PR #4944, opened 2019-08-12
Okhttp bug fix

g2vinay:


sdk/keyvault/microsoft-azure-keyvault/src/main/java/com/microsoft/azure/keyvault/authentication/KeyVaultCredentials.java
```diff
@@ -68,12 +68,12 @@ public Response intercept(Chain chain) throws IOException {
...
 
-                    if (response.code() == 200) {
+                    if (response.code() !=  401) {
                         return response;
-                    } else if (response.code() != 401) {
-                        throw new IOException("Unexpected unauthorized response.");
                     }
+
                     authenticatedRequestPair = buildAuthenticatedRequest(originalRequest, response);
+                    response.close();
```
anuchandy: seems `buildAuthenticatedRequest` can throw exception if so `response.close();` won't be hit and that can leak connection.

g2vinay: Valid point.
Added a try finally block to ensure response gets cleaned up via close().

PR #4963, opened 2019-08-13
Enabling javadoc production for template.

mitchdenny:
Working on getting the unified pipeline going for the template project so we can use it for validating the deployment pipeline end-to-end.

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -735,7 +735,8 @@
       <id>template-module</id>
       <activation>
         <property>
-          <name>include-template</name>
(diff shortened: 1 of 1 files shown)


JonathanGiles:
I'd rather this be done in a separate profile if possible, so that our published JavaDocs don't include the template files in them.

mitchdenny:
> I'd rather this be done in a separate profile if possible, so that our published JavaDocs don't include the template files in them.

Adjusted this its similar to what we had before, but its driven by a variable that I set on the ```java - template - ci``` (public) and ```java - template``` (private)  pipelines.

PR #4964, opened 2019-08-13
Disable Jetty by default.

mitchdenny:


sdk/keyvault/pom.service.xml
```diff
@@ -9,22 +9,10 @@
   <packaging>pom</packaging>
   <version>1.0.0</version><!-- Need not change for every release-->
   <modules>
-
-    <!-- Track 1 -->
```
weshaggard: What is the plan to re-enable track 1?

mitchdenny: We discussed a temporary convention recently. Let's see if we can capture it here:

1. Introduce a ```ci.data.yml``` file which sits along side the existing ```ci.yml``` file.
2. This file would turn on Jetty and specify the different set of artifacts.
3. Introduce profiles into ```pom.service.xml``` so that we can build the track 1 libraries or the track 2 libraries through the same aggregate.
4. Create a pipeline called ```java - [service] - ci.data``` (public) and ```java - [service] - data``` (internal).

mitchdenny: This would only apply when:

1. We have track 1 and track 2 libraries where we want to/need to seperate them.
2. Can't deprecate the need to build and release the track 1 libraries.

The preference will be to:

1. Co exist in the same pipeline
2. Disable track 1 builds if we don't need to build them anymore.
3. Setup a pipeline as described above.

Longer term we'll be removing these track 1 pipelines.

weshaggard: OK in this case then given both set of code still exists in the repo I'd go ahead and setup the track 1 CI as well so that we don't have folks submitting changes to it without any validation.

JonathanGiles:
Can't wait to get this merged!

JonathanGiles:
Those CI times are super nice.

mitchdenny:
I think that this PR is ready to go. When it is merged all pipelines that include track 1 and track 2 libraries will continue to build as usual (including starting Jetty). Then for libraries which have no track 1 elements we can simply set the EnableJetty flag to false and Jetty will disappear from the build times.

For libraries that contain track 1 and track 2 libraries work needs to be done on a per library basis to split out the track 1 libraries into a separate build. This PR contains an example of how this is done for KeyVault (using profiles).

The template yml is setup so that Jetty will start unless it is specifically disabled, but the approach to the ```pom.service.xml``` is such that if ```EnableJetty``` is missing or false, only the track 2 components will build. So for the split out track 1 pipelines it is mandatory that ```EnableJetty``` setting is set to true so it all functions correctly.

mitchdenny:
Spoke too soon - looks like the track 1 build is failing. Not sure why.

mitchdenny:
OK - it looks like some recorded tests didn't come across in the move. Will need to go back and find them and put them in the right place.

sdk/keyvault/ci.yml
```diff
@@ -33,6 +33,8 @@ pr:
     exclude:
       - sdk/keyvault/microsoft-azure
 
+
+
```
g2vinay: whitespace not needed.

sdk/keyvault/ci.yml
```diff
@@ -33,6 +33,8 @@ pr:
...
 stages:
   - template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
```
g2vinay: Does the CI auto detect which tests to run (Track 1 or Track 2) based on code changes ?

sdk/keyvault/ci.yml
mitchdenny: Was needed to do some triggering ;) But will remove now.

sdk/keyvault/ci.yml
mitchdenny: I'll be doing some further tweaks to the path filters.

mitchdenny: We have two pipeline definitions and two yaml files. Each pipeline runs the appropriate set of tests. If you change files in both track 1 and track 2 -- two separate pipelines will run but I expect this to be the rare case. In fact the track 1 builds I think will be temporary.

PR #4965, opened 2019-08-13
Bumped version of jackon to 2.9.9.3 because of CVE.

mitchdenny:
This PR increments the version of ```jackson-databind``` to ```2.9.9.3``` because of a CVE. Not all of the ```jackson-databind``` libraries got an update which has a result means that the property that existed previously needed to get split.

Perhaps a larger question here is whether we should continue setting the dependency versions in the parent POMs here given the direction we want to go with a ```versions.txt``` file.

Whilst this PR will stop GitHub and CG screaming at us, I think we need to do some follow-up work and remove all of the parent POM inherited properties from the version specifiers in the children.

This would also mean that tools like Dependabot would do a better job submitting update PRs.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -6,7 +6,7 @@
   <groupId>com.azure</groupId>
   <artifactId>azure-sdk-parent</artifactId>
   <packaging>pom</packaging>
-  <version>1.2.1</version>
+  <version>1.3.0</version>
 
   <name>Microsoft Azure SDK for Java Parent</name>
   <description>Parent POM for Microsoft Azure SDK for Java</description>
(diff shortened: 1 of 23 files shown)


parent/pom.xml
```diff
@@ -100,6 +100,7 @@
 
     <!-- Dependency Versions -->
     <jackson.version>2.9.9</jackson.version>
```
weshaggard: I think we also need to update this version to at least 2.9.9.2. 

See https://dev.azure.com/azure-sdk/internal/_componentGovernance/106501/alert/1778976?typeId=390820

weshaggard: Never mind I see that CG issue is only for databind.

PR #4968, opened 2019-08-13
Update parent/pom.xml to 1.3.0

conniey:
Noticed that parent/pom.xml was 1.2.0. This updates it in case we forget to bump the version before next release.


parent/pom.xml
```diff
@@ -6,7 +6,7 @@
   <groupId>com.azure</groupId>
   <artifactId>azure-sdk-parent</artifactId>
   <packaging>pom</packaging>
-  <version>1.2.0</version>
```
srnagar: Any reason to bump this up if there are no changes to the parent pom? We should increment the version only if we make any changes to this file.

mitchdenny: No - we shouldn't update parent POMs that frequently, however we have a jackson databind CVE open at the moment that will require this POM file to be modified. So I suspect we are going to be pushing a new release soon anyway!

conniey:
Defer updating pom until someone updates the version.

Issue #4713, opened 2019-07-30
Standardize Testing Framework


alzimmermsft:
Use a common testing framework across Blobs, Files, Queues.

joshfree:
I'm assuming this is tracking converting some Storage java/junit tests to groovy/spock?

PR #4971, opened 2019-08-13
Playback and Record for Azure Storage Blob

alzimmermsft:
Fixes part of #4713 

**Note: ~1000 of the files are for playback, those can be ignored**

- Creates a playback and record framework for Azure Storage Blobs (uses Groovy and Spock for unit testing).
- Adds playback records (will be added after draft review).
- Updates Azure Core Test to fix edge cases and extend functionality needed by storage tests.
- Adds `NetworkCallException` to handle serialization and deserialization of exceptions thrown in the pipeline during playback and record.
- Adds recording handling for pipeline exceptions.
- Default serialization of the response body no longer assumes compression, fixes exception thrown by record policy.
- Adds OffsetDateTime creation in `TestResourceNamer`.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
@@ -1114,7 +1114,7 @@ class BlobAPITest extends APISpec {
             metadata.put(key2, value2)
         }
 
-        String status = bu2.startCopyFromURLWithResponse(bc.getBlobUrl(), metadata, null, null, null, null)
+        def status = bu2.startCopyFromURLWithResponse(bc.getBlobUrl(), metadata, null, null, null, null)
             .headers().value("x-ms-copy-status")
 
         OffsetDateTime start = OffsetDateTime.now()


conniey:
I started to look at this, but the diff is 1024files, I don't know if I can manage that many. 😄

sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
```diff
@@ -104,40 +126,77 @@ public RecordNetworkCallPolicy(RecordedData recordedData) {
...
             });
         }
     }
+
+    private String redactUserDelegationKey(String content) {
+        if (!content.contains("UserDelegationKey")) {
+            return content;
+        }
+
+        content = redactionReplacement(content, DELEGATIONKEY_KEY_PATTERN.matcher(content), Base64.getEncoder().encodeToString("REDACTED".getBytes(StandardCharsets.UTF_8)));
```
sima-zhu: Since you have the helper method, it is better to simplify the second param to string

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -96,222 +94,345 @@ class APISpec extends Specification {
...
+    static SharedKeyCredential alternateCredential
+    static SharedKeyCredential blobCredential
+    static SharedKeyCredential premiumCredential
+    static TestMode testMode
 
-    /*
-    URLs to various kinds of accounts.
-     */
-    BlobServiceClient primaryServiceURL
+    BlobServiceClient primaryServiceClient
```
sima-zhu: I'd suggest we use the same way to define the variable. Xstore team suggests to use **def** as much as possible.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -96,222 +94,345 @@ class APISpec extends Specification {
...
-        return ctx.getCurrentFeature().name.replace(' ', '').toLowerCase()
-    }
+        primaryServiceClient = setClient(primaryCredential)
+        alternateServiceClient = setClient(alternateCredential)
+        blobStorageServiceClient = setClient(blobCredential)
+        premiumServiceClient = setClient(premiumCredential)
 
-    def generateContainerName() {
-        generateContainerName(specificationContext, iterationNo, entityNo++)
+        cu = primaryServiceClient.getContainerClient(generateContainerName())
```
sima-zhu: cu -> containerClient

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -96,222 +94,345 @@ class APISpec extends Specification {
...
+        @Override
+        ProxyOptions get() {
+            return new ProxyOptions(ProxyOptions.Type.HTTP, new InetSocketAddress("localhost", 8888))
+        }
+    }
 
-        primaryServiceURL = getGenericServiceURL(primaryCreds)
-        cu = primaryServiceURL.getContainerClient(containerName)
-        cu.create()
+    def generateContainerName() {
```
sima-zhu: Is this still in use?

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
```diff
@@ -141,7 +160,12 @@ private String applyReplacementRule(String text) {
     }
 
     private static String removeHost(String url) {
-        URI uri = URI.create(url);
-        return String.format("%s?%s", uri.getPath(), uri.getQuery());
+        UrlBuilder urlBuilder = UrlBuilder.parse(url);
+
+        if (urlBuilder.query().containsKey("sig")) {
```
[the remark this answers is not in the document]
alzimmermsft: It's useful to keep the rest of the SAS token the same as that is what is being matched in the playback client, it could help catch scenarios where we accidentally break the generation of it.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
alzimmermsft: I'd prefer not to use `def` here as it cannot infer type as it is assigned during runtime and it makes it much harder to program as IntelliSense is lost.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public NetworkCallError(Throwable throwable) {
+        this.throwable = throwable;
+        this.className = throwable.getClass().getName();
+    }
+
+    /**
+     * @return the thrown throwable as the class it was thrown as by converting is using its class name.
+     */
+    public Throwable get() {
```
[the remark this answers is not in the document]
alzimmermsft: Not entirely certain what you mean by this, do you want it to return `<T extends Throwable> T`?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
alzimmermsft: Yes, it's used in 30 places based on IntelliJ's find all references.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
alzimmermsft: I think James is changing this in another PR and I don't want to cause a ton of tiny merge conflicts.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
alzimmermsft: I'd rather handle the encoding here as the other redactions don't require an encoding.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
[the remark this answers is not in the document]
sima-zhu: Talked offline. Will make the exception class generic.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/SASTest.groovy
```diff
@@ -669,9 +610,9 @@ class SASTest extends APISpec {
         e.getMessage().contains(parameter)
 
         where:
-        version | creds       || parameter
-        null    | primaryCreds | "version"
-        "v"     | null         | "sharedKeyCredentials"
+        version | creds             || parameter
```
sima-zhu: I guess the table format of `||` is typo right here. Can you make a singe `|`?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlockBlobAPITest.groovy
```diff
@@ -652,9 +651,9 @@ class BlockBlobAPITest extends APISpec {
         validateBlobProperties(response, cacheControl, contentDisposition, contentEncoding, contentLanguage, contentMD5, contentType)
 
         where:
-        cacheControl | contentDisposition | contentEncoding | contentLanguage | contentMD5                                                   | contentType
-        null         | null               | null            | null            | null                                                         | null
-        "control"    | "disposition"      | "encoding"      | "language"      | MessageDigest.getInstance("MD5").digest(defaultData.array()) | "type"
+        cacheControl | contentDisposition | contentEncoding | contentLanguage | contentMD5                                                                | contentType
```
sima-zhu: The format is weird. Remove extra space

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobOutputStreamTest.groovy
```diff
@@ -0,0 +1,74 @@
+package com.azure.storage.blob
+
+import com.azure.storage.common.Constants
+import spock.lang.Ignore
+
+class BlobOutputStreamTest extends APISpec {
+    private static int FOUR_MB = 4 * Constants.MB
+
+    @Ignore
```
sima-zhu: When do you want them run?
Could you make them run in live or leave the reason of ignore.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
```diff
@@ -682,7 +660,9 @@ class BlobAPITest extends APISpec {
         setup:
         String leaseID = setupBlobLeaseCondition(bu, receivedLeaseID)
 
-        Thread.sleep(16000) // Wait for the lease to expire to ensure we are actually renewing it
+        // If running in live mode wait for the lease to expire to ensure we are actually renewing it
+        sleepIfRecord(16000)
```
sima-zhu: It seems a long time to run. Can live test finish in 60 mins with the time span like this?
Might need to ignore in live tests

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
```diff
@@ -121,7 +124,20 @@ public HttpClient port(int i) {
                 }
             }
 
-            bytes = rawBody.getBytes(StandardCharsets.UTF_8);
+            String contentType = networkCallRecord.response().get("Content-Type");
+
+            // octet-stream's are written to disk using Arrays.toString() which creates an output such as "[12, -1]".
+            if (contentType != null && contentType.contains("octet-stream")) {
```
conniey: "application/octet-stream" mime-type is well known. I'd use `.equalsIgnoreCase()` because they are also case-insensitive.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -95,222 +93,345 @@ class APISpec extends Specification {
...
+        if (testMode == TestMode.RECORD) {
+            builder.addPolicy(interceptorManager.getRecordPolicy())
         }
 
-        try {
-            blobStorageServiceURL = getGenericServiceURL(getGenericCreds("BLOB_STORAGE_"))
+        return builder.credential(credential).buildBlobClient()
+    }
+
+    BlobClient getBlobClient(SharedKeyCredential credential, String endpoint, String blobName) {
```
sima-zhu: Since you have several different clients depend on same builder. Could you make the helper method for the builder instead of client?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -437,83 +555,47 @@ class APISpec extends Specification {
...
-    This method returns a stub of an HttpResponse. This is for when we want to test policies in isolation but don't care
-     about the status code, so we stub a response that always returns a given value for the status code. We never care
-     about the number or nature of interactions with this stub.
-     */
-
-    def getStubResponse(int code) {
-        return Stub(HttpResponse) {
-            statusCode() >> code
+    // Only sleep if test is running in live mode
+    def sleepIfRecord(long milliseconds) {
```
sima-zhu: Why you take million sec here? Can we take second?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
sima-zhu: I thought the value is second. Ignore the comments here

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobOutputStreamTest.groovy
alzimmermsft: These were always ignored, when they were in a JUnit test class the `@Test` was commented out.

Will look to enable these in a follow-up PR.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/DownloadResponseMockFlux.java
```diff
@@ -31,15 +31,16 @@
...
+
         switch (this.scenario) {
             case DR_TEST_SCENARIO_SUCCESSFUL_ONE_CHUNK:
-                this.scenarioData = APISpec.getRandomData(512 * 1024);
+                this.scenarioData = apiSpec.getRandomData(512 * 1024);
                 break;
             case DR_TEST_SCENARIO_SUCCESSFUL_MULTI_CHUNK:
             case DR_TEST_SCENARIO_SUCCESSFUL_STREAM_FAILURES:
-                this.scenarioData = APISpec.getRandomData(1024);
+                this.scenarioData = apiSpec.getRandomData(1024);
```
sima-zhu: I think the static method is enough. It is weird to take in the APISpec

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
alzimmermsft: `Thread.sleep` is in milliseconds, running the blob tests in live mode doesn't come close to hitting the 60 minute timeout.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
alzimmermsft: That is what `Thread.sleep` takes.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
```diff
@@ -102,40 +123,77 @@ public RecordNetworkCallPolicy(RecordedData recordedData) {
...
         }
 
-        String contentType = response.headerValue("content-type");
+        String contentType = response.headerValue("Content-Type");
         if (contentType == null) {
             return Mono.just(responseData);
-        } else if (contentType.contains("json") || response.headerValue("content-encoding") == null) {
-            return response.bodyAsString().switchIfEmpty(Mono.just("")).map(content -> {
-                responseData.put("Body", content);
+        } else if (contentType.contains("octet-stream")) {
```
conniey: There are a lot of special string literals "octet-stream", "Body", "Content-Encoding". It would be nice to consolidate these in either a constants file or at the top of the class so we don't miss updating one.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/DownloadResponseMockFlux.java
alzimmermsft: The method can't be static as it requires test context to properly record the data.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
conniey: Same comment about application/octet-stream

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
conniey: in the RFC, headers are case insensitive. It is by convention that you see "Content-Type" rather than "content-type". But they amount to the same thing.

sima-zhu: Playback client is not validator. Is it better to ignore the case for all the hard-code string checking?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/SASTest.groovy
alzimmermsft: Spock allows for `||` in the data table, I don't think it causes any differences in the test run but its mean to be a visual indicator of what the test is compared to in the end.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
```diff
@@ -850,9 +830,9 @@ class BlobAPITest extends APISpec {
 
         where:
         leaseTime | breakPeriod | remainingTime
-        -1        | null        | 0
-        -1        | 20          | 25
-        20        | 15          | 16
+            -1        | null        | 0
```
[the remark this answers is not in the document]
alzimmermsft: Fixed

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
conniey: RecordNetworkCallPolicy encounters the same issues. I added a comment about that.

alzimmermsft: While RFC headers are case-insensitive the data structure, `Map`, we use to store this information in playback isn't. This is another cleanup we should do to either use `HttpHeaders` or a case-insensitive `Map` implementation.

conniey: SO has revealed that there is a way to have case insensitive key comparison. One thing that .NET provides over HashMap.
https://stackoverflow.com/a/22336599

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
alzimmermsft: Will look into this change after the PR goes through.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
alzimmermsft: Updated to the case ignores check.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
alzimmermsft: Updated

sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
sima-zhu: I mean you can have the signature like
redactionReplacement(String content, String regex, String replacement)
Not a big fan of method call in signature. Hard to log and debug. The third param  makes sense to me as some key might use other way to encode.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/DownloadResponseMockFlux.java
sima-zhu: Not a big blocker for this. Might be better to add the randomData method in TestRandomNamer

PR #4978, opened 2019-08-13
[ServiceBus] Update SDK version to 3.0.0

bainian12345:
Update SDK version to 3.0.0 for RBAC breaking changes

### DIFF
--- a/sdk/servicebus/microsoft-azure-servicebus/pom.xml
+++ b/sdk/servicebus/microsoft-azure-servicebus/pom.xml
@@ -6,7 +6,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.microsoft.azure</groupId>
   <artifactId>azure-servicebus</artifactId>
-  <version>2.0.0</version>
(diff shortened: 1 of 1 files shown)


mitchdenny:
This means that you won't have ever published a 2.0.0 version (other than previews) - is that intentional?

bainian12345:
There seems to be a 2.0.0 version released on 15-march-2019 isn't there?
[[https://search.maven.org/search?q=g:com.microsoft.azure%20AND%20a:azure-servicebus&core=gav](https://search.maven.org/search?q=g:com.microsoft.azure%20AND%20a:azure-servicebus&core=gav)](https://search.maven.org/search?q=g:com.microsoft.azure%20AND%20a:azure-servicebus&core=gav)

---
In reply to: [521100029](https://github.com/Azure/azure-sdk-for-java/pull/4978#issuecomment-521100029) [](ancestors = 521100029)

mitchdenny:
My bad - so there is.

Issue #4980, opened 2019-08-14
Miss one storage error code "AuthenticationError"


sima-zhu:
Storage queue returns StorageErrorException of "AuthenticationError" when we passed in the metadata with capital letter.

There is no such constant error code in swagger in java.

sima-zhu:
@rickle-msft @tg-msft

tg-msft:
You'll want to add this to https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-queue/swagger/README.md with a transform like:

```yaml
directive:
- from: swagger-document
  where: $.definitions.ErrorCode
  transform: >
    if (!$.includes("AuthenticationError")) {
        $.push("AuthenticationError");
    }
```

And then regenerate following the instructions at the top of that file.

kurtzeborn:
@alzimmermsft, you were suspicious this is not the error return they actually use here.  Please take a look.

alzimmermsft:
Closing as this isn't an expected response code from the service based on their documentation.

https://docs.microsoft.com/en-us/rest/api/storageservices/common-rest-api-error-codes
https://docs.microsoft.com/en-us/rest/api/storageservices/queue-service-error-codes

PR #4981, opened 2019-08-14
Update Storage swagger

tg-msft:
There are two commits in this PR that might be easier to review individually.

The first changes to the official versions of the Storage swagger and adds README.md transformations address any deltas.  I've regenerated and there was only one code change I couldn't trace back to the swagger which I'll call out below.

The second commit updates blobs to support STG69 features.  This will require the feature crew to take over this PR and update any calls to the protocol layer that added additional parameters _(I assume null will be fine to get started)_.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
+++ b/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
@@ -66,7 +66,7 @@
       <Class name="com.microsoft.azure.keyvault.authentication.KeyVaultCredentials"/>
       <Class name="com.azure.storage.blob.implementation.AzureBlobStorageBuilder"/>
       <Class name="com.azure.storage.file.implementation.AzureFileStorageBuilder"/>
-      <Class name="com.azure.storage.file.implementation.AzureQueueStorageBuilder"/>
+      <Class name="com.azure.storage.queue.implementation.AzureQueueStorageBuilder"/>
     </Or>
     <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE"/>
   </Match>


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobHierarchyListSegment.java
```diff
@@ -6,17 +6,14 @@
...
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
-
 import java.util.ArrayList;
 import java.util.List;
 
 /**
  * The BlobHierarchyListSegment model.
  */
 @JacksonXmlRootElement(localName = "Blobs")
-@JsonDeserialize(using = CustomHierarchicalListingDeserializer.class)
```
tg-msft: This is the only change in the first commit that wasn't just moving things around and I'm not sure what in the swagger changed to stop generating it.

sima-zhu:
https://github.com/Azure/azure-sdk-for-java/issues/4980

I created an issue for missing error constant in StorageErrorCode. The issue only exists in Storage queue. Could you please take a look? It might something wrong with the service or missing the error code in swagger.
I have a PR manually added them in.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobHierarchyListSegment.java
alzimmermsft: This was added back manually last I checked, @jianghaolu is that correct?

rickle-msft: @alzimmermsft correct. This has never been generated; it's always been added manually. We have to add it back every time we generate. I don't know if I could find it, but I'm pretty sure I filed an issue against autorest.java to try to get the deserialization fixed a while back

tg-msft: Okay - thanks.  I feel less crazy now.  😄 

@alzimmermsft - can you both add this back and then add a note in the README.md for Blobs to add this back in after regenerating?  We might be able to address this with a regex at some point, but I don't want to block on getting that working now.

sdk/storage/azure-storage-blob/swagger/README.md
```diff
@@ -0,0 +1,849 @@
...
+        "description": "UTC date/time value generated by the service that identifies the version of the blob. This header is returned for requests made against version 2018-11-09 and above."
+      };
+    }
+```
+
+### /{containerName}/{blob}?comp=copy&sync
+``` yaml
+directive:
+- from: swagger-document
+  where: $["x-ms-paths"]["/{containerName}/{blob}?comp=copy&sync"]
```
weshaggard: Can't this be combined with the comp=copy version directly above this?

tg-msft: I'm not aware of a way to do that.

weshaggard: I thought the where clauses were json path syntax and if so I'd expect something like `/{containerName}/{blob}?comp=copy*` to work. Although I'm no expert in json path or these autorest directives so perhaps that is just wishful thinking.

tg-msft: Maybe?  [The only documentation I've found](https://github.com/Azure/autorest/tree/master/Samples/3b-custom-transformations) doesn't mention that explicitly, but it does look like JSON Path syntax.  I don't think you can use a wildcard `*` inside a string like that, but there's probably a way using those arcane JSON Path filter expressions.  I don't want to block progress on getting this work done though so we can explore that after this PR.

tg-msft:
Nice, these changes look great.

@rickle-msft -- mind taking a look at Alan's commits in particular to CR the way he's changing the clients to call the new 2019-02-02 protocol layer?

alzimmermsft:
#5001 Calls out the changes that new features around Content-MD5 and x-ms-content-crc64 need to be handled as well as other potential changes that need to be handled.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobHierarchyListSegment.java
alzimmermsft: Added it back and added a comment in the README about this issue.

jianghaolu: This is because Storage is returning something like

```xml
<BlobItems>
<BlobPrefix>prefix1</BlobPrefix>
<Blob>blob1</Blob>
<BlobPrefix>prefix2</BlobPrefix>
<Blob>blob2</Blob>
</BlobItems>
```

And without this annotation blob prefix list will only have `prefix2` and blob list will only have `blob2`. Assuming this XML format is correct, I'm wondering if there's a way to represent this in Swagger. If we can represent it in Swagger, then I can conditionally generate this annotation in Java code.

PR #4982, opened 2019-08-14
Optimize checkout

mitchdenny:
This PR introduces explicit checkout steps to the pipeline to optimize the pipelines. Where possible we will eliminate checkout.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -58,7 +58,7 @@ jobs:
     steps:
       - checkout: self
         fetchDepth: 1
-        
+
       - script: |
           echo "##vso[build.addbuildtag]Scheduled"
         displayName: 'Tag scheduled builds'
@@ -90,6 +90,9 @@ jobs:
       vmImage: 'ubuntu-16.04'
 
     steps:
+      - checkout: self
+        fetchDepth: 1
+        
       - task: UsePythonVersion@0
         displayName: 'Use Python 3.6'
         inputs:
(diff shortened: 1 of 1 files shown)


JonathanGiles:
Checkout is much faster across the board now - looks good to me.

JonathanGiles:
It does appear that all branches are being cloned - is it worth just configuring it to clone master only?

mitchdenny:
Yep - a fair chunk of time is spent on the server building a set of pack files to transfer to the client and I was wondering whether we could optimize by having a greater fetch depth to trick GitHub into reusing pack files but its probably diminishing returns - and the gh-pages thing will have a bigger impact anyway.

mitchdenny:
> It does appear that all branches are being cloned - is it worth just configuring it to clone master only?

If I can avoid falling out to a manual clone in a script command I would like to. I want to drop the gh-pages content because I think that will have a much larger impact in the volume of data being brought down. The extra refs for the various branches don't add much weight I don't think.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -56,6 +56,9 @@ jobs:
           JavaVersion: '1.11'
 
     steps:
+      - checkout: self
```
weshaggard: Can this not be setup at a higher level to avoid duplication?

mitchdenny: There is a variable that can be set ```Agent.Source.Git.ShallowFetchDepth``` which does it globally. I'll do another PR which takes this route.

PR #4988, opened 2019-08-14
Seeing if we can speed up spotbugs

### LINKED ISSUE #4974: Evaluate options to make spotbugs run faster - start with "effort"
Spotbugs may soon become the longest pole in building Track 2 client libraries. This task is to look into "effort" for spotbugs and tune it to improve build times.  https://spotbugs.readthedocs.io/en/stable/effort.html

JonathanGiles:
by lowering the effort and not forking the JVM for each module.

Fixes #4974

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -164,7 +164,7 @@
           </dependency>
         </dependencies>
         <configuration>
-          <effort>min</effort>
+          <effort>max</effort>
           <threshold>Low</threshold>
           <xmlOutput>true</xmlOutput>
           <spotbugsXmlOutputDirectory>${project.build.directory}/spotbugs</spotbugsXmlOutputDirectory>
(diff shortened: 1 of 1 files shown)


JonathanGiles:
This shaves about 25 seconds off of the Analyze job, so I think it is worth merging.

pom.client.xml
```diff
@@ -164,12 +164,13 @@
           </dependency>
         </dependencies>
         <configuration>
-          <effort>Max</effort>
+          <effort>min</effort>
```
srnagar: Maybe we should run nightly builds with `max` or `more` and publish the [results here](https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-java/staging/spotbugsXml.html#{%22tab%22:%22summary%22,%22includeFixed%22:false,%22release%22:-1,%22priority%22:4})

Min does very little and may not catch some issues.
https://spotbugs.readthedocs.io/en/stable/effort.html

JonathanGiles: My expectation is even at min we should get everything we want.

JonathanGiles: But I should add, there was almost no difference between min and max on my machine. The only win was from not forking the JVM for each module, so I am also happy to return this to max.

PR #4996, opened 2019-08-14
Increasing clone depth because some builds are failing.

mitchdenny:


weshaggard:
Do we understand why some builds are failing because of depth 1?

mitchdenny:
> Do we understand why some builds are failing because of depth 1?

My understanding (based on the [PR that first introduced shallow fetching](https://github.com/microsoft/azure-pipelines-agent/pull/620/files)) is that it is some kind of race condition where the pack that Git prepares for the client doesn't contain the commit that you end up wanting to check out. I'm not sure why it happens but I recall that increasing the depth seems to mitigate the issue.

mitchdenny:
I'm closing this down. Looks like it might have just been a glitch it hasn't been happening since then.

PR #5002, opened 2019-08-15
Renaming IterableResponse to IterableStream

JonathanGiles:


sdk/core/azure-core/src/main/java/com/azure/core/util/IterableStream.java
```diff
@@ -1,39 +1,40 @@
...
 
 /**
- * This class provides utility to iterate over values. All the values are preserved even if they are traversed multiple times.
+ * This class provides utility to iterate over values using standard 'for-each' style loops, or to convert them into a
+ * Stream and operate in that fashion. All the values are preserved even if they are traversed multiple times.
  *
  * <p><strong>Code sample using Stream</strong></p>
  *
- * {@codesnippet com.azure.core.http.rest.iterableResponse.stream}
+ * {@codesnippet com.azure.core.util.iterableStream.stream}
```
conniey: Glad you have codesnippets.. in the UX study, they had troubles with this class.

PR #3631, opened 2019-05-13
Add support for user proxy configuration

conniey:
* Adds support for Event Hubs users to specify proxy configuration.

Dependent on: https://github.com/Azure/qpid-proton-j-extensions/pull/18

PR #5019, opened 2019-08-16
Adding support for proxy configuration in EventProcessorHost

conniey:
### Summary
* This builds on top of PR #3631 (which adds support for ProxyConfiguration in the low level Track 1 Event Hubs client library).
  * The diff is less daunting if you see the comparison to the parent branch: 
     https://github.com/conniey/azure-sdk-for-java/compare/proxyConfiguration...conniey:proxyConfiguration-eph
* Adds proxy configuration support into EventProcessorHost

### Dependencies
- [x] Merge #3631 
- [ ] Release of azure-storage 8.3.0 (which is a private drop atm)

### TODO
- [ ] Test end-to-end scenario with a proxy.
- [ ] Add unit tests.
- [ ] Investigate integration tests.


parent/pom.xml
```diff
@@ -105,7 +105,7 @@
     <spotbugs.version>4.0.0-beta3</spotbugs.version>
     <spotbugs.maven.version>3.1.12</spotbugs.maven.version>
     <proton-j-version>0.31.0</proton-j-version>
-    <qpid-proton-j-extensions-version>1.2.0</qpid-proton-j-extensions-version>
+    <qpid-proton-j-extensions-version>1.2.1</qpid-proton-j-extensions-version>
```
srnagar: Since we making changes to the versions here, we should increment the parent version too and update the modules that would need these new versions.

sdk/eventhubs/microsoft-azure-eventhubs/src/main/java/com/microsoft/azure/eventhubs/ProxyConfiguration.java
```diff
@@ -0,0 +1,140 @@
...
+package com.microsoft.azure.eventhubs;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.PasswordAuthentication;
+import java.util.Arrays;
+import java.util.Objects;
+
+public class ProxyConfiguration implements AutoCloseable {
```
srnagar: javadoc

sdk/eventhubs/microsoft-azure-eventhubs/src/main/java/com/microsoft/azure/eventhubs/ProxyConfiguration.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+
+    /**
+     * Gets the proxy address.
+     *
+     * @return The proxy address. Returns {@code null} if user creates proxy credentials with
+     * {@link ProxyConfiguration#SYSTEM_DEFAULTS}.
+     */
+    public java.net.Proxy proxyAddress() {
```
srnagar: Import package instead.

sdk/eventhubs/microsoft-azure-eventhubs/src/main/java/com/microsoft/azure/eventhubs/ProxyConfiguration.java
```diff
@@ -0,0 +1,140 @@
...
+        return proxyAddress;
+    }
+
+    /**
+     * Gets credentials to authenticate against proxy with.
+     *
+     * @return The credentials to authenticate against proxy with. Returns {@code null} if no credentials were set. This
+     * occurs when user uses {@link ProxyConfiguration#SYSTEM_DEFAULTS}.
+     */
+    public PasswordAuthentication credentials() {
```
srnagar: Track 1 library uses `get` and `set` prefix.

sdk/eventhubs/microsoft-azure-eventhubs/src/main/java/com/microsoft/azure/eventhubs/ProxyConfiguration.java
```diff
@@ -0,0 +1,140 @@
...
+     */
+    public boolean isProxyAddressConfigured() {
+        return proxyAddress != null && proxyAddress.address() != null;
+    }
+
+    @Override
+    public void close() {
+        // It is up to us to clear the password field when we are done using it.
+        if (credentials != null) {
+            Arrays.fill(credentials.getPassword(), '\0');
```
srnagar: Can you just use `null` instead of `'\0'`?

conniey:
Closing for now because storage does not support multiple proxy configurations at the same time.

PR #5028, opened 2019-08-17
Update Storage Files swagger to 2019-02-02

tg-msft:
Also adding a transform to automatically add the serialization attribute back on hierarchical blobs when regenerating.

I'd appreciate if @alzimmermsft could take over this PR and get it building/passing tests like the other swagger change.  (You can see the tweaks I made to get .NET working at https://github.com/Azure/azure-sdk-for-net/pull/7319)

### DIFF
--- a/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/implementation/AzureFileStorageBuilder.java
+++ b/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/implementation/AzureFileStorageBuilder.java
@@ -30,22 +30,6 @@ public final class AzureFileStorageBuilder {
     }
 
     /*
-     * Only update is supported: - Update: Writes the bytes downloaded from the source url into the specified range.
-     */
-    private String fileRangeWriteFromUrl;
-
-    /**
-     * Sets Only update is supported: - Update: Writes the bytes downloaded from the source url into the specified range.
-     *
-     * @param fileRangeWriteFromUrl the fileRangeWriteFromUrl value.
-     * @return the AzureFileStorageBuilder.
-     */
-    public AzureFileStorageBuilder fileRangeWriteFromUrl(String fileRangeWriteFromUrl) {
-        this.fileRangeWriteFromUrl = fileRangeWriteFromUrl;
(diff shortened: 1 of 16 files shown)


sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/implementation/AzureFileStorageBuilder.java
```diff
@@ -29,6 +29,22 @@ public AzureFileStorageBuilder version(String version) {
         return this;
     }
 
+    /*
+     * Only update is supported: - Update: Writes the bytes downloaded from the source url into the specified range.
+     */
```
rickle-msft: Why is this on the builder. This seems like a parameter that should be passed into an api

alzimmermsft: @tg-msft Do you have more insight on what this gets used for?

tg-msft: It came with the new swagger:

```json
        "FileRangeWriteFromUrl": {
            "name": "x-ms-write",
            "x-ms-client-name": "fileRangeWriteFromUrl",
            "in": "header",
            "description": "Only update is supported: - Update: Writes the bytes downloaded from the source url into the specified range.",
            "required": true,
            "type": "string",
            "enum": [
                "update"
            ],
            "default": "update",
            "x-ms-enum": {
                "name": "FileRangeWriteFromUrlType",
                "modelAsString": false
            }
        },
```

rickle-msft: I am pretty sure that comes from [this api](https://docs.microsoft.com/en-us/rest/api/storageservices/put-range). In normal write operations, it's used to specify whether the operation is to update the data or clear the range. When specifying a source url (hence the "fromURL", the only viable option is update.

This should just be an internal constant that is used for the writeRangeFromURL method.

PR #5034, opened 2019-08-19
Event Processor load balancer

srnagar:
This is a draft PR for load balancing Event Processors as per the algorithm outlined here - https://gist.github.com/srnagar/7ef8566cfef0673288275c450dc99590

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancerTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancerTest.java
@@ -93,9 +93,9 @@ public class PartitionBasedLoadBalancerTest {
         });
     }
 
-    void sleep(int i) {
+    void sleep(int secondsToSleep) {
         try {
-            TimeUnit.SECONDS.sleep(i);
+            TimeUnit.SECONDS.sleep(secondsToSleep);
         } catch (InterruptedException ex) {
 
         }
@@ -241,7 +241,7 @@ public class PartitionBasedLoadBalancerTest {
             }
         });
 
-        sleep(3);
+        sleep(6);
         partitionOwnership = partitionManager.listOwnership(eventHubName,
             consumerGroupName).collectList().block();
 
@@ -300,7 +300,7 @@ public class PartitionBasedLoadBalancerTest {
             eventHubAsyncClient, eventHubName, consumerGroupName, "owner", TimeUnit.SECONDS.toSeconds(5),
             partitionPumpManager);
         loadBalancer.loadBalance();
-        sleep(2);
+        sleep(5);
         verify(eventHubAsyncClient, atLeast(1)).getPartitionIds();
         verify(eventHubAsyncClient, never()).createConsumer(any(), any(), any());
         verify(eventHubConsumer, never()).receive();


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -4,31 +4,22 @@
...
 import reactor.core.scheduler.Scheduler;
 import reactor.core.scheduler.Schedulers;
 
 /**
  * This is the starting point for event processor.
  * <p>
- * Event Processor based application consists of one or more instances of {@link EventProcessor} which are
- * set up to consume events from the same Event Hub + consumer group and to balance the workload across different
- * instances and track progress when events are processed.
+ * Event Processor based application consists of one or more instances of {@link EventProcessor} which are set up to
```
jsquire: nit:  Missing hyphen.   `Event Processor-based`

jsquire: nit:  This reads a bit awkward to me.  Maybe consider something like "... which have been configured to consume events..."

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -43,18 +34,13 @@
 
     private static final long INTERVAL_IN_SECONDS = 10; // run every 10 seconds
```
jsquire: nit:  Run what?   The comment doesn't provide context for how this is applied.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -64,28 +50,22 @@
      * @param eventHubAsyncClient The {@link EventHubAsyncClient}.
      * @param consumerGroupName The consumer group name used in this event processor to consumer events.
      * @param partitionProcessorFactory The factory to create new partition processor(s).
-     * @param initialEventPosition Initial event position to start consuming events.
-     * @param partitionManager The partition manager.
-     * @param eventHubName The Event Hub name.
+     * @param initialEventPosition The event position to start processing events from a partition if no checkpoint is
```
jsquire: Should we consider having this on a per-partition level?   Since checkpoints and other concepts apply to a given partition, it seems limiting to have this as a single unit applied to all.  My vote would be to remove this and offer guidance on how to bootstrap a checkpoint if there's a need to set a position other than "first."   If we're going to keep it as a parameter, I believe that we really need to make this a map of `partitionId => eventPosition`

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,333 @@
...
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import reactor.core.publisher.Mono;
+import reactor.util.function.Tuple2;
+
+/**
+ * This class is responsible for balancing the load of processing events from all partitions of an Event Hub by
+ * distributing the number of partitions uniformly among all the  active {@link EventProcessor}s.
```
jsquire: nit: extra space "all  the"

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,333 @@
...
+    private final PartitionPumpManager partitionPumpManager;
+
+    /**
+     * Creates an instance of PartitionBasedLoadBalancer for the given Event Hub name and consumer group.
+     *
+     * @param partitionManager The partition manager that this load balancer will use to read/update ownership details.
+     * @param eventHubAsyncClient The asynchronous Event Hub client used to consume events.
+     * @param eventHubName The Event Hub name the {@link EventProcessor} is associated with.
+     * @param consumerGroupName The consumer group name the {@link EventProcessor} is associated with.
+     * @param ownerId The owner identifier for the {@link EventProcessor} this load balancer is associated with.
```
jsquire: nit:  This reads a bit awkward.  Maybe consider "The identifier of the `Event Processor` that owns this load balancer" or similar?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,333 @@
...
+     * This method works with the given partition ownership details and Event Hub partitions to evaluate whether the
+     * current Event Processor should take on the responsibility of processing more partitions.
+     */
+    private Mono<Void> loadBalance(final Tuple2<Map<String, PartitionOwnership>, List<String>> tuple) {
+
+        Map<String, PartitionOwnership> partitionOwnershipMap = tuple.getT1();
+        List<String> partitionIds = tuple.getT2();
+
+        if (ImplUtils.isNullOrEmpty(partitionIds)) {
+            // ideally, should never happen.
```
jsquire: "Ideally"  or "under normal circumstances"...  What I'm asking is whether this is something that would potentially occur and is expected or is an indicator that something is wrong.  The comment implies the former, but the implementation seems to imply the latter.

jsquire: Assuming this is reading the actual Event Hub metadata, this is impossible and would indicate a service failure of some sort.  Based on that, I think the comment is the artifact that is out of sync here.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,333 @@
...
+            return Mono.error(new IllegalStateException("There are no partitions in Event Hub " + this.eventHubName));
+        }
+
+        if (!isValid(partitionOwnershipMap)) {
+            // User data is corrupt.
+            return Mono.error(new IllegalStateException("Invalid partitionOwnership data from PartitionManager"));
+        }
+
+        /*
+         * Remove all partitions ownerships that have not be modified for a long time. This means that the previous
```
jsquire: "A long time" is ambiguous.   You may wish to consider offering context around how that is determined.  "Remove ownership that have not been modified within the configured period" or similar?

_(also, "ownership" is both singular and plural)_

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,333 @@
...
+                Collectors.groupingBy(PartitionOwnership::ownerId, mapping(Function.identity(), toList())));
+
+        // add the current event processor to the map if it doesn't exist
+        ownerPartitionMap.putIfAbsent(this.ownerId, new ArrayList<>());
+
+        /*
+         * Find the minimum number of partitions every event processor should own when the load is
+         * evenly distributed.
+         */
+        int minPartitionsPerEventProcessor = partitionIds.size() / ownerPartitionMap.size();
```
jsquire: nit: This may be a lack of understanding of language idioms, but you're making several calls to the `size()` methods in this logic.   Maybe it is worth considering reading them once and caching in a local variable for efficiency?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,333 @@
...
+     */
+    private boolean shouldOwnMorePartitions(final int minPartitionsPerEventProcessor,
+        final Map<String, List<PartitionOwnership>> ownerPartitionMap) {
+
+        int numberOfPartitionsOwned = ownerPartitionMap.get(this.ownerId).size();
+
+        int leastPartitionsOwnedByAnyEventProcessor =
+            ownerPartitionMap.values().stream().min(Comparator.comparingInt(List::size)).get().size();
+
+        if (numberOfPartitionsOwned < minPartitionsPerEventProcessor
```
jsquire: nit:  You could remove the conditional and just return the result of this expression.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,333 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    /*
+     * This method will create a new map of partition id and PartitionOwnership containing only those partitions
+     * that are actively owned. All entries in the original map returned by PartitionManager that haven't been
+     * modified for a duration of time greater than the allowed inactivity time limit are assumed to be owned by
+     * dead event processors. These will not be included in the map returned by this method.
```
jsquire: I wonder if we should add a tiny bit of random jitter to the time calculations.   If a bunch of event processors started running at just about the same time, then we're likely to see contention.  A bit of jitter to have them working at a slightly earlier or later duration for each calculation may help alleviate that.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -64,28 +50,22 @@
...
-            .requireNonNull(partitionManager, "partitionManager cannot be null");
-        this.initialEventPosition = Objects
-            .requireNonNull(initialEventPosition, "initialEventPosition cannot be null");
-        this.eventHubName = Objects
-            .requireNonNull(eventHubName, "eventHubName cannot be null");
+        PartitionManager partitionManager) {
+
         this.identifier = UUID.randomUUID().toString();
         logger.info("The instance ID for this event processors is {}", this.identifier);
+        this.partitionPumpManager = new PartitionPumpManager(partitionManager, partitionProcessorFactory,
```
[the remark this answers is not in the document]
srnagar: 😄 I used the same name Track 1 was using which I am not entirely satisfied with. Do you have a better name suggestion?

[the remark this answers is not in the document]
jsquire: Unfortunately, not a wonderful one.   Maybe something like `PartitionReaderManager` or similar?    Either way, I think you're good.  I just have a fun mental image every time I read about their "pumps"

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/CheckpointManager.java
```diff
@@ -31,7 +31,7 @@
      * is called from this CheckpointManager, this ETag will be used to provide <a href="https://en.wikipedia.org/wiki/Optimistic_concurrency_control">optimistic
      * concurrency</a>.
      */
-    CheckpointManager(String ownerId, PartitionContext partitionContext, PartitionManager partitionManager,
+    public CheckpointManager(String ownerId, PartitionContext partitionContext, PartitionManager partitionManager,
         String eTag) {
```
conniey: Since this can be constructed publicly, I'd add some error checking.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
```diff
@@ -395,9 +395,8 @@ public EventProcessor buildEventProcessor() {
         EventPosition initialEventPosition =
             this.initialEventPosition == null ? EventPosition.earliest()
                 : this.initialEventPosition;
-
         return new EventProcessor(buildAsyncClient(), this.consumerGroupName,
```
conniey: I don't think we have conventions around this, but `this.` is used inconsistently here to reference member variables. (ie. partitionManager doesn't have this). I'd prefer it without `this.` if there aren't any conflicting local variables (makes it more succinct), but I won't lose sleep if this change isn't in. 😆

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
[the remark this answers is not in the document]
conniey: They definitely have a lot of "pumps".. (ie. PartitionPump). 🤣

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * This class is responsible for balancing the load of processing events from all partitions of an Event Hub by
+ * distributing the number of partitions uniformly among all the  active {@link EventProcessor}s.
+ * <p>
+ * This load balancer will retrieve partition ownership details from the {@link PartitionManager} to find the number of
+ * active {@link EventProcessor}s. It uses the last modified time to decide if an EventProcessor is active. If a
+ * partition ownership entry has not be updated for a specified duration of time, the owner of that partition is
+ * considered inactive and the partition is available for other EventProcessors to own.
+ * </p>
+ * More details can be found <a href="https://gist.github.com/srnagar/7ef8566cfef0673288275c450dc99590">here</a>.
```
conniey: It may be good for this to live in a more official place, like one of our READMEs. (Tho, at this point, our README needs a table of contents. The Azure identity README has a TOC.)

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,331 @@
...
+            .subscribe();
+    }
+
+    /*
+     * This method works with the given partition ownership details and Event Hub partitions to evaluate whether the
+     * current Event Processor should take on the responsibility of processing more partitions.
+     */
+    private Mono<Void> loadBalance(final Tuple2<Map<String, PartitionOwnership>, List<String>> tuple) {
+
+        Map<String, PartitionOwnership> partitionOwnershipMap = tuple.getT1();
```
conniey: Why is this returning `Mono<Void>`? The work itself is synchronous until we return Mono.empty() at the very end.  One way to make this more async is to return a `Mono.fromRunnable(() -> {
// work in here.
});`

So the work is run asynchronously and completes when it ends.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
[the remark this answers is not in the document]
joshfree: MessageDispatcher?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -116,7 +102,8 @@ public synchronized void start() {
         }
         logger.info("Starting a new event processor instance with id {}", this.identifier);
         scheduler = Schedulers.newElastic("EventProcessor");
-        runner = scheduler.schedulePeriodically(this::run, INITIAL_DELAY, INTERVAL_IN_SECONDS, TimeUnit.SECONDS);
+        runner = scheduler.schedulePeriodically(partitionBasedLoadBalancer::loadBalance, INITIAL_DELAY,
+            INTERVAL_IN_SECONDS /* TODO: make this configurable */, TimeUnit.SECONDS);
```
joshfree: can you file a bug that references this todo?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,331 @@
...
+         * If the number of partitions in Event Hub is not evenly divisible by number of active event processors,
+         * a few Event Processors may own 1 additional partition than the minimum when the load is balanced. Calculate
+         * the number of event processors that can own additional partition.
+         */
+        int numberOfEventProcessorsWithAdditionalPartition = numberOfPartitions % numberOfActiveEventProcessors;
+
+        if (isLoadBalanced(minPartitionsPerEventProcessor, numberOfEventProcessorsWithAdditionalPartition,
+            ownerPartitionMap)) {
+            // If the partitions are evenly distributed among all active event processors, no change required.
+            return Mono.empty();
```
ShivangiReja: Lets say we have an Event Hubs with 3 partitions and we follows the below steps-
* Create Event Processor instance `processor1`
* call `processor1.start()`
* after some delay call `processor1.stop()`
* now  `processor1` owns all the 3 partitions and the load is balanced.
* if we call `processor1.start()` again, `isLoadBalanced()` will return true and here if we returns `Mono.empty()` it means we will stuck in the loop forever(if I understood correctly)?

 I think in this case we should create a new pump and start receiving events again from all the partitions?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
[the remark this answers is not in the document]
jsquire: I'd be a bit cautious with the word `Message` in the Event Hubs context.  Easy to confuse whether it acts on events or AMQP messages.... but the `Dispatcher` part of that, I really like!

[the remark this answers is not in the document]
jsquire: ... and because I feel that it is a moral imperative, each time I visit this PR:

image

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
jsquire: ... or do we just reset the state to assume that things are not balanced each time we stop?   
... or do we assume things are balanced and then slowly realize they're not and re balance them?


I don't have a strong opinion, I think either case works - its just a matter of how aggressive we pursue work stealing.  However, I'd suggest that we avoid throwing away an instance and triggering an allocation for a new one if we just need a state reset.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
joshfree: ditto; I'd personally limit `this` to use in getters/setters/ctors when required.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,352 @@
...
+    private boolean shouldOwnMorePartitions(final int minPartitionsPerEventProcessor,
+        final Map<String, List<PartitionOwnership>> ownerPartitionMap) {
+
+        int numberOfPartitionsOwned = ownerPartitionMap.get(this.ownerId).size();
+
+        int leastPartitionsOwnedByAnyEventProcessor =
+            ownerPartitionMap.values().stream().min(Comparator.comparingInt(List::size)).get().size();
+
+        return numberOfPartitionsOwned < minPartitionsPerEventProcessor
+            || numberOfPartitionsOwned == leastPartitionsOwnedByAnyEventProcessor;
```
ShivangiReja: I'm not sure if this second condition makes sense.

Scenario:
There are 6 partitions total, and 4 EventProcessors.
The expected distribution would be 2 EventProcessors owning 2 partitions, and 2 EventProcessors owning 1 partition.

Let's say our EventProcessor currently owns 1 partition. In this case, `numberOfPartitionsOwned === leastPartitionsOwnedByAnyEventProcessor` will be true, indicating that the EventProcessor should own more partitions. This would be incorrect though since based on the scenario, we know we've reached an optimal load balance.

I think the EventProcessor should own more partitions if:
1. numberOfPartitionsOwned is less than the minimum partitions per event processor.
2. The sum of partitions actively owned by any processor is less than the total number of partitions __AND__ the numberOfPartitionsOwned is less than the `minPartitionsPerEventProcessor + 1`.

srnagar: We won't reach this code if the load is already balanced. On line 189, there's a check to see if the load is already balanced. This method gets called only when the load is not balanced.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,352 @@
...
+        this.partitionPumpManager = partitionPumpManager;
+    }
+
+    /**
+     * This is the main method responsible for load balancing. This method is expected to be invoked by the {@link
+     * EventProcessor} periodically. Every call to this method will result in this {@link EventProcessor} owning <b>at
+     * most one</b> new partition.
+     * <p>
+     * The load is considered balanced when no active EventProcessor owns 2 partitions more than any other active
+     * EventProcessor.Given that each invocation to this method results in ownership claim of at most one partition,
```
ShivangiReja: nit: missing space `EventProcessor.Given`

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,352 @@
...
+                ownerPartitionMap)) {
+                // If the partitions are evenly distributed among all active event processors, no change required.
+                logger.info("Load is balanced");
+                return;
+            }
+
+            if (!shouldOwnMorePartitions(minPartitionsPerEventProcessor, ownerPartitionMap)) {
+                // This event processor already has enough partitions and shouldn't own more.
+                logger.info("This event processor owns {} partitions and shouldn't own more",
+                    ownerPartitionMap.get(ownerId));
```
ShivangiReja: Do you mean to print the number of partitions owned, or the array? Right now I think it will do the array and not the length.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,352 @@
...
+             * it's partitions are now available for others to own or because event processors are just
+             * starting up and gradually claiming partitions to own or new partitions were added to Event Hub.
+             * Find any partition that is not actively owned and claim it.
+             *
+             * OR
+             *
+             * Find a partition to steal from another event processor. Pick the event processor that has owns the highest
+             * number of partitions.
+             */
+            String partitionToClaim = partitionIds.parallelStream()
```
ShivangiReja: Just curious, why are we using `parallelStream()` instead of regular `stream()`? are we making any service calls/ IO calls ?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,352 @@
...
+            logger.info("Partition manager returned {} ownership records", partitionOwnershipMap.size());
+            logger.info("EventHubAsyncClient returned {} partitions", numberOfPartitions);
+            if (!isValid(partitionOwnershipMap)) {
+                // User data is corrupt.
+                throw Exceptions
+                    .propagate(new IllegalStateException("Invalid partitionOwnership data from PartitionManager"));
+            }
+
+            /*
+             * Remove all partitions' ownership that have not be modified for a configuration period of time. This means
```
ShivangiReja: nit: ownership that have not `been` modified

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,353 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import reactor.core.Exceptions;
+import reactor.core.publisher.Mono;
+import reactor.util.function.Tuple2;
+
+/**
+ * This class is responsible for balancing the load of processing events from all partitions of an Event Hub by
+ * distributing the number of partitions uniformly among all the  active {@link EventProcessor}s.
+ * <p>
```
conniey: nit: `the  active` -> `the active`. 
Does `{@link EventProcessor}s` render nicely in javadocs? `{@link EventProcessor EventProcessors}` may look better. Same with the one below.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionBasedLoadBalancer.java
```diff
@@ -0,0 +1,353 @@
...
+     * @param ownerId The identifier of the {@link EventProcessor} that owns this load balancer.
+     * @param inactiveTimeLimitInSeconds The time in seconds to wait for an update on an ownership record before
+     * assuming the owner of the partition is inactive.
+     * @param partitionPumpManager The partition pump manager that keeps track of all EventHubConsumers and partitions
+     * that this {@link EventProcessor} is processing.
+     */
+    public PartitionBasedLoadBalancer(final PartitionManager partitionManager,
+        final EventHubAsyncClient eventHubAsyncClient,
+        final String eventHubName, final String consumerGroupName, final String ownerId,
+        final long inactiveTimeLimitInSeconds, final PartitionPumpManager partitionPumpManager) {
```
conniey: `inactiveTimeLimitInSeconds` I wonder if it would be easier to understand if we used a class like Duration rather than a long.

PR #5035, opened 2019-08-19
Add Impressions tracking link to READMEs

g2vinay:


sdk/storage/microsoft-azure-storage-blob/README.md
```diff
@@ -186,3 +186,5 @@ When sending pull requests, please send non-breaking PRs to the dev branch and b
 * [Azure Storage Service](http://azure.microsoft.com/en-us/documentation/services/storage/)
 * [Azure Storage Team Blog](http://blogs.msdn.com/b/windowsazurestorage/)
 * [Javadoc](http://azure.github.io/azure-storage-java/)
+
+![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-java/sdk/storage/microsoft-azure-storage-blob/README.png)
```
alzimmermsft: Do we want to add this for track 1 libraries?

alzimmermsft:
Looks like build is failing due to the Analyze step due to the azure-storage-file README missing `* Azure .+ client library for Java`

g2vinay:
> Looks like build is failing due to the Analyze step due to the azure-storage-file README missing * Azure .+ client library for Java

Looks like a false positive, because the header exists.
I'll revert the changes to file and introduce again,

scbedd:
> > Looks like build is failing due to the Analyze step due to the azure-storage-file README missing * Azure .+ client library for Java
> 
> Looks like a false positive, because the header exists.
> I'll revert the changes to file and introduce again,

A false positive because the header exists? It's a regex....

`# Azure File client library for Java`

against regex:

`Azure .+ client library for Java`

...is failing. BOM maybe? There is a known issue with that right now.

sdk/storage/microsoft-azure-storage-blob/README.md
g2vinay: Most of the Track 1 libraries have this already, but they need some correction.
Will tackle Track 1 links in a separate PR.

PR #5042, opened 2019-08-19
Cosmos Direct TCP port from azure-cosmosdb v2.6.1

David-Noble-at-work:
This port brings v3 into alignment with v2.6.1 of the Cosmos SDK for Java. Except for RntbdTransportClient.java which has some adjustments for Reactor the implementations are now equivalent.

### DIFF
--- a/sdk/cosmos/microsoft-azure-cosmos-examples/src/main/java/com/azure/data/cosmos/examples/AccountSettings.java
+++ b/sdk/cosmos/microsoft-azure-cosmos-examples/src/main/java/com/azure/data/cosmos/examples/AccountSettings.java
@@ -7,17 +7,17 @@ import org.apache.commons.lang3.StringUtils;
 
 /**
  * Contains the account configurations for Sample.
- * 
+ *
  * For running tests, you can pass a customized endpoint configuration in one of the following
  * ways:
  * <ul>
- * <li>-DACCOUNT_KEY="[your-key]" -ACCOUNT_HOST="[your-endpoint]" as JVM
+ * <li>-DACCOUNT_KEY="[your-key]" -DACCOUNT_HOST="[your-endpoint]" as JVM
  * command-line option.</li>
- * <li>You can set ACCOUNT_KEY and ACCOUNT_HOST as environment variables.</li>
+ * <li>You can set COSMOS_ACCOUNT_KEY and COSMOS_ACCOUNT_HOST as environment variables.</li>
  * </ul>
- * 
+ *
  * If none of the above is set, emulator endpoint will be used.
- * Emulator http cert is self signed. If you are using emulator, 
+ * Emulator http cert is self signed. If you are using emulator,
  * make sure emulator https certificate is imported
  * to java trusted cert store:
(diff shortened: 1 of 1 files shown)


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/Configs.java
```diff
@@ -68,9 +76,11 @@ public SslContext getSslContext() {
     }
 
     public Protocol getProtocol() {
-        String protocol = getJVMConfigAsString(PROTOCOL, DEFAULT_PROTOCOL.toString());
+        String protocol = System.getProperty(PROTOCOL_PROPERTY, firstNonNull(
+            emptyToNull(System.getenv().get(PROTOCOL_ENVIRONMENT_VARIABLE)),
+            DEFAULT_PROTOCOL.name()));
```
moderakh: given the fact that protocol name is different in v2 and v3, Kushagra did some coding here to ensure our CI works for both v2 and v3. As v3 is on a different repo with new CIs, I am not sure how much of that is still applicable but good to check.

@kushagraThapar does this break anything in the CI?

David-Noble-at-work: Thanks for the review. Kushagra may know something more but I've not seen that CI is broken by the change. l'll wait to hear back from him before resolving this issue.

sdk/cosmos/microsoft-azure-cosmos-benchmark/pom.xml
```diff
@@ -43,28 +43,46 @@ Licensed under the MIT License.
...
+      <version>${metrics.version}</version>
     </dependency>
 
     <dependency>
       <groupId>io.dropwizard.metrics</groupId>
       <artifactId>metrics-graphite</artifactId>
+      <version>${metrics.version}</version>
     </dependency>
 
+    <dependency>
```
christopheranderson: ✅ We have approval for these dependencies (in email)

sdk/cosmos/microsoft-azure-cosmos/pom.xml
```diff
@@ -113,6 +113,18 @@ Licensed under the MIT License.
       <artifactId>guava</artifactId>
     </dependency>
 
+    <dependency>
```
christopheranderson: ✅ We have approval for these dependencies (in email)

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/BridgeInternal.java
```diff
@@ -52,6 +53,10 @@ public static Document documentFromObject(Object document, ObjectMapper mapper)
         return Document.FromObject(document, mapper);
     }
 
+    public static void monitorTelemetry(MeterRegistry registry) {
```
christopheranderson: ⚠ This is public API change exposing an unapproved public dependency, but we've documented that this public API shouldn't be depended on, so I'm not seeking approval at this time. Approved.

PR #5057, opened 2019-08-20
Fix Install Linting Issues

alzimmermsft:
Removed newlines from regex ORs to prevent them from failing to match. Fixed JavadocPackage ignore as that isn't a custom check.

PR #5058, opened 2019-08-20
A minor fix that bring back JavadocPackage check 

mssfang:
A minor fix that bring back JavadocPackage check


conniey:
This overlaps with @alzimmermsft 's PR #5057. I don't think you need this.

PR #5061, opened 2019-08-20
Amqp tracing support

samvaity:
- Added support for OpenTelemetry for AMQP Libraries
- EventHubs Send instrumented with Context propogation (amqp)
- EventHubs Process instrumented with Context propogation (amqp)

### DIFF
--- a/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TracerProvider.java
+++ b/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TracerProvider.java
@@ -6,6 +6,7 @@ import com.azure.core.amqp.exception.AmqpException;
 import com.azure.core.implementation.tracing.ProcessKind;
 import com.azure.core.implementation.tracing.Tracer;
 import com.azure.core.util.Context;
+import com.azure.core.util.logging.ClientLogger;
 import reactor.core.publisher.Signal;
 
 import java.util.ArrayList;
@@ -15,7 +16,7 @@ import java.util.Objects;
 import static com.azure.core.implementation.tracing.Tracer.OPENTELEMETRY_SPAN_KEY;
 
 public class TracerProvider {
-
+    private final ClientLogger logger = new ClientLogger(TracerProvider.class);
     private final List<Tracer> tracers = new ArrayList<>();
 
     public TracerProvider(Iterable<Tracer> tracers) {
@@ -41,7 +42,8 @@ public class TracerProvider {
     public Context startSpan(Context context, ProcessKind processKind) {
         Context local = Objects.requireNonNull(context, "'context' cannot be null");
         Objects.requireNonNull(processKind, "'processKind' cannot be null");
-        String spanName = "Azure.eventhubs." + processKind.getProcessKind();
+        String spanName = getSpanName(processKind);
+
         for (Tracer tracer : tracers) {
             local = tracer.start(spanName, local, processKind);
         }
@@ -118,5 +120,25 @@ public class TracerProvider {
         for (Tracer tracer : tracers) {
             tracer.end(statusMessage, throwable, context);
         }
+    }
+
+    private String getSpanName(ProcessKind processKind) {
+        String spanName = "Azure.eventhubs.";
+        switch (processKind) {
+            case SEND:
+                spanName += "send";
+                break;
+            case RECEIVE:
+                spanName += "message";
+                break;
+            case PROCESS:
+                spanName += "process";
+                break;
+            default:
+                logger.warning("Unknown processKind type: {}", processKind);
+                break;
+        }
+
+        return spanName;
     }
 }
--- a/sdk/core/azure-core/src/main/java/com/azure/core/implementation/tracing/ProcessKind.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/implementation/tracing/ProcessKind.java
@@ -10,28 +10,13 @@ public enum ProcessKind {
     /**
      * Amqp Send Message process call to send data.
      */
-    SEND("send"),
+    SEND,
     /**
      * Amqp message process call to receive data.
      */
-    RECEIVE("message"),
+    RECEIVE,
     /**
      * Custom process call to process received messages.
      */
-    PROCESS("process");
-
-    private final String processKind;
-
-    ProcessKind(String processKind) {
-        this.processKind = processKind;
-    }
-
-    /**
-     * Gets the AMQP process type.
-     *
-     * @return AMQP process type for the calling method.
-     */
-    public String getProcessKind() {
-        return processKind;
-    }
+    PROCESS
 }


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TraceUtil.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.azure.core.amqp.exception.AmqpException;
+import com.azure.core.implementation.tracing.TracerProxy;
+import com.azure.core.util.Context;
+import reactor.core.publisher.Signal;
+
+import java.util.Optional;
+
+public final class TraceUtil {
+
```
JonathanGiles: Add a private no-op constructor here just to defend against people creating instances of this class for no reason.

samvaity:
> I'm guessing this still needs updating before PR review is ready?

Yes some TODO's mentioned [here](https://github.com/Azure/azure-sdk-for-java/pull/5061#issue-309246264). But this will be the initial work for supporting Tracing in AMQP.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
```diff
@@ -171,6 +174,17 @@ public EventData addProperty(String key, Object value) {
...
+    /**
+     * Creates a new {@link Context} object with all the keys and values provided
+     *
+     * @param context The key value pair object
+     * @return The updated EventData object.
+     */
+    public EventData context(Context context) {
+        this.context = context;
+        return this;
+    }
```
JonathanGiles: @conniey Do you want to give your thoughts on this? Do you want Context passed in to the constructor or is this ok?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/tracing/Tracer.java
```diff
@@ -47,6 +84,15 @@
...
     void end(int responseCode, Throwable error, Context context);
 
+    /**
+     * Completes the current tracing span.
+     *
+     * @param errorCondition the error message that occurred during the call.
+     * @param error Potential throwable that happened during the span.
+     * @param context Additional metadata that is passed through the call stack.
+     */
+    void end(String errorCondition, Throwable error, Context context);
```
conniey: This is oddly specific to AMQP. Is it better to have an AMQP specific Tracer?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
conniey: I'd prefer the fields to be immutable, so having this as a constructor parameter is better, imho.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncProducer.java
```diff
@@ -110,6 +115,10 @@
     private final EventHubProducerOptions senderOptions;
     private final Mono<AmqpSendLink> sendLinkMono;
     private final boolean isPartitionSender;
+    private final String entityPath = Tracer.OPENTELEMETRY_AMQP_ENTITY_PATH;
```
conniey: Why allocate fields for something that is a constant?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -44,6 +49,9 @@
     private static final long INTERVAL_IN_SECONDS = 10; // run every 10 seconds
     private static final long INITIAL_DELAY = 0; // start immediately
     private static final long OWNERSHIP_EXPIRATION_TIME_IN_MILLIS = TimeUnit.SECONDS.toMillis(30);
+    private static final String SPAN_CONTEXT = Tracer.OPENTELEMETRY_AMQP_EVENT_SPAN_CONTEXT;
```
conniey: These fields don't need to exist.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -229,14 +237,48 @@ private void receiveEvents(PartitionOwnership partitionOwnership) {
...
+                startScopedTracingSpan(eventData, processSpanContext);
+                partitionProcessor.processEvent(eventData).subscribe(unused -> {
+                }, partitionProcessor::processError);
+                endScopedTracingSpan(processSpanContext);
+            },
                 partitionProcessor::processError,
-                // Currently, there is no way to distinguish if the receiver was closed because
-                // another receiver with higher/same owner level(epoch) connected or because
-                // this event processor explicitly called close on this consumer.
+            // Currently, there is no way to distinguish if the receiver was closed because
```
conniey: This can be undone? It looks like the comments are not at the same indentation level.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/AmqpSendLink.java
```diff
@@ -45,4 +45,11 @@
      * Gets the context for this AMQP send link.
      */
     ErrorContext getErrorContext();
+
+    /**
+     * The remote endpoint path this link is connected to.
+     *
+     * @return The remote endpoint path this link is connected to.
+     */
+    String getHostname();
```
conniey: This hostname is common to an AmqpLink. Update azure-core-amqp's AmqpLink to have this method

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -23,34 +31,57 @@
     private static final Tracer TRACER = Tracing.getTracer();
     private static final String OPENTELEMETRY_SPAN_KEY = com.azure.core.implementation.tracing.Tracer.OPENTELEMETRY_SPAN_KEY;
     private static final String OPENTELEMETRY_SPAN_NAME_KEY = com.azure.core.implementation.tracing.Tracer.OPENTELEMETRY_SPAN_NAME_KEY;
+    private static final String DIAGNOSTIC_ID = com.azure.core.implementation.tracing.Tracer.OPENTELEMETRY_DIAGNOSTIC_ID_KEY;
```
conniey: This seems like a duplicated memory allocation. You can import the constant as a static import.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TraceUtil.java
```diff
@@ -0,0 +1,87 @@
...
+import com.azure.core.implementation.tracing.TracerProxy;
+import com.azure.core.util.Context;
+import reactor.core.publisher.Signal;
+
+/**
+ * Helper class to help start tracing spans.
+ */
+public class TraceUtil {
+
+    private static final String OPENTELEMETRY_SPAN_KEY = com.azure.core.implementation.tracing.Tracer.OPENTELEMETRY_SPAN_KEY;
```
srnagar: You could import this field statically instead of redefining another constant here.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/tracing/TracerProxy.java
```diff
@@ -38,6 +38,26 @@ public static Context start(String methodName, Context context) {
         return local;
     }
 
+    /**
```
conniey: Still not a fan of TracerProxy being a static class. makes it hard to test

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TraceUtil.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Starts the tracing span for the current service call.
+     *
+     * @param context Context information about the current service call.
+     * @return The updated context containing the span context.
+     */
+    public static Context start(String methodName, Context context) {
+        String spanName = String.format("Azure.eventhubs.%s", methodName);
```
srnagar: `String.format()` is generally slower than just using `+`. You could instead use `"Azure.eventhubs" + methodName`. Given that `start()` call is made every single time a service call is made, we should try and avoid this.

https://kylewbanks.com/blog/java-string-concatenation-vs-stringbuilder-vs-string-format-performance
https://stackoverflow.com/questions/513600/should-i-use-javas-string-format-if-performance-is-important

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TraceUtil.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Starts a new scoped tracing span for the current service call.
+     *
+     * @param context Context information about the current service call.
+     * @return The updated context containing the span context.
+     */
+    public static Context startScopedSpan(String methodName, Context context) {
+        String spanName = String.format("Azure.eventhubs.%s", methodName);
```
srnagar: same here.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TraceUtil.java
```diff
@@ -0,0 +1,87 @@
...
+        context = TracerProxy.setSpanName(spanName, context);
+        return TracerProxy.startScopedSpan(spanName, context);
+    }
+
+    /**
+     * Given a context containing the current tracing span the span is marked completed with status info from {@link Signal}
+     *
+     * @param context Additional metadata that is passed through the call stack.
+     * @param signal The signal indicates the status and contains the metadata we need to end the tracing span.
+     *
```
srnagar: nit: remove this line

sdk/core/azure-core/src/main/java/com/azure/core/implementation/tracing/TracerProxy.java
conniey: Can we instead, pass a "Tracer" into EventHubClient as a ctor parameter and propagate that? That way, we can mock the other calls.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/tracing/TracerProxy.java
```diff
@@ -76,4 +96,38 @@ public static Context setSpanName(String spanName, Context context) {
...
+
+    /**
+     * For each tracer plugged into the SDK a new context is extracted from the event's diagnostic Id.
+     *
+     * @param diagnosticId Unique identifier of an external call from producer to the queue.
+     */
+    public static Context extractContext(String diagnosticId) {
+        Context local = Context.NONE;
+        for (Tracer tracer : tracers) {
+            local = tracer.extractContext(diagnosticId);
```
srnagar: If there are multiple tracers, wouldn't this `local` context be overwritten and only the last value retained?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncProducer.java
```diff
@@ -296,19 +305,46 @@
...
                     final int batchSize = size > 0 ? size : MAX_MESSAGE_LENGTH_BYTES;
                     final BatchOptions batchOptions = new BatchOptions()
                         .partitionKey(partitionKey)
                         .maximumSizeInBytes(batchSize);
 
-                    return events.collect(new EventDataCollector(batchOptions, 1, () -> link.getErrorContext()));
+                    return events.map(eventData -> {
+                        Context parentContext = eventData.context();
+                        Context entityContext = parentContext.addData(entityPath, link.getEntityPath());
+                        sendSpanContext.set(TraceUtil.start("send", entityContext.addData(hostName, link.getHostname())));
```
srnagar: Using the static method to `start` will be hard to unit test.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/AmqpSendLink.java
srnagar: Should this be called the `getEndpoint()` since the javadoc refers this to endpoint?

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
srnagar: You can use static imports instead for all these constants.

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/implementation/AmqpPropagationFormatUtil.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Parse span context and get the diagnostic Id for that event.
+     *
+     * @param spanContext the span context.
+     * @retrun The dignostic Id
+     */
+    public static String getDiagnosticId(SpanContext spanContext) {
+        char[] chars = new char[55];
```
conniey: Is there a spec to look at for this?

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/implementation/AmqpTraceUtil.java
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.tracing.opentelemetry.implementation;
+
+import io.opencensus.trace.Status;
+
+public final class AmqpTraceUtil {
```
conniey: This is a lot of duplication that doesn't belong in this library. I'd add a reference to azure-core-amqp and use those string constants instead.

conniey: Where did the mapping from ErrorCondition to status codes come from? maybe that can live in azure-core-amqp if they are common. I know there are a few in azure-core-amqp.

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -23,34 +31,57 @@
...
-    public Context start(String methodName, Context context) {
-        Span parentSpan = (Span) context.getData(OPENTELEMETRY_SPAN_KEY).orElse(TRACER.getCurrentSpan());
-        String spanName = (String) context.getData(OPENTELEMETRY_SPAN_NAME_KEY).orElse(methodName);
-
-        SpanBuilder spanBuilder = TRACER.spanBuilderWithExplicitParent(spanName, parentSpan);
-        Span span = spanBuilder.startSpan();
-
+    public Context start(String spanName, Context context) {
+        Span span = startSpanWithExplicitParent(spanName, context);
+        if (context.getData(ENTITY_PATH).isPresent() && span.getOptions().contains(Span.Options.RECORD_EVENTS)) {
```
lmolkova: I think I'm missing  something
1. we start span from explicit parent - fine
2. if there is entity path, we add it to span - fine
3. why we setContextData  if there is no entity path OR span is sampled out? - this doesn't look right

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -23,34 +31,57 @@
...
+            // If span is sampled in, add additional TRACING attributes
+            addSpanRequestAttributes(span, context, spanName);
+        } else {
+            // Add diagnostic Id to Context
+            context = setContextData(span);
```
lmolkova: I think I'm missing something

we start span from explicit parent - fine
if there is entity path, we add it to span - fine
why we setContextData if there is no entity path OR span is sampled out? - this doesn't look right

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -73,4 +104,71 @@ public void setAttribute(String key, String value, Context context) {
...
+        Optional<Object> spanOptional = eventContext.getData(OPENTELEMETRY_SPAN_KEY); // TODO: we need this to be the parent span key
+
+        if (!spanOptional.isPresent()) {
+            logger.warning("Failed to find span to link it.");
+            return;
+        }
+        SpanContext spanContext = (SpanContext) spanContextOptional.get();
+        Span span = (Span) spanOptional.get();
+
+        span.addLink(Link.fromSpanContext(spanContext, PARENT_LINKED_SPAN));
```
lmolkova: can you please leave a TODO here (or where we call this method) that once we switch to OpenTelemetry, we need to add links before span is starter (and no link type - parent - is needed)

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
lmolkova: I see, you distinguish 'message' from 'send' span this way. Is there a more obvious and stable way?

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/implementation/AmqpPropagationFormatUtil.java
samvaity: This is the spec - https://gist.github.com/lmolkova/e4215c0f44a49ef824983382762e6b92#file-minimal_instrumentation-md

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncProducer.java
conniey: I'm hoping doing something more DI-ish like this will remove the need for it.
https://github.com/Azure/azure-sdk-for-java/pull/5061#discussion_r316944552

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
lmolkova: Also 'send' span should have client kind, 'message' can have no kind and 'process' should be 'server' kind - we should somehow add it here or in other appropriate method

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/AmqpSendLink.java
conniey: It's that weird string like: eh-namespace.servicebus.windows.net. I'm not sure it's an endpoint because we can't use https:// or sb:// to connect to it. In the track1 code they called it "hostname" so I maintained that name.

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
samvaity: I think we can directly use the span name to distinguish that. It would be well-formed by the time it reaches here. `Azure.eventhubs.send` and `Azure.eventhubs.message`

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/implementation/AmqpPropagationFormatUtil.java
conniey: Can we add documentation in the javadocs how this gets created? A year from now, someone else looking at this would have no idea.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/AmqpSendLink.java
samvaity: yeah, the javadoc was the same from above `EntityPath`. Updated it.

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
lmolkova: can we do this without relying on the span name? E.g. explicitly call setContextData when needed? 
It seems like if span name changes or we need to add other kids of spans, we'd have special-case all names here. 
This is also a bit hard to read and follow.

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -6,51 +6,82 @@
...
-    public Context start(String methodName, Context context) {
-        Span parentSpan = (Span) context.getData(OPENTELEMETRY_SPAN_KEY).orElse(TRACER.getCurrentSpan());
-        String spanName = (String) context.getData(OPENTELEMETRY_SPAN_NAME_KEY).orElse(methodName);
-
-        SpanBuilder spanBuilder = TRACER.spanBuilderWithExplicitParent(spanName, parentSpan);
-        Span span = spanBuilder.startSpan();
+    public Context start(String spanName, Context context) {
+        SpanBuilder spanBuilder = startSpanWithExplicitParent(spanName, context);
+        Span span;
+        if (spanName.contains("send")) {
```
lmolkova: what if we add receive? will we need to special-case it too? Is there a more explicit way to assign kind to span?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -229,14 +236,48 @@ private void receiveEvents(PartitionOwnership partitionOwnership) {
...
             .createPartitionProcessor(partitionContext, checkpointManager);
         partitionProcessor.initialize().subscribe();
-
+        final AtomicReference<Context> processSpanContext = new AtomicReference<>(Context.NONE);
         consumer.receive().subscribeOn(Schedulers.newElastic("PartitionPump"))
-            .subscribe(eventData -> partitionProcessor.processEvent(eventData).subscribe(unused -> {
-            }, partitionProcessor::processError),
+            .subscribe(eventData -> {
+                startScopedTracingSpan(eventData, processSpanContext);
+                partitionProcessor.processEvent(eventData).subscribe(unused -> {
```
lmolkova: I recall this is asyn call `partitionProcessor.processEvent`, right? If so, `endScopedTracingSpan(processSpanContext)` happens before user code returns? Or am I missing something?

sdk/tracing/tracing-opentelemetry/src/main/java/com/azure/tracing/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -6,51 +6,82 @@
...
+        return context.addData(OPENTELEMETRY_SPAN_KEY, spanBuilder.startSpan());
+    }
 
-        return context.addData(OPENTELEMETRY_SPAN_KEY, span);
+    @Override
+    public Context startScopedSpan(String spanName, Context context) {
+        Span span;
+        if (context.getData(SPAN_CONTEXT).isPresent()) {
+            span = startSpanWithRemoteParent(spanName, context);
+        } else {
```
lmolkova: this is not an obvious way to distinguish 'process'  span from others. If there a more explicit way that does not rely on presence of context or span name?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -229,14 +236,48 @@ private void receiveEvents(PartitionOwnership partitionOwnership) {
...
                 // Currently, there is no way to distinguish if the receiver was closed because
                 // another receiver with higher/same owner level(epoch) connected or because
                 // this event processor explicitly called close on this consumer.
                 () -> partitionProcessor.close(CloseReason.LOST_PARTITION_OWNERSHIP));
     }
+
+    /*
+     * Starts a new process tracing span and attached context the EventData object for users.
+     */
+    private void startScopedTracingSpan(EventData eventData, AtomicReference<Context> processSpanContext) {
```
lmolkova: this method doesn't seem like general-purpose one and only creates process spans, maybe is should be called like that?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -229,14 +236,48 @@ private void receiveEvents(PartitionOwnership partitionOwnership) {
...
+        if (!processSpanContext.get().getData("scope").isPresent()) {
+            return;
+        }
+        Closeable close = (Closeable) processSpanContext.get().getData("scope").get();
+        try {
+            close.close();
+        } catch (IOException ioException) {
+            logger.error("EventProcessor.run() endTracingSpan().close() failed with an error %s", ioException);
+        }
+        TraceUtil.endTracingSpan(processSpanContext.get(), null);
```
lmolkova: what if processing results with an exception on the user side, how do we set error status for processing span?

samvaity: the doOnEach will get us OnComplete and OnError so then we can just pass down the error and have the user error message in the description for Status.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/tracing/TracerProxy.java
samvaity: updated to keep adding the context for each tracer to `local` context object.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
samvaity: same as [this](https://github.com/Azure/azure-sdk-for-java/pull/5061#discussion_r317320920)
doOnEach will wait till an item is emitted with success or error signal.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
samvaity: In this, we have a scenario where we need the updated `EventData` object with the new span Context attached to it. But to do so, 

1. We would either need to add a method similar to `addProperty` to achieve it which returns the updated EventData object.
2. Or create a new `EventData` object from the parameterized constructor(with Context) using the existing event data object (overkill?)

Thoughts?

PR #5078, opened 2019-08-21
Tune management build triggering

mitchdenny:
The ```Azure.azure-sdk-for-java-mgmt ``` build triggers on every change in the repo currently which is unnecessary. For the time being I'm proposing that we stop triggering on the ```sdk/``` folder until we restructure the management plane into the ```sdk/``` folder at which point we might need to have some finer grained rules.

### DIFF
--- a/eng/pipelines/mgmt.yml
+++ b/eng/pipelines/mgmt.yml
@@ -3,8 +3,6 @@ trigger:
     include:
     - master
   paths:
-    include:
-    - sdk/core/
     exclude:
     - sdk/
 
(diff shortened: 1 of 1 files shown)


eng/pipelines/mgmt.yml
```diff
@@ -1,6 +1,23 @@
 trigger:
-  - master
-  
+  branches:
+    include:
+    - master
+  paths:
+    include:
+    - sdk/core/
```
weshaggard: Why include sdk/core?

mitchdenny: Because there is core-management now if I recall correctly and I'm assuming they have a dependency on that?

mitchdenny: Actually looking through this now - it doesn't appear to have a dependency. So I'll remove this until they do.

PR #5080, opened 2019-08-21
Fix tests

sima-zhu:


sdk/storage/pom.service.xml
```diff
@@ -9,7 +9,7 @@
   <packaging>pom</packaging>
   <version>1.0.0</version><!-- Need not change for every release-->
   <modules>
-<!--   <module>microsoft-azure-storage-blob</module>-->
+   <module>microsoft-azure-storage-blob</module>
```
alzimmermsft: Why did we enable the track 1 library to run CI during check in?

alzimmermsft:
Looks like a few of the Blobs tests run during playback even though they don't have playback records, to get this to pass add an assume true statement on test mode being record to disable that.

parent/pom.xml
```diff
@@ -107,8 +107,8 @@
     <proton-j-version>0.31.0</proton-j-version>
     <qpid-proton-j-extensions-version>1.2.0</qpid-proton-j-extensions-version>
     <log4j-api.version>2.11.1</log4j-api.version>
-    <groovy-eclipse-compiler.version>2.9.3-01</groovy-eclipse-compiler.version>
-    <groovy-eclipse-batch.version>2.4.16-03</groovy-eclipse-batch.version>
+    <groovy-eclipse-compiler.version>3.4.0-01</groovy-eclipse-compiler.version>
```
alzimmermsft: Make sure nothing else is using this dependency

sima-zhu: Sure. The plugin is intended the run Java and Groovy code together in maven, I believe no other client using the plugin for now.

sdk/storage/pom.service.xml
sima-zhu: Comment out.

parent/pom.xml
alzimmermsft: Perfect, just wanted to make sure we weren't accidentally bumping a dependency on other libraries.

sdk/storage/pom.service.xml
```diff
@@ -15,7 +15,7 @@
     <module>../core/azure-core-http-netty</module>
     <module>../identity/azure-identity</module>
     <module>azure-storage-common</module>
-    <module>azure-storage-blob</module>
+<!--    <module>azure-storage-blob</module>-->
```
alzimmermsft: I don't think disabling all of blobs during CI is the correct way to fix the two tests that are accidentally running.

sima-zhu: I am going to fix it for next commit. 
Is the failure testcases only run in record mode?

```
[ERROR] serviceSasSignatures string to sign(com.azure.storage.blob.SASTest)  Time elapsed: 0.001 s  <<< ERROR!
org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object 'id' with class 'java.lang.String' to class 'com.azure.storage.blob.models.SignedIdentifier'
        at com.azure.storage.blob.SASTest.serviceSasSignatures string to sign(SASTest.groovy:569)

[ERROR] ServiceSASSignatureValues assertGenerateOk(com.azure.storage.blob.SASTest)  Time elapsed: 0 s  <<< ERROR!
org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object '0000' with class 'java.lang.String' to class 'com.azure.storage.blob.models.SignedIdentifier'
        at com.azure.storage.blob.SASTest.ServiceSASSignatureValues assertGenerateOk(SASTest.groovy:856)
```

Issue #5013, opened 2019-08-15
Cosmos CFP: remove Thread.Sleep() call from PartitionProcessorImpl


milismsft:
Currently the call to Thrad.sleep() is to ensure that we respect the user set schedule for the REST API calls. Ultimately we want this to be a "non-blocking" code using Reactor functionality; we need to ensure that the first REST API call not delayed.

PR #5085, opened 2019-08-22
Cosmos CFP: fix for stale lease and remove blocking call

milismsft:
Fix for CFP host failing to take over stale leases that still have ownership set.
Remove last blocking call (see #5013)

### DIFF
--- a/sdk/cosmos/microsoft-azure-cosmos/src/test/java/com/azure/data/cosmos/rx/ChangeFeedProcessorTest.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/test/java/com/azure/data/cosmos/rx/ChangeFeedProcessorTest.java
@@ -106,7 +106,7 @@ public class ChangeFeedProcessorTest extends TestSuiteBase {
         try {
             Thread.sleep(2 * CHANGE_FEED_PROCESSOR_TIMEOUT);
         } catch (InterruptedException e) {
-            e.printStackTrace();
+            log.error(e.getMessage());
         }
 
         changeFeedProcessor.stop().subscribeOn(Schedulers.elastic()).timeout(Duration.ofMillis(CHANGE_FEED_PROCESSOR_TIMEOUT)).subscribe();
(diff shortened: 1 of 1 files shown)


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/LeaseStoreManagerImpl.java
```diff
@@ -435,16 +435,15 @@ public LeaseStoreManagerBuilderDefinition hostName(String hostName) {
...
 
     private String getDocumentId(String leaseToken)
     {
         return this.getPartitionLeasePrefix() + leaseToken;
     }
 
-    private String getPartitionLeasePrefix()
-    {
+    private String getPartitionLeasePrefix() {
         return this.settings.getContainerNamePrefix() + "..";
```
moderakh: we should consider making these as defined constants.

sdk/cosmos/microsoft-azure-cosmos/src/test/java/com/azure/data/cosmos/rx/ChangeFeedProcessorTest.java
```diff
@@ -40,7 +50,7 @@
     private CosmosContainer createdFeedCollection;
     private CosmosContainer createdLeaseCollection;
     private List<CosmosItemProperties> createdDocuments;
-    private static Map<String, CosmosItemProperties> receivedDocuments;
+    private static volatile Map<String, CosmosItemProperties> receivedDocuments;
```
moderakh: why is this volatile?

milismsft: The initialization of the map was happening inside a Reactor thread; this is no longer the case, so I'll remove the "volatile".

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/LeaseStoreManagerImpl.java
milismsft: changed to a constant

PR #5090, opened 2019-08-22
Remove parent/pom.xml for Event Hubs Track 1

conniey:


mitchdenny:
@conniey looks like you'll need to remove the intermediate parent POM reference from the other packages.

pom.data.xml
```diff
@@ -543,7 +543,9 @@
   <modules>
     <module>./sdk/batch/microsoft-azure-batch</module>
     <module>./sdk/cosmos</module>
```
conniey: @mitchdenny Do I need to bump the version of pom.data.xml because I updated the modules?

PR #5093, opened 2019-08-23
Update event processor API based on user study

srnagar:
The intent of this change is to address feedback from user studies for event processor. The changes include:
- A separate builder for creating new instance of Event Processor
- An abstract class instead of interface for Partition Processor that requires user to only implement processEvent. Other methods can be optionally overridden.
- Builder allows users to provide a lambda function to process events or for advanced users, partition processor factory is the alternate option.


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
```diff
@@ -218,6 +228,15 @@ public ByteBuffer body() {
         return body.duplicate();
     }
 
+    /**
```
conniey: Since we're updating this API surface area. Would you be opposed to adding this surface area to the EventData of yours, too? @jsquire @ramya-rao-a @YijunXieMS

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessor.java
```diff
@@ -24,20 +25,22 @@
...
+ * set up to consume events from the same Event Hub, consumer group to balance the workload across different instances
+ * and track progress when events are processed. Based on the number of instances running, each Event Processor may own
+ * 0 or more partitions to balance the workload among all the instances.
+ *
  * <p>
- * Event Processor based application consists of one or more instances of {@link EventProcessor} which are
- * set up to consume events from the same Event Hub + consumer group and to balance the workload across different
- * instances and track progress when events are processed.
+ * Event Processor based application consists of one or more instances of {@link EventProcessor} which are set up to
+ * consume events from the same Event Hub + consumer group and to balance the workload across different instances and
```
JonathanGiles: Don't use '+' when you mean to use 'and'. Also, this paragraph could do with a comma or two.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * <p>
+ * To create an instance of {@link EventProcessor} that processes events with user-provided lambda function and
+ * checkpoints after successfully processing every event, configure the following fields:
+ *
+ * <ul>
+ * <li>Consumer group name.</li>
+ * <li>{@link EventHubAsyncClient} - An asynchronous Event Hub client the {@link EventProcessor} will use for
+ * consuming events.</li>
+ * <li>{@link #processEvent(Function)} - A lambda function that will be called when new events are received.</li>
```
JonathanGiles:
```suggestion
 * <li>{@link #processEvent(Function)} - A callback for processing new events as they are received.</li>
```

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * Sets the lambda function that will be used to process events from all partitions this {@link EventProcessor} is
+     * responsible for.
+     *
+     * @param processEvent The lambda function that will be used to process events.
+     * @return The updated {@link EventProcessorBuilder} instance.
+     */
+    public EventProcessorBuilder processEvent(Function<EventData, Mono<Void>> processEvent) {
```
JonathanGiles: Do you need `Function<EventData, Mono<Void>>`? You don't document what the user is supposed to actually return, and I am not sure you have a use case for it anyway? Could it be `Consumer<EventData>` instead?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+        this.processEvent = processEvent;
+        return this;
+    }
+
+    /**
+     * Sets the {@link PartitionManager} the {@link EventProcessor} will use for storing partition ownership and
+     * checkpoint information.
+     *
+     * <p>
+     * Users can, optionally, provide their own implementation of {@link PartitionManager} which will store ownership
```
JonathanGiles: Put the opening and closing tag on the same line as the paragraph text.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+     * @param partitionProcessorFactory The factory that creates new {@link PartitionProcessor} for each partition.
+     * @return The updated {@link EventProcessorBuilder} instance.
+     */
+    public EventProcessorBuilder partitionProcessorFactory(PartitionProcessorFactory partitionProcessorFactory) {
+        this.partitionProcessorFactory = partitionProcessorFactory;
+        return this;
+    }
+
+    /**
+     * Sets the initial event position. If this property is not set and if checkpoint for a partition doesn't exist,
```
JonathanGiles: As you know, the first sentence is the most important sentence, as it gets shown in the summary table at the top of the JavaDoc. I suggest you consider something like this: 

```suggestion
     * Sets the initial event position, defaulting to {@link EventPosition#earliest()} if this property is not set.
```

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
ramya-rao-a: In Javascript, we allow users to pass in any type of data for the `body`.
We have a default encode/decode behavior which the user can override as well.

jsquire: I've no objection, and think there's value in this.   That said, we need to have a consensus around the naming for it.     ` BodyAsString` feels on the awkward side to me, though I can't think of anything that I like better.

@ramya-rao-a:  As the arbiter of all things Event Hubs, is this the name that we should go with, or do you want to raise it up before we officially take the change?

jsquire: @ramya-rao-a :  If we're taking this across the board, could we either have confirmation in a Teams thread with the requested API shape or (preferably) in an issue for each language so that we can plan for it?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorBuilder.java
```diff
@@ -0,0 +1,187 @@
...
+    private EventHubAsyncClient eventHubAsyncClient;
+    private ProcessEventConsumer processEvent;
+
+    /**
+     * Sets the Event Hub client the {@link EventProcessor} will use to connect to the Event Hub for consuming events.
+     *
+     * @param eventHubAsyncClient The Event Hub asynchronous client for consuming events.
+     * @return The updated {@link EventProcessorBuilder} instance.
+     */
+    public EventProcessorBuilder eventHubClient(EventHubAsyncClient eventHubAsyncClient) {
```
conniey: For the EventHubClient, we could pass in an overload that accepts the synchronous one? It wouldn't matter to us since we can access that underlying async client internally, but it'll make it so that users don't have to construct another if they want to synchronously send.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorBuilder.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**
+     * Sets the callback for processing new events as they are received from all partitions this {@link EventProcessor}
+     * is responsible for. Instance of a {@link CheckpointManager} is also provided for updating checkpoints.
+     *
+     * @param processEvent A callback for processing new events as they are received.
+     * @return The updated {@link EventProcessorBuilder} instance.
+     */
+    public EventProcessorBuilder processEvent(ProcessEventConsumer processEvent) {
```
ramya-rao-a: @srnagar As discussed offline, lets pull the simple `processEvent` that does not allow user to checkpoint out of the current preview and keep only the factory

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionProcessor.java
```diff
@@ -3,29 +3,67 @@
...
+    private final PartitionContext partitionContext;
+    private final CheckpointManager checkpointManager;
+
+    /**
+     * Creates a new instance of PartitionProcessor with the given partition context and checkpoint manager
+     *
+     * @param partitionContext The partition information specific to this PartitionProcessor instance.
+     * @param checkpointManager The checkpoint manager this PartitionProcessor can use to update checkpoints.
+     */
+    public PartitionProcessor(PartitionContext partitionContext, CheckpointManager checkpointManager) {
```
ramya-rao-a: As discussed offline, please consider
- passing partition context in the individual methods instead of the constructor
- having the updateCheckpoint() methods in the partition context and not exposing CheckpointManager to the user

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorBuilder.java
srnagar: As discussed, deferring it for later. Synchronous client may lead the user to believe that the event processor is going to use synchronous APIs even though we can get access to the async client underneath. We can discuss this option for P4.

eng/spotbugs-aggregate-report/pom.xml
```diff
@@ -59,6 +59,7 @@
                 <source>..\..\sdk\core\azure-core\src\samples\java</source>
                 <source>..\..\sdk\core\azure-core-amqp\src\main\java</source>
                 <source>..\..\sdk\core\azure-core-http-netty\src\main\java</source>
+                <source>..\..\sdk\core\azure-core-http-okhttp\src\main\java</source>
```
conniey: sort in alphabetical order.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionProcessor.java
```diff
@@ -70,32 +44,36 @@ public CheckpointManager checkpointManager() {
...
      * This method is called when an error occurs while receiving events from Event Hub. An error also marks the end of
      * event data stream.
      *
+     * @param partitionContext The partition information where the error occurred.
      * @param throwable The {@link Throwable} that caused this method to be called.
      */
-    public void processError(Throwable throwable) {
-        logger.warning("Error occurred in partition processor for partition {} ", partitionContext.partitionId());
+    public void processError(PartitionContext partitionContext, Throwable throwable) {
+        logger.warning("Error occurred in partition processor for partition {} ", partitionContext.partitionId(),
```
conniey: Does this get rendered properly? iirc the underlying slf4j .warn method does not take a varargs and then a throwable.

srnagar: This is not slf4j issue. Adding exception as the last param in slf4j `warn` method works as expected. But our `ClientLogger` removes the exception for any log level other than `verbose`. I am not sure why we do that but I prefer to have the users of `ClientLogger` pass in the exception which can contain useful information and we may change the `ClientLogger` to not remove exceptions later.

PR #5094, opened 2019-08-23
Update Swagger for Blobs and Files

alzimmermsft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/AppendBlobsImpl.java
```diff
@@ -25,19 +25,18 @@
 import com.azure.storage.blob.models.AppendBlobsCreateResponse;
 import com.azure.storage.blob.models.AppendPositionAccessConditions;
 import com.azure.storage.blob.models.BlobHTTPHeaders;
-import com.azure.storage.blob.models.CustomerProvidedKeyInfo;
+import com.azure.storage.blob.models.CpkInfo;
```
tg-msft: Don't block the PR on this, but we should make sure we change this name back at some point.

sdk/storage/azure-storage-blob/swagger/README.md
```diff
@@ -881,3 +881,33 @@ directive:
         "public final class BlobHierarchyListSegment {",
         "@JsonDeserialize(using = CustomHierarchicalListingDeserializer.class)\npublic final class BlobHierarchyListSegment {");
 ```
+
+
+### Replace ByteBuf with ByteBuffer
```
tg-msft: Glad this worked.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/AppendBlobsImpl.java
alzimmermsft: This would be a change in the Swagger itself.

PR #5102, opened 2019-08-26
Enabling Cosmos for Unified Pipelines

mitchdenny:
This PR converts the existing Cosmos PR validation into the full unified pipeline model. One of the things that I anticipated when I introduced the ```archetype-sdk-client.yml``` template names was that we would have builds that break the mold and we would need other files. In these scenarios we introduce another template with the name ```[service]-sdk-client.yml```. This file is mostly the same as the archetype template but has any additional ingredients embedded. This means that we can mostly push down the complexity of a particular service into a seperate file and not clutter up the ```ci.yml``` file.

That is why this change includes a new template file which is referenced by the ```sdk/cosmos/ci.yml``` file - everything else is that file is the same, but the ```eng/pipelines/templates/stages/cosmos-sdk-client.yml``` file contains all the complexity that had shown up in the original version of the ```ci.yml``` file.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
@@ -22,7 +22,9 @@ parameters:
 jobs:
   - job: ${{ parameters.TestName }}
     timeoutInMinutes: ${{ parameters.TimeoutInMinutes }}
-
+    
+    variables:
+      - template: ../variables/globals.yml
     strategy:
       matrix: ${{ parameters.Matrix }}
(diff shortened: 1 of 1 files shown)


mitchdenny:
@HarshaNalluru pinging you on this PR because this is an example of how I am going to "normalize" the cosmos pipeline in the JS repo so I wanted to give you a heads up since the JS repo is probably that next one I'll tackle after finishing the Java repo.

mitchdenny:
Good call. Should probably be a more localised variable

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #5103, opened 2019-08-26
Repo restructure, start with keyvault

weidongxu-microsoft:
Restructure Java SDK repository according to https://azure.github.io/azure-sdk/policies_repostructure.html

Updated keyvault service.
E.g. from
$(azure-libraries-for-java-folder)/keyvault/resource-manager/v2016_10_01
to
$(azure-libraries-for-java-folder)/sdk/keyvault/azure-mgmt-keyvault/v2016_10_01


mitchdenny:
@weidongxu-microsoft thanks for picking up this work! We discussed previously about path length issues. We had a chat about this in our team room and we've been debating the pros and cons of forcing developers on Windows to turn on long path support which is imperfect and doing the work in the repo structure to make it a non-issue given a reasonable Git working directory location on disk.

Then I think that there is an open question about how we want to do PR and CI validation. We could continue to keep the management plane and the client builds completely separate, but there is an opportunity here (in this repo at least) to consolidate them together.

There are a few factors there, we wouldn't want the management plane builds to adversely slow down the client library builds and vice versa. The unified pipelines we are using in the Java repo can actually publish all the way to Maven Central now, although the number of pipelines we have is fairly narrowly scoped compared to the ARM API surface that you guys cover.

weidongxu-microsoft:
@mitchdenny 
Yes, as you correctly noted, the CI configuration will need to be adjusted as well, upon restructure. We will separate client/management and adjust pipeline accordingly.

I think at current stage we probably do not want to public mgmt library automatically. The autorest.java still have bug, and mgmt lib does not have meaningful tests yet. I guess maintainer had to manually test the new code before releasing.
I think it could be helpful to have a template test automation (probably the same record/playback) built in with mgmt package (though it be empty right now), so that maintainer would have at least some ease to add basic test next time they upgrade the mgmt lib.

weidongxu-microsoft:
@mitchdenny @JonathanGiles 

keyvault should be ready.

mgmt pipeline is maintained as separated with client/date pipeline (I will further verify it after pipeline yaml get merged, since paths filter depends on master content).

1. Change output-folder in [swagger spec](https://github.com/Azure/azure-rest-api-specs/) java
2. Move mgmt SDK to location inside `sdk` and conform to [Policies: Repository Structure](https://azure.github.io/azure-sdk/policies_repostructure.html)
3. Update pom under `profiles/2018-03-01-hybrid` and `profiles/2019-03-01-hybrid`, and pom inside mgmt SDK
4. Update ci.yaml to exclude mgmt SDK from client/data pipeline
5. Add test if doable (might need to update pom)
6. Raise PR and verify build pass and mgmt pipeline contains this mgmt SDK in new location

eng/pipelines/mgmt.yml
```diff
@@ -3,16 +3,27 @@ trigger:
...
     - master
   paths:
-    exclude:
-    - sdk/
+    include:
+    - '/pom.management.xml'
+    - '/eng/'
+    - '/parent/'
+    - '/profiles/'
+    - '*/resource-manager/*'
```
mitchdenny: This kind of include filter won't work. It only works by matching the front part of the string with the changes in the commit/PR. Globbing syntax isn't supported (sadly).

mitchdenny:
This looks reasonable to me - thanks again for looking at this @weidongxu-microsoft. I wanted to loop @JimSuplizio who is also working in the Java repo quite a bit from an engineering systems team perspective since this will likely impact the approach he is taking to some of his work.

eng/pipelines/mgmt.yml
weidongxu-microsoft: Got it. I will double check. Alternatively I will try to figure out some way to fast abort the CI if change is not related to management SDK.

weshaggard: I'm not sure how much more work it will take but I'd like to see the mgmt build merged with the client/data-plane build ci.yml approach so that we can simply filter by service directory instead of trying to manage these path filters like this.

weidongxu-microsoft: Yes, this is another option.
One concern is that pipeline seems to be created manually (one for each yaml added), at present, and the number of management service build is far greater than client/data-plane build.
Another concern is that this mgmt.yaml pipeline was added quite recently hence I am a bit reluctant to move away from it.

weshaggard: @mitchdenny has some tools he can share to help with creating and management of the pipelines.  While the mgmt.yml file was recently created and it was a close translation from what we used to have in jenkins it was always a goal to converge the engineering systems between mgmt and data.

weidongxu-microsoft: Got the background. I could do this.

mitchdenny: @weidongxu-microsoft check out the pipeline generator tool in https://github.com/azure/azure-sdk-tools. We use this to automatically generate all the Java and JS pipelines for both internal and external builds. I think getting the management plane builds running might just involve adding the management plane libraries to pom.service.xml in the appropriate directories - then it should "just work".

There are things we would need to work out around logistics, for example would management plane like to release on the same cadence as the client libraries. In some ways that makes sense because if there is a new feature that management plane has been modified to support, then shouldn't there be corresponding support in the client libraries (if they exist - they won't in many cases).

I've got an idea. Why don't we get this change over onto a branch called ```feature/something``` and then we can play around with some of these ideas? That branch would include the restructure work you've got here.

weidongxu-microsoft:
@joshfree

CODEOWNER updated. Later entry should take effect.

eng/pipelines/mgmt.yml
weidongxu-microsoft: Agreed. At present state mgmt CI cannot reuse eng\pipelines\templates\stages\archetype-sdk-client.yml directly, hence I will make a feature branch to try out the CI.

weshaggard:
@weidongxu-microsoft what is the plan here? Do you intend to open up a new PR with these changes?

weidongxu-microsoft:
@weshaggard 
Yes, opened https://github.com/Azure/azure-sdk-for-java/pull/5378 on feature/ci-mgmt branch to continue the work.
feature branch would enable us to trigger CI very much alike that on master. However testing on CI configuration is still tricking so I might even start something on local mocked repo so I could do CI triggering freely.

Issue #4009, opened 2019-06-19
[FEATURE REQ] Allow Passing Custom Thread Factory to Blob Store HTTP Client


original-brownbear:
**Is your feature request related to a problem? Please describe.**

We are trying to integrate Azure blob storage SDK v11 into the Elasticsearch Azure repository plugin in https://github.com/elastic/elasticsearch/issues/43309. One challenge we ran into is the fact that the SDK v11 does require certain permissions that our security manager setup does not grant by default.
One example would be Jackson's `"java.lang.RuntimePermission" "accessDeclaredMembers"`:

```
Caused by: com.microsoft.rest.v2.RestException: HTTP response has malformed headers
   at com.microsoft.rest.v2.protocol.HttpResponseDecoder.decode(HttpResponseDecoder.java:68)
   at com.microsoft.rest.v2.policy.DecodingPolicyFactory$DecodingPolicy.lambda$sendAsync$0(DecodingPolicyFactory.java:34)
   at io.reactivex.internal.operators.single.SingleFlatMap$SingleFlatMapCallback.onSuccess(SingleFlatMap.java:76)
   at io.reactivex.internal.operators.single.SingleDoOnSuccess$DoOnSuccess.onSuccess(SingleDoOnSuccess.java:60)
   at io.reactivex.internal.operators.single.SingleDoOnError$DoOnError.onSuccess(SingleDoOnError.java:52)
   at io.reactivex.internal.operators.single.SingleCreate$Emitter.onSuccess(SingleCreate.java:67)
   at com.microsoft.rest.v2.http.NettyClient$HttpClientInboundHandler.channelRead(NettyClient.java:925)
   at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:374)
   at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:360)
   at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:352)
   at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.fireChannelRead(CombinedChannelDuplexHandler.java:438)
   at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:323)
   at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:297)
[...]

PR #5104, opened 2019-08-26
Allow instantiation of HttpClient sub-types

JonathanGiles:
This introduces a builder API for creating a NettyAsyncHttpClient instances. The next step would be to provide API in the client libraries to accept an HttpClient instance, so that people may customise their HTTP clients.

This may also help to resolve issues such as #4009.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
@@ -11,6 +11,7 @@ import com.azure.core.http.HttpPipelineNextPolicy
 import com.azure.core.http.HttpRequest
 import com.azure.core.http.HttpResponse
 import com.azure.core.http.ProxyOptions
+import com.azure.core.http.netty.NettyAsyncHttpClientBuilder
 import com.azure.core.http.policy.HttpLogDetailLevel
 import com.azure.core.http.policy.HttpPipelinePolicy
 import com.azure.core.http.rest.Response
@@ -381,24 +382,17 @@ class APISpec extends Specification {
     }
 
     private HttpClient getHttpClient() {
-        HttpClient client
+        NettyAsyncHttpClientBuilder builder = new NettyAsyncHttpClientBuilder()
         if (testMode == TestMode.RECORD) {
-            client = HttpClient.createDefault().wiretap(true)
-        } else {
-            client = interceptorManager.getPlaybackClient()
-        }
+            builder.wiretap(true)
 
-        if (Boolean.parseBoolean(ConfigurationManager.getConfiguration().get("AZURE_TEST_DEBUGGING"))) {
-            return client.proxy(PROXY_OPTIONS)
-        } else {
-            return client
-        }
-    }
+            if (Boolean.parseBoolean(ConfigurationManager.getConfiguration().get("AZURE_TEST_DEBUGGING"))) {
+                builder.proxy(new ProxyOptions(ProxyOptions.Type.HTTP, new InetSocketAddress("localhost", 8888)))
+            }
 
-    private Supplier<ProxyOptions> PROXY_OPTIONS = new Supplier<ProxyOptions>() {
-        @Override
-        ProxyOptions get() {
-            return new ProxyOptions(ProxyOptions.Type.HTTP, new InetSocketAddress("localhost", 8888))
+            return builder.build();
+        } else {
+            return interceptorManager.getPlaybackClient()
         }
     }
 


eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -121,6 +121,8 @@
   <!-- Suppress warnings for Event Processor until the usage of "Client" is discussed and resolved: https://github.com/Azure/azure-sdk/issues/321 -->
   <suppress checks="com.azure.tools.checkstyle.checks.ServiceClientBuilderCheck" files=".*EventHubClientBuilder.java"/>
 
+  <suppress checks="com.azure.tools.checkstyle.checks.ServiceClientBuilderCheck" files="com.azure.core.http.netty.NettyAsyncHttpClientBuilder.java"/>
```
JonathanGiles: @mssfang The checkstyle rule here is requiring that I suppress this issue, because it requires that all builders have the `@ServiceClientBuilder` annotation. We should find a way to make this check more specific or less demanding, as there will be cases such as this where it is a builder but not a service client. Can you please give it some thought and get back to me?

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+    public NettyAsyncHttpClientBuilder() {    }
+
+    /**
+     *
+     * @return A new NettyAsyncHttpClient instance
+     * @throws IllegalStateException If proxy type is unknown.
+     */
+    public NettyAsyncHttpClient build() {
+        HttpClient nettyHttpClient = HttpClient.create()
+            .port(port)
```
JonathanGiles: @alzimmermsft Can you please validate that this is sane when the user does not specify a port, and therefore the port value is zero. Thanks!

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+    public NettyAsyncHttpClientBuilder port(int port) {
+        this.port = port;
+        return this;
+    }
+
+    /**
+     * Sets the NIO event loop group that will be used to handle IO.
+     *
+     * @param nioEventLoopGroup the thread factory
+     * @return a HttpClient with the NIO event loop group applied
```
JonathanGiles: All of these return comments are incorrect - it isn't actually an HttpClient, just a builder configured to return one. @alzimmermsft Did you want to do a pass on docs?

JonathanGiles:
@alzimmermsft Could you please add a checkstyle suppression so that the following issue does not show up. This is one of the only places that external dependencies can be exposed! :-) Thanks.

`NettyAsyncHttpClientBuilder.java:104:58: Class 'NioEventLoopGroup', is a class from external dependency. You should not use it as a method argument type. [ExternalDependencyExposed]`

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+                ProxyProvider.Proxy nettyProxy;
+                switch (proxyOptions.type()) {
+                    case HTTP: nettyProxy = ProxyProvider.Proxy.HTTP; break;
+                    case SOCKS4: nettyProxy = ProxyProvider.Proxy.SOCKS4; break;
+                    case SOCKS5: nettyProxy = ProxyProvider.Proxy.SOCKS5; break;
+                    default:
+                        throw logger.logExceptionAsWarning(new IllegalStateException("Unknown Proxy type '" + proxyOptions.type() + "' in use. Not configuring Netty proxy."));
+                }
+
+                return tcpConfig.proxy(ts -> ts.type(nettyProxy).address(proxyOptions.address()));
```
JonathanGiles: To avoid a future bug, we should move the configuration of the proxy into the if block above by inverting the null check.

JonathanGiles:
There is still documentation to come, but marking this as ready for review for people to give their feedback on the changes, and to also kick off the CI builds.

rickle-msft:
This looks fine to me as a draft

JonathanGiles:
Docs are done. Please review now. Thanks.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
```diff
@@ -21,55 +21,49 @@
...
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.util.Objects;
 import java.util.function.BiFunction;
-import java.util.function.Function;
-import java.util.function.Supplier;
 
 /**
- * HttpClient that is implemented using reactor-netty.
+ * This class implements the {@link HttpClient} interface with a Netty-based implementation. Creating an instance of
```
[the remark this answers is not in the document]
JonathanGiles: Will change.

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpClient.java
```diff
@@ -28,28 +26,4 @@
...
         return HttpClientProviders.createInstance();
     }
-
-    /**
-     * Apply the provided proxy configuration to the HttpClient.
-     *
-     * @param proxyOptions the proxy configuration supplier
-     * @return a HttpClient with proxy applied
-     */
-    HttpClient proxy(Supplier<ProxyOptions> proxyOptions);
```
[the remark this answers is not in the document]
JonathanGiles: We have now gone to a place where the HttpClient itself is immutable. This configuration is now set in the builder of the HttpClient instead.

JonathanGiles:
> Looks good. :) Any unit tests to ensure the properties are set correctly would be nice.

Somewhat annoyingly, the Reactor-Netty HttpClient (not our wrapper around it) does not expose the configuration of, e.g., port, wiretap, etc. Once it is set, you cannot determine what it was set to. I don't think I have any ability to confirm that the builder is properly configuring the reactor-netty HttpClient, so I don't have any tests yet. I could test that the builder is storing the values properly, but I'm not sure that is super useful.

Issue #5066, opened 2019-08-20
Final check on javadoc and logging for File


sima-zhu:

sima-zhu:
Added extra space between each block in groovy tests.

PR #5108, opened 2019-08-26
Implement unexpectlengthexception in azure core

sima-zhu:
Make changes in azure core. Added tests in storage.

Will make up more tests in azure core unit tests.

### DIFF
--- a/sdk/storage/azure-storage-file/src/test/resources/session-records/FileAPITestsUploadDataLengthMismatch0.json
+++ b/sdk/storage/azure-storage-file/src/test/resources/session-records/FileAPITestsUploadDataLengthMismatch0.json
@@ -1,65 +1,68 @@
 {
   "networkCallRecords" : [ {
     "Method" : "PUT",
-    "Uri" : "https://azstoragesdkaccount.file.core.windows.net/fileapitestsuploaddatalengthmismatch0088664c40f869f?restype=share",
-    "Headers" : {
-      "x-ms-version" : "2019-02-02",
-      "User-Agent" : "azsdk-java-azure-storage-file/12.0.0-preview.3 1.8.0_221; Windows 10 10.0"
-    },
-    "Response" : {
-      "x-ms-version" : "2019-02-02",
-      "Server" : "Windows-Azure-File/1.0 Microsoft-HTTPAPI/2.0",
-      "ETag" : "\"0x8D73090253F6BBB\"",
-      "Last-Modified" : "Tue, 03 Sep 2019 16:59:55 GMT",
-      "retry-after" : "0",
-      "Content-Length" : "0",
-      "StatusCode" : "201",
-      "x-ms-request-id" : "49c06a7b-201a-00a9-7179-62e59e000000",
-      "Date" : "Tue, 03 Sep 2019 16:59:54 GMT",
-      "x-ms-client-request-id" : "ab01920a-fc56-429b-a3a8-b8db352aad3f"
-    },
-    "Exception" : null
-  }, {
-    "Method" : "PUT",
-    "Uri" : "https://azstoragesdkaccount.file.core.windows.net/fileapitestsuploaddatalengthmismatch0088664c40f869f/fileapitestsuploaddatalengthmismatch062054dafc6056d",
-    "Headers" : {
-      "x-ms-version" : "2019-02-02",
-      "User-Agent" : "azsdk-java-azure-storage-file/12.0.0-preview.3 1.8.0_221; Windows 10 10.0"
-    },
-    "Response" : {
-      "x-ms-version" : "2019-02-02",
-      "x-ms-file-permission-key" : "2189138948451189390*13168958551941065216",
-      "x-ms-file-id" : "13835128424026341376",
-      "Server" : "Windows-Azure-File/1.0 Microsoft-HTTPAPI/2.0",
-      "x-ms-file-creation-time" : "2019-09-03T16:59:55.8095491Z",
-      "Last-Modified" : "Tue, 03 Sep 2019 16:59:55 GMT",
-      "retry-after" : "0",
-      "StatusCode" : "201",
-      "x-ms-request-server-encrypted" : "true",
-      "Date" : "Tue, 03 Sep 2019 16:59:54 GMT",
-      "ETag" : "\"0x8D7309025757E83\"",
-      "x-ms-file-attributes" : "Archive",
-      "x-ms-file-change-time" : "2019-09-03T16:59:55.8095491Z",
-      "x-ms-file-parent-id" : "0",
-      "Content-Length" : "0",
-      "x-ms-request-id" : "49c06a7e-201a-00a9-7279-62e59e000000",
-      "x-ms-client-request-id" : "3cde7104-3d8d-4ae2-8d07-3ad4a2d02350",
-      "x-ms-file-last-write-time" : "2019-09-03T16:59:55.8095491Z"
-    },
-    "Exception" : null
-  }, {
-    "Method" : "PUT",
-    "Uri" : "https://azstoragesdkaccount.file.core.windows.net/fileapitestsuploaddatalengthmismatch0088664c40f869f/fileapitestsuploaddatalengthmismatch062054dafc6056d?comp=range",
+    "Uri" : "https://sima.file.core.windows.net/fileapitestsuploaddatalengthmismatch0086095fc04acf8?restype=share",
     "Headers" : {
       "x-ms-version" : "2019-02-02",
       "User-Agent" : "azsdk-java-azure-storage-file/12.0.0-preview.3 1.8.0_221; Windows 10 10.0",
+      "x-ms-client-request-id" : "411538b5-1c19-4266-a02e-cdad97ad2842"
+    },
+    "Response" : {
+      "x-ms-version" : "2019-02-02",
+      "Server" : "Windows-Azure-File/1.0 Microsoft-HTTPAPI/2.0",
+      "ETag" : "\"0x8D7318D7D4CD061\"",
+      "Last-Modified" : "Wed, 04 Sep 2019 23:13:25 GMT",
+      "retry-after" : "0",
+      "Content-Length" : "0",
+      "StatusCode" : "201",
+      "x-ms-request-id" : "8b41ae39-e01a-000c-1476-632163000000",
+      "Date" : "Wed, 04 Sep 2019 23:13:25 GMT",
+      "x-ms-client-request-id" : "411538b5-1c19-4266-a02e-cdad97ad2842"
+    },
+    "Exception" : null
+  }, {
+    "Method" : "PUT",
+    "Uri" : "https://sima.file.core.windows.net/fileapitestsuploaddatalengthmismatch0086095fc04acf8/fileapitestsuploaddatalengthmismatch088578ada9488e0",
+    "Headers" : {
+      "x-ms-version" : "2019-02-02",
+      "User-Agent" : "azsdk-java-azure-storage-file/12.0.0-preview.3 1.8.0_221; Windows 10 10.0",
+      "x-ms-client-request-id" : "68e6301e-e84e-4f04-b318-39323ecd2e7f"
+    },
+    "Response" : {
+      "x-ms-version" : "2019-02-02",
+      "x-ms-file-permission-key" : "9147088710610530974*15442866184092965392",
+      "x-ms-file-id" : "13835128424026341376",
+      "Server" : "Windows-Azure-File/1.0 Microsoft-HTTPAPI/2.0",
+      "x-ms-file-creation-time" : "2019-09-04T23:13:26.2920881Z",
+      "Last-Modified" : "Wed, 04 Sep 2019 23:13:26 GMT",
+      "retry-after" : "0",
+      "StatusCode" : "201",
+      "x-ms-request-server-encrypted" : "true",
+      "Date" : "Wed, 04 Sep 2019 23:13:25 GMT",
+      "ETag" : "\"0x8D7318D7D8F80B1\"",
+      "x-ms-file-attributes" : "Archive",
+      "x-ms-file-change-time" : "2019-09-04T23:13:26.2920881Z",
+      "x-ms-file-parent-id" : "0",
+      "Content-Length" : "0",
+      "x-ms-request-id" : "8b41ae47-e01a-000c-1576-632163000000",
+      "x-ms-client-request-id" : "68e6301e-e84e-4f04-b318-39323ecd2e7f",
+      "x-ms-file-last-write-time" : "2019-09-04T23:13:26.2920881Z"
+    },
+    "Exception" : null
+  }, {
+    "Method" : "PUT",
+    "Uri" : "https://sima.file.core.windows.net/fileapitestsuploaddatalengthmismatch0086095fc04acf8/fileapitestsuploaddatalengthmismatch088578ada9488e0?comp=range",
+    "Headers" : {
+      "x-ms-version" : "2019-02-02",
+      "User-Agent" : "azsdk-java-azure-storage-file/12.0.0-preview.3 1.8.0_221; Windows 10 10.0",
+      "x-ms-client-request-id" : "9229d03e-c16e-4110-8a24-e3e639e12e43",
       "Content-Type" : "application/octet-stream"
     },
     "Response" : null,
     "Exception" : {
-      "ClassName" : "com.azure.core.implementation.exception.UnexpectedLengthException",
+      "ClassName" : "com.azure.core.exception.UnexpectedLengthException",
       "ErrorMessage" : "Request body emitted 7 bytes more than the expected 6 bytes."
     }
   } ],
-  "variables" : [ "fileapitestsuploaddatalengthmismatch0088664c40f869f", "fileapitestsuploaddatalengthmismatch062054dafc6056d" ]
+  "variables" : [ "fileapitestsuploaddatalengthmismatch0086095fc04acf8", "fileapitestsuploaddatalengthmismatch088578ada9488e0" ]
 }
--- a/sdk/storage/azure-storage-file/src/test/resources/session-records/FileAPITestsUploadDataLengthMismatch1.json
+++ b/sdk/storage/azure-storage-file/src/test/resources/session-records/FileAPITestsUploadDataLengthMismatch1.json
@@ -1,65 +1,68 @@
 {
   "networkCallRecords" : [ {
     "Method" : "PUT",
-    "Uri" : "https://azstoragesdkaccount.file.core.windows.net/fileapitestsuploaddatalengthmismatch156914517d63eba?restype=share",
-    "Headers" : {
-      "x-ms-version" : "2019-02-02",
-      "User-Agent" : "azsdk-java-azure-storage-file/12.0.0-preview.3 1.8.0_221; Windows 10 10.0"
-    },
-    "Response" : {
-      "x-ms-version" : "2019-02-02",
-      "Server" : "Windows-Azure-File/1.0 Microsoft-HTTPAPI/2.0",
-      "ETag" : "\"0x8D7309025C5C75A\"",
-      "Last-Modified" : "Tue, 03 Sep 2019 16:59:56 GMT",
-      "retry-after" : "0",
-      "Content-Length" : "0",
-      "StatusCode" : "201",
-      "x-ms-request-id" : "d333a65e-d01a-0130-1779-622c09000000",
-      "Date" : "Tue, 03 Sep 2019 16:59:55 GMT",
-      "x-ms-client-request-id" : "01c43a0d-c7b5-4d93-9f11-b53f3787a846"
-    },
-    "Exception" : null
-  }, {
-    "Method" : "PUT",
-    "Uri" : "https://azstoragesdkaccount.file.core.windows.net/fileapitestsuploaddatalengthmismatch156914517d63eba/fileapitestsuploaddatalengthmismatch103738ab4356fff",
-    "Headers" : {
-      "x-ms-version" : "2019-02-02",
-      "User-Agent" : "azsdk-java-azure-storage-file/12.0.0-preview.3 1.8.0_221; Windows 10 10.0"
-    },
-    "Response" : {
-      "x-ms-version" : "2019-02-02",
-      "x-ms-file-permission-key" : "2189138948451189390*13168958551941065216",
-      "x-ms-file-id" : "13835128424026341376",
-      "Server" : "Windows-Azure-File/1.0 Microsoft-HTTPAPI/2.0",
-      "x-ms-file-creation-time" : "2019-09-03T16:59:56.3881057Z",
-      "Last-Modified" : "Tue, 03 Sep 2019 16:59:56 GMT",
-      "retry-after" : "0",
-      "StatusCode" : "201",
-      "x-ms-request-server-encrypted" : "true",
-      "Date" : "Tue, 03 Sep 2019 16:59:55 GMT",
-      "ETag" : "\"0x8D7309025CDC661\"",
-      "x-ms-file-attributes" : "Archive",
-      "x-ms-file-change-time" : "2019-09-03T16:59:56.3881057Z",
-      "x-ms-file-parent-id" : "0",
-      "Content-Length" : "0",
-      "x-ms-request-id" : "d333a660-d01a-0130-1879-622c09000000",
-      "x-ms-client-request-id" : "79581f67-42d9-4440-a769-4ef25cfcea77",
-      "x-ms-file-last-write-time" : "2019-09-03T16:59:56.3881057Z"
-    },
-    "Exception" : null
-  }, {
-    "Method" : "PUT",
-    "Uri" : "https://azstoragesdkaccount.file.core.windows.net/fileapitestsuploaddatalengthmismatch156914517d63eba/fileapitestsuploaddatalengthmismatch103738ab4356fff?comp=range",
+    "Uri" : "https://sima.file.core.windows.net/fileapitestsuploaddatalengthmismatch199346919df5b70?restype=share",
     "Headers" : {
       "x-ms-version" : "2019-02-02",
       "User-Agent" : "azsdk-java-azure-storage-file/12.0.0-preview.3 1.8.0_221; Windows 10 10.0",
+      "x-ms-client-request-id" : "8c4ace67-c493-4de5-9fbb-8f0d9af8e507"
+    },
+    "Response" : {
+      "x-ms-version" : "2019-02-02",
+      "Server" : "Windows-Azure-File/1.0 Microsoft-HTTPAPI/2.0",
+      "ETag" : "\"0x8D7318D7E0E1A50\"",
+      "Last-Modified" : "Wed, 04 Sep 2019 23:13:27 GMT",
+      "retry-after" : "0",
+      "Content-Length" : "0",
+      "StatusCode" : "201",
+      "x-ms-request-id" : "4093b47a-f01a-005c-3e76-633e6b000000",
+      "Date" : "Wed, 04 Sep 2019 23:13:26 GMT",
+      "x-ms-client-request-id" : "8c4ace67-c493-4de5-9fbb-8f0d9af8e507"
+    },
+    "Exception" : null
+  }, {
+    "Method" : "PUT",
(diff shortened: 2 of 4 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -117,6 +120,9 @@ public SerializerAdapter serializer() {
      * @return a {@link Mono} that emits HttpResponse asynchronously
      */
     public Mono<HttpResponse> send(HttpRequest request, Context contextData) {
+        if(request.body() != null){
+            request.body().map(ByteBuffer::reset);
```
alzimmermsft: Is this happening because the `UnexpectedLengthException` is reading the `ByteBuffer`? If so, is there any way the check could be implemented to not read the ByteBuffer and need it to be reset?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -154,6 +164,78 @@ public Object invoke(Object proxy, final Method method, Object[] args) {
...
+//                    public Publisher<ByteBuffer> apply(final byte[] bytes) {
+//                        return Flux.just(ByteBuffer.wrap(bytes));
+//                    }
+//                });
+//            }
+//        });
+//    }
+
+    private int len(ByteBuffer input) {
+        int result = input.remaining();
```
alzimmermsft: Is this method necessary? Can't we just use `ByteBuffer::remaining` or `input.remaining()` wherever this is being called?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -290,6 +372,8 @@ private HttpRequest configRequest(HttpRequest request, SwaggerMethodParser metho
                 if (!bodyContentString.isEmpty()) {
                     request.body(bodyContentString);
                 }
+            } else if(bodyContentObject instanceof ByteBuffer) {
+                request.body(Flux.just(((ByteBuffer) bodyContentObject)));
```
alzimmermsft: Extra set of parentheses here.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -154,6 +164,78 @@ public Object invoke(Object proxy, final Method method, Object[] args) {
...
 
+    private Flux<ByteBuffer> validateLength(final HttpRequest request, final Flux<ByteBuffer> bbFlux) {
+        if (bbFlux == null) {
+            return Flux.empty();
+        }
+        return Flux.defer(new Supplier<Publisher<ByteBuffer>>() {
+            @Override
+            public Publisher<ByteBuffer> get() {
+                Long expectedLength = Long.valueOf(request.headers().value("Content-Length"));
+                List<Integer> bufferLengthList = new ArrayList<>();
```
alzimmermsft: Why are we buffering the running total to a `List` instead of tracking it in a `long`?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
alzimmermsft: Actually, it looks like we don't do any reading of the `ByteBuffer`, so can this be removed?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -154,6 +164,78 @@ public Object invoke(Object proxy, final Method method, Object[] args) {
...
+                return bbFlux.doOnEach(s -> {
+                    if (s.isOnNext()) {
+                        Long currentTotalLength = Long.valueOf(
+                            bufferLengthList.stream().reduce(Integer::sum).orElse(0)) + s.get().remaining();
```
alzimmermsft: Is it possible for this or the other reduce method running into an overflow issue?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlockBlobClient.java
```diff
@@ -137,20 +137,9 @@ public BlockBlobItem upload(InputStream data, long length) throws IOException {
...
-                while (read < count) {
-                    read += data.read(cache, read, (int) count - read);
-                }
-                return ByteBuffer.wrap(cache);
-            }));
-
+        Flux<ByteBuffer> fbb = Utility.convertStreamToByteBuffer(data, length, BlockBlobAsyncClient.BLOB_DEFAULT_UPLOAD_BLOCK_SIZE);
         Mono<Response<BlockBlobItem>> upload = blockBlobAsyncClient
-            .uploadWithResponse(fbb.subscribeOn(Schedulers.elastic()), length, headers, metadata, accessConditions, context);
+            .uploadWithResponse(fbb, length, headers, metadata, accessConditions, context);
```
alzimmermsft: Did we mean to remove the `subscribeOn` here?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClientBuilder.java
```diff
@@ -114,7 +114,7 @@ public BlobServiceAsyncClient buildAsyncClient() {
             policies.add(new SASTokenCredentialPolicy(sasTokenCredential));
         }
         HttpPolicyProviders.addBeforeRetryPolicies(policies);
-        policies.add(new RequestRetryPolicy(retryOptions));
+       // policies.add(new RequestRetryPolicy(retryOptions));
```
alzimmermsft: Did this mean to be removed? What does the retry policy do if the error is `UnexpectedLengthException`?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AppendBlobAPITest.groovy
```diff
@@ -188,13 +189,11 @@ class AppendBlobAPITest extends APISpec {
...
         exceptionType.isInstance(e)
-
         where:
         data                     | dataSize            | exceptionType
         null                     | defaultDataSize     | NullPointerException
-        defaultInputStream.get() | defaultDataSize + 1 | IndexOutOfBoundsException
-        // TODO (alzimmer): This doesn't throw an error as the stream is larger than the stated size
-        //defaultInputStream.get()    | defaultDataSize - 1 | StorageException
+        defaultInputStream.get() | defaultDataSize + 1 | UnexpectedLengthException
+        defaultInputStream.get()    | defaultDataSize - 1 | UnexpectedLengthException
```
alzimmermsft: nit: Spacing in the data grid

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
```diff
@@ -482,4 +488,25 @@ public static SharedKeyCredential getSharedKeyCredential(HttpPipeline httpPipeli
         }
         return null;
     }
+
+    public static Flux<ByteBuffer> convertStreamToByteBuffer (InputStream data, long length, int blockSize ) {
```
alzimmermsft: Mind fixing the space before the parentheses?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: It is not needed. Good catch. Will remove,

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: Lambda only allow us to take in final variable. I have tried Long/long, both not work

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: Good call.
Make the reduce take in long.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: Removed.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: Removed.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClientBuilder.java
sima-zhu: No. Just for testing. Will uncomment.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
sima-zhu: The method can be ignored in this PR. Will merge in the correct one from playback client one.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlockBlobClient.java
sima-zhu: It removed because previous implementation in azure core. 
Current way of taking request body does not affect by this. Will add it back.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AppendBlobAPITest.groovy
sima-zhu: Done.

sdk/storage/azure-storage-file/pom.xml
```diff
@@ -88,7 +88,11 @@
 
     <sourceDirectory>src/main/java</sourceDirectory>
     <testSourceDirectory>src/test/java</testSourceDirectory>
-
+    <testResources>
+      <testResource>
+        <directory>${basedir}/src/test/resources</directory>
```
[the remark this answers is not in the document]
sima-zhu: This is how groovy use the test resource folder. I did not see any file operation in blob. @alzimmermsft do you know how you do this in blob groovy test?

sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/spock/FileAPITests.groovy
```diff
@@ -195,6 +191,23 @@ class FileAPITests extends APISpec {
         FileTestHelper.assertExceptionStatusCodeAndMessage(e, 416, StorageErrorCode.INVALID_RANGE)
     }
 
+    @Unroll
+    def "Upload data length mismatch"() {
```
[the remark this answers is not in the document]
sima-zhu: Thanks! A little annoying if I have this in this PR. Will do in the issue.
https://github.com/Azure/azure-sdk-for-java/issues/5066

sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/spock/FileAsyncAPITests.groovy
```diff
@@ -141,6 +142,26 @@ class FileAsyncAPITests extends APISpec {
         defaultData.clear()
     }
 
+    @Unroll
+    def "Upload data length mismatch"() {
```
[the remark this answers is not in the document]
sima-zhu: https://github.com/Azure/azure-sdk-for-java/issues/5066

sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/spock/FileTestHelper.groovy
```diff
@@ -202,10 +204,15 @@ class FileTestHelper {
 
     static String createRandomFileWithLength(int size, String folder, String fileName) {
         def path = Paths.get(folder)
+        if (path == null) {
+            logger.logExceptionAsError("The folder path does not exist.")
```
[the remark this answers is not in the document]
sima-zhu: Good catch. Fixed

sdk/storage/azure-storage-file/pom.xml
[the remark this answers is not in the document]
alzimmermsft: Blobs doesn't use any file resources as part of testing, aside from the session-records which the playback framework handles the logic for that. Blobs generates and tears down file creation for the tests that need it.

sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/spock/FileTestHelper.groovy
```diff
@@ -202,10 +204,15 @@ class FileTestHelper {
...
+            return null
```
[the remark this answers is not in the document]
sima-zhu: Fixed and throw

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
```diff
@@ -47,7 +48,8 @@ public Throwable get() {
 
             case "java.net.UnknownHostException":
                 return new UnknownHostException(throwable.getMessage());
-
+//            case "":
```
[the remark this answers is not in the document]
sima-zhu: Fixed it.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -154,6 +160,40 @@ public Object invoke(Object proxy, final Method method, Object[] args) {
...
+    private Flux<ByteBuffer> validateLength(final HttpRequest request) {
+        final Flux<ByteBuffer> bbFlux = request.body();
+        if (bbFlux == null) {
+            return Flux.empty();
+        }
+        return Flux.defer(new Supplier<Publisher<ByteBuffer>>() {
+            @Override
+            public Publisher<ByteBuffer> get() {
+                Long expectedLength = Long.valueOf(request.headers().value("Content-Length"));
+                List<Long> bufferLengthList = new ArrayList<>();
```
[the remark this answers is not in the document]
sima-zhu: I tried. Lambda does not take in non final variable. No easy way to do sum up.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -154,6 +160,40 @@ public Object invoke(Object proxy, final Method method, Object[] args) {
...
+                return bbFlux.doOnEach(s -> {
+                    if (s.isOnNext()) {
+                        Long currentLength = Long.valueOf(s.get().remaining());
+                        Long currentTotalLength = bufferLengthList.stream().reduce(Long::sum).orElse(0L)
+                            + currentLength;
+                        if (currentTotalLength > expectedLength) {
+                            throw new UnexpectedLengthException(
```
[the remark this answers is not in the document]
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -154,6 +160,36 @@ public Object invoke(Object proxy, final Method method, Object[] args) {
...
+        }
+
+        return Flux.defer(() -> {
+            Long expectedLength = Long.valueOf(request.headers().value("Content-Length"));
+            final AtomicLong currentTotalLength = new AtomicLong(0L);
+            return bbFlux.doOnEach(s -> {
+                if (s.isOnNext()) {
+                    ByteBuffer byteBuffer = s.get();
+                    int currentLength = (byteBuffer == null) ? 0 : byteBuffer.remaining();
+                    currentTotalLength.getAndAdd(currentLength);
```
anuchandy: AtomicLong introduces synchronization internally which is not needed when we're in onNext/onComplete. 

The guarantee that a reactive publisher provide is - it will always serialize the call to multiple onNext and final onComplete, so no concurrent execution of onNext means AtomicLong internal synchronization is unnecessary.  

If the reason for using AtomicLong is to enable mutation within the lambda, then we could use a long array with one element 

```
final long[] currentTotalLengh = new long[1];
```

then use `currentTotalLength[0]` wherever needed like:

```
currentTotalLength[0] = currentTotalLength[0] + currentLength;
```

sima-zhu: The AtomicLong locked on multi-thread. Is there any chance one reactor request on client side will invoke different thread?

sima-zhu: I will take the array one as you suggested.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
```diff
@@ -40,6 +46,7 @@
 import java.util.function.Function;
 
 public final class Utility {
+    private static final ClientLogger LOGGER = new ClientLogger(Utility.class);
```
alzimmermsft: This goes against our logging rules of having no static loggers.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/PageBlobAPITest.groovy
```diff
@@ -201,14 +202,14 @@ class PageBlobAPITest extends APISpec {
         bc.uploadPages(new PageRange().start(0).end(PageBlobClient.PAGE_BYTES * 2 - 1), data)
 
         then:
-        def e = thrown(Exception)
+        def e = thrown(exceptionType)
         exceptionType.isInstance(e)
```
alzimmermsft: Don't need this line anymore, just the `thrown(exceptionType)`

sima-zhu: Good catch.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
anuchandy: The onNext can be called from multiple threads depending on how upstream is designed to produce and emit values (e.g. each value computed in separate threads and emitted from same thread). But these onNext  calls never occur concurrently i.e. two invocation on onNext never overlaps, that's the guarantee that reactive manifesto promises.

sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/spock/FileAPITests.groovy
```diff
@@ -195,6 +196,23 @@ class FileAPITests extends APISpec {
...
+        when:
+        primaryFileClient.uploadWithResponse(defaultData, size, 0, null)
+        then:
+        def e = thrown(UnexpectedLengthException)
+        e.getMessage().contains(errMsg)
+        cleanup:
+        defaultData.clear()
+        where:
+        size | errMsg
+        6 | "more than"
```
alzimmermsft: These tests are throwing an `IllegalStateException` instead of `UnexpectedLengthException`

sima-zhu: Do you mean these tsts should throw `IllegalStateException ` ? I don't quite get why.

alzimmermsft:
Need to clean up the unit tests that are failing with a different error than expected.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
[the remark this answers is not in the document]
sima-zhu: Do this as Anu suggested:
`final long[] currentTotalLengh = new long[1];`

sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/spock/FileAPITests.groovy
sima-zhu: Update the json files.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AppendBlobAPITest.groovy
```diff
@@ -264,10 +263,10 @@ class AppendBlobAPITest extends APISpec {
 
         when:
         bc.appendBlockWithResponse(defaultInputStream.get(), defaultDataSize, bac, null, null)
-
```
jaschrep-msft: Nit: Don't delete these blank lines. Having a separation between labeled blocks is really useful for readability.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AppendBlobAPITest.groovy
```diff
@@ -325,6 +325,8 @@ class AppendBlobAPITest extends APISpec {
         ByteArrayOutputStream downloadStream = new ByteArrayOutputStream(1024)
         destURL.download(downloadStream)
         downloadStream.toByteArray() == Arrays.copyOfRange(data, 2 * 1024, 3 * 1024)
+        cleanup:
+        defaultInputStream.get().reset()
```
jaschrep-msft: `defaultInputStream` is a `Supplier<InputStream>` that news up a stream of the bytes of a preset string and just gives us that. These cleanup blocks are just getting a brand new stream only to call `reset()` and discard the reference for the GC to pick up.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AppendBlobAPITest.groovy
sima-zhu: Will revert the changes

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
```diff
@@ -50,7 +51,7 @@ class BlobAPITest extends APISpec {
         HttpHeaders headers = response.headers()
 
         then:
-        body == defaultData
+        defaultData.compareTo(body) == 0
```
jaschrep-msft: This was already idiomatic to groovy. `==` checks to see if `Comparable` is implemented and calls `objA.compareTo(objB) == 0` if so. Otherwise, it calls `objA.equals(objB)`.

(future reference: if you want to compare reference equality in groovy, you write `objA.is(objB)`)

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AppendBlobAPITest.groovy
sima-zhu: Done.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
```diff
@@ -482,4 +488,46 @@ public static SharedKeyCredential getSharedKeyCredential(HttpPipeline httpPipeli
...
+     * entity length and the input length.
+     *
+     * @param data The input data which needs to convert to ByteBuffer.
+     * @param length The expected input data length.
+     * @param blockSize The size of each ByteBuffer.
+     * @return {@link ByteBuffer} which contains the input data.
+     * @throws UnexpectedLengthException when input data length mismatch input length.
+     * @throws RuntimeException When I/O error occurs.
+     */
+    public static Flux<ByteBuffer> convertStreamToByteBuffer(InputStream data, long length, int blockSize) {
```
jaschrep-msft: I feel like this is a common enough case that this should be in core? We can't be the only library moving data and need to do this conversion.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
sima-zhu: Revert the changes back. Thanks for the info!

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AppendBlobAPITest.groovy
sima-zhu: Revert it back.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
sima-zhu: Will discuss with @JonathanGiles if we can put up the helper method in Azure core. Leave it here right now as only storage has this kind of needs.

sima-zhu: Double check the Utility methods. There are bunch  a few which do not specific to storage. We can several of them into Azure core based on the needs of other clients. It is better to limit the scope of the utility to storage before preview 3.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/AppendBlobClient.java
```diff
@@ -159,24 +161,14 @@ public AppendBlobItem appendBlock(InputStream data, long length) {
      * @param context Additional context that is passed through the Http pipeline during the service call.
      *
      * @return A {@link Response} whose {@link Response#value() value} contains the append blob operation.
+     * @throws UnexpectedLengthException when the length of data does not match the input {@code length}.
```
JonathanGiles: You cannot throw UnexpectedLengthException here as it is not public API. You either need to make it public API, or not throw it from public API.

sima-zhu: Put the exception in public exception folder

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -154,6 +159,36 @@ public Object invoke(Object proxy, final Method method, Object[] args) {
...
 
+    private Flux<ByteBuffer> validateLength(final HttpRequest request) {
+        final Flux<ByteBuffer> bbFlux = request.body();
+        if (bbFlux == null) {
+            return Flux.empty();
+        }
+
+        return Flux.defer(() -> {
+            Long expectedLength = Long.valueOf(request.headers().value("Content-Length"));
+            final long[] currentTotalLength = new long[1];
```
anuchandy: I guess what we're doing here is correct. I asked the question in SOF and got a response from David (the main contributor of rx) - https://stackoverflow.com/questions/57776012/mutating-array-elements-in-reactive-publisher-onnext-doonnext-methods/
The last thing to rule out is whether to use volatile keyword to ensure immediate update visibility across thread, waiting for his response.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/PageBlobClient.java
```diff
@@ -170,21 +172,14 @@ public PageBlobItem uploadPages(PageRange pageRange, InputStream body) {
...
-                byte[] cache = new byte[PAGE_BYTES];
-                int read = 0;
-                while (read < PAGE_BYTES) {
-                    read += body.read(cache, read, PAGE_BYTES - read);
-                }
-
-                return ByteBuffer.wrap(cache);
-            }));
+        Objects.requireNonNull(body);
+        long length = pageRange.end() - pageRange.start() + 1;
```
mssfang: nit: final modifier for unmodifiable variable.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
anuchandy: we're good, David confirmed that framework takes care of memory visibility, no need for volatile. https://stackoverflow.com/questions/57776012/mutating-array-elements-in-reactive-publisher-onnext-doonnext-methods/57786588?noredirect=1#comment102032427_57786588

sima-zhu: Thanks Anu for taking care of this! Learned a lot from the thread.

PR #5115, opened 2019-08-26
Avoid buffering for download in storage

jianghaolu:


alzimmermsft:
The DownloadAsyncResponse gets used which handles all the logic for reliable download.

jianghaolu:
Is `DownloadAsyncResponse` replacing `DownloadResponse`? If I misunderstand here I can add it back.

jianghaolu:
The main fix is in the decoder... Removing DownloadResponse.java is just an add-on since I see it's not being used.

PR #5120, opened 2019-08-27
Adding parent POMs to cache key.

mitchdenny:
This PR addresses an issue where the parent POM is being stored in the cache and changes in its contents weren't causing invalidation from build to build resulting in the build not running plugins that were recently configured. Could have just rotated the salt, but this is a more correct approach to have this happen automagically in the future.


mitchdenny:
Ah, I had this in a other PR that required it which I ended up not needing. I'll add this back in.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #5084, opened 2019-08-22
Proposal to create a base type of PagedFlux

JonathanGiles:
This allows other types to be created in cases where a subclass of PagedResponse is used (e.g. Azure Search). Not complete by any means.

PR #5128, opened 2019-08-27
Intorduce PagedFluxBase

g2vinay:
Introduces PagedFluxBase class.
Ability to add single page retriever for non-paginated collection responses.


g2vinay:
Adding this feature via:
https://github.com/Azure/azure-sdk-for-java/pull/5084

PR #5150, opened 2019-08-28
Okhttp HttpClient plugin for azure-core

anuchandy:
This PR introduces an implementation of azure-core HttpClient for OkHttp and tests.

### Design choices:

1. While Netty allows writing request contents chunk by chunk, OkHttp requires consumer to aggregate it. We’ve used okio.buffer type for aggregation which is pooled. See [this](https://github.com/anuchandy/azure-sdk-for-java/blob/b196bd96caafcdaccdcec8d038105b8884028415/sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java#L122).
2. OkHttp response content can be retrieved as Inputstream, bytes[] or String. bytes[] and String options drain the socket, we used InputStream option to have a back pressure aware Flux<ByteBuffer>. See [this](https://github.com/anuchandy/azure-sdk-for-java/blob/b196bd96caafcdaccdcec8d038105b8884028415/sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java#L263).
3. The send(HttpRequest) method itself is back pressure aware, we used MonoSink::onRequest. See [this](https://github.com/anuchandy/azure-sdk-for-java/blob/b196bd96caafcdaccdcec8d038105b8884028415/sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java#L38).
4. We’ve a builder that allows configuring OkHttp settings


sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
```diff
@@ -0,0 +1,320 @@
...
+
+    @Override
+    public HttpClient port(int port) {
+        throw new RuntimeException("Port is property of okhttp3.Request not client, TODO: remove it, once core HttpClient is fixed");
+    }
+
+    /**
+     * An implementation of azure-core HttpResponse for OkHttp.
+     */
+    public static class OkHttpResponse extends HttpResponse {
```
JonathanGiles: This can be private?

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Set the read timeout, default is 120 seconds.
+     *
+     * @param timeout the timeout
+     * @param unit the time unit for the timeout
+     * @return the builder
+     */
+    public OkHttpAsyncHttpClientBuilder readTimeout(long timeout, TimeUnit unit) {
```
JonathanGiles: Make these timeout methods use `Duration` instead, if possible.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+     */
+    public HttpClient build() {
+        if (connectionPool != null) {
+            this.httpClientBuilder = httpClientBuilder.connectionPool(connectionPool);
+        }
+        if (dispatcher != null) {
+            this.httpClientBuilder = httpClientBuilder.dispatcher(dispatcher);
+        }
+        return new OkHttpAsyncHttpClient(this.httpClientBuilder.build());
+    }
```
JonathanGiles: I think this builder pattern needs consideration. For example, it would not be possible for person to create two HttpClient instances from this builder where the first one has a dispatcher and the second one does not, because they are unable to null it out. I would probably remove the constructor that takes the HttpClientBuilder and to only create that in the build() method here.

JonathanGiles: Yes, we should support that

anuchandy:
@JonathanGiles @jianghaolu PR updated, please re-review.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Add network layer interceptors to Http request pipeline.
+     *
+     * @param networkInterceptors the interceptors to add
+     * @return the builder
+     */
+    public OkHttpAsyncHttpClientBuilder networkInterceptors(List<Interceptor> networkInterceptors) {
+        this.networkInterceptors = Objects.requireNonNull(networkInterceptors);
+        return this;
+    }
```
JonathanGiles: I think this should be removed.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
```diff
@@ -0,0 +1,318 @@
...
+     * @return the Mono emitting okhttp request
+     */
+    private static Mono<okhttp3.Request> toOkHttpRequest(HttpRequest request) {
+        return Mono.just(new okhttp3.Request.Builder())
+                .map(rb -> {
+                    rb.url(request.url());
+                    if (request.headers() != null) {
+                        return rb.headers(okhttp3.Headers.of(request.headers().toMap()));
+                    } else {
+                        return rb.headers(okhttp3.Headers.of((new HashMap<>())));
```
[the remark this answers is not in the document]
anuchandy: It's a typo, thanks for catching

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
anuchandy: This is to reuse the builder so that caller can clear previous selection by passing empty list or a list with brand new interceptors, any other idea to handle this use case?

JonathanGiles: You should update the JavaDoc to say that this replaces all previously-set interceptors on the builder.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
```diff
@@ -0,0 +1,320 @@
...
+     *
+     *
+     * @param bbFlux stream of java.nio.ByteBuffer representing request content
+     * @param headers the headers associated with the original request
+     * @return the okhttp3.RequestBody
+     */
+    private static okhttp3.RequestBody toOkHttpRequestBody(Flux<ByteBuffer> bbFlux, HttpHeaders headers) {
+        okio.ByteString byteString = okio.ByteString.EMPTY;
+        if (bbFlux != null) {
+            byteString = aggregate(bbFlux).block();
```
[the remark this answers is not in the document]
jianghaolu: For an async operation I'd prefer it not happening on the caller thread. I agree with you on option 2.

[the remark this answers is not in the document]
anuchandy: Right, the real async operation we care (http call execution) happens in OkHttp thread for both options. 

Even in option2 for one case, block for byte buffer  aggregation happens in caller thread but I wouldn't worry about it as caller has option to have async aggregation by doing subscribeOn. The possible cases are explained below and also added as code comment for future reference.

The blocking behavior `toOkHttpRequest(r).subscribe` call:

The `okhttp3.Request` emitted by `toOkHttpRequest(r)` is chained from the body of request `Flux<ByteBuffer>`:
   1. If `Flux<ByteBuffer>` synchronous and `send(r)` caller does not apply `subscribeOn` then
      subscribe block on caller thread.
   2. If `Flux<ByteBuffer>` synchronous and `send(r)` caller apply `subscribeOn` then
      does not block caller thread but block on scheduler thread.
   3. If `Flux<ByteBuffer>` asynchronous then `Subscribe` does not block caller thread
      but block on the thread backing flux. This ignore any `subscribeOn` applied to `send(r)`

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
```diff
@@ -0,0 +1,318 @@
...
+        @Override
+        public Mono<String> bodyAsString() {
+            if (this.responseBody() == null) {
+                return Mono.empty();
+            } else {
+                return Mono.using(() -> this.responseBody(),
+                        rb -> {
+                            try {
+                                String content = rb.string();
+                                return content.length() == 0 ? Mono.empty() : Mono.just(content);
```
srnagar: Can this instead be `return Mono.just(content)`? Any reason to return `Mono.empty()` instead of an empty string?

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
```diff
@@ -0,0 +1,318 @@
...
+        @Override
+        public Mono<byte[]> bodyAsByteArray() {
+            if (this.responseBody() == null) {
+                return Mono.empty();
+            } else {
+                return Mono.using(() -> this.responseBody(),
+                        rb -> {
+                            try {
+                                byte[] content = rb.bytes();
+                                return content.length == 0 ? Mono.empty() : Mono.just(content);
```
srnagar: Same here - empty array vs `Mono.empty()`. I am not sure if there's a difference between a `null` responseBody vs a non-null responsebody with empty byte array.

Should we instead return `Mono.just(content)` where content could potentially be empty?

sdk/core/azure-core-http-okhttp/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core</artifactId>
+      <version>1.0.0-preview.4</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp</artifactId>
+      <version>4.1.0</version>
```
srnagar: Add this version to parent/pom.xml where all versions are declared as properties. It's easier to maintain all external dependency versions in one place.

sdk/core/azure-core-http-okhttp/src/main/resources/META-INF/services/com.azure.core.implementation.http.spi.HttpClientProvider
```diff
@@ -0,0 +1 @@
+com.azure.core.http.okhttp.implementation.OkHttpClientProvider
```
srnagar: Just for my understanding, what is the purpose of this file?

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Builder to configure and build an implementation of com.azure.core.http.HttpClient for OkHttp.
+ */
+public class OkHttpAsyncHttpClientBuilder {
```
srnagar: Should the builder just be named as `OkHttpClientBuilder` instead of adding `Async`. Other client builders don't add `Async` to the classname.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+    private final ClientLogger logger = new ClientLogger(OkHttpAsyncHttpClientBuilder.class);
```
srnagar: Looks like the logger is not used.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
```diff
@@ -0,0 +1,318 @@
...
+import java.util.HashMap;
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
+ * HttpClient implementation for OkHttp.
+ */
+class OkHttpAsyncHttpClient implements HttpClient {
+    private final ClientLogger logger = new ClientLogger(OkHttpAsyncHttpClient.class);
+    private final okhttp3.OkHttpClient httpClient;
```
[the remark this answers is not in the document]
anuchandy: Done in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222).

sdk/core/azure-core-http-okhttp/src/test/java/com/azure/core/http/okhttp/OkHttpClientTests.java
```diff
@@ -0,0 +1,313 @@
...
+                    .verifyError(IOException.class);
+        } finally {
+            ss.close();
+        }
+    }
+
+    @Ignore("This flakey test fails often on MacOS. https://github.com/Azure/azure-sdk-for-java/issues/4357.")
+    @Test
+    public void testConcurrentRequests() throws NoSuchAlgorithmException {
+//        long t = System.currentTimeMillis();
```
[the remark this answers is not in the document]
anuchandy: Done in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222).

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+import okhttp3.Interceptor;
+import okhttp3.OkHttpClient;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Builder to configure and build an implementation of com.azure.core.http.HttpClient for OkHttp.
```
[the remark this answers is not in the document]
anuchandy: Done in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222).

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+        this.okHttpClient = Objects.requireNonNull(okHttpClient, "okHttpClient == null");
+    }
+
+    /**
+     * Add a network layer interceptor to Http request pipeline.
+     *
+     * @param networkInterceptor the interceptor to add
+     * @return the builder
+     */
+    public OkHttpAsyncHttpClientBuilder networkInterceptor(Interceptor networkInterceptor) {
```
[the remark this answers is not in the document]
anuchandy: Done in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222) for both Netty and OkHttp client builders.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+    public OkHttpAsyncHttpClientBuilder readTimeout(Duration readTimeout) {
+        // readTimeout can be null
+        this.readTimeout = readTimeout;
+        return this;
+    }
+
+    /**
+     * Sets the connection timeout.
+     *
+     * The default read timeout is 60 seconds.
```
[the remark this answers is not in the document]
anuchandy: Done in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222).

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+        this.okHttpClient = null;
+    }
+
+    /**
+     * Creates OkHttpAsyncHttpClientBuilder from the builder of an existing OkHttpClient.
+     *
+     * @param okHttpClient the httpclient
+     */
+    public OkHttpAsyncHttpClientBuilder(okhttp3.OkHttpClient okHttpClient) {
+        this.okHttpClient = Objects.requireNonNull(okHttpClient, "okHttpClient == null");
```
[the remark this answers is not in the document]
anuchandy: Done in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222), added similar message for all null checks.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+        this.readTimeout = readTimeout;
+        return this;
+    }
+
+    /**
+     * Sets the connection timeout.
+     *
+     * The default read timeout is 60 seconds.
+     *
+     * @param connectionTimeout the timeout
```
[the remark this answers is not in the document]
anuchandy: Done in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222).

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+        this.networkInterceptors = Objects.requireNonNull(networkInterceptors);
+        return this;
+    }
+
+    /**
+     * Sets the read timeout.
+     *
+     * The default read timeout is 120 seconds.
+     *
+     * @param readTimeout the timeout
```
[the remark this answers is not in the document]
anuchandy: Done in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222).

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
anuchandy: Not sure about removing "Async" word from builder name, I followed reactory-netty HttpClient builder naming convention. 

Also the naming pattern we follow is: `[productName][Async]HttpClientBuilder` where product name is `Netty`, `OkHttp`. 

With 'Async' removed, the pattern is `[productName]HttpClientBuilder` so `OkHttpHttpClientBuilder`, the word Http repeats. If change it to `OkHttpClientBuilder` as suggested then it is no longer following the pattern. 

@JonathanGiles thoughts?

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
anuchandy: Good question @srnagar :)
 
Q1: – *if there's a difference between a null responseBody ..*:
 
`okhttp3.Response::body()` getter will not return null for server returned responses. It can be null only if we build response manually with null body (e.g. for mocking) or for the cases described here [ref](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response/body/). Those null cases are not applicable for OkHttp HttpClient impl. I added the null check as a defensive check, those might never hit.
 
Q2: - *using Mono.just(content) [where content=new byte[0]] vs Mono.empty()*
 
If content part is absent in a http wire response then content getters in `okhttp3.ResponseBody` instance uses Java empty semantic. i.e. `okhttp3.ResponseBody::bytes()` getter return `new byte[0]`, `okhttp3.ResponseBody::string()` returns "".
 
Similarly Netty uses empty `ByteBuf` i.e. `ByteBuf` of size 0 if content part is absent.
 
In reactive, the pattern of using `Flux::empty()/Mono::empty()` for absence of value is common.
 
If we look at reactor-netty as an example, it translate an empty `ByteBuf` from Netty to `Flux::empty()`.
 
Our reactor-netty azure-core HttpClient plugin follows the same pattern.
 
Any implementation of azure-core HttpClient is expected to follow this pattern, hence okHttp HttpClient plugin.
 
---
 
It looks like I need to clarify above points in the code to make it readable for others. I've opened a [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222)  same behavior but better readability. It also centralize resource management.
 
Thanks for raising this question.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
anuchandy: Removed in the [PR](https://github.com/Azure/azure-sdk-for-java/pull/5222).

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
anuchandy: see below comment

sdk/core/azure-core-http-okhttp/src/main/resources/META-INF/services/com.azure.core.implementation.http.spi.HttpClientProvider
anuchandy: This is the configuration file that `ServiceLoader` API uses to detect `OkHttpClientProvider` as an implementation of `Service Provider`. More info can be found [here](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html).

Issue #4193, opened 2019-06-29
Validate length of flux content matches with Content-Length header 


anuchandy:
**Is your feature request related to a problem? Please describe.**
Storage SDK has blob upload methods where user can provide flux emitting the content and length. The length become Content-Length header. A validation is missing that ensure length of data emitted by flux is equal to length param.

**Describe the solution you'd like**
instead of passing the flux to reactor-netty directly, append operators that does this check and pass updated flux to reactor-netty.

**Describe alternatives you've considered**
Adding this validation in storage SDK side is possible but with lot of  code duplication since we have multiple upload methods.

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ X] Description Added
- [X ] Expected solution specified

anuchandy:
When this validation fails `UnexpectedLengthException`

```java
/**
 * Indicates that a stream emitted an unexpected number of bytes.
 */
public class UnexpectedLengthException extends IllegalStateException {
    private final long bytesRead;
    private final long bytesExpected;

    /**
     * Creates an UnexpectedLengthException.
     * @param message the message
     * @param bytesRead the number of bytes actually read in the stream
     * @param bytesExpected the number of bytes expected to be read in the stream
     */
    public UnexpectedLengthException(String message, long bytesRead, long bytesExpected) {
        super(message);
        this.bytesRead = bytesRead;
        this.bytesExpected = bytesExpected;
    }

    /**
     * @return the number of bytes actually read in the stream
     */
    public long bytesRead() {
        return bytesRead;
    }

    /**
     * @return the number of bytes expected to be read in the stream
     */
    public long bytesExpected() {
        return bytesExpected;
    }
}
```

jianghaolu:
@joshfree this will unblock many Storage tests.

sima-zhu:
Verified the current structure.
RestProxy is the front door in azure core which can take in the request body. 
Will add the validation in RestProxy.

PR #5159, opened 2019-08-29
Serialize and deserialize azure core exception type.

sima-zhu:
**Summary:** 
1. The PR is to generic exception type and deserialize it for client library tests. 
2. Remove the stack trace from record json files.
3. Added the exception to storage blob, uncommon related tests and fix bugs in storage.

**Problem to solve:**
Currently, playback can only deserialize specific exception type using switch case. Moreover, the exception must have default constructor. Try to generic the process so people do not need to continue adding different exceptions. Also, the UnexpectedLengthException asks value constructor, which is also a shortage for the playback. 

**Issue related:**
https://github.com/Azure/azure-sdk-for-java/issues/4193


sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
```diff
@@ -165,4 +171,15 @@ private static String removeHost(String url) {
 
         return String.format("%s%s", urlBuilder.path(), urlBuilder.queryString());
     }
+
+    private void constructExceptionType(final NetworkCallRecord networkCallRecord) throws Exception {
+        if (networkCallRecord.exception() == null) {
+            return;
+        }
+        if (networkCallRecord.exception() != null && networkCallRecord.exception().argTypes() == null) {
```
alzimmermsft: We don't need the not null check, already checked for being null right before this

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
```diff
@@ -28,47 +33,131 @@ public NetworkCallError() {
...
      *
      * @param throwable Throwable thrown during a network call.
+     * @throws RuntimeException when I/O error occurs.
      */
     public NetworkCallError(Throwable throwable) {
-        this.throwable = throwable;
-        this.className = throwable.getClass().getName();
+        try {
+            this.className = throwable.getClass();
+            this.argTypes = throwable.getClass().getConstructors()[0].getParameterTypes();
```
alzimmermsft: Do we always want to target the first constructor or a specific constructor of the Throwable?

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
```diff
@@ -28,47 +33,131 @@ public NetworkCallError() {
...
-     * @param className Class name of the throwable.
+     * @return The array of argument values.
      */
-    public void className(String className) {
-        this.className = className;
+    public Object[] argValues() {
+        return ImplUtils.clone(this.argValues);
+    }
+
+    private Object getDefaultValue(Class<?> type) throws Exception {
```
alzimmermsft: Do we want to let the exception bubble up as is or wrap it in a `RuntimeException`?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClientBuilder.java
```diff
@@ -114,7 +114,7 @@ public BlobServiceAsyncClient buildAsyncClient() {
             policies.add(new SASTokenCredentialPolicy(sasTokenCredential));
         }
         HttpPolicyProviders.addBeforeRetryPolicies(policies);
-        policies.add(new RequestRetryPolicy(retryOptions));
+       // policies.add(new RequestRetryPolicy(retryOptions));
```
alzimmermsft: Why did we comment out this line of code? This is a critical line of code.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlockBlobAPITest.groovy
```diff
@@ -188,6 +188,20 @@ class BlockBlobAPITest extends APISpec {
         uncommittedBlock.hasNext()
     }
 
+    def "test"() {
```
alzimmermsft: This is going to be removed right?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/RetryTest.groovy
```diff
@@ -4,6 +4,7 @@
 package com.azure.storage.blob
 
 import com.azure.core.http.HttpResponse
+import com.azure.core.implementation.exception.UnexpectedLengthException
```
alzimmermsft: Why did this import get added here?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
```diff
@@ -291,6 +292,7 @@ class ServiceAPITest extends APISpec {
         thrown(StorageException)
     }
 
+    @Ignore
```
alzimmermsft: Why are these tests being ignored now?

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
```diff
@@ -28,47 +33,131 @@ public NetworkCallError() {
...
+        // Get default value of non-primitive type
+        if (!type.isPrimitive()) {
+            return type.getConstructor().newInstance();
```
alzimmermsft: Is there always a guarantee for an empty constructor?

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
```diff
@@ -28,47 +33,131 @@ public NetworkCallError() {
...
+     * Gets the throwable argument default values which needed for new instance constructor. This is used during deserialization the construct the throwable
      * as the actual class that was thrown.
      *
-     * @param className Class name of the throwable.
+     * @return The array of argument values.
      */
-    public void className(String className) {
-        this.className = className;
+    public Object[] argValues() {
+        return ImplUtils.clone(this.argValues);
```
alzimmermsft: I don't think this is going to work as you expect, this will clone the array with the same references (aka a shallow copy).

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
sima-zhu: It is checking networkCallRecord being null or not. There is no checking for networkCallRecord.exception()

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
sima-zhu: We usually don't care too much about which constructor to use. Record take in the first one, and playback is able to deserialize the same. What we care about is the exception type.
However, it is better to have null check right here. Will update it,

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
sima-zhu: The exception will be caught in NetworkCallError constructor and throw the error message which help to debug.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClientBuilder.java
sima-zhu: Just for quick expose the error msg. Will uncomment. Good catch.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlockBlobAPITest.groovy
sima-zhu: Done.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/RetryTest.groovy
sima-zhu: This because I removed the UnexpectedLengthException from blob. RequestRetryTestFactory use the one in Azure core now.

```
    def "Retries non replyable flux"() {
        setup:
        RequestRetryTestFactory retryTestFactory = new RequestRetryTestFactory(RequestRetryTestFactory.RETRY_TEST_SCENARIO_NON_REPLAYABLE_FLOWABLE, retryTestOptions)

        when:
        retryTestFactory.send(retryTestURL).block()

        then:
        def e = thrown(IllegalStateException)
        e.getMessage().startsWith("The request failed because")
        e.getCause() instanceof UnexpectedLengthException
    }
```

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
sima-zhu: The record did not work for me. The live tests also failed at `AuthenticationMismatchError`
I will create an issue and link it in code. Issue opened  https://github.com/Azure/azure-sdk-for-java/issues/5172

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
sima-zhu: ImplUtil will do a deep copy.
The class<?> is able to copy to new place. What we need is the class path.
The object[] , primitive is able to copy. if it is non-primitive the default value we use is null.
Copy will not give us wrong result.
Since not every exception expose the getter to the field, I am not able to get real value but the dummy default one to make it work in record and playback.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/models/NetworkCallError.java
sima-zhu: There is no need for non-checking as the this.argTypes is known to be not null.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
sima-zhu: As below.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
alzimmermsft: Right but you check for null and return them check for not null which will always be true.

sima-zhu: Right. I am reading wrong line.
Fixed it.

JonathanGiles:
Can you please update the description to explain what this PR is attempting to achieve, and why. Is there an issue filed that describes this? Please provide more context.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
```diff
@@ -68,11 +73,7 @@ public PlaybackClient(RecordedData recordedData, Map<String, String> textReplace
 
             return Mono.error(new IllegalStateException("==> Unexpected request: " + incomingMethod + " " + incomingUrl));
         }
-
-        if (networkCallRecord.exception() != null) {
-            throw logger.logExceptionAsWarning(Exceptions.propagate(networkCallRecord.exception().get()));
-        }
-
+        constructExceptionType(networkCallRecord);
```
JonathanGiles: You should change this. At present the `constructExceptionType` call is odd, as it may throw an exception but it won't if the `networkCallRecord.exception()` is null. You should reinstitute the if block you removed, and then make constructExceptionType return an exception that is then definitively thrown.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/exception/UnexpectedLengthException.java
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.implementation.exception;
+
+public class UnexpectedLengthException extends IllegalStateException {
```
JonathanGiles: You are making this class non-public API. Do you ever document anywhere in the storage APIs that this is thrown in certain situations?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
JonathanGiles: Can you please provide more context on this - I don't know why they are being ignored either.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/exception/UnexpectedLengthException.java
sima-zhu: Good call. I will update the API which will throw the exception. This is not intended to use by client. It is for http request validation.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
sima-zhu: I have a PR for addressing the issue. 
We can uncomment the ignore once the PR gets in.
https://github.com/Azure/azure-sdk-for-java/pull/5176

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
sima-zhu: the step only takes affect when exception is not null. 
Will add a checking before calling the method.
Thanks for catching this.

JonathanGiles: My suggestion and what you implemented are slightly different. Please re-read and clarify with me if necessary.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
sima-zhu: User delegate key is asking service principle ids.
We use our personal account previously to get the tests passed. However, the live tests are using the team account which use other service principle id and key.
The PR above is to update the key and add missing creds we need to use.
Once the PR got merged in, live tests should pass and I will uncomment here.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlockBlobClient.java
```diff
@@ -134,23 +136,15 @@ public BlockBlobItem upload(InputStream data, long length) throws IOException {
...
-                    read += data.read(cache, read, (int) count - read);
-                }
-                return ByteBuffer.wrap(cache);
-            }));
-
+        Objects.requireNonNull(data);
+        Flux<ByteBuffer> fbb = Utility.convertStreamToByteBuffer(data, length, BlockBlobAsyncClient.BLOB_DEFAULT_UPLOAD_BLOCK_SIZE);
         Mono<Response<BlockBlobItem>> upload = blockBlobAsyncClient
-            .uploadWithResponse(fbb.subscribeOn(Schedulers.elastic()), length, headers, metadata, accessConditions, context);
+            .uploadWithResponse(fbb, length, headers, metadata, accessConditions, context);
```
alzimmermsft: Did we want to remove the `subscribeOn` here?

sima-zhu:
Close it as this is not the approach we take

PR #5162, opened 2019-08-29
Removes parent POM for keyvault track 1 libraries.

mitchdenny:
This PR removes the track 1 parent POM for KeyVault (an intermediate POM). This means that we lift some version numbers into each of the children and change the parent to point to the main track 1 parent POM.


sdk/keyvault/microsoft-azure-keyvault-complete/pom.xml
```diff
@@ -41,22 +41,27 @@ the MIT License. See License.txt in the project root for license information. --
     <dependency>
       <groupId>com.microsoft.azure</groupId>
       <artifactId>azure-keyvault</artifactId>
+      <version>1.2.2</version>
```
g2vinay: we can extract this into a property.

g2vinay: applies to each pom

mitchdenny: For consistency with the way we are specifying internal dependencies within core (just as one example) I think we should leave this as explicit. I think moving to explicit version references will also help when we shift to automated tooling to synchronize versions (because that tooling won't then need to go and interpret properties.

Issue #3923, opened 2019-06-14
[Event Hubs] Replace constant strings with resource strings


conniey:
Replace all the user strings (ie. error strings) with resource strings that can possibly be globalised later.

mclay:
please assign me this task and let me know which modules are affected from this

JonathanGiles:
@conniey Can you please give @mclay the details so he may own this task. Thanks!

conniey:
@mclay Hey! Thanks for offering to take a look at this. :D The modules affected by this issue are: [azure-messaging-eventhubs](https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/eventhubs/azure-messaging-eventhubs) and [azure-messaging-eventhubs-checkpointstore-blob](https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob).

I was hoping to have something like Key Vault where they have a [`*.properties` file](https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/keyvault/azure-security-keyvault-secrets/src/main/resources) in their resource folder with user-facing messages in it that we could point to in our code base. 

I am open to other suggestions if you've seen it done in the wild in other ways.

mclay:
@conniey could you please reopen this one because i have to also replace the constant string values in azure-messaging-eventhubs-checkpointstore-blob.

conniey:
@mclay Thanks for letting me know. I forgot about that module. :(

mclay:
hi @conniey, i finished the remaining module and it's ready for review

PR #5169, opened 2019-08-29
EventHubs: global constant strings 

mssfang:
Create a file to contains all global constant string for user errors. Refer to issue https://github.com/Azure/azure-sdk-for-java/issues/3923

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubErrorCodeStrings.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubErrorCodeStrings.java
@@ -97,7 +97,8 @@ class EventHubErrorCodeStrings {
 
     private static synchronized void loadProperties() {
         if (errorStrings == null) {
-            try (InputStream fileInputStream = EventHubErrorCodeStrings.class.getClassLoader().getResource(ERROR_STRINGS_FILE_NAME).openStream()) {
+            try (InputStream fileInputStream = EventHubErrorCodeStrings.class.getClassLoader()
+                .getResource(ERROR_STRINGS_FILE_NAME).openStream()) {
                 errorStrings = new Properties();
                 errorStrings.load(fileInputStream);
             } catch (IOException ex) {
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/resources/eventhubsErrorStrings.properties
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/resources/eventhubsErrorStrings.properties
@@ -24,7 +24,8 @@ max_frame_size_require_positive_num='maxFrameSize' must be a positive number.
 payload_exceeded_max_size=Error sending. Size of the payload exceeded maximum message size: %s kb.
 entity_send_failed=Entity(%s): send operation failed. Please see cause for more details.
 entity_send_failed_delivery=Entity(%s): send operation failed while advancing delivery(tag: %s).
-entity_send_failed_schedule_retry=Entity(%s): send operation failed while scheduling a retry on Reactor, see cause for more details.
+entity_send_failed_schedule_retry=\
+  Entity(%s): send operation failed while scheduling a retry on Reactor, see cause for more details.
 entity_send_timeout=Entity(%s): Send operation timed out.
 unable_close_connection_to_service=Unable to close connection to service.
 no_starting_position_set=No starting position was set.
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/EventHubErrorCodeStrings.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/EventHubErrorCodeStrings.java
@@ -147,7 +147,8 @@ class EventHubErrorCodeStrings {
 
     private static synchronized void loadProperties() {
         if (errorStrings == null) {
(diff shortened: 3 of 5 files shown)


conniey:
@JonathanGiles @mssfang
I was thinking of how Android does it, with a strings.xml file. `/resources/values/`, /resources/values-en-us/ file, etc.
I'd need to do a bit more research to see what java's built in approach is.

https://developer.android.com/guide/topics/resources/localization
https://developer.android.com/guide/topics/resources/string-resource

JonathanGiles:
> @JonathanGiles @mssfang
> I was thinking of how Android does it, with a strings.xml file. `/resources/values/`, /resources/values-en-us/ file, etc.
> I'd need to do a bit more research to see what java's built in approach is.
> 
> https://developer.android.com/guide/topics/resources/localization
> https://developer.android.com/guide/topics/resources/string-resource

I can give a lot of context here. I've done this in previous open source projects and there is a similar way that it is done in Java, except with standard properties files rather than xml :-) Vinay also did this for Key Vault, so there is some prior art already.

conniey:
Ahh, I see now! https://www.baeldung.com/java-resourcebundle was trying to google the name of this feature.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubErrorCodeStrings.java
```diff
@@ -0,0 +1,282 @@
...
+package com.azure.messaging.eventhubs;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * The EventHubs common error code Strings
+ */
+public class EventHubErrorCodeStrings {
```
JonathanGiles: This should not be public API.

mssfang: removed and updated

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
```diff
@@ -137,7 +137,8 @@ public EventData(String body) {
      * @throws NullPointerException if {@code message} is null.
      */
     EventData(Message message) {
-        Objects.requireNonNull(message, "'message' cannot be null.");
+        Objects.requireNonNull(message,
+            EventHubErrorCodeStrings.getErrorString(EventHubErrorCodeStrings.MESSAGE_CANNOT_NULL));
```
JonathanGiles: Static import to make the lines shorter (so that they don't wrap). If you static import the getErrorString method and the EventHubErrorCodeStrings types, it would just be `getErrorString(MESSAGE_CANNOT_NULL)`.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubErrorCodeStrings.java
```diff
@@ -0,0 +1,287 @@
...
+     * @return The {@link String value} containing the error message.
+     */
+    static String getErrorString(String propertyName) {
+        loadProperties();
+        return errorStrings.getProperty(propertyName);
+    }
+
+    private static synchronized void loadProperties() {
+        if (errorStrings == null) {
+            try (InputStream fileInputStream = EventHubErrorCodeStrings.class.getClassLoader().getResource((ERROR_STRINGS_FILE_NAME)).openStream()) {
```
JonathanGiles:
```suggestion
            try (InputStream fileInputStream = EventHubErrorCodeStrings.class.getClassLoader().getResource(ERROR_STRINGS_FILE_NAME).openStream()) {
```

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubErrorCodeStrings.java
```diff
@@ -0,0 +1,287 @@
...
+     */
+    static final String TRACER_PROVIDER_CANNOT_NULL = "null_tracer_provider";
+
+    /**
+     * The value of {@code tryTimeout} key is null
+     */
+    static final String TRY_TIME_OUT_CANNOT_NULL = "null_try_time_out";
+
+
+    // EMPTY STRINGS
```
JonathanGiles: What does this mean?

sdk/eventhubs/azure-messaging-eventhubs/src/main/resources/eventhubsErrorStrings.properties
```diff
@@ -0,0 +1,76 @@
...
+null_partition_manager='partitionManager' cannot be null.
+null_initial_event_position='initialEventPosition' cannot be null.
+null_eventhub_name='eventHubName' cannot be null.
+null_consumer_group_name='consumerGroupName' cannot be null.
+null_owner_id='ownerId' cannot be null.
+null_authentication='authentication' cannot be null.
+null_property_key='key' cannot be null.
+null_property_value='value' cannot be null.
+null_message='message' cannot be null.
+null_tracer_provider='tracerProvider' cannot be null.
```
JonathanGiles: Use a wildcard rather than have a different string for each parameter type.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubErrorCodeStrings.java
mssfang: Trying to add more description for the errors messages of "something can not be empty strings" . Removed this comment already

mssfang:
No longer needed.

PR #5181, opened 2019-08-30
Refactor new client libraries to ship with full Java module support

JonathanGiles:
The intent of this change is to prepare Java SDK for Java 9+. The changes include
- Adding module-info.java for all modules
- Configuring visibility of packages to other modules
- Support for groovy-java tests in Storage modules 
- Configuring pom.client.xml to include additional command line args to run the tests
- Some renaming and moving of Java classes to new packages/modules to be Java 9 compliant.

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -347,7 +347,8 @@
             <excludePackageNames>
               *.impl*:
               *.implementation*:
-              com.azure.tools.checkstyle*
+              com.azure.tools.checkstyle*:
+              com.azure.core.test*
             </excludePackageNames>
             <groups>
               <group>
@@ -375,10 +376,6 @@
               <group>
                 <title>Azure Core - Management</title>
                 <packages>com.azure.core.management*</packages>
-              </group>
-              <group>
-                <title>Azure Core - Test</title>
-                <packages>com.azure.core.test*</packages>
               </group>
               <group>
                 <title>Azure App Configuration</title>


pom.client.xml
```diff
@@ -679,6 +701,31 @@
...
           </plugin>
+          <!-- Surefire plugin is broken, https://issues.apache.org/jira/browse/SUREFIRE-1501 -->
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>${maven-surefire-plugin.version}</version>
+            <configuration>
+              <argLine>
+                <!-- KeyVault tests fail without these exports -->
+                --add-exports com.azure.core/com.azure.core.implementation.http=ALL-UNNAMED
```
conniey: Are these issues that we need to fix later? If so, maybe creating issues for them help.

eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
```diff
@@ -59,6 +59,11 @@
     <Bug pattern="PZLA_PREFER_ZERO_LENGTH_ARRAYS"/>
   </Match>
 
+  <Match>
```
conniey: Maybe we should add a reason for this spotbugs exclude? I thought our goal was to avoid adding excludes for track 2 library.

sdk/core/azure-core-http-netty/pom.xml
```diff
@@ -107,6 +113,7 @@
     <dependency>
       <groupId>com.github.tomakehurst</groupId>
       <artifactId>wiremock-standalone</artifactId>
+      <version>2.24.1</version>
```
conniey: I thought explicit version numbers like this should be in our parent/pom.xml

parent/pom.xml
```diff
@@ -163,7 +163,7 @@
     <!-- Maven Tool Versions -->
     <maven-build-tools.version>1.0.2</maven-build-tools.version>
     <sdk-build-tools.version>1.0.0</sdk-build-tools.version>
-    <maven-javadoc.version>3.0.1</maven-javadoc.version>
+    <maven-javadoc.version>3.1.1</maven-javadoc.version>
```
JonathanGiles: @srnagar Remind me - does this cause us trouble w.r.t our generated JavaDoc (recall we had bugs with both versions we wanted to use)

pom.client.xml
```diff
@@ -647,30 +640,59 @@
...
             <version>${maven-compiler-plugin.version}</version>
+            <!-- setup defaults for compile and testCompile -->
             <configuration>
-              <source>11</source>
-              <target>11</target>
+              <source>9</source>
+              <target>9</target>
               <showWarnings>true</showWarnings>
-              <failOnWarning>true</failOnWarning>
+              <failOnWarning>false</failOnWarning> <!-- FIXME: this is false for now as it introduces a build failure -->
```
JonathanGiles: Do we still have build failures if this is true?

pom.client.xml
JonathanGiles: This is a section I would like to dive more deeply into, to make sure we need all of these (and can't, e.g. export them in the module-info)

sdk/core/azure-core-amqp/src/main/java/module-info.java
```diff
@@ -0,0 +1,14 @@
...
+module com.azure.amqp {
+    requires com.azure.core;
+
+    requires reactor.core;
+
+    exports com.azure.core.amqp;
+    exports com.azure.core.amqp.exception;
+
+    exports com.azure.core.amqp.implementation to
+        com.azure.messaging.eventhubs;
```
JonathanGiles: Add a FIXME or TODO in here, so that we know to revisit this (and all other 'exports .. to' lines).

sdk/core/azure-core-http-okhttp/pom.xml
```diff
@@ -90,6 +96,7 @@
     <dependency>
       <groupId>com.github.tomakehurst</groupId>
       <artifactId>wiremock-standalone</artifactId>
+      <version>2.24.1</version>
```
JonathanGiles: Let's take this version out, and make sure parent is the same.

sdk/core/azure-core-http-netty/pom.xml
JonathanGiles: Yep!

sdk/core/azure-core-test/src/main/java/com/azure/core/test/MyRestException.java
```diff
@@ -0,0 +1,25 @@
...
+     */
+    public MyRestException(String message, HttpResponse response, HttpBinJSON body) {
+        super(message, response, body);
+    }
+
+    @Override
+    public HttpBinJSON getValue() {
+        return (HttpBinJSON) super.getValue();
+    }
+}
```
JonathanGiles: I'm not sure why this class exists?

sdk/core/azure-core/src/test/java/com/azure/core/ConfigurationTests.java
```diff
@@ -15,6 +16,7 @@
 /**
  * Tests the configuration API.
  */
+@Ignore("TEMPORARY")
```
JonathanGiles: Can you try removing this?

sdk/identity/azure-identity/src/main/java/module-info.java
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module com.azure.identity {
+    // FIXME this is unfortunate - java.desktop is used to open the users browser
+    requires java.desktop;
```
JonathanGiles: Can you file a tech debt on this please?

sdk/keyvault/azure-keyvault-keys/src/main/java/module-info.java
```diff
@@ -0,0 +1,22 @@
...
+    exports com.azure.security.keyvault.keys.models.webkey;
+
+    opens com.azure.security.keyvault.keys to com.fasterxml.jackson.databind;
+    opens com.azure.security.keyvault.keys.cryptography to com.fasterxml.jackson.databind;
+    opens com.azure.security.keyvault.keys.cryptography.models to com.fasterxml.jackson.databind;
+    opens com.azure.security.keyvault.keys.implementation to com.fasterxml.jackson.databind;
+    opens com.azure.security.keyvault.keys.models to com.fasterxml.jackson.databind;
+    opens com.azure.security.keyvault.keys.models.webkey to com.fasterxml.jackson.databind;
+
+
```
JonathanGiles: Get rid of empty lines here

sdk/storage/azure-storage-blob/pom.xml
```diff
@@ -120,37 +122,148 @@
...
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>${maven-surefire-plugin.version}</version>
+            <configuration>
+              <argLine>
+                --add-exports com.azure.core/com.azure.core.implementation.http=ALL-UNNAMED
+                --add-exports com.azure.core/com.azure.core.implementation.serializer.jackson=ALL-UNNAMED
+                --add-exports com.azure.core/com.azure.core.implementation.util=ALL-UNNAMED
+                --add-opens com.azure.storage.blob/com.azure.storage.blob=ALL-UNNAMED
+                --add-opens com.azure.storage.blob/com.azure.storage.blob.implementation=ALL-UNNAMED
```
JonathanGiles: Lets dive into whether these are all required.

sdk/core/azure-core/pom.xml
```diff
@@ -152,6 +152,7 @@
       <!-- RestProxyTests is inherited by tests in azure-core-http-netty -->
       <plugin>
         <artifactId>maven-jar-plugin</artifactId>
+        <version>3.1.2</version>
```
conniey: i noticed a couple of plugin versions. is this something enforced in parent/pom.xml

pom.client.xml
srnagar: Yes, I will report an issue for this.

eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
srnagar: I will add a comment for this and report an issue. This was added for the azure-core-test package.

sdk/core/azure-core-http-netty/pom.xml
srnagar: Resolved.

pom.client.xml
srnagar: Yes, the build will still fail because we are exporting some implementation packages to client library modules and a clean build will not know in advance all these modules and logs warning about "<modulename> is not known"

sdk/core/azure-core-amqp/src/main/java/module-info.java
srnagar: Added comments to fix these later.

sdk/core/azure-core-http-okhttp/pom.xml
srnagar: yup, updated.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/MyRestException.java
srnagar: I moved it from azure-core to azure-core-test. It's used in a bunch of tests.

sdk/core/azure-core/src/test/java/com/azure/core/ConfigurationTests.java
srnagar: Removed.

sdk/identity/azure-identity/src/main/java/module-info.java
srnagar: Will create an issue.

sdk/keyvault/azure-keyvault-keys/src/main/java/module-info.java
srnagar: Done.

sdk/storage/azure-storage-blob/pom.xml
srnagar: Discussed this offline and we will need to keep it until we fix the implementation package usage in other modules.

sdk/core/azure-core-http-okhttp/pom.xml
```diff
@@ -61,6 +61,12 @@
       <artifactId>azure-core</artifactId>
       <version>1.0.0-preview.5</version>
     </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-test</artifactId>
+      <version>1.0.0-preview.5</version>
+      <scope>compile</scope>
```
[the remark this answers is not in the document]
srnagar: Yes, removing it.

sdk/core/azure-core/pom.xml
srnagar: Removed version from here.

pom.client.xml
JonathanGiles: I am fine with that - let's get two issues, one for module-info `exports .. to` statements, and the other for the build-time `add-opens` statements. Once those are filed I am fine with merging what we have today.

pom.client.xml
JonathanGiles: Please file an issue. Of all the issues to be filed, this is the one I want resolved most urgently. Do you have any plans on how to resolve this?

sdk/core/azure-core-test/src/main/java/com/azure/core/test/MyRestException.java
JonathanGiles: Ok. I can't recall off the top of my head if we publish the javadoc for this module, but if we do, I think we should stop it. This isn't an end-user facing library, so it shouldn't be included. If we do, please can you exclude that package?

JonathanGiles:
Please remember to update the storage blob module-info to export the new `specialized` package.

sdk/core/azure-core-test/pom.xml
```diff
@@ -52,10 +52,76 @@
...
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.github.tomakehurst</groupId>
+      <artifactId>wiremock-standalone</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
```
alzimmermsft: I thought we were trying to remove our dependency on Jetty, why was this added here?

srnagar: Some tests were moved from azure-core to azure-core-test and to run them this is required. When we remove the dependency on Jetty, we have to update this as well.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/entities/Slide.java
```diff
@@ -40,6 +41,9 @@ public String title() {
      * @return The content strings of the slide.
      */
     public String[] items() {
-        return items;
+        if (items == null) {
```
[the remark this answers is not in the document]
srnagar: It's recommended to return an empty array/collection instead of null.

http://www.informit.com/articles/article.aspx?p=2133373&seqNum=5

sdk/core/azure-core/src/main/java/com/azure/core/annotation/ReturnValueWireType.java
```diff
@@ -14,7 +16,7 @@
  * Supported values are:
  *
  * <ol>
- *     <li>{@link com.azure.core.implementation.Base64Url}</li>
+ *     <li>{@link Base64Url}</li>
```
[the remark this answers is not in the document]
srnagar: Updated other links too.

sdk/core/azure-core-test/pom.xml
```diff
@@ -52,10 +52,76 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
-
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
```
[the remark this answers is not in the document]
srnagar: No, these are used in src/main/java packages, so, they should not be test dependencies.

parent/pom.xml
srnagar: Yeah, I figured out what the issue was and fixed it. We can now use the latest version of javadoc plugin that fixed external links issue we had with 3.0.1 version of the plugin.

So, our javadocs will now have links to Mono, Flux and other reactor and netty types.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/MyRestException.java
srnagar: We used to publish javadocs for test module. I have removed it now.

PR #5185, opened 2019-08-30
Reactor-Netty HttpClient changes

kushagraThapar:
Changes: 

* Used Reactor-Core and Reactor-Netty latest release. 
* New TCP Configurations with idle connection timeout, connect timeout, connection keep alive, etc. 
* Used Reactor-Netty's response APIs to convert Flux<ByteBuf> / Mono<ByteBuf> to Flux<String> / Mono<String>

### DIFF
--- a/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/RxGatewayStoreModel.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/RxGatewayStoreModel.java
@@ -17,7 +17,6 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpResponseStatus;
-import io.netty.util.internal.StringUtil;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;


sdk/cosmos/microsoft-azure-cosmos-benchmark/src/main/java/com/azure/data/cosmos/benchmark/Configuration.java
```diff
@@ -55,7 +55,7 @@
     private int documentDataFieldSize = 20;
 
     @Parameter(names = "-maxConnectionPoolSize", description = "Max Connection Pool Size")
-    private Integer maxConnectionPoolSize = 1000;
+    private Integer maxConnectionPoolSize = 500;
```
moderakh: we have many different endpoint in the direct mode.

with `maxConnectionPoolSize = 500;` does it mean the connection pool size to each endpoint is at most 500 or does it mean total number of connections to all endpoints at most is 500?

as a separate question why are we changing this to 500?

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/ConnectionPolicy.java
```diff
@@ -18,7 +18,7 @@
     private static final int DEFAULT_MEDIA_REQUEST_TIMEOUT_IN_MILLIS = 300 * 1000;
     private static final int DEFAULT_IDLE_CONNECTION_TIMEOUT_IN_MILLIS = 60 * 1000;
 
-    private static final int DEFAULT_MAX_POOL_SIZE = 1000;
+    private static final int DEFAULT_MAX_POOL_SIZE = 500;
```
moderakh: why are we changing to 500?

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/BufferedHttpResponse.java
```diff
@@ -53,6 +55,11 @@ public HttpHeaders headers() {
         return bodyAsByteArray().flatMapMany(bytes -> Flux.just(Unpooled.wrappedBuffer(bytes)));
     }
 
+    @Override
+    public Flux<InputStream> bodyAsInputStream() {
+        return bodyAsByteArray().flatMapMany(bytes -> Flux.just(new ByteArrayInputStream(bytes)));
```
moderakh: why do we need `bodyAsInputStream()` ? why can't we fully rely on the materialized String or Bytes?

whey you wrap in ByteArrayInputStream memory will be GCed later. this kind of is similar to inmobi issue

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpResponse.java
```diff
@@ -68,6 +69,13 @@
      */
     public abstract Flux<ByteBuf> body();
 
+    /**
+     * Get the response content as InputStream.
+     *
+     * @return this response content as InputStream
+     */
+    public abstract Flux<InputStream> bodyAsInputStream();
```
moderakh: why can't we just rely on Flux<Byte> or Flux<String> ?

we are not doing attachment anymore, so we shouldn't need to do InputStream I think

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpClient.java
```diff
@@ -28,12 +30,40 @@ static HttpClient createFixed(HttpClientConfig httpClientConfig) {
             throw new IllegalArgumentException("HttpClientConfig is null");
         }
 
-        if (httpClientConfig.getMaxPoolSize() == null) {
-            return new ReactorNettyClient(ConnectionProvider.fixed(httpClientConfig.getConfigs().getReactorNettyConnectionPoolName()), httpClientConfig);
+        Integer maxIdleConnectionTimeoutInMillis = 60 * 1000;
```
moderakh: should be defined as constant

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpClient.java
```diff
@@ -28,12 +30,40 @@ static HttpClient createFixed(HttpClientConfig httpClientConfig) {
...
+        }
+
+        //  Default pool size
+        Integer maxPoolSize = 500;
+        if (httpClientConfig.getMaxPoolSize() != null) {
+            maxPoolSize = httpClientConfig.getMaxPoolSize();
         }
-        return new ReactorNettyClient(ConnectionProvider.fixed(httpClientConfig.getConfigs().getReactorNettyConnectionPoolName(), httpClientConfig.getMaxPoolSize()), httpClientConfig);
+
+        int connectionAcquireTimeoutInMillis = 45 * 1000;
```
moderakh: should be defined as constant

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpClient.java
```diff
@@ -28,12 +30,40 @@ static HttpClient createFixed(HttpClientConfig httpClientConfig) {
...
 
-        if (httpClientConfig.getMaxPoolSize() == null) {
-            return new ReactorNettyClient(ConnectionProvider.fixed(httpClientConfig.getConfigs().getReactorNettyConnectionPoolName()), httpClientConfig);
+        Integer maxIdleConnectionTimeoutInMillis = 60 * 1000;
+        if (httpClientConfig.getMaxIdleConnectionTimeoutInMillis() != null) {
+            maxIdleConnectionTimeoutInMillis = httpClientConfig.getMaxIdleConnectionTimeoutInMillis();
+        }
+
+        //  Default pool size
+        Integer maxPoolSize = 500;
```
moderakh: should be defined as constant

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/ReactorNettyClient.java
```diff
@@ -31,30 +32,51 @@
...
+                tcpClient =
+                    tcpClient.proxy(typeSpec -> typeSpec.type(ProxyProvider.Proxy.HTTP).address(this.httpClientConfig.getProxy()));
             }
+            tcpClient =
+                tcpClient.secure(sslContextSpec -> sslContextSpec.sslContext(this.httpClientConfig.getConfigs().getSslContext()));
+            if (LoggerFactory.getLogger(REACTOR_NETWORK_LOG_CATEGORY).isTraceEnabled()) {
+                tcpClient = tcpClient.wiretap(REACTOR_NETWORK_LOG_CATEGORY, LogLevel.INFO);
+            }
+            //  By default, keep alive is enabled on http client
+            tcpClient = tcpClient.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, 30 * 1000);
```
moderakh: we should consider to make 30 *100 a constant.

sdk/cosmos/microsoft-azure-cosmos-benchmark/src/main/java/com/azure/data/cosmos/benchmark/Configuration.java
kushagraThapar: It is at most 500 concurrent connections (combining all endpoints). 

Reactor Netty has done some performance benchmarking, and they suggest 500 as max connection pool size. 
Can't find the article right now, will post it here once I find it.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/ConnectionPolicy.java
kushagraThapar: Same reason as above.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/BufferedHttpResponse.java
kushagraThapar: Good point, and since attachment is not used at all, I will try to get rid of this.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpResponse.java
kushagraThapar: Yes, attachment is no longer supported, will get rid of this completely.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpClient.java
kushagraThapar: Will do.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpClient.java
kushagraThapar: Will do.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpClient.java
kushagraThapar: Will do.

sdk/cosmos/microsoft-azure-cosmos-benchmark/src/main/java/com/azure/data/cosmos/benchmark/Configuration.java
kushagraThapar: Changed it back to 1000 connections.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/Configs.java
```diff
@@ -55,6 +55,10 @@
     private static final int CPU_CNT = Runtime.getRuntime().availableProcessors();
     private static final int DEFAULT_DIRECT_HTTPS_POOL_SIZE = CPU_CNT * 500;
 
+    //  Reactor Netty Constants
+    private static final int MAX_IDLE_CONNECTION_TIMEOUT_IN_MILLIS = 60 * 1000;
+    private static final int CONNECTION_ACQUIRE_TIMEOUT_IN_MILLIS = 45 * 1000;
```
moderakh: question: why do we have a different timeout value? any reason this one is 45s and the other is 60s? is there a rationale behind this?

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/RxGatewayStoreModel.java
```diff
@@ -17,14 +17,14 @@
 import io.netty.buffer.Unpooled;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.util.internal.StringUtil;
```
moderakh: I would use StringUtils from apache commons lang not from netty

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/directconnectivity/StoreReader.java
```diff
@@ -158,13 +158,13 @@ public StoreReader(
                                 return Flux.error(e);
                             }
                         }
-                ).onErrorResume(t -> {
-
+                ).onErrorResume(throwable -> {
+                    Throwable unwrappedException = Exceptions.unwrap(throwable);
```
moderakh: some places we do unwrap and some places we don't. why

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/Configs.java
kushagraThapar: Yes, these are default values for reactor netty for these variables. I wanted to explicitly state them in our code. 
So that we can change them whenever we want.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/RxGatewayStoreModel.java
kushagraThapar: This is unused :) 
I will remove it.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/directconnectivity/StoreReader.java
kushagraThapar: Because of the way these exceptions are thrown from the above layer.

mitchdenny:

PR #5186, opened 2019-08-30
Add PagedFlux and PagedIterable to Files and Queues

alzimmermsft:


sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
```diff
@@ -639,45 +667,14 @@ public String getShareSnapshotId() {
...
-    }
-
     private List<FileRef> convertResponseAndGetNumOfResults(DirectorysListFilesAndDirectoriesSegmentResponse response) {
         List<FileRef> fileRefs = new ArrayList<>();
         if (response.value().segment() != null) {
             response.value().segment().directoryItems().forEach(directoryItem -> fileRefs.add(new FileRef(directoryItem.name(), true, null)));
             response.value().segment().fileItems().forEach(fileItem -> fileRefs.add(new FileRef(fileItem.name(), false, fileItem.properties())));
         }
+
+        fileRefs.sort(Comparator.comparing(FileRef::name));
```
jaschrep-msft: Why are we adding this sort? This is already a segment of a paged response, what's the point of spending cycles sorting every 1000 file subset of a larger list?

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryClient.java
```diff
@@ -288,8 +289,8 @@ public DirectorySetMetadataInfo setMetadata(Map<String, String> metadata) {
      *                   If the request does not specify maxresults or specifies a value greater than 5,000, the server will return up to 5,000 items.
      * @return {@link FileRef File info} in this directory with prefix and max number of return results.
      */
-    public Iterable<FileRef> listFilesAndDirectories(String prefix, Integer maxResults) {
-        return directoryAsyncClient.listFilesAndDirectories(prefix, maxResults).toIterable();
+    public PagedIterable<FileRef> listFilesAndDirectories(String prefix, Integer maxResults) {
```
jaschrep-msft: We need timeouts in the sync cases. You can look at similar operations in the blob service for an example on how to do it. Make sure you check all the sync clients for this.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileAsyncClient.java
```diff
@@ -854,10 +856,19 @@ private void channelCleanUp(AsynchronousFileChannel channel) {
      * @param range Optional byte range which returns file data only from the specified range.
      * @return {@link FileRange ranges} in the files that satisfy the requirements
      */
-    public Flux<FileRange> listRanges(FileRange range) {
+    public PagedFlux<FileRange> listRanges(FileRange range) {
```
jaschrep-msft: This is not a paged operation. `marker` isn't being used anywhere.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/ShareAsyncClient.java
```diff
@@ -444,9 +447,18 @@ public FileAsyncClient getFileClient(String filePath) {
      * @return The stored access policies specified on the queue.
      * @throws StorageException If the share doesn't exist
      */
-    public Flux<SignedIdentifier> getAccessPolicy() {
-        return azureFileStorageClient.shares().getAccessPolicyWithRestResponseAsync(shareName, Context.NONE)
-            .flatMapMany(response -> Flux.fromIterable(response.value()));
+    public PagedFlux<SignedIdentifier> getAccessPolicy() {
```
jaschrep-msft: This is not a paged operation.

sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueAsyncClient.java
```diff
@@ -295,9 +298,18 @@ public URL getQueueUrl() {
      * @return The stored access policies specified on the queue.
      * @throws StorageException If the queue doesn't exist
      */
-    public Flux<SignedIdentifier> getAccessPolicy() {
-        return postProcessResponse(client.queues().getAccessPolicyWithRestResponseAsync(queueName, Context.NONE))
-            .flatMapMany(response -> Flux.fromIterable(response.value()));
+    public PagedFlux<SignedIdentifier> getAccessPolicy() {
```
jaschrep-msft: This is not a paged operation.

sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueAsyncClient.java
```diff
@@ -544,11 +556,20 @@ public URL getQueueUrl() {
      * @throws StorageException If the queue doesn't exist or {@code maxMessages} or {@code visibilityTimeout} is
      * outside of the allowed bounds
      */
-    public Flux<DequeuedMessage> dequeueMessages(Integer maxMessages, Duration visibilityTimeout) {
+    public PagedFlux<DequeuedMessage> dequeueMessages(Integer maxMessages, Duration visibilityTimeout) {
```
jaschrep-msft: This is not a paged response

sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueAsyncClient.java
```diff
@@ -594,10 +615,18 @@ public URL getQueueUrl() {
      * metadata about the message.
      * @throws StorageException If the queue doesn't exist or {@code maxMessages} is outside of the allowed bounds
      */
-    public Flux<PeekedMessage> peekMessages(Integer maxMessages) {
-        return postProcessResponse(client.messages()
-            .peekWithRestResponseAsync(queueName, maxMessages, null, null, Context.NONE))
-            .flatMapMany(response -> Flux.fromIterable(response.value()));
+    public PagedFlux<PeekedMessage> peekMessages(Integer maxMessages) {
```
jaschrep-msft: This is not a paged response.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
alzimmermsft: The REST spec states that the response is lexically sorted with directories and files intermingled and I wanted to align on this. https://docs.microsoft.com/en-us/rest/api/storageservices/list-directories-and-files#remarks

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryClient.java
alzimmermsft: Created issue #5189

alzimmermsft:
@jaschrep-msft The [specification](https://azure.github.io/azure-sdk/java_design.html#java-pagination-pagedflux) states to use `PagedFlux` in the case of non-paged list returns. While I'll agree that `PagedFlux` doesn't tightly align with non-paged responses it allows for all headers and response data to be accessible which is another tenant of the Track 2 clients.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
jaschrep-msft: If the rest spec says it's sorted, wouldn't it already be sorted?

jaschrep-msft:
@alzimmermsft I was directed to not use this in the blob client. Opening an issue for the blob client to use this for its non-paged listing operations.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
sima-zhu: They should sort in our return type in fileRef.
The line here is trying to sort out based on name instead of sort separately based on dir and file.

PR #5188, opened 2019-08-30
Storage: Basic Request Validation

jaschrep-msft:
Introduced a pipeline policy that applies general-purpose assertions to
responses. Can add these configurations individually to, or to all of,
the blob, file, and queue packages.

Added a validation to check if the x-ms-client-request-id header is echoed correctly, should it be present in the response.

### DIFF
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/BaseClientBuilder.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/BaseClientBuilder.java
@@ -110,7 +110,7 @@ public abstract class BaseClientBuilder<T extends BaseClientBuilder<T>> {
      */
     private HttpPipelinePolicy makeValidationPolicy() {
         ResponseValidationPolicyBuilder builder = new ResponseValidationPolicyBuilder()
-            .optionalEcho(Constants.HeaderConstants.CLIENT_REQUEST_ID); // echo client request id
+            .addOptionalEcho(Constants.HeaderConstants.CLIENT_REQUEST_ID); // echo client request id
 
         applyServiceSpecificValidations(builder);
 
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ResponseValidationPolicyBuilder.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ResponseValidationPolicyBuilder.java
@@ -37,11 +37,15 @@ public class ResponseValidationPolicyBuilder {
      * @param headerName The header to validate.
      * @return This policy.
      */
-    public ResponseValidationPolicyBuilder optionalEcho(String headerName) {
+    public ResponseValidationPolicyBuilder addOptionalEcho(String headerName) {
         assertions.add((httpResponse, logger) -> {
+            String requestHeaderValue = httpResponse.request().headers().value(headerName);
             String responseHeaderValue = httpResponse.headerValue(headerName);
-            if (responseHeaderValue != null && !responseHeaderValue.equals(httpResponse.request().headers().value(headerName))) {
-                throw logger.logExceptionAsError(new RuntimeException());
+            if (responseHeaderValue != null && !responseHeaderValue.equals(requestHeaderValue)) {
+                throw logger.logExceptionAsError(new RuntimeException(String.format(
+                    "Unexpected header value. Expected response to echo `%s: %s`. Got value `%s`.",
+                    headerName, requestHeaderValue, responseHeaderValue
+                )));
             }
         });
 


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ResponseValidationPolicyBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+     * Fluently applies an optional validation to this policy where, if the response contains the given header, asserts
+     * its value is an echo of the value provided in the request.
+     *
+     * @param headerName The header to validate.
+     * @return This policy.
+     */
+    public ResponseValidationPolicyBuilder optionalEcho(String headerName) {
+        assertions.add((httpResponse, logger) -> {
+            String responseHeaderValue = httpResponse.headerValue(headerName);
+            if (responseHeaderValue != null && !responseHeaderValue.equals(httpResponse.request().headers().value(headerName))) {
```
gapra-msft: Does httpResponse.request().headers().value(headerName) throw if headerName is invalid?

jaschrep-msft: Javadocs say it returns null.

jaschrep-msft:
Addresses issue #4511

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ResponseValidationPolicyBuilder.java
gapra-msft: Should be fine if it returns null

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ResponseValidationPolicyBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Fluently applies an optional validation to this policy where, if the response contains the given header, asserts
+     * its value is an echo of the value provided in the request.
+     *
+     * @param headerName The header to validate.
+     * @return This policy.
+     */
+    public ResponseValidationPolicyBuilder optionalEcho(String headerName) {
```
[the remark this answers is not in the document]
jaschrep-msft: That matches the client builders as well. Will fix.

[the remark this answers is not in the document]
jaschrep-msft: Oops, it looks like we log a blank RuntimeException. I can have it construct a message based on the header name.

As for leaving a log message in general, I'd say not on this API, as it's very obviously doing one and only one check, and can easily make a log message as to what was expected and not gotten. In the future, if we need to pass in arbitrary consumers to do more intricate assertions, those APIs can take in log messages. Or the consumer can have the log message on its own.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ResponseValidationPolicyBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+        assertions.add((httpResponse, logger) -> {
+            String responseHeaderValue = httpResponse.headerValue(headerName);
+            if (responseHeaderValue != null && !responseHeaderValue.equals(httpResponse.request().headers().value(headerName))) {
+                throw logger.logExceptionAsError(new RuntimeException());
```
sima-zhu: Do you need to log specific error msg here? How customers checkout what happened if no error log out.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/BaseClientBuilder.java
```diff
@@ -98,6 +102,29 @@ protected HttpPipeline buildPipeline() {
...
+        return builder.build();
+    }
+
+    /**
+     * Applies validation of general-purpose requests to builder. Method-specific validations should be performed in the
+     * convenience layer.
+     *
+     * @param builder Builder to assemble assertions together.
+     */
+    protected abstract void applyServiceSpecificValidations(ResponseValidationPolicyBuilder builder);
```
sima-zhu: I did not see any implementation on each builder. Is this method a placeholder or leave it here for some other purpose?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ResponseValidationPolicyBuilder.java
jaschrep-msft: Mentioned in one of Alan's comments that I missed this. Will be addressed.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/BaseClientBuilder.java
jaschrep-msft: This this is overridden with `final` in BaseBlobClientBuilder, BaseFileClientBuilder, and BaseQueueClientBuilder. They are currently all three no-ops, but this is laying the groundwork for CPK to validate that the key-hash returned is the same as the one sent over the wire. CPK is a blob-only feature, so there was no need to add extra steps to file and queue's pipelines.

PR #5190, opened 2019-08-30
Misc fixes

srnagar:
This change includes Netty library samples directory to spotbugs reporting and add codesnippets to javadocs.


eng/spotbugs-aggregate-report/pom.xml
```diff
@@ -59,6 +59,7 @@
                 <source>..\..\sdk\core\azure-core\src\samples\java</source>
                 <source>..\..\sdk\core\azure-core-amqp\src\main\java</source>
                 <source>..\..\sdk\core\azure-core-http-netty\src\main\java</source>
+                <source>..\..\sdk\core\azure-core-http-netty\src\samples\java</source>
```
JonathanGiles: Can you also add okHttp in here too please?

srnagar: Looks like this PR adds okHttp #5209.

PR #5206, opened 2019-09-02
Dep version update

JonathanGiles:
This PR is trying to align dependencies such that builds that were failing will start to pass.

### DIFF
--- a/sdk/keyvault/microsoft-azure-keyvault-test/pom.xml
+++ b/sdk/keyvault/microsoft-azure-keyvault-test/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.azure</groupId>
     <artifactId>azure-data-sdk-parent</artifactId>
-    <version>1.3.0</version>
+    <version>1.2.0</version>
     <relativePath>../../../pom.data.xml</relativePath>
   </parent>
 


pom.client.xml
```diff
@@ -6,7 +6,7 @@
   <groupId>com.azure</groupId>
   <artifactId>azure-client-sdk-parent</artifactId>
   <packaging>pom</packaging>
-  <version>1.3.0</version>
+  <version>1.4.0</version>
```
mitchdenny: I think this can remain ```1.3.0``` since it hasn't been published yet.

pom.data.xml
```diff
@@ -6,7 +6,7 @@
   <groupId>com.azure</groupId>
   <artifactId>azure-data-sdk-parent</artifactId>
   <packaging>pom</packaging>
-  <version>1.2.0</version>
+  <version>1.3.0</version>
```
mitchdenny: This can remain ```1.2.0``` since it hasn't been published yet.

mitchdenny: I reverted it back.

pom.client.xml
mitchdenny: I reverted it back

PR #5207, opened 2019-09-02
Fix example URLs (which have been refactored)

jdubois:


jdubois:
Sorry for the noise: I was in a branch with that error, I see this has already been fixed on master! So closing this.

PR #5224, opened 2019-09-03
System properties

### LINKED ISSUE #5146: [Event Hubs] Expose SystemProperties
* Expose message_annotations (ie. SystemProperties) in EventData that is received.
* Remove the well-known ones that we always extract (ie. sequenceNumber, offset).

conniey:
* Sets system properties on creation time of EventData from a Proton-j message.
* Removes the overlapping properties from the system properties map such as:
  * enqueuedTime
  * partitionKey
  * offset
  * sequencenumber

Fixes #5146


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
```diff
@@ -352,9 +352,37 @@ public int hashCode() {
      */
     private static class SystemProperties extends HashMap<String, Object> {
         private static final long serialVersionUID = -2827050124966993723L;
+        private final String offset;
```
jsquire: I thought we were going to make `offset` a numeric;  the service should never send back an offset of "-1" or "@latest" - it should always be a concrete number of bytes.

conniey: Ah yes. I will create an issue and address this in another PR. thanks for the catch!

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
```diff
@@ -352,9 +352,37 @@ public int hashCode() {
...
+        }
 
         SystemProperties(final Map<String, Object> map) {
-            super(Collections.unmodifiableMap(map));
+            super(map);
+            this.offset = getSystemProperty(OFFSET_ANNOTATION_NAME.getValue());
+            this.partitionKey = getSystemProperty(PARTITION_KEY_ANNOTATION_NAME.getValue());
+
+            final Date enqueuedTimeValue = getSystemProperty(ENQUEUED_TIME_UTC_ANNOTATION_NAME.getValue());
+            this.enqueuedTime = enqueuedTimeValue != null ? enqueuedTimeValue.toInstant() : null;
```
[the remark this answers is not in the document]
conniey: Not sure. I copied this logic from the Track 1 library.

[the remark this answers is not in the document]
conniey: I added the error. Makes sense to throw it.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/Checkpoint.java
```diff
@@ -106,7 +106,7 @@ public Checkpoint ownerId(String ownerId) {
      *
      * @return The offset of the last successfully processed event to store as checkpoint.
      */
-    public String offset() {
+    public long offset() {
```
srnagar:
```suggestion
    public Long offset() {
```

PR #5230, opened 2019-09-04
Fixed the issue which request id did not take down in json file.

sima-zhu:
Storage has a new pipeline which check the equality of request and response x-ms-client-request-id.

The playback tests all failed because our playback json did not have the field in request header.

Added the field into json files. 
TODO: Will add storage playback jsons after approval. 

Failed pipeline:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=101548&view=logs&jobId=a36c942c-7952-53c6-0566-20a200a74a93


alzimmermsft:
I think you merged in another open PR into this PR

PR #5231, opened 2019-09-04
Turn off shallow cloning.

mitchdenny:
This PR turns off shallow cloning in. I believe that shallow cloning and some replication/caching on the GitHub side might be causing intermittent build failures.


eng/pipelines/templates/variables/globals.yml
```diff
@@ -1,6 +1,6 @@
 variables:
   DefaultOptions: '--batch-mode --fail-at-end -Dmaven.wagon.http.pool=false --settings eng/settings.xml'
   LoggingOptions: '-Dorg.slf4j.simpleLogger.defaultLogLevel=error -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn'
-  Agent.Source.Git.ShallowFetchDepth: 1
+  #Agent.Source.Git.ShallowFetchDepth: 1
```
weshaggard: Did you want to try and up the depth as a first step to dealing with the intermittent failures?

mitchdenny: I think for now I just want to stop the bleeding. Since we made the gh-pages change clone times should be way better anyway.

mitchdenny:
Merged after a quick chat with @JonathanGiles

PR #5232, opened 2019-09-04
Turn spotbugs back on in PR.

mitchdenny:
This PR removes the condition on spotbugs so it runs on PRs, this to catch issues that have been failing builds in CI.


mitchdenny:
Merging after a quick chat with @JonathanGiles

Issue #4515, opened 2019-07-22
File SMB Properties REST Parity


rickle-msft:
service feature

Issue #5039, opened 2019-08-19
Determine Defaults for New Required File Parameters


alzimmermsft:
With the update to service version 2019-02-02 there are new required properties that need defaults determined. Methods to fix are:

- DirectoryClient.create
- FileClient.create
- FileClient.setHttpHeaders

PR #5233, opened 2019-09-04
Storage restparity

gapra-msft:


sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
```diff
@@ -150,28 +152,39 @@ public DirectoryAsyncClient getSubDirectoryClient(String subDirectoryName) {
...
-    Mono<Response<DirectoryInfo>> createWithResponse(Map<String, String> metadata, Context context) {
-        // TODO (alzimmer): These properties are dummy defaults to allow the new service version to be used. Remove these and use correct defaults when known (https://github.com/Azure/azure-sdk-for-java/issues/5039)
-        String fileAttributes = "None";
-        String filePermission = "inherit";
-        String fileCreationTime = "now";
-        String fileLastWriteTime = "now";
-        return postProcessResponse(azureFileStorageClient.directorys().
-            createWithRestResponseAsync(shareName, directoryPath, fileAttributes, fileCreationTime, fileLastWriteTime,
-                null, metadata, filePermission, null, context))
+    public Mono<Response<DirectoryInfo>> createWithResponse(FileSmbProperties smbProperties, String filePermission,
```
gapra-msft: The way this is exposed is different in directories and files. I put the filePermission in the file properties which may or may not make sense since it is never returned from the service in these calls. I could take filePermission out of the file fileproperties and make it look closer to the way directories expose it

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/FileProperties.java
```diff
@@ -4,32 +4,69 @@
...
+    private String contentLanguage;
+    private OffsetDateTime copyCompletionTime;
+    private String copyStatusDescription;
+    private String copyId;
+    private String copyProgress;
+    private String copySource;
+    private CopyStatusType copyStatus;
+    private Boolean isServerEncrypted;
+    private FileSmbProperties smbProperties;
+    private String filePermission;
```
gapra-msft: Should I remove this and make it an extra parameter in the setfileproperties and createfile calls?

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/FileProperties.java
```diff
@@ -4,32 +4,69 @@
...
+     * @param contentEncoding The Content-Encoding request header.
+     * @param contentLanguage The Content-Language request header.
+     * @param cacheControl The Cache-Control request header.
+     * @param contentMD5 The MD5 hash of the file to check the message content integrity.
+     * @param contentDisposition The Content-Disposition header that specifies how to process the response.
+     * @param smbProperties The SMB properties of the file.
+     * @param filePermission The file permission of the file. smbProperties.filePermissionKey and filePermission must
+     *                       not both be set simultaneously
+     */
+    public FileProperties(final String contentType, final String contentEncoding, final String contentLanguage,
```
gapra-msft: I kind of want to add builder style setters for the settable properties to make it easier to build the object

gapra-msft: Another addition would be to maybe add a setter for the http headers from a filehttpheaders object?

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileSmbProperties.java
```diff
@@ -0,0 +1,278 @@
...
+     */
+    private static String parseFileSMBDate(OffsetDateTime time) {
+        return time.format(DateTimeFormatter.ofPattern(FileConstants.SMB_DATE_STRING));
+    }
+
+    /**
+     * Creates a new {@link FileSmbProperties} from a {@link FilesCreateResponse}.
+     * @param response The {@link FilesCreateResponse}.
+     */
+    FileSmbProperties(FilesCreateResponse response) {
```
alzimmermsft: For all of these creation methods are the same properties accessed with the same names? It might be worse but a thought could be taking a HttpHeaders object and pulling the values out manually, that way there would only be one method to maintain.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileSmbProperties.java
```diff
@@ -0,0 +1,278 @@
...
+import com.azure.storage.file.models.FilesDownloadResponse;
+import com.azure.storage.file.models.FilesGetPropertiesResponse;
+import com.azure.storage.file.models.FilesSetHTTPHeadersResponse;
+import com.azure.storage.file.models.NtfsFileAttributes;
+
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.EnumSet;
+
+public class FileSmbProperties {
```
alzimmermsft: Would it make sense to split this into two classes, one for request data and another for response data?

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileConstants.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Default value for several SMB file headers.
+     */
+    public static final String PRESERVE = "Preserve";
+
+    /**
+     * Stores a reference to the date/time pattern expected for File SMB properties
+     */
+    public static final String SMB_DATE_STRING = "yyyy-MM-dd'T'HH:mm:ss.SSSSSSS'Z'";
```
alzimmermsft: Is it required that Files use a more precise date string? Could this use `Utility.ISO8601_PATTERN`? They're both ISO8601 but the one in utility is precise down to the second.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileExtensions.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class FileExtensions {
+
+    /**
+     * Verifies that the file permission and file permission key are not both set and if the file permission is set,
+     * the file permission is of valid length.
+     * @param filePermission The file permission.
+     * @param filePermissionKey The file permission key.
+     */
+    public static void filePermissionAndKeyHelper(String filePermission, String  filePermissionKey) {
```
alzimmermsft: I think this should be removed and allow a request to be sent to the service and fail there. If the service changed to allow both values or a larger file permission size this would need to be updated or it would inappropriately fail requests.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileConstants.java
```diff
@@ -0,0 +1,42 @@
...
+
+    static final class MessageConstants {
```
alzimmermsft: Depending on the comment about removing the `FileExtensions` class this may need to be removed as well.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/NtfsFileAttributes.java
```diff
@@ -0,0 +1,183 @@
...
+     *         which is described at {@link #toAttributes(String)}.
+     */
+    public static String toString(EnumSet<NtfsFileAttributes> ntfsAttributes) {
+        if (ntfsAttributes == null) {
+            return Constants.EMPTY_STRING;
+        }
+
+        final StringBuilder builder = new StringBuilder();
+
+        if (ntfsAttributes.contains(NtfsFileAttributes.READ_ONLY)) {
```
alzimmermsft: Could add a small helper function which takes in the StringBuilder, EnumSet, Enum to look for, and String to add if enum is found.

sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/spock/DirectoryAPITests.groovy
```diff
@@ -21,6 +25,8 @@ class DirectoryAPITests extends APISpec {
     def directoryPath
     def shareName
     static def testMetadata
+    static def smbProperties
+    static def filePermission = "O:S-1-5-21-2127521184-1604012920-1887927527-21560751G:S-1-5-21-2127521184-1604012920-1887927527-513D:AI(A;;FA;;;SY)(A;;FA;;;BA)(A;;0x1200a9;;;S-1-5-21-397955417-626881126-188441444-3053964)S:NO_ACCESS_CONTROL"
```
alzimmermsft: Is any of this file permission string sensitive information or contain any secrets? If so we'll need to find another way to get this information for running the tests (also you'll need to regenerate any secrets this used since they're publicly visible).

alzimmermsft:
This fixes #4515 and #5039

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileConstants.java
gapra-msft: The service call will fail with a less precise date

sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/spock/DirectoryAPITests.groovy
gapra-msft: This isn't sensitive information from what I know of.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileExtensions.java
gapra-msft: The permission size is required to be smaller than 8kb since headers cant be longer than that (request wont even go through) so I think it is best to keep that check. As for the other check, it is a conceptual check because it doesnt make sense to send both (it's always an either or).

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileSmbProperties.java
gapra-msft: I think it would be convenient to keep them in the same class due to the convenience of the user being able to just extract the FileSMBProperties and change what they want if they want to ever re set them. But I'm open to the idea. Anyone else have any major issues with one class?

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileSmbProperties.java
gapra-msft: This is a good idea, I might first finish up the get and create share since its an important part of this feature then take some time to refactor this

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileConstants.java
seanmcc-msft: I'm pretty sure the service expects this format exactly, down to the millionth place.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
rickle-msft: I think it's probably best to be consistent between the two.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
```diff
@@ -150,28 +152,39 @@ public DirectoryAsyncClient getSubDirectoryClient(String subDirectoryName) {
...
+
+    Mono<Response<DirectoryInfo>> createWithResponse(FileSmbProperties smbProperties, String filePermission,
+        Map<String, String> metadata, Context context) {
+        FileSmbProperties properties = smbProperties == null ? new FileSmbProperties() : smbProperties;
+
+        // Checks that file permission and file permission key are valid
+        FileExtensions.filePermissionAndKeyHelper(filePermission, properties.filePermissionKey());
+
+        // If file permission and file permission key are both not set then set default value
+        filePermission = properties.filePermission(filePermission, FileConstants.FILE_PERMISSION_INHERIT);
```
rickle-msft: It doesn't look like you check whether permission and key are both not set? Unless that happens in filePermission().

Should we also be accepting the values you're setting for the ntfs stuff instead of setting constants? (Sorry idk much about this feature. I'll trust what you say)

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
gapra-msft: Currently working on extracting filePermission

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
gapra-msft: I created package private methods in FileSmbproperties. that method filePermission(String, String) checks the internal filePermissionKey and the permission passed in for both null

gapra-msft: The NTFS stuff is like properties essentially the service will expect a string like this Directory|ReadOnly|Something, so I have an NTFSAttributesEnum that the user can use and we hide the conversion to the expected String if that makes sense. The service also returns the ntfsattributes the same way so I also have a method that converts it back into the enumset

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/FileProperties.java
gapra-msft: Doing this to be consistent

rickle-msft: As discussed, I think it's better to pull it out so it's consistent with directories

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/FileProperties.java
rickle-msft: There's an open issue on this sdk to update the interfaces for all these helper types to use get and set prefixes. I'd advise that whatever you do is consistent with that so you don't hit merge conflicts/your work isn't undone.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileExtensions.java
alzimmermsft: Agreed that it will fail to send. Generally we've erred on the side of less checks and allowing the service to handle invalid requests and send back an error message, that way if the service changes we don't need to do anything.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/FileProperties.java
gapra-msft: Talked to Alan, he said the change will happen after this preview, so it's best to stay consistent with what exists.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileSmbProperties.java
```diff
@@ -0,0 +1,292 @@
...
+        return time.format(DateTimeFormatter.ofPattern(FileConstants.SMB_DATE_STRING));
+    }
+
+    /**
+     * Given an <code>OffsetDateTime</code>, generates a {@code String} representing a date in the format needed for
+     * file SMB properties
+     * @param time the <code>OffsetDateTime</code> to be interpreted as a {@code String}
+     * @return The {@code String} representing the date
+     */
+    private static String parseFileSMBDateString(OffsetDateTime time) {
```
alzimmermsft: This is never called causing a Spotbugs issue

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/FileProperties.java
```diff
@@ -199,4 +243,81 @@ public CopyStatusType copyStatus() {
...
+        return this;
+    }
+
+    /**
+     * Sets the content md5.
+     * @param contentMD5 The content md5.
+     * @return the updated FileProperties object.
+     */
+    public FileProperties contentMD5(byte[] contentMD5) {
+        this.contentMD5 = contentMD5;
```
alzimmermsft: `ImplUtils.clone` the byte array as you do in the constructor, this is causing a Spotbugs issue

PR #5238, opened 2019-09-04
Adding okhttp dependency to spotbugs pom

anuchandy:


anuchandy:
Since changing eng files doesn't run the check, i tried it locally and it came out clean.

```
"mvn -f pom.client.xml -Djava-lts -DskipTests -Dgpg.skip spotbugs:check checkstyle:checkstyle-aggregate
```

PR #5253, opened 2019-09-05
Add synchronous Event Hub Consumer

### LINKED ISSUE #4911: Add synchronous EventHubConsumer
* Design what our synchronous receive API will be.
* Implement that.

conniey:
* Adds support for a synchronous event hub consumer.

__TODO__

* [x] Fix failing unit test
* [x] Add integration tests

Fixes #4911

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConsumer.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConsumer.java
@@ -82,9 +82,11 @@ public class EventHubConsumer implements Closeable {
         Objects.requireNonNull(maximumWaitTime, "'maximumWaitTime' cannot be null.");
 
         if (maximumMessageCount < 1) {
-            throw new IllegalArgumentException("'maximumMessageCount' cannot be less than 1.");
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("'maximumMessageCount' cannot be less than 1."));
         } else if (maximumWaitTime.isNegative() || maximumWaitTime.isZero()) {
-            throw new IllegalArgumentException("'maximumWaitTime' cannot be zero or less.");
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("'maximumWaitTime' cannot be zero or less."));
         }
 
         final Flux<EventData> events = Flux.create(emitter -> {


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConsumer.java
```diff
@@ -30,15 +36,22 @@
...
 
     EventHubConsumer(EventHubAsyncConsumer consumer, EventHubConsumerOptions options) {
-        this.consumer = Objects.requireNonNull(consumer, "'consumer' cannot be null.");
-        this.options = Objects.requireNonNull(options, "'options' cannot be null.");
+        Objects.requireNonNull(options, "'options' cannot be null.");
 
-        //TODO (conniey): Keep track of the last sequence number as each method invoked.
-        this.consumer.receive().windowTimeout(options.prefetchCount(), this.options.retry().tryTimeout());
+        this.consumer = Objects.requireNonNull(consumer, "'consumer' cannot be null.");
+        this.timeout = options.retry().tryTimeout();
```
JonathanGiles: Is it weird to have an argument of type `EventHubConsumerOptions` that you take only the retry try timeout out of and then discard the rest? Could that be the argument into the constructor instead (or doesn't it really matter as this is non-public API)?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConsumer.java
```diff
@@ -60,9 +74,43 @@
...
+     *     zero or a negative duration.
      */
     public IterableStream<EventData> receive(int maximumMessageCount, Duration maximumWaitTime) {
-        return new IterableStream<>(Flux.empty());
+        Objects.requireNonNull(maximumWaitTime, "'maximumWaitTime' cannot be null.");
+
+        if (maximumMessageCount < 1) {
+            throw new IllegalArgumentException("'maximumMessageCount' cannot be less than 1.");
+        } else if (maximumWaitTime.isNegative() || maximumWaitTime.isZero()) {
+            throw new IllegalArgumentException("'maximumWaitTime' cannot be zero or less.");
```
jsquire: (extreme) nit:  Same error condition for these two exceptions, but different wording.   It may be helpful, if not superfluous, to use consistent verbiage.

PR #5254, opened 2019-09-05
Cosmos emulator start-up timeout

mitchdenny:
This PR sets a budget of 10 minutes for Cosmos emulator start-up which is already crazy long (I would anticipate a minute or so tops).

### DIFF
--- a/eng/pipelines/templates/stages/cosmos-sdk-client.yml
+++ b/eng/pipelines/templates/stages/cosmos-sdk-client.yml
@@ -72,7 +72,7 @@ stages:
 
         - powershell: |
             Write-Host "Starting Comsos DB Emulator"
-            Start-Process "$env:Temp\Azure Cosmos DB Emulator\SourceDir\Azure Cosmos DB Emulator\CosmosDB.Emulator.exe" "/NoExplorer /NoUI /DisableRateLimiting /PartitionCount=100 /Consistency=Strong" -Verb RunAs
+            Start-Process "$env:Temp\Azure Cosmos DB Emulator\SourceDir\Azure Cosmos DB Emulator\CosmosDB.Emulator.exe" "/NoExplorer /NoUI /DisableRateLimiting /PartitionCount=100 /Consistency=Strong /overrides=eseLogFileSize:1024;" -Verb RunAs
           displayName: 'Run Public Cosmos DB Emulator'
 
         - powershell: |


mitchdenny:
@moderakh we are seeing Cosmos emulator start-up taking a long time. Generally when it starts successfully its under 10 minutes (in the 5-7 minute range). But often enough it just sits there in a StartPending state until the job times out.

This PR enforces a 10 minute timeout so that the job will fail so we aren't consuming an agent for nearly an hour waiting for the emulator to start - but we need to fix this for our builds to be reliable.

Do you have any ideas why the emulator wouldn't be starting reliably. Is there any logging we can get to figure out where it is getting stuck?

moderakh:
@milismsft FYI as this is emulator FYI.

milismsft:
The emulator startup time is heavily impacted by the quality of the agent's VM configuration and whatever the agent is doing at that particular time. We've been through these kind of issues a number of times with the DevOps guys. 

Having the agent configured with SSD's instead of regular HDD will drastically improve the startup time. Another option is to configure the emulator's startup to pass "/overrides=eseLogFileSize:1024;" as an argument.

moderakh:
@mitchdenny could you try Millis suggestion on emulator? If that works that should be a better solution as it will reduce the emulator start up time.

mitchdenny:
> @mitchdenny could you try Millis suggestion on emulator? If that works that should be a better solution as it will reduce the emulator start up time.

Done. Although I've got some more failing tests on my side :(

milismsft:
@moderakh @mitchdenny I noticed that we are starting the emulator with 100 partitions; that's 4x the default and the emulator startup will be at least 2x slower. Do we actually have tests that require so many partitions or is this done to work around some of the tests not properly deleting/removing the databases/collections after they are done? Can we try with /PartitionCount=50 instead?

mitchdenny:
> @moderakh @mitchdenny I noticed that we are starting the emulator with 100 partitions; that's 4x the default and the emulator startup will be at least 2x slower. Do we actually have tests that require so many partitions or is this done to work around some of the tests not properly deleting/removing the databases/collections after they are done? Can we try with /PartitionCount=50 instead?

Adding @danieljurek and @g2vinay for input on this question. I didn't set the original partition count so I'm not sure if there was a reason for this.

mitchdenny:
The cosmos emulator is currently behaving itself, closing this down for now.

PR #5261, opened 2019-09-05
Removed getting parent client

### LINKED ISSUE #5075: Makes Client Hierarchy Walking Consistent
There are open discussions around how clients should be able to interact with the hierarchy for the service (Account -> Container -> Blob and Account -> Share -> Directory/File). Once that is decided there may be updates required for our clients. Additionally, we should look to reduce the number of methods in ContainerClient that retrieve BlobClients, potential solution is leveraging BlobClient's as<BlobType> methods.

alzimmermsft:
Fixes #5075 

Removes the ability to get the parent client from a child client (Blob -> Container).


alzimmermsft:
Closing as this is being rolled into #5399

PR #5283, opened 2019-09-06
Netty client perf & feature improvements 

### LINKED ISSUE #5305: Investigate Connection Pool Reuse
When using Fiddler it appears there is a new CONNECT tunneling request made each time a new request is sent. This adds performance overhead which should be investigated in fixing/improving.

jianghaolu:
resolves #5305

- Bump up reactor netty to 0.8.11.RELEASE which has huge improvement gains
- Use a fixed pool that will potentially help mitigate premature connection closures (https://github.com/reactor/reactor-netty/issues/796)
- Open up more configurations on the builder for user customization


kushagraThapar:
Reactor netty 0.8.x is still immature. 0.9.0.M3 is the version which has a bunch of improvements. Specially with connection pool. They have moved away from netty fixed channel pool to their own reactor connection pool. 

Please do load testing before merging this in, specially with concurrent connections. I would also recommend testing the memory consumption, because in 0.8.x versions, they had an issue where closed connections were not going back to the pool.

jianghaolu:
@kushagraThapar - Thanks for the info! I've tested with the 0.9.0.RC1 - but it seems to use more memory than the stable 0.8.x versions. I'll be keeping an eye out for it when 0.9.0 stabilizes. 

> in 0.8.x versions, they had an issue where closed connections were not going back to the pool.

I was not aware of this. Do you have a link to a related issue? I'll be staying away from fixed size pools in this case.

jianghaolu:
Probably a no-go for preview 3 for the following reasons:
- Changing the pool to fixed is actually a bit more complicated - the requests are timed out both in the pool and in storage's retry policy. We can remove the timeout in retry policy, but for reactor-netty's pool config, we have to put in Int.MAX_VALUE to avoid channel acquisition timeout. We should look for a more elegant solution, ideally with the new pool 0.9.0 is providing. The requests are timing out because we are not honoring backpressure in our upload/download methods.
- The fixed pool in 0.8.x could run out of connections, if it's true that they might leak closed connections.

I'll do as much as I can do for 0.8.x in this PR and we can re-evaluate after preview 3. If still not good, we can talk to reactor-netty team about their schedule - if 0.9.0.RELEASE will be out before our GA we can starting depending on their RC releases.

JonathanGiles:
Let's ask @smaldini to weigh in on this eagerly, rather than wait for too long.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -123,4 +138,30 @@ public NettyAsyncHttpClientBuilder setNioEventLoopGroup(NioEventLoopGroup nioEve
...
+    /**
+     * Sets custom configurations for the reactor netty client directly on the reactor netty {@link HttpClient}. The
+     * provided configurations in this method is applied first, before the other convenience methods, thus having the
+     * least priority and may be overwritten. For example, you can set raw configurations as shown below:
+     *
+     * {@codesnippet com.azure.core.http.netty.NettyAsyncHttpClientBuilder#reactorNettyConfiguration}
+     *
+     * @param rawConfiguration configurations on the raw reactor netty {@link HttpClient}
+     * @return the updated NettyAsyncHttpClientBuilder object
+     */
```
JonathanGiles: I'm slightly worried about the low precedence of this compared with the overwriting that will occur for port, wiretap, and tcpConfiguration in the build method. I would put this at a higher precedence - what do you think?

jianghaolu: I guess you are probably right. Users will probably use this as their last advanced resort and the configurations made here are probably later than the other convenience methods.

kushagraThapar:
@JonathanGiles , @jianghaolu https://github.com/reactor/reactor-netty/releases/tag/v0.9.0.RELEASE is out. When do guys plan to update azure-sdk to this version?

This release solves most of the issues mentioned in my comment. We should plan to depend on this soon.

JonathanGiles:
We've already upgraded!

kushagraThapar:
> We've already upgraded!

Awesome, I will upgrade Cosmos SDK this week.

jianghaolu:
Closing since @anuchandy has updated the dependency.

PR #5304, opened 2019-09-09
Increase memory for Maven

mitchdenny:
This PR increases the amount of memory allocated to the JVM running underneath Maven for certain steps of the analyze job. We may consider making this global but wanted to localize this change for now.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -69,7 +69,7 @@ jobs:
           mavenPomFile: sdk/${{parameters.ServiceDirectory}}/pom.service.xml
           goals: 'package'
           options: '$(DefaultOptions) $(ProfileFlag) "-DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -DskipTests -Dinclude-template' # We include template-module so we ensure it always builds in CI
-          mavenOptions: '$(LoggingOptions)'
+          mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
           javaHomeOption: 'JDKVersion'
           jdkVersionOption: $(JavaVersion)
           jdkArchitectureOption: 'x64'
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -132,7 +132,7 @@ jobs:
         inputs:
           mavenPomFile: pom.client.xml
           options: '$(DefaultOptions) -Djava-lts -DskipTests -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true'
-          mavenOptions: '$(LoggingOptions)'
+          mavenOptions: '-Xmx3072m $(LoggingOptions)'
```
weshaggard: Any particular reason you picked this value? What is the default?

mitchdenny: Its in use elsewhere in the build. I can't remember where I got this value from - but I think I saw it elsewhere in our builds somewhere. Could potentially reduce or increase but this seems to work.

I did a little bit of reading. Some sites suggest that the allocated memory is 1/4 of system ram. The hosted VMs have ~7GB of RAM. So this would lift that quite a bit.

mitchdenny: Any other issues @weshaggard  would be good to unblock @g2vinay

weshaggard: No issues. I was just curious about why this particular number was chosen.

weshaggard:
Did you want to update the other references again? https://github.com/Azure/azure-sdk-for-java/search?q=-Xmx3072m&unscoped_q=-Xmx3072m

mitchdenny:
@weshaggard I'll do another PR with tidy up here  for the test templates but I wanted to get the team unblocked.

PR #5310, opened 2019-09-09
Reduce memory footprint of integration tests

conniey:
* Sometimes tests fail with "GC overhead limit reached"
* When tests are in playback mode, they still try to mock things like Selectable and the Reactor. This is removed since all our integration tests are skipped if AZURE_TEST_MODE is not RECORD.
* ApiTestBase renamed to IntegrationTestBase.
* Clean up mocks in EventProcessor and LoakPartitionProcessor tests

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventProcessorTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventProcessorTest.java
@@ -68,6 +68,15 @@ public class EventProcessorTest {
 
     @After
     public void teardown() {
+        consumer1 = null;
+        consumer2 = null;
+        consumer3 = null;
+        eventData1 = null;
+        eventData2 = null;
+        eventData3 = null;
+        eventData4 = null;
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/TestUtils.java
```diff
@@ -31,6 +31,8 @@
  * Contains helper methods for working with AMQP messages
  */
 final class TestUtils {
+    static final String TEST_CONNECTION_STRING = "Endpoint=sb://test-event-hub.servicebus.windows.net/;SharedAccessKeyName=dummyaccount;SharedAccessKey=ctzMq410TV3wS7upTBcunJTDLEJwMAZuFPfr0mrrA08=;EntityPath=non-existent-hub;";
```
srnagar: This `SharedAccessKey` looks legit.

conniey: Ahaha. It's definitely not. 😂 I'll make it more fake looking

PR #5315, opened 2019-09-10
Update README for checkpoint store

srnagar:
This change is to update README for checkpoint store plugin that is intended to be used with `EventProcessor` for persisting checkpoints and partition ownership details.


sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/README.md
```diff
@@ -1,4 +1,8 @@
-# Azure Checkpoint Store client library for Java using Storage Blobs.
+# Azure Checkpoint Store client library for Java using Storage Blobs
```
YijunXieMS: Azure EventHubs Checkpoint Store... ? Azure Checkpoint Store scope sounds too broad.

PR #5318, opened 2019-09-10
fix containerservice parent pom

weidongxu-microsoft:


weidongxu-microsoft:
The parent pom (containerservice/resource-manager/pom.xml) should be a aggregate pom, which would not publish jar.

But still I could not find any artifact related to these poms, either on maven or sonatype.

PR #5321, opened 2019-09-10
Disable storage and checkpoint store modules

srnagar:
Adding all the dependent modules to EH build is causing OOM in CI pipeline build. This change is to temporarily disable checkpoint store plugin module and it's storage dependencies to allow for event hubs release. 

Checkpoint store will be re-enabled after Storage Blobs preview.3 is released.

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -170,7 +170,7 @@
           <spotbugsXmlOutputDirectory>${project.build.directory}/spotbugs</spotbugsXmlOutputDirectory>
           <excludeFilterFile>spotbugs/spotbugs-exclude.xml</excludeFilterFile>
           <failOnError>true</failOnError>
-          <fork>false</fork>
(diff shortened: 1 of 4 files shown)


sdk/eventhubs/pom.service.xml
```diff
@@ -34,12 +34,12 @@
         <module>../core/azure-core</module>
         <module>../core/azure-core-test</module>
         <module>../core/azure-core-amqp</module>
-        <module>../core/azure-core-http-netty</module>
-        <module>../identity/azure-identity</module>
-        <module>../storage/azure-storage-common</module>
-        <module>../storage/azure-storage-blob</module>
+<!--        <module>../core/azure-core-http-netty</module>-->
```
joshfree: Please add a tracking bug # to the

srnagar:
No longer required.

PR #5326, opened 2019-09-10
Fetch version and artifect Id from property file.

sima-zhu:
Clean up the issues in TODO.

Put queue changes first. Will update other module once it got approved.

### DIFF
--- a/sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueConfiguration.java
+++ b/sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueConfiguration.java
@@ -2,11 +2,46 @@
 // Licensed under the MIT License.
 package com.azure.storage.queue;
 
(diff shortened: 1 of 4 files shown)


alzimmermsft:
This change is for Preview 4

sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueConfiguration.java
```diff
@@ -2,11 +2,46 @@
...
+    private final ClientLogger logger = new ClientLogger(QueueClient.class);
+    private final Properties properties = new Properties();
+
+    public QueueConfiguration() {
+        loadProperties();
+    }
+
+    private void loadProperties() {
+        try {
+            properties.load(this.getClass().getClassLoader().getResourceAsStream("client.properties"));
```
alzimmermsft: Instead of loading properties should we being using reflections and getting the package information?

PR #5346, opened 2019-09-11
Migration of azure-core-amqp implementation 

conniey:
* Migrates implementation interfaces of azure-core-amqp from azure-messaging-eventhubs.
* Removes EventDataUtil and replaces with MessageSerializer interface that deserializes/serializes objects to a proton-j message.

### DIFF
--- a/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RequestResponseChannel.java
+++ b/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RequestResponseChannel.java
@@ -103,7 +103,8 @@ public class RequestResponseChannel implements Closeable {
         this.receiveLink.setTarget(receiverTarget);
         this.receiveLink.setSenderSettleMode(SenderSettleMode.SETTLED);
         this.receiveLink.setReceiverSettleMode(ReceiverSettleMode.SECOND);
-        this.receiveLinkHandler = handlerProvider.createReceiveLinkHandler(connectionId, hostname, linkName, entityPath);
+        this.receiveLinkHandler = handlerProvider.createReceiveLinkHandler(connectionId, hostname, linkName,
+            entityPath);
         BaseHandler.setHandler(this.receiveLink, receiveLinkHandler);
 
         this.subscription = receiveLinkHandler.getDeliveredMessages().map(this::decodeDelivery).subscribe(message -> {


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ConnectionStringProperties.java
```diff
@@ -21,10 +21,10 @@
     private static final String ENTITY_PATH = "EntityPath";
     private static final String ERROR_MESSAGE_FORMAT = "Could not parse 'connectionString'. Expected format: "
         + "'Endpoint={endpoint};SharedAccessKeyName={sharedAccessKeyName};"
-        + "SharedAccessKey={sharedAccessKey};EntityPath={eventHubName}'. Actual: %s";
+        + "SharedAccessKey={sharedAccessKey};EntityPath={entityPath}'. Actual: %s";
 
     private final URI endpoint;
-    private final String eventHubName;
+    private final String entityPath;
```
srnagar: I thought we wanted to change from path to name. Are we going back to using entity path?

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/EndpointStateNotifierBase.java
```diff
@@ -56,7 +58,10 @@ public AmqpEndpointState getCurrentState() {
     void notifyError(Throwable error) {
         Objects.requireNonNull(error);
 
-        logger.error("Error occurred. {}", error.toString());
+        final String stackTrace = Stream.of(error.getStackTrace()).map(StackTraceElement::toString)
```
srnagar: I think we should consider updating ClientLogger to provide as much information about an exception (including stacktrace) as possible. For some reason, ClientLogger filters out exceptions before sending it to slf4j logger. If we don't do that, then slf4j will log the entire stacktrace.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ConnectionOptions.java
```diff
@@ -0,0 +1,125 @@
...
+import com.azure.core.amqp.models.ProxyConfiguration;
+import com.azure.core.credentials.TokenCredential;
+import reactor.core.scheduler.Scheduler;
+
+import java.util.Objects;
+
+/**
+ * A set of options for creating the AMQP connection.
+ */
+public class ConnectionOptions {
```
srnagar: Add `@Immutable` annotation.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ReactorHandlerProvider.java
```diff
@@ -61,7 +62,8 @@ ConnectionHandler createConnectionHandler(String connectionId, String hostname,
      * @param openTimeout Duration to wait for the session to open.
      * @return A new {@link SessionHandler}.
      */
-    SessionHandler createSessionHandler(String connectionId, String host, String sessionName, Duration openTimeout) {
+    public SessionHandler createSessionHandler(String connectionId, String host, String sessionName,
```
srnagar: All the method in this class should use `hostname` instead of `host` to be consistent with `ConnectionOptions` class.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/models/package-info.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * This package contains models used when AMQP protocol.
```
srnagar:
```suggestion
 * This package contains models used in AMQP protocol.
```

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventDataBatch.java
```diff
@@ -7,8 +7,8 @@
 import com.azure.core.amqp.exception.AmqpException;
 import com.azure.core.amqp.exception.ErrorCondition;
 import com.azure.core.util.logging.ClientLogger;
-import com.azure.messaging.eventhubs.implementation.AmqpConstants;
-import com.azure.messaging.eventhubs.implementation.ErrorContextProvider;
+import com.azure.core.amqp.implementation.AmqpConstants;
```
srnagar: With modules in Java 9, we should figure out if we really need to access implementation packages from other modules.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncClient.java
```diff
@@ -70,29 +71,31 @@
...
     private final EventHubProducerOptions defaultProducerOptions;
     private final EventHubConsumerOptions defaultConsumerOptions;
     private final TracerProvider tracerProvider;
 
-    EventHubAsyncClient(ConnectionOptions connectionOptions, TracerProvider tracerProvider,
+    EventHubAsyncClient(ConnectionOptions options, TracerProvider tracerProvider, MessageSerializer messageSerializer,
                         Mono<EventHubConnection> eventHubConnectionMono) {
 
-        this.connectionOptions = Objects.requireNonNull(connectionOptions, "'connectionOptions' cannot be null.");
+        this.options = Objects.requireNonNull(options, "'connectionOptions' cannot be null.");
```
srnagar: The error message should match the variable name.  `"'options' cannot be null."`

Also, I liked the previous name -  `connectionOptions`

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ConnectionStringProperties.java
conniey: In AMQP, this is the entity path, when we surface it in our client API, it is EventHubName

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/EndpointStateNotifierBase.java
conniey: I'll create an issue for it and add a TODO in this to fix this logic when the issue is resolved.
https://github.com/Azure/azure-sdk-for-java/issues/5446

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventDataBatch.java
conniey: Unfortunately this is needed because we didn't want to expose proton-j reactor APIs in the public API.

PR #5366, opened 2019-09-13
Fix group Id for cognitive services.

mitchdenny:
This PR fixes a group ID that we lost when I adjusted the parent POMs for the cognitive services libraries.


mitchdenny:
Holding this PR open until the current incident impacting GitHub checks and Azure Pipelines is resolved.

Issue #5341, opened 2019-09-11
[BUG] IllegalStateException block()/blockFirst()/blockLast() are blocking, which is not supported in thread


rvervaek:
**Describe the bug**
When uploading content via a BlobOutputStream an IllegalStateException is thrown from reactor-core's BlockingSingleSubscriber class. 

***Exception or Stack Trace***
IllegalStateException: block()/blockFirst()/blockLast() are blocking, which is not supported in thread reactor-http-nio-4

 0 = {StackTraceElement@19693} "reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:77)"
 1 = {StackTraceElement@19694} "reactor.core.publisher.Mono.block(Mono.java:1494)"
 2 = {StackTraceElement@19695} "com.azure.core.implementation.RestProxy.createResponse(RestProxy.java:456)"
 3 = {StackTraceElement@19696} "com.azure.core.implementation.RestProxy.handleRestResponseReturnType(RestProxy.java:393)"
 4 = {StackTraceElement@19697} "com.azure.core.implementation.RestProxy.lambda$handleRestReturnType$9(RestProxy.java:528)"
 5 = {StackTraceElement@19698} "reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:118)"
 6 = {StackTraceElement@19699} "reactor.core.publisher.FluxContextStart$ContextStartSubscriber.onNext(FluxContextStart.java:103)"
 7 = {StackTraceElement@19700} "org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.onNext(ScopePassingSpanSubscriber.java:96)"
 8 = {StackTraceElement@19701} "reactor.core.publisher.FluxDoOnEach$DoOnEachSubscriber.onNext(FluxDoOnEach.java:159)"
 9 = {StackTraceElement@19702} "org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.onNext(ScopePassingSpanSubscriber.java:96)"
 10 = {StackTraceElement@19703} "reactor.core.publisher.Operators$MonoSubscriber.complete(Operators.java:1515)"
 11 = {StackTraceElement@19704} "reactor.core.publisher.MonoFlatMap$FlatMapInner.onNext(MonoFlatMap.java:241)"
[...]

PR #5376, opened 2019-09-15
Rest proxy remove block from Schedulers

sima-zhu:
This is the fix which customer reported issue:
https://github.com/Azure/azure-sdk-for-java/issues/5341

Here is the reactor issue for addressing the problem of invoking Mono/Flux.block calls inside NonBlockingThread.
https://github.com/reactor/reactor-core/issues/1102

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
@@ -523,7 +523,7 @@ public class RestProxy implements InvocationHandler {
                         {httpRequest, responseStatusCode, responseHeaders}));
                 case 4:
                     return Mono.just(createResponse(ctor, new Object[] {httpRequest, responseStatusCode,
-                                      responseHeaders, bodyAsObject}));
+                        responseHeaders, bodyAsObject}));
                 case 5:
                     return response.getDecodedHeaders()
                         .map((Function<Object, Response<?>>) headers -> {


sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -521,12 +525,37 @@ private static Exception instantiateUnexpectedException(UnexpectedExceptionInfor
...
+                                            responseHeaders, bodyAsObject, null);
+                                    }
+                                    return ctor.newInstance(httpRequest, responseStatusCode,
+                                        responseHeaders, bodyAsObject, headers);
+                                } catch (IllegalAccessException | InvocationTargetException
+                                    | InstantiationException e) {
+                                    throw logger.logExceptionAsError(Exceptions.propagate(e));
+                                }
+                            });
+//                        Function<Object, Response<?>> headersToResponse = (deserializedHeaders) -> {
```
alzimmermsft: Remove the commented out code

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -521,12 +525,37 @@ private static Exception instantiateUnexpectedException(UnexpectedExceptionInfor
...
+                        return Mono.just(ctor.newInstance(httpRequest, responseStatusCode, responseHeaders));
                     case 4:
-                        return ctor.newInstance(httpRequest, responseStatusCode, responseHeaders, bodyAsObject);
+                        return Mono.just(ctor.newInstance(httpRequest, responseStatusCode, responseHeaders,
+                            bodyAsObject));
                     case 5:
-                        return ctor.newInstance(httpRequest, responseStatusCode, responseHeaders, bodyAsObject,
-                            response.getDecodedHeaders().block());
+                        return response.getDecodedHeaders()
+                            .defaultIfEmpty(NullObject.INSTANCE)
```
alzimmermsft: Is there a case when `getDecodedHeaders` returns null? If not this statement can be removed.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -521,12 +525,37 @@ private static Exception instantiateUnexpectedException(UnexpectedExceptionInfor
...
+                            .map(headers -> {
```
alzimmermsft: Should this be switched to a `flatMap` and follow a similar pattern as `handleRestResponseReturnType` which you changed in the PR as well.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: It could be. This is for deserializedHeaders which is not required to have for all clients.

jianghaolu: Can we do something like
```
return response.getDecodedHeaders()
    .map(headers -> ctor.newInstance(httpRequest, responseStatusCode,
                                        responseHeaders, bodyAsObject, headers))
    .switchIfEmpty(Mono.defer(() -> ctor.newInstance(httpRequest, responseStatusCode, responseHeaders, bodyAsObject, null)));
```

I understand what you are trying to do - but we should be able to work around the null value here.

anuchandy: you could use `switchIfEmpty` operator and avoid using `NullObject` pattern

anuchandy:
```
response.getDecodedHeaders()
    .map(headers -> {
       // code to create an instance of response that takes headers
    })
    .switchIfEmpty( // code to create an instance of response that does not take header);
```

sima-zhu: Done.

sima-zhu: With help with @anuchandy , get this piece of code use reactor way of handling null. Done.

sima-zhu: With help with @anuchandy , get this piece of code use reactor way of handling null. Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: map works well in this case and make code less overhead.
I would prefer to keep using map here.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -515,30 +517,42 @@ private static Exception instantiateUnexpectedException(UnexpectedExceptionInfor
...
         throw logger.logExceptionAsError(new RuntimeException("Cannot find suitable constructor for class " + cls));
     }
 
+    private BiFunction<Constructor<? extends Response<?>>,Object[], Response<?>> newInstanceHelper = (ctor, args) -> {
+        try {
+            return ctor.newInstance(args);
+        } catch (IllegalAccessException | InvocationTargetException | InstantiationException e) {
+            throw logger.logExceptionAsError(Exceptions.propagate(e));
+        }
+    };
```
anuchandy: nice work @sima-zhu :). I learned something today.

Just one question - Do we need an implementation of BiFunction to achieve this? or just a method is sufficient?, something like:

```java
private static Response<?> createResponse(Constructor<? extends Response<?>> ctr, Object[] args) {
    try {
	    return ctor.newInstance(args);
	} catch (IllegalAccessException | InvocationTargetException | InstantiationException e) {
	    throw logger.logExceptionAsError(Exceptions.propagate(e));
	}
}
```

sima-zhu: Yes, I think it should be enough. Making functional as it would easy for any lambda function. Since it does not take function as parameter, I agree to convert it to method.

PR #5378, opened 2019-09-16
Feature/ci mgmt

weidongxu-microsoft:
Create feature branch to test out ci.mgmt pipeline.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -23,8 +23,8 @@
 /sdk/storage/                				@rickle-msft @jaschrep-msft @gapra-msft @alzimmermsft @sima-zhu
 
 # Management Plane
-/**/resource-manager/                               @yaohaizh
-/sdk/**/azure-mgmt-*/                               @yaohaizh
+/**/resource-manager/                            	@yaohaizh
+/sdk/**/azure-mgmt-*/                            	@yaohaizh
 
 ###########
 # Eng Sys


weidongxu-microsoft:
@mitchdenny
As your suggestion I've made a feature branch to test out the CI integration. Still trying to figure the rules, since ci.data and ci.client should not run here, and new ci.mgmt should run.

weshaggard:
@weidongxu-microsoft do you have any ETA on this work? We really would like to get the root of the repo cleared up and into this new structure so folks coming to the repo can more easily see the readme file and the structure will become more consistent with the other language repos. Java is the last repo to conform to this new structure.

weidongxu-microsoft:
@weshaggard I am still working on it (but at a lower priority). Code restructuring is solved. But there is lots of try-and-error while configuring the pipeline (with too much yaml intangled and too little documationation for guidance), with the gap need to be tried-out and I am not comfortable to do this even on the feature branch. i probably will try that on another fork under my github account to faciliate (without affecting others) to see which triggers with on all these pipelines.

weshaggard:
I'd suggest we do this in steps if we can. First step is to restructure then the next step is change the pipelines.

weidongxu-microsoft:
@mitchdenny @joshfree

What is your opinion?

Do we try to do it in steps (with first step just move the repositories and config the mgmt.yaml with say 50 lines of explicit excludes on client/data-plane lib), then config the pipeline for the moved mgmt lib; or we still try to figure out the pipeline in this PR as well?

If later does anyone have better way to test the triggering of pipeline without actually sending dummy PR to do the trigger? It is because this PR also trigger client/data-plane build which was excluded and should not be trigger and I've no idea why.

mitchdenny:
Because you are developing on a branch, you can just queue your build against that branch specifically. You do this via the Azure Pipelines UX. Find your pipeline, run it, and select the branch.

weidongxu-microsoft:
@mitchdenny 

I've manually added the ADO pipeline for keyvault mgmt.
https://dev.azure.com/azure-sdk/internal/_build?definitionId=1217&_a=summary
https://dev.azure.com/azure-sdk/public/_build?definitionId=1218&_a=summary
Please take a review. Thanks.

There appears to be 2 pipeline with same config, one in public, one in internal. What is the purpose?

I still do not know how to use PipelineGenerator. Would you give me a sample command line with parameters (e.g. I've no idea what "Name of an environment variable which contains a PAT." is).

weidongxu-microsoft:
@JimSuplizio @weshaggard 

Here is still restructure + pipeline combined. Please let me know your opinion on it.

Note since the transfer to azure.core is not done on mgmt yet, these are all legacy mgmt SDKs (and as Jim mentioned in email, some of them will not build. One cause might be that the "azure-arm-client-runtime" package later than 1.6.5 no longer have "test" jar published, hence for pom had to specify 1.6.5 as test.)

And CI is not using the shared yaml template, but using a similar configure as [eng/pipelines/mgmt.yml](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/mgmt.yml)
I am not sure whether in future we would have certain period that the legacy code / pipeline would co-exist with new code / pipeline based on azure.core, and whether we need to prepare for it at present (say, `ci.mgmt-legacy.yml`, instead?).

mitchdenny:
@weidongxu-microsoft I've had a look through this and I think that this is a reasonable approach for now. Once this is merged, I'd actually like to evaluate whether we can convert the management, track 1 and track 2 client pipelines on a per service directory basis.

The reason is that I am conscious of a current performance issue in Azure Pipelines where they make a request to GitHub for each pipeline associated with the repository (and templates it references). Until they optimize this doubling up on all the pipelines could be a perf/stability issue. I think in time that will be fixed though.

We shouldn't block on this now - I just wanted to give you an indication of where my head is at.

weidongxu-microsoft:
@mitchdenny Thanks for your comments. I will wait a few days for @JimSuplizio @weshaggard 's input as well.

BTW, do you have sample usage on PipelineGenerator? I assume this is the tool to create CI pipeline automatically in batch. And when we resturcture multiple services, We will need the tool to configure the CI pipeline (instead of manually here).

mitchdenny:
I'll send you a link offline, our invocation of the pipeline generator sits in a template on a private repo.

weshaggard:
I agree with the approach of just updating the existing mgmt.yml files for now to account for the restructuring. It looks like you have currently only moved the keyvault directories so I assume you plan to do the others as well? Perhaps in another PR.

weidongxu-microsoft:
@weshaggard 

This PR is actually separating the pipeline (and make new dedicated CI for keyvault mgmt) as well.

If this one is OK, next plan is PR no.2 to move one more service, and configure the pipeline generator job to pick up it (with yaml) and create an CI in DevOp automatically.
And then we could do in batch in PR no.3 ++.

History is this:
At first we planned to move code only, and continue using mgmt.yml for CI. However this got complicated by DevOp not supporting wildcard in yaml include/exclude clause so we had difficulty to either include the specific mgmt folder or exclude data/client folder under sdk.
And after that we are trying to sperate the CI as well. And complications with that make this PR take quite some time to form.
Now I think Mitch and I could agree that the sperated mgmt CI in keyvault is in good shape and it is now possible to continue with other services.

@JimSuplizio I would still like your opinion. I do not want to surprise you with any CI change or change in sdk folder, since you probably work on CI and other utils under sdk folder as well (e.g. I do not want to move this and suddently your python script now updating thing that it does not support to update).

weidongxu-microsoft:
Jim raised an issue that there be quite many CI pipeline does not define SdkType=client. I will take some time to review this case as well.

mitchdenny:
> Jim raised an issue that there be quite many CI pipeline does not define SdkType=client. I will take some time to review this case as well.

This is by design. The default (from memory) is that the pipelines are treated as track 1 pipelines (client). But for those that aren't we define them as data. All it really controls is the spin up of Jetty at this point since the new track 2 libraries don't require it.

weidongxu-microsoft:
@mitchdenny Thanks. The issue is that previously there is at most 2-value in profile selecting, so it is such:
```
    <profile>
      <id>data</id>
      <activation>
        <property>
          <name>env.SDKTYPE</name>
          <value>data</value>
        </property>
      </activation>
      ...
    </profile>
    <profile>
      <id>client</id>
      <activation>
        <property>
          <name>env.SDKTYPE</name>
          <value>!data</value>
        </property>
      </activation>
      ...
    </profile>
```
And the activation for client is loose as `!data`. So SdkType could be `client`, could be not defined, could be anything other than `data`.

Now after we added mgmt, it would be 3-value, and maven does not support e.g. `!data && !mgmt` in profile activation. So I modified it to such:
```
    <profile>
      <id>data</id>
      <activation>
        <property>
          <name>env.SDKTYPE</name>
          <value>data</value>
        </property>
      </activation>
      ...
    </profile>
    <profile>
      <id>client</id>
      <activation>
        <property>
          <name>env.SDKTYPE</name>
          <value>client</value>
        </property>
      </activation>
      ...
    </profile>
    <profile>
      <id>mgmt</id>
      <activation>
        <property>
          <name>env.SDKTYPE</name>
          <value>mgmt</value>
        </property>
      </activation>
      ...
    </profile>
```
And the activation for client would thus be strictly as `client`.

So far I think this would not be cause, since either CI for client already defined SdkType=client, or we can spec `!mgmt` to existing CI without SdkType defined and add `mgmt` for newly added ci.mgmt.

mitchdenny:
With this change we'll need to go through all our Java pipelines and set a variable which controls this. Its not the end of the world, although perhaps this is then something that we should embed in the ci.yml, ci.data.yml and ci.mgmt.yml files instead of the definition.

At some point in the future we'll ditch ci.data.yml as it won't be necessary and then we could potentially look at converging ci.yml and ci.mgmt.yml (thats a bit further off though).

So in this PR then could you add a variable called SDKTYPE=client to ci.yml, SDKTYPE=data for ci.data.yml and SDKTYPE=management for ci.mgmt.yml?

That we we won't need to have the custom config per pipeline...

weidongxu-microsoft:
@mitchdenny 
Do you mean you would like to have SdkType in yaml, instead of in DevOp variable?
OK. I think it is reasonable to concentrated on yaml and avoid customize CI on DevOp.

mitchdenny:
Exactly. When we just had two flavors of build we could get away with sprinking the variable in the definition here and there and relying on !data to opt client on by default (since we only have a handful of track 1 builds). But this change would require it everywhere so we may as well put it in the YAML.

JimSuplizio:
@mitchdenny @weidongxu-microsoft  if we change to set SDKType=client for client builds that means that developers building client pipelines are going to always have the SDKType variable set in order to build client libraries. Right now, the way things are setup the SDKType only needs to be set for Data. Before making this type of change it should probably get approval from @JonathanGiles because I'm not sure that requiring an environment variable for client builds, when one wasn't required before, is going to fly. It's unfortunate that maven won't let allow multiple properties (or a single property with multiple value checks) inside of the activation tags for a profile.

weidongxu-microsoft:
@JimSuplizio 
IMHO, the implication of adding SdkType=client precondition is really for the developers.

CI we got all kinds of variable/template/parameters and not everyone would required to know what kind of setting would work for CI or not, mainly we just depends on existing setting and CI running there.

However if any developer is actually using this pom.service.xml for daily build, they might found out that suddenly it does not build their library (because of this added env). Of course they would only need to specifically specify the profile via "-Pclient", similar to "-Pdata".

Personally I am kind of in the camp of make "maven clean package" just work for everybody, so I sort of prefer a configure that without any setting, maven would build all of the libs. But this is not very related to this PR.

mitchdenny:
> because I'm not sure that requiring an environment variable for client builds, when one wasn't required before, is going to fly. It's unfortunate that maven won't let allow multiple properties (or a single property with multiple value checks) inside of the activation tags for a profile.

Hrm. Good point Jim - it might actually be easier to introduce a 2nd aggregator POM pom.mgmt.xml which is similar to pom.service.xml (names TBC) which is focused on management plane to split this out.

mitchdenny:
I would be in favor of moving to a model where we have a single ```pom.xml``` file in each ```sdk/[service]``` directory which replaces ```pom.service.xml```. This would be a future aggregate of client, data and management plane.

weidongxu-microsoft:
@mitchdenny 
I will process with a 2nd pom.mgmt.xml. It will minimal the impact on client/data.

weidongxu-microsoft:
Link to a prior PR https://github.com/Azure/azure-sdk-for-java/pull/5103

weidongxu-microsoft:
https://github.com/Azure/azure-sdk-for-java/issues/5682

.github/CODEOWNERS
```diff
@@ -24,6 +24,7 @@
 
 # Management Plane
 /**/resource-manager/        				@yaohaizh
+/sdk/**/azure-mgmt-*/               @yaohaizh
```
joshfree: nit: can you align the spacing with the rest of the file?  It looks like you're off by a tab or two

joshfree:
Thank you @weidongxu-microsoft for starting the migration of management plane SDKs under the /sdk root folder!  This will really help clean up the azure-sdk-for-java repo and get it clean and consistent with the other azure-sdk-for-* repo directory structures. 😄

.github/CODEOWNERS
weidongxu-microsoft: Sure. It's actually a few tabs on line above. I will change to spaces for all.

PR #5385, opened 2019-09-16
Add AuthorizationCodeCredential

### LINKED ISSUE #5306: Auth code flow for identity

jianghaolu:
resolved #5306

### DIFF
--- a/sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/AuthorizationCodeCredentialBuilder.java
+++ b/sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/AuthorizationCodeCredentialBuilder.java
@@ -34,7 +34,7 @@ public class AuthorizationCodeCredentialBuilder extends AadCredentialBuilderBase
 
     /**
      * Sets redirect URI for the Oauth 2.0 login request, which must be
-     * registered as a valid redirect URI on the application. the authorization code
+     * registered as a valid redirect URI on the application. The authorization code
      * will be sent to this URI so it must be listening on this server and is able
      * to complete the {@link AuthorizationCodeCredential} construction from there.
      * This is also called Reply URLs in some contexts.


sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/AuthorizationCodeCredentialBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+     * will be sent to this URL so it must be listening on this server and is able
+     * to complete the {@link AuthorizationCodeCredential} construction from there.
+     *
+     * @param redirectUri the redirect URI to send the authorization code
+     * @return the AuthorizationCodeCredentialBuilder itself
+     */
+    public AuthorizationCodeCredentialBuilder redirectUri(String redirectUri) {
+        this.redirectUri = redirectUri;
+        return this;
+    }
```
JonathanGiles: This method JavaDoc mentions "URL" a few times, but the API itself mentions "URI' a few times. Can you standardise on one?

sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/AuthorizationCodeCredentialBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+    public String buildLoginUrl(String... scopes) {
+        ValidationUtil.validate(getClass().getSimpleName(), new HashMap<String, Object>() {{
+                put("clientId", clientId);
+                put("redirectUri", redirectUri);
+            }});
+        return String.format("%s/oauth2/v2.0/authorize?response_type=code&response_mode=query&prompt"
+                    + "=select_account&client_id=%s&redirect_uri=%s&state=%s&scope=%s",
+                identityClientOptions.getAuthorityHost(), clientId, redirectUri, UUID.randomUUID(),
+                String.join(" ", scopes));
+    }
```
JonathanGiles: What is the use case for this method? Do you have sample code on how users would use it?

jianghaolu: The auth code flow is a 2 step process - ask the user to login at a login URL and then redirect back to the server with the authorization code to complete the token acquisition.

The credential only works when authorization code is acquired. .NET and Python are not providing any utilities for step 1 but directly ask users for the authorization code (which will make this very simple). But Java has been providing a utility method to get the login URL in the v1 runtime (https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-authentication/src/main/java/com/microsoft/azure/credentials/DelegatedTokenCredentials.java#L113) so I'm hoping to provide something similar.

The usage will look like
```java
// Create builder
AuthorizationCodeCredentialBuilder builder = new AuthorizationCodeCredentialBuilder()
    .clientId(clientId)
    .redirectUri("https://mywebservice.com/login");
// Get login URL
String loginUrl = builder.buildLoginUrl(scopes);
// Defer to user login
renderWebPageFunction(200, "Please login at " + loginUrl); // pretend this renders a web page
// Wait for auth code to come back from the AAD authority (through redirectUri), or a callback in real world web servers
String authCode = routes.get("https://mywebservice.com/login").block().getQueryParameters().get("authCode");
// complete credential construction
AuthorizationCodeCredential credential = builder.authorizationCode(authCode).build();
// Ready to use
SecretClient secretClient = new SecretClientBuilder.credential(credential).build();
```

I'm not super happy with this design but it works. I'd love to have new ideas on how to improve this.

JonathanGiles: Completely unrelatedly, one aspect of this that bothers me is that to open a users browser requires our module-info to declare a requirement on `java.desktop`. This makes this code far heavier than ideal, and I would like to explore better options on how we could deal with this. I've filed #5510 to track this.

JonathanGiles: My instinct at this stage is to say that this API is purely additive and can be provided in a follow-up release. I would be inclined for now to remove this functionality from the API, optionally moving the code into a sample application that you can point users to until such time that proper API is provided.

jianghaolu: We discussed this at the identity scrum. Scott proposed changing this method into an async callback that can ask users to provide an auth code. I'm prototyping this later today.

sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/AuthorizationCodeCredentialBuilder.java
jianghaolu: Changed everything to redirect URI but also added note this is also called the "Reply URL" in some contexts.

sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/AuthorizationCodeCredentialBuilder.java
jianghaolu: Removing for now.

sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/AuthorizationCodeCredentialBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+     * @return the AuthorizationCodeCredentialBuilder itself
+     */
+    public AuthorizationCodeCredentialBuilder authorizationCode(String authCode) {
+        this.authCode = authCode;
+        return this;
+    }
+
+    /**
+     * Sets redirect URI for the Oauth 2.0 login request, which must be
+     * registered as a valid redirect URI on the application. the authorization code
```
JonathanGiles:
```suggestion
     * registered as a valid redirect URI on the application. The authorization code
```

joshfree:
@schaabs PTAL and rebase and merge assuming you approve as-is

Issue #5372, opened 2019-09-13
[FEATURE REQ] Enable Multi-Proxy Support in Track 2 Azure.Core


joshfree:
**Is your feature request related to a problem? Please describe.**
Some customers expect to run in an environment with a global proxy setting.  Our Track 2 SDKs should be able to have a proxy setting independent of that and not overwrite the global setting.

**Describe the solution you'd like**
The ability to have multi-proxy support so that no connection can overwrite the proxy settings for another connection.

**Additional context**
This is a feature request we should investigate for Preview 4

alzimmermsft:
PR #5387 adds documentation and samples on how to configure a proxy for both HttpClient implementations that we have plugin packages for, does this help resolve the multi-proxy support issue?

conniey:
This is an ask from one of our customers too, who don't want a global proxy configuration. Track 1 Event Hubs has the concept of a ProxyConfiguration that people can pass when creating the client. This also exists in Track 2 Event Hubs.

joshfree:
@jianghaolu can you please prioritize this for the November preview release?

jianghaolu:
Maybe I'm missing something here but why not just create another HttpClient with a different proxy setup?

conniey:
This may already be supported in Track 2, just have to ensure it works. In Track 1 storage libraries, they use HttpURLConnection, which only supports the global proxy configuration.

joshfree:
Just following up here -- we should create a quick sample showing Digest and Basic Authenticated Proxy working on Track 2.  If there are any feature gaps in Azure.Core around Digest then we should open up additional tracking items to complete this work in an upcoming monthly SDK update.

amishra-dev:
@joshfree, samples is a good idea (just showing how to do 2 pipelines with 2 proxies). And as long as we can confirm internally that it works and we can integrate this with eventhub, I think we are good.

jianghaolu:
this is a simple example of talking to Key vault and Storage using different proxies: 
```
public class Main {
    private static final ProxyOptions proxy1 = new ProxyOptions(ProxyOptions.Type.HTTP,
            new InetSocketAddress("10.127.70.25", 8085))
        .setCredentials("azureuser", "StrongPass!123");

    private static final ProxyOptions proxy2 = new ProxyOptions(ProxyOptions.Type.HTTP,
            new InetSocketAddress("10.127.70.25", 8888));

    public static void main(String... args) throws Exception {
[...]

PR #5387, opened 2019-09-16
Add examples for configuring Netty and OkHttp clients with a proxy

alzimmermsft:
Adds documentation to address #5372 

Adds documentation and examples around setting proxies in our Netty and OkHttp client implementations.

### DIFF
--- a/sdk/core/azure-core-http-netty/README.md
+++ b/sdk/core/azure-core-http-netty/README.md
@@ -29,11 +29,10 @@ The following sections provide several code snippets covering some of the most c
 
 ### Create a Simple Client
 
-Create the simplest Netty Http client.
+Create a Netty Http client that uses port 80 and has no proxy.
 
 ```java
-HttpClient client = new NettyAsyncHttpClientBuilder()
-    .build();
+HttpClient client = new NettyAsyncHttpClientBuilder().build();
 ```
 
 ### Create a Client with Proxy
--- a/sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
+++ b/sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
@@ -63,7 +63,7 @@ public class NettyAsyncHttpClientBuilder {
                             nettyProxy = ProxyProvider.Proxy.SOCKS5;
                             break;
                         default:
-                            throw logger.logExceptionAsWarning(new IllegalStateException(
+                            throw logger.logExceptionAsError(new IllegalStateException(
                                 String.format("Unknown Proxy type '%s' in use. Not configuring Netty proxy.",
                                     proxyOptions.getType())));
                     }
--- a/sdk/core/azure-core-http-okhttp/README.md
+++ b/sdk/core/azure-core-http-okhttp/README.md
@@ -29,11 +29,10 @@ The following sections provide several code snippets covering some of the most c
 
 ### Create a Simple Client
 
-Create the simplest OkHttp client.
+Create an OkHttp client using a connection timeout of 60 seconds and a read timeout of 120 seconds.
 
 ```java
-HttpClient client = new OkHttpAsyncHttpClientBuilder()
-    .build();
+HttpClient client = new OkHttpAsyncHttpClientBuilder().build();
 ```
 
 ### Create a Client with Proxy


sdk/core/azure-core-http-netty/README.md
```diff
@@ -22,6 +22,30 @@ Azure Core Netty HTTP client is a plugin for the azure-core HTTP client API.
...
+- [Create a Simple Client](#create-a-simple-client)
+- [Create a Client with Proxy](#create-a-client-with-proxy)
+
+### Create a Simple Client
+
+Create the simplest Netty Http client.
+
+```java
+HttpClient client = new NettyAsyncHttpClientBuilder()
+    .build();
```
JonathanGiles: I always like seeing `.build()` on the same line if there is no other calls. Thoughts?

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -68,7 +69,7 @@ public OkHttpAsyncHttpClientBuilder addNetworkInterceptor(Interceptor networkInt
      * @param networkInterceptors the interceptors to add
      * @return the updated OkHttpAsyncHttpClientBuilder object
      */
-    public OkHttpAsyncHttpClientBuilder setNetworkInterceptors(List<Interceptor> networkInterceptors) {
+    public OkHttpAsyncHttpClientBuilder networkInterceptors(List<Interceptor> networkInterceptors) {
```
JonathanGiles: Good catch!

sdk/core/azure-core-http-netty/README.md
```diff
@@ -22,6 +22,30 @@ Azure Core Netty HTTP client is a plugin for the azure-core HTTP client API.
...
 ## Examples
 
+The following sections provide several code snippets covering some of the most common client configuration scenarios.
+
+- [Create a Simple Client](#create-a-simple-client)
+- [Create a Client with Proxy](#create-a-client-with-proxy)
+
+### Create a Simple Client
+
+Create the simplest Netty Http client.
```
srnagar: Would it be useful to elaborate a bit on what a simple Netty client would do? For e.g. what are the default values it would be configured with if someone creates this client.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -63,10 +63,9 @@ public NettyAsyncHttpClient build() {
                             nettyProxy = ProxyProvider.Proxy.SOCKS5;
                             break;
                         default:
-                            throw logger.logExceptionAsWarning(
-                                new IllegalStateException(
-                                    "Unknown Proxy type '" + proxyOptions.getType()
-                                        + "' in use. Not configuring Netty proxy."));
+                            throw logger.logExceptionAsWarning(new IllegalStateException(
```
srnagar: Is this going to not use Proxy or will this result in failing to create a HttpClient? If we fail to create the HttpClient, it should be logged as error. If we create a client without proxy, then warning is fine.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -68,7 +69,7 @@ public OkHttpAsyncHttpClientBuilder addNetworkInterceptor(Interceptor networkInt
...
         this.networkInterceptors = Objects.requireNonNull(networkInterceptors, "networkInterceptors cannot be null.");
```
srnagar: Should this check for both null and empty?

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
alzimmermsft: Good catch

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
alzimmermsft: Only checking nullness should suffice, no `Interceptors` is allowed.

PR #5394, opened 2019-09-17
Minor cleanups for azure-core:

JonathanGiles:
1) Move AzureEnvironment from azure-core to azure-core-management
2) Adding a convenience 'stream()' method to HttpHeaders class
3) Renaming HttpPipeline.httpClient() to HttpPipeline.getHttpClient()
4) Renamed HttpRequest.buffer() to HttpRequest.clone()
5) Making HttpResponse immutable by setting the internal HttpRequest to the constructor, rather than having it on the setter


sdk/core/azure-core/src/main/java/com/azure/core/http/HttpResponse.java
```diff
@@ -15,7 +15,11 @@
  * The type representing response of {@link HttpRequest}.
  */
 public abstract class HttpResponse implements Closeable {
-    private HttpRequest request;
+    private final HttpRequest request;
```
conniey: Yes for changing this to final. Woo!

PR #5405, opened 2019-09-18
Return HttpClient interface rather than Netty impl

JonathanGiles:
Make the NettyAsyncHttpClient class non-public, and have the NettyAsyncHttpClientBuilder return HttpClient instead.


sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
```diff
@@ -9,6 +9,7 @@
 import com.azure.core.http.HttpRequest;
 import com.azure.core.http.HttpResponse;
 import com.azure.core.http.ProxyOptions;
+import com.azure.core.http.netty.NettyAsyncHttpClientBuilder;
```
alzimmermsft: Import isn't needed

JonathanGiles: It is - for the JavaDoc reference.

PR #5410, opened 2019-09-18
First pass at cleaning up the Configuration API

JonathanGiles:


sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/SharedTokenCacheCredential.java
```diff
@@ -44,11 +42,11 @@
      * @param identityClientOptions the options for configuring the identity client
      */
     SharedTokenCacheCredential(String username, String clientID, IdentityClientOptions identityClientOptions) {
-        this.configuration = ConfigurationManager.getConfiguration().clone();
+        this.configuration = Configuration.getGlobalConfiguration().clone();
 
         if (username == null) {
-            if (configuration.contains(BaseConfigurations.AZURE_USERNAME)) {
-                this.username = configuration.get(BaseConfigurations.AZURE_USERNAME);
+            if (configuration.contains(Configuration.PROPERTY_AZURE_USERNAME)) {
```
alzimmermsft: Same as the other if/else question

alzimmermsft:
Need to fix the Configuration tests

JonathanGiles:
The two test failures are related to a timeout on storage: `com.azure.storage.blob.RetryTest`. I am fine with merging as-is, based on feedback from Alan earlier.

sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
```diff
@@ -19,6 +19,115 @@
      */
     public static final Configuration NONE = new NoopConfiguration();
 
+    // Default properties - these are what we read from the environment
+    /**
+     * URI of the proxy for HTTP connections.
+     */
+    public static final String PROPERTY_HTTP_PROXY = "HTTP_PROXY";
```
conniey: What if the user has a proxy configured, but also requires authentication? Is there any username, password, or authentication type (digest vs basic vs none) variable name we should use?

sdk/core/azure-core/src/test/java/com/azure/core/ConfigurationTests.java
```diff
@@ -90,28 +89,28 @@ public void foundConfigurationIsConverted() {
...
-        boolean initialTracingDisabled = Boolean.parseBoolean(ConfigurationManager.getConfiguration().get(BaseConfigurations.AZURE_TRACING_DISABLED));
-        System.setProperty(BaseConfigurations.AZURE_TRACING_DISABLED, Boolean.toString(!initialTracingDisabled));
-        assertNotEquals(initialTracingDisabled, ConfigurationManager.getConfiguration().get(BaseConfigurations.AZURE_TRACING_DISABLED));
+        boolean initialTracingDisabled = Boolean.parseBoolean(Configuration.getGlobalConfiguration().get(Configuration.PROPERTY_AZURE_TRACING_DISABLED));
+        System.setProperty(Configuration.PROPERTY_AZURE_TRACING_DISABLED, Boolean.toString(!initialTracingDisabled));
+        assertNotEquals(initialTracingDisabled, Configuration.getGlobalConfiguration().get(Configuration.PROPERTY_AZURE_TRACING_DISABLED));
 
         // Cleanup the test
-        System.setProperty(BaseConfigurations.AZURE_TRACING_DISABLED, Boolean.toString(initialTracingDisabled));
+        System.setProperty(Configuration.PROPERTY_AZURE_TRACING_DISABLED, Boolean.toString(initialTracingDisabled));
```
conniey: I would wrap this in a `try/finally` block, if the test fails at line 111, this will never get reset. Same with line 103.

sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
JonathanGiles: I don't believe we have anything defined in the spec for this.

sdk/core/azure-core/src/test/java/com/azure/core/ConfigurationTests.java
JonathanGiles: Tests never fail! :-)

conniey: giphy

sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
conniey: Hmm.. Feels like a gap in the spec, but part of another conversation. :)

JonathanGiles: Yes, please do chase up on it separately so we can get it clarified, if necessary.

PR #5416, opened 2019-09-18
Split Event Hubs related connection and mapping from AMQPConnection

conniey:
Removes some in-class instantiations and rely on more dependency injection to increase testibility and reuse in Service Bus.
* Remove ManagementNode from ReactorConnection and into EventHubReactorConnection
  * This is not part of an "AMQP Connection", it is a concept that we added in for Event Hubs
  * This allows us to dump ReactorConnection into azure-core-amqp for ServiceBus to use
* Make mapping of responses from Management node more generic
  * Remove AmqpResponseMapper and replace with ManagementResponseMapper
* Abstract hard coded string for "audience" from AzureTokenManagerProvider
  * This manages TokenProviders. :D

Related to: https://github.com/Azure/azure-sdk-for-java/pull/5346

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubConsumerIntegrationTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubConsumerIntegrationTest.java
@@ -249,9 +249,10 @@ public class EventHubConsumerIntegrationTest extends IntegrationTestBase {
             // Act
             final IterableStream<EventData> receive = consumer.receive(100, Duration.ofSeconds(3));
 
-            System.out.println("Sending second batch.");
+            logger.info("Sending second batch.");
             producer.send(events2);
-            System.out.println("Receiving second batch.");
+
+            logger.info("Receiving second batch.");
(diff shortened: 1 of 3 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubConsumerIntegrationTest.java
```diff
@@ -229,7 +230,10 @@ public void multipleConsumers() {
      * Verify that we can receive until the timeout multiple times.
      */
     @Test
-    public void receiveUntilTimeoutMultipleTimes() {
+    public void receiveUntilTimeoutMultipleTimes() throws IOException {
+        this.consumer.close();
```
srnagar: Not sure why this is required but if you intended to close out the consumer that was created from previous tests, you can instead use this:

```
@After
public void cleanUp() {
    this.consumer.close();
    this.consumer = null;
}
```

conniey: Oh, that was because I don't use the consumer created in beforeTest() and create my own. I'll remove it. It's not necessary.

PR #5419, opened 2019-09-18
Block blob client side encryption

gapra-msft:
I am in the process of moving this into a new PR. If you haven't reviewed this yet, please wait.
Moved to #5528

### DIFF
--- a/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 package com.azure.storage.blob.cryptography;
 
 import com.azure.core.http.HttpHeader;
@@ -59,7 +62,7 @@ public class BlobDecryptionPolicy implements HttpPipelinePolicy {
      * the key and use it.
      *
      * @param key An object of type {@link IKey} that is used to wrap/unwrap the content encryption key.
-     * @param keyResolver  The key resolver used to select the correct key for decrypting existing blobs.
+     * @param keyResolver The key resolver used to select the correct key for decrypting existing blobs.
      */
     BlobDecryptionPolicy(IKey key, IKeyResolver keyResolver) {
         this.keyWrapper = key;
@@ -69,14 +72,14 @@ public class BlobDecryptionPolicy implements HttpPipelinePolicy {
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
         // 1. Expand the range of download for decryption
-        String RANGE_HEADER = "x-ms-range";
+        String rangeHeader = "x-ms-range";
         EncryptedBlobRange encryptedRange = EncryptedBlobRange.getEncryptedBlobRangeFromHeader(
-            context.getHttpRequest().getHeaders().getValue(RANGE_HEADER));
+            context.getHttpRequest().getHeaders().getValue(rangeHeader));
 
         // Assumption: Download is the only API on an encrypted client that sets x-ms-range
         // Only set the x-ms-range header if it already exists
-        if(context.getHttpRequest().getHeaders().getValue(RANGE_HEADER) != null) {
-            context.getHttpRequest().getHeaders().put(RANGE_HEADER, encryptedRange.toBlobRange().toString());
+        if (context.getHttpRequest().getHeaders().getValue(rangeHeader) != null) {
+            context.getHttpRequest().getHeaders().put(rangeHeader, encryptedRange.toBlobRange().toString());
         }
 
         // 2. Replace the body of the response with a decrypted version of the body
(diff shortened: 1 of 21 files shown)


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,403 @@
...
+    /**
+     * Package-private constructor for use by {@link EncryptedBlobClientBuilder}.
+     */
+    EncryptedBlockBlobAsyncClient(AzureBlobStorageImpl constructImpl, String snapshot, CpkInfo cpk,
+        BlobEncryption encryptionPolicy) {
+        super(constructImpl, snapshot, cpk);
+        this.encryptionPolicy = encryptionPolicy;
+    }
+
+    public EncryptedBlockBlobAsyncClient wrap(BlockBlobAsyncClient client, IKey key, IKeyResolver keyResolver) {
```
gapra-msft: I need to work more on this, but can someone check if I have the right idea here? Is there some attributes I'm missing

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,403 @@
...
+    static final int BLOB_DEFAULT_UPLOAD_BLOCK_SIZE = 4 * Constants.MB;
+    private static final int BLOB_MAX_UPLOAD_BLOCK_SIZE = 100 * Constants.MB;
+    private final ClientLogger logger = new ClientLogger(EncryptedBlockBlobAsyncClient.class);
+
+    private final BlobEncryption encryptionPolicy;
+
+    /**
+     * Package-private constructor for use by {@link EncryptedBlobClientBuilder}.
+     */
+    EncryptedBlockBlobAsyncClient(AzureBlobStorageImpl constructImpl, String snapshot, CpkInfo cpk,
```
gapra-msft: Opinions on having encryptedblockblobasyncclient instead just wrap a blockblobasyncclient (that way the commitblock and stageblock will not be repeated in this class)

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlobRange.java
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * How many bytes to download, including the adjustments for encryption block boundaries and the IV.
+     * Must be greater than or equal to 0 if specified.
+     */
+    private Long adjustedDownloadCount;
+
+    static EncryptedBlobRange getEncryptedBlobRangeFromHeader(String stringRange) {
+        // Null case
+        if (stringRange == null) {
```
alzimmermsft: Should this use `ImplUtils.isNullOrEmpty`?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlobRange.java
```diff
@@ -0,0 +1,139 @@
...
+            return new EncryptedBlobRange(null);
+        }
+        // Non-null case
+        String trimmed = stringRange.substring(6); // Trim off the "bytes=" part
```
alzimmermsft: Instead of using a magic number could we get the character position of the first `=`?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
alzimmermsft: Wrapping BlockBlobAsyncClient should work if the method internals are the same outside of this client having an additional policy.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
alzimmermsft: Might want to make this method part of the builder instead of a constructor?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
gapra-msft: It's not just an additional policy, it also encrypts the data before sending it on it's way through the pipeline

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
gapra-msft: already done in my next iteration

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,491 @@
...
+     */
+    BlobDecryptionPolicy(IKey key, IKeyResolver keyResolver) {
+        this.keyWrapper = key;
+        this.keyResolver = keyResolver;
+    }
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        // 1. Expand the range of download for decryption
+        String rangeHeader = "x-ms-range";
```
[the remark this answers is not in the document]
gapra-msft: I did that but this is the only method it is called in so I made it local, would it be better to just make it a global?

gapra-msft:
Moved to #5528

PR #5421, opened 2019-09-18
Get URL with resource name and snapshot

sima-zhu:
Consist with blob getURL which returned the URL with scheme, host, resource name and snapshot

### DIFF
--- a/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
+++ b/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
@@ -101,13 +101,13 @@ public class DirectoryAsyncClient {
      * @throws RuntimeException If the directory is using a malformed URL.
      */
     public URL getDirectoryUrl() {
-        String directoryURLString = String.format("%s/%s/%s", azureFileStorageClient.getUrl(),
-            shareName, directoryPath);
+        StringBuilder directoryURLString = new StringBuilder(azureFileStorageClient.getUrl()).append("/")
+            .append(shareName).append("/").append(directoryPath);
         if (snapshot != null) {
-            directoryURLString = String.format("%s?snapshot=%s", directoryURLString, snapshot);
+            directoryURLString.append("?snapshot=").append(snapshot);
         }
         try {
-            return new URL(directoryURLString);
+            return new URL(directoryURLString.toString());
         } catch (MalformedURLException e) {
             throw logger.logExceptionAsError(new RuntimeException(
                 String.format("Invalid URL on %s: %s" + getClass().getSimpleName(), directoryURLString), e));
--- a/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileAsyncClient.java
+++ b/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileAsyncClient.java
@@ -128,12 +128,13 @@ public class FileAsyncClient {
      * @throws RuntimeException If the file is using a malformed URL.
      */
     public URL getFileUrl() {
-        String fileURLString = String.format("%s/%s/%s", azureFileStorageClient.getUrl(), shareName, filePath);
+        StringBuilder fileURLString = new StringBuilder(azureFileStorageClient.getUrl()).append("/")
+            .append(shareName).append("/").append(filePath);
         if (snapshot != null) {
-            fileURLString = String.format("%s?snapshot=%s", fileURLString, snapshot);
+            fileURLString.append("?snapshot=").append(snapshot);
         }
         try {
-            return new URL(fileURLString);
+            return new URL(fileURLString.toString());
         } catch (MalformedURLException e) {
             throw logger.logExceptionAsError(new RuntimeException(
                 String.format("Invalid URL on %s: %s" + getClass().getSimpleName(), fileURLString), e));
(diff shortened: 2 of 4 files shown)


sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
```diff
@@ -100,12 +100,16 @@
      * @throws RuntimeException If the directory is using a malformed URL.
      */
     public URL getDirectoryUrl() {
+        String directoryURLString = String.format("%s/%s/%s", azureFileStorageClient.getUrl(),
```
alzimmermsft: Do we split out the share and directory path in the URL? If not we don't need this `String.format`.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileAsyncClient.java
```diff
@@ -127,12 +127,15 @@
      * @throws RuntimeException If the file is using a malformed URL.
      */
     public URL getFileUrl() {
+        String fileURLString = String.format("%s/%s/%s", azureFileStorageClient.getUrl(), shareName, filePath);
```
alzimmermsft: Same question as before

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/ShareAsyncClient.java
```diff
@@ -95,12 +95,15 @@
      * @throws RuntimeException If the share is using a malformed URL.
      */
     public URL getShareUrl() {
+        String shareURLString = String.format("%s/%s", azureFileStorageClient.getUrl(), shareName);
```
alzimmermsft: Same question

sima-zhu:
> Are there changes that are needed for Blobs as well?

James has the implementation done for blob. However, we are using two different ways of getting the URL. Will discuss him which is a better way to go and update all three clients.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
sima-zhu: The format is like
`https://sima.file.core.windows.net/sharename/directoryPath`

alzimmermsft:
I think you merged another PR into this PR given there are BlobOuputStream and BlobInputStream changes now.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
```diff
@@ -101,12 +101,16 @@
...
+            shareName, directoryPath);
+        if (snapshot != null) {
+            directoryURLString = String.format("%s?snapshot=%s", directoryURLString, snapshot);
```
jaschrep-msft: Instead of doing multiple String::format calls across if statements, why not use a string builder?

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileAsyncClient.java
```diff
@@ -128,12 +128,15 @@
...
+        if (snapshot != null) {
+            fileURLString = String.format("%s?snapshot=%s", fileURLString, snapshot);
```
jaschrep-msft: StringBuilder comment.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/ShareAsyncClient.java
```diff
@@ -96,12 +96,15 @@
...
+        if (snapshot != null) {
+            shareURLString = String.format("%s?snapshot=%s", shareURLString, snapshot);
```
jaschrep-msft: StringBuilder comment.

sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueAsyncClient.java
```diff
@@ -86,7 +86,7 @@
      */
     public URL getQueueUrl() {
         try {
-            return new URL(client.getUrl());
+            return new URL(String.format("%s/%s", client.getUrl(), queueName));
         } catch (MalformedURLException ex) {
             logger.error("Queue URL is malformed");
             throw logger.logExceptionAsError(new RuntimeException("Queue URL is malformed"));
```
jaschrep-msft:
```java
logger.error("Queue URL is malformed");
throw logger.logExceptionAsError(new RuntimeException("Queue URL is malformed"));
```
Does this log twice?

sima-zhu: Good catch.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
sima-zhu: Any benefits of StringBuilder?

sima-zhu: Made the changes.

PR #5429, opened 2019-09-19
Fix signing by removing artifact from publish list.

mitchdenny:
This PR addresses the storage pipeline failing during signing.


alzimmermsft:
@gapra-msft When we officially open the encryption PR we will want to un-commented these lines.

PR #5435, opened 2019-09-19
Optimisation for RestProxy createResponse API.

JonathanGiles:


sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -0,0 +1,216 @@
...
+     * 2. (httpRequest, statusCode, headers, body)
+     * 3. (httpRequest, statusCode, headers)
+     *
+     * @param responseClass the response class
+     * @return optional with located constructor, empty optional if there is no match
+     */
+    Optional<ResponseConstructor> get(Class<? extends Response<?>> responseClass) {
+
+        if (!this.cache.containsKey(responseClass)) {
+            this.cache.put(responseClass, Arrays.stream(responseClass.getDeclaredConstructors())
```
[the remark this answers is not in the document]
anuchandy: Through out the life time of a cache instance, it's going to contain only "One of the Ctr" of a "Specific Response type". Hence if we use tuple `{int, classOf(Response)}` as key and if cache has key `{4, FooResponse.class}` then `{3, FooResponse.class}` cannot co-exists. The smallest level of granularity is Response type.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -0,0 +1,216 @@
...
+                                                    Object.class),
+                                            ctrMethodHandle,
+                                            ctrMethodHandle.type())
+                                            .getTarget().invoke();
+                                    return Optional.of(new ResponseConstructor(5, f5));
+                                default:
+                                    return Optional.empty();
+                            }
+
+                        } catch (Throwable t) {
```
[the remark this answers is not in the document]
anuchandy: It’s because invoke throws Throwable,  `MethodHandle::invoke(Object... args) throws Throwable` :(

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * A concurrent cache of {@link Response} constructors.
+ */
+final class ResponseConstructorsCache {
+    private final ClientLogger logger = new ClientLogger(ResponseConstructorsCache.class);
+    private final Map<Class<?>, Optional<ResponseConstructor>> cache = new ConcurrentHashMap<>();
```
JonathanGiles: We shouldn't need to have the `ResponseConstructor` wrapped with an `Optional` here. If it is in the map, we should be certain it exists.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -0,0 +1,213 @@
...
+                            return Optional.of(new ResponseConstructor(5, f5));
+                        default:
+                            return Optional.<ResponseConstructor>empty();
+                    }
+
+                } catch (Throwable t) {
+                    throw logger.logExceptionAsError(new RuntimeException(t));
+                }
+            }));
+        return this.cache.get(responseClass);
```
JonathanGiles: You don't need to call get - you can just return directly from the `computeIfAbsent` call

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -0,0 +1,213 @@
...
+        this.cache.computeIfAbsent(responseClass, responseCls -> Arrays.stream(responseCls.getDeclaredConstructors())
+            .filter(constructor -> {
+                int paramCount = constructor.getParameterCount();
+                return paramCount >= 3 && paramCount <= 5;
+            })
+            .sorted(Comparator.comparingInt(Constructor::getParameterCount))
+            .findFirst()
+            .flatMap(constructor -> {
+                try {
+                    MethodHandles.Lookup lookup = MethodHandles.lookup();
```
JonathanGiles: We discussed moving this out and reusing it - is that possible?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -0,0 +1,213 @@
...
+                    MethodHandle ctrMethodHandle = lookup.unreflectConstructor(constructor);
+                    switch (constructor.getParameterCount()) {
+                        case 3:
+                            Object f3 = LambdaMetafactory.metafactory(lookup,
+                                    "apply",
+                                    MethodType.methodType(ResponseFunc3.class),
+                                    MethodType.methodType(Object.class,
+                                            Object.class,
+                                            int.class,
+                                            Object.class),
```
JonathanGiles: Should these method type calls be more specific, e.g. `HttpRequest.class, int.class, HttpHeaders.class`?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -0,0 +1,213 @@
...
+                                            Object.class,
+                                            Object.class),
+                                    ctrMethodHandle,
+                                    ctrMethodHandle.type())
+                                    .getTarget().invoke();
+                            return Optional.of(new ResponseConstructor(5, f5));
+                        default:
+                            return Optional.<ResponseConstructor>empty();
+                    }
+
```
JonathanGiles: Get rid of whitespace

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -0,0 +1,213 @@
...
+            })
+            .sorted(Comparator.comparingInt(Constructor::getParameterCount))
+            .findFirst()
+            .flatMap(constructor -> {
+                try {
+                    MethodHandles.Lookup lookup = MethodHandles.lookup();
+                    MethodHandle ctrMethodHandle = lookup.unreflectConstructor(constructor);
+                    switch (constructor.getParameterCount()) {
+                        case 3:
+                            Object f3 = LambdaMetafactory.metafactory(lookup,
```
JonathanGiles: Can the type of `f3` be more specific than `Object`?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -0,0 +1,213 @@
...
+                            "Response constructor with expected parameters not found."));
+            }
+        }
+    }
+
+    @FunctionalInterface
+    private interface ResponseFunc3 {
+        Object apply(Object httpRequest,
+                     int responseStatusCode,
+                     Object responseHeaders);
```
JonathanGiles: Can we type these fully?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
anuchandy: agree, given this is in hot path, we needs to be careful about any extra allocation. I'm seeing that  even usage of Stream and associated operators for doing this very basic logic (filter, sorting an array with few element [in worst case 3 elements] and mappings) is causing allocation and adding overhead. 

Interestingly, there is already a bug in the current code base, we do sorting in ascending order and take the first, in reality we need to take the last because out of ctrs with 3 to 5 params we need the one with most number of params.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
anuchandy: nice catch!

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
anuchandy: LambdaMetafactory 's `.invoke()` returns `Object` (this method is not generic) and this Object is of `ResponseFunc3`.

We could cast it to the specific type `ResponseFunc3` but our cache value is of type `Object`. We do casting when we use it to invoke ctr.

I was initially worried about usage of `Object` but I'm seeing this is happening quite often when working with low level LambdaMetaFactory.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
anuchandy: this is something I'm still exploring, the scope of `MethodHandles.lookup`.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
anuchandy: I'm testing using specific types for non-primitive (in j8 & j11) to see whether we can do so.

anuchandy: seems we can use specific types, will update the PR.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
anuchandy: yes it's possible.

anuchandy:
@JonathanGiles thanks! I  will add few more stuffs to this PR, Should be ready for final review by EOD tomorrow 09/26.

anuchandy:
@JonathanGiles, I added JMH based profiling to compare response caching and creation based on LambdaMetaFactory (LMF)/Reflection. The goals are:

1. If any one doubt the gain then code can be pulled and tried.
2. It seems in each version of JDK, reflection is getting improved, this benchmark helps to validate this with future JDKs.

anuchandy:
Here is the JMH result on Java 8:

LambdaMetaFactory_Reflection

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
anuchandy: yes possible, switched to singleton Lookup.

anuchandy:
JMH result on Java 8 [CachedLambdaMetaFactory vs CachedReflection vs NonCachedReflection]

MetaFactory_ReflectionCached_ReflectionNonCached

JonathanGiles:
@g2vinay FYI

anuchandy:
@g2vinay by default JMH outputs result in table format. To get the graph, follow below steps:

1. In IntelliJ for `ResponseConstructorsCacheBenchMark` "Run Configuration", specify `-rf json` as "program argument"
2. Run the bench mark i.e. `ResponseConstructorsCacheBenchMark::main()`
3. JMH will save the result as json format in a file with name `jmh-result.json`
4. Go to https://jmh.morethan.io and upload this file to generate graph

anuchandy:
@JonathanGiles the JMH framework produces lot of files at compile time like:

```
core/azure-core/target/generated-test-sources/test-annotations/com/azure/core/implementation/generated/
ResponseConstructorsCacheBenchMark_jmhType_B2.java
```

our checkstyle is getting executed for all these files and failing, how can we skip these files from being validated?

JonathanGiles:
We can add a suppression rule for generated-test-sources

anuchandy:
@JonathanGiles @mssfang I've added suppression for `generated-test-sources`

```
  <!-- Checkstyle rules should not check files in generated-test-sources -->
   <suppress checks="[a-zA-Z0-9]*" files="[/\\]generated-test-sources[/\\]"/>
```

sdk/core/azure-core/src/test/java/com/azure/core/implementation/ResponseConstructorsCacheBenchMark.java
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.implementation;
+
+import com.azure.core.http.rest.Response;
+import com.azure.core.implementation.util.TypeUtil;
+import org.openjdk.jmh.Main;
```
srnagar: Would this work for all distributions of JDK? If not, and if it's only benchmarking, can we move this to a separate module?

anuchandy: I had a quick chat with @karianna and @JonathanGiles on this.

JMH works best on OpenJDK or OpenJDK derivatives (could be enhanced for Eclipse OpenJ9 or Azul Zing benchmarking), but at this moment our sdks are targeting only OpenJDK variants. It’s unlikely that we’ll be using a non OpenJDK derivative on Azure.

JonathanGiles:
In the fullness of time (i.e. in the next few weeks), I have asked @g2vinay to create a centralised benchmarking suite inside the /eng directory of our repo. At this stage I would like to see this benchmark moved into there also.

Issue #5431, opened 2019-09-19
ServiceBus: Message lock auto-renew


storey247:
Working with the c# ServiceBus SDK it would seem that there is an auto-renew of the locks that happens behind the scenes for long running message related workflows. However, in the Java world this functionality doesn't exist. This means when I am working with messages I have to always write boilerplate code that will execute a scheduled thread pool to renew the locks, or my messages get abandoned etc.

**Describe the solution you'd like**
Having looked at the SDK it would seem that the `MessageReceiver` could have an additional property added to set auto-renew on or possibly take a configuration object to its constructor. If this is set to true then we could have a new object of type `AutoRenewMessage` that implements `IMessage` or even could inherit `Message` and extend it.

`MessageReciever` can be extended to have a `ScheduledThreadPool` that will keep track of these `AutoRenewMessage` using a new `Runnable` for example: 

```
public class MessageLockAutoRenewTask implements Runnable {

    private final IMessageReceiver receiver;
    private final IMessage message;

    public MessageLockAutoRenewTask(IMessageReceiver receiver, IMessage message) {
        this.receiver = receiver;
        this.message = message;
    }

    @Override
    public void run() {
        try {
            this.receiver.renewMessageLock(message.getLockToken());
        } catch (InterruptedException ex) {
            // do something else here
        } catch (ServiceBusException ex) {
            // do something here
        }
    }
}
```

If the `MessageReceiver` was to be closed/disposed then the `ScheduledThreadPool` can be abandoned and all executing tasks will naturally close down.

When a message is abandoned, DLQ'd or completed it would just be a matter of checking if the `IMessage` is of type `Message` or `AutoRenewMessage` and then making sure the object stopped its scheduled task.

**Describe alternatives you've considered**

- Extend the existing MessageReciever to have a new AutoRenewMessageReciever and then maintain message states on a concurrent collection.
- Implement a new intermediary class similar to the c# library whereby all calls to the `CoreMessageReciever` go through this intermediary object and it tracks the state of messages in flight and renews the locks etc.

**Additional context**
- I am currently working on a POC of this new code and was just wanting to raise this Issue as a proposal before I get round to raising a PR to get feedback regarding the approach and if the PR would be accepted?

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Description Added
- [x] Expected solution specified

jfggdl:
[...]

PR #5444, opened 2019-09-20
Implementing auto renew locks for servicebus

storey247:
This is a POC linked to issue #5431 

It is a new MessageReciever implementation that is responsible for adding the additional timer necessary to automatically renew locks when they are going to expire. It will also clean up after itself when the receiver is closed or if messages are completed/abandoned etc.

Feedback welcome :)


JonathanGiles:
@yvgopal Please review this PR and merge or close, as appropriate. Thanks.

yvgopal:
Before we review the code, the question to ask is "Is it necessary to put this in the SDK?".  Application can always start a renew loop after receiving the message. As it was discussed in the thread #5431, I still believe it is not necessary to be included in the SDK at this point in time. So I will put this on hold, until we decide to include it as a feature in the SDK. Thanks for submitting the PR.

PR #5452, opened 2019-09-20
Remove Event Hubs logic from ReactorSession. Fix NullPointerException

conniey:
This mitigates the cognitive burden in the large PR: https://github.com/Azure/azure-sdk-for-java/pull/5346. 
* Split Event Hubs related knowledge from ReactorSession
* Fix NullPointerException that occurred in CustomIOHandler
* Fix NullPointerException in SessionHandler

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/EventHubReactorSession.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/EventHubReactorSession.java
@@ -16,11 +16,27 @@ import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 
-public class EventHubReactorSession extends ReactorSession implements EventHubSession {
+/**
+ * An AMQP session for Event Hubs.
+ */
+class EventHubReactorSession extends ReactorSession implements EventHubSession {
     private static final Symbol EPOCH = Symbol.valueOf(AmqpConstants.VENDOR + ":epoch");
     private static final Symbol RECEIVER_IDENTIFIER_NAME = Symbol.valueOf(AmqpConstants.VENDOR + ":receiver-name");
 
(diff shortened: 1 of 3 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/EventHubReactorSession.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.qpid.proton.amqp.Symbol;
+import org.apache.qpid.proton.amqp.UnknownDescribedType;
+import org.apache.qpid.proton.engine.Session;
+import reactor.core.publisher.Mono;
+
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+
+public class EventHubReactorSession extends ReactorSession implements EventHubSession {
```
conniey: I'll add documentation to this.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/handler/CustomIOHandler.java
```diff
@@ -32,7 +32,13 @@ public void onTransportClosed(Event event) {
 
     @Override
     public void onUnhandled(Event event) {
-        // logger.verbose("Unhandled event: {}, {}", event.getEventType(), event.toString());
-        super.onUnhandled(event);
+        // logger.verbose("Unhandled event: {}, {}", event.getEventType(), event);
```
srnagar: You can uncomment this. Since it's verbose, it's most likely not logged and if verbose logging is enabled, then I guess logging this is okay.

conniey: Yeah. It's very, very verbose when it is enabled, because there are so many reactor events we don't bother to listen to. I guess that's why it's DEBUG.

PR #5455, opened 2019-09-20
Storage InputStream and OutputStream

sima-zhu:
The PR is to abstract storage input and output stream into common layer. Implemented StorageFileInputStream and StorageFileOutputStream. Make BlobInputStream and BlobOutputStream extends the common one.

Todo list:
1. Rewrite some of the comments.
2. Extract some common conversion to azure core if any.
3. Tests

### DIFF
--- a/sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/APISpec.groovy
+++ b/sdk/storage/azure-storage-file/src/test/java/com/azure/storage/file/APISpec.groovy
@@ -41,7 +41,7 @@ class APISpec extends Specification {
     // Test name for test method name.
     def methodName
 
-    def testMode = getTestMode()
+    static def testMode = getTestMode()
     String connectionString
 
     // If debugging is enabled, recordings cannot run as there can only be one proxy at a time.


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
```diff
@@ -0,0 +1,385 @@
...
+package com.azure.storage.common;
+
+import com.azure.core.util.logging.ClientLogger;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+/**
+ * Provides an input stream to read a given blob resource.
```
jaschrep-msft: This has been extracted away to work with more than just blobs.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobInputStream.java
```diff
@@ -81,21 +50,16 @@
...
      *
-     * @param blobClient A {@link BlobAsyncClientBase} object which represents the blob that this stream is associated
-     * with.
+     * @param blobClient A {@link BlobAsyncClient} object which represents the blob that this stream is associated with.
      * @param accessCondition An {@link BlobAccessConditions} object which represents the access conditions for the
      * blob.
      * @throws StorageException An exception representing any error which occurred during the operation.
      */
-    BlobInputStream(final BlobAsyncClientBase blobClient, final BlobAccessConditions accessCondition)
+    BlobInputStream(final BlobAsyncClient blobClient, final BlobAccessConditions accessCondition)
```
rickle-msft: I think this should be just BlobClient. Streams are used in the sync world, so it doesn't make sense to require a customer to pass in an async client, especially since we always block anyway.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
```diff
@@ -0,0 +1,391 @@
...
+/**
+ * Provides an input stream to read a given storage resource.
+ */
+public abstract class StorageInputStream extends InputStream {
+    private final ClientLogger logger = new ClientLogger(StorageInputStream.class);
+
+    /**
+     * A flag to determine if the stream is faulted, if so the last error will be thrown on next operation.
+     */
+    private volatile boolean streamFaulted;
```
rickle-msft: >privat [](start = 4, length = 6)

I think this should be protected. I don't see a need for the parent and child types to have their own streamFaulted fields.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
```diff
@@ -0,0 +1,391 @@
...
+
+    /**
+     * Holds the last exception this stream encountered.
+     */
+    private IOException lastError;
```
rickle-msft: >private [](start = 4, length = 7)

protected?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
```diff
@@ -0,0 +1,391 @@
...
+
+    /**
+     * Holds an absolute byte position of the current read position.
+     */
+    private long currentAbsoluteReadPosition;
+
+    /**
+     * Holds the absolute byte position of the start of the current buffer.
+     */
+    private long bufferStartOffset;
```
rickle-msft: protected?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
```diff
@@ -0,0 +1,391 @@
...
+
+    /**
+     * Holds the length of the current buffer in bytes.
+     */
+    private int bufferSize;
```
rickle-msft: protected?

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/StorageFileInputStream.java
```diff
@@ -0,0 +1,94 @@
...
+     * Initializes a new instance of the StorageFileInputStream class. Note that if {@code fileRangeOffset} is not
+     * {@code 0} or {@code fileRangeLength} is not {@code null}, there will be no content MD5 verification.
+     *
+     * @param fileAsyncClient A {@link FileAsyncClient} object which represents the blob
+     * that this stream is associated with.
+     * @param fileRangeOffset The offset of file range data to begin stream.
+     * @param fileRangeLength How much data the stream should return after fileRangeOffset.
+     * @throws StorageException An exception representing any error which occurred during the operation.
+     */
+    StorageFileInputStream(final FileAsyncClient fileAsyncClient, long fileRangeOffset, Long fileRangeLength)
```
rickle-msft: >FileAsyncClient [](start = 33, length = 15)

sync client

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageOutputStream.java
```diff
@@ -0,0 +1,155 @@
...
+    /*
+     * Holds the write threshold of number of bytes to buffer prior to dispatching a write. For block blob this is the
+     * block size, for page blob this is the Page commit size.
+     */
+    final int writeThreshold;
+
+    /*
+     * Holds the last exception this stream encountered.
+     */
+    volatile IOException lastError;
```
rickle-msft: protected

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageOutputStream.java
```diff
@@ -0,0 +1,155 @@
...
+ * StorageOutputStream
+ */
+public abstract class StorageOutputStream extends OutputStream {
+    final ClientLogger logger = new ClientLogger(StorageOutputStream.class);
+
+    /*
+     * Holds the write threshold of number of bytes to buffer prior to dispatching a write. For block blob this is the
+     * block size, for page blob this is the Page commit size.
+     */
+    final int writeThreshold;
```
rickle-msft: protected

rickle-msft:
>     }

I think if you make commit() an abstract method on the base and then have it be a no op on the file output stream, then you don't have to duplicate any of this. This whole thing can move into the base. 

---
Refers to: sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java:80 in 0aa814b. [](commit_id = 0aa814b1f5f476f5ab18d636bbb8b5412e34b29e, deletion_comment = False)

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobInputStream.java
sima-zhu: Change to sync will dramatically change the PR, which we can not fit the PR before Friday.
Async to Sync client change is an internal implementation change, so it is ok to do a post preview PR if it is needed.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
sima-zhu: Done.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
sima-zhu: Done.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
sima-zhu: Done.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/StorageFileInputStream.java
sima-zhu: Prefer no change for this PR before preview

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageOutputStream.java
sima-zhu: Done.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageOutputStream.java
sima-zhu: This is only used in StorageOutputStream, all child class will pass in the writeThreshold to base class

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageOutputStream.java
sima-zhu: Done.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageOutputStream.java
sima-zhu: I made it private.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
sima-zhu: Done. Good point

sima-zhu:
> > ```
> > }
> > ```
> 
> I think if you make commit() an abstract method on the base and then have it be a no op on the file output stream, then you don't have to duplicate any of this. This whole thing can move into the base.
> 
> Refers to: sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java:80 in [0aa814b](https://github.com/Azure/azure-sdk-for-java/commit/0aa814b1f5f476f5ab18d636bbb8b5412e34b29e). [](commit_id = [0aa814b](https://github.com/Azure/azure-sdk-for-java/commit/0aa814b1f5f476f5ab18d636bbb8b5412e34b29e), deletion_comment = False)

@rickle-msft  Could you clarify which is the base BlobOutputStream or StorageOutputStream?

I did not find commit() in StorageFileOutputStream, It is an abstract in BlobOutputStream and only used by blockBlobStream. The reason I cannot move to the base is it asks for client which is only available in child stream.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobInputStream.java
joshfree: Can this feedback be addressed by adding a BlobsyncClient overload in a later update?

sima-zhu: Only private or protected method and constructor needs to update for this change. Our public APIs do not take client as param. We can simply change to sync instead of adding sync overload.

PR #5457, opened 2019-09-20
Move Storage queue internal model types to implementation

jianghaolu:
The config is at https://github.com/Azure/azure-sdk-for-java/pull/5457/files#diff-fa4268ef9b2ae13e8190a1ff1d145bf7R36

### DIFF
--- a/sdk/storage/azure-storage-queue/pom.xml
+++ b/sdk/storage/azure-storage-queue/pom.xml
@@ -79,20 +79,32 @@
 
   <profiles>
     <profile>
-      <id>java8storage</id>
+      <id>java8</id>
       <activation>
         <jdk>[1.8,9)</jdk>
       </activation>
       <build>
         <plugins>
           <plugin>
-            <!-- Skip maven compiler and use gmavenplus plugin instead -->
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueAsyncClient.java
```diff
@@ -20,13 +20,13 @@
 import com.azure.storage.queue.implementation.AzureQueueStorageImpl;
 import com.azure.storage.queue.models.DequeuedMessage;
 import com.azure.storage.queue.models.EnqueuedMessage;
-import com.azure.storage.queue.models.MessageIdUpdateHeaders;
-import com.azure.storage.queue.models.MessageIdsUpdateResponse;
+import com.azure.storage.queue.implementation.MessageIdUpdateHeaders;
```
alzimmermsft: Do we want the model classes to just move to `implementation` or `models.implementation`?

jianghaolu: @JonathanGiles

alzimmermsft: Or `implementation.models`

PR #5459, opened 2019-09-23
Fixed repository type.

mitchdenny:
This is just a quick PR to fix a typo I made in the ```sdk/search/ci.yml``` definition.

### DIFF
--- a/sdk/search/ci.yml
+++ b/sdk/search/ci.yml
@@ -4,7 +4,7 @@
 resources:
   repositories:
     - repository: azure-sdk-build-tools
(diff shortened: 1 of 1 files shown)


sdk/search/ci.yml
```diff
@@ -4,7 +4,7 @@
 resources:
   repositories:
     - repository: azure-sdk-build-tools
-      type: github
+      type: git
```
weshaggard: Out of curiosity how does this error manifest itself? Does it simply fail to find this template?

PR #5469, opened 2019-09-23
Use Version Based Profile to Compile Groovy

alzimmermsft:
Updates the Blob and Queue encryption packages to use version based profiles to compile the Groovy tests to support Java 8 and Java 9+.

### DIFF
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/RequestRetryOptions.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/RequestRetryOptions.java
@@ -144,7 +144,7 @@ public final class RequestRetryOptions {
         long delay;
         switch (this.retryPolicyType) {
             case EXPONENTIAL:
-                delay = ((tryCount - 1) * (tryCount - 1) - 1L) * this.retryDelayInMs;
+                delay = (powOfTwo(tryCount - 1) - 1L) * this.retryDelayInMs;
                 break;
 
             case FIXED:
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob-cryptography/pom.xml
```diff
@@ -72,32 +72,148 @@
...
+              </dependency>
+            </dependencies>
+          </plugin>
+
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>${maven-surefire-plugin.version}</version>
+            <configuration>
+              <argLine>
```
srnagar: You can remove this argLine property until you have source code that requires these `exports` and `opens`.

sdk/storage/azure-storage-queue-cryptography/pom.xml
```diff
@@ -72,32 +72,148 @@
...
+              </dependency>
+            </dependencies>
+          </plugin>
+
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>${maven-surefire-plugin.version}</version>
+            <configuration>
+              <argLine>
```
srnagar: same here

alzimmermsft: Removed

sdk/storage/azure-storage-blob-cryptography/pom.xml
alzimmermsft: Removed

PR #5474, opened 2019-09-23
Replace EventDataUtil and ManagementResponseMapper with MessageSerializer. Add EventHubClient.getEventHubName

conniey:
* Makes EventHubClient.getEventHubName() a public property to align with all the other languages.
* Increase the testability and reusability by removing static EventDataUtil class and replacing it with MessageSerializer interface
  * Service Bus and Event Hubs serialize/deserialize AMQP messages in different ways.
* Remove ManagementResponseMapper and roll logic into MessageSerializer

Removes cognitive burden in: https://github.com/Azure/azure-sdk-for-java/pull/5346 by changing less classes.

### DIFF
--- a/sdk/core/azure-core-amqp/src/main/java/module-info.java
+++ b/sdk/core/azure-core-amqp/src/main/java/module-info.java
@@ -5,6 +5,7 @@ module com.azure.amqp {
     requires com.azure.core;
 
     requires reactor.core;
+    requires proton.j;
 
     exports com.azure.core.amqp;
     exports com.azure.core.amqp.exception;


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncClient.java
```diff
@@ -95,6 +98,15 @@
             .setScheduler(connectionOptions.getScheduler());
     }
 
+    /**
+     * Gets the Event Hub name this client interacts with.
+     *
+     * @return The Event Hub name this client interacts with.
+     */
+    public String getEventHubName() {
```
srnagar: Do we want to make this public?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncProducer.java
```diff
@@ -16,16 +11,24 @@
...
 import com.azure.core.util.logging.ClientLogger;
+import com.azure.messaging.eventhubs.implementation.AmqpConstants;
 import com.azure.messaging.eventhubs.implementation.AmqpSendLink;
 import com.azure.messaging.eventhubs.implementation.ErrorContextProvider;
-import com.azure.messaging.eventhubs.implementation.EventDataUtil;
+import com.azure.messaging.eventhubs.implementation.MessageSerializer;
 import com.azure.messaging.eventhubs.models.BatchOptions;
 import com.azure.messaging.eventhubs.models.EventHubProducerOptions;
 import com.azure.messaging.eventhubs.models.SendOptions;
+import org.apache.qpid.proton.amqp.messaging.MessageAnnotations;
```
srnagar: Since this is in event hubs package, can we avoid using qpid packages directly? 

(Ignore if this is planned for another PR).

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
```diff
@@ -1,42 +1,41 @@
...
 import org.apache.qpid.proton.amqp.Symbol;
 import org.apache.qpid.proton.amqp.messaging.AmqpValue;
 import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
 import org.apache.qpid.proton.amqp.messaging.Data;
 import org.apache.qpid.proton.amqp.messaging.MessageAnnotations;
+import org.apache.qpid.proton.amqp.messaging.Section;
 import org.apache.qpid.proton.message.Message;
 
+import java.nio.ByteBuffer;
+import java.util.Date;
```
srnagar: Is it possible to move away from `java.util.Date`? IntelliJ keeps complaining about its usage as it's obsolete.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/MessageSerializer.java
```diff
@@ -0,0 +1,37 @@
+package com.azure.messaging.eventhubs.implementation;
+
+import org.apache.qpid.proton.message.Message;
+
+/**
+ * Serializer to translate objects to and from proton-j {@link Message messages}.
+ */
+public interface MessageSerializer {
```
srnagar: Would this be a candidate for core-amqp module instead of Event Hub module?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClient.java
```diff
@@ -55,6 +55,15 @@
             .setScheduler(connectionOptions.getScheduler());
     }
 
+    /**
+     * Gets the Event Hub name this client interacts with.
+     *
+     * @return The Event Hub name this client interacts with.
+     */
+    public String getEventHubName() {
```
srnagar: Should this be public?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
```diff
@@ -100,6 +105,110 @@ private static Message toAmqpMessage(String partitionKey, EventData eventData) {
         return message;
     }
 
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T> T deserialize(Message message, Class<T> clazz) {
+        Objects.requireNonNull(message, "'message' cannot be null.");
+        Objects.requireNonNull(clazz, "'class' cannot be null.");
```
srnagar:
```suggestion
        Objects.requireNonNull(clazz, "'clazz' cannot be null.");
```

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncClient.java
conniey: Yes. In one of our stand-ups every other language mentioned they had this exposed, so we were the only ones who did not.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncProducer.java
conniey: We already expose the proton-j `Message` class because an AmqpSendLink only takes Messages. This isn't publicly exposed though.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
conniey: It's a type that is serialized and returned from qpid-proton-j. We'd have to update the underlying library to use Instant if we wanted to remove this.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/MessageSerializer.java
conniey: Yup! This will be moved there in the PR https://github.com/Azure/azure-sdk-for-java/pull/5346

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncClient.java
conniey: For example, .NET has this public:

https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/eventhub/Azure.Messaging.EventHubs/src/EventHubClient.cs#L30-L42

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClient.java
conniey: See: https://github.com/Azure/azure-sdk-for-java/pull/5474#discussion_r327369142

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/MessageSerializer.java
conniey: Actually. I'll move it there now.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
JonathanGiles: FWIW, I don't mind using this internally, I just don't want us to use Date in our public APIs. This is a good candidate for a CheckStyle rule.

PR #5480, opened 2019-09-24
Add module-info to appconfig

srnagar:
This change is to add module-info.java to AppConfiguration module and update module names to be consistent.

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/module-info.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/module-info.java
@@ -6,4 +6,7 @@ module com.azure.data.appconfiguration {
 
     opens com.azure.data.appconfiguration.implementation to com.fasterxml.jackson.databind;
     opens com.azure.data.appconfiguration.models to com.fasterxml.jackson.databind;
+
+    exports com.azure.data.appconfiguration;
+    exports com.azure.data.appconfiguration.models;
 }


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/module-info.java
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module com.azure.data.appconfiguration {
+    requires com.azure.core;
+
```
alzimmermsft: This should `exports com.azure.data.appconfiguration` and `exports com.azure.data.appconfiguration.models`

sdk/core/azure-core-http-okhttp/src/main/java/module-info.java
```diff
@@ -1,16 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-module azure.core.http.okhttp {
+module com.azure.core.http.okhttp {
 
     requires transitive com.azure.core;
+    requires com.azure.core.test;
```
alzimmermsft: Do we want to make `com.azure.core.test` `requires transitive`?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/module-info.java
srnagar: Good point! Updated

sdk/core/azure-core-http-okhttp/src/main/java/module-info.java
conniey: I'm curious why we would want to make it transitive? I thought `transitive` is if you wanted anyone consuming azconfig to implicitly reads azure-core-test. I don't think we ship anything publicly that requires this?

srnagar: I don't think anyone wanting to use okhttp module would also want to use test module. If required we can update later.

PR #5492, opened 2019-09-24
Convert credential scopes param to TokenRequest

jianghaolu:
.NET PR for reference https://github.com/Azure/azure-sdk-for-net/pull/7655


sdk/core/azure-core/src/main/java/com/azure/core/credentials/TokenRequest.java
```diff
@@ -0,0 +1,51 @@
...
+        return scopes;
+    }
+
+    /**
+     * Sets the scopes required for the token.
+     * @param scopes the scopes required for the token
+     * @return the TokenRequest itself
+     */
+    public TokenRequest setScopes(List<String> scopes) {
+        this.scopes.clear();
```
conniey: Should there be any error handling? (ie. scopes == null)

sdk/identity/azure-identity/src/test/java/com/azure/identity/credential/UsernamePasswordCredentialTest.java
```diff
@@ -62,10 +63,10 @@ public void testValidUserCredential() throws Exception {
         // test
         UsernamePasswordCredential credential =
             new UsernamePasswordCredentialBuilder().clientId(clientId).username(username).password(password).build();
-        AccessToken token = credential.getToken(scopes1).block();
+        AccessToken token = credential.getToken(request1).block();
```
conniey: Should be making use of StepVerifier instead of .block() on these tests.

jianghaolu:
Couldn't run tests on my PC with JDK 8. Will fix the tests on my Mac when I get in the office.

Issue #4481, opened 2019-07-18
Blob Encryption


kurtzeborn:
@kurtzeborn commented on [Thu Jul 18 2019](https://github.com/Azure/azure-sdk/issues/518)

Issue #5511, opened 2019-09-25
Introduce opinionated azure-core


JonathanGiles:
We should consider introducing an opinionated azure-core POM that pre-selects some plugin dependencies, for example, azure-core-http-netty will be a listed dependency.

We may choose to call this what is now `azure-core`, and to introduce another pom that is `azure-core-lite` (pending a better name). We could then simplify our documentation to simply tell users to use `azure-core`, and to direct them to our documentation (in the wiki or elsewhere) for situations where they want to use a different HTTP client.

Another approach is to include netty by default and document how people can exclude netty instead. This is possible in Maven using the following syntax:

```xml
   <dependency>
        <groupId>org.springframework.security</groupId>
        <artifactId>spring-security-taglibs</artifactId>
        <version>${spring-security.version}</version>
        <exclusions>
            <exclusion>
                <groupId>org.springframework.security</groupId>
                <artifactId>spring-security-web</artifactId>
            </exclusion>
        </exclusions>
    </dependency>
```

JonathanGiles:
Fixed by #5577

PR #5528, opened 2019-09-25
Storage block blob encryption

gapra-msft:
Addresses #4481

### DIFF
--- a/sdk/storage/azure-storage-blob-cryptography/pom.xml
+++ b/sdk/storage/azure-storage-blob-cryptography/pom.xml
@@ -44,11 +44,6 @@
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
-      <artifactId>azure-keyvault-keys</artifactId>
-      <version>4.0.0-preview.5</version>
-    </dependency>
-    <dependency>
-      <groupId>com.azure</groupId>
       <artifactId>azure-core-test</artifactId>
       <version>1.0.0-preview.6</version>
       <scope>test</scope>
--- a/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
@@ -8,7 +8,6 @@ import com.azure.core.cryptography.AsyncKeyEncryptionKey;
 import com.azure.core.cryptography.AsyncKeyEncryptionKeyResolver;
 import com.azure.core.http.policy.UserAgentPolicy;
 import com.azure.core.implementation.util.ImplUtils;
-import com.azure.security.keyvault.keys.cryptography.models.KeyWrapAlgorithm;
 import com.azure.core.http.HttpPipeline;
 import com.azure.core.http.policy.HttpPipelinePolicy;
 import com.azure.core.util.logging.ClientLogger;
@@ -39,7 +38,7 @@ import java.util.Objects;
  * <li>Credential set through {@link #credential(SharedKeyCredential)} , {@link #sasToken(String)}, or
  * {@link #connectionString(String)} if the container is not publicly accessible.
  * <li>Key and key wrapping algorithm (for encryption) and/or key resolver (for decryption) must be specified
- * through {@link #key(AsyncKeyEncryptionKey, KeyWrapAlgorithm)} and {@link #keyResolver(AsyncKeyEncryptionKeyResolver)}
+ * through {@link #key(AsyncKeyEncryptionKey, String)} and {@link #keyResolver(AsyncKeyEncryptionKeyResolver)}
  * </ul>
  *
  * <p>
@@ -61,7 +60,7 @@ public final class EncryptedBlobClientBuilder extends BaseBlobClientBuilder<Encr
 
     private AsyncKeyEncryptionKey keyWrapper;
     private AsyncKeyEncryptionKeyResolver keyResolver;
-    private KeyWrapAlgorithm keyWrapAlgorithm;
+    private String keyWrapAlgorithm;
 
     /**
      * Creates a new instance of the EncryptedBlobClientBuilder
@@ -208,10 +207,10 @@ public final class EncryptedBlobClientBuilder extends BaseBlobClientBuilder<Encr
      * Sets the encryption key parameters for the client
      *
      * @param key An object of type {@link AsyncKeyEncryptionKey} that is used to wrap/unwrap the content encryption key
-     * @param keyWrapAlgorithm The {@link KeyWrapAlgorithm} used to wrap the key.
+     * @param keyWrapAlgorithm The {@link String} used to wrap the key.
      * @return the updated EncryptedBlobClientBuilder object
      */
-    public EncryptedBlobClientBuilder key(AsyncKeyEncryptionKey key, KeyWrapAlgorithm keyWrapAlgorithm) {
+    public EncryptedBlobClientBuilder key(AsyncKeyEncryptionKey key, String keyWrapAlgorithm) {
         this.keyWrapper = key;
         this.keyWrapAlgorithm = keyWrapAlgorithm;
         return this;
--- a/sdk/storage/pom.service.xml
+++ b/sdk/storage/pom.service.xml
@@ -14,7 +14,6 @@
     <module>../core/azure-core-test</module>
     <module>../core/azure-core-http-netty</module>
     <module>../identity/azure-identity</module>
-    <module>../keyvault/azure-keyvault-keys</module>
     <module>azure-storage-common</module>
     <module>azure-storage-blob</module>
     <module>azure-storage-blob-batch</module>
--- a/eng/jacoco-test-coverage/pom.xml
+++ b/eng/jacoco-test-coverage/pom.xml
@@ -30,7 +30,7 @@
     <azure-core-test.version>1.0.0-preview.6</azure-core-test.version>
     <azure-data-appconfiguration.version>1.0.0-preview.5</azure-data-appconfiguration.version>
     <azure-identity.version>1.0.0-preview.5</azure-identity.version>
-    <azure-keyvault.version>4.0.0-preview.4</azure-keyvault.version>
+    <azure-keyvault.version>4.0.0-preview.5</azure-keyvault.version>
     <azure-messaging-eventhubs.version>5.0.0-preview.4</azure-messaging-eventhubs.version>
     <azure-messaging-eventhubs-checkpointstore-blob.version>1.0.0-preview.2</azure-messaging-eventhubs-checkpointstore-blob.version>
     <azure-storage-common.version>12.0.0-preview.4</azure-storage-common.version>
--- a/eng/spotbugs-aggregate-report/pom.xml
+++ b/eng/spotbugs-aggregate-report/pom.xml
@@ -25,7 +25,7 @@
     <azure-core.version>1.0.0-preview.6</azure-core.version>
     <azure-data-appconfiguration.version>1.0.0-preview.5</azure-data-appconfiguration.version>
     <azure-identity.version>1.0.0-preview.5</azure-identity.version>
-    <azure-keyvault.version>4.0.0-preview.4</azure-keyvault.version>
+    <azure-keyvault.version>4.0.0-preview.5</azure-keyvault.version>
     <azure-messaging-eventhubs.version>5.0.0-preview.4</azure-messaging-eventhubs.version>
     <azure-messaging-eventhubs-checkpointstore-blob.version>1.0.0-preview.2</azure-messaging-eventhubs-checkpointstore-blob.version>
     <azure-storage-common.version>12.0.0-preview.4</azure-storage-common.version>
--- a/sdk/keyvault/azure-keyvault-certificates/pom.xml
+++ b/sdk/keyvault/azure-keyvault-certificates/pom.xml
@@ -12,7 +12,7 @@
 
   <groupId>com.azure</groupId>
   <artifactId>azure-keyvault-certificates</artifactId>
-  <version>4.0.0-preview.4</version>
+  <version>4.0.0-preview.5</version>
 
   <name>Microsoft Azure client library for KeyVault Certificates</name>
   <description>This module contains client library for Microsoft Azure KeyVault Certificates.</description>
--- a/sdk/keyvault/azure-keyvault-keys/pom.xml
+++ b/sdk/keyvault/azure-keyvault-keys/pom.xml
@@ -13,7 +13,7 @@
 
   <groupId>com.azure</groupId>
   <artifactId>azure-keyvault-keys</artifactId>
-  <version>4.0.0-preview.4</version>
+  <version>4.0.0-preview.5</version>
 
   <name>Microsoft Azure client library for KeyVault Keys</name>
   <description>This module contains client library for Microsoft Azure KeyVault Keys.</description>
--- a/sdk/keyvault/azure-keyvault-secrets/pom.xml
+++ b/sdk/keyvault/azure-keyvault-secrets/pom.xml
@@ -11,7 +11,7 @@
 
   <groupId>com.azure</groupId>
   <artifactId>azure-keyvault-secrets</artifactId>
-  <version>4.0.0-preview.4</version>
+  <version>4.0.0-preview.5</version>
 
   <name>Microsoft Azure client library for KeyVault Secrets</name>
   <description>This module contains client library for Microsoft Azure KeyVault Secrets.</description>
--- a/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
@@ -8,7 +8,6 @@ import com.azure.core.cryptography.AsyncKeyEncryptionKey;
 import com.azure.core.http.rest.Response;
 import com.azure.core.implementation.util.FluxUtil;
 import com.azure.core.util.logging.ClientLogger;
-import com.azure.security.keyvault.keys.cryptography.models.KeyWrapAlgorithm;
 import com.azure.storage.blob.BlobAsyncClient;
 import com.azure.storage.blob.implementation.AzureBlobStorageImpl;
 import com.azure.storage.blob.models.AccessTier;
@@ -77,15 +76,15 @@ public class EncryptedBlobAsyncClient extends BlobAsyncClient {
     private final AsyncKeyEncryptionKey keyWrapper;
 
     /**
-     * A {@link KeyWrapAlgorithm} that is used to wrap/unwrap the content key during encryption.
+     * A {@link String} that is used to wrap/unwrap the content key during encryption.
      */
-    private final KeyWrapAlgorithm keyWrapAlgorithm;
+    private final String keyWrapAlgorithm;
 
     /**
      * Package-private constructor for use by {@link EncryptedBlobClientBuilder}.
      */
     EncryptedBlobAsyncClient(AzureBlobStorageImpl constructImpl, String snapshot, String accountName,
-        AsyncKeyEncryptionKey key, KeyWrapAlgorithm keyWrapAlgorithm) {
+        AsyncKeyEncryptionKey key, String keyWrapAlgorithm) {
         super(constructImpl, snapshot, null, accountName);
         this.keyWrapper = key;
         this.keyWrapAlgorithm = keyWrapAlgorithm;
@@ -259,10 +258,10 @@ public class EncryptedBlobAsyncClient extends BlobAsyncClient {
             keyWrappingMetadata.put(CryptographyConstants.AGENT_METADATA_KEY,
                 CryptographyConstants.AGENT_METADATA_VALUE);
 
-            return this.keyWrapper.wrapKey(keyWrapAlgorithm.toString(), aesKey.getEncoded())
+            return this.keyWrapper.wrapKey(keyWrapAlgorithm, aesKey.getEncoded())
                 .map(encryptedKey -> {
                     WrappedKey wrappedKey = new WrappedKey(
-                        this.keyWrapper.getKeyId().block(), encryptedKey, keyWrapAlgorithm.toString());
+                        this.keyWrapper.getKeyId().block(), encryptedKey, keyWrapAlgorithm);
 
                     // Build EncryptionData
                     EncryptionData encryptionData = new EncryptionData()
--- a/sdk/storage/azure-storage-blob-cryptography/src/main/java/module-info.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/main/java/module-info.java
@@ -4,7 +4,6 @@
 module com.azure.storage.blob.cryptography {
     requires transitive com.azure.core;
     requires transitive com.azure.storage.common;
-    requires com.azure.security.keyvault.keys;
     requires com.azure.storage.blob;
     requires com.fasterxml.jackson.dataformat.xml;
 
--- a/sdk/storage/azure-storage-blob-cryptography/src/samples/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilderJavaDocCodeSnippets.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/samples/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilderJavaDocCodeSnippets.java
@@ -5,7 +5,6 @@ package com.azure.storage.blob.specialized.cryptography;
 
 import com.azure.core.cryptography.AsyncKeyEncryptionKey;
 import com.azure.core.cryptography.AsyncKeyEncryptionKeyResolver;
-import com.azure.security.keyvault.keys.cryptography.models.KeyWrapAlgorithm;
 
 /**
  *
@@ -15,7 +14,7 @@ public class EncryptedBlobClientBuilderJavaDocCodeSnippets {
 
     private AsyncKeyEncryptionKey key = JavaDocCodeSnippetsHelpers.getKey();
     private AsyncKeyEncryptionKeyResolver keyResolver = JavaDocCodeSnippetsHelpers.getKeyResolver();
-    private KeyWrapAlgorithm keyWrapAlgorithm = KeyWrapAlgorithm.RSA_OAEP;
+    private String keyWrapAlgorithm = "keyWrapAlgorithm";
 
     /**
      * Code snippet for {@link EncryptedBlobClientBuilder#buildEncryptedBlobAsyncClient()}
--- a/sdk/storage/azure-storage-blob-cryptography/src/test/java/com/azure/storage/blob/specialized/cryptography/APISpec.groovy
+++ b/sdk/storage/azure-storage-blob-cryptography/src/test/java/com/azure/storage/blob/specialized/cryptography/APISpec.groovy
@@ -16,9 +16,6 @@ import com.azure.core.test.TestMode
 import com.azure.core.test.utils.TestResourceNamer
 import com.azure.core.util.Configuration
 import com.azure.core.util.logging.ClientLogger
-import com.azure.security.keyvault.keys.cryptography.KeyEncryptionKeyClientBuilder
-import com.azure.security.keyvault.keys.cryptography.models.KeyWrapAlgorithm
-import com.azure.security.keyvault.keys.models.webkey.JsonWebKey
 import com.azure.storage.blob.BlobAsyncClient
 import com.azure.storage.blob.BlobClient
 import com.azure.storage.blob.BlobProperties
@@ -218,7 +215,7 @@ class APISpec extends Specification {
                                                          SharedKeyCredential credential, String endpoint,
                                                          HttpPipelinePolicy... policies) {
         EncryptedBlobClientBuilder builder = new EncryptedBlobClientBuilder()
-            .key(key, KeyWrapAlgorithm.RSA_OAEP_256)
+            .key(key, "keyWrapAlgorithm")
             .keyResolver(keyResolver)
             .endpoint(endpoint)
             .httpClient(getHttpClient())
@@ -262,14 +259,6 @@ class APISpec extends Specification {
         }
 
         return builder
-    }
-
-    KeyEncryptionKeyClientBuilder getKeyClientBuilder(String keyId, SecretKey secretKey) {
-        return new KeyEncryptionKeyClientBuilder()
-        .jsonWebKey(JsonWebKey.fromAes(secretKey))
-        .keyIdentifier(keyId)
-        .credential(new BasicAuthenticationCredential("fakeuname", "fakepwd"))
-
     }
 
     def generateContainerName() {
--- a/sdk/storage/azure-storage-blob-cryptography/src/test/java/com/azure/storage/blob/specialized/cryptography/EncryptedFlux.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/test/java/com/azure/storage/blob/specialized/cryptography/EncryptedFlux.java
@@ -4,7 +4,6 @@
 package com.azure.storage.blob.specialized.cryptography;
 
 import com.azure.core.cryptography.AsyncKeyEncryptionKey;
-import com.azure.security.keyvault.keys.cryptography.models.KeyWrapAlgorithm;
 import org.reactivestreams.Subscriber;
 import reactor.core.CoreSubscriber;
 import reactor.core.publisher.Flux;
@@ -141,7 +140,7 @@ public class EncryptedFlux extends Flux<ByteBuffer> {
         this.plainText = spec.getRandomData(DOWNLOAD_SIZE - 2); // This will yield two bytes of padding... for fun.
 
         EncryptedBlob encryptedBlob = new EncryptedBlobAsyncClient(null, null, null, key,
-            KeyWrapAlgorithm.RSA_OAEP)
+            "keyWrapAlgorithm")
             .encryptBlob(Flux.just(this.plainText)).block();
         this.cipherText = APISpec.collectBytesInBuffer(encryptedBlob.getCiphertextFlux()).block();
         this.encryptionData = encryptedBlob.getEncryptionData();
--- a/sdk/storage/azure-storage-blob-cryptography/src/test/java/com/azure/storage/blob/specialized/cryptography/EncyptedBlockBlobAPITest.groovy
+++ b/sdk/storage/azure-storage-blob-cryptography/src/test/java/com/azure/storage/blob/specialized/cryptography/EncyptedBlockBlobAPITest.groovy
@@ -2,8 +2,6 @@ package com.azure.storage.blob.specialized.cryptography
 
 import com.azure.core.cryptography.AsyncKeyEncryptionKey
 import com.azure.core.cryptography.AsyncKeyEncryptionKeyResolver
-import com.azure.security.keyvault.keys.cryptography.models.KeyWrapAlgorithm
-import com.azure.storage.blob.BlobContainerAsyncClient
 import com.azure.storage.blob.BlobContainerClient
 
 import com.azure.storage.blob.models.BlobAccessConditions


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+    /**
+     * Gets and validates {@link EncryptionData} from a Blob's metadata
+     *
+     * @param metadata {@code Map} of String -> String
+     *
+     * @return {@link EncryptionData}
+     */
+    private EncryptionData getAndValidateEncryptionData(Map<String, String> metadata) {
```
gapra-msft: TODO: Pass through the encryptedDataString instead of Metadata

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,419 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import static com.azure.core.implementation.util.FluxUtil.withContext;
+import static com.azure.storage.blob.PostProcessor.postProcessResponse;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class EncryptedBlockBlobAsyncClient extends BlobAsyncClient {
```
gapra-msft: TODO: Discuss how to deal with duplicate code here like commitBlock and stageBlock calls.
Prior conv:
gapra
Opinions on having encryptedblockblobasyncclient instead just wrap a blockblobasyncclient (that way the commitblock and stageblock will not be repeated in this class)

 alzimmermsft
Wrapping BlockBlobAsyncClient should work if the method internals are the same outside of this client having an additional policy.

 gapra-msft
It's not just an additional policy, it also encrypts the data before sending it on it's way through the pipeline

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,489 @@
...
+            throw logger.logExceptionAsError(new IllegalArgumentException("Key and KeyResolver cannot both be null"));
+        }
+        this.keyWrapper = key;
+        this.keyResolver = keyResolver;
+    }
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        // 1. Expand the range of download for decryption
+        String rangeHeader = "x-ms-range";
```
gapra-msft: Should this be moved to global variable even though its the only place it is used? Another option is to put it in the common constants.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,489 @@
...
+                of downloaded data are in position to be used as the IV for the data actually requested and we are
+                in the desired state.
+                 */
+                byte[] iv = new byte[ENCRYPTION_BLOCK_SIZE];
+                /*
+                Adjusting the range by <= 16 means we only adjusted to align on an encryption block boundary
+                (padding will add 1-16 bytes as it will prefer to pad 16 bytes instead of 0) and therefore the key
+                is in the metadata.
+                 */
+                if (encryptedBlobRange.offsetAdjustment() <= ENCRYPTION_BLOCK_SIZE) {
```
gapra-msft: This could be made into an if/else where the else creates the empty byte array?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobClient.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Package-private constructor for use by {@link BlobClientBuilder}.
+     */
+    EncryptedBlockBlobClient(EncryptedBlockBlobAsyncClient encryptedBlockBlobAsyncClient) {
+        super(encryptedBlockBlobAsyncClient);
+        this.encryptedBlockBlobAsyncClient = encryptedBlockBlobAsyncClient;
+    }
+
+    // TODO (gapra) : Test this
```
gapra-msft: Ned to remove this comment

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,416 @@
...
+    /**
+     * Package-private constructor for use by {@link EncryptedBlobClientBuilder}.
+     */
+    EncryptedBlockBlobAsyncClient(AzureBlobStorageImpl constructImpl, String snapshot,
+        BlobEncryptionPolicy encryptionPolicy) {
+        super(constructImpl, snapshot, null);
+        this.encryptionPolicy = encryptionPolicy;
+    }
+
+    // TODO (gapra) : Test this
```
gapra-msft: Need to remove this comment

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
rickle-msft: I think putting it in constants is a good idea

---
In reply to: [328297902](https://github.com/Azure/azure-sdk-for-java/pull/5528#discussion_r328297902) [](ancestors = 328297902)

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -0,0 +1,265 @@
...
+import java.util.Objects;
+
+/**
+ * This class provides a fluent builder API to help aid the configuration and instantiation Storage Blob clients.
+ *
+ * <p>
+ * The following information must be provided on this builder:
+ *
+ * <ul>
+ * <li>the endpoint through {@code .endpoint()}, including the container name and blob name, in the format of
```
rickle-msft: >including the container name and blob name [](start = 49, length = 42)

"if the container and blob name are not specified, they must be specified through {@code .blobName} and {@code .containerName}."

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -0,0 +1,265 @@
...
+ * {@code https://{accountName}.blob.core.windows.net/{containerName}/{blobName}}.
+ * <li>the credential through {@code .credential()} or {@code .connectionString()} if the container is not publicly
+ * accessible.
```
rickle-msft: Also the key must be provided

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -0,0 +1,265 @@
...
+    public EncryptedBlockBlobClient buildEncryptedBlockBlobClient(BlockBlobClient client) {
+        Objects.requireNonNull(encryptionPolicy);
+        Objects.requireNonNull(decryptionPolicy);
+        AzureBlobStorageImpl impl = new AzureBlobStorageBuilder()
+            .url(client.getBlobUrl().toString())
+            .pipeline(addDecryptionPolicy(client.getHttpPipeline(),
+                client.getHttpPipeline().getHttpClient()))
+            .build();
+        return new EncryptedBlockBlobClient(
+            new EncryptedBlockBlobAsyncClient(impl, client.getSnapshotId(), encryptionPolicy));
```
rickle-msft: It's unfortunate that this can't call into buildAsync. I tend to refactor some stuff like this so 114-120 are in a common place in case we need to change our configurations, but I'll leave it to you if you think that's overkill.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    /**
+     * Creates a builder instance that is able to configure and construct Storage Blob clients.
+     *
+     * @param key An object of type {@link IKey} that is used to wrap/unwrap the content encryption key.
+     * @param keyResolver The key resolver used to select the correct key for decrypting existing blobs.
+     * @return the updated EncryptedBlobClientBuilder object
+     */
+    public EncryptedBlobClientBuilder keyAndKeyResolver(IKey key, IKeyResolver keyResolver) {
```
rickle-msft: >keyAndKeyResolver [](start = 38, length = 17)

Why are these passed together?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobEncryptionPolicy.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Represents a blob encryption policy that is used to perform envelope encryption of Azure storage blobs.
+ * <a href="https://docs.microsoft.com/en-us/azure/storage/common/storage-client-side-encryption-java?toc=%2fazure%2fstorage%2fblobs%2ftoc.json">Azure Docs</a>
+ * for more information.
+ */
+final class BlobEncryptionPolicy {
```
rickle-msft: I wonder if we should rename this now that "policy" has become overloaded. Perhaps just BlobEncryptor? Or even just moving the methods on here to EncryptedBlockBlobClient might work. Since this is an internal type, and the methods are only used within EncryptedBlockBlobClient, and this is never really passed around, perhaps a separate type is not even necessary. Probably the reason it's been contained in this type for v11 and track 2 is just because that's what we did in track 1.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,416 @@
...
+     * @param numBuffers The maximum number of buffers this method should allocate. Must be at least two. Typically, the
+     * larger the number of buffers, the more parallel, and thus faster, the upload portion of this operation will be.
+     * The amount of memory consumed by this method may be up to blockSize * numBuffers.
+     * @param headers {@link BlobHTTPHeaders}
+     * @param metadata {@link Metadata}
+     * @param tier {@link AccessTier} for the destination blob.
+     * @param accessConditions {@link BlobAccessConditions}
+     * @return A reactive response containing the information of the uploaded block blob.
+     */
+    public Mono<Response<BlockBlobItem>> uploadWithResponse(Flux<ByteBuffer> data, int blockSize, int numBuffers,
```
rickle-msft: >uploadWithResponse [](start = 41, length = 18)

I think we should be able to get rid of this and just encrypt the data then pass through to the regular async client's version. That would also let us get rid of the internal version of stageBlock and commitBlockList

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,416 @@
...
+        Mono<Flux<ByteBuffer>> dataFinal = encryptionPolicy.prepareToSendEncryptedRequest(data, metadataFinal);
+        // Readjust the length to account for padding.
+        long lengthFinal =  length + (16 - length % 16);
+
+        return dataFinal.flatMap(df ->
+            postProcessResponse(this.azureBlobStorage.blockBlobs().uploadWithRestResponseAsync(null, null, df,
+                lengthFinal, null, metadataFinal, tier, null, headers,
+                accessConditionsFinal.getLeaseAccessConditions(), null /*cpk*/,
+                accessConditionsFinal.getModifiedAccessConditions(), context))
+            .map(rb -> new SimpleResponse<>(rb, new BlockBlobItem(rb.getDeserializedHeaders()))));
```
rickle-msft: Ideally this would pass through to the unencrypted client, but I seem to recall some conversation about why that wasn't possible for this api in particular. Can you leave a comment as to what that reason is/was

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,489 @@
...
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        // 1. Expand the range of download for decryption
+        String rangeHeader = "x-ms-range";
+        HttpHeaders requestHeaders = context.getHttpRequest().getHeaders();
+        EncryptedBlobRange encryptedRange = EncryptedBlobRange.getEncryptedBlobRangeFromHeader(
+            requestHeaders.getValue(rangeHeader));
+
+        // Assumption: Download is the only API on an encrypted client that sets x-ms-range
+        // Only set the x-ms-range header if it already exists
```
rickle-msft: Another option here is checking the context. The client would have to insert a "decrypt: true" key/value into the context and then this policy could check for that.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,489 @@
...
+            // response
+            if (httpResponse.getRequest().getHttpMethod() == HttpMethod.GET && httpResponse.getBody() != null) {
+                HttpHeaders responseHeaders = httpResponse.getHeaders();
+                /*
+                We will need to know the total size of the data to know when to finalize the decryption. If it was
+                not set originally with the intent of downloading the whole blob, update it here.
+                 */
+                encryptedRange.withAdjustedDownloadCount(Long.parseLong(responseHeaders.getValue("Content-Length")));
+                boolean padding = encryptedRange.toBlobRange().getOffset()
+                    + encryptedRange.toBlobRange().getCount() > (blobSize(responseHeaders) - ENCRYPTION_BLOCK_SIZE);
```
rickle-msft: Can you comment on this line? I think it's that we expect padding only if we are at the end of the blob

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobDecryptionPolicy.java
rickle-msft: It's true. Up to you which you prefer. 

---
In reply to: [328298432](https://github.com/Azure/azure-sdk-for-java/pull/5528#discussion_r328298432) [](ancestors = 328298432)

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/EncryptedBlobClientBuilder.java
gapra-msft: Changed to pass separately

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/cryptography/BlobEncryptionPolicy.java
gapra-msft: Moving to encryptedblockblobasyncclient

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -0,0 +1,313 @@
...
+/**
+ * This class provides a fluent builder API to help aid the configuration and instantiation of Storage Blob clients.
+ *
+ * <p>
+ * The following information must be provided on this builder:
+ *
+ * <ul>
+ * <li>the endpoint through {@code .endpoint()}, including the container name and blob name, in the format of
+ * {@code https://{accountName}.blob.core.windows.net/{containerName}/{blobName}}.
+ * <li>if the container and blob name are not specified, they must be specified through {@code .blobName90} and
```
gapra-msft: Typo with blobName()

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/BaseClientBuilder.java
```diff
@@ -109,7 +110,7 @@ protected HttpPipeline buildPipeline() {
      *
      * @return The validation policy.
      */
-    private HttpPipelinePolicy makeValidationPolicy() {
+    protected HttpPipelinePolicy makeValidationPolicy() {
```
gapra-msft: will revert this back since this does not need to be protected

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/UploadBufferPool.java
```diff
@@ -28,8 +28,10 @@
  * to be available, and either store the incomplete buffer to be filled on the next write or return the filled buffer to
  * be sent. Filled buffers can be uploaded in parallel and should return buffers to the pool after the upload completes.
  * Once the source terminates, it should call flush.
+ *
+ * RESERVED FOR INTERNAL USE ONLY
```
gapra-msft: Will revert changes to this as well since now the encrypted client calls into blob client methods

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/BaseClientBuilder.java
```diff
@@ -118,6 +119,13 @@ private HttpPipelinePolicy makeValidationPolicy() {
         return builder.build();
     }
 
+    /**
+     * Adds an optional encryption policy that decrypts encrypted blobs.
+     * @param policies The list of policies to add an optional encryption policy to.
+     */
+    protected void addOptionalEncryptionPolicy(List<HttpPipelinePolicy> policies) {
```
alzimmermsft: I think this should be added into each encryption client builder instead of the BaseClientBuilder as this should only be applicable to two builders.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobRange.java
```diff
@@ -17,6 +17,15 @@
     private final long offset;
     private final Long count;
 
+    /**
+     * Specifies the download operation to start from the offset position (zero-based) and download the
+     * rest of the entire blob to the end. An empty BlobRange effectively specifies the whole blob.
+     */
+    public BlobRange() {
```
alzimmermsft: Was this added to simplify some code? I'm not completely against it but given the immutability of this class it is easily replicated by `new BlobRange(0)`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -263,10 +266,9 @@ as we can guarantee we only need at most two buffers for any call to write (two
                 return getBlockBlobAsyncClient().stageBlockWithResponse(blockId, Flux.just(buffer), buffer.remaining(),
                     accessConditionsFinal.getLeaseAccessConditions())
                     // We only care about the stageBlock insofar as it was successful, but we need to collect the ids.
-                    .map(x -> {
-                        pool.returnBuffer(buffer);
-                        return blockId;
-                    }).flux();
+                    .map(x -> blockId)
```
alzimmermsft: Love this change!

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptionAgent.java
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
gapra-msft: TODO: refactor names in this method to have get set prefixes

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
alzimmermsft:
```suggestion
 * <li>Container and blob name if not specified in the {@code endpoint}, set through {@link #containerName(String)} and {@link #blobName(String)} respectively.
```

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -0,0 +1,313 @@
...
+ * {@code .containerName()}.
+ * <li>the credential through {@code .credential()} or {@code .connectionString()} if the container is not publicly
+ * accessible.
+ * <li>the key and key wrapping algorithm (for encryption) and/or key resolver (for decryption) must be specified
+ * through {@code .key()} and {@code .keyResolver()}
```
alzimmermsft:
```suggestion
 * through {@link #key()} and {@link #keyResolver()}
```

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/WrappedKey.java
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
gapra-msft: Will refactor this class to use get set prefixes

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -0,0 +1,313 @@
...
+ *
+ * <p>
+ * The following information must be provided on this builder:
+ *
+ * <ul>
+ * <li>the endpoint through {@code .endpoint()}, including the container name and blob name, in the format of
+ * {@code https://{accountName}.blob.core.windows.net/{containerName}/{blobName}}.
+ * <li>if the container and blob name are not specified, they must be specified through {@code .blobName90} and
+ * {@code .containerName()}.
+ * <li>the credential through {@code .credential()} or {@code .connectionString()} if the container is not publicly
```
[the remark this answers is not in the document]
gapra-msft: This could imply the container the blob is in?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobRange.java
gapra-msft: You're right. Just removed it

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/BaseClientBuilder.java
gapra-msft: we call this in our base build pipeline method so the decryption policy can be added in the beginning of the pipeline. If I just added it in the two clients that need it then I would need to rewrite a lot of the base builder

sdk/storage/azure-storage-blob-cryptography/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+```xml
+<dependency>
+  <groupId>com.azure</groupId>
+  <artifactId>azure-storage-blob</artifactId>
+  <version>12.0.0-preview.3</version>
+</dependency>
+```
+
+### Default HTTP Client
```
srnagar: We should have Netty as the default Http client. See https://github.com/Azure/azure-sdk-for-java/issues/5511

This section should be updated accordingly. Refer README for blobs.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,485 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static com.azure.storage.blob.specialized.cryptography.CryptographyConstants.ENCRYPTION_BLOCK_SIZE;
+
+public class BlobDecryptionPolicy implements HttpPipelinePolicy {
```
srnagar: Add Javadoc.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,485 @@
...
+     *
+     * @param encryptedDataString {@code String} of encrypted metadata
+     *
+     * @return {@link EncryptionData}
+     */
+    private EncryptionData getAndValidateEncryptionData(String encryptedDataString) {
+        if (encryptedDataString == null) {
+            throw logger.logExceptionAsError(new IllegalStateException(CryptographyConstants.DECRYPT_UNENCRYPTED_BLOB));
+        }
+        ObjectMapper objectMapper = new ObjectMapper();
```
srnagar: You can create a static instance of this `ObjectMapper` to avoid creating lots of these objects for every `decryptBlob` call.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,485 @@
...
+        try {
+            EncryptionData encryptionData = objectMapper.readValue(encryptedDataString, EncryptionData.class);
+
+            if (encryptionData == null) {
+                throw logger.logExceptionAsError(new IllegalStateException(
+                    CryptographyConstants.DECRYPT_UNENCRYPTED_BLOB));
+            }
+
+            Objects.requireNonNull(encryptionData.contentEncryptionIV());
```
srnagar: Add an error message to indicate what was null everywhere you are using `Objects.requireNonNull`

```suggestion
            Objects.requireNonNull(encryptionData.contentEncryptionIV(), "contentEncryptionIV in encryptionData cannot be null");
```

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,485 @@
...
+
+    /**
+     * Gets and validates {@link EncryptionData} from a Blob's metadata
+     *
+     * @param encryptedDataString {@code String} of encrypted metadata
+     *
+     * @return {@link EncryptionData}
+     */
+    private EncryptionData getAndValidateEncryptionData(String encryptedDataString) {
+        if (encryptedDataString == null) {
```
srnagar: This should throw NPE instead - https://azure.github.io/azure-sdk/java_implementation.html#java-errors-system-errors

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -0,0 +1,325 @@
...
+    private String snapshot;
+
+    private AsyncKeyEncryptionKey keyWrapper;
+    private AsyncKeyEncryptionKeyResolver keyResolver;
+    private KeyWrapAlgorithm keyWrapAlgorithm;
+
+    /**
+     * Creates a new instance of the EncryptedBlobClientBuilder
+     */
+    public EncryptedBlobClientBuilder() {
```
srnagar: This is redundant. You can remove this constructor.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    private AzureBlobStorageImpl constructImpl() {
+        Objects.requireNonNull(blobName, "'blobName' cannot be null.");
```
srnagar: Don't need this check as the `blobName()` builder method already checks for null.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,363 @@
...
+import java.nio.file.StandardOpenOption;
+import java.security.GeneralSecurityException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+public class EncryptedBlockBlobAsyncClient extends BlobAsyncClient {
```
srnagar: Add javadoc and all the applicable annotations for this client.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlockBlobClient.java
```diff
@@ -0,0 +1,85 @@
...
+import com.azure.storage.blob.models.ParallelTransferOptions;
+import com.azure.storage.blob.specialized.BlockBlobClient;
+import com.azure.storage.common.Utility;
+import reactor.core.publisher.Mono;
+
+import java.io.UncheckedIOException;
+import java.time.Duration;
+import java.util.Map;
+
+public class EncryptedBlockBlobClient extends BlobClient {
```
srnagar: Add javadoc and annotations to the class and methods.

sdk/storage/pom.service.xml
```diff
@@ -14,6 +14,8 @@
     <module>../core/azure-core-test</module>
     <module>../core/azure-core-http-netty</module>
     <module>../identity/azure-identity</module>
+    <module>../keyvault/microsoft-azure-keyvault-core</module>
```
srnagar: Should we be taking a dependency on track 1 Key Vault?

@JonathanGiles

sdk/storage/azure-storage-blob-cryptography/pom.xml
```diff
@@ -42,30 +42,49 @@
...
-<!--      <scope>test</scope>-->
-<!--    </dependency>-->
-<!--    &lt;!&ndash; https://mvnrepository.com/artifact/org.spockframework/spock-core &ndash;&gt;-->
-<!--    <dependency>-->
-<!--      <groupId>org.spockframework</groupId>-->
-<!--      <artifactId>spock-core</artifactId>-->
-<!--      <scope>test</scope>-->
-<!--    </dependency>-->
-<!--    <dependency>-->
+    <dependency>
```
srnagar: Since this is a new module, please add `module-info.java` as well.

sdk/storage/pom.service.xml
gapra-msft: This dependency is purely for testing purposes so we can make sure track 1 and track 2 SDKs encryption is cross compatible.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlockBlobAsyncClient.java
gapra-msft: I added Javadoc, but what annotations need to be added to clients? I dont see any on BlobClient from when I pulled from master this morning.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
gapra-msft: blob name is also optionally set in endpoint so it could have been null there.

sdk/storage/pom.service.xml
srnagar: This is included in pom.service.xml as one of the sources to build. So, our pipelines will build track 1 and track 2 libraries.

gapra-msft: just removed from pom.service.xml

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/CryptographyConstants.java
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.storage.blob.specialized.cryptography;
+
+import com.azure.storage.blob.BlobConfiguration;
+
+public final class CryptographyConstants {
```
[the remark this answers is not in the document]
gapra-msft: Making this package private

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
gapra-msft: I was just following the pattern of BlobClientBuilder which also has the redundant constructor.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BaseBlobClientBuilder.java
```diff
@@ -46,7 +47,7 @@ public T customerProvidedKey(CustomerProvidedKey key) {
      * @return the {@code UserAgentPolicy} that will be used in the {@link HttpPipeline}.
      */
     @Override
-    protected final UserAgentPolicy getUserAgentPolicy() {
+    protected UserAgentPolicy getUserAgentPolicy() {
```
gapra-msft: This cant be final since we want blob-cryptography's user agent policy to say azure-storage-blob-cryptography instead of azure-storage-blob

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptionData.java
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
gapra-msft: Change this class to get set prefixes

sdk/storage/azure-storage-blob-cryptography/pom.xml
gapra-msft: @srnagar Once this PR gets merged and approved, could you add the module-info.java file?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlockBlobAsyncClient.java
gapra-msft: Added service client annotations

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,510 @@
...
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static com.azure.storage.blob.specialized.cryptography.CryptographyConstants.ENCRYPTION_BLOCK_SIZE;
+
+/**
+ * This is a decryption policy in an {@link com.azure.core.http.HttpPipeline} to decrypt data in an
+ * {@link EncryptedBlockBlobAsyncClient} or {@link EncryptedBlockBlobClient} download request. The range will be
+ * expanded for decryption purposes and then decrypt the body when the response comes in. This policy must only be used
+ * on encrypted blob clients.
```
rickle-msft: No longer true given our recent changes.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
```diff
@@ -0,0 +1,510 @@
...
+                of downloaded data are in position to be used as the IV for the data actually requested and we are
+                in the desired state.
+                 */
+                    byte[] iv;
+
+                /*
+                Adjusting the range by <= 16 means we only adjusted to align on an encryption block boundary
+                (padding will add 1-16 bytes as it will prefer to pad 16 bytes instead of 0) and therefore the key
+                is in the metadata.
+                 */
```
rickle-msft: Formatting 🙈

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
rickle-msft: This is actually by design. The guidelines state throw a NPE when null is unexpected, but null is a valid value here, it just means that there's nothing to parse. We use that in decrypt blob to indicate skipping the decryption step and go right to trimming. 

---
In reply to: [332684738](https://github.com/Azure/azure-sdk-for-java/pull/5528#discussion_r332684738) [](ancestors = 332684738)

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,346 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * This class provides a client side encryption client that contains generic blob operations for Azure Storage Blobs.
+ * Operations allowed by the client are uploading, downloading and copying a blob, retrieving and setting metadata,
+ * retrieving and setting HTTP headers, and deleting and un-deleting a blob. The upload and download operation allow for
+ * encryption and decryption of the data client side.
```
rickle-msft: > side. [](start = 47, length = 6)

Can we add a call out that setting metadata in particular is unsafe and should only be done with caution?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,346 @@
...
+     * Creates a new block blob, or updates the content of an existing block blob, with the content of the specified
+     * file.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.specialized.cryptography.EncryptedBlockBlobAsyncClient.uploadFromFile#String-ParallelTransferOptions-BlobHTTPHeaders-Map-AccessTier-BlobAccessConditions}
+     *
+     * @param filePath Path to the upload file
+     * @param parallelTransferOptions {@link ParallelTransferOptions} to use to upload from file. Number of parallel
+     *        transfers parameter is ignored.
```
rickle-msft: I don't think this is true in the encrypted case since we pass it to buffered upload.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlockBlobAsyncClient.java
```diff
@@ -0,0 +1,346 @@
...
+     * @param headers {@link BlobHTTPHeaders}
+     * @param metadata Metadata to associate with the blob.
+     * @param tier {@link AccessTier} for the destination blob.
+     * @param accessConditions {@link BlobAccessConditions}
+     * @return An empty response
+     * @throws IllegalArgumentException If {@code blockSize} is less than 0 or greater than 100MB
+     * @throws UncheckedIOException If an I/O error occurs
+     */
+    // TODO (gapra) : Investigate if this is can be parallelized, and include the parallelTransfers parameter.
```
rickle-msft: >/ TODO (gapra) : Investigate if this is can be parallelized, and include the parallelTransfers parameter. [](start = 5, length = 105)

Remove

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlockBlobAsyncClient.java
gapra-msft: Good catch

sdk/storage/azure-storage-blob-cryptography/pom.xml
gapra-msft: Added module file

sdk/storage/azure-storage-blob-cryptography/README.md
```diff
@@ -0,0 +1,154 @@
...
+All client libraries, by default, use Netty HTTP client. Adding the above dependency will automatically configure 
+Storage Blob to use Netty HTTP client. 
+
+### Alternate HTTP client
+If, instead of Netty it is preferable to use OkHTTP, there is a HTTP client available for that too. Exclude the default
+Netty and include OkHTTP client in your pom.xml.
+
+```xml
+<!-- Add Storage Blob dependency without Netty HTTP client -->
```
srnagar: Update the comments and artifact name.

sdk/storage/azure-storage-blob-cryptography/README.md
```diff
@@ -0,0 +1,154 @@
...
+## Next steps
+
+## Contributing
+This project welcomes contributions and suggestions. Most contributions require you to agree to a Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us the rights to use your contribution. For details, visit https://cla.microsoft.com.
+
+When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repos using our CLA.
+
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the Code of Conduct FAQ or contact opencode@microsoft.com with any additional questions or comments.
+
+![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-java/sdk/tracing/README.png)
```
srnagar: This should use storage blob cryptography instead of tracing.

sdk/storage/azure-storage-blob-cryptography/README.md
```diff
@@ -0,0 +1,154 @@
...
+Blob storage is designed for:
+
+- Serving images or documents directly to a browser.
+- Storing files for distributed access.
+- Streaming video and audio.
+- Writing to log files.
+- Storing data for backup and restore, disaster recovery, and archiving.
+- Storing data for analysis by an on-premises or Azure-hosted service.
+
+## Examples
```
srnagar: Add some examples on how to use this module.

PR #5529, opened 2019-09-25
use pom.data.xml instead of pom.client.xml to invoke jetty for data service pipeline builds

JimSuplizio:
The **Start Jetty** task is only being run when the **SdkType** is **data** however the mavenPomFile was using the pom.client.xml which was causing needless compilation of track 2 sources during a track 1 build.


JimSuplizio:
This item requires more investigation and a separate issue has been created for that. I'm closing this PR out and dropping these changes.

PR #5532, opened 2019-09-25
Rename the sas permission

sima-zhu:


gapra-msft:
Just looks like there are some tests that are failing. Make sure to fix those and should be fine to merge.

PR #5533, opened 2019-09-25
Renaming 'fields' to 'selects'

mssfang:
Rename all 'fields' to selects
fix ComplexConfiguration's wrong package path.

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/samples/java/com/azure/data/appconfiguration/ConfigurationSets.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/samples/java/com/azure/data/appconfiguration/ConfigurationSets.java
@@ -4,9 +4,9 @@
 package com.azure.data.appconfiguration;
 
 import com.azure.data.appconfiguration.credentials.ConfigurationClientCredentials;
+import com.azure.data.appconfiguration.models.ComplexConfiguration;
 import com.azure.data.appconfiguration.models.ConfigurationSetting;
 import com.azure.data.appconfiguration.models.SettingSelector;
-import com.azure.models.ComplexConfiguration;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import reactor.core.publisher.Flux;


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelector.java
```diff
@@ -202,18 +202,18 @@ public SettingSelector setRange(Range range) {
 
     @Override
     public String toString() {
-        String fields;
-        if (ImplUtils.isNullOrEmpty(this.fields)) {
-            fields = "ALL_FIELDS";
+        String selects;
+        if (ImplUtils.isNullOrEmpty(this.selects)) {
+            selects = "SELECT_ALL";
```
alzimmermsft: @conniey Do you remember if `ALL_FIELDS` triggered the service to return something special? Looking at their documentation there is no mention of this and we likely should pass nothing in this case to get the default selection.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelects.java
```diff
@@ -12,7 +12,7 @@
  * @see SettingSelector
  * @see ConfigurationAsyncClient
  */
-public enum SettingFields {
+public enum SettingSelects {
```
alzimmermsft: Maybe name this to `FieldSelections`? We should bring this naming up in the stand-up

alzimmermsft: Let's change this from being an `enum` to `ExpandableStringEnum` to help make this version independent.

conniey: There's a teams discussion in the azconfig channel about this. I think the name should be agreed on by all language folks before updating.

mssfang:
As AzConfig Channel discussed. All languages will stick with "fields". So no change needed.

PR #5537, opened 2019-09-25
Change last_modified and locked JsonProperty READ_ONLY

mssfang:
Change last_modified and locked JsonProperty READ_ONLY


alzimmermsft:
Why are we changing the JSON serialization to readonly? We only needed to remove the setter method for these properties and that looks to be done already.

mssfang:
Ahh. Misunderstand it. Closing this PR

PR #5538, opened 2019-09-25
Core cryptography interfaces

g2vinay:
Introduces Crypto interfaces in core.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/cryptography/AsyncKeyEncryptionKeyResolver.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/cryptography/AsyncKeyEncryptionKeyResolver.java
@@ -14,7 +14,7 @@ public interface AsyncKeyEncryptionKeyResolver {
      * Retrieves the {@link AsyncKeyEncryptionKey} corresponding to the specified {@code keyId}
      *
      * @param keyId The key identifier of the key encryption key to retrieve
-     * @return The encryption key corresponding to the specified {@code keyId}
+     * @return The key encryption key corresponding to the specified {@code keyId}
      */
     Mono<AsyncKeyEncryptionKey> resolveKey(String keyId);
 }
--- a/sdk/core/azure-core/src/main/java/com/azure/core/cryptography/KeyEncryptionKeyResolver.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/cryptography/KeyEncryptionKeyResolver.java
@@ -12,7 +12,7 @@ public interface KeyEncryptionKeyResolver {
      * Retrieves the {@link KeyEncryptionKey} corresponding to the specified {@code keyId}
      *
      * @param keyId The key identifier of the key encryption key to retrieve
-     * @return The encryption key corresponding to the specified {@code keyId}
+     * @return The key encryption key corresponding to the specified {@code keyId}
      */
     KeyEncryptionKey resolveKey(String keyId);
 }


sdk/core/azure-core/src/main/java/com/azure/core/cryptography/IKeyEncryptionKey.java
```diff
@@ -0,0 +1,31 @@
...
+
+import reactor.core.publisher.Mono;
+
+
+/**
+ * A key which is used to encrypt, or wrap, another key.
+ */
+public interface IKeyEncryptionKey {
+    
+    /**
```
gapra-msft: Could there be a getKeyId() method? This was also a change that was made in .NET and is required for a preview 4 storage feature

sdk/core/azure-core/src/main/java/com/azure/core/cryptography/IKeyEncryptionKey.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.azure.core.cryptography;
+
+import reactor.core.publisher.Mono;
+
+
+/**
+ * A key which is used to encrypt, or wrap, another key.
+ */
+public interface IKeyEncryptionKey {
```
JonathanGiles: Should this just be called `EncryptionKey`, and the interface below be called `EncryptionKeyResolver`?

g2vinay: Updated, Introduced Sync and Async variants,

sdk/core/azure-core/src/main/java/com/azure/core/cryptography/IKeyEncryptionKey.java
g2vinay: Added it

sdk/core/azure-core/src/main/java/com/azure/core/cryptography/IKeyEncryptionKey.java
schaabs: Actually while it might seem verbose this should be `KeyEncryptionKey` and `KeyEncryptionKeyResolver`. The methods that `KeyEncryptionKey` exposes, `wrapKey` and `unwrapKey` are specific to encrypting a ___key___, not arbitrary data. An `EncryptionKey` interface would have methods `encrypt` and `decrypt`. 

Originally in track 1 there was the `IKey` interface which had methods,  `encrypt`, `decrypt`, `wrapKey`, `unwrapKey`, `sign`, and `verify`.  Here we have scoped an interface to only provide the methods needed by storage for their client side encryption implementation. But we want to name it appropriately so that we could reintroduce the full interface if necessary. The full set of interfaces would look something like this:

- KeyEncryptionKey 
  - wrapKey
  - unwrapKey
- EncryptionKey
  - encrypt
  - decrypt
- SignatureKey
  - sign
  - verify

Also you could possibly tie these all together in an interface to have an interface that was equivalent to track 1's `IKey`, although you'd need to rename it to fit the java guidelines, so something like `CryptographicKey` possibly.

sdk/core/azure-core/src/main/java/com/azure/core/cryptography/AsyncEncryptionKey.java
```diff
@@ -0,0 +1,39 @@
...
+ * A key which is used to asynchronously encrypt, or wrap, another key.
+ */
+public interface AsyncEncryptionKey {
+
+    /**
+     * Retrieves the key identifier.
+     *
+     * @return A {@link Mono} containing key identifier.
+     */
+    Mono<URI> getKeyId();
```
schaabs: For compatibility reasons this needs to be a string rather than a URI. @jaschrep-msft originally raised this issue and in the .NET SDK and is addressing it in this PR: https://github.com/Azure/azure-sdk-for-net/pull/7799

sdk/core/azure-core/src/main/java/com/azure/core/cryptography/IKeyEncryptionKey.java
JonathanGiles: Your explanation makes sense. @g2vinay be sure to include some of what Scott has stated here in the class-level javadocs to clarify to users what this API is setting out to achieve.

@schaabs Because I'm no expert on crypto, my initial instinct is to have it all in one interface. My reasoning being is that I'd rather be overwhelmed in one interface rather than be overwhelmed by multiple interfaces! :-) However, for this I'm deferring to you, I just wanted to share a naive opinion on crypto API intuitiveness.

sdk/core/azure-core/src/main/java/com/azure/core/cryptography/AsyncEncryptionKey.java
g2vinay: updated to String.

sdk/core/azure-core/src/main/java/com/azure/core/cryptography/IKeyEncryptionKey.java
g2vinay: updated names to KeyEncryptionKey and KeyEncryptionKeyResolver

PR #5539, opened 2019-09-25
Fix parameter names

conniey:
* Change ExpandableStringEnum.values to getValues to follow our Java spec
* Change parameter name from `getNew` to `tokenSupplier`.


JonathanGiles:
Hmm, there was a communication breakdown here. I spoke with @samvaity about this earlier and said to not do the values() rename - this is actually just recreating the standard values() method on enum types, so it made sense to leave it as is.

conniey:
@JonathanGiles https://github.com/Azure/azure-sdk-for-java/pull/5555 Its reverted here.

PR #5547, opened 2019-09-25
downloadToFile Max overload method rename

sima-zhu:
Based on the API review, we decided to have downloadToFile max overload API returns `Mono<Response<BlobProperties>>` and  `Mono<Response<FileProperties>> ` as async and 
`Response<BlobProperties>` and  `Response<FileProperties> ` as sync, 
which is also in sync with .NET API.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
@@ -450,10 +450,10 @@ public class BlobClientBase {
      * @param timeout An optional timeout value beyond which a {@link RuntimeException} will be raised.
      * @param context Additional context that is passed through the Http pipeline during the service call.
      * @return The response of download blob properties.
-     * @throws UncheckedIOException If an I/O error occurs
+     * @throws UncheckedIOException If an I/O error occurs.
      */
     public Response<BlobProperties> downloadToFileWithResponse(String filePath, BlobRange range, Integer blockSize,
-            ReliableDownloadOptions options,BlobAccessConditions accessConditions, boolean rangeGetContentMD5,
+            ReliableDownloadOptions options, BlobAccessConditions accessConditions, boolean rangeGetContentMD5,
             Duration timeout, Context context) {
         Mono<Response<BlobProperties>> download = client.downloadToFileWithResponse(filePath, range, blockSize,
             options, accessConditions, rangeGetContentMD5, context);


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -559,27 +560,39 @@ public HttpPipeline getHttpPipeline() {
...
-                    .flatMap(dar -> FluxUtil.writeFile(dar.body(options), channel,
-                        chunk.getOffset() - (range == null ? 0 : range.getOffset()))))
-                .then(), this::downloadToFileCleanup);
+            channel -> blobDownloadInChunk(channel, range, blockSize, options, accessConditions, rangeGetContentMD5,
+                context),
+            this::downloadToFileCleanup);
+
+    }
+
+    private Mono<Response<BlobProperties>> blobDownloadInChunk(final AsynchronousFileChannel channel, BlobRange range,
```
alzimmermsft: Did we need to add this private method?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -559,27 +560,39 @@ public HttpPipeline getHttpPipeline() {
...
+                              Integer blockSize, ReliableDownloadOptions options, BlobAccessConditions accessConditions,
+                              boolean rangeGetContentMD5, Context context) {
+        Mono<Response<BlobProperties>> asyncGetPropertiesResponse = getPropertiesWithResponse(accessConditions);
```
alzimmermsft: The way this is being used may trigger two `getProperties` requests, can we verify this isn't happening.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -559,27 +560,39 @@ public HttpPipeline getHttpPipeline() {
...
+        Mono<BlobRange> fullBlobRange = (range == null) ? getFullBlobRange(asyncGetPropertiesResponse)
+            : Mono.just(range);
+        return fullBlobRange.flatMapMany(rg -> Flux.fromIterable(sliceBlobRange(rg, blockSize)))
```
alzimmermsft: Could we look at what `downloadWithResponse` is doing to see if we could put this all into a single stream?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
```diff
@@ -408,16 +408,19 @@ public void download(OutputStream stream) {
...
      * @param filePath A non-null {@link OutputStream} instance where the downloaded data will be written.
+     * @return The properties of the download blob.
      * @throws UncheckedIOException If an I/O error occurs
      */
-    public void downloadToFile(String filePath) {
-        downloadToFile(filePath, null, null, null, null, false, null, Context.NONE);
+    public BlobProperties downloadToFile(String filePath) {
+        downloadToFileWithResponse(filePath, null, null, null, null,
+            false, null, Context.NONE);
+        return getProperties();
```
alzimmermsft: Why are we calling `getProperties` here? Just call `getValue` on the call above.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
sima-zhu: I am trying to make the public method a little clear. Cannot understand the logic with too many lambda involved in one statement.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
sima-zhu: I don't think it triggered twice. The blob range used the response directly.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
sima-zhu: Make changes to the method. My mainline merge messed them a little. Only call the maxOverload now.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
sima-zhu: downloadWithProperties return DownloadAsyncResponse. Do we want to align with that?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: Looking at this again the method doesn't get triggered to run until `.then` is called with it. `Monos` and `Fluxes` are deferred until a subscription is made.

https://projectreactor.io/docs/core/release/reference/#_operators

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,40 @@ public boolean isSnapshot() {
      * @throws IllegalArgumentException If {@code blockSize} is less than 0 or greater than 100MB.
      * @throws UncheckedIOException If an I/O error occurs.
      */
-    public Mono<Void> downloadToFile(String filePath, BlobRange range, Integer blockSize,
-        ReliableDownloadOptions options, BlobAccessConditions accessConditions, boolean rangeGetContentMD5) {
-        return withContext(context -> downloadToFile(filePath, range, blockSize, options, accessConditions,
+    public Mono<Response<BlobProperties>> downloadToFileWithResponse(String filePath, BlobRange range,
+                                    Integer blockSize, ReliableDownloadOptions options,
```
conniey: How did you get the indentation? I thought the standard was to either:
1. align with the method parameters above
1. or, do an 8 space indentation
https://stackoverflow.com/a/6023423

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,40 @@ public boolean isSnapshot() {
...
+            this::downloadToFileCleanup);
+
+    }
+
+    private Mono<Response<BlobProperties>> blobDownloadInChunk(final AsynchronousFileChannel channel, BlobRange range,
+                                    Integer blockSize, ReliableDownloadOptions options,
+                                    BlobAccessConditions accessConditions, boolean rangeGetContentMD5,
+                                    Context context) {
+        Mono<Response<BlobProperties>> asyncGetPropertiesResponse = getPropertiesWithResponse(accessConditions);
+        Mono<BlobRange> fullBlobRange = (range == null) ? getFullBlobRange(asyncGetPropertiesResponse)
```
conniey: I'd be consistent about ( ) for your ternary operators. There are braces here for the conditional statement, but the line below, 581, there is not.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
conniey: From my experience, it will be evaluated twice.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
conniey: Have you considered the .concat operator?

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileClient.java
```diff
@@ -315,7 +323,7 @@ public FileDownloadInfo downloadWithProperties() {
      *
      * <p>Delete the file</p>
      *
-     * {@codesnippet com.azure.storage.file.fileClient.delete}
+     * {q@codesnippet com.azure.storage.file.fileClient.delete}
```
samvaity: extra `q` at the start

sima-zhu: Done. Good catch

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
sima-zhu: Would add () for ternary.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
sima-zhu: Done.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: More on the stream structure not the response type.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,43 @@ public boolean isSnapshot() {
...
+
+    }
+
+    private Mono<Response<BlobProperties>> processInRange(final AsynchronousFileChannel channel,
+                                                          final Response<BlobProperties> blobPropertiesResponse,
+                                                          BlobRange range, Integer blockSize,
+                                                          ReliableDownloadOptions options,
+                                                          BlobAccessConditions accessConditions,
+                                                          boolean rangeGetContentMD5, Context context) {
+        return Mono.justOrEmpty(range).switchIfEmpty(Mono.just(new BlobRange(0,
```
alzimmermsft: This still results in two calls to getProperties

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -586,7 +603,8 @@ private void downloadToFileCleanup(AsynchronousFileChannel channel) {
     }
 
     private Mono<BlobRange> getFullBlobRange(BlobAccessConditions accessConditions) {
```
alzimmermsft: Does this get called anymore?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,43 @@ public boolean isSnapshot() {
      * @throws IllegalArgumentException If {@code blockSize} is less than 0 or greater than 100MB.
      * @throws UncheckedIOException If an I/O error occurs.
      */
-    public Mono<Void> downloadToFile(String filePath, BlobRange range, Integer blockSize,
-        ReliableDownloadOptions options, BlobAccessConditions accessConditions, boolean rangeGetContentMD5) {
-        return withContext(context -> downloadToFile(filePath, range, blockSize, options, accessConditions,
+    public Mono<Response<BlobProperties>> downloadToFileWithResponse(String filePath, BlobRange range,
+                                                                     Integer blockSize, ReliableDownloadOptions options,
```
conniey: This comment applies. This is mushed against the right margins.
https://github.com/Azure/azure-sdk-for-java/pull/5547#discussion_r329232537

conniey: I think a lot of these have really long names and the indentation should be fixed in this class. These params are hard to read because they're all on new lines.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,43 @@ public boolean isSnapshot() {
...
-                .then(), this::downloadToFileCleanup);
+            channel -> getPropertiesWithResponse(accessConditions).flatMap(response -> processInRange(channel, response,
+                range, blockSize, options, accessConditions, rangeGetContentMD5, context)),
+            this::downloadToFileCleanup);
+
+    }
+
+    private Mono<Response<BlobProperties>> processInRange(final AsynchronousFileChannel channel,
+                                                          final Response<BlobProperties> blobPropertiesResponse,
+                                                          BlobRange range, Integer blockSize,
```
conniey: As a nit, to consistently use `final` in your parameters or not.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,43 @@ public boolean isSnapshot() {
...
+                                                          ReliableDownloadOptions options,
+                                                          BlobAccessConditions accessConditions,
+                                                          boolean rangeGetContentMD5, Context context) {
+        return Mono.justOrEmpty(range).switchIfEmpty(Mono.just(new BlobRange(0,
+            blobPropertiesResponse.getValue().getBlobSize()))).flatMapMany(rg ->
+            Flux.fromIterable(sliceBlobRange(rg, blockSize)))
+            .flatMap(chunk -> this.download(chunk, accessConditions, rangeGetContentMD5, context)
+                .subscribeOn(Schedulers.elastic())
```
conniey: Why are you explicitly subscribing on the elastic scheduler? why not parallel? Are there an indefinite blocking tasks in the download that can grow unbounded?
https://projectreactor.io/docs/core/release/api/reactor/core/scheduler/Schedulers.html

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
sima-zhu: It is been removed.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
sima-zhu: This is done by @jianghao. I did not have the full background on this.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
sima-zhu: I have made a change on the stream calling. Could you take a look at the most recent code?

sima-zhu: Talked offline. Keep the most recent changes.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
sima-zhu: Thanks for the link. Changed the indentation.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,39 @@ public boolean isSnapshot() {
...
+            Integer blockSize, ReliableDownloadOptions options, BlobAccessConditions accessConditions,
+            boolean rangeGetContentMD5) {
+        return withContext(context -> downloadToFileWithResponse(filePath, range, blockSize, options, accessConditions,
             rangeGetContentMD5, context));
     }
 
-    Mono<Void> downloadToFile(String filePath, BlobRange range, Integer blockSize, ReliableDownloadOptions options,
-        BlobAccessConditions accessConditions, boolean rangeGetContentMD5, Context context) {
+    Mono<Response<BlobProperties>> downloadToFileWithResponse(String filePath, BlobRange range, Integer blockSize,
+        ReliableDownloadOptions options, BlobAccessConditions accessConditions, boolean rangeGetContentMD5,
```
conniey: Should be 8 spaces, not 4.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,39 @@ public boolean isSnapshot() {
...
-                .flatMapMany(rg -> Flux.fromIterable(sliceBlobRange(rg, blockSize)))
-                .flatMap(chunk -> this.download(chunk, accessConditions, rangeGetContentMD5, context)
-                    .subscribeOn(Schedulers.elastic())
-                    .flatMap(dar -> FluxUtil.writeFile(dar.body(options), channel,
-                        chunk.getOffset() - (range == null ? 0 : range.getOffset()))))
-                .then(), this::downloadToFileCleanup);
+            channel -> getPropertiesWithResponse(accessConditions).flatMap(response -> processInRange(channel, response,
+                range, blockSize, options, accessConditions, rangeGetContentMD5, context)),
+            this::downloadToFileCleanup);
+
```
conniey: nit: empty new line.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,39 @@ public boolean isSnapshot() {
...
+    }
+
+    private Mono<Response<BlobProperties>> processInRange(AsynchronousFileChannel channel,
+                Response<BlobProperties> blobPropertiesResponse, BlobRange range, Integer blockSize,
```
conniey: This is 12 spaces, not 8.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -545,27 +546,39 @@ public boolean isSnapshot() {
...
+                ReliableDownloadOptions options, BlobAccessConditions accessConditions, boolean rangeGetContentMD5,
+                Context context) {
+        return Mono.justOrEmpty(range).switchIfEmpty(Mono.just(new BlobRange(0,
+            blobPropertiesResponse.getValue().getBlobSize()))).flatMapMany(rg ->
+            Flux.fromIterable(sliceBlobRange(rg, blockSize)))
```
conniey: In general, new line splits are preferred on the higher level rather than this lower level break.

```java
return Mono.justOrEmpty(range)
    .switchIfEmpty(Mono.just(new BlobRange(0,  blobPropertiesResponse.getValue().getBlobSize())))
    .flatMapMany(rg -> Flux.fromIterable(sliceBlobRange(rg, blockSize)))
    .flatMap(chunk -> ...
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
conniey: Would be nice to ger an answer on this? From the documentation, it appears boundedElastic() is preferred over elastic(), anyways.

https://projectreactor.io/docs/core/release/reference/index.html#schedulers

Issue #5551, opened 2019-09-26
Change enum SettingFields to ExpandableStringEnum


mssfang:
In order to be able to expand enums dynamically,

JonathanGiles:
What is the use case for requiring this?

mssfang:
We enable it so service team to be able to add enum when they have new enums. We can talk about it more with Alan.

PR #5552, opened 2019-09-26
Changing SettingFields from Enum to ExpandableStringEnum

mssfang:
refactor SettingFields from Enum to ExpandableStringEnum
fixes: #5551


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingFields.java
```diff
@@ -12,40 +13,39 @@
  * @see SettingSelector
  * @see ConfigurationAsyncClient
  */
-public enum SettingFields {
+public class SettingFields extends ExpandableStringEnum<SettingFields> {
```
alzimmermsft: Let's make this a final class

PR #5557, opened 2019-09-26
OkHttp:  Don't send null header values (consistent with netty), support for subscription disposal

anuchandy:
1.  Null header value is not valid, already skipped that in NettyClient, doing same in OkHttp
2. Attempt to cancel OkHttp request upon downstream subscription disposal


anuchandy:
@g2vinay this fixes the issue you ran into when using azConfig with okhttp.

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
```diff
@@ -74,7 +79,13 @@
             .map(rb -> {
                 rb.url(request.getUrl());
                 if (request.getHeaders() != null) {
-                    return rb.headers(okhttp3.Headers.of(request.getHeaders().toMap()));
+                    Map<String, String> headers = new HashMap<>();
+                    for (HttpHeader hdr : request.getHeaders()) {
+                        if (hdr.getValue() != null) {
```
alzimmermsft: null or null or empty string?

anuchandy: Empty string as header value supported in Netty & OkHttp.

PR #5567, opened 2019-09-27
Fix errors caused by upgrading to latest reactor versions

srnagar:
OkHttpClient and NettyClient tests were failing after upgrading to new reactor version.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -115,7 +115,7 @@
     <slf4j-log4j12.version>1.7.0</slf4j-log4j12.version>
     <reactor-core.version>3.3.0.RELEASE</reactor-core.version>
     <reactor-netty.version>0.9.0.RELEASE</reactor-netty.version>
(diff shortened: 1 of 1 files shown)


anuchandy:
LGTM.

Some context:

1. Behaviour of Mono created from "using" factory changed in this release of reactor-core. In previous version, resource clean up used to be lazy by default which means upstream will wait for all downstream applied operators to complete then release the resource. In new version, default became eager means resource get clean up as soon as immediate operator is done. This was causing response body to be closed before we get a chance to consume it. 

2. In old and new release, reactor-netty's ByteBufFlux is one shot means can be consumed only once. In old release second consumption throws error (like we try to read from closed socket). In new version, after first subscription connection is recycled but second consumption gets mono.empty, no more error

anuchandy:
It's a breaking change, if any customer applied resumeOnError on flux backed by ByteBufFlux then this operator is  no-op in new release :(

PR #5571, opened 2019-09-27
Add Jim to codeowners and update link to docs

weshaggard:
cc @JimSuplizio


JimSuplizio:
LGTM - Thanks for doing this Wes.

PR #5575, opened 2019-09-27
Enables "EnableReceiverRuntimeMetric" feature from Track 1

### LINKED ISSUE #5147: [Event Hubs] Expose EnableReceiverRuntimeMetric
Enable EnableReceiverRuntimeMetric and update Consumer with these properties on each event received.

conniey:
* Exposes Track 1 feature as `trackLastEnqueuedEventProperties`
* Adds EventHubConsumerOptions.trackLastEnqueuedEventProperties
* Exposes it on EventHub*Consumer
* Moves translation of EventPosition into a string expression into EventHubReactorSession, so EventHubConsumer does not care about how this is used internally

Fixes #5147

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubAsyncConsumerIntegrationTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubAsyncConsumerIntegrationTest.java
@@ -99,13 +99,14 @@ public class EventHubAsyncConsumerIntegrationTest extends IntegrationTestBase {
                     EventPosition.fromEnqueuedTime(Instant.now()));
                 consumers[i] = consumer;
 
-                final Disposable subscription = consumer.receive().take(numberOfEvents).subscribe(event -> {
-                        logger.info("Event[{}] received. partition: {}", event.getSequenceNumber(), partitionId);
-                    }, error -> Assert.fail("An error should not have occurred:" + error.toString()),
-                    () -> {
-                        logger.info("Disposing of consumer now that the receive is complete.");
-                        countDownLatch.countDown();
-                    });
+                final Disposable subscription = consumer.receive().take(numberOfEvents)
+                    .subscribe(
+                        event -> logger.info("Event[{}] received. partition: {}", event.getSequenceNumber(), partitionId),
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/EventHubConsumerOptions.java
```diff
@@ -202,6 +222,17 @@ public int getPrefetchCount() {
...
     }
 
+    /**
+     * Gets whether or not the consumer should request information on the last enqueued event on its associated
+     * partition, and track that information as events are received.
+     *
+     * @return {@code true} if the resulting consumer will keep track of the last enqueued information for that
+     *     partition; {@code false} otherwise.
+     */
+    public boolean getLastEnqueuedEventProperties() {
```
srnagar: For auto-generated methods that return `boolean`, `get` is an acceptable prefix but in this case, should it be `isLastEnqueuedEventPropertiesTracked` to match `trackLastEnqueuedEventProperties`?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/EventHubReactorSession.java
```diff
@@ -67,27 +82,60 @@
...
+                isInclusiveFlag,
+                eventPosition.getSequenceNumber());
+        }
+
+        if (eventPosition.getEnqueuedDateTime() != null) {
+            String ms;
+            try {
+                ms = Long.toString(eventPosition.getEnqueuedDateTime().toEpochMilli());
+            } catch (ArithmeticException ex) {
+                ms = Long.toString(Long.MAX_VALUE);
```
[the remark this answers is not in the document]
conniey: I'm good with this. I'd prefer to fail fast. 🤔 It was carry over from Track 1.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/EventHubConsumerOptions.java
conniey: Thanks for the catch. I missed that.

PR #5577, opened 2019-09-27
Opinionated client libraries

### LINKED ISSUE #5511: Introduce opinionated azure-core

srnagar:
The intent of this change is to make Netty HTTP client the default for all client libraries as this is the most common scenario for non-Android users. For Android users, Netty can be excluded from the dependency and OkHTTP can be included.

Fixes #5511

### DIFF
--- a/sdk/storage/azure-storage-blob/README.md
+++ b/sdk/storage/azure-storage-blob/README.md
@@ -31,7 +31,7 @@ Storage Blob to use Netty HTTP client.
 
 ### Alternate HTTP client
 If, instead of Netty it is preferable to use OkHTTP, there is a HTTP client available for that too. Exclude the default
-Netty and include OkHTTP client in your pom.xml
+Netty and include OkHTTP client in your pom.xml.
 
 ```xml
 <!-- Add Storage Blob dependency without Netty HTTP client -->
--- a/sdk/appconfiguration/azure-data-appconfiguration/README.md
+++ b/sdk/appconfiguration/azure-data-appconfiguration/README.md
@@ -32,7 +32,7 @@ AppConfiguration to use Netty HTTP client.
 
 ### Alternate HTTP client
 If, instead of Netty it is preferable to use OkHTTP, there is a HTTP client available for that too. Exclude the default
-Netty and include OkHTTP client in your pom.xml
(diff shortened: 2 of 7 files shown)


sdk/keyvault/azure-keyvault-certificates/pom.xml
```diff
@@ -1,81 +1,82 @@
...
 
-       <dependency>
-        <groupId>com.azure</groupId>
-        <artifactId>azure-core-http-netty</artifactId>
-        <version>1.0.0-preview.5</version>
-        <scope>test</scope>
-       </dependency>
 
-    </dependencies>
+  </dependencies>
```
conniey: Phew!

sdk/storage/azure-storage-blob/README.md
```diff
@@ -26,21 +26,27 @@ definition, such as text or binary data.
 ```
 
 ### Default HTTP Client
-All client libraries support a pluggable HTTP transport layer. Users can specify an HTTP client specific for their needs by including the following dependency in the Maven pom.xml file:
+
+### Alternate HTTP client
+Netty and include OkHTTP client in your pom.xml
```
conniey: nit: You're missing a period to end off these sentences.

JonathanGiles:
Fixes #5511

PR #5613, opened 2019-09-30
Update return type for download API

### LINKED ISSUE #5409: Investigate a Better Return Type for Blob Download
Currently, the download method returns `Mono<Flux<ByteBuffer>>` to handle allowing the response to return once the headers and initial portion of the download stream returns and to delay reading of the body stream until requested. While the return type is technically correct it is daunting for someone newer to reactive coding and we should simplify this into a class that better handles this complexity in its internal implementation.

An option for this is using `DownloadAsyncResponse` (might want to rename it) in the non-maximal and maximal overloads (which partially goes against the API pattern).

srnagar:
The intent of this change is to simplify the return type for `download()` API in Blob client. Instead of returning a `Mono<Flux<ByteBuffer>>`, just return `Flux<ByteBuffer>` which is easier for the user.

Fixes #5409

### DIFF
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
@@ -1032,7 +1032,7 @@ class BlockBlobAPITest extends APISpec {
          buffering properly to allow for retries even given this source behavior.
          */
         bac.upload(Flux.just(defaultData), defaultDataSize).block()
-        def nonReplayableFlux = bac.download().block()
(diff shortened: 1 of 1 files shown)


JonathanGiles:
What about the `public Mono<Response<Flux<ByteBuffer>>> downloadWithResponse(..)` method? Any thoughts there?

srnagar:
`What about the public Mono<Response<Flux<ByteBuffer>>> downloadWithResponse(..) method? Any thoughts there?`

Discussed this with @alzimmermsft and decided that this is consistent with other APIs that return `Mono<Response<T>>` when the API is `withResponse` overload. In this case, `T` is `Flux<ByteBuffer>`. So, this API will remain as-is. Please let me know if you have other suggestions for this.

PR #5617, opened 2019-10-01
Remove SAS Generation Methods from Clients

### LINKED ISSUE #5541: Synchronize Java and .NET Generate SAS Methods
Java and .NET look drastically different with respects to generating SAS tokens. Java has many methods attached to clients to aid in this and this should be changed to follow what .NET has.

alzimmermsft:
Fixes #5541

### DIFF
--- a/sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/BlobServiceClientJavaDocCodeSnippets.java
+++ b/sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/BlobServiceClientJavaDocCodeSnippets.java
@@ -12,12 +12,7 @@ import com.azure.storage.blob.models.PublicAccessType;
 import com.azure.storage.blob.models.RetentionPolicy;
 import com.azure.storage.blob.models.StorageAccountInfo;
 import com.azure.storage.blob.models.StorageServiceProperties;
-import com.azure.storage.common.AccountSASPermission;
-import com.azure.storage.common.AccountSASResourceType;
-import com.azure.storage.common.AccountSASService;
-import com.azure.storage.common.Constants;
-import com.azure.storage.common.IpRange;
-import com.azure.storage.common.SASProtocol;
+
 import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobSasConstructor.java
```diff
@@ -0,0 +1,503 @@
...
+import com.azure.storage.common.credentials.SharedKeyCredential;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.time.OffsetDateTime;
+
+/**
+ * This class provides a fluent API to help aid in the creation of Azure Blob SAS Tokens.
+ */
+public class BlobSasConstructor {
```
JonathanGiles: 'Constructor' feels like an unfortunate suffix.

JonathanGiles: I almost feel compelled to let you call this a Builder with the required CheckStyle suppression.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobSasConstructor.java
```diff
@@ -0,0 +1,503 @@
...
+            sb.append(Utility.urlEncode(param)).append('=').append(Utility.urlEncode(value.toString()));
+        }
+    }
+
+    private String formatQueryDate(OffsetDateTime dateTime) {
+        if (dateTime == null) {
+            return null;
+        } else {
+            return Utility.ISO_8601_UTC_DATE_FORMATTER.format(dateTime);
+        }
```
JonathanGiles: We ain't paying for all these lines of code! Use ternary operator.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobSasConstructor.java
alzimmermsft: I think the `Builder` suffix feels more fluid. Will get additional input on this class before continuing forward on implementing this and using it in Files and Queues.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobSasConstructor.java
alzimmermsft: Will cleanup if the class is kept

PR #5620, opened 2019-10-01
Use pageRange in PageBlob.

sima-zhu:
PageBlobClient mixused of pageRange and blobRange.

Change the blobRange in PageBlob to pagePange.

Reference: https://docs.microsoft.com/en-us/rest/api/storageservices/operations-on-page-blobs

### DIFF
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/AppendBlobAPITest.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/AppendBlobAPITest.groovy
@@ -161,7 +161,6 @@ class AppendBlobAPITest extends APISpec {
 
     def "Append block defaults"() {
         setup:
-
         def appendResponse = bc.appendBlockWithResponse(defaultInputStream.get(), defaultDataSize, null, null, null)
 
         when:
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/PageBlobAPITest.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/PageBlobAPITest.groovy
@@ -518,7 +518,7 @@ class PageBlobAPITest extends APISpec {
         def response = bc.clearPagesWithResponse(new PageRange().setStart(0).setEnd(PageBlobClient.PAGE_BYTES - 1), null, null, null)
 
         then:
-        bc.getPageRanges(new BlobRange(0)).getPageRange().size() == 0
+        bc.getPageRanges(new PageRange().setStart(0).setEnd(PageBlobClient.PAGE_BYTES - 1)).getPageRange().size() == 0
         validateBasicHeaders(response.getHeaders())
         response.getValue().getContentMD5() == null
         response.getValue().getBlobSequenceNumber() == 0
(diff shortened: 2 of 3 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/PageBlobAsyncClient.java
```diff
@@ -373,24 +372,28 @@
...
-        return withContext(context -> getPageRangesWithResponse(blobRange, accessConditions, context));
+        return withContext(context -> getPageRangesWithResponse(pageRange, accessConditions, context));
     }
 
-    Mono<Response<PageList>> getPageRangesWithResponse(BlobRange blobRange, BlobAccessConditions accessConditions,
+    Mono<Response<PageList>> getPageRangesWithResponse(PageRange pageRange, BlobAccessConditions accessConditions,
         Context context) {
-        blobRange = blobRange == null ? new BlobRange(0) : blobRange;
+        if (pageRange == null) {
+            // Throwing is preferred to Single.error because this will error out immediately instead of waiting until
```
alzimmermsft: Why are we throwing an error here now? Also, if we keep this let's not use RxJava terminology.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/PageBlobAsyncClient.java
```diff
@@ -421,25 +424,30 @@
...
 
-    Mono<Response<PageList>> getPageRangesDiffWithResponse(BlobRange blobRange, String prevSnapshot,
+    Mono<Response<PageList>> getPageRangesDiffWithResponse(PageRange pageRange, String prevSnapshot,
         BlobAccessConditions accessConditions, Context context) {
-        blobRange = blobRange == null ? new BlobRange(0) : blobRange;
+        if (pageRange == null) {
+            // Throwing is preferred to Single.error because this will error out immediately instead of waiting until
+            // subscription.
+            throw logger.logExceptionAsError(new IllegalArgumentException("pageRange cannot be null."));
+        }
```
alzimmermsft: Same question here, if it is because there is no constructor for ’PageRange’ like ’BlobRange’ is there a possible substitute?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/AppendBlobAPITest.groovy
```diff
@@ -161,6 +161,7 @@ class AppendBlobAPITest extends APISpec {
 
     def "Append block defaults"() {
         setup:
+
```
alzimmermsft: Is this change needed?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/PageBlobAsyncClient.java
sima-zhu: PageRange required to specify the start and end. I don't think we need to set default to it.
https://docs.microsoft.com/en-us/rest/api/storageservices/put-page

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/AppendBlobAPITest.groovy
sima-zhu: No, will revert. Will update tests in another commit.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/PageBlobAsyncClient.java
alzimmermsft: PutRange needs this check, but get pages doesn't (which this uses).

https://docs.microsoft.com/en-us/rest/api/storageservices/get-page-ranges

sima-zhu: Make sense.

sima-zhu:
Close the PR as the changes do not apply according to the rest API doc.

PR #5621, opened 2019-10-01
Adds support for proxies in Event Hubs

### LINKED ISSUE #3580: [Event Hubs] Ensure Proxy support aligns with Track 2 guidelines
See if we need to support proxy configuration for Java.

conniey:
* Adds support for proxies identical to features available in Track 1
   * Users can specify ProxyConfiguration. If the proxy is configured, then the WebSocketsProxyConnectionHandler is used when connecting through proton-j to AMQP message broker.
* Adds tests for WebSocketsProxyConnectionHandler

Fixes #3580

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/jproxy/SimpleProxy.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/jproxy/SimpleProxy.java
@@ -11,7 +11,6 @@ import java.nio.channels.AsynchronousServerSocketChannel;
 import java.nio.channels.AsynchronousSocketChannel;
 import java.nio.channels.CompletionHandler;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
 /**
@@ -62,7 +61,6 @@ public class SimpleProxy implements ProxyServer {
         private final ClientLogger logger;
         private final Consumer<Throwable> onErrorHandler;
         private final AtomicBoolean isRunning;
-        private final AtomicReference<ProxyNegotiationHandler> negotiationHandler = new AtomicReference<>();
 
         private SocketListener(ClientLogger logger, Consumer<Throwable> onErrorHandler, AtomicBoolean isRunning) {
             this.logger = logger;
@@ -78,16 +76,11 @@ public class SimpleProxy implements ProxyServer {
                 logger.error("Error occurred while getting remote address.", e);
             }
 
-            if (negotiationHandler.get() != null) {
-                logger.error("Negotiation handler has already been set. We are not setting another.");
-                return;
-            }
-
             logger.info("Setting negotiation handler.");
             serverSocket.accept(serverSocket, this);
 
             try {
-                negotiationHandler.set(new ProxyNegotiationHandler(client));
+                new ProxyNegotiationHandler(client);
             } catch (IOException e) {
                 onErrorHandler.accept(e);
             }
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/proxy/ProxyReceiveTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/proxy/ProxyReceiveTest.java
@@ -8,7 +8,6 @@ import com.azure.core.util.logging.ClientLogger;
 import com.azure.messaging.eventhubs.EventHubAsyncClient;
 import com.azure.messaging.eventhubs.EventHubAsyncConsumer;
 import com.azure.messaging.eventhubs.EventHubClientBuilder;
-import com.azure.messaging.eventhubs.TestUtils;
 import com.azure.messaging.eventhubs.implementation.IntegrationTestBase;
 import com.azure.messaging.eventhubs.implementation.IntegrationTestEventData;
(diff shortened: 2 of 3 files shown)


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/handler/WebSocketsProxyConnectionHandler.java
```diff
@@ -0,0 +1,220 @@
...
+     * @param proxies List of proxies available.
+     *
+     * @return {@code true} if the first proxy in the list is an HTTP proxy and is an IP address.
+     */
+    private static boolean isProxyAddressLegal(final List<Proxy> proxies) {
+        // we look only at the first proxy in the list
+        // if the proxy can be translated to InetSocketAddress
+        // only then - can we parse it to hostName and Port
+        // which is required by qpid-proton-j library reactor.connectToHost() API
+        return proxies != null
```
conniey: This is the implementation in Track 1. I wonder if we should search through the list and get any that match rather than the first.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/handler/WebSocketsProxyConnectionHandler.java
```diff
@@ -0,0 +1,229 @@
...
+        final String[] hostNameParts = hostName.split(":");
+        if (hostNameParts.length != 2) {
+            return;
+        }
+
+        int port;
+        try {
+            port = Integer.parseInt(hostNameParts[1]);
+        } catch (NumberFormatException ignore) {
+            return;
```
srnagar: Add a log message to help debug issues.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/handler/WebSocketsProxyConnectionHandler.java
```diff
@@ -0,0 +1,229 @@
...
+        final boolean isProxyConfigured = proxySelector != null
+            || (proxyConfiguration != null && proxyConfiguration.isProxyAddressConfigured());
+
+        if (!isProxyConfigured || ImplUtils.isNullOrEmpty(hostName)) {
+            return;
+        }
+
+        final String[] hostNameParts = hostName.split(":");
+        if (hostNameParts.length != 2) {
+            return;
```
srnagar: Add a log message.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/handler/WebSocketsProxyConnectionHandler.java
```diff
@@ -0,0 +1,229 @@
...
+     * @param connectionId Identifier for this connection.
+     * @param amqpHostname Hostname of the AMQP message broker. The hostname of the proxy is exposed in {@link
+     *     #getHostname()}.
+     * @throws NullPointerException if {@code amqpHostname} or {@code proxyConfiguration} is null.
+     */
+    public WebSocketsProxyConnectionHandler(String connectionId, String amqpHostname,
+                                            ProxyConfiguration proxyConfiguration) {
+        super(connectionId, amqpHostname);
+        this.amqpHostname = Objects.requireNonNull(amqpHostname, "'amqpHostname' cannot be null.");
+        this.proxyConfiguration = Objects.requireNonNull(proxyConfiguration, "'proxyConfiguration' is required.");
```
srnagar: Use the same error message to be consistent. `'proxyConfiguration' cannot be null.`

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/handler/WebSocketsProxyConnectionHandler.java
```diff
@@ -0,0 +1,229 @@
...
+            : null;
+        final String password = proxyConfiguration.hasUserDefinedCredentials()
+            ? new String(proxyConfiguration.getCredential().getPassword())
+            : null;
+
+        return new com.microsoft.azure.proton.transport.proxy.ProxyConfiguration(type,
+            proxyConfiguration.getProxyAddress(), username, password);
+    }
+
+    private ProxyAuthenticationType getProtonAuthType(com.azure.core.amqp.models.ProxyAuthenticationType type) {
```
srnagar: Would be better to import track 2 types and use fully qualified names for track 1 (or com.microsoft.azure) types.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/handler/WebSocketsProxyConnectionHandler.java
srnagar: Don't have enough context about this. We should leave this as-is to ensure easy migration for clients from track 1 to track 2. Optionally, we can add a configuration to scan all proxies to find a match.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/exception/ErrorCondition.java
```diff
@@ -98,7 +98,11 @@
     /**
      * Tracking Id for an exception.
      */
-    TRACKING_ID_PROPERTY("com.microsoft:tracking-id");
+    TRACKING_ID_PROPERTY("com.microsoft:tracking-id"),
```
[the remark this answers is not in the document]
conniey: This property is populated by the Event Hubs service. The service team would have to update their server side logic if this wanted to be changed.

PR #5629, opened 2019-10-01
Fixed missing renaming APIs and bugs in convertStreamToByteBuffer

sima-zhu:


sima-zhu:
> Were the rest of the SASPermission objects renamed as well already? Like BlobSASPermission, FileSASPermission, etc?

Yes. The PR is to address the miss part from another renaming PR.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/ContainerSASPermission.java
```diff
@@ -77,108 +77,108 @@ public static ContainerSASPermission parse(String permString) {
...
     /**
      * Sets the read permission status.
      *
-     * @param read Permission status to set
+     * @param readPermission Permission status to set
      * @return the updated ContainerSASPermission object
      */
-    public ContainerSASPermission setRead(boolean read) {
-        this.read = read;
+    public ContainerSASPermission setReadPermission(boolean readPermission) {
```
alzimmermsft: Hasn't the name for the parameter been `hasXPermission`?

sima-zhu: Done.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
```diff
@@ -514,9 +514,9 @@ public static SharedKeyCredential getSharedKeyCredential(HttpPipeline httpPipeli
                 byte[] cache = new byte[(int) count];
                 int lastIndex = data.read(cache);
                 currentTotalLength[0] += lastIndex;
-                if (currentTotalLength[0] < count) {
```
jaschrep-msft: I don't understand what this array of size 1 is (or why it's even an array) and so I don't understand this change.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
```diff
@@ -27,11 +27,13 @@ import com.azure.storage.blob.models.StorageErrorCode
 import com.azure.storage.blob.models.StorageErrorException
 import com.azure.storage.blob.models.StorageException
 import com.azure.storage.common.policy.RequestRetryOptions
+
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
 import spock.lang.Requires
 import spock.lang.Unroll
 
+import java.lang.reflect.Field
```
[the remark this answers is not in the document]
sima-zhu: Will remove.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
sima-zhu: This is because we need to variable to remember the current total length.
However, the lambda will complain the variable must be immutable if having  something like `int sum`. 

This is the best way of recording the current length in lambda, based on the discussion with team and @anuchandy 

The fix here is to compare the chunk we currently process, so it should compare the reading bytes and the expected to have for the chunk, not the currentTotalLength.

PR #5641, opened 2019-10-01
Removed azure-core-management so that it does not get built releasing

hemanttanwar:


joshfree:
```
# DO NOT EDIT THIS FILE	# DO NOT EDIT THIS FILE
# This file is generated automatically and any changes will be lost.
```

JonathanGiles:
Yeah, let's get confirmation from @mitchdenny

mitchdenny:
That's fine. Think of that more like a scare tactic for now :) it will hopefully be code generated in the future.

PR #5648, opened 2019-10-02
Change URL Getters to String in Storage

### LINKED ISSUE #5602: Change getLeaseUrl to getResourceUrl
Blobs doesn't have a lease resource type and given that the lease client could either be a blob or container use the generalized resource naming.

samvaity:
Closes #5602

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/LeaseAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/LeaseAsyncClient.java
@@ -69,7 +69,7 @@ public final class LeaseAsyncClient {
      *
      * @return URL of the lease client.
      */
-    public String getLeaseUrl() {
+    public String getResourceUrl() {
         return this.client.getUrl();
     }
 
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/LeaseClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/LeaseClient.java
@@ -49,8 +49,8 @@ public final class LeaseClient {
      *
      * @return URL of the lease client.
      */
-    public String getLeaseUrl() {
-        return client.getLeaseUrl();
+    public String getResourceUrl() {
+        return client.getResourceUrl();
     }
 
     /**


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -108,21 +106,28 @@ public BlobAsyncClientBase getSnapshotClient(String snapshot) {
...
-                String.format("Invalid URL on %s: %s" + getClass().getSimpleName(), azureBlobStorage.getUrl()), e));
+    public String getBlobUrl() {
+        if (!this.isSnapshot()) {
+            return azureBlobStorage.getUrl();
+        } else {
+            return String.format("%s?snapshot=%s", azureBlobStorage.getUrl(), snapshot);
         }
     }
 
+    // public String getBlobUrl() {
```
alzimmermsft: Remove the commented out code.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -108,21 +106,28 @@ public BlobAsyncClientBase getSnapshotClient(String snapshot) {
...
-            }
-            return urlBuilder.toURL();
-        } catch (MalformedURLException e) {
-            throw logger.logExceptionAsError(new RuntimeException(
-                String.format("Invalid URL on %s: %s" + getClass().getSimpleName(), azureBlobStorage.getUrl()), e));
+    public String getBlobUrl() {
+        if (!this.isSnapshot()) {
+            return azureBlobStorage.getUrl();
+        } else {
+            return String.format("%s?snapshot=%s", azureBlobStorage.getUrl(), snapshot);
```
alzimmermsft: Should make a check that there are no query parameters already attached to the URL, maybe check for a `?` as I believe that is a reserved URL character.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
```diff
@@ -716,7 +715,7 @@ class BlobAPITest extends APISpec {
         setup:
         def copyDestBlob = cc.getBlobClient(generateBlobName()).asBlockBlobClient()
         def headers =
-            copyDestBlob.startCopyFromURLWithResponse(bc.getBlobUrl(), null, null, null, null, null, null, null).getHeaders()
+            copyDestBlob.startCopyFromURLWithResponse(new URL(bc.getBlobUrl()), null, null, null, null, null, null, null).getHeaders()
```
JonathanGiles: If this method requires a URL, then there should be a follow-up change (either here or in a separate PR) to changes these method calls to also take a String url.

alzimmermsft: This is an interesting case as the Swagger interface is expecting a URL parameter, I'm wondering if it could make more sense as a String otherwise this method will need to handle the creation of the URL.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
samvaity: updated.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
samvaity: Have created this issue to track this - #5656

JonathanGiles: I am fine with tackling this in a follow-up issue

PR #5650, opened 2019-10-02
Deleted updateSetting and updated setSetting

mssfang:
Deleted updateSetting and corresponding sample and test
Updated setSetting method, also tests and samples

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
@@ -78,7 +78,8 @@ public final class ConfigurationAsyncClient {
      *
      * @param key The key of the configuration setting to add.
      * @param value The value associated with this configuration setting key.
-     * @param label Optional, the label of the configuration setting to add.
+     * @param label The label of the configuration setting to create or update, or optionally, null if a setting with
+     * label is desired.
      * @return The {@link ConfigurationSetting} that was created, if a key collision occurs or the key is an invalid
      * value (which will also throw HttpResponseException described below).
      * @throws IllegalArgumentException If {@code key} is {@code null}.
@@ -170,7 +171,8 @@ public final class ConfigurationAsyncClient {
      * {@codesnippet com.azure.data.appconfiguration.configurationasyncclient.setSetting#string-string-string}
      *
      * @param key The key of the configuration setting to create or update.
-     * @param label Optional, the label of the setting to create or update.
+     * @param label The label of the configuration setting to create or update, or optionally, null if a setting with
+     * label is desired.
      * @param value The value of this configuration setting.
      * @return The {@link ConfigurationSetting} that was created or updated, if the key is an invalid value (which will
      * also throw HttpResponseException described below).
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClient.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClient.java
@@ -55,7 +55,8 @@ public final class ConfigurationClient {
      *
      * @param key The key of the configuration setting to add.
      * @param value The value associated with this configuration setting key.
-     * @param label Optional, the label of the configuration setting to add.
+     * @param label The label of the configuration setting to create or update, or optionally, null if a setting with
+     * label is desired.
      * @return The {@link ConfigurationSetting} that was created, or {@code null}, if a key collision occurs or the key
      * is an invalid value (which will also throw ServiceRequestException described below).
      * @throws IllegalArgumentException If {@code key} is {@code null}.
@@ -130,7 +131,8 @@ public final class ConfigurationClient {
      * {@codesnippet com.azure.data.appconfiguration.ConfigurationClient.setSetting#String-String-String}
      *
      * @param key The key of the configuration setting to create or update.
-     * @param label Optional, the label of the configuration setting to create or update.
+     * @param label The label of the configuration setting to create or update, or optionally, null if a setting with
+     * label is desired.
      * @param value The value of this configuration setting.
      * @return The {@link ConfigurationSetting} that was created or updated, or {@code null}, if the key is an invalid
      * value (which will also throw ServiceRequestException described below).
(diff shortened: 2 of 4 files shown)


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
```diff
@@ -195,40 +197,10 @@
...
-     *
-     * <p><strong>Code Samples</strong></p>
-     *
-     * <p>Add a setting with the key "prodDBConnection", label "westUS", and value "db_connection".</p>
-     *
-     * {@codesnippet com.azure.data.appconfiguration.configurationasyncclient.setSettingWithResponse#ConfigurationSetting}
+     * {@codesnippet com.azure.data.appconfiguration.configurationasyncclient.setSettingWithResponse#ConfigurationSetting-boolean}
      *
      * @param setting The configuration setting to create or update.
+     * @param onlyIfUnchanged A boolean indicates if using setting's ETag as If-Match's value.
```
conniey: This should be geared towards customers. When would a customer want to set "onlyIfUnchanged" to true? The parameter name is odd, "ifUnchanged" feels sufficient.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClient.java
```diff
@@ -119,15 +119,16 @@ public ConfigurationSetting addSetting(ConfigurationSetting setting) {
     }
 
     /**
-     * Creates or updates a configuration value in the service with the given key.
+     * Creates or updates a configuration value in the service with the given key, optional {@code label}
```
conniey:
```suggestion
     * Creates or updates a configuration value in the service with the given key and optional {@code label}.
```

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
alzimmermsft: Agreed, I prefer `ifChanged` as well

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
```diff
@@ -240,125 +212,30 @@
...
+        return withContext(context -> setSetting(setting, onlyIfUnchanged, context));
     }
 
-    Mono<Response<ConfigurationSetting>> setSetting(ConfigurationSetting setting, Context context) {
+    Mono<Response<ConfigurationSetting>> setSetting(ConfigurationSetting setting, boolean onlyIfUnchanged,
+                                                    Context context) {
         // Validate that setting and key is not null. The key is used in the service URL so it cannot be null.
         validateSetting(setting);
 
+        final String ifMatchETag = onlyIfUnchanged ? getETagValue(setting.getETag()) : "*";
```
alzimmermsft: I don't think we should pass an `ETag` if there is no check for unchanged, this will cause the set to fail if the setting doesn't exist.

sdk/appconfiguration/azure-data-appconfiguration/src/samples/java/com/azure/data/appconfiguration/HelloWorld.java
```diff
@@ -40,7 +40,7 @@ public static void main(String[] args) throws NoSuchAlgorithmException, InvalidK
         // We subscribe and wait for the service call to complete then print out the contents of our newly added setting.
         // If an error occurs, we print out that error. On completion of the subscription, we delete the setting.
         // .block() exists there so the program does not end before the deletion has completed.
-        client.setSetting(key, "world").subscribe(
+        client.setSetting(key, null, "world").subscribe(
```
alzimmermsft: Do we want to use the NO_LABEL static here to show what passing null means?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClient.java
```diff
@@ -166,8 +169,8 @@ public ConfigurationSetting setSetting(String key, String value) {
      * @throws HttpResponseException If {@code key} is an empty string.
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
-    public ConfigurationSetting setSetting(ConfigurationSetting setting) {
-        return setSettingWithResponse(setting, Context.NONE).getValue();
+    public ConfigurationSetting setSetting(ConfigurationSetting setting, boolean onlyIfUnchanged) {
```
alzimmermsft: Merge this with the maximal overload.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClient.java
```diff
@@ -196,96 +200,9 @@ public ConfigurationSetting setSetting(ConfigurationSetting setting) {
...
-     * @throws HttpResponseException If {@code key} is an empty string.
-     */
-    @ServiceMethod(returns = ReturnType.SINGLE)
-    public Response<ConfigurationSetting> updateSettingWithResponse(ConfigurationSetting setting, Context context) {
-        return updateSetting(setting, context);
-    }
-
-    private Response<ConfigurationSetting> updateSetting(ConfigurationSetting setting, Context context) {
-        return client.updateSetting(setting, context).block();
+    public Response<ConfigurationSetting> setSettingWithResponse(ConfigurationSetting setting, boolean onlyIfUnchanged,
```
alzimmermsft: Should this be annotated with `ServiceMethod`?

sdk/appconfiguration/azure-data-appconfiguration/src/samples/java/com/azure/data/appconfiguration/ConfigurationAsyncClientJavaDocCodeSnippets.java
```diff
@@ -16,6 +16,7 @@
  */
 public class ConfigurationAsyncClientJavaDocCodeSnippets {
 
+    private static final String NO_LABEL = null;
```
alzimmermsft: Doesn't this exist in the `ConfigurationSetting` class? If the one in `ConfigurationSetting` isn't public lets  make it so.

PR #5655, opened 2019-10-02
Move post processor to policy

sima-zhu:
To remove post processor, we have to deal with scrubEtag, and throw StorageException.
1. Added ScrubEtagPolicy into each builder.
2. Change transformor to throw StorageException instead of StorageErrorException.

### DIFF
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
@@ -17,9 +17,8 @@ import java.nio.charset.Charset;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-import static io.netty.handler.codec.http.HttpHeaders.Names.ETAG;
-
 public class ScrubEtagPolicy implements HttpPipelinePolicy {
+    private static final String ETAG = "eTag";
 
     private HttpResponse innerHttpResponse;
     private HttpHeaders headers;


joshfree:
> Only update a few for testing purpose. Will continue work on this in Preview 5.

We should aim to be API-complete in Preview 4.  Any changes that add new public surface area should be prioritized for Preview 4 code complete.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileServiceAsyncClient.java
```diff
@@ -393,8 +393,10 @@ public ShareAsyncClient getShareAsyncClient(String shareName, String snapshot) {
         Integer quotaInGB, Context context) {
         ShareAsyncClient shareAsyncClient = new ShareAsyncClient(azureFileStorageClient, shareName, null);
 
-        return postProcessResponse(shareAsyncClient.createWithResponse(metadata, quotaInGB, context))
-            .map(response -> new SimpleResponse<>(response, shareAsyncClient));
+//        return postProcessResponse(shareAsyncClient.createWithResponse(metadata, quotaInGB, context))
```
rickle-msft: Delete

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/ExceptionMappingPolicy.java
```diff
@@ -0,0 +1,35 @@
...
+import com.azure.core.http.HttpPipelineCallContext;
+import com.azure.core.http.HttpPipelineNextPolicy;
+import com.azure.core.http.HttpResponse;
+import com.azure.core.http.policy.HttpPipelinePolicy;
+import reactor.core.publisher.Mono;
+
+/**
+ * This is a exception mapping policy offers functionality to mapping the error to specific exception.
+ */
+public class ExceptionMappingPolicy implements HttpPipelinePolicy {
```
anuchandy: Should we move `ExceptionMappingPolicy` types and `ScrubEtagPolicy` to implementation package? to avoid appearing them in Javadoc and to enable us to modify/delete it without breaking GAed lib.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/ExceptionMappingPolicy.java
```diff
@@ -0,0 +1,35 @@
...
+     */
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        return next.process()
+            .onErrorResume(StorageErrorException.class, resume ->
+                    resume.getResponse()
+                        .getBodyAsString()
+                        .switchIfEmpty(Mono.just(""))
+                        .flatMap(body -> Mono.error(new StorageException(resume, body)))
```
anuchandy: empty and null should be treated differently. can we have something like below?

```java
.getBodyAsString()
.flatMap(body -> Mono.error(new StorageException(resume, body)))
.switchIfEmpty(Mono.defer(() -> Mono.error(new StorageException(resume, null))));
```

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/ExceptionMappingPolicy.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        return next.clone().process()
+            .onErrorResume(StorageErrorException.class, resume ->
+                    resume.getResponse()
+                        .getBodyAsString()
+                        .switchIfEmpty(Mono.just(""))
+                        .flatMap(body -> Mono.error(new StorageException(resume, body)))
```
anuchandy: same comment here as above w.r.t empty vs null body.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/StorageException.java
```diff
@@ -19,34 +20,29 @@
...
+    public StorageException(String message, HttpResponse response, Object value) {
+        super(message, response, value);
     }
 
     /**
      * @return The error code returned by the service.
      */
     public StorageErrorCode getErrorCode() {
-        return this.errorCode;
+        return StorageErrorCode.fromString(super.getResponse().getHeaders().getValue(ERROR_CODE));
```
alzimmermsft: Why did we remove this from the constructor? We are forcing a parsing of the response headers each time it is accessed.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/StorageException.java
```diff
@@ -19,34 +20,29 @@
...
     }
 
     /**
      * @return The message returned by the service.
      */
     public String getServiceMessage() {
-        return this.message;
+        return super.getMessage();
```
alzimmermsft: Make sure that this message and the old message are the same.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/StorageException.java
```diff
@@ -19,34 +20,29 @@
  */
 public final class StorageException extends HttpResponseException {
     private static final String ERROR_CODE = "x-ms-error-code";
```
alzimmermsft: You can use the Constant in common instead of this local String.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/BaseClientBuilder.java
```diff
@@ -25,6 +25,7 @@
 import com.azure.storage.common.policy.RequestRetryOptions;
 import com.azure.storage.common.policy.RequestRetryPolicy;
 import com.azure.storage.common.policy.ResponseValidationPolicyBuilder;
+import com.azure.storage.common.policy.ScrubEtagPolicy;
```
alzimmermsft: This builder is going to be removed as an FYI.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+        if (eTagHeader == null) {
+            return unprocessedResponse;
+        }
+        String eTag = eTagHeader.getValue();
+
+        eTag = eTag.replace("\"", "");
+        headers = unprocessedResponse.getHeaders();
+        headers.put(eTagHeader.getName(), eTag);
+        innerHttpResponse = unprocessedResponse;
+        return new InnerHttpResponse(unprocessedResponse.getRequest());
```
alzimmermsft: Why are we wrapping this response instead of just returning the `unprocessedResponse`?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Wraps any potential error responses from the service and applies post processing of the response's eTag header to
+     * standardize the value.
+     *
+     * @return an updated response with post processing steps applied.
+     */
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        return next.process()
+            .flatMap(response -> {
```
alzimmermsft: This could just be `.flatMap(response -> Mono.justOrEmpty(scrubETagHeader(response))`.

Also, do we need to have `justOrEmpty` as this shouldn't be able to be null.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+import com.azure.core.http.policy.HttpPipelinePolicy;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public class ScrubEtagPolicy implements HttpPipelinePolicy {
+    private static final String ETAG = "eTag";
```
alzimmermsft: I believe there is a constant in common for `eTag`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/StorageException.java
sima-zhu: The rest deserilizer in restProxy had a hard time to deserialize the statement in constructor.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/StorageException.java
sima-zhu: Our queue and file has some tests on error code. They got right messages.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/BaseClientBuilder.java
sima-zhu: OK. I will resolve this after the PR merged in.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/StorageException.java
sima-zhu: Done.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
sima-zhu: Our httpResponse is final one. We cannot modify its headers.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
sima-zhu: Done.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
sima-zhu: ETAG in common is "ETag", what the response header has in name is "eTag".

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/ExceptionMappingPolicy.java
sima-zhu: Deleted the policy.  For the ScrubEtagPolicy, put it in common policy.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
alzimmermsft: I believe `final` indicates that the memory location reference cannot change, not the values contained within that reference. Mind testing this out with this removed?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
alzimmermsft: The `get` method on `HttpHeaders` uses a case insensitive implementation, so this should be fine. Additionally, `ETag` is the more correct implementation as that is what the documentation states is the header return name.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
sima-zhu: Sure. I have local record tests passed. Will trigger the live test in pipeline

sima-zhu: We returned a new HttpResponse back instead of making changes to old one.

sima-zhu:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=143686&view=logs&s=96ac2280-8cb4-5df5-99de-dd2da759617d

sima-zhu:
Live tests failed because of some changes before the change.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
sima-zhu: Cannot find the ETAG anymore. Leave it as it is.

Issue #5526, opened 2019-09-25
[Feature] Make RetryPolicy Configurable in Builder


alzimmermsft:
The default RetryPolicy is used right now and may not fit everyone's use case, make it settable in the builder.

PR #5669, opened 2019-10-03
add custom retry policy

mssfang:
fixes: https://github.com/Azure/azure-sdk-for-java/issues/5526

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
@@ -282,8 +282,8 @@ public final class ConfigurationClientBuilder {
     /**
      * Sets the {@link RetryPolicy} that is used when each request is sent.
      *
-     * The default retry policy will be used if not provided.
-     *
+     * The default retry policy will be used if not provided {@link ConfigurationClientBuilder#buildAsyncClient()}
+     * to build {@link ConfigurationAsyncClient} or {@link ConfigurationClient}.
      * @param retryPolicy RetryPolicy applied to each request.
      * @return The updated ConfigurationClientBuilder object.
      */


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -279,6 +279,19 @@ public ConfigurationClientBuilder configuration(Configuration configuration) {
         return this;
     }
 
+    /**
+     * Sets the retry policy that is used during construction of the service client.
```
alzimmermsft:
```suggestion
     * Sets the retry policy that is used when each request is sent.
```

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -279,6 +279,19 @@ public ConfigurationClientBuilder configuration(Configuration configuration) {
...
+     *
+     * The default retry policy will be used if not provided.
```
alzimmermsft: Wrap this in <p></p>, maybe link the default constructor.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -279,6 +279,19 @@ public ConfigurationClientBuilder configuration(Configuration configuration) {
...
+     *
+     * @param retryPolicy The custom retry policy
```
alzimmermsft:
```suggestion
     * @param retryPolicy RetryPolicy applied to each request.
```

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
mssfang: I delete this constructor instead of having a setter to set

mssfang: Because it is a builder pattern. Also we need to renaming all to getter and setter

PR #5680, opened 2019-10-03
Removed SASTokenCredential and SASTokenCredentialPolicy

### LINKED ISSUE #5636: Remove SAS Token Credential and Corresponding Policy
Remove these classes and just maintain the SAS token on the URI.

alzimmermsft:
Fixes #5636 

Removes `SASTokenCredential` and `SASTokenCredentialPolicy` and updates the builders to have a `sasToken` method.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
+++ b/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
@@ -200,4 +200,7 @@
 
   <!-- Checkstyle rules should not check files in generated-test-sources -->
   <suppress checks="[a-zA-Z0-9]*" files="[/\\]generated-test-sources[/\\]"/>
+
+  <!-- Allows the SasTokenCredentialPolicy in Implementation -->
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
```diff
@@ -86,11 +86,11 @@
      * @param queryString Query string to parse
      * @return a mapping of query string pieces as key-value pairs.
      */
-    public static TreeMap<String, String[]> parseQueryStringSplitValues(final String queryString) {
+    public static Map<String, String[]> parseQueryStringSplitValues(final String queryString) {
         return parseQueryStringHelper(queryString, (value) -> urlDecode(value).split(","));
     }
 
-    private static <T> TreeMap<String, T> parseQueryStringHelper(final String queryString,
+    private static <T> Map<String, T> parseQueryStringHelper(final String queryString,
```
anuchandy: reg line 102: just double checking, is it always required that query string part of SaS url to be in `a=b` format? not sure about rules around blob url, any chance it could be something like `a=b&c=&d&e=f` here c and d has no value.

alzimmermsft: Discussed offline, there is an expectation for the query parameters to have values. This is a small guard that should be done anyways in the future.

PR #5685, opened 2019-10-04
add ResourceBodyNotFoundException

mssfang:
Add ResourceBodyNotFoundException for Response<T> with status code 304 but access response.getValue() will return ResourceBodyNotFoundException


mssfang:
Thanks for your feedback. Agreed with your options. HttpResponseException will be sufficient.  Will close this PR without merged.

Issue #4154, opened 2019-06-27
Logging improvements


JonathanGiles:
The ClientLogger API in azure-core should be improved to enable new features to be supported. The features we want to offer are:

1. **Support for logging redaction**: The ClientLogger API, as the central location for all logging activities in all Java client libraries, might benefit from having a pluggable API for string redaction, before the message is output into the console, output files, etc. For example, consumers do not want their secret keys, passwords, etc to be published into log files, where they could more easily end up in the wrong hands. One approach we could take would be to have a pluggable API (either an SPI or just a static method on ClientLogger) to register string redactors. Before any logging message is output, the string will run through all redactors. Ideally we could make this more performant by only having redactors execute if the logging message has come from that service in the first place.

2. **Custom logging strategies**: We should provide a plugin API that allows for client libraries to augment the default logging framework with additional strategies. For example, if a client library wishes to write all WARNING level logging information into a rotating log file, even in the absence of user configuration or opt-in, this strategy should be enabled by a custom plugin that can read all log messages as they flow through the ClientLogger.

To support these requirements, we might need to explore allowing for hierarchical loggers to be defined, so that the pluggable features for redaction and custom logging strategies do not need to be defined against every ClientLogger instance.

joshfree:
@JonathanGiles is going to investigate this for Preview 4 / Preview 5

JonathanGiles:
For logging redaction, based on the [.Net approach](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/Pipeline/DiagnosticsOptions.cs#L35-L37), I propose the following:

Where today our client builders offer a `httpLogLevelDetail(HttpLogLevelDetail logLevel)` method, we would remove this and replace it with a `httpOptions(HttpOptions options)` method. This would be a fluent API along the following lines:

``` java
public class HttpOptions {
    private HttpLogLevelDetail logLevel;
    private List<String> allowedHeaderNames;
    private List<String> allowedQueryParamNames;

   // set / get logLevel

   // set / get / add whitelisted names 
}
```

We would then update the HttpLoggingPolicy to inspect these two lists and only output white listed values.

alzimmermsft:
Thoughts on adding a Function that is able to process the response and return a logging level? Storage had functionality in previous clients that would escalate a logging level based on response time.

srnagar:
[...]

PR #5688, opened 2019-10-04
Adding HttpLogOptions for Logging improvements

samvaity:
Reference #4154

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
@@ -180,7 +180,9 @@ public class HttpLoggingPolicy implements HttpPipelinePolicy {
                 logger.info("<-- {} {} ({} ms, {} body)", response.getStatusCode(), url, tookMs, bodySize);
             }
 
-            formatAllowableHeaders(httpLogOptions.getAllowedHeaderNames(), response.getHeaders(), logger);
+            if (httpLogLevel.shouldLogHeaders()) {
+                formatAllowableHeaders(httpLogOptions.getAllowedHeaderNames(), response.getHeaders(), logger);
+            }
 
             if (httpLogLevel.shouldLogBody()) {
                 long contentLength = getContentLength(response.getHeaders());


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -207,14 +208,14 @@ public ConfigurationClientBuilder credential(ConfigurationClientCredentials cred
     }
 
     /**
-     * Sets the logging level for HTTP requests and responses.
+     * Sets the logging configuration for HTTP requests and responses.
      *
-     * @param logLevel The amount of logging output when sending and receiving HTTP requests/responses.
+     * @param logOptions The configuration of logging output when sending and receiving HTTP requests/responses.
```
JonathanGiles:
```suggestion
     * @param logOptions The logging configuration to use when sending and receiving HTTP requests/responses.
```

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationAsyncClientTest.java
```diff
@@ -39,13 +40,13 @@ protected void beforeTest() {
             client = clientSetup(credentials -> new ConfigurationClientBuilder()
                     .credential(credentials)
                     .httpClient(interceptorManager.getPlaybackClient())
-                    .httpLogDetailLevel(HttpLogDetailLevel.BODY_AND_HEADERS)
+                .httpLogOptions(new HttpLogOptions(HttpLogDetailLevel.BODY_AND_HEADERS))
```
JonathanGiles: Bad indentation here

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
```diff
@@ -0,0 +1,73 @@
...
+
+package com.azure.core.http.policy;
+
+import java.util.Set;
+
+/**
+ * The log configurations for HTTP messages.
+ */
+public class HttpLogOptions {
+    public HttpLogOptions(final HttpLogDetailLevel logLevel) {
```
JonathanGiles: I would make this a no-arg constructor and have a fluent API. Have each set method return 'this', so that it can be chained.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
```diff
@@ -0,0 +1,73 @@
...
+    public Set<String> getAllowedHeaderNames() {
+        return allowedHeaderNames;
+    }
+
+    /**
+     * Sets the given whitelisted headers that should be logged.
+     *
+     * @param allowedHeaderNames The list of whitelisted header names from the user.
+     */
+    public void setAllowedHeaderNames(final Set<String> allowedHeaderNames) {
```
JonathanGiles: have an 'addAllowedHeaderName' option as well to take a single String.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
```diff
@@ -0,0 +1,73 @@
...
+    public Set<String> getAllowedQueryParamNames() {
+        return allowedQueryParamNames;
+    }
+
+    /**
+     * Sets the given whitelisted query params to be displayed in the logging info.
+     *
+     * @param allowedQueryParamNames The list of whitelisted query params from the user.
+     */
+    public void setAllowedQueryParamNames(final Set<String> allowedQueryParamNames) {
```
JonathanGiles: Same here - have an 'add'.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
```diff
@@ -0,0 +1,73 @@
...
+        this.allowedQueryParamNames = allowedQueryParamNames;
+    }
+
+    private HttpLogDetailLevel logLevel;
+    private Set<String> allowedHeaderNames;
+    private Set<String> allowedQueryParamNames;
```
JonathanGiles: Move fields to top.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogDetailLevel.java
```diff
@@ -32,7 +32,12 @@
     /**
      * Logs everything in HEADERS and BODY.
      */
-    BODY_AND_HEADERS;
+    BODY_AND_HEADERS,
+
+    /**
+     * Logs everything in QUERY PARAMS.
+     */
+    QUERY_PARAMS;
```
JonathanGiles: This doesn't quite make sense - if the user opts in for QUERY_PARAMS, that means that they can't also get BODY or BODY_AND_HEADERS.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -115,7 +120,40 @@ public HttpLoggingPolicy(HttpLogDetailLevel detailLevel, boolean prettyPrintJSON
...
+    private void formatAllowableQueryParams(Set<String> allowedQueryParamNames, HttpRequest request,
+                                            ClientLogger logger) {
+        StringBuilder sb = new StringBuilder();
+        String queryString = request.getUrl().getQuery();
+        String[] queryParams = queryString.split("&");
+        for (String queryParam : queryParams) {
+            String[] queryPair = queryParam.split("=");
+            if (2 != queryPair.length) {
+                throw new IllegalArgumentException("Parse failed for " + queryString);
+            }
```
anuchandy: It would be better if not do any request validation (and throwing of errors) in logging framework. I would rather do:

```java
 String[] queryPair = queryParam.split("=", 2);
 if (queryPair.length == 2) {
   if(allowedQueryParamNames.contains(queryPair[0])) {
       sb.append(queryParam);
   } else {
       sb.append(queryPair[0]).append("=").append(REDACTED_PLACEHOLDER);
   }
 } else {
   sb.append(queryParam);
 }
```

this way we leave parameters like `c` and `d` in `?a=b&c=&d` as it is.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -115,7 +120,40 @@ public HttpLoggingPolicy(HttpLogDetailLevel detailLevel, boolean prettyPrintJSON
...
+            } else {
+                logger.info(REDACTED_PLACEHOLDER);
+            }
+        }
+    }
+
+    private void formatAllowableQueryParams(Set<String> allowedQueryParamNames, HttpRequest request,
+                                            ClientLogger logger) {
+        StringBuilder sb = new StringBuilder();
+        String queryString = request.getUrl().getQuery();
```
anuchandy: I guess we need a null check for `queryString`, if i remember correctly `getQuery()` can return null if there is no query segment in the uri. just double check.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
samvaity: Yeah, even I was thinking of the same to not do any validation. But just put out there to know it better. Will remove it, 👍

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogDetailLevel.java
samvaity: so should we keep the requirement to log query params on only BODY or BODY_AND_HEADERS or both?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -115,7 +120,40 @@ public HttpLoggingPolicy(HttpLogDetailLevel detailLevel, boolean prettyPrintJSON
         return reqBodyLoggingMono;
     }
 
-    private Function<HttpResponse, Mono<HttpResponse>> logResponseDelegate(final ClientLogger logger, final URL url,
+    private void formatAllowableHeaders(Set<String> allowedHeaderNames, HttpRequest request, ClientLogger logger) {
+        for (HttpHeader header : request.getHeaders()) {
+            if (allowedHeaderNames.contains(header)) {
+                logger.info(header.toString());
+            } else {
+                logger.info(REDACTED_PLACEHOLDER);
```
anuchandy: For a query parameter pair we log key as it is and value as REDACTED i.e. 
i.e. `a=b` as `a=REDACTED`, should we do the same for header as well? e.g: `Authorization:REDACTED`?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -115,7 +120,40 @@ public HttpLoggingPolicy(HttpLogDetailLevel detailLevel, boolean prettyPrintJSON
         return reqBodyLoggingMono;
     }
 
-    private Function<HttpResponse, Mono<HttpResponse>> logResponseDelegate(final ClientLogger logger, final URL url,
+    private void formatAllowableHeaders(Set<String> allowedHeaderNames, HttpRequest request, ClientLogger logger) {
```
anuchandy: @samvaity / @JonathanGiles does guideline proposes different meaning for `empty vs null` value for `allowedHeaderNames`? i.e. something like empty set means redact all headers and null means log all headers?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogDetailLevel.java
anuchandy: `BASIC` indicates log url (that includes query parameter) and skip headers & body. Wondering why we need `QUERY_PARAMS` enum option for a specific part of url, Does guideline enforce it?

samvaity: I don't think the guideline enforces to have the `QUERY_PARAMS` enum. It can be removed, but still, not sure if we need `BODY_AND_HEADERS` with it.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
samvaity: I think we should also add something like a "*", to say log all headers may be..

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
samvaity: yup, added the same thing with Headers.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
JonathanGiles: At present null and empty set should mean the same thing - to not print any headers.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogDetailLevel.java
JonathanGiles: Yes, remove `QUERY_PARAMS`. Leave the rest as-is I would think.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -89,8 +90,8 @@
      * The constructor with defaults.
      */
     public ConfigurationClientBuilder() {
-        httpLogDetailLevel = HttpLogDetailLevel.NONE;
         policies = new ArrayList<>();
+        httpLogOptions = new HttpLogOptions().setLogLevel(HttpLogDetailLevel.NONE);
```
JonathanGiles: It probably makes sense to just have the log level be NONE by default.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -207,14 +208,14 @@ public ConfigurationClientBuilder credential(ConfigurationClientCredentials cred
...
-     * @param logLevel The amount of logging output when sending and receiving HTTP requests/responses.
+     * @param logOptions The logging configuration to use when sending and receiving HTTP requests/responses.
      * @return The updated ConfigurationClientBuilder object.
-     * @throws NullPointerException If {@code logLevel} is {@code null}.
+     * @throws NullPointerException If {@code logOptions} is {@code null}.
      */
-    public ConfigurationClientBuilder httpLogDetailLevel(HttpLogDetailLevel logLevel) {
-        httpLogDetailLevel = Objects.requireNonNull(logLevel);
+    public ConfigurationClientBuilder httpLogOptions(HttpLogOptions logOptions) {
+        httpLogOptions = Objects.requireNonNull(logOptions, "Http log options cannot be null.");
```
JonathanGiles: We should probably allow null now that I think about it - this lets the user clear out the configuration in the builder so that the next time they call build, they will get an HttpLoggingPolicy with null options. This also means you need to have HttpLoggingPolicy know what to do if the options are null!

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Sets the level of detail to log on Http messages.
+     *
+     * @param logLevel The {@link HttpLogDetailLevel}.
+     * @return The updated HttpLogOptions object.
+     * @throws NullPointerException If {@code logLevel} is {@code null}.
+     */
+    public HttpLogOptions setLogLevel(final HttpLogDetailLevel logLevel) {
+        this.logLevel = Objects.requireNonNull(logLevel);
```
JonathanGiles: Maybe rather than require non-null, if null is passed in just set it to NONE. Be sure to document that though.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -26,28 +27,29 @@
...
-     * @param detailLevel The HTTP logging detail level.
-     * @param prettyPrintJSON If true, pretty prints JSON message bodies when logging.
-     *     If the detailLevel does not include body logging, this flag does nothing.
+     * @param httpLogOptions The HTTP logging configuration options.
+     * @param prettyPrintJSON If true, pretty prints JSON message bodies when logging. If the detailLevel does not
+     * include body logging, this flag does nothing.
      */
-    public HttpLoggingPolicy(HttpLogDetailLevel detailLevel, boolean prettyPrintJSON) {
-        this.detailLevel = detailLevel;
+    public HttpLoggingPolicy(HttpLogOptions httpLogOptions, boolean prettyPrintJSON) {
```
JonathanGiles: I'm curious if, outside of unit tests, anyone calls this constructor. If no one does, I would like to see this constructor made package-private for now, because this kind of boolean config should be in the HttpLogOptions object.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -69,19 +71,22 @@ public HttpLoggingPolicy(HttpLogDetailLevel detailLevel, boolean prettyPrintJSON
...
+
+        if (httpLogOptions.getLogLevel().shouldLogQueryParams()) {
+            formatAllowableQueryParams(httpLogOptions.getAllowedQueryParamNames(), request, logger);
         }
+
         //
         Mono<Void> reqBodyLoggingMono = Mono.empty();
         //
-        if (detailLevel.shouldLogBody()) {
+        if (httpLogOptions.getLogLevel().shouldLogBody()) {
```
JonathanGiles: Whenever I see a method call (like `getLogLevel()`) multiple times I prefer to see it referred to in a local variable.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -115,7 +120,46 @@ public HttpLoggingPolicy(HttpLogDetailLevel detailLevel, boolean prettyPrintJSON
...
+        StringBuilder sb = new StringBuilder();
+        for (HttpHeader header : request.getHeaders()) {
+            sb.append(header.getName()).append(":");
+            if (allowedHeaderNames.contains(header.getName())) {
+                sb.append(header.getValue());
+            } else {
+                sb.append(REDACTED_PLACEHOLDER);
+            }
+        }
+        logger.info(sb.toString());
```
JonathanGiles: Should there be a newline or whitespace included in this concatenation operation?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -128,18 +172,25 @@ public HttpLoggingPolicy(HttpLogDetailLevel detailLevel, boolean prettyPrintJSON
...
                 for (HttpHeader header : response.getHeaders()) {
-                    logger.info(header.toString());
+                    sb.append(header.getName()).append(":");
+                    if (httpLogOptions.getAllowedHeaderNames().contains(header.getName())) {
+                        sb.append(header.getValue());
+                    } else {
+                        sb.append(REDACTED_PLACEHOLDER);
+                    }
                 }
+                logger.info(sb.toString());
```
JonathanGiles: This header concatenation code should be centralised in this class rather than duplicated for request and response headers.

Issue #4509, opened 2019-07-22
Batch Delete


rickle-msft:
@rickle-msft commented on [Mon Jul 22 2019](https://github.com/Azure/azure-sdk-for-python/issues/6431)

rickle-msft:
service feature

PR #5693, opened 2019-10-04
Batching in Azure Storage Blobs

alzimmermsft:
Implements #4509, #4510, and #5161 

Adds batching into Azure Storage Blobs and implements the batching APIs for Blob delete and setTier.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
@@ -3,49 +3,45 @@
 
 package com.azure.storage.blob;
 
-import com.azure.core.annotation.ServiceClient;
 import com.azure.core.credentials.TokenCredential;
 import com.azure.core.http.HttpPipeline;
-import com.azure.core.http.rest.PagedFlux;
-import com.azure.core.http.rest.PagedResponse;
 import com.azure.core.http.rest.Response;
 import com.azure.core.http.rest.SimpleResponse;
-import com.azure.core.implementation.http.PagedResponseBase;
-import com.azure.core.implementation.util.FluxUtil;
+import com.azure.core.http.rest.VoidResponse;
 import com.azure.core.util.Context;
-import com.azure.core.util.logging.ClientLogger;
 import com.azure.storage.blob.implementation.AzureBlobStorageBuilder;
 import com.azure.storage.blob.implementation.AzureBlobStorageImpl;
-import com.azure.storage.blob.implementation.models.ServicesListContainersSegmentResponse;
-import com.azure.storage.blob.models.BlobContainerItem;
-import com.azure.storage.blob.models.CpkInfo;
+import com.azure.storage.blob.models.ContainerItem;
 import com.azure.storage.blob.models.KeyInfo;
-import com.azure.storage.blob.models.ListBlobContainersOptions;
+import com.azure.storage.blob.models.ListContainersOptions;
+import com.azure.storage.blob.models.Metadata;
 import com.azure.storage.blob.models.PublicAccessType;
+import com.azure.storage.blob.models.ServicesListContainersSegmentResponse;
 import com.azure.storage.blob.models.StorageAccountInfo;
 import com.azure.storage.blob.models.StorageServiceProperties;
 import com.azure.storage.blob.models.StorageServiceStats;
 import com.azure.storage.blob.models.UserDelegationKey;
+import com.azure.storage.common.IPRange;
+import com.azure.storage.common.SASProtocol;
 import com.azure.storage.common.Utility;
+import com.azure.storage.common.credentials.SharedKeyCredential;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-import java.time.Duration;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.time.OffsetDateTime;
-import java.util.Map;
-import java.util.function.Function;
 
-import static com.azure.core.implementation.util.FluxUtil.withContext;
-import static com.azure.storage.blob.implementation.PostProcessor.postProcessResponse;
+import static com.azure.storage.blob.PostProcessor.postProcessResponse;
 
 /**
- * Client to a storage account. It may only be instantiated through a {@link BlobServiceClientBuilder}. This class does
- * not hold any state about a particular storage account but is instead a convenient way of sending off appropriate
- * requests to the resource on the service. It may also be used to construct URLs to blobs and containers.
+ * Client to a storage account. It may only be instantiated through a {@link BlobServiceClientBuilder}. This class does not
+ * hold any state about a particular storage account but is instead a convenient way of sending off appropriate requests
+ * to the resource on the service. It may also be used to construct URLs to blobs and containers.
  *
  * <p>
- * This client contains operations on a blob. Operations on a container are available on {@link
- * BlobContainerAsyncClient} through {@link #getBlobContainerAsyncClient(String)}, and operations on a blob are
- * available on {@link BlobAsyncClient}.
+ * This client contains operations on a blob. Operations on a container are available on {@link ContainerAsyncClient}
+ * through {@link #getContainerAsyncClient(String)}, and operations on a blob are available on {@link BlobAsyncClient}.
  *
  * <p>
  * Please see <a href=https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction>here</a> for more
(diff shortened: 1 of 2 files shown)


pom.client.xml
```diff
@@ -693,8 +693,8 @@
             <version>${maven-compiler-plugin.version}</version>
             <!-- setup defaults for compile and testCompile -->
             <configuration>
-              <source>9</source>
-              <target>9</target>
```
JonathanGiles: For now lets revert this to keep focus on the batching

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
```diff
@@ -0,0 +1,413 @@
...
+import java.util.Locale;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
+
+/**
+ * This class provides
```
JonathanGiles: ...no javadoc

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
```diff
@@ -0,0 +1,413 @@
...
+            }
+
+            if (policy instanceof SharedKeyCredentialPolicy) {
+                batchHeadersPolicySet = true;
+                // The batch policy needs to be added before the SharedKey policy to run preparation cleanup.
+                policies.add(new BatchHeadersPolicy());
+            }
+
+            policies.add(pipeline.getPolicy(i));
+        }
```
JonathanGiles: This is slightly traumatising.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
```diff
@@ -0,0 +1,413 @@
...
+     *
+     * @param containerName The container of the blob.
+     * @param blobName The name of the blob.
+     * @return a {@link Response} that will be used to associate this operation to the response when the batch is
+     * submitted.
+     * @throws UnsupportedOperationException If this batch has already added an operation of another type.
+     */
+    public Response<Void> delete(String containerName, String blobName) {
+        return deleteHelper(buildClient(containerName, blobName), null, null);
+    }
```
JonathanGiles: If this adds an operation (but doesn't actually perform it), it seems weird that this returns `Response<Void>`?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
```diff
@@ -0,0 +1,413 @@
...
+     * @param blobName The name of the blob.
+     * @param deleteOptions Delete options for the blob and its snapshots.
+     * @param blobAccessConditions Additional access conditions that must be met to allow this operation.
+     * @return a {@link Response} that will be used to associate this operation to the response when the batch is
+     * submitted.
+     * @throws UnsupportedOperationException If this batch has already added an operation of another type.
+     */
+    public Response<Void> delete(String containerName, String blobName,
+        DeleteSnapshotsOptionType deleteOptions, BlobAccessConditions blobAccessConditions) {
+        return deleteHelper(buildClient(containerName, blobName), deleteOptions, blobAccessConditions);
```
JonathanGiles: Why is it necessary to build a client for each operation?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
```diff
@@ -0,0 +1,413 @@
...
+        }
+
+        while (!batchOperationQueue.isEmpty()) {
+            Mono<? extends Response> batchOperation = batchOperationQueue.pop();
+            Disposable disposable = batchOperation.subscribe();
+            while (!disposable.isDisposed()) {
+                // Wait until the batch operation has processed in the pipeline.
+                // This is used as opposed to block as it won't trigger an exception if ran in a Reactor thread.
+            }
+        }
```
JonathanGiles: Is it imperative that each operation complete before the next starts? Can you run them concurrently with a flatMap / collect style approach. The current code goes into a very tight while loop which isn't ideal.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -445,4 +459,135 @@ public String getAccountUrl() {
         return postProcessResponse(this.azureBlobStorage.services().getAccountInfoWithRestResponseAsync(context))
             .map(rb -> new SimpleResponse<>(rb, new StorageAccountInfo(rb.getDeserializedHeaders())));
     }
+
+    /**
+     * Submits a batch operation.
+     *
+     * <p>If any request in a batch fails this will throw a {@link StorageException}.</p>
+     *
+     * <p><strong>Code Samples</strong></p>
```
JonathanGiles: Need code samples

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -445,4 +459,135 @@ public String getAccountUrl() {
...
+     *
+     * @param batch Batch to submit.
+     * @return An empty response indicating that the batch operation has completed.
```
JonathanGiles: Document how the user gets the response information for each operation.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClient.java
```diff
@@ -353,4 +354,40 @@ public StorageAccountInfo getAccountInfo() {
 
         return Utility.blockWithOptionalTimeout(response, timeout);
     }
+
+    /**
+     * Submits a batch operation.
+     *
+     * <p>If any request in a batch fails this will throw a {@link StorageException}.</p>
+     *
+     * <p><strong>Code Samples</strong></p>
```
JonathanGiles: TODO

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
rickle-msft: I think because we need each request to go through a pipeline so it can get the data and auth stuff added, and for that we need a client per operation since each client can only point to one resource

alzimmermsft: That is correct, each request needs to be funneled through a pipeline to build up the request. In preview 5 we can look into making this implementation more flexible and efficient.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
alzimmermsft: The planned implementation is a little round about where the submission will return a response that becomes populated once the batch submission returns. Could change the response type to a `Future` to better indicate that this doesn't do anything yet and document that the batch must be submitted via `submitBatch` before the response will have a value.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
alzimmermsft: I think this could be simplified to just adding the batch policy before the credential policy.

rickle-msft: As long as it still gets the signing from the credentials but doesn't get retries

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
rickle-msft: I think Mono instead of Future would be preferable so we don't introduce a separate paradigm just for this. It might be ok to have this as an "async only" feature where sync customers who want to use it just have to block because I agree that there should be some sort of type-enforced means of communicating that the return value isn't valid until the batch completes.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
alzimmermsft: Remove the removal of policies, retry policies shouldn't trigger here given the response is `Mono.empty()`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
alzimmermsft: Could add a mutable policy that changes the URL of the request based on a `Context` value. That way there would be a single client and single policy to handle the URL switching.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobBatch.java
alzimmermsft: Will try out a Flux flatMapping that has a single tight loop instead of this numerous tight loop implementation.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -3,48 +3,45 @@
 
 package com.azure.storage.blob;
```
tg-msft: Is this autogenerated?  Why the massive diff delta?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/ServicesImpl.java
```diff
@@ -97,7 +98,7 @@ public ServicesImpl(AzureBlobStorageImpl client) {
         Mono<ServicesGetAccountInfoResponse> getAccountInfo(@HostParam("url") String url, @HeaderParam("x-ms-version") String version, @QueryParam("restype") String restype, @QueryParam("comp") String comp, Context context);
 
         @Post("")
-        @ExpectedResponses({200})
+        @ExpectedResponses({202})
```
tg-msft: Did you add a readme transform?  Is it in your PR?

sdk/core/azure-core/src/main/java/module-info.java
```diff
@@ -73,6 +73,7 @@
         com.azure.security.keyvault.secrets,    // FIXME this should not be a long-term solution
         com.azure.storage.common,               // FIXME this should not be a long-term solution
         com.azure.storage.blob,                 // FIXME this should not be a long-term solution
+        com.azure.storage.blob.batch,           // FIXME this should not be a long-term solution
```
joshfree: can you file a tracking issue for all of these FIXMEs

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+  <version>12.0.0-preview.4</version>
+</dependency>
+```
+
+### Default HTTP Client
+All client libraries, by default, use Netty HTTP client. Adding the above dependency will automatically configure 
+Storage Blob to use Netty HTTP client. 
+
+### Alternate HTTP client
+If, instead of Netty it is preferable to use OkHTTP, there is a HTTP client available for that too. Exclude the default
```
joshfree: nit: remove comma after If

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,113 @@
...
+     * {@codesnippet com.azure.storage.blob.BlobServiceAsyncClient.submitBatchWithResponse#BlobBatch-boolean}
+     *
+     * @param batch Batch to submit.
+     * @param throwOnAnyFailure Flag to indicate if an exception should be thrown if any request in the batch fails.
+     * @return A response only containing header and status code information, used to indicate that the batch operation
+     * has completed.
+     * @throws StorageException If {@code throwOnAnyFailure} is {@code true} and any request in the {@link BlobBatch}
+     * failed or the batch request is malformed.
+     */
+    public Mono<Response<Void>> submitBatchWithResponse(BlobBatch batch, boolean throwOnAnyFailure) {
```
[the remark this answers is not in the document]
joshfree: lets go on throwOnFailure for Preview 4

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
```diff
@@ -0,0 +1,123 @@
...
+    static Mono<Response<Void>> mapBatchResponse(BlobBatch batch, ServicesSubmitBatchResponse batchResponse,
+        boolean throwOnAnyFailure) {
+        /*
+         * Content-Type will contain the boundary for each batch response. The expected format is:
+         * "Content-Type: multipart/mixed; boundary=batchresponse_66925647-d0cb-4109-b6d3-28efe3e1e5ed"
+         */
+        String contentType = batchResponse.getDeserializedHeaders().getContentType();
+
+        // Split on the boundary [ "multipart/mixed; boundary", "batchresponse_66925647-d0cb-4109-b6d3-28efe3e1e5ed"]
+        String boundary = contentType.split("=", 2)[1];
```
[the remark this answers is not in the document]
joshfree: agreed

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/BlobServiceClientJavaDocCodeSnippets.java
```diff
@@ -240,4 +246,48 @@ public void getAccountInfoWithResponse() {
         StorageAccountInfo accountInfo = client.getAccountInfoWithResponse(timeout, context).getValue();
         // END: com.azure.storage.blob.BlobServiceClient.getAccountInfoWithResponse#Duration-Context
     }
+
```
joshfree: It looks like all of our batching tests are batching just 2 operations.  Can we test with an N greater than 2 as well?

alzimmermsft:
Closed in favor of #5734

PR #5696, opened 2019-10-04
5638-storage-getaccountname

hemanttanwar:
Expose account name  in all the clients.
The account name is first part of the host string. We can parse the endpoint or connection string and get account name.


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/LeaseAsyncClient.java
```diff
@@ -52,6 +52,7 @@
     private final boolean isBlob;
     private final String leaseId;
     private final AzureBlobStorageImpl client;
+    private final String accountName;
 
     LeaseAsyncClient(HttpPipeline pipeline, String url, String leaseId, boolean isBlob) {
```
alzimmermsft: Why not pass the accountName from the LeaseClientBuilder?

hemanttanwar: agree

PR #5698, opened 2019-10-04
Issues/5530 use consistent naming

### LINKED ISSUE #5530: Use consistent naming for methods/parameters.
There are several places where we inconsistently name things. Should update code to be consistent.
* Use `username` instead of `userName`
* Use Http, Url, etc, rather than HTTP, URL, etc
  * Create Checkstyle rule for this.
  * Example is: shouldLogURL -> shouldLogUrl

chrischild:
Created a new Checkstyle to look for specific words that should follow normal camelcasing standards.

Since it seems like there could be a growing list of words I added the ability to pass in a set of words that should be checked against.

Currently I checked in the checkstyle.xml that has an example of usage.

Since I did not have a set of concrete words I have not gone and changed any of the outstanding naming issues that were not fixed in [#5544](https://github.com/Azure/azure-sdk-for-java/pull/5544). If provided with a full list I can do that as well.

Fixes #5530


chrischild:
Removed the error message for missing disallowedWords property. It was a hack and didn't feel like it was a good solution. Changed to now have 2 default values to search for, URL and HTTP. This can still be overwritten by setting the disallowedWords property.

chrischild:
Tried to squash some commits and that didn't seem to work so reset to previous good commit.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,152 @@
...
+    private boolean applyToPublic = true;
+
+    /**
+     * Adds words that Classes, Methods and Variables that should follow Camelcasing standards
+     * @param disallowedWords words that should follow normal Camelcasing standards
+     */
+    public final void setDisallowedWords(String... disallowedWords) {
+        if (this.disallowedWords != null) {
+            Collections.addAll(this.disallowedWords, disallowedWords);
+        }
```
JonathanGiles: The `if` check here will never be null, as you're creating `disallowedWords` as a field, and that is what the `this.disallowedWords` refers to. I would suggest changing the code to something like this:

```java
    public final void setDisallowedWords(String... words) {
        if (words == null || words.length == 0) {
            this.disallowedWords.clear();
        }
        Collections.addAll(this.disallowedWords, words);
    }
```

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,152 @@
...
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class DisallowedWordsCheck extends AbstractCheck {
+    private final String[] DEFAULT_VALUES = new String[] { "URL", "HTTP" };
```
JonathanGiles: Perhaps remove this field and have it just configured through the XML config?

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml
```diff
@@ -378,6 +378,16 @@ page at http://checkstyle.sourceforge.net/config.html -->
...
     <module name="com.azure.tools.checkstyle.checks.GoodLoggingCheck"/>
 
+    <!-- CUSTOM CHECKS -->
+    <!-- Public API classes should follow Camelcase rules: -->
+    <!-- 1. ClassName -->
+    <!-- 2. methodName -->
+    <!-- 3. variableName -->
+    <!-- 4. longerVariableOrMethodName -->
+    <!--<module name="com.azure.tools.checkstyle.checks.DisallowedWordsCheck">-->
+    <!--<property name="disallowedWords" value="URL, HTTP, XML" />-->
```
JonathanGiles: URL, HTTP, XML, JSON, SAS, CPK

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,152 @@
...
+    private Set<String> disallowedWords = new HashSet<>(Arrays.asList(DEFAULT_VALUES));
+    private String errorMessage = "%s, All Public API Classes, Fields and Methods should follow " +
+        "Camelcase standards for the following words: %s.";
+    private boolean applyToPublic = true;
```
JonathanGiles: This field can be removed - we should only apply this rule to public API.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * Should we check member with given modifiers.
+     *
+     * @param token
+     *                modifiers of member to check.
+     * @return true if we should check such member.
+     */
+    private boolean shouldCheckInScope(DetailAST token) {
```
JonathanGiles: This might be better named 'isPublicAPI'. The method needs to check not just for PUBLIC modifier, but also for PROTECTED modifier, as either of them is considered public API that we need to guard against.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,152 @@
...
+        TokenTypes.METHOD_DEF};
+    }
+
+    @Override
+    public void visitToken(DetailAST token) {
+        switch (token.getType()) {
+            case TokenTypes.CLASS_DEF:
+            case TokenTypes.METHOD_DEF:
+            case TokenTypes.VARIABLE_DEF:
+                String tokenName = token.findFirstToken(TokenTypes.IDENT).getText();
```
JonathanGiles: This line can move inside the 'if' block.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,152 @@
...
+                }
+                beginIndex = -1;
+            }
+        }
+        // if abbreviation at the end of name (example: scaleX)
+        if (abbrStarted) {
+            final int endIndex = tokenName.length() - 1;
+            result = getAbbreviationIfIllegal(tokenName, beginIndex, endIndex);
+        }
+        return result;
```
JonathanGiles: This whole method could potentially be simplified to use the Java String `contains` method looping over the set of disallowed words?

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
chrischild: Yes I have this fixed locally I just have been waiting to fix anything that came from the review.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
chrischild: Well, really, it could be taken away. The only thing that could happen is that someone using the Checkstyle would need to look up how to use it and really that should be a given.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
chrischild: Good call. The code I used is from a checkstyle class but this is definitely a better solution.

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml
chrischild: So I have run the new checkstyle against the code and I probably fixed 100+ infractions against these words. In total I have 20ish files to commit. Did you want me to create a new branch for these changes or keep using this one?

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
chrischild: Ended up taking the property away since someone should look at the class if it doesn't work as they think it should. A default didn't really help or make sense.

chrischild:
Just wanted to check in.

chrischild:
Just popping in again and wanted to point out that I have created a new branch for all the files that were changed. I found that there were quite a few, over 100, and I think it would pollute this PR to check it into the same branch. The new branch is [here](https://github.com/Azure/azure-sdk-for-java/commit/7f9fec33109e745c8d2f985a337847bd52ab911c)

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,129 @@
...
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class DisallowedWordsCheck extends AbstractCheck {
+    private Set<String> disallowedWords = new HashSet<>(Arrays.asList());
```
conniey:
```suggestion
    private final Set<String> disallowedWords = new HashSet<>(); 
```

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,129 @@
...
+    private String errorMessage = "%s, All Public API Classes, Fields and Methods should follow " +
```
conniey: Would suggest using a constant for this:

private static final String ERROR_MESSAGE = "...."

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,129 @@
...
+        TokenTypes.VARIABLE_DEF,
+        TokenTypes.METHOD_DEF};
+    }
+
+    @Override
+    public void visitToken(DetailAST token) {
+        switch (token.getType()) {
+            case TokenTypes.CLASS_DEF:
+            case TokenTypes.METHOD_DEF:
+                if (isPublicAPI(token)) {
```
conniey: nit: isPublicAPI -> isPublicApi

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,129 @@
...
+                    String tokenName = token.findFirstToken(TokenTypes.IDENT).getText();
+                    boolean found = getDisallowedWords(tokenName);
```
conniey: Can avoid the variable initialisation by using:
`if (getDisallowedWords(tokenName)) {`

conniey: On that note, it's odd that this returns a Boolean when I expect some disallowed words? a better method name could be: hasDisallowedWords(tokenName)

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @param token
+     *                modifiers of member to check.
+     * @return true if we should check such member.
+     */
+    private boolean isPublicAPI(DetailAST token) {
+        final DetailAST modifiersAST =
+            token.findFirstToken(TokenTypes.MODIFIERS);
+        final boolean isStatic = modifiersAST.findFirstToken(TokenTypes.LITERAL_STATIC) != null;
+        final boolean isPublic = modifiersAST
```
conniey: How about using [TokenUtil.getScopeFromMods](https://checkstyle.org/apidocs/index.html?com/puppycrawl/tools/checkstyle/utils/TokenUtil.html).

```java
final Scope scope = TokenUtil.getScopeFromMods(token);
return scope == Scope.PUBLIC || scope == Scope.PROTECTED) && !isStatic;
```

eng/code-quality-reports/src/test/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheckTests.java
```diff
@@ -0,0 +1,62 @@
+package com.azure.tools.checkstyle.checks;
```
conniey: Add that header:
```xml
```

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml
```diff
@@ -378,6 +378,16 @@ page at http://checkstyle.sourceforge.net/config.html -->
...
 
+    <!-- CUSTOM CHECKS -->
+    <!-- Public API classes should follow Camelcase rules: -->
+    <!-- 1. ClassName -->
+    <!-- 2. methodName -->
+    <!-- 3. variableName -->
+    <!-- 4. longerVariableOrMethodName -->
+    <!--<module name="com.azure.tools.checkstyle.checks.DisallowedWordsCheck">-->
+    <!--<property name="disallowedWords" value="URL, HTTP, XML, JSON, SAS, CPK" />-->
+    <!--</module>-->
```
JonathanGiles: When you enable this, what issues do you see? If none, please do enable this rule!

chrischild: Will do.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
chrischild: No problem and I'll should I also add API to the list of words?

chrischild:
How do you feel about the naming I used? I also think BlacklistedWords might be better but either works.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
chrischild: Does not look like the Checkstyle version being used has that method.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
conniey: Yes, please. :D

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/DisallowedWordsCheck.java
conniey: Looks like it is in now: ScopeUtil.getScopeFromMods(token)

conniey:
@chrischild I like blacklist, it's concise. I won't lose sleep if you decide not to change it though.

chrischild:
Sorry I'm just fixing an issue with Git I found.

chrischild:
GitHub does not allow you to re-open a branch if it has been recreated apparently so I had to create a new PR [here](https://github.com/Azure/azure-sdk-for-java/pull/6054).

Apologies but I was committing some info that I didn't want to be.

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/BlacklistedWordsCheck.java
```diff
@@ -0,0 +1,108 @@
...
+            default:
+                // Checkstyle complains if there's no default block in switch
+                break;
+        }
+    }
+
+    /**
+     * Should we check member with given modifiers.
+     *
+     * @param token
```
mssfang: nit: is it format better if keeping
``` Java
@param token modifiers of the member to check.
```

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/BlacklistedWordsCheck.java
```diff
@@ -0,0 +1,108 @@
...
+        final boolean isStatic = modifiersAST.findFirstToken(TokenTypes.LITERAL_STATIC) != null;
+        final boolean isPublic = modifiersAST
+            .findFirstToken(TokenTypes.LITERAL_PUBLIC) != null;
+        final boolean isProtected = modifiersAST.findFirstToken(TokenTypes.LITERAL_PROTECTED) != null;
+        return (isPublic || isProtected) && !isStatic;
+    }
+
+    /**
+     * Gets the disallowed abbreviation contained in given String.
+     * @param tokenName
```
mssfang: same here. One line instead of two

PR #5702, opened 2019-10-04
Log more error details in azure-core

sima-zhu:
Need to continue working on serializer, jackson, util


sima-zhu:
In a wrong branch

PR #5703, opened 2019-10-04
Log more error details in azure core

sima-zhu:
Need to continue working on serializer, jackson, ~~util~~.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerMethodParser.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerMethodParser.java
@@ -33,7 +33,6 @@ import com.azure.core.implementation.serializer.HttpResponseDecodeData;
 import com.azure.core.implementation.serializer.SerializerAdapter;
 import com.azure.core.implementation.util.ImplUtils;
 import com.azure.core.implementation.util.TypeUtil;
-import com.azure.core.util.logging.ClientLogger;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 


sdk/core/azure-core/src/main/java/com/azure/core/credentials/BasicAuthenticationCredential.java
```diff
@@ -44,10 +44,11 @@ public BasicAuthenticationCredential(String username, String password) {
         String credential = username + ":" + password;
         String encodedCredential;
         try {
+            // Please change the error message everytime you update the named charset.
             encodedCredential = Base64Util.encodeToString(credential.getBytes("UTF8"));
         } catch (UnsupportedEncodingException e) {
             // The encoding is hard-coded, so if it's unsupported, it needs to be fixed right here.
-            throw logger.logExceptionAsError(new RuntimeException(e));
+            throw logger.logExceptionAsError(new RuntimeException("Please fixed the unsupported encoding UTF8.", e));
```
srnagar: Usually, the message in exceptions should inform what went wrong and shouldn't include corrective actions to fix the error. In this case, there's nothing the user could do to fix it as the encoding format is hardcoded in this method. It's better to revert this change and let the message in `UnsupportedEncodingException` be used by the logger which is more informative.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HostPolicy.java
```diff
@@ -39,7 +39,7 @@ public HostPolicy(String host) {
             context.getHttpRequest().setUrl(urlBuilder.setHost(host).toURL());
             result = next.process();
         } catch (MalformedURLException e) {
-            result = Mono.error(e);
+            result = Mono.error(new RuntimeException("Please check the host URL: " + urlBuilder.setHost(host), e));
```
srnagar: Same here. Error message shouldn't include corrective actions. "Host URL is invalid" would be better.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/PortPolicy.java
```diff
@@ -40,7 +40,8 @@ public PortPolicy(int port, boolean overwrite) {
             try {
                 context.getHttpRequest().setUrl(urlBuilder.setPort(port).toURL());
             } catch (MalformedURLException e) {
-                return Mono.error(e);
+                return Mono.error(new RuntimeException(
+                    String.format("Failed to set port %d to http request.", port), e));
```
srnagar: 👍 - this is good as the message is clear about what went wrong.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerMethodParser.java
```diff
@@ -53,6 +54,7 @@
  * method.
  */
 public class SwaggerMethodParser implements HttpResponseDecodeData {
+    private ClientLogger logger = new ClientLogger(SwaggerInterfaceParser.class);
```
srnagar: The argument to `ClientLogger` constructor should match the classname in which it's initialized.

```suggestion
    private ClientLogger logger = new ClientLogger(SwaggerMethodParser.class);
```

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -317,6 +317,7 @@ private boolean matchStatus(PollResponse<T> currentPollResponse, OperationStatus
             .onErrorResume(throwable -> {
                 // We should never get here and since we want to continue polling
                 //Log the error
+                logger.info("Failed to apply delay and call poll operation. Details: " + throwable.getMessage());
```
srnagar: Failure messages should at least be a warning.

sdk/core/azure-core/src/main/java/com/azure/core/credentials/BasicAuthenticationCredential.java
sima-zhu: Good point. Will reword that

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HostPolicy.java
```diff
@@ -39,7 +39,8 @@ public HostPolicy(String host) {
             context.getHttpRequest().setUrl(urlBuilder.setHost(host).toURL());
             result = next.process();
         } catch (MalformedURLException e) {
-            result = Mono.error(e);
+            result = Mono.error(new RuntimeException(String.format("Host URL %s is invalid ",
+                urlBuilder.setHost(host)), e));
```
conniey: Are you also invoking a method in here to set the host? Is this a desired side-effect?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HostPolicy.java
```diff
@@ -39,7 +39,8 @@ public HostPolicy(String host) {
             context.getHttpRequest().setUrl(urlBuilder.setHost(host).toURL());
             result = next.process();
         } catch (MalformedURLException e) {
-            result = Mono.error(e);
+            result = Mono.error(new RuntimeException(String.format("Host URL %s is invalid ",
```
conniey:
```suggestion
            result = Mono.error(new RuntimeException(String.format("Host URL '%s' is invalid.", 
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -176,11 +176,12 @@ private String prettyPrintIfNeeded(ClientLogger logger, String contentType, Stri
         return result;
     }
 
-    private long getContentLength(HttpHeaders headers) {
+    private long getContentLength(ClientLogger logger, HttpHeaders headers) {
         long contentLength = 0;
         try {
             contentLength = Long.parseLong(headers.getValue("content-length"));
         } catch (NumberFormatException | NullPointerException ignored) {
+            logger.warning("Http logging cannot parse the content-length: " + headers.getValue("content-length"));
```
conniey: I would use placeholders so that evaluation of headers.getValue isn't performed if the log level is not warning. In this case, 
```java
logger.warning("HTTP logging cannot parse the content-length: {}", headers.getValue("content-length"));
```

https://www.slf4j.org/manual.html#typical_usage

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -117,7 +120,8 @@
                                 bodyAsObject,
                                 decodedHeaders);
                         } catch (IllegalAccessException | InvocationTargetException | InstantiationException e) {
-                            throw logger.logExceptionAsError(Exceptions.propagate(e));
+                            throw logger.logExceptionAsError(new RuntimeException("Failed to deserialize 5-parameter"
+                                + "response with decoded headers. Error Details: " + e.getMessage(), e));
```
conniey:
```suggestion
                                + " response with decoded headers. Error Details: " + e.getMessage(), e));
```

conniey: The downside of doing string concatenations like this where it is split between multiple lines is the actual string may not be formatted (ie. missing spaces) correctly.

conniey: This suggestion also applies to the other messages above.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerInterfaceParser.java
```diff
@@ -18,6 +19,7 @@
  * interface.
  */
 public class SwaggerInterfaceParser {
+    private ClientLogger logger = new ClientLogger(SwaggerInterfaceParser.class);
```
conniey: private final.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerMethodParser.java
```diff
@@ -53,6 +54,7 @@
  * method.
  */
 public class SwaggerMethodParser implements HttpResponseDecodeData {
+    private ClientLogger logger = new ClientLogger(SwaggerMethodParser.class);
```
conniey: private final. Unless you plan on reassigning this somewhere else in the class.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -317,6 +317,7 @@ private boolean matchStatus(PollResponse<T> currentPollResponse, OperationStatus
             .onErrorResume(throwable -> {
                 // We should never get here and since we want to continue polling
                 //Log the error
+                logger.warning("Failed to apply delay and call poll operation. Details: " + throwable.getMessage());
```
conniey: Use slf4j placeholders.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/PortPolicy.java
alzimmermsft: Do we want to log these messages before sending them back through the reactive stream or will the be logged upstream?

alzimmermsft: Failed to set the HTTP request port to %d.

Thoughts on this? It feels a little more fluid in wording.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/ProtocolPolicy.java
```diff
@@ -40,7 +40,8 @@ public ProtocolPolicy(String protocol, boolean overwrite) {
             try {
                 context.getHttpRequest().setUrl(urlBuilder.setScheme(protocol).toURL());
             } catch (MalformedURLException e) {
-                return Mono.error(e);
+                return Mono.error(new RuntimeException(
+                    String.format("Failed to set protocol %s to http request.", protocol), e));
```
alzimmermsft: Similar comments for this and other policy error messages.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -65,7 +65,8 @@ public RetryPolicy(int maxRetries, Duration delayDuration) {
                     return attemptAsync(context, next, originalHttpRequest, tryCount + 1)
                         .delaySubscription(this.delayDuration);
                 } else {
-                    return Mono.error(err);
+                    return Mono.error(new RuntimeException("Max retries exceeded. Error Details: " + err.getMessage(),
```
alzimmermsft: Do we want to add how many retried were attempted?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -96,7 +97,8 @@
                         responseStatusCode,
                         responseHeaders));
                 } catch (IllegalAccessException | InvocationTargetException | InstantiationException e) {
-                    throw logger.logExceptionAsError(Exceptions.propagate(e));
+                    throw logger.logExceptionAsError(new RuntimeException("Failed to deserialize 3-parameter "
+                        + "response. Error Details: " + e.getMessage(), e));
```
alzimmermsft: Are we logging twice here when these fail? Once here and again for failing to deserialize response.

Additionally I'm not sure we should log these so specifically, it seems failing to decide response is sufficient.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -161,7 +161,8 @@ public Object invoke(Object proxy, final Method method, Object[] args) {
             }
 
         } catch (Exception e) {
-            throw logger.logExceptionAsError(Exceptions.propagate(e));
+            throw logger.logExceptionAsError(new RuntimeException(
+                "Error occurs during sending request and deserializing response. Error Details: "+ e.getMessage(), e));
```
alzimmermsft: Should use past tense for occurs, but do we want to log here as it seems it would be another generic log message after a specific log message. There might be times where we would want to ignore the logging rules.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerInterfaceParser.java
```diff
@@ -47,15 +49,16 @@ public SwaggerInterfaceParser(Class<?> swaggerInterface, SerializerAdapter seria
             if (hostAnnotation != null && !hostAnnotation.value().isEmpty()) {
                 this.host = hostAnnotation.value();
             } else {
-                throw new MissingRequiredAnnotationException(Host.class, swaggerInterface);
+                throw logger.logExceptionAsError(new MissingRequiredAnnotationException(Host.class, swaggerInterface));
```
alzimmermsft: These logs are more for us than a customer right? They wouldn't be able to do anything for this.

@JonathanGiles would we want debugging only logging?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerMethodParser.java
```diff
@@ -109,7 +111,8 @@
             requiredAnnotationOptions.add(Delete.class);
             requiredAnnotationOptions.add(Post.class);
             requiredAnnotationOptions.add(Patch.class);
-            throw new MissingRequiredAnnotationException(requiredAnnotationOptions, swaggerMethod);
+            throw logger.logExceptionAsError(new MissingRequiredAnnotationException(requiredAnnotationOptions,
```
alzimmermsft: Same comment as the other swagger class logging

alzimmermsft:
Overall I'm happy with the better logging just need to iron a few things.

sdk/core/azure-core/src/main/java/com/azure/core/credentials/BasicAuthenticationCredential.java
```diff
@@ -44,6 +44,7 @@ public BasicAuthenticationCredential(String username, String password) {
         String credential = username + ":" + password;
         String encodedCredential;
         try {
+            // Please change the error message everytime you update the named charset.
```
JonathanGiles: I don't understand what this message is saying to future developers. What error message? I would rather not have this kind of message left in code comments unless it is much clearer about what needs to happen.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HostPolicy.java
JonathanGiles: Yeah, this looks really odd. We shouldn't expect side effects from an exception.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
JonathanGiles: Yes

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerInterfaceParser.java
JonathanGiles: Yeah, customers should hopefully never see this, as it would mean we've shipped something that didn't work. I'm ok with it being an Error, so that we definitely see it during development.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HostPolicy.java
sima-zhu: Yes. Good catch. 
The right logging is String.format("Host URL %s is invalid ", host)

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerInterfaceParser.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerMethodParser.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
sima-zhu: Good catch. The error message show up in stack trace twice. Removed the one in each block. 

Having the parameter info is to help developer quickly determine which response constructor failed.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: Try to log some content out. Have trouble to find any. Revert it back.
@anuchandy please suggest if any.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerInterfaceParser.java
sima-zhu: Revert it back.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerMethodParser.java
sima-zhu: Revert it back.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/PortPolicy.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/ProtocolPolicy.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/PortPolicy.java
sima-zhu: It is actually a good place to log what the failed property looks like.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
conniey: This isn't resolved.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -65,7 +64,8 @@
                 try {
                     return (Constructor<? extends Response<?>>) constructor;
                 } catch (Throwable t) {
-                    throw logger.logExceptionAsError(new RuntimeException(t));
+                    throw logger.logExceptionAsError(new RuntimeException("Failed to cast the constructor to Response."
+                        + "Error Details: " + t.getMessage(), t));
```
conniey: Do you need `"Error Details:" + t.getMessage`? In previous log messages you didn't have this, and you've already passed the inner exception.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -128,7 +131,8 @@
                                 bodyAsObject,
                                 null));
                         } catch (IllegalAccessException | InvocationTargetException | InstantiationException e) {
-                            throw logger.logExceptionAsError(Exceptions.propagate(e));
+                            throw logger.logExceptionAsError(new RuntimeException("Failed to deserialize 5-parameter"
+                                + "response without decoded headers. ", e));
```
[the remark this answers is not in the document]
conniey: I noticed this because of the split, there isn't a space between "5-parameter" and "response". And same applies for all the other messages above. The split here doesn't hit the 120 character max either.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -48,6 +49,7 @@
     static Mono<Object> decode(HttpResponse httpResponse, SerializerAdapter serializer,
                                HttpResponseDecodeData decodeData) {
         ensureRequestSet(httpResponse);
+        final ClientLogger logger = new ClientLogger(HttpResponseBodyDecoder.class);
```
conniey: Creating a new instance every time could get expensive (if there was complex initialisation logic). But sadly, this is a static method/class.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -56,13 +58,13 @@
...
                                 decodeData.getUnexpectedException(httpResponse.getStatusCode()).getExceptionBodyType(),
-                                null,
-                                serializer,
-                                SerializerEncoding.fromHeaders(httpResponse.getHeaders()));
+                                null, serializer, SerializerEncoding.fromHeaders(httpResponse.getHeaders()));
                             return decodedErrorEntity == null ? Mono.empty() : Mono.just(decodedErrorEntity);
                         } catch (IOException | MalformedValueException ignored) {
                             // This translates in RestProxy as a RestException with no deserialized body.
                             // The response content will still be accessible via the .response() member.
+                            logger.warning("Failed to deserialize the error entity. Error details: "
```
conniey: You should pass the actual error in logger.warning as a parameter rather than manually printing out the message.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseHeaderDecoder.java
```diff
@@ -46,9 +46,7 @@
                     return Mono.justOrEmpty(deserializeHeaders(httpResponse.getHeaders(), serializer, decodeData));
                 } catch (IOException e) {
                     return Mono.error(new HttpResponseException(
-                        "HTTP response has malformed headers",
-                        httpResponse,
-                        e));
+                        "HTTP response has malformed headers", httpResponse, e));
```
conniey:
```suggestion
                        "HTTP response has malformed headers.", httpResponse, e));
```

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/jackson/JacksonAdapter.java
```diff
@@ -117,12 +117,14 @@ public String serialize(Object object, SerializerEncoding encoding) throws IOExc
 
     @Override
     public String serializeRaw(Object object) {
+        final ClientLogger logger = new ClientLogger(JacksonAdapter.class);
```
conniey: Why are we creating a new one here instead of having a private final one in the class?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/jackson/JacksonAdapter.java
```diff
@@ -117,12 +117,14 @@ public String serialize(Object object, SerializerEncoding encoding) throws IOExc
...
         if (object == null) {
             return null;
         }
         try {
             return serialize(object, SerializerEncoding.JSON).replaceAll("^\"*", "").replaceAll("\"*$", "");
         } catch (IOException ex) {
+            logger.warning("Failed to serialize {} to JSON.", object.getClass());
```
conniey: Log the exception as the last parameter in this.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/util/FluxUtil.java
```diff
@@ -61,7 +61,8 @@ private static void accept(ByteArrayOutputStream byteOutputStream, ByteBuffer by
         try {
             byteOutputStream.write(byteBufferToArray(byteBuffer));
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            throw new RuntimeException("Error occurs when write ByteBuffer to ByteArrayOutputStream. "
+                + "Error details: " + e.getMessage(), e);
```
conniey: Should not have to print these error details. It'll be duplicated in the inner throwable. Same with one below,.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -317,6 +317,7 @@ private boolean matchStatus(PollResponse<T> currentPollResponse, OperationStatus
             .onErrorResume(throwable -> {
                 // We should never get here and since we want to continue polling
                 //Log the error
+                logger.warning("Failed to apply delay and call poll operation. Details: {}" + throwable.getMessage());
```
conniey: Instead of logging the message manually, pass the exception as a parameter. Our logger should be exposing these details.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/jackson/JacksonAdapter.java
sima-zhu: Removed

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/jackson/JacksonAdapter.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/util/FluxUtil.java
sima-zhu: If not logged out the error message, then it lost the IOException details.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
sima-zhu: I want to expose the detail message from inner exception to upper layer. Otherwise it will show up a little under. Hard to trace.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -63,8 +63,7 @@
                         } catch (IOException | MalformedValueException ignored) {
                             // This translates in RestProxy as a RestException with no deserialized body.
                             // The response content will still be accessible via the .response() member.
-                            logger.warning("Failed to deserialize the error entity. Error details: "
-                                + ignored.getMessage());
+                            logger.warning("Failed to deserialize the error entity. ", ignored);
```
conniey:
```suggestion
                            logger.warning("Failed to deserialize the error entity.", ignored);
````

File: sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java [hunk 30]
```diff
@@ -63,8 +63,7 @@
                         } catch (IOException | MalformedValueException ignored) {
```
conniey: The name is misleading because this is not actually ignored. Should be error or ex or e, etc.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -317,7 +317,7 @@ private boolean matchStatus(PollResponse<T> currentPollResponse, OperationStatus
             .onErrorResume(throwable -> {
                 // We should never get here and since we want to continue polling
                 //Log the error
-                logger.warning("Failed to apply delay and call poll operation. Details: {}" + throwable.getMessage());
+                logger.warning("Failed to apply delay and call poll operation. ", throwable);
```
conniey:
```suggestion
                logger.warning("Failed to apply delay and call poll operation.", throwable); 
```

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/BasicExample.java
```diff
@@ -27,102 +27,118 @@
      * @throws RuntimeException If the downloaded data doesn't match the uploaded data
      */
     public static void main(String[] args) throws IOException {
+//
```
conniey: Why is this commented out?

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/BasicExample.java
```diff
@@ -27,102 +27,118 @@
...
-        /*
-         * Create more blobs before listing.
-         */
-        for (int i = 0; i < 3; i++) {
-            String sampleData = "Samples";
-            InputStream dataInBlobs = new ByteArrayInputStream(sampleData.getBytes(Charset.defaultCharset()));
-            blobContainerClient.getBlobClient("myblobsforlisting" + System.currentTimeMillis()).getBlockBlobClient()
-                .upload(dataInBlobs, sampleData.length());
-            dataInBlobs.close();
+    private static Throwable b() {
```
conniey: b? a?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
sima-zhu: After a second thought. It might be useful to revert it back to origin one.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/util/FluxUtil.java
conniey: It's a problem with our ClientLogger, not this logic.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
conniey: It's a problem with our ClientLogger that should be resolved. Unless we explicitly say in the spec that we should be logging Throwable.getMessage().

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/BasicExample.java
sima-zhu: That's something unnecessary. Revert.

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/BasicExample.java
sima-zhu: Unnecessary change. Revert

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
sima-zhu: Missed it. Checked in my changes.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -221,11 +221,13 @@ private String prettyPrintIfNeeded(ClientLogger logger, String contentType, Stri
...
     }
 
-    private long getContentLength(HttpHeaders headers) {
+    private long getContentLength(ClientLogger logger, HttpHeaders headers) {
         long contentLength = 0;
         try {
             contentLength = Long.parseLong(headers.getValue("content-length"));
         } catch (NumberFormatException | NullPointerException ignored) {
+            logger.warning("Http logging cannot parse the content-length {}."
+                + headers.getValue("content-length"));
```
conniey: Should be passed as a parameter. And pass the 'ignored' exception as the last parameter. Though I would change its name from ignored.

```java
logger.warning("Could not parse the HTTP header content-length: '{}'.",
    headers.getValue("content-length", ignored));
```

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
```diff
@@ -96,7 +95,8 @@
                         responseStatusCode,
                         responseHeaders));
                 } catch (IllegalAccessException | InvocationTargetException | InstantiationException e) {
-                    throw logger.logExceptionAsError(Exceptions.propagate(e));
+                    throw logger.logExceptionAsError(new RuntimeException("Failed to deserialize 3-parameter"
```
conniey: You're changing the behaviour here by not using Exceptions.propagate(e). Same with the others below.

```java
throw logger.logExceptionAsError(Exceptions.propagate(
    new RuntimeException("Failed to deserialize 3-parameter response", e)));
```

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
```diff
@@ -159,8 +159,7 @@ public Object invoke(Object proxy, final Method method, Object[] args) {
                 return handleHttpResponse(request, asyncDecodedResponse, methodParser, methodParser.getReturnType(),
                     context);
             }
-
-        } catch (Exception e) {
+        } catch (IOException e) {
```
conniey: You're limiting the exception caught here to IOException from Exception. Is this what you meant to do?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerMethodParser.java
```diff
@@ -53,6 +54,7 @@
  * method.
  */
 public class SwaggerMethodParser implements HttpResponseDecodeData {
+    private final ClientLogger logger = new ClientLogger(SwaggerMethodParser.class);
```
conniey: There doesn't appear to be any use of this logger. Is it needed.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/util/FluxUtil.java
```diff
@@ -61,7 +61,7 @@ private static void accept(ByteArrayOutputStream byteOutputStream, ByteBuffer by
         try {
             byteOutputStream.write(byteBufferToArray(byteBuffer));
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            throw new RuntimeException("Error occurs when write ByteBuffer to ByteArrayOutputStream.", e);
```
conniey:
```suggestion
            throw new RuntimeException("Error occurred writing ByteBuffer to ByteArrayOutputStream.", e);
```

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -361,6 +361,7 @@ private boolean matchStatus(PollResponse<T> currentPollResponse, OperationStatus
             .onErrorResume(throwable -> {
                 // We should never get here and since we want to continue polling
                 //Log the error
```
conniey: You can remove this line //Log the error

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -399,7 +402,8 @@ private static Type extractEntityTypeFromReturnType(HttpResponseDecodeData decod
                     // Get Type of 'T' from OperationStatus<T>
                     token = TypeUtil.getTypeArgument(token);
                 }
-            } catch (Exception ignored) {
+            } catch (ClassNotFoundException ignored) {
+                logger.warning("Failed to find class 'com.azure.core.management.implementation.OperationStatus'.");
```
srnagar: This should log and throw. ClassNotFoundException means the necessary libraries are not loaded at runtime. Instead of logging and ignoring it, we should propagate this exception upwards.

sima-zhu: @anuchandy Could you give me some insights on whether we need to throw exception or silence pass when the method encountered   ClassNotFoundException ?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerMethodParser.java
sima-zhu: Removed.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/RestProxy.java
sima-zhu: The error should catch as specific as possible. Based on the code, only unchecked IOException should catch right here. All other exception should handle inside of the methods. Did not see the point of catching general Exception.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ResponseConstructorsCache.java
sima-zhu: Trying to log the paramCount for debugging purpose.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
sima-zhu: Done.

PR #5704, opened 2019-10-04
Add remove to HttpHeaders

alzimmermsft:
Adds a method to remove headers from `HttpHeaders`.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/HttpHeaders.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/HttpHeaders.java
@@ -7,13 +7,14 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/HttpHeaders.java
```diff
@@ -79,22 +80,33 @@ public HttpHeader get(String name) {
...
+     * Removes the {@link HttpHeader header} with the provided header name. {@code Null} is returned if the header
+     * isn't found.
      *
-     * @param name the name of the header to look for
-     * @return The String value of the header, or null if the header isn't found
+     * @param name the name of the header to remove.
+     * @return the header if removed, null otherwise.
+     */
+    public HttpHeader remove(String name) {
+        return headers.remove(formatKey(name));
```
srnagar: Change `headers` map to use `ConcurrentHashMap`.

alzimmermsft: Done

Issue #3706, opened 2019-05-20
Improve documentation of Context


alzimmermsft:
Context needs documentation that explains how it is able to be used.

joshfree:
@samvaity PTAL

samvaity:
Closed in https://github.com/Azure/azure-sdk-for-java/pull/5705

PR #5705, opened 2019-10-04
Documentation and code snippet update for Context

samvaity:
Reference #3706


sdk/core/azure-core/src/samples/java/com/azure/core/util/ContextJavaDocCodeSnippets.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Code snippets for {@link Context}
+ */
+public class ContextJavaDocCodeSnippets {
+
+    /**
+     * Code snippet for {@link Context#Context(Object, Object)}
+     */
+    public void constructContextObject() {
```
alzimmermsft: I'm not sure if we have any well known `Context` values that get passed through other than tracing, they would be good to use in these samples as they demonstrate real world scenarios.

sdk/core/azure-core/src/main/java/com/azure/core/util/Context.java
```diff
@@ -13,7 +13,9 @@
 /**
  * {@code Context} offers a means of passing arbitrary data (key-value pairs) to pipeline policies.
  * Most applications do not need to pass arbitrary data to the pipeline and can pass {@code Context.NONE} or
- * {@code null}. Each context object is immutable. The {@code addData(Object, Object)} method creates a new
+ * {@code null}.
+ * <p>
+ * Each context object is immutable. The {@code addData(Object, Object)} method creates a new
```
conniey: You can use `{@link #addData(Object, Object)}`. If the method gets renamed, this reference also gets renamed.

sdk/core/azure-core/src/main/java/com/azure/core/util/Context.java
```diff
@@ -34,8 +36,12 @@
     /**
      * Constructs a new {@link Context} object.
      *
-     * @param key the key
-     * @param value the value
+     * <p><strong>Code samples</strong></p>
+     *
+     * {@codesnippet com.azure.core.util.context#object-object}
+     *
+     * @param key the given key with which the specified value should be associated
```
conniey:
```suggestion
     * @param key The key with which the specified value should be associated.
```

sdk/core/azure-core/src/main/java/com/azure/core/util/Context.java
```diff
@@ -34,8 +36,12 @@
...
+     * @param value the given value to be associated with the specified key
```
conniey:
```suggestion
     * @param value The value to be associated with the specified key.
```

conniey: Same with instances below.

sdk/core/azure-core/src/samples/java/com/azure/core/util/ContextJavaDocCodeSnippets.java
```diff
@@ -0,0 +1,84 @@
...
+        // BEGIN: com.azure.core.util.context.getData#object
+        final String key1 = "Key1";
+        final String value1 = "first-value";
+
+        // Create a context object with given key and value
+        Context context = new Context(key1, value1);
+
+        // Look for the specified key in the returned context object
+        Optional<Object> optionalObject = context.getData(key1);
+        System.out.printf("Key1 value : %s%n", optionalObject.get().toString());
```
conniey: I would make use of optional in case key1's value doesn't exist.
```java
if (optionalObject.isPresent()) {
    System.out.printf("Key1 value: %s%n", optionalObject.get());
} else {
    System.out.println("Key1 does not exist or have data.");
}
```

sdk/core/azure-core/src/samples/java/com/azure/core/util/ContextJavaDocCodeSnippets.java
```diff
@@ -0,0 +1,84 @@
...
+    public void contextOfObject() {
+        // BEGIN: com.azure.core.util.context.of#map
+        final String key1 = "Key1";
+        final String value1 = "first-value";
+        Map<Object, Object> keyValueMap = new HashMap<>();
+        keyValueMap.put(key1, value1);
+
+        // Create a context using the provided key value pair map
+        Context keyValueContext = Context.of(keyValueMap);
+        System.out.printf("Key1 value %s%n", keyValueContext.getData(key1).get().toString());
```
conniey: The .toString() isn't necessary, the `%s` formatter will call .toString() on it.

sdk/core/azure-core/src/samples/java/com/azure/core/util/ContextJavaDocCodeSnippets.java
samvaity: I added tracing relevant Context object uses in a few places. Since these were code snippets and outside of the tracing package could not have a complete scenario-based snippet.

PR #5354, opened 2019-09-12
Add Ink Recognition SDK

minaltolpadi:
This introduces an SDK to access the Ink Recognizer service which recognizes handwritten content in applications.

PR #5708, opened 2019-10-06
Add Cognitive Service - Ink Recognizer Track 1 SDK

mssfang:
Previously, we moved Ink Recognizer into azure-sdk-for-java in PR #5354.

When this PR merged, it will be able to see all spotbugs and checkstyle errors.

@minaltolpadi  You can start to fix all spotbugs and checkstyle errors introduced when you ran 

mvn install -f sdk\cognitiveservices\ms-azure-cs-inkrecognizer\pom.xml -DskipTests -Dgpg.skip

after this PR is merged.

### DIFF
--- a/sdk/cognitiveservices/ms-azure-cs-inkrecognizer/pom.xml
+++ b/sdk/cognitiveservices/ms-azure-cs-inkrecognizer/pom.xml
@@ -29,32 +29,26 @@
         <dependency>
             <groupId>com.squareup.okhttp3</groupId>
             <artifactId>okhttp</artifactId>
-            <version>4.1.0</version>
         </dependency>
         <dependency>
             <groupId>io.projectreactor</groupId>
             <artifactId>reactor-core</artifactId>
-            <version>3.2.3.RELEASE</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.9.9</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
-            <version>2.9.9</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
-            <version>2.9.9</version>
         </dependency>
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
@@ -64,7 +58,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.1</version>
                 <configuration>
                     <source>1.8</source>
                     <target>1.8</target>


pom.data.xml
```diff
@@ -284,6 +284,10 @@
                 <title>Azure Storage</title>
                 <packages>com.microsoft.azure.storage*</packages>
               </group>
+              <group>
```
srnagar: Update the version of this pom file as this module is newly introduced which warrants a new version. Customers using the current version shouldn't automatically see this inkrecognizer module.

pom.data.xml
```diff
@@ -548,6 +552,7 @@
   <modules>
     <module>./sdk/batch/microsoft-azure-batch</module>
     <module>./sdk/applicationinsights/microsoft-azure-applicationinsights-query</module>
+    <module>./sdk/cognitiveservices/ms-azure-cs-inkrecognizer/pom.xml</module>
```
srnagar: The module name is inconsistent with the rest. All other modules start with `microsoft-`. This starts with `ms-`.

sdk/cognitiveservices/ms-azure-cs-inkrecognizer/pom.xml
```diff
@@ -1,59 +1,54 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT License. -->
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+    <artifactId>azure-cognitiveservices-inkrecognizer</artifactId>
+    <groupId>com.microsoft.azure.cognitiveservices</groupId>
+    <version>1.1.0-preview.1</version>
```
srnagar: This is a new module. Why is the version `1.1.0` instead of `1.0.0`?

pom.data.xml
weshaggard: @srnagar I don't believe we need to bump the version for this pom for dev-time plugin configurations. It is only necessary for changes that are needed for runtime changes to the parents, for example the version properties for the dependencies.

sdk/cognitiveservices/ms-azure-cs-inkrecognizer/pom.xml
weshaggard: Yes this should be `1.0.0-preview.1` if it is a new module.

pom.data.xml
weshaggard: For some things like cogservice we abbreviated them because of long file path issues. So I'd suggest sticking with the "ms-" for this one and be consistent with the rest of cogservices directory.

srnagar: I don't think length of filepath is an issue because the applicationinsights module above this line has 67 chars and inkrecognizer with `microsoft` is only 56 chars. Also, the longest module name in cogservices is `custom-vision-prediction` which is 66 chars (1 less than applicationinsights).

`microsoft-azure-applicationinsights-query`
`microsoft-azure-cs-inkrecognizer`
`microsoft-azure-cs-customvision-prediction`

weshaggard: We abbreviated both ms and cs in all those directory names. We can certainly decide to alter that but it shouldn't be part of this PR. For this PR it should remain consistent with the other cogservices libraries directories.

PR #5709, opened 2019-10-06
Fix proxy creation and proxy tests

### LINKED ISSUE #5694: Fix SimpleProxy in Event Hubs proxy tests
SimpleProxy is negotiating the proxy handler twice, so it is resulting in a bunch of garbage.

conniey:
* Fixes bug in proxy connection where the proxy connect request was being made to the proxy rather than to the AMQP messaging service (ie. in WebSocketsProxyConnectionHandler).  This resulted in a connect request to itself, causing the proxy integration tests to fail.
* Re-write SimpleProxy into multiple classes to make it easier to maintain.

Fixes #5694

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/proxy/ProxyReceiveTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/proxy/ProxyReceiveTest.java
@@ -16,7 +16,6 @@ import com.azure.messaging.eventhubs.models.EventHubProducerOptions;
 import com.azure.messaging.eventhubs.models.EventPosition;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
 import org.junit.Rule;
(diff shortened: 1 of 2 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/jproxy/ConnectionProperties.java
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+package com.azure.messaging.eventhubs.jproxy;
```
JonathanGiles: Why is this test package called `jproxy`?

conniey: This is pulled from Track 1 where they had pulled in a simple jproxy written by one of the devs. I don't have any preferences if you want to name it something else.
https://github.com/SreeramGarlapati/jproxy

Issue #5485, opened 2019-09-24
[BUG] Azure Storage SDK12 inputStream buffer handling is incorrect


alexander-riss:
**Describe the bug**
The com.azure.storage.common.Utility convertStreamToByteBuffer method expects to always be able to read the full size of the requested buffer. This can not be expected from an InputStream and should not be confused with an EOF signal

***Code Snippet***
com.azure.storage.common.Utility:512

**Expected behavior**
Reading less than the expected buffer size must be handled by the code and not be treated like an error. An inputStream might still be in the process of receiving data and block until more becomes available (before finally hitting the EOF).

amishra-dev:
@gapra-msft can you please take a look.

alexander-riss:
sample stack trace when e.g. reading data from an http client to a blob object:
```
com.azure.core.exception.UnexpectedLengthException: Request body emitted 7435 bytes less than the expected 114077 bytes.
	at com.azure.storage.common.Utility.lambda$null$3(Utility.java:514)
	at reactor.core.publisher.MonoCallable.call(MonoCallable.java:91)
	at reactor.core.publisher.FluxConcatMap$ConcatMapImmediate.drain(FluxConcatMap.java:402)
	at reactor.core.publisher.FluxConcatMap$ConcatMapImmediate.onNext(FluxConcatMap.java:244)
	at reactor.core.publisher.FluxMapFuseable$MapFuseableSubscriber.onNext(FluxMapFuseable.java:121)
	at reactor.core.publisher.FluxJust$WeakScalarSubscription.request(FluxJust.java:99)
	at reactor.core.publisher.FluxMapFuseable$MapFuseableSubscriber.request(FluxMapFuseable.java:162)
	at reactor.core.publisher.FluxConcatMap$ConcatMapImmediate.onSubscribe(FluxConcatMap.java:229)
	at reactor.core.publisher.FluxMapFuseable$MapFuseableSubscriber.onSubscribe(FluxMapFuseable.java:90)
	at reactor.core.publisher.FluxJust.subscribe(FluxJust.java:70)
	at reactor.core.publisher.FluxMapFuseable.subscribe(FluxMapFuseable.java:63)
	at reactor.core.publisher.FluxConcatMap.subscribe(FluxConcatMap.java:121)
	at reactor.core.publisher.FluxPeek.subscribe(FluxPeek.java:83)
	at reactor.core.publisher.Flux.subscribe(Flux.java:7921)
[...]

PR #5716, opened 2019-10-08
Throw exception until hit the EOF

sima-zhu:
https://github.com/Azure/azure-sdk-for-java/issues/5485

### DIFF
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
@@ -550,7 +550,7 @@ public final class Utility {
                     throw Exceptions.propagate(ioe);
                 }
             })
-            .takeUntil(p -> p.readBytes() == -1)
+            .takeUntil(p -> p.readBytes() == -1 || currentTotalLength[0] > length)
             .filter(p -> p.readBytes() > 0)
             .flatMap(p -> {
                 if (currentTotalLength[0] < length) {


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
```diff
@@ -538,12 +538,12 @@ public static SharedKeyCredential getSharedKeyCredential(HttpPipeline httpPipeli
                 long count = pos + blockSize > length ? length - pos : blockSize;
                 byte[] cache = new byte[(int) count];
                 int lastIndex = data.read(cache);
-                currentTotalLength[0] += lastIndex;
-                if (lastIndex < count) {
+                if (lastIndex == -1 && currentTotalLength[0] < length) {
```
alzimmermsft: This is getting closer to an implementation that would handle a stream that is buffering slowly but there is an issue that this will have random chunks of data missing. With the current implementation given a scenario where I have 5 100 byte chunks I could have this happen:

Read 100
Read 100
Read 10 (slow buffering)
Read 10 (slow buffering)
Read 100
Throw UnexpectedLengthException, expected 500 got 320

But there could still be more data loading into the stream.

Instead there should be a looping mechanism on each concatMap emissions where if I read that 10 I try to read more until one of the following scenarios

1. I have filled by current buffer, wrap this and trigger the next concatMap emission
2. I have hit end of file, wrap this and trigger on complete
3. I have hit end of file and I got less than expected, trigger an error

Now the scenario should look like

Read 100
Read 100,
Read 10 (slow buffering), read 10 (slow buffering), read 80
Read 100
Read 100
Everything is good here, send the data to the service.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
```diff
@@ -531,33 +532,61 @@ public static SharedKeyCredential getSharedKeyCredential(HttpPipeline httpPipeli
...
-                byte[] cache = new byte[(int) count];
-                int lastIndex = data.read(cache);
-                currentTotalLength[0] += lastIndex;
-                if (lastIndex < count) {
+        return Flux.just(true)
+            .repeat()
+            .map(ignore -> {
+                byte[] buffer = new byte[blockSize];
+                try {
+                    int numBytes = data.read(buffer);
```
anuchandy: Now that we've repeat() and `length` I guess we can simplify this a bit and avoid extra allocation of `Pair` type & avoid extra filter , onComplete and flatmap. We could also move validations to within this map. I didn't test below code but this is something in my mind.

```java
public static Flux<ByteBuffer> convertStreamToByteBuffer(InputStream data, long length, int blockSize) {
    final long[] currentTotalLength = new long[1];
    return Flux.just(true)
        .repeat()
        .map(ignore -> {
             byte[] buffer = new byte[blockSize];
             int totalBytesRead = 0;
             while (totalBytesRead < blockSize) {
                 int numBytesRead = data.read(buffer, totalBytesRead, blockSize - totalBytesRead);
                 if (numBytesRead == -1)  {
                     break;
                 } else {
                    totalBytesRead += numBytesRead; 
                 }
             }
             if (totalBytesRead == 0) {
                 // If totalBytesRead is zero after the above loop then means either we reached EOF
                 // or blockSize is zero
                 if (currentTotalLength[0] < length) {
                     throw "Stream has less than provided length!"
                 } else {
                    // Emit an empty ByteBuffer that get filtered out by below takeUntil.
                    return ByteBuffer.wrap(buffer, 0, 0);
                 }
             } else {
                 currentTotalLength[0] += totalBytesRead;
                 if (currentTotalLength[0] > length) {
                     throw "Stream has more than provided length!"
                 } else {
                    return ByteBuffer.wrap(buffer, 0, totalBytesRead);
                 }
             }
         })
         .takeUntil(bb -> !bb.hasRemaining());
```

sima-zhu: https://github.com/Azure/azure-sdk-for-java/issues/5754

An issue to track the simplification.

PR #5719, opened 2019-10-08
Added config option to provide preference list for Connection policy

shettyh:
This will help to benchmark the cross-region read/write latencies


shettyh:
Closing the MR as it seems like `PreferenceList` is not supported in latest libraries. Changed according to this https://github.com/Azure/azure-cosmosdb-java/tree/master/benchmark/src/main/java/com/microsoft/azure/cosmosdb/benchmark

PR #5724, opened 2019-10-08
Preview 4: Final API Review, Update java doc and sample

mssfang:
Final API changes based on the feedback given on the API review Tool
Update javadoc, code samples
Update JSON record

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientCredentials.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientCredentials.java
@@ -65,7 +65,7 @@ class ConfigurationClientCredentials {
     /**
      * Gets the base URI of the Azure App Configuration instance based on the provided connection string.
      *
-     * @return The string value of the base URI of the configuration service extracted from connection string provided.
+     * @return The base url of the configuration service extracted from connection string provided.
      */
     String getBaseUri() {
         return this.credentials.baseUri().toString();


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -311,10 +307,18 @@ private URL getBuildEndpoint(ConfigurationClientCredentials buildCredentials) {
...
             return endpoint;
         } else if (buildCredentials != null) {
-            return buildCredentials.getBaseUri();
+            return getURLFromString(buildCredentials.getAppConfigurationURL());
         } else {
             return null;
         }
     }
+
+    private URL getURLFromString(String endpoint) {
```
alzimmermsft: Instead of adding this you should also change the constructor for `ConfigurationAsyncClient` to take a `String` instead of URL, all it is doing is calling `toString` on the passed `URL`. Kind of pointless to add all this exception wrapping and conversion handling.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/credentials/ConfigurationClientCredentials.java
```diff
@@ -67,10 +67,10 @@ public ConfigurationClientCredentials(String connectionString)
     /**
      * Gets the base URI of the Azure App Configuration instance based on the provided connection string.
      *
-     * @return The base URI of the configuration service extracted from connection string provided.
+     * @return The String of base URI of the configuration service extracted from connection string provided.
      */
-    public URL getBaseUri() {
-        return this.credentials.baseUri();
+    public String getAppConfigurationURL() {
```
alzimmermsft:
```suggestion
    public String getAppConfigurationUrl() {
```

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/credentials/ConfigurationClientCredentials.java
```diff
@@ -67,10 +67,10 @@ public ConfigurationClientCredentials(String connectionString)
     /**
      * Gets the base URI of the Azure App Configuration instance based on the provided connection string.
      *
-     * @return The base URI of the configuration service extracted from connection string provided.
+     * @return The String of base URI of the configuration service extracted from connection string provided.
```
alzimmermsft:
```suggestion
     * @return The base url of the configuration service extracted from connection string provided.
```

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/credentials/ConfigurationClientCredentials.java
alzimmermsft: Does this method need to be `public` if we are going to move this class into that same package as the builder and clients?

Did you ping @annelo-msft about this change?

PR #5734, opened 2019-10-08
Batching in Azure Storage Blobs

alzimmermsft:
Implements #4509, #4510, and #5161

Adds batching into Azure Storage Blobs and implements the batching APIs for Blob delete and setTier.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
+++ b/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
@@ -542,7 +542,10 @@
   </Match>
   <!-- It is fine to have un-used variables, unread fields, anonymous static inner classes in javadoc code samples. -->
   <Match>
-    <Class name="~.*JavaDoc(CodeSnippets|CodeSamples)"/>
+    <Or>
+      <Class name="~.*JavaDoc(CodeSnippets|CodeSamples)"/>
+      <Class name="com.azure.storage.blob.batch.ReadmeCodeSamples"/>
+    </Or>
     <Bug pattern="DLS_DEAD_LOCAL_STORE,
                   URF_UNREAD_FIELD,
                   SIC_INNER_SHOULD_BE_STATIC_ANON,
--- a/sdk/storage/azure-storage-blob-batch/src/samples/java/com/azure/storage/blob/batch/ReadmeCodeSamples.java
+++ b/sdk/storage/azure-storage-blob-batch/src/samples/java/com/azure/storage/blob/batch/ReadmeCodeSamples.java
@@ -5,24 +5,26 @@ package com.azure.storage.blob.batch;
 
 import com.azure.core.http.rest.Response;
 import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.BlobServiceClientBuilder;
 import com.azure.storage.blob.models.AccessTier;
 import com.azure.storage.blob.models.BlobAccessConditions;
 import com.azure.storage.blob.models.DeleteSnapshotsOptionType;
 import com.azure.storage.blob.models.LeaseAccessConditions;
 
+import java.util.Arrays;
 import java.util.List;
 
 /**
  * Code samples for the READEME.md
  */
 public class ReadmeCodeSamples {
-    private BlobServiceClient blobServiceClient;
-    private BlobBatchClient blobBatchClient;
-    private List<String> blobUrls;
-    private String blobUrl;
-    private String blobUrl2;
-    private String blobUrlWithSnapshot;
-    private String blobUrlWithLease;
+    private BlobServiceClient blobServiceClient = new BlobServiceClientBuilder().buildClient();
+    private BlobBatchClient blobBatchClient = new BlobBatchClientBuilder(blobServiceClient).buildClient();
+    private String blobUrl = "https://account.core.windows.net/containerName/blobName";
+    private String blobUrl2 = "https://account.core.windows.net/containerName/blobName2";
+    private String blobUrlWithSnapshot = "https://account.core.windows.net/containerName/blobName?snapshot=<DateTime>";
+    private String blobUrlWithLease = "https://account.core.windows.net/containerName/blobNameWithLease";
+    private List<String> blobUrls = Arrays.asList(blobUrl, blobUrl2, blobUrlWithSnapshot, blobUrlWithLease);
 
     public void creatingBlobBatchClient() {
         BlobBatchClient blobBatchClient = new BlobBatchClientBuilder(blobServiceClient).buildClient();


sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+<dependency>
+  <groupId>com.azure</groupId>
+  <artifactId>azure-storage-blob</artifactId>
+  <version>12.0.0-preview.4</version>
+</dependency>
+```
+
+### Default HTTP Client
+All client libraries, by default, use Netty HTTP client. Adding the above dependency will automatically configure 
+Storage Blob to use Netty HTTP client.
```
srnagar: Storage Blob Batch

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+### Default HTTP Client
+All client libraries, by default, use Netty HTTP client. Adding the above dependency will automatically configure 
+Storage Blob to use Netty HTTP client. 
+
+### Alternate HTTP client
+If, instead of Netty it is preferable to use OkHTTP, there is a HTTP client available for that too. Exclude the default
+Netty and include OkHTTP client in your pom.xml.
+
+```xml
+<!-- Add Storage Blob dependency without Netty HTTP client -->
```
srnagar: Storage Blob Batch

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+<dependency>
+    <groupId>com.azure</groupId>
+    <artifactId>azure-storage-blob</artifactId>
```
srnagar:
```suggestion
    <artifactId>azure-storage-blob-batch</artifactId>
```

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+-  Java Development Kit (JDK) with version 8 or above
+- [Azure Subscription][azure_subscription]
+- [Create Storage Account][storage_account]
+
+### Adding the package to your product
+
+```xml
+<dependency>
+  <groupId>com.azure</groupId>
+  <artifactId>azure-storage-blob</artifactId>
```
srnagar:
```suggestion
  <artifactId>azure-storage-blob-batch</artifactId>
```

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param accessTier {@link AccessTier} to set on the blobs.
+     * @param blobUrls Blobs to set the access tier.
+     * @return
+     * @throws StorageException If the batch is malformed or any of the set access tier operations fail.
+     */
+    public Flux<Response<Void>> setBlobsAccessTier(AccessTier accessTier, String... blobUrls) {
+        return batchingHelper((batch, blobUrl) -> batch.setTier(blobUrl, accessTier), blobUrls);
+    }
+
+    private <T> Flux<Response<T>> batchingHelper(BiFunction<BlobBatch, String, Response<T>> generator,
```
rickle-msft: Can you leave some docs here?

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
+# Azure Storage Blobs client library for Java
```
srnagar: Batch

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+
+Azure Blob storage is Microsoft's object storage solution for the cloud. Blob
```
srnagar: This description should be updated too.

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+    <artifactId>azure-storage-blob</artifactId>
+    <exclusions>
+      <exclusion>
+        <groupId>com.azure</groupId>
+        <artifactId>azure-core-http-netty</artifactId>
+      </exclusion>
+    </exclusions>
+</dependency>
+
+<!-- Add OkHTTP client to use with Storage Blob -->
```
srnagar: Storage Blob Batch

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+
+```java
+BlockBlobClient blockBlobClient = containerClient.getBlobClient("myblockblob").getBlockBlobClient();
+String dataSample = "samples";
+try (ByteArrayInputStream dataStream = new ByteArrayInputStream(dataSample.getBytes())) {
+    blockBlobClient.upload(dataStream, dataSample.length());
+}
+```
+
+### Uploading a blob from `File`
```
srnagar: Need to update README with examples for using the batch APIs.

sdk/storage/azure-storage-blob-batch/pom.xml
```diff
@@ -0,0 +1,278 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>com.azure</groupId>
+    <artifactId>azure-client-sdk-parent</artifactId>
+    <version>1.4.0</version>
```
srnagar: This should be 1.5.0 since you incremented the version of `pom.client.xml`.

sdk/storage/azure-storage-blob-batch/pom.xml
```diff
@@ -0,0 +1,278 @@
...
+        <enabled>false</enabled>
+      </snapshots>
+    </pluginRepository>
+  </pluginRepositories>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core</artifactId>
+      <version>1.0.0-preview.5</version>
```
srnagar: This may have to be core preview 6 if we bump the version up before another release.

sdk/storage/azure-storage-blob-batch/pom.xml
```diff
@@ -0,0 +1,278 @@
...
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-test</artifactId>
+      <version>1.0.0-preview.5</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-http-netty</artifactId>
+      <version>1.0.0-preview.5</version>
+      <scope>test</scope>
```
srnagar: Netty should be a default HTTP client making this an opinionated client library. Users can exclude this, if required. So, the scope shouldn't be `test` here.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    /**
+     * Constructs a {@link BlobBatch} using the {@link BlobServiceAsyncClient#getHttpPipeline()
+     * BlobServiceAsyncClient's HttpPipeline} to build a modified {@link HttpPipeline} that is used to prepare the
+     * requests in the batch.
+     *
+     * @param client {@link BlobServiceAsyncClient} used to construct the batch.
+     */
+    public BlobBatch(BlobServiceAsyncClient client) {
```
srnagar: I am not sure how users will interact with this but from looking at this, there's a batch that can be created using sync or async client but all the APIs here are returning sync types.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,445 @@
...
+     * @param blobName The name of the blob.
+     * @param accessTier The tier to set on the blob.
+     * @param leaseAccessConditions Lease access conditions that must be met to allow this operation.
+     * @return a {@link Response} that will be used to associate this operation to the response when the batch is
+     * submitted.
+     * @throws UnsupportedOperationException If this batch has already added an operation of another type.
+     */
+    public Response<Void> setTier(String containerName, String blobName, AccessTier accessTier,
+        LeaseAccessConditions leaseAccessConditions) {
+        return setTierHelper(String.format("%s/%s", containerName, blobName), accessTier, leaseAccessConditions);
```
srnagar: String.format is very expensive in terms of performance. I would not recommend using it in a high-volume code path (maybe okay for exceptions) but it's better to use StringBuilder or just plain string concat (which gets converted to StringBuilder by the compiler internally).

https://kylewbanks.com/blog/java-string-concatenation-vs-stringbuilder-vs-string-format-performance
https://stackoverflow.com/questions/513600/should-i-use-javas-string-format-if-performance-is-important

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,113 @@
...
+import reactor.core.publisher.Mono;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.BiFunction;
+
+import static com.azure.core.implementation.util.FluxUtil.withContext;
+import static com.azure.storage.blob.implementation.PostProcessor.postProcessResponse;
+
+public final class BlobBatchAsyncClient {
```
srnagar: Javadoc and annotations.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,113 @@
...
+     *
+     * <p><strong>Code samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.BlobServiceAsyncClient.submitBatch#BlobBatch}
+     *
+     * @param batch Batch to submit.
+     * @return An empty response indicating that the batch operation has completed.
+     * @throws StorageException If any request in the {@link BlobBatch} failed or the batch request is malformed.
+     */
+    public Mono<Void> submitBatch(BlobBatch batch) {
```
srnagar: Missing annotations.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,113 @@
...
+        return postProcessResponse(this.client.services().submitBatchWithRestResponseAsync(
+            batch.getBody(), batch.getContentLength(), batch.getContentType(), context))
+            .flatMap(response -> BlobBatchHelper.mapBatchResponse(batch, response, throwOnAnyFailure));
+    }
+
+    /**
+     * Deletes the passed blobs.
+     *
+     * @param blobUrls Blobs to delete.
+     * @return
```
srnagar: Missing return javadoc

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,113 @@
...
+    public Flux<Response<Void>> deleteBlobs(String... blobUrls) {
+        return batchingHelper(BlobBatch::delete, blobUrls);
+    }
+
+    /**
+     * Sets the {@link AccessTier} on the passed blobs.
+     *
+     * @param accessTier {@link AccessTier} to set on the blobs.
+     * @param blobUrls Blobs to set the access tier.
+     * @return
```
srnagar: Same here

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchClient.java
```diff
@@ -0,0 +1,71 @@
...
+import com.azure.core.http.rest.Response;
+import com.azure.core.util.Context;
+import com.azure.storage.blob.models.AccessTier;
+import com.azure.storage.blob.models.StorageException;
+import com.azure.storage.common.Utility;
+
+import java.time.Duration;
+import java.util.stream.Stream;
+
+public final class BlobBatchClient {
```
srnagar: Javadoc and annotations.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchClientBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.azure.core.http.HttpPipeline;
+import com.azure.core.util.logging.ClientLogger;
+import com.azure.storage.blob.BlobServiceAsyncClient;
+import com.azure.storage.blob.BlobServiceClient;
+
+/**
+ *
+ */
+public final class BlobBatchClientBuilder {
```
srnagar: Add javadocs

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchClientBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+    public BlobBatchClientBuilder(BlobServiceClient client) {
+        this.accountUrl = client.getAccountUrl();
+        this.pipeline = client.getHttpPipeline();
+    }
+
+    /**
+     *
+     * @param client
+     */
+    public BlobBatchClientBuilder(BlobServiceAsyncClient client) {
```
srnagar: Shouldn't the builder be instantiated with a default constructor and the client should be a setter?

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/package-info.java
```diff
@@ -0,0 +1 @@
+package com.azure.storage.blob.batch;
```
srnagar: Add javadoc and copyright license.

alzimmermsft: Added

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
alzimmermsft: Added comments

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,418 @@
...
+    private static final String CONTENT_ID = "Content-Id";
+    private static final String BATCH_BOUNDARY_TEMPLATE = "batch_%s";
+    private static final String REQUEST_CONTENT_TYPE_TEMPLATE = "multipart/mixed; boundary=%s";
+    private static final String BATCH_OPERATION_CONTENT_TYPE = "Content-Type: application/http";
+    private static final String BATCH_OPERATION_CONTENT_TRANSFER_ENCODING = "Content-Transfer-Encoding: binary";
+    private static final String BATCH_OPERATION_CONTENT_ID_TEMPLATE = "Content-ID: %d";
+    private static final String HTTP_VERSION = "HTTP/1.1";
+    private static final String OPERATION_TEMPLATE = "%s %s %s";
+    private static final String HEADER_TEMPLATE = "%s: %s";
+    private static final String NEWLINE = "\r\n";
```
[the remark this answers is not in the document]
tg-msft: We don't want to do that - this is an HTTP newline that needs to be platform independent.  It can never change from `\r\n`.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchClient.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    public Response<Void> submitBatchWithResponse(BlobBatch batch, boolean throwOnAnyFailure, Duration timeout,
+        Context context) {
+        return Utility.blockWithOptionalTimeout(client.submitBatchWithResponse(batch, throwOnAnyFailure, context),
+            timeout);
+    }
+
+    public Stream<Response<Void>> deleteBlobs(String... blobUrls) {
+        return client.deleteBlobs(blobUrls).toStream();
+    }
```
rickle-msft: docs. and on setBlobsAccessTier

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,113 @@
...
+     * @throws StorageException If the batch is malformed or any of the set access tier operations fail.
+     */
+    public Flux<Response<Void>> setBlobsAccessTier(AccessTier accessTier, String... blobUrls) {
```
rickle-msft: >(AccessTier accessTie [](start = 50, length = 21)

Possible alternative is accepting a list of tiers and a parallel list of blobUrls to make it possible for each blob to get a specific tier. I don't feel strongly about this though.

sdk/storage/azure-storage-blob-batch/src/samples/java/com/azure/storage/blob/batch/BlobBatchJavaDocCodeSnippets.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Code snippet for {@link BlobBatch#setTier(String, AccessTier, LeaseAccessConditions)}
+     */
+    public void addSetTierWithUrl() {
+        // BEGIN: com.azure.storage.blob.BlobBatch.setTier#String-AccessTier-LeaseAccessConditions
+        Response<Void> setTierResponse = batch.setTier("{url of blob}", AccessTier.HOT,
+            new LeaseAccessConditions().setLeaseId("{lease ID}"));
+        // END: com.azure.storage.blob.BlobBatch.setTier#String-AccessTier-LeaseAccessConditions
+    }
+}
```
rickle-msft: Can we also have some sort of samples file that shows an end-to-end sample of how to construct a batch, submit it, and then process the responses once they have been filled?

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,418 @@
...
+            batchPipelineBuilder.policies(pipeline.getPolicy(i));
+        }
+
+        if (!batchHeadersPolicySet) {
+            batchPipelineBuilder.policies(this::cleanseHeaders);
+        }
+
+        this.batchClient = new BlobClientBuilder()
+            .endpoint(accountUrl)
+            .blobName("")
```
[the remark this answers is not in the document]
alzimmermsft: Given there is a `HttpPipelinePolicy` handling configuration of the blob name and container name in the URL there needs to be something to instantiate the blob client with.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,418 @@
...
+            }
+
+            batchPipelineBuilder.policies(pipeline.getPolicy(i));
+        }
+
+        if (!batchHeadersPolicySet) {
+            batchPipelineBuilder.policies(this::cleanseHeaders);
+        }
+
+        this.batchClient = new BlobClientBuilder()
```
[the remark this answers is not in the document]
alzimmermsft: Done

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
```diff
@@ -0,0 +1,132 @@
...
+     * This pattern matches finding the "Content-Id" of the batch response.
+     */
+    private static final Pattern CONTENT_ID_PATTERN = Pattern
+        .compile("Content-ID:\\s?(\\d+)", Pattern.CASE_INSENSITIVE);
+
+    /*
+     * This pattern matches finding the status code of the batch response.
+     */
+    private static final Pattern STATUS_CODE_PATTERN = Pattern
+        .compile("HTTP\\/\\d\\.\\d\\s?(\\d+)\\s?\\w+", Pattern.CASE_INSENSITIVE);
```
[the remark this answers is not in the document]
alzimmermsft: The other option is splitting the line on `" "` and taking the second value, both are highly dependent on what this service guarantees to return.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
```diff
@@ -0,0 +1,132 @@
...
+
+                // Split the batch response body into batch operation responses.
+                for (String subResponse : body.split("--" + boundary)) {
+                    // This is a split value that isn't a response.
+                    if (!subResponse.contains("application/http")) {
+                        continue;
+                    }
+
+                    // The batch operation response will be delimited by two new lines.
+                    String[] subResponseSections = subResponse.split("\r\n\r\n");
```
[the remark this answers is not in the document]
alzimmermsft: Same as Ted's other comment this it HTTP specific and should be "\r\n" but I will make a static value for it.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,445 @@
...
+    private static final String OPERATION_TEMPLATE = "%s %s %s";
+    private static final String HEADER_TEMPLATE = "%s: %s";
+    private static final String NEWLINE = "\r\n";
+
+    private static final int[] EXPECTED_DELETE_STATUS_CODES = { 202 };
+    private static final int[] EXPECTED_SET_TIER_STATUS_CODES = { 200, 202 };
+
+    private final ClientLogger logger = new ClientLogger(BlobBatch.class);
+
+    private final BlobAsyncClient batchClient;
```
rickle-msft: >batchClient [](start = 34, length = 11)

Can you rename this to blobClient? Calling it batchClient feels misleading (and makes reading some of the code below confusing)

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,445 @@
...
+        }
+
+        Disposable disposable = Flux.fromStream(batchOperationQueue.stream())
+            .flatMap(batchOperation -> batchOperation)
+            .subscribe();
+
+        while (!disposable.isDisposed()) {
+            // Wait until the batch operation has processed in the pipeline.
+            // This is used as opposed to block as it won't trigger an exception if ran in a Reactor thread.
+        }
```
rickle-msft: Think I left a comment on the other PR, but this I think can just be .blockLast or equivalent

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,113 @@
...
+     * has completed.
+     * @throws StorageException If {@code throwOnAnyFailure} is {@code true} and any request in the {@link BlobBatch}
+     * failed or the batch request is malformed.
+     */
+    public Mono<Response<Void>> submitBatchWithResponse(BlobBatch batch, boolean throwOnAnyFailure) {
+        return withContext(context -> submitBatchWithResponse(batch, throwOnAnyFailure, context));
+    }
+
+    Mono<Response<Void>> submitBatchWithResponse(BlobBatch batch, boolean throwOnAnyFailure, Context context) {
+        return postProcessResponse(this.client.services().submitBatchWithRestResponseAsync(
```
rickle-msft: >postProcessResponse [](start = 15, length = 19)

This, I think, is in the process of getting removed

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchOperationResponse.java
```diff
@@ -0,0 +1,157 @@
...
+            }
+
+            @Override
+            public Mono<String> getBodyAsString() {
+                return Mono.just(body);
+            }
+
+            @Override
+            public Mono<String> getBodyAsString(Charset charset) {
+                return Mono.just(body);
```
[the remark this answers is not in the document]
alzimmermsft: Using it.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: It could but I went with this approach as it is safer for use inside Reactor threads, they will throw an exception by default if a blocking call is made in them.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
alzimmermsft: This will happen in preview 5 as it is a non-API breaking change and that PR can roll this up.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: Done

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
alzimmermsft: For now I am going to move forward with the single access tier and list of blob urls to stay aligned with .NET, in the future if feedback requests that functionality it'll be an easy add.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchClient.java
alzimmermsft: Done

sdk/storage/azure-storage-blob-batch/src/samples/java/com/azure/storage/blob/batch/BlobBatchJavaDocCodeSnippets.java
alzimmermsft: I'll add a non-Javadoc snippets sample file that will go through an end-to-end scenario.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,445 @@
...
+     * {@codesnippet com.azure.storage.blob.BlobBatch.setTier#String-AccessTier-LeaseAccessConditions}
+     *
+     * @param blobUrl URI of the blob.
+     * @param accessTier The tier to set on the blob.
+     * @param leaseAccessConditions Lease access conditions that must be met to allow this operation.
+     * @return a {@link Response} that will be used to associate this operation to the response when the batch is
+     * submitted.
+     * @throws UnsupportedOperationException If this batch has already added an operation of another type.
+     */
+    public Response<Void> setTier(String blobUrl, AccessTier accessTier, LeaseAccessConditions leaseAccessConditions) {
```
rickle-msft: >Response<Void> [](start = 11, length = 14)

I think I also left this on the other PR, but Mono<Response> makes more sense to me here as the operation isn't completed until the batch is submitted. Mono.create should give you a way to fill in the value of the mono once the batch has been submitted and processed.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,445 @@
...
+        context.getHttpRequest().setHeader(CONTENT_ID, context.getData(BATCH_REQUEST_CONTENT_ID).get().toString());
+
+        // Set the request URL to the correct endpoint.
+        try {
+            UrlBuilder requestUrl = UrlBuilder.parse(context.getHttpRequest().getUrl());
+            requestUrl.setPath(context.getData(BATCH_REQUEST_URL_PATH).get().toString());
+            context.getHttpRequest().setUrl(requestUrl.toURL());
+        } catch (MalformedURLException ex) {
+            throw Exceptions.propagate(logger.logExceptionAsError(new IllegalStateException(ex)));
+        }
```
rickle-msft: This doesn't belong in a cleanseHeaders method. Can you either rename or refactor?

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,445 @@
...
+                batchHeadersPolicySet = true;
+                // The batch policy needs to be added before the SharedKey policy to run preparation cleanup.
+                batchPipelineBuilder.policies(this::cleanseHeaders);
+            }
+
+            batchPipelineBuilder.policies(pipeline.getPolicy(i));
+        }
+
+        if (!batchHeadersPolicySet) {
+            batchPipelineBuilder.policies(this::cleanseHeaders);
```
rickle-msft: It doesn't look like you ever call build? Perhaps line 140 is supposed to pass batchPipelineBuilder.build() instead of pipeline?

alzimmermsft: Fixed

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: Made a new private method name `setRequestUrl`

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
rickle-msft: One additional thought about why Mono is preferred is because otherwise you can't know for sure when the batch has finished and will be stuck polling. Your comment about the error from assertResponseReceived in the BatchResponse (or whatever that check is called) is recoverable by actually submitting the batch, but then there's no way of knowing when it is safe to access the response after that. Just because you call submit batch, doesn't mean the response will be immediately filled and ok to access. 

---
In reply to: [333123167](https://github.com/Azure/azure-sdk-for-java/pull/5734#discussion_r333123167) [](ancestors = 333123167)

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: Removed constructors for this class, need to get the `BlobBatch` from the client.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: This is like any other method that does string manipulation, it should only trigger once for each method call and shouldn't normally be called in a tight loop.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
alzimmermsft: Added

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
alzimmermsft: Added

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
alzimmermsft: Added

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
alzimmermsft: Added

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: This can be changed in preview 5 if we find this pattern to be bad

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,445 @@
...
+        }
+
+        return next.process();
+    }
+
+    /*
+     * This will "send" the batch operation request when triggered, it simply acts as a way to build and write the
+     * batch operation into the overall request and then returns nothing as the response.
+     */
+    private Mono<HttpResponse> setupBatchOperation(HttpRequest request) {
```
rickle-msft: This naming is also a little confusing as this also serves as the "HttpClient". It may be sufficient to include that in the comments, but it'd be helpful to have that information somewhere to help tie the pieces together.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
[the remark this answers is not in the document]
alzimmermsft: Luckily this is an implementation detail and we can rev this in preview 5.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
```diff
@@ -0,0 +1,132 @@
...
+    private static final Pattern CONTENT_ID_PATTERN = Pattern
+        .compile("Content-ID:\\s?(\\d+)", Pattern.CASE_INSENSITIVE);
+
+    /*
+     * This pattern matches finding the status code of the batch response.
+     */
+    private static final Pattern STATUS_CODE_PATTERN = Pattern
+        .compile("HTTP\\/\\d\\.\\d\\s?(\\d+)\\s?\\w+", Pattern.CASE_INSENSITIVE);
+
+    private static final ClientLogger LOGGER = new ClientLogger(BlobBatchHelper.class);
```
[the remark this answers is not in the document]
alzimmermsft: Changed to pass through a logger from the calling class.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchClientBuilder.java
alzimmermsft: This is following what `LeaseClientBuilder` currently does, we can switch this up a little in preview 5 if need be

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
+# Azure Storage Blobs Batch client library for Java
+
+Azure Blob storage is Microsoft's object storage solution for the cloud. Blob
+storage is optimized for storing massive amounts of unstructured data.
+Unstructured data is data that does not adhere to a particular data model or
+definition, such as text or binary data.
```
srnagar: Add another line here to describe the purpose of this module specifically. Something like "This client library provides APIs to perform bulk/batch operations on Azure Blob storage."

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,263 @@
...
+     *
+     * <p><strong>Code samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.batch.BlobBatchAsyncClient.submitBatch#BlobBatch-boolean}
+     *
+     * @param batch Batch to submit.
+     * @param throwOnAnyFailure Flag to indicate if an exception should be thrown if any request in the batch fails.
+     * @return A response only containing header and status code information, used to indicate that the batch operation
+     * has completed.
+     * @throws StorageException If {@code throwOnAnyFailure} is {@code true} and any request in the {@link BlobBatch}
```
srnagar: This should not actually throw an exception. Instead, the error should be propagated through the error channel of the `Mono` . Same for all the APIs here that return a Publisher (Mono/Flux).

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -0,0 +1,263 @@
...
+            }
+
+            @Override
+            public void close() {
+            }
+        };
+    }
+
+
+//    private <T> PagedFlux<Response<T>> batchingHelper(List<String> blobUrls,
```
srnagar: Delete instead of commenting out the code.

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+```
+
+### Downloading a blob to local path
+
+Download blob to local file using BlobClient.
+```java
+blobClient.downloadToFile("downloaded-file.jpg");
+```
+
+### Enumerating blobs
```
srnagar: Add samples for using the batch client.

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+[samples_storage_error]: src/samples/java/com/azure/storage/blob/StorageErrorHandlingExample.java
+[samples_list_containers]: src/samples/java/com/azure/storage/blob/ListContainersExample.java
+[samples_metadata]: src/samples/java/com/azure/storage/blob/SetMetadataAndHTTPHeadersExample.java
+[samples_identity]: src/samples/java/com/azure/storage/blob/AzureIdentityExample.java
+[cla]: https://cla.microsoft.com
+[coc]: https://opensource.microsoft.com/codeofconduct/
+[coc_faq]: https://opensource.microsoft.com/codeofconduct/faq/
+[coc_contact]: mailto:opencode@microsoft.com
+
+![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-java/sdk/storage/azure-storage-blob/README.png)
```
srnagar:
```suggestion
![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-java/sdk/storage/azure-storage-blob-batch/README.png)
```

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+[storage_account]: https://docs.microsoft.com/azure/storage/common/storage-quickstart-create-account?tabs=azure-portal
+[storage_account_create_ps]: https://docs.microsoft.com/azure/storage/common/storage-quickstart-create-account?tabs=azure-powershell
+[storage_account_create_cli]: https://docs.microsoft.com/azure/storage/common/storage-quickstart-create-account?tabs=azure-cli
+[storage_account_create_portal]: https://docs.microsoft.com/azure/storage/common/storage-quickstart-create-account?tabs=azure-portal
+[azure_cli]: https://docs.microsoft.com/cli/azure
+[azure_sub]: https://azure.microsoft.com/free/
+[identity]: https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/identity/azure-identity/README.md
+[error_codes]: https://docs.microsoft.com/rest/api/storageservices/blob-service-error-codes
+[samples]: src/samples
+[samples_basic]: src/samples/java/com/azure/storage/blob/BasicExample.java
```
srnagar: Will this path work? This is a different directory for batch.

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,330 @@
...
+
+## Contributing
+
+This project welcomes contributions and suggestions. Most contributions require you to agree to a Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us the rights to use your contribution. For details, visit https://cla.microsoft.com.
+
+When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repos using our CLA.
+
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the Code of Conduct FAQ or contact opencode@microsoft.com with any additional questions or comments.
+
+![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-java%2Fsdk%2Fstorage%2FAzure.Storage.Blobs%2FREADME.png)
```
srnagar:
```suggestion
![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-java%2Fsdk%2Fstorage%2FAzure.Storage.Blobs.Batch%2FREADME.png)
```

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
rickle-msft: XD you sneaky mom!

---
In reply to: [333121225](https://github.com/Azure/azure-sdk-for-java/pull/5734#discussion_r333121225) [](ancestors = 333121225)

sdk/storage/azure-storage-blob-batch/README.md
alzimmermsft: Delete this line in favor of the impression at the bottom of the file, should make sure if both are needed later.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
```diff
@@ -0,0 +1,137 @@
...
+
+        int contentId;
+        if (contentIdMatcher.find()) {
+            contentId = Integer.parseInt(contentIdMatcher.group(1));
+        } else {
+            throw logger.logExceptionAsError(
+                new IllegalStateException("Batch operation response doesn't contain a 'Content-Id' header."));
+        }
+
+        return batch.getBatchRequest(contentId).setResponseReceived();
```
rickle-msft: >.setResponseReceived() [](start = 47, length = 22)

This should be called after setBodyOrPotentiallyThrow to ensure that all the fields are filled in before it is marked as valid for reading.

sdk/storage/azure-storage-blob-batch/README.md
alzimmermsft: Removed

sdk/storage/azure-storage-blob-batch/README.md
alzimmermsft: Fixed

sdk/storage/azure-storage-blob-batch/README.md
alzimmermsft: Done

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
alzimmermsft: Given this is more of an implementation detail I will clean this up in Preview 5

pom.client.xml
```diff
@@ -401,6 +401,10 @@
                 <title>Azure Storage - Blobs</title>
                 <packages>com.azure.storage.blob*</packages>
               </group>
+              <group>
+                <title>Azure Storage Blob - Batch</title>
+                <packages>com.azure.storage.blob.batch*</packages>
```
conniey: Will this conflict with the regex above? It seems like blob* will match first before blob.batch*

alzimmermsft: It might, mind writing me an issue to fix this after this PR goes in? It is blocking a few releases from going out.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,427 @@
...
+
+        // 'flatMap' the requests to trigger them to run through the pipeline.
+        Disposable disposable = Flux.fromStream(batchOperationQueue.stream())
+            .flatMap(batchOperation -> batchOperation)
+            .subscribe();
+
+        /* Wait until the 'Flux' is disposed of (aka complete) instead of blocking as this will prevent Reactor from
+         * throwing an exception if this was ran in a Reactor thread.
+         */
+        while (!disposable.isDisposed()) {
```
conniey: This feels weird, runs in a tight loop, and could consume unnecessary CPU cycles.

What about:
```java
        return Flux.fromStream(batchOperationQueue.stream())
            .flatMap(batchOperation -> batchOperation)
            .then(Mono.fromRunnable(() -> {
                batchRequest.add(ByteBuffer.wrap(String.format("--%s--%s", batchBoundary, BlobBatchHelper.HTTP_NEWLINE).getBytes(StandardCharsets.UTF_8)));
            }))
            .thenMany(Flux.fromIterable(batchRequest));
```

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -208,4 +210,7 @@
 
   <!-- Allows the SasTokenCredentialPolicy in Implementation -->
   <suppress checks="com.azure.tools.checkstyle.checks.HttpPipelinePolicy" files="com.azure.storage.common.implementation.policy.SasTokenCredentialPolicy.java"/>
+
+  <!-- Empty while loop waiting for Reactor stream completion -->
```
conniey: This suppression could be removed if you removed that while loop.

pom.client.xml
conniey: Hm, since this is the documentation that gets published with our release, it would be worthwhile to verify this works?

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: I've tried using this and it doesn't work for some reason. Given this is implementation detail and there is more work for batching next preview I'll write up an issue.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
```diff
@@ -0,0 +1,137 @@
...
+        String boundary = contentType.split("=", 2)[1];
+
+        return FluxUtil.collectBytesInByteBufferStream(batchResponse.getValue())
+            .flatMap(byteArrayBody -> {
+                String body = new String(byteArrayBody, StandardCharsets.UTF_8);
+
+                // Split the batch response body into batch operation responses.
+                for (String subResponse : body.split("--" + boundary)) {
+                    // This is a split value that isn't a response.
+                    if (!subResponse.contains("application/http")) {
```
conniey: IIRC HTTP response headers and mime types aren't case sensitive.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,427 @@
...
+     * @param blobName The name of the blob.
+     * @param accessTier The tier to set on the blob.
+     * @param leaseAccessConditions Lease access conditions that must be met to allow this operation.
+     * @return a {@link Response} that will be used to associate this operation to the response when the batch is
+     * submitted.
+     * @throws UnsupportedOperationException If this batch has already added an operation of another type.
+     */
+    public Response<Void> setBlobAccessTier(String containerName, String blobName, AccessTier accessTier,
+        LeaseAccessConditions leaseAccessConditions) {
+        return setBlobAccessTierHelper(String.format("%s/%s", containerName, blobName), accessTier,
```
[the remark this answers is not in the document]
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
```diff
@@ -0,0 +1,137 @@
...
+                continue;
+            }
+
+            if (line.startsWith("HTTP")) {
+                Matcher statusCodeMatcher = STATUS_CODE_PATTERN.matcher(line);
+                if (statusCodeMatcher.find()) {
+                    batchOperationResponse.setStatusCode(Integer.parseInt(statusCodeMatcher.group(1)));
+                }
+            } else {
+                String[] headerPieces = line.split(":\\s*", 2);
```
conniey: If the line doesn't contain that split `":\\s*"`, it'll only result in a single array item, causing your following code "headerPieces[1]" to fail with an ArrayOutOfBounds exception.

sdk/storage/azure-storage-blob-batch/pom.xml
```diff
@@ -0,0 +1,277 @@
...
+    <sourceDirectory>src/main/java</sourceDirectory>
+    <testSourceDirectory>src/test/java</testSourceDirectory>
+    <testResources>
+      <testResource>
+        <directory>${basedir}/src/test/resources</directory>
+      </testResource>
+    </testResources>
+  </build>
+
+  <profiles>
```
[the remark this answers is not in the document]
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
```diff
@@ -0,0 +1,137 @@
...
+    static Mono<SimpleResponse<Void>> mapBatchResponse(BlobBatch batch, ServicesSubmitBatchResponse batchResponse,
+        boolean throwOnAnyFailure, ClientLogger logger) {
+        /*
+         * Content-Type will contain the boundary for each batch response. The expected format is:
+         * "Content-Type: multipart/mixed; boundary=batchresponse_66925647-d0cb-4109-b6d3-28efe3e1e5ed"
+         */
+        String contentType = batchResponse.getDeserializedHeaders().getContentType();
+
+        // Split on the boundary [ "multipart/mixed; boundary", "batchresponse_66925647-d0cb-4109-b6d3-28efe3e1e5ed"]
+        String boundary = contentType.split("=", 2)[1];
```
conniey: I would be proactive about ensuring that the split actually results in 2 items, lest you get an ArrayIndex out of bounds exception.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
```diff
@@ -0,0 +1,137 @@
...
+            });
+    }
+
+    private static BlobBatchOperationResponse<?> getBatchOperation(BlobBatch batch, String responseBatchInfo,
+        ClientLogger logger) {
+        Matcher contentIdMatcher = CONTENT_ID_PATTERN.matcher(responseBatchInfo);
+
+        int contentId;
+        if (contentIdMatcher.find()) {
+            contentId = Integer.parseInt(contentIdMatcher.group(1));
```
conniey: The regex you are using, actually matches the group optionally, so this group(1) you're assuming exists here can fail.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
alzimmermsft: Will update this to be case-insensitive, I believe the service is fairly strict on the casing return

sdk/storage/azure-storage-blob-batch/README.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+### Advanced Batching
+
+Deleting blobs in a batch that have different pre-requisites.
+```java
+BlobBatch blobBatch = blobBatchClient.getBlobBatch();
+
+// Delete a blob.
+Response<Void> deleteResponse = blobBatch.deleteBlob(blobUrl);
```
samvaity: should this be deleteBlobWithResponse?
Also, the methods below.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,427 @@
...
+     *
+     * {@codesnippet com.azure.storage.blob.batch.BlobBatch.deleteBlob#String-String}
+     *
+     * @param containerName The container of the blob.
+     * @param blobName The name of the blob.
+     * @return a {@link Response} that will be used to associate this operation to the response when the batch is
+     * submitted.
+     * @throws UnsupportedOperationException If this batch has already added an operation of another type.
+     */
+    public Response<Void> deleteBlob(String containerName, String blobName) {
```
samvaity: Are we not supporting `withResponse` overloads in this?

alzimmermsft: Batching has an interesting implementation where using the `BlobBatch` API instantly returns an empty response until it is submitted via `BlobBatchClient.submitBatch(BlobBatch)`

sdk/storage/azure-storage-blob-batch/README.md
alzimmermsft: This is as it should be given how batching is implemented.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,427 @@
...
+/**
+ * This class allows for batching of multiple Azure Storage operations in a single request via {@link
+ * BlobBatchClient#submitBatch(BlobBatch)} or {@link BlobBatchAsyncClient#submitBatch(BlobBatch)}.
+ *
+ * <p>Azure Storage Blob batches are homogeneous which means a {@link #deleteBlob(String) delete} and {@link
+ * #setBlobAccessTier(String, AccessTier) set tier} are not allowed to be in the same batch.</p>
+ *
+ * {@codesnippet com.azure.storage.blob.batch.BlobBatch.illegalBatchOperation}
+ *
+ * <p>Please refer to the <a href="https://docs.microsoft.com/en-us/rest/api/storageservices/blob-batch">Azure Docs</a>
```
conniey:
```suggestion
 * <p>Please refer to the <a href="https://docs.microsoft.com/rest/api/storageservices/blob-batch">Azure Docs</a>
```

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,427 @@
...
+            .filter(header -> !X_MS_VERSION.equalsIgnoreCase(header.getName()))
+            .forEach(header -> appendWithNewline(batchRequestBuilder,
+                String.format(HEADER_TEMPLATE, header.getName(), header.getValue())));
+
+        batchRequestBuilder.append(BlobBatchHelper.HTTP_NEWLINE);
+
+        batchRequest.add(ByteBuffer.wrap(batchRequestBuilder.toString().getBytes(StandardCharsets.UTF_8)));
+        batchMapping.get(contentId).setRequest(request);
+
+        return Mono.empty();
```
conniey: I would replace this with 
```java
return Mono.fromRunnable(() -> {
// stuff
});
```
So that the string concatonation and so forth can happen asynchronously.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,427 @@
...
+    private final Map<Integer, BlobBatchOperationResponse<?>> batchMapping;
+
+    private final AtomicInteger contentId;
+    private final String batchBoundary;
+    private final String contentType;
+
+    private BlobBatchType batchType;
+
+    BlobBatch(String accountUrl, HttpPipeline pipeline) {
+        this.contentId = new AtomicInteger(0);
```
conniey: new AtomicInteger() is sufficient. The default is 0.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,427 @@
...
+        for (int i = 0; i < pipeline.getPolicyCount(); i++) {
+            HttpPipelinePolicy policy = pipeline.getPolicy(i);
+
+            if (policy instanceof SharedKeyCredentialPolicy) {
+                batchHeadersPolicySet = true;
+                // The batch policy needs to be added before the SharedKey policy to run preparation cleanup.
+                batchPipelineBuilder.policies(this::cleanseHeaders, this::setRequestUrl);
+            }
+
+            batchPipelineBuilder.policies(pipeline.getPolicy(i));
```
conniey:
```suggestion
            batchPipelineBuilder.policies(policy);
```

conniey: You've already evaluated it in line 99.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    /**
+     * Adds a delete blob operation to the batch.
+     *
+     * <p><strong>Code sample</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.batch.BlobBatch.deleteBlob#String-DeleteSnapshotsOptionType-BlobAccessConditions}
+     *
+     * @param blobUrl URI of the blob.
```
conniey: These are actually URLs, that you support, the documentation is misleading.

https://danielmiessler.com/study/url-uri/

alzimmermsft: Will update in Preview 5

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: Will update in Preview 5

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: Will update in Preview 5

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchHelper.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/5756

alzimmermsft:
GitHub isn't reporting step passes properly and random transient error in a core test.

Issue #3752, opened 2019-05-22
JavaDoc review and improvements for azure core public API


JonathanGiles:
We need to review all public API in Azure core to ensure it is high-quality, fully implemented, clearly-explained, and meeting our high expectations for JavaDoc in Azure Java SDK.

joshfree:
@anuchandy could you help load balance this one?

anuchandy:
@josefree I will also take a look from my side.

joshfree:
Closing issue given #5653 #5735 shipped in the November update

PR #5735, opened 2019-10-08
azure-core JavaDocs and code snippet clean-up

conniey:
* Updates documentation for classes in packages:
  * com.azure.core.http.rest
  * com.azure.core.util
  * com.azure.core.exception
* Updates javadoc code snippets for PagedFlux and PagedFluxBase.
  * They were side-effect operators like doOnSubscribe, doOnNext rather than a subscribe overload.
  * Use `PagedFlux.subscribe(CoreSubscriber)` overload rather than the Flux one.
* Fixed potential NullPointerException if `StreamResponse.getValue()` returns null.

Related to: #3752

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
@@ -37,27 +37,27 @@ public class ClientLogger {
     /*
      * Indicates that log level is at verbose level.
      */
-    public static final int VERBOSE_LEVEL = 1;
+    private static final int VERBOSE_LEVEL = 1;
 
     /*
      * Indicates that log level is at information level.
      */
-    public static final int INFORMATIONAL_LEVEL = 2;
+    private static final int INFORMATIONAL_LEVEL = 2;
 
     /*
      * Indicates that log level is at warning level.
      */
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -35,29 +35,29 @@
     private final Logger logger;
 
     /*
-     * Indicate that log level is at verbose level.
+     * Indicates that log level is at verbose level.
      */
-    private static final int VERBOSE_LEVEL = 1;
+    public static final int VERBOSE_LEVEL = 1;
```
conniey: @alzimmermsft @JonathanGiles 
One thing I wanted to ask if there were negative repercussions to making this public. From an external user's point of view, it's hard to know what the log levels are without going to the source.

I have a link in:

JonathanGiles: Yeah, ideally we wouldn't do this

conniey: How do external customers decide which integer to set AZURE_LOG_LEVEL if we move this back to being private? If I set AZURE_LOG_LEVEL to 10, what behaviour should expect.

PR #5742, opened 2019-10-09
Update tracing readme

samvaity:
Update tracing readme to use correct package dependency and configuring http client.

### DIFF
--- a/sdk/tracing/azure-core-tracing-opencensus/README.md
+++ b/sdk/tracing/azure-core-tracing-opencensus/README.md
@@ -10,8 +10,9 @@ documentation][api_documentation] | [Samples][samples]
 ## Getting started
 ### Prerequisites
 - Java Development Kit (JDK) with version 8 or above
-- opencensus-api package with version 0.20.0 or above 
+- opencensus-api package with version 0.20.0
 - [Maven][maven]
+
 ```xml
 <dependency>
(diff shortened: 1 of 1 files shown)


sdk/tracing/azure-core-tracing-opencensus/README.md
```diff
@@ -10,54 +10,56 @@ documentation][api_documentation] | [Samples][samples]
 ## Getting started
 ### Prerequisites
 - Java Development Kit (JDK) with version 8 or above
+- opencensus-api package with version 0.20.0 or above
```
JimSuplizio: There have been [5 versions of opencensus-api](https://search.maven.org/search?q=g:io.opencensus%20AND%20a:opencensus-api&core=gav) since 0.20.0  and we're building this artifact with the 0.20.0 version. Can we reliably make the statement that version 0.20.0 or above will work? Also, should we be using the latest version, 0.24.0 to build?

samvaity: I think with the current changes around the tracing package we can continue to support just 0.20.0 for now, since this should go away in the near future and be replaced with opentelemetry package dependency. 

Also, since we already have core published with 0.20.0 version of this dependency I will update this to say we support only 0.20.0.

Issue #5132, opened 2019-08-27
`NoImplInPublicAPI` suppressed check in OpenTelemetryTracer


samvaity:
Confirm and check if we need the `NoImplInPublicAPI` in OpenTelemetryTracer class

samvaity:
Fixed in #5747

PR #5747, opened 2019-10-09
Integration tests for Tracing

### LINKED ISSUE #5156: Add end to end tracing testing for libraries (Http and Amqp)
Add more concrete and end to end tests for testing tracing in Amqp and Http libraries using the plugin package.

samvaity:
Fixes #5156, #5132

### DIFF
--- a/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
+++ b/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
@@ -76,9 +76,6 @@
   <suppress checks="FinalClass" files="RequestResponseLink.java"/>
   <suppress checks="FinalClass" files="Util.java"/>
 
-  <!-- External dependency exposed in the public API  -->
-  <suppress checks="com.azure.tools.checkstyle.checks.ExternalDependencyExposedCheck" files="com.azure.core.tracing.opencensus.implementation.HttpTraceUtil.java"/>
-
   <!-- Star imports are ok in unit tests - it makes it easier to import all assertions, etc -->
   <suppress checks="AvoidStarImport" files=".*[/\\]src[/\\]test[/\\]java[/\\].*\.java"/>
 
--- a/sdk/core/azure-core-tracing-opencensus/CHANGELOG.md
+++ b/sdk/core/azure-core-tracing-opencensus/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Release History
 
 ## Version 1.0.0-preview.4 (2019-10-25)
-For details on the Azure SDK for Java (October 2019 Preview) release refer to the [release announcement](https://aka.ms/azure-sdk-preview5-java).
+For details on the Azure SDK for Java (November 2019 Preview) release refer to the [release announcement](https://aka.ms/azure-sdk-preview5-java).
 
 Version 1.0.0-preview.4 added tracing support for AMQP and HTTP client libraries across different languages and platforms as possible.
 See this package's
--- a/sdk/core/azure-core-tracing-opencensus/README.md
+++ b/sdk/core/azure-core-tracing-opencensus/README.md
@@ -53,7 +53,7 @@ Netty and include OkHTTP client in your pom.xml.
 <dependency>
   <groupId>com.azure</groupId>
   <artifactId>azure-core-http-okhttp</artifactId>
-  <version>1.0.0-preview.7</version>
+  <version>1.0.0</version>
 </dependency>
 ```
 [//]: # ({x-version-update-end})
@@ -194,7 +194,7 @@ These following samples provide example code for additional scenarios commonly e
     * Create new version of an existing secret.
     * List versions of an existing secret.
 #### Publish multiple Events to an Event Hub
-* [PublishEvents.java][sample_publish_events] - Contains sample for publishing multiple events with tracing support.
+* [PublishEvents][sample_publish_events] - Contains sample for publishing multiple events with tracing support.
 
 ### Additional Documentation
 For more extensive documentation on OpenCensus, see the [API reference documentation][opencensus].
--- a/sdk/core/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/implementation/AmqpTraceUtil.java
+++ b/sdk/core/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/implementation/AmqpTraceUtil.java
@@ -7,8 +7,7 @@ import io.opencensus.trace.Status;
 
 public final class AmqpTraceUtil {
 
-    private AmqpTraceUtil() {
-    }
+    private AmqpTraceUtil() { }
 
     /**
      * Parses an OpenTelemetry Status from AMQP Error Condition.
--- a/sdk/core/azure-core-tracing-opencensus/src/samples/AsyncListKeyVaultSecrets.md
+++ b/sdk/core/azure-core-tracing-opencensus/src/samples/AsyncListKeyVaultSecrets.md
@@ -11,7 +11,7 @@ Following documentation describes instructions to run a sample program for listi
 <dependency>
   <groupId>com.azure</groupId>
   <artifactId>azure-security-keyvault-secrets</artifactId>
-  <version>4.0.0-preview.5</version>
+  <version>4.0.0</version>
   <exclusions>
     <exclusion>
       <groupId>com.azure</groupId>
@@ -22,12 +22,12 @@ Following documentation describes instructions to run a sample program for listi
 ```
 [//]: # ({x-version-update-end})
 [//]: # ({x-version-update-start;com.azure:azure-core-http-okhttp;current})
-```
+```xml
 <!-- Add OkHTTP client to use with KeyVault Secrets -->
 <dependency>
   <groupId>com.azure</groupId>
   <artifactId>azure-core-http-okhttp</artifactId>
-  <version>1.0.0-preview.4</version>
+  <version>1.0.0</version>
 </dependency>
 ```
 [//]: # ({x-version-update-end})
--- a/sdk/core/azure-core-tracing-opencensus/src/samples/AsyncQueueClientEnqueueMessages.md
+++ b/sdk/core/azure-core-tracing-opencensus/src/samples/AsyncQueueClientEnqueueMessages.md
@@ -12,7 +12,7 @@ messages asynchronously on queue client with tracing instrumentation for Java SD
 <dependency>
   <groupId>com.azure</groupId>
   <artifactId>azure-storage-queue</artifactId>
-  <version>12.0.0-preview.4</version>
+  <version>12.0.0</version>
 </dependency>
 ```
 [//]: # ({x-version-update-end})
--- a/sdk/core/azure-core-tracing-opencensus/src/samples/ListeKeyVaultSecrets.md
+++ b/sdk/core/azure-core-tracing-opencensus/src/samples/ListeKeyVaultSecrets.md
@@ -11,7 +11,7 @@ Following documentation describes instructions to run a sample program for listi
 <dependency>
     <groupId>com.azure</groupId>
     <artifactId>azure-security-keyvault-secrets</artifactId>
-    <version>4.0.0-preview.5</version>
+    <version>4.0.0</version>
     <exclusions>
       <exclusion>
         <groupId>com.azure</groupId>
@@ -22,12 +22,12 @@ Following documentation describes instructions to run a sample program for listi
 ```
 [//]: # ({x-version-update-end})
 [//]: # ({x-version-update-start;com.azure:azure-core-http-okhttp;current})
-```
+```xml
 <!-- Add OkHTTP client to use with KeyVault Secrets -->
 <dependency>
   <groupId>com.azure</groupId>
   <artifactId>azure-core-http-okhttp</artifactId>
-  <version>1.0.0-preview.4</version>
+  <version>1.0.0</version>
 </dependency>
 ```
 ### Adding the Azure core tracing OpenCensus plugin package to your project:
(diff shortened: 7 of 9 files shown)


sdk/tracing/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/OpenCensusTracer.java
```diff
@@ -258,7 +256,11 @@ private void addSpanRequestAttributes(Span span, Context context, String spanNam
      * @return The component name contained in the context.
      */
     private static String parseComponentValue(String spanName) {
-        return spanName.substring(spanName.indexOf(".") + 1, spanName.lastIndexOf("."));
+        if (spanName.length() > 0 && spanName.indexOf(".") != -1 && spanName.lastIndexOf(".") != -1) {
```
joshfree: the jit is hopefully going to optimize away these redundant calls to indexOf; but just in case, could you refactor the code to use temporary ints and avoid the duplicate API calls?

joshfree: should we guard against a null string as well?

sdk/tracing/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/implementation/AmqpTraceUtil.java
```diff
@@ -24,7 +24,7 @@ public static Status parseStatusMessage(String statusMessage, Throwable error) {
                 ? Status.UNKNOWN.withDescription(message)
                 : Status.UNKNOWN.withDescription(error.getClass().getSimpleName());
 
-        } else if (statusMessage.equalsIgnoreCase("success")) {
+        } else if (statusMessage != null && statusMessage.equalsIgnoreCase("success")) {
```
joshfree: given the first if-block returns, could you move this to the next line down as a regular if-block instead to make it slightly more clear.  Similarly, do you need to handle the statusMessage==null case separately so it doesn't also flow down to the Status.UNKNOWN.withDescription below?

sdk/tracing/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -51,27 +183,92 @@ public void startSpanOverloadNullPointerException() {
...
+    @Test
+    public void endSpanNoSuccessErrorMessageTest() {
+        // Arrange
+        RecordEventsSpanImpl recordEventsSpan = (RecordEventsSpanImpl) tracer.getCurrentSpan();
+
+        // Act
+        openCensusTracer.end(null, null, tracingContext);
+
+        // Assert
+        Assert.assertEquals(recordEventsSpan.getStatus().getCanonicalCode().toString(), "UNKNOWN");
```
joshfree: I know this is test code, but consider moving the string literals into a constant

joshfree:
please be sure to squash this commit :)

sdk/tracing/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -51,27 +183,92 @@ public void startSpanOverloadNullPointerException() {
...
+        RecordEventsSpanImpl existingSpan = (RecordEventsSpanImpl) tracer.spanBuilder("existing-span").startSpan();
+        Context traceContext = tracingContext.addData(SPAN_CONTEXT, existingSpan.getContext());
+        RecordEventsSpanImpl parentSpanImpl = (RecordEventsSpanImpl) parentSpan;
+
+        // Act
+        openCensusTracer.addLink(traceContext);
+
+        //Assert
+        // TODO: existing -> parentSpanImpl comes after after add Link, confirm this behavior?
+        Assert.assertEquals(parentSpanImpl.getPrev(), existingSpan);
```
samvaity: @lmolkova I do not remember this very clearly but I think we introduced something called as a `Link.fromSpanContext` to mimic otel's behavior of adding Link.

But the way it should have worked is adding a link after the parentSpan, correct?

sdk/tracing/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -36,13 +77,111 @@ public void startSpanNullPointerException() {
...
+    }
+
+    @Test
+    public void startSpanProcessKindProcess() {
+        // Arrange
+        tracer.spanBuilder(OPENCENSUS_SPAN_KEY).startScopedSpan();
+        SpanId parentSpanId = parentSpan.getContext().getSpanId();
+
+        // Act
+        final Context updatedContext = openCensusTracer.start(METHOD_NAME, tracingContext, ProcessKind.PROCESS);
```
lmolkova: This typically happens when there is no parent, could we test it works?

sdk/tracing/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -36,13 +77,111 @@ public void startSpanNullPointerException() {
...
+
+        // verify span attributes
+        Map<String, AttributeValue> attributeMap = recordEventsSpan.toSpanData().getAttributes().getAttributeMap();
+        Assert.assertEquals(attributeMap.get(COMPONENT), AttributeValue.stringAttributeValue(COMPONENT_VALUE));
+        Assert.assertEquals(attributeMap.get(MESSAGE_BUS_DESTINATION), AttributeValue.stringAttributeValue(ENTITY_PATH_VALUE));
+        Assert.assertEquals(attributeMap.get(PEER_ENDPOINT), AttributeValue.stringAttributeValue(HOSTNAME_VALUE));
+    }
+
+    @Test
+    public void startSpanProcessKindReceive() {
```
lmolkova: I recall we decided against instrumenting Receive call

sdk/tracing/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -51,27 +190,97 @@ public void startSpanOverloadNullPointerException() {
...
+        Context traceContext = tracingContext.addData(SPAN_CONTEXT, existingSpan.getContext());
+        RecordEventsSpanImpl parentSpanImpl = (RecordEventsSpanImpl) parentSpan;
+
+        // Act
+        openCensusTracer.addLink(traceContext);
+
+        //Assert
+        // TODO: existing -> parentSpanImpl comes after after add Link, confirm this behavior?
+        Assert.assertEquals(parentSpanImpl.getPrev(), existingSpan);
+        Assert.assertEquals(parentSpanImpl.toSpanData().getChildSpanCount().intValue(), 1);
```
lmolkova: This is not related to links - not needed

sdk/tracing/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
lmolkova: With opencensus, links are added one the span after span is created, with opentelemetry - before creation on builder.

What do you mean by 'after the parentSpan'?

sdk/tracing/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
samvaity: Receive is the message span. 
I think with eventhubs we wanted to call it this way as "message" didn't macth any of th existing terms we used in the package.

sdk/tracing/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
samvaity: So the way we have the link is, we have parentSpan followed by existing span.

samvaity: Removed this and now checking if the two spans are related by the same TraceId

sdk/core/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -35,13 +79,119 @@ public void startSpanNullPointerException() {
...
+        final SpanId parentSpanId = parentSpan.getContext().getSpanId();
+
+        // Act
+        final Context updatedContext = openCensusTracer.start(METHOD_NAME, tracingContext);
+
+        // Assert
+        assertSpanWithExplicitParent(updatedContext, parentSpanId);
+        final RecordEventsSpanImpl recordEventsSpan =
+            (RecordEventsSpanImpl) updatedContext.getData(PARENT_SPAN_KEY).get();
+        Assert.assertNull(recordEventsSpan.getKind());
```
lmolkova: just curious, why check for null kind?

sdk/core/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -50,27 +200,99 @@ public void startSpanOverloadNullPointerException() {
...
+        final Context traceContext = tracingContext.addData(SPAN_CONTEXT_KEY, testSpan.getContext());
+        final RecordEventsSpanImpl parentSpanImpl = (RecordEventsSpanImpl) parentSpan;
+
         // Act
-        final Context updatedContext = openCensusTracer.start(METHOD_NAME, Context.NONE, ProcessKind.SEND);
+        openCensusTracer.addLink(traceContext);
+
+        //Assert
+        // verify both spans share the same link traceId
+        Assert.assertEquals(parentSpanImpl.toSpanData().getContext().getTraceId(),
```
lmolkova: this seems to be an incorrect validation. Links do not flow into trace-id.

So i guess what happens in this test:
setUp creates parent span
testSpan will get same TraceId - because it's a child
traceContext gets context from traceSpan
link is created from traceContext (which shares trace-id with parentSpan and testSpan
Then you check if parentSpan has the same traceId as link

What should be tested:
link should have random context - new one
you add it  to the span
and then compare links on the parent span to what you've added

sdk/core/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
samvaity: Just to validate it isn't getting set anywhere in the workflow.

sdk/core/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
samvaity: updated.

sdk/core/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -35,13 +79,119 @@ public void startSpanNullPointerException() {
...
+        Assert.assertEquals(attributeMap.get(PEER_ENDPOINT), AttributeValue.stringAttributeValue(HOSTNAME_VALUE));
+    }
+
+    @Test
+    public void startSpanProcessKindReceive() {
+        // Arrange
+        final SpanId parentSpanId = parentSpan.getContext().getSpanId();
+
+        // Act
+        final Context updatedContext = openCensusTracer.start(METHOD_NAME, tracingContext, ProcessKind.RECEIVE);
```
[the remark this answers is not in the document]
samvaity: updated it to `message`.

sdk/core/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -35,13 +80,119 @@ public void startSpanNullPointerException() {
...
+        final Map<String, AttributeValue> attributeMap = recordEventsSpan.toSpanData().getAttributes()
+            .getAttributeMap();
+        Assert.assertEquals(attributeMap.get(COMPONENT), AttributeValue.stringAttributeValue(COMPONENT_VALUE));
+        Assert.assertEquals(attributeMap.get(MESSAGE_BUS_DESTINATION),
+            AttributeValue.stringAttributeValue(ENTITY_PATH_VALUE));
+        Assert.assertEquals(attributeMap.get(PEER_ENDPOINT), AttributeValue.stringAttributeValue(HOSTNAME_VALUE));
+    }
+
+    @Test
+    public void startSpanProcessKindReceive() {
```
lmolkova:
```suggestion
    public void startSpanProcessKindMessage() {
```

PR #5757, opened 2019-10-09
Fix Send Span error in eventhubs

samvaity:
`sendSpanContext` was getting updated on every event causing the order of spans to be messed up. So updated it to be set only once for the entire event batch.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncProducer.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncProducer.java
@@ -357,13 +357,13 @@ public class EventHubAsyncProducer implements Closeable {
                     final AtomicReference<Boolean> isFirst = new AtomicReference<>(true);
                     return events.map(eventData -> {
                         Context parentContext = eventData.getContext();
-                        if (isFirst.get()) {
+                        if (isFirst.getAndSet(false)) {
                             // update sendSpanContext only once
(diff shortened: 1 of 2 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubAsyncProducer.java
```diff
@@ -354,11 +354,16 @@
                         .setPartitionKey(partitionKey)
                         .setMaximumSizeInBytes(batchSize);
 
+                    final AtomicReference<Boolean> isFirst = new AtomicReference<>(true);
```
conniey: AtomicBoolean is sufficient. Then you can do:

```java
if (!isFirst.getAndSet(true)) {
    // update sendSpanContext only once

    Context entityContext = parentContext.addData(ENTITY_PATH, link.getEntityPath()); 
    sendSpanContext.set(tracerProvider 
        .startSpan(entityContext.addData(HOST_NAME, link.getHostname()), ProcessKind.SEND));
}

// add span context on event data
return setSpanContext(eventData, parentContext);
```

conniey: Would suggest writing a test for this, too.

samvaity: updated, with the above code and fixed the already existing test cases for this.

Issue #5671, opened 2019-10-03
[API Review] Auto-gen layer queue


sima-zhu:
- [x] CorsRule `getMaxAgeInSeconds()` why not milliseconds and return long

- [x] Should these classes (e.g. [EqueuedMessage](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/models/EnqueuedMessage.java)) be made immutable? Remove setter and final member data.

- [x] Rename Logging to LogData?

- [x] isIncludeAPIs() in Metrics return boolean instead of Boolean

- [x] `setName` in QueueItem rename to `setQueueName`

- [x] QueueMessage class is not used anywhere, so moved to implementation.

- [x] `getMaxResults()` in QueueItem return int instead of Integer?

- [x] Extract StorageException to common module.

- [x] `setLogging` in StorageServiceProperties whether can make it clear to differentiate from client-side logging.

- [x] The necessity of `StorageServiceStats` as it only has one property field in class.

Reference:  https://apiview.azurewebsites.net/Assemblies/Review/2cefcabd89fe4b008f29c7300ba1c3c2

rickle-msft:
- Because the value is in seconds. So why would we report it in miliseconds?
- These are generated types. We'd love to have them immutable but were told that the generator can't determine when types are supposed to be immutable. 
- This is called Logging because that's the xml value the service accepts. I understand the confusion with client side logging, but I also don't think LogData is accurate. Maybe something like ServerLoggingConfigurations
- I think this can be a boolean. If metrics are enabled, it's a required parameter, so that should be ok.
- I'm fine with setName to be setQueueName. I think there's maybe a risk of "stuttering", but if Jonathan thinks it's more clear, then ok.
- I'm seeing QueueMessage is used in QueueAsyncClient (478 and 696(
- I believe this actually applies to QueueSegmentOptions, but it should be an Integer. It's optional. 
- Left a comment inline about merging the StorageException types, but basically there are generator limits to this. 
- This seems like it should be in line with whatever we decide for 3.
- This is the return value for getServiceStats, and this reflects how the service structures this data. Presumably if they created an object here, they are leaving space for future additions and we should do the same.

sima-zhu:
*Regarding: Should these classes (e.g. EqueuedMessage) be made immutable? Remove setter and final member data.*

[...]

PR #5776, opened 2019-10-10
API changes for queue and file

sima-zhu:
1. Rename FileRef -> StorageFileItem (align with .NET and FIleItem has been taken)

2. Rename to get* for getter method in FileUploadRangeFromUrlInfo

3. Removed the FileExtension and make the helper private

4. Move QueueMessage to implementation folder.

### DIFF
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
@@ -299,8 +299,8 @@ public final class Utility {
      */
     public static void assertInBounds(final String param, final long value, final long min, final long max) {
         if (value < min || value > max) {
-            throw new IllegalArgumentException(String.format(Locale.ROOT,
-                Constants.MessageConstants.PARAMETER_NOT_IN_RANGE, param, min, max));
+            throw LOGGER.logExceptionAsError(new IllegalArgumentException(String.format(Locale.ROOT,
+                Constants.MessageConstants.PARAMETER_NOT_IN_RANGE, param, min, max)));
         }
     }
 


sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
```diff
@@ -842,18 +843,37 @@ public String getAccountName() {
...
-                .forEach(fileItem -> fileReferences.add(new FileReference(fileItem.getName(), false,
+                .forEach(fileItem -> storageFileItems.add(new StorageFileItem(fileItem.getName(), false,
                     fileItem.getProperties())));
         }
 
-        return new ArrayList<>(fileReferences);
+        return new ArrayList<>(storageFileItems);
+    }
+
+    /**
```
gapra-msft: Can this method be made package private so you dont have to repeat the code in File and Directory?

sima-zhu: If make it package private, then we need to have an object to make the call. It is a little overhead, as DirectoryAsyncClient and FileAsyncClient does not have the other client initialized.
So better not to make it package private.

alzimmermsft: The other option is having a package private static method which won't need to initialization overhead and would serve the same purpose.

sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/models/QueuesSegmentOptions.java
```diff
@@ -73,7 +73,7 @@ public QueuesSegmentOptions setPrefix(String prefix) {
     /**
      * @return the maximum number of queues to include in a single response
      */
-    public Integer getMaxResults() {
+    public int getMaxResults() {
```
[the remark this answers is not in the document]
anuchandy: @rickle-msft [mentioned](https://github.com/Azure/azure-sdk-for-java/issues/5671#issuecomment-538097440) that "I believe this actually applies to QueueSegmentOptions, but it should be an Integer. It's optional.", so swagger is correct and we can't present it as primitive int.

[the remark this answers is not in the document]
sima-zhu: Revert it as the default int = 0 is not allowed by service. Thanks for comments.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/DirectoryAsyncClient.java
sima-zhu: Do you mean we have another StorageFileUtility under File?

sima-zhu: Prefer to leave it here and abstract these if we have more helper methods needed. As this is internal changes, it will introduce any breaking changes.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileAsyncClient.java
```diff
@@ -1236,4 +1237,23 @@ public String getAccountName() {
...
+     */
+    private void validateFilePermissionAndKey(String filePermission, String  filePermissionKey) {
+        if (filePermission != null && filePermissionKey != null) {
+            throw logger.logExceptionAsError(new IllegalArgumentException(
+                FileConstants.MessageConstants.FILE_PERMISSION_FILE_PERMISSION_KEY_INVALID));
+        }
+
+        if (filePermission != null) {
+            Utility.assertInBounds("filePermission",
+                filePermission.getBytes(StandardCharsets.UTF_8).length, 0, 8 * Constants.KB);
```
anuchandy: The `assertInBounds` is throwing an exception https://github.com/Azure/azure-sdk-for-java/blob/fec0fa1d35e7b13654815d927a338746e7582b9a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java#L300 but we're not logging it

anuchandy: may be we can pass logger to assertInBounds and let it use to log err, this way we use a client logger attached to the correct class it's gets called from

anuchandy: @sima-zhu since this is not a release blocker we can track this separately if you think so.

Issue #5760, opened 2019-10-10
[ArchBoard] Support for API / service version specifier in client builder API


JonathanGiles:
At present we do not follow spec in our APIs, in that a service / API version can not be set by the user optionally. We should compare with the other languages to ensure we are meeting expectations.

joshfree:
Discussed offline - we will add this in the next wave

JonathanGiles:
Further discussion offline - we will do that this in the next 1.5 weeks after all.

JonathanGiles:
@pakrym What is the name of the enum you use in .NET? Have you considered it being different per client library? E.g. KeyVaultKeyServiceVersion, StorageBlobServiceVersion, etc?

pakrym:
`ServiceVersion`, it's nested inside client-specific option classes https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/keyvault/Azure.Security.KeyVault.Secrets/src/SecretClientOptions.cs#L32 so we didn't do differnt names.

pakrym:
I don't think we've spent a lot of time considering nested vs not-nested enums, honestly.

srnagar:
Every Azure service has its own version of APIs that it supports. Java client libraries should be able to interact with multiple service APIs. 

Users should be able to create a client for any Azure service by specifying the service API version that they want to use. The client should interact with the Azure service using that specific API version.

This allows users to stick to a specific API version and if a newer service API is available that can potentially add/remove features, users are not impacted by that.

One way to do this is by having an enumerated list of all service API versions a client library supports and the user can optionally provide one of the enumerated service versions to the client builder to create the client. A client created with a specific service version will use the request/response protocol mapped to that version.


For e.g.
AppConfig Service has two API versions - v1, v2

AppConfig client library will define an enum:

```java
public enum ServiceVersion {
 V1,
 V2
}

public class ConfigurationClientBuilder {
   ...
   public ConfigurationClientBuilder serviceVersion(ServiceVersion version) {
     this.version = version;
     return this;
   }

   public ConfigurationClient buildClient() {
      // build client with provided version 
      return configurationClient;
   }
}
```


```java
public static void main(String[] args) {
    ConfigurationClient client = new ConfigurationClientBuilder()
                                                  .serviceVersion(ServiceVersion.V2)
                                                  .credentials(credentials)
                                                  .buildClient();
    client.addSetting(...); // this should use V2 service API

[...]

PR #5782, opened 2019-10-11
Allow users to specify the Service Version to use when sending requests

samvaity:
Add method in builders for users to specify the service/api version to use when sending HTTP requests.
Reference: #5760


rickle-msft:
Is this in the guidelines that we have to support specifying versions? We've historically been very intentional about avoiding this because it turns into spaghetti code very quickly. If the service changes a model type or changes behavior, we have to have different code paths to handle that forever or otherwise risk breaking/having incorrect behavior

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -292,6 +293,19 @@ public ConfigurationClientBuilder retryPolicy(RetryPolicy retryPolicy) {
...
 
+    /**
+     * Sets the {@link ServiceVersion} that is used when making API requests.
+     *
+     * The latest service version will be used if not provided {@link ServiceVersion}
+     *
+     * @param version {@link ServiceVersion} of the service to be used when making requests.
+     * @return The updated ConfigurationClientBuilder object.
+     */
+    public ConfigurationClientBuilder serviceVersion(ServiceVersion version) {
```
srnagar: The `buildClient` and `buildAsyncClient` methods should pass this version to the client constructor.

The client must use the appropriate `ServiceInterface` class to pass to `RestProxy.create();`. This can be done as next step but the version should be passed on to the client.

sdk/keyvault/azure-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientBuilder.java
```diff
@@ -73,6 +73,7 @@
     HttpLogOptions httpLogOptions;
     final RetryPolicy retryPolicy;
     Configuration configuration;
+    ServiceVersion version;
```
srnagar: All these fields have to be `private`.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ServiceVersion.java
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.storage.common;
+
+/**
+ * The versions of Azure Storage supported by this client library.
+ */
+public enum ServiceVersion {
```
srnagar: Is it possible to have different service versions for blobs, files and queues?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ServiceVersion.java
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.data.appconfiguration;
+
+/**
+ * The versions of Azure App Configuration supported by this client library.
+ */
+public enum ServiceVersion {
+    V1_0("1.0");
```
srnagar: The service version should also be included in the user-agent string.

sdk/keyvault/azure-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientBuilder.java
samvaity: These same fields are accessed [here](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/keyvault/azure-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/KeyEncryptionKeyClientBuilder.java#L71) in the `KeyEncryptionKeyClientBuilder` and so cannot be made private.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ServiceVersion.java
samvaity: I will probably update this for Storage to have ServiceVersion's in each of the clients, similar to KeyVault.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
samvaity: Added Added `serviceVersion` field to async client constructors.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ServiceVersion.java
samvaity: Added `versionString` to user-agent string.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -305,6 +307,19 @@ public ConfigurationClientBuilder retryPolicy(RetryPolicy retryPolicy) {
         return this;
     }
 
+    /**
+     * Sets the {@link ServiceVersion} that is used when making API requests.
+     *
+     * The latest service version will be used if not provided {@link ServiceVersion}
```
JonathanGiles:
```suggestion
     * If a service version is not provided, the service version that will be used will be the latest known service version based on the version of the client library being used. If no service version is specified, updating to a newer version the client library will have the result of potentially moving to a newer service version.
```

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ServiceVersion.java
JonathanGiles: Use the service naming scheme (e.g. `V2019_05_09`)

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ServiceVersion.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    /**
+     * Gets the latest service version supported by this client library
+     *
+     * @return the latest {@link ServiceVersion}
+     */
+    public static ServiceVersion getLatest() {
+        return V1_0;
+    }
```
JonathanGiles: Consider introducing a `ServiceVersion` interface in azure-core that has `getVersionString()`.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -62,12 +62,12 @@ public UserAgentPolicy(String userAgent) {
      * @param sdkVersion Version of the client library.
      * @param configuration Configuration store that will be checked for the AZURE_TELEMETRY_DISABLED.
      */
-    public UserAgentPolicy(String sdkName, String sdkVersion, Configuration configuration) {
+    public UserAgentPolicy(String sdkName, String sdkVersion, Configuration configuration, String versionString) {
```
JonathanGiles: With a `ServiceVersion` interface, you can then have this `UserAgentPolicy` take the `ServiceVersion` directly.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
JonathanGiles: Use this text everywhere, or some variation, to make it clear to the user how things work.

amishra-dev:
@JonathanGiles this PR must not be approved. We should not commit to supporting the dev's passing random service API versions.
/cc @joshfree fyi

anuchandy:
FYI - there is also discussion happening in the original issue https://github.com/Azure/azure-sdk-for-java/issues/5760 that this PR attempt to address.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/AppConfigurationServiceVersion.java
```diff
@@ -6,12 +6,12 @@
 /**
  * The versions of Azure App Configuration supported by this client library.
  */
-public enum ServiceVersion implements com.azure.core.http.ServiceVersion {
+public enum AppConfigurationServiceVersion implements com.azure.core.http.ServiceVersion {
```
srnagar: You can use import instead of a fully qualified class name.

sdk/keyvault/azure-keyvault-certificates/src/main/java/com/azure/security/keyvault/certificates/CertificateServiceVersion.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * The versions of Azure Key Vault supported by this client library.
+ */
+public enum CertificateServiceVersion implements com.azure.core.http.ServiceVersion {
+    V7_0("7.0");
+
+    private final String version;
+
+    CertificateServiceVersion(String version) {
```
srnagar: Make this constructor private

sdk/keyvault/azure-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientBuilder.java
srnagar: It is preferable to have package-private getters instead of making the fields package-private.

sdk/keyvault/azure-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/ServiceVersion.java
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.security.keyvault.keys.cryptography;
+
+/**
+ * The versions of Azure Key Vault supported by this client library.
+ */
+public enum ServiceVersion implements com.azure.core.http.ServiceVersion {
```
srnagar: This must be named as `CryptographyServiceVersion`

sdk/keyvault/azure-keyvault-certificates/src/main/java/com/azure/security/keyvault/certificates/CertificateServiceVersion.java
samvaity: Aren't enum constructors private or package-private by default? Gives me a redundant access modifier.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/AzureBlobStorageBuilder.java
```diff
@@ -33,15 +34,15 @@ public AzureBlobStorageBuilder url(String url) {
     /*
      * Specifies the version of the operation to use for this request.
      */
-    private String version;
+    private BlobServiceVersion version;
```
alzimmermsft: Don't modify this, it is codegen, same goes for all the Impl class changes. Instead toString the enum before using it.

sdk/core/azure-core/src/main/java/com/azure/core/http/ServiceVersion.java
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.http;
```
JonathanGiles: This doesn't need to be under .http, as we might also use this for amqp, etc. Same with the comment below.

JonathanGiles: Consider perhaps putting this under .util

sdk/keyvault/azure-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/KeyServiceVersion.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.azure.security.keyvault.keys;
+
+import com.azure.core.http.ServiceVersion;
+
+/**
+ * The versions of Azure Key Vault supported by this client library.
+ */
+public enum KeyServiceVersion implements ServiceVersion {
+    V7_0("7.0");
```
JonathanGiles: It might pay to confer with @schaabs about whether the service versions of Key Vault will always move in lock-step (for keys, secrets, and certs), or if there is a possibility that they may be versioned independently.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceVersion.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.azure.storage.blob;
+
+import com.azure.core.http.ServiceVersion;
+
+/**
+ * The versions of Azure App Configuration supported by this client library.
+ */
+public enum BlobServiceVersion implements ServiceVersion {
+    V2019_02_02("2019-02-02");
```
JonathanGiles: What string does the service expect to see? Is it prefixed with `V`, and the blob code is doing that? What string do the version docs state - with or without the `V` prefix. It just feels like this is a little disconnected.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileServiceVersion.java
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.storage.file;
+
+import com.azure.core.http.ServiceVersion;
+
+/**
+ * The versions of Azure App Configuration supported by this client library.
```
JonathanGiles: Copy / paste error. It's in a few places so please do a double-check!

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceVersion.java
rickle-msft: The expected header format is just "YYYY-MM-DD"; no 'v' or anything else.

sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueServiceVersion.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.azure.storage.queue;
+
+import com.azure.core.util.ServiceVersion;
+
+/**
+ * The versions of Azure Storage Queue supported by this client library.
+ */
+public enum QueueServiceVersion implements ServiceVersion {
+    V2018_03_28("2018-03-28");
```
samvaity: Need to confirm this with @tg-msft

sdk/keyvault/azure-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/KeyServiceVersion.java
schaabs: They have in the past, but not necessarily moving forward. They are adding another resource type which will share the Keys protocol but doesn't implement the Secrets or Certificates protocols. Given that it's somewhat likely that they would start to version these separately.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
```diff
@@ -45,10 +46,11 @@
 
     private final AzureBlobStorageImpl client;
 
-    BlobBatchAsyncClient(String accountUrl, HttpPipeline pipeline) {
+    BlobBatchAsyncClient(String accountUrl, HttpPipeline pipeline, BlobServiceVersion version) {
         this.client = new AzureBlobStorageBuilder()
             .url(accountUrl)
             .pipeline(pipeline)
+            .version(version.getVersionString())
```
samvaity: Do we want to update this `version` too?
But it falls under one of the codegen classes.

sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueServiceVersion.java
samvaity: Confirmed by the storage team, we should use 2019-02-02 as the version.

sdk/core/azure-core/src/main/java/com/azure/core/util/ServiceVersion.java
```diff
@@ -0,0 +1,17 @@
...
+ * A generic interface for sending HTTP requests using the provided service version.
+ */
+public interface ServiceVersion {
+
+    /**
+     * Gets the string representation of the {@link ServiceVersion}
+     *
+     * @return the string representation of the {@link ServiceVersion}
+     */
+    String getVersionString();
```
srnagar: Do we need to name it as `getVersionString` instead of just `getVersion` as it returns `String` anyway?

JonathanGiles: I agree.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
JonathanGiles: It is not public API so don't worry about changing that, but we should get @alzimmermsft and / or @jianghaolu to review this to ensure that our builders always have a `serviceVersion` API and not `version` or any other naming convention.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceVersion.java
JonathanGiles: In this case I wonder if we should name the enum values with a prefix of `V` (as we have here), or if we should just use the string as-is. I'm not convinced the `V` really adds any value.

samvaity: As discussed offline, keeping the V prefix to stay consistent with .NET.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchAsyncClient.java
jianghaolu: This is generated code - Swagger specifies it as "version" thus generated like this. I'm not going to try to guess this is actually the `serviceVersion` property in the code generator.

PR #5784, opened 2019-10-11
Remove Builder Inheritence Structure

### LINKED ISSUE #5608: Review the Need for the Builder Inheritence Structure
This has led to some situations where we are extending from non-public API and have a lot of protected API. Additionally, it has required some less than desirable methods to be created such as `getClazz` which will type check the return conversion instead of suppressing unchecked casting warnings.

alzimmermsft:
Fixes #5608 

Removes `BaseClientBuilder` from Storage Common and updates all builders inheriting from it to implement there own methods. Helper classes were added to each package to help abstract common logic.

### DIFF
--- a/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
@@ -334,10 +334,6 @@ public final class EncryptedBlobClientBuilder {
      * with blobs in the root container, it is best to set the endpoint to the account url and specify the blob name
      * separately using the {@link EncryptedBlobClientBuilder#blobName(String) blobName} method.</p>
      *
-     * <p><strong>Code Samples</strong></p>
-     *
-     * {@codesnippet com.azure.storage.blob.specialized.BlobClientBase.Builder.endpoint#String}
-     *
      * @param endpoint URL of the service
      * @return the updated EncryptedBlobClientBuilder object
      * @throws IllegalArgumentException If {@code endpoint} is {@code null} or is a malformed URL.
@@ -366,10 +362,6 @@ public final class EncryptedBlobClientBuilder {
 
     /**
      * Sets the name of the container that contains the blob.
-     *
-     * <p><strong>Code Samples</strong></p>
-     *
-     * {@codesnippet com.azure.storage.blob.specialized.BlobClientBase.Builder.containerName#String}
      *
      * @param containerName Name of the container. If the value {@code null} or empty the root container, {@code $root},
      * will be used.
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClientBuilder.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClientBuilder.java
@@ -235,10 +235,6 @@ public final class BlobClientBuilder {
      * with blobs in the root container, it is best to set the endpoint to the account url and specify the blob name
      * separately using the {@link BlobClientBuilder#blobName(String) blobName} method.</p>
      *
-     * <p><strong>Code Samples</strong></p>
-     *
-     * {@codesnippet com.azure.storage.blob.specialized.BlobClientBase.Builder.endpoint#String}
-     *
      * @param endpoint URL of the service
      * @return the updated BlobClientBuilder object
      * @throws IllegalArgumentException If {@code endpoint} is {@code null} or is a malformed URL.
@@ -267,10 +263,6 @@ public final class BlobClientBuilder {
 
     /**
      * Sets the name of the container that contains the blob.
-     *
-     * <p><strong>Code Samples</strong></p>
-     *
-     * {@codesnippet com.azure.storage.blob.specialized.BlobClientBase.Builder.containerName#String}
      *
      * @param containerName Name of the container. If the value {@code null} or empty the root container, {@code $root},
      * will be used.
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerClientBuilder.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerClientBuilder.java
@@ -121,10 +121,6 @@ public final class BlobContainerClientBuilder {
     /**
      * Sets the service endpoint, additionally parses it for information (SAS token, container name)
      *
-     * <p><strong>Code Samples</strong></p>
-     *
-     * {@codesnippet com.azure.storage.blob.BlobContainerClientBuilder.endpoint#String}
-     *
      * @param endpoint URL of the service
      * @return the updated BlobContainerClientBuilder object
      * @throws IllegalArgumentException If {@code endpoint} is {@code null} or is a malformed URL.
@@ -244,10 +240,6 @@ public final class BlobContainerClientBuilder {
 
     /**
      * Sets the name of the container.
-     *
-     * <p><strong>Code Samples</strong></p>
-     *
-     * {@codesnippet com.azure.storage.blob.specialized.BlobClientBase.Builder.containerName#String}
      *
      * @param containerName Name of the container. If the value {@code null} or empty the root container, {@code $root},
      * will be used.
--- a/sdk/storage/azure-storage-common/src/main/java/module-info.java
+++ b/sdk/storage/azure-storage-common/src/main/java/module-info.java
@@ -8,13 +8,13 @@ module com.azure.storage.common {
     exports com.azure.storage.common.credentials;
     exports com.azure.storage.common.policy;
 
-    exports com.azure.storage.common.implementation.credentials to
+    exports com.azure.storage.common.implementation.credentials to // FIXME this should not be a long-term solution
         com.azure.storage.blob,
         com.azure.storage.blob.cryptography,
         com.azure.storage.file,
         com.azure.storage.queue;
 
-    exports com.azure.storage.common.implementation.policy to
+    exports com.azure.storage.common.implementation.policy to // FIXME this should not be a long-term solution
         com.azure.storage.blob,
         com.azure.storage.blob.cryptography,
         com.azure.storage.file,


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -149,38 +364,38 @@ public EncryptedBlobClientBuilder endpoint(String endpoint) {
...
-
     /**
-     * Sets the name of the container this client is connecting to.
-     * @param containerName the name of the container
-     * @return the updated EncryptedBlobClientBuilder  object
+     * Sets the name of the container that contains the blob.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.specialized.BlobClientBase.Builder.containerName#String}
```
srnagar: The codesnippet tags need to be updated to reflect new classnames in all builders. 

In this instance, do we really need code sample on using a setter method?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClientBuilder.java
```diff
@@ -137,26 +266,25 @@ public BlobClientBuilder endpoint(String endpoint) {
     }
 
     /**
-     * Sets the name of the container this client is connecting to.
+     * Sets the name of the container that contains the blob.
      *
      * <p><strong>Code Samples</strong></p>
      *
      * {@codesnippet com.azure.storage.blob.specialized.BlobClientBase.Builder.containerName#String}
```
srnagar: Remove this.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerClientBuilder.java
```diff
@@ -104,8 +138,8 @@ public BlobContainerClientBuilder endpoint(String endpoint) {
             this.containerName = parts.getBlobContainerName();
 
             String sasToken = parts.getSasQueryParameters().encode();
-            if (ImplUtils.isNullOrEmpty(sasToken)) {
-                super.sasToken(sasToken);
+            if (!ImplUtils.isNullOrEmpty(sasToken)) {
```
srnagar: Was this a bug in the previously released preview?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerClientBuilder.java
```diff
@@ -116,27 +150,190 @@ public BlobContainerClientBuilder endpoint(String endpoint) {
...
+        return this;
+    }
+
+    /**
+     * Sets the name of the container.
      *
      * <p><strong>Code Samples</strong></p>
      *
-     * {@codesnippet com.azure.storage.blob.BlobContainerClientBuilder.containerName#String}
+     * {@codesnippet com.azure.storage.blob.specialized.BlobClientBase.Builder.containerName#String}
```
srnagar: This can be removed too.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/package-info.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Package containing implementation specific utilities for Azure Storage Blobs.
+ */
+package com.azure.storage.blob.implementation.util;
```
srnagar: nit: implementation packages don't require package-info.

sdk/storage/azure-storage-common/src/main/java/module-info.java
```diff
@@ -7,4 +7,16 @@
     exports com.azure.storage.common;
     exports com.azure.storage.common.credentials;
     exports com.azure.storage.common.policy;
+
+    exports com.azure.storage.common.implementation.credentials to
```
srnagar: Add a `TODO` comment to remove these later.

rickle-msft:
Why are we removing this?

alzimmermsft:
> 
> 
> Why are we removing this?

While the inheritance structure removes code duplication, which is great, it introduces a maintenance and functionality pain points.

- Any time we need to make changes to the base level builder we would need to re-release every package that depended on it, this could result in a lot more releases.
- The base builder binds some functionality to all downstream builders that they may not use (OAuth in Files, no credential support everywhere, etc).
- The base builder forces there to be protected API that a customer could inherit and override, making that public API (even if we politely ask them not to do so).
- A lot of the code de-duplication that the base builder offers could be replicated in static stand alone helper classes.
- It's not good to inherit from non-public API in public API (the Blob, File, and Queue base builders).

In the future if we think this change was wrong it is easier to add an inheritance structure than remove it.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
alzimmermsft: Removed

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClientBuilder.java
alzimmermsft: Done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerClientBuilder.java
alzimmermsft: Yes, this appears to have been release with the incorrect boolean check.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerClientBuilder.java
alzimmermsft: Done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/package-info.java
alzimmermsft: We appear to auto-generate package-infos in other implementation packages, I will go ahead and leave this for now.

sdk/storage/azure-storage-common/src/main/java/module-info.java
alzimmermsft: Done

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -50,14 +68,27 @@
...
     private String containerName;
     private String blobName;
     private String snapshot;
 
+    private SharedKeyCredential sharedKeyCredential;
+    private TokenCredential tokenCredential;
+    private SasTokenCredential sasTokenCredential;
+
+    private HttpClient httpClient;
+    private final List<HttpPipelinePolicy> additionalPolicies = new ArrayList<>();
```
conniey: nit: keeping private final and private variables grouped.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -79,13 +111,51 @@ private AzureBlobStorageImpl constructImpl() {
...
 
-        checkValidEncryptionParameters();
+        if (httpPipeline != null) {
+            return new AzureBlobStorageBuilder()
+                .url(String.format("%s/%s/%s", endpoint, containerName, blobName))
+                .pipeline(httpPipeline)
+                .build();
+        }
+
+        String userAgentName = BlobCryptographyConfiguration.NAME;
```
conniey: These are only used once. Do we need variables for these two?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClientBuilder.java
```diff
@@ -52,13 +82,21 @@ public BlobServiceClient buildClient() {
...
     public BlobServiceAsyncClient buildAsyncClient() {
-        HttpPipeline pipeline = super.getPipeline();
-        if (pipeline == null) {
-            pipeline = super.buildPipeline();
-        }
+        HttpPipeline pipeline = (httpPipeline != null) ? httpPipeline : BuilderHelper.buildPipeline(() -> {
+            if (sharedKeyCredential != null) {
+                return new SharedKeyCredentialPolicy(sharedKeyCredential);
+            } else if (tokenCredential != null) {
+                return new BearerTokenAuthenticationPolicy(tokenCredential, String.format("%s/.default", endpoint));
```
conniey: These special strings are used in multiple places, should there be some constants class?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClientBuilder.java
```diff
@@ -87,12 +125,161 @@ public BlobServiceClientBuilder endpoint(String endpoint) {
...
-    String endpoint() {
-        return super.endpoint;
+    /**
+     * Sets the configuration object used to retrieve environment configuration values during building of the client.
+     *
+     * @param configuration Configuration store used to retrieve environment configurations.
+     * @return the updated BlobServiceClientBuilder object
+     */
+    public BlobServiceClientBuilder configuration(Configuration configuration) {
+        this.configuration = configuration;
```
conniey: There are Objects.requireNotNull checks in other parameters, but not this one?

alzimmermsft: This one has a default if the value is `null` when being used.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClientBuilder.java
alzimmermsft: I think we can promote this later as it will need to go into public API.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/BuilderHelper.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+/**
+ * This class provides helper methods for common builder patterns.
+ */
+final class BuilderHelper {
```
sima-zhu: Better to put the builderHelper to implementation as well

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileClientBuilder.java
```diff
@@ -60,16 +70,30 @@
...
-
+@ServiceClientBuilder(serviceClients = {
+    FileClient.class, FileAsyncClient.class,
+    DirectoryClient.class, DirectoryAsyncClient.class
+})
+public class FileClientBuilder {
     private final ClientLogger logger = new ClientLogger(FileClientBuilder.class);
 
+    private String endpoint;
+    private String accountName;
```
sima-zhu: What's the reason to have accountName here?

sdk/storage/azure-storage-blob-batch/src/test/java/com/azure/storage/blob/batch/APISpec.groovy
```diff
@@ -231,7 +230,9 @@ class APISpec extends Specification {
             builder.addPolicy(policy)
         }
 
-        addOptionalRecording(builder)
+        if (testMode == TestMode.RECORD && recordLiveMode) {
```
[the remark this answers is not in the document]
alzimmermsft: That's what the `recordLiveMode` flag is about, there are some tests that are annotated with `@Requires(liveMode())` which won't be recorded.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/BuilderHelper.java
alzimmermsft: This one is package private so it is hidden from public API, no need to move it now.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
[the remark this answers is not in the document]
alzimmermsft: I was just following the pattern of other client builders. We can follow-up on this later as it is purely an implementation detail.

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileClientBuilder.java
alzimmermsft: We added being able to get the account name from every client. It is just another way to help move around and understanding what is being used.

joshfree:
This merge broke the storage team's ADLS feature work; I know it was opened for 3 days before being merged but please tag storage devs on open PRs in progress, while we lock down Preview 5. 😅

PR #5799, opened 2019-10-11
Updates to Poller implementation.

conniey:
* Using `AtomicReference<PollerResponse<T>>` to ensure atomic access to `pollResponse`, since this variable can be accessed from any thread.
* The result of `activationOperation`, if it is set, is set as the first `PollerResponse<T>`, so that the first pollerOperation can leverage this.
* Improves test performance and stability using StepVerifier and Mockito.
  * Also uncovered some bugs in the Poller implementation and feature gaps regarding cancellation.

### DIFF
--- a/sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
@@ -62,7 +62,7 @@ public class PollerTests {
     /**
      * Verify we cannot pass in null pollInterval.
      */
-    @Test(expected = IllegalArgumentException.class)
+    @Test(expected = NullPointerException.class)
     public void constructorPollIntervalNull() {
         new Poller<>(null, pollOperation, () -> Mono.just(new Response("Foo")), cancelOperation);
     }
@@ -70,7 +70,7 @@ public class PollerTests {
     /**
      * Verify we cannot pass in null pollInterval.
      */
-    @Test(expected = IllegalArgumentException.class)
+    @Test(expected = NullPointerException.class)
     public void constructorPollOperationNull() {
         new Poller<>(Duration.ofSeconds(1), null, () -> Mono.just(new Response("Foo")), cancelOperation);
     }


sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -81,26 +82,27 @@
     /*
      * This will save last poll response.
      */
-    private PollResponse<T> pollResponse;
+    private final AtomicReference<PollResponse<T>> pollResponse = new AtomicReference<>(
```
hemanttanwar: Good that we are using AtomicReference.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -247,10 +257,10 @@ public void cancelOperation() throws UnsupportedOperationException {
      * @return A {@link Mono} that returns {@link PollResponse}. This will call poll operation once.
      */
     public Mono<PollResponse<T>> poll() {
-        return this.pollOperation.apply(this.pollResponse)
+        return pollOperation.apply(pollResponse.get())
```
hemanttanwar: earlier in my review I received to comment to use  this for class variable. Also this which class uses this all other places. So for consistency, shouldn't we use `this.` here as well.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -374,9 +384,11 @@ private boolean matchStatus(PollResponse<T> currentPollResponse, OperationStatus
      * We will use  {@link PollResponse#getRetryAfter} if it is greater than zero otherwise use poll interval.
      */
     private Duration getCurrentDelay() {
-        return (this.pollResponse != null
-            && this.pollResponse.getRetryAfter() != null
-            && this.pollResponse.getRetryAfter().toNanos() > 0) ? this.pollResponse.getRetryAfter() : this.pollInterval;
+        final PollResponse<T> current = pollResponse.get();
```
hemanttanwar: `this.`  ?

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
```diff
@@ -336,132 +197,111 @@ public void subscribeToOnlyFinalEventSuccessfullyCompleteInNSecondsTest() throws
...
+        Poller<Response> poller = new Poller<>(pollInterval, pollOperation, null,
+            ignored -> new PollResponse<Response>(OperationStatus.USER_CANCELLED, null));
+
+        // Assert
+        StepVerifier.create(poller.getObserver())
+            .expectNext(inProgress)
+            .then(() -> poller.setAutoPollingEnabled(false))
+            .expectNoEvent(waitTime)
+            .then(() -> poller.cancelOperation())
+            .verifyError();
```
hemanttanwar: This usecase ,user stop polling after N second. I think user is not calling  `cancelOperation` . Should we remove the call to `cancelOperation()` here ?

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
```diff
@@ -221,113 +145,50 @@ public void run() {
...
-    }
-
-    /* Test where SDK Client is subscribed to only final/last response.
-     * The last response in this case will be PollResponse.OperationStatus.SUCCESSFULLY_COMPLETED
-     * This scenario is setup where source will generate successful response returned after few in progress response.
-     * But the subscriber is only interested in last response, The test will ensure subscriber
-     * only gets last PollResponse.OperationStatus.SUCCESSFULLY_COMPLETED . */
-
-    @Test
-    public void subscribeToOnlyFinalEventSuccessfullyCompleteInNSecondsTest() throws Exception {
```
hemanttanwar: Why are we removing this test ? 
This is one of the usecase which was part mentioned in requirements. Are we testing this somewhere else ?
`subscriber is only interested in last response`

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -152,7 +154,7 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
      */
     public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse<T>>> pollOperation,
                   Supplier<Mono<T>> activationOperation, Consumer<Poller<T>> cancelOperation) {
-        if (pollInterval == null || pollInterval.toNanos() <= 0) {
+        if (pollInterval == null || pollInterval.compareTo(Duration.ZERO) <= 0) {
```
srnagar: If it's null, then we must throw NPE and if the value is invalid, throw `IllegalArgumentException`.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
conniey: Not necessary. There isn't a conflict in member names versus variable names in this scope.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
conniey: I had reviews where for terseness and it does't conflict, to prefer not to use this.

conniey: I would rather remove `this.`. For me, it only adds cognitive burden.

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
conniey: Yes we are. In all the other tests, we ensure we get the last response.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -163,22 +165,29 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
...
         this.pollInterval = pollInterval;
         this.pollOperation = pollOperation;
-        this.pollResponse = new PollResponse<>(OperationStatus.NOT_STARTED, null);
+
+        // When the first item is emitted, we set the poll response to it. So the first invocation of pollOperation can
+        // leverage this value.
+        final Mono<T> onActivation = activationOperation == null
+            ? Mono.empty()
+            : activationOperation.get().map(response -> {
+                pollResponse.set(new PollResponse<>(OperationStatus.NOT_STARTED, response));
```
srnagar: `pollResponse` is already initialized with `NOT_STARTED` when the field is declared. Do we need to reinitialize it here?

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
conniey: I'll add this one back.

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
conniey: Fixed.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
srnagar: I am not sure if using AtomicReference here is adding much value as the code below doesn't really ensure atomicity.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
hemanttanwar: ok good for me.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
conniey: Removed and reverted back to volatile.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
conniey: I reverted this back.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
hemanttanwar: I think underneath ClientLogger API  does this check so it will take care of this `Objects.requireNonNull`

pom.client.xml
```diff
@@ -718,14 +718,19 @@
             <version>${maven-surefire-plugin.version}</version>
             <configuration>
               <argLine>
-                <!-- KeyVault tests fail without these exports -->
+                <!-- Mockito uses reflection to create PollResponses in tests. -->
+                --add-opens com.azure.core/com.azure.core.util.polling=ALL-UNNAMED
```
srnagar: nit: Keep all `opens` args together. There's a bunch of `opens` after `exports`.

conniey: Fixed

PR #5800, opened 2019-10-11
Align File Upload and Download Methods with Blobs

### LINKED ISSUE #5785: Align Blob and File Download And Upload Methods
The standard download and upload methods, not the specialized any size or from/to file, need to take in consistent parameters for Blobs and Files.

Files should move from using `ByteBuffer` to `InputStream`.

alzimmermsft:
Fixes #5785 

Aligns the base File upload and download methods with what Blobs is using.

### DIFF
--- a/sdk/storage/azure-storage-file/src/samples/java/com/azure/storage/file/FileAsyncJavaDocCodeSamples.java
+++ b/sdk/storage/azure-storage-file/src/samples/java/com/azure/storage/file/FileAsyncJavaDocCodeSamples.java
@@ -209,18 +209,18 @@ public class FileAsyncJavaDocCodeSamples {
     }
 
     /**
-     * Generates a code sample for using {@link FileAsyncClient#upload(Flux, long, long)}
+     * Generates a code sample for using {@link FileAsyncClient#uploadWithResponse(Flux, long, Long)}
      */
     public void uploadDataMaxOverloadAsync() {
         FileAsyncClient fileAsyncClient = createAsyncClientWithSASToken();
-        // BEGIN: com.azure.storage.file.fileAsyncClient.upload#flux-long-long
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileClient.java
```diff
@@ -334,25 +344,38 @@ public FileDownloadInfo downloadWithProperties() {
...
+
+        Mono<Response<Void>> download = fileAsyncClient.downloadWithResponse(range, rangeGetContentMD5, context)
+            .flatMapMany(response -> response.getValue().doOnNext(buffer -> {
+                try {
+                    stream.write(FluxUtil.byteBufferToArray(buffer));
+                } catch (IOException ex) {
+                    throw logger.logExceptionAsError(new UncheckedIOException(ex));
+                }
+            }).map(buffer -> response))
+            .last()
```
anuchandy: If `response.getValue()` emits `Flux.empty()` then `last()` operator can throw ` NoSuchElementException` Ref: https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#last--

anuchandy: I guess we could simplify it a bit by using flatmap() and reduce() something like below (untested):

```java
Mono<Void> download = fileAsyncClient.downloadWithResponse(range, rangeGetContentMD5, context)
  .flatmap(response -> response.getValue().reduce(() -> stream, (stream, bb) -> { do write;, return stream; });
```

alzimmermsft: Updated to use this

PR #5804, opened 2019-10-12
Ported NPE from v2

kushagraThapar:
This PR is a port of v2 fix: https://github.com/Azure/azure-cosmosdb-java/pull/267


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/directconnectivity/AddressResolver.java
```diff
@@ -548,6 +551,7 @@ private ResolutionResult handleRangeAddressResolutionFailure(
             throw BridgeInternal.setResourceAddress(new PartitionKeyRangeGoneException(errorMessage), request.getResourceAddress());
         }
 
+        logger.debug("handleRangeAddressResolutionFailure returns null");
         return null;
```
moderakh: is this return null correct?

kushagraThapar: Yes, this null return is correct :)

mitchdenny:

Issue #5699, opened 2019-10-04
[ArchBoard] Merge auto-generated and exposed models in Storage


conniey:
There are autogenerated model classes from the storage REST APIs that have the same name as the models exposed in the client library. These are confusing because customers may not know which ones to use. (They don't live in the .implementation package.)

This also results in a spotbugs issue that we've had to suppressed for a few of them.
> [ERROR] Confusing to have methods com.azure.storage.blob.BlobContainerProperties.getETag() and com.azure.storage.blob.models.BlobContainerProperties.getEtag() [com.azure.storage.blob.BlobContainerProperties] At BlobContainerProperties.java:[line 50] NM_CONFUSING

Example:
* https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerProperties.java#L49
* https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerProperties.java#L100

joshfree:
@sima-zhu can you please drive this to resolution for Preview 5?

PR #5808, opened 2019-10-13
Renaming autogen models and  applying x-ms-client-name for etag property

anuchandy:
We generate `com.azure.storage.blob.models.BlobProperties` which is composed in `BlobItem` model. `BlobItem` represents an item in LIST blobs API.

We manually author `com.azure.storage.blob.BlobProperties` which represents result of GET blob API.

Models with same name in different namespaces can create user confusion. Merging two into one model type is Risky.  Using same type can result in subset of properties to be present or absent depending on which API is used. e.g. service may return less information about the resource in LIST and full information in GET.

This PR renames - autogenerated `com.azure.storage.blob.models.BlobProperties` to `com.azure.storage.blob.models.BlobItemProperties` (as it is composed in `BlobItem`).

Following the same convention autogenerated `com.azure.storage.blob.models.BlobContainerProperties` is renamed in `com.azure.storage.blob.models.BlobContainerItemProperties` (as it is composed in BlobContainerItem).

This PR also applies x-ms-client name attribute to `etag` properties, that name it to `eTag`.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerItemProperties.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerItemProperties.java
@@ -13,9 +13,9 @@ import java.time.OffsetDateTime;
 /**
  * Properties of a container.
  */
-@JacksonXmlRootElement(localName = "BlobContainerProperties")
+@JacksonXmlRootElement(localName = "Properties")
 @Fluent
-public final class BlobContainerProperties {
+public final class BlobContainerItemProperties {
     /*
      * The lastModified property.
      */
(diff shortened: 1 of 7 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerItemProperties.java
```diff
@@ -13,9 +13,9 @@
 /**
  * Properties of a container.
  */
-@JacksonXmlRootElement(localName = "BlobContainerProperties")
```
alzimmermsft: This was released using the wrong XML property wasn't it?

alzimmermsft:
Should the eTag naming be pushed as a change to the swagger file?

alzimmermsft:
Mind referencing the GitHub issues fixed here so we can close them when this merged?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerItemProperties.java
anuchandy: Yes, it was supposed to be "Properties"

anuchandy:
Yes, we need to propagate etag x-ms-client-name to swagger, will sync with storage team on that.

anuchandy:
Fixes: https://github.com/Azure/azure-sdk-for-java/issues/5699

anuchandy:
@alzimmermsft - created a swagger issue to track etag naming https://github.com/Azure/azure-sdk-for-java/issues/5815, will remove it from transformtion readme when swagger work is done.

PR #5820, opened 2019-10-14
poller upgrade

g2vinay:
Migrates Poller from KV Certs API to core.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
@@ -80,7 +80,7 @@ public final class PollResponse<T> {
          * Creates or finds a {@link OperationStatus} from its string representation.
          * @param name a name to look for
          * @param isComplete a status to indicate if the operation is complete or not.
-         * @throws IllegalArgumentException  if {@code name} matches a pre-configured {@link OperationStatus} but
+         * @throws IllegalArgumentException if {@code name} matches a pre-configured {@link OperationStatus} but
          * {@code isComplete} doesn't match its pre-configured complete status.
          * @return the corresponding {@link OperationStatus}
          */
--- a/sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
@@ -21,7 +21,6 @@ import java.util.List;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
-import static junit.framework.TestCase.assertEquals;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;


sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -86,14 +87,20 @@
...
      * Indicate to poll automatically or not when poller is created.
      * default value is false;
      */
     private boolean autoPollingEnabled;
 
+    /*
+     * This will be called when final result needs to be retrieved after polling has completed.
+     */
+    private final Supplier<Mono<R>> fetchResultOperation;
+
```
conniey: nit: unneeded extra line.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -122,11 +129,13 @@
      *     operation, it should be handled by the client library and return a valid {@link PollResponse}. However if
      *     the poll operation returns {@link Mono#error(Throwable)}, the {@link Poller} will disregard it and continue
      *     to poll.
+     * @param fetchResultOperation the operation to be called to fetch final result after polling has been completed.
```
conniey:
```suggestion
     * @param fetchResultOperation The operation to be called to fetch final result after polling has been completed.
```

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -122,11 +129,13 @@
...
      * @throws IllegalArgumentException if {@code pollInterval} is less than or equal to zero and if
      *     {@code pollInterval} or {@code pollOperation} are {@code null}
      */
-    public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse<T>>> pollOperation) {
-        this(pollInterval, pollOperation, null, null);
+    public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse<T>>> pollOperation,
```
conniey: Is it required to supply a final fetch operation? I can see the use case in create certificate but we don't want to fetch the blob when it's complete copying.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -219,11 +232,22 @@ public void cancelOperation() throws UnsupportedOperationException {
...
-        this.cancelOperation.accept(this);
+        return this.cancelOperation.apply(this);
+    }
+
+    /**
+     * Returns the final result if the polling operation has been completed.
+     * @return A {@link Mono} containing the final output.
+     */
+    public Mono<R> result() {
+        if (!getStatus().equals(PollResponse.OperationStatus.SUCCESSFULLY_COMPLETED)) {
```
conniey: getStatus can be equal to null.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -219,11 +232,22 @@ public void cancelOperation() throws UnsupportedOperationException {
...
+            return Mono.error(new IllegalAccessException("The poll operation has not successfully completed."));
```
conniey: Is IllegalAccess the correct exception?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -219,11 +232,22 @@ public void cancelOperation() throws UnsupportedOperationException {
...
+        }
+        return fetchResultOperation.get();
```
conniey: Possible to have a null fetchResultOperation.

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
```diff
@@ -11,21 +11,24 @@
...
 
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static org.junit.Assert.*;
```
conniey: Remove import `*`

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
```diff
@@ -11,21 +11,24 @@
...
 
 public class PollerTests {
 
     private final ClientLogger logger = new ClientLogger(PollerTests.class);
     private boolean debug = true;
     int count;
+    private final String CERTIFICATE_NAME = "CERTIFICATE_NAME";
```
conniey: private static final

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
```diff
@@ -51,6 +54,11 @@
         };
     }
 
+    private Supplier<Mono<CertificateOutput>>
```
conniey: use `@Mock`.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
```diff
@@ -37,35 +37,45 @@
      * {@code USER_CANCELLED} or {@code FAILED}.
      */
     public static final class OperationStatus extends ExpandableStringEnum<OperationStatus> {
+
+        private boolean completed;
```
conniey: Is this necessary? can it be inferred from OperationStatus?

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
g2vinay: removed

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
g2vinay: fixed.

sdk/core/azure-core/src/test/java/com/azure/core/util/polling/PollerTests.java
g2vinay: its a method

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
conniey: In Processor not? it seems inferred from your usages.

conniey: I'm curious why not? it seems like you're inferring it from those status' below.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
```diff
@@ -37,35 +37,46 @@
...
         /**
          * Creates or finds a {@link OperationStatus} from its string representation.
          * @param name a name to look for
+         * @param isComplete a status to indicate if the operation is complete or not.
          * @return the corresponding {@link OperationStatus}
          */
-        public static OperationStatus fromString(String name) {
-            return fromString(name, OperationStatus.class);
+        public static OperationStatus fromString(String name, boolean isComplete) {
+            OperationStatus status =  fromString(name, OperationStatus.class);
```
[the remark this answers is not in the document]
conniey: nit: double space.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -82,7 +83,18 @@
     /*
      * This will be called when cancel operation is triggered.
      */
-    private final Consumer<Poller<T>> cancelOperation;
+    private final Function<Poller<T, R>, Mono<T>> cancelOperation;
+
+    /*
```
conniey: why not keep these together?

conniey: final and member variables.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -215,22 +228,35 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
...
         final PollResponse<T> response = this.pollResponse;
         if (response != null && response.getStatus() != OperationStatus.IN_PROGRESS) {
-            return;
+            return Mono.empty();
         }
 
-        // Time to call cancel
-        this.cancelOperation.accept(this);
+        //Time to call cancel
+        return this.cancelOperation.apply(this);
```
conniey: What is fcancelOepration throws an exception>

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -146,21 +154,25 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
...
      * @throws IllegalArgumentException if {@code pollInterval} is less than or equal to zero and if
      *      {@code pollInterval} or {@code pollOperation} are {@code null}
      */
     public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse<T>>> pollOperation,
-                  Supplier<Mono<T>> activationOperation, Consumer<Poller<T>> cancelOperation) {
+                  Supplier<Mono<R>> fetchResultOperation, Supplier<Mono<T>> activationOperation,
+                  Function<Poller<T, R>, Mono<T>> cancelOperation) {
 
         Objects.requireNonNull(pollInterval, "'pollInterval' cannot be null.");
+        Objects.requireNonNull(fetchResultOperation, "'fetchResultOperation' cannot be null.");
```
conniey: What do you expect it to be in cases where we don't care about a fetch operation?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
g2vinay: if some user supplies 'TOTALLY_COMPLETED' as final status, then we do not know that is the final status.
Poller.isComplete relies on this boolean too now to signal is Poller completed or not.

conniey: Thanks for clarifying!

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: () -> Mono.empty
This should be enough for fetch operation, if R is Void in Poller<T,R>

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: If its a REST call in cancelOperation then the behavoiur will be same as an API call.
User can specify .onError on the Mono returned to them from cancelOperation.
if needed in CancelOperation, Mono.error should be returned there.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: grouped together

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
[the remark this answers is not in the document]
g2vinay: fixed.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
```diff
@@ -37,35 +37,48 @@
...
 
         /**
          * Creates or finds a {@link OperationStatus} from its string representation.
          * @param name a name to look for
+         * @param isComplete a status to indicate if the operation is complete or not.
          * @return the corresponding {@link OperationStatus}
          */
-        public static OperationStatus fromString(String name) {
-            return fromString(name, OperationStatus.class);
+        public static OperationStatus fromString(String name, boolean isComplete) {
```
srnagar: How does this work if the user provided a string `IN_PROGRESS` and `true`? We should at least have a check to ensure the user-provided string doesn't match any status that is defined by us. If it does, then the boolean value should also match.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -198,14 +210,15 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
...
+     * @param fetchResultOperation The operation to be called to fetch final result after polling has been completed.
      * @param cancelOperation cancel operation if cancellation is supported by the service. It can be {@code null}
      *      which will indicate to the {@link Poller} that cancel operation is not supported by Azure service.
      * @throws IllegalArgumentException if {@code pollInterval} is less than or equal to zero and if
      * {@code pollInterval} or {@code pollOperation} are {@code null}
      */
     public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse<T>>> pollOperation,
-                  Consumer<Poller<T>> cancelOperation) {
-        this(pollInterval, pollOperation, null, cancelOperation);
+                  Supplier<Mono<R>> fetchResultOperation, Function<Poller<T, R>, Mono<T>> cancelOperation) {
```
srnagar: I see Poller being returned in sync clients too. Do we need to have Pollers that take Functions/Suppliers that are not `Mono`?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -215,22 +228,35 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
      * It will call cancelOperation if status is {@link OperationStatus#IN_PROGRESS} otherwise it does nothing.
      *
      * @throws UnsupportedOperationException when the cancel operation is not supported by the Azure service.
+     * @return A {@link Mono} containing the response.
```
srnagar: Make this clear that this is the poller response (T) and not the final response (R).

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -215,22 +228,35 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
...
      */
-    public void cancelOperation() throws UnsupportedOperationException {
+    public Mono<T> cancelOperation() throws UnsupportedOperationException {
```
srnagar: Anything returning a Publisher shouldn't have `throws`. It should be propagated through the error channel.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -175,7 +187,7 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
         this.fluxHandle = asyncPollRequestWithDelay()
             .flux()
             .repeat()
-            .takeUntil(pollResponse -> hasCompleted())
+            .takeUntil(pollResponse -> isComplete())
             .share()
             .delaySubscription(onActivation);
```
conniey: One thing I was wondering is if the onActivation fails, should we keep polling? delaySubscription only delays it.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: fixed.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: updated

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: nope,
In Sync clients, user will choose to block / block until,  if they want.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
g2vinay: added the check

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
```diff
@@ -37,35 +37,59 @@
...
+         * @throws IllegalArgumentException if invalid {2code isComplete} is provided for a pre-configured
+         * {@link OperationStatus} with {@code name}
          * @return the corresponding {@link OperationStatus}
          */
-        public static OperationStatus fromString(String name) {
-            return fromString(name, OperationStatus.class);
+        public static OperationStatus fromString(String name, boolean isComplete) {
+            OperationStatus status = fromString(name, OperationStatus.class);
+
+            if (status != null) {
```
conniey: if status == null, I think we should throw something because this should not happen. Not sure what you'd do with a null status.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
```diff
@@ -37,35 +37,59 @@
...
+                for (OperationStatus opStatus : values(OperationStatus.class)) {
```
conniey: Why a for loop? The one returned `fromString(name, OperationStatus.class)` will return one from the dictionary if it was added before. Then you can do:
```java
OperationStatus status = fromString(name, OperationStatus.class);

if (status.isComplete() != isComplete) {
  // throw.
}

return status;
```

g2vinay: The intention is to check against only defined by us.
So replaced with a static map and check against it.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
g2vinay: looking at expandable String enum code, it returns null if String name is null.
So, we are returning null for null name input, across all classes who extend ExpandableStringEnum.
This code will be consistent too.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
```diff
@@ -37,35 +39,69 @@
...
+            opStatusMap.put(FAILED.toString(), FAILED);
+            opStatusMap.put(USER_CANCELLED.toString(), USER_CANCELLED);
+            operationStatusMap = Collections.unmodifiableMap(opStatusMap);
+        }
 
         /**
          * Creates or finds a {@link OperationStatus} from its string representation.
          * @param name a name to look for
+         * @param isComplete a status to indicate if the operation is complete or not.
+         * @throws IllegalArgumentException if invalid {2code isComplete} is provided for a pre-configured
```
conniey: `{2code isComplete} ` -> `{@code isComplete} `

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
JonathanGiles: This still appears to throw?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
```diff
@@ -37,35 +39,69 @@
...
+            opStatusMap.put(USER_CANCELLED.toString(), USER_CANCELLED);
+            operationStatusMap = Collections.unmodifiableMap(opStatusMap);
+        }
 
         /**
          * Creates or finds a {@link OperationStatus} from its string representation.
          * @param name a name to look for
+         * @param isComplete a status to indicate if the operation is complete or not.
+         * @throws IllegalArgumentException if invalid {@code isComplete} is provided for a pre-configured
+         * {@link OperationStatus} with {@code name}
```
JonathanGiles: This throws statement isn't clear enough for an end user.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -215,22 +233,35 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
...
         }
 
-        // Time to call cancel
-        this.cancelOperation.accept(this);
+        //Time to call cancel
+        return this.cancelOperation.apply(this);
+    }
+
+    /**
+     * Returns the final result if the polling operation has been completed.
```
JonathanGiles: There should be documentation here about what happens if the operation has not been completed and this method is called.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -215,22 +233,35 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
...
+     * @return A {@link Mono} containing the final output.
+     */
+    public Mono<R> result() {
```
JonathanGiles: Why isn't this getResult()?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
JonathanGiles: I don't understand enough yet, but almost certainly this is not the right exception type.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -283,13 +315,14 @@ public void cancelOperation() throws UnsupportedOperationException {
      * <p>It will enable auto-polling if it was disable by user.
      *
      * @param timeout The duration for which execution is blocked and waits for polling to complete.
-     * @return returns final {@link PollResponse} when polling is complete as defined in {@link OperationStatus}.
+     * @return The final output once Polling completes.
```
JonathanGiles:
```suggestion
     * @return The final output once polling completes.
```

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -268,13 +299,14 @@ public void cancelOperation() throws UnsupportedOperationException {
      *
      * <p>It will enable auto-polling if it was disabled by the user.
      *
-     * @return A {@link PollResponse} when polling is complete.
+     * @return The final output once Polling completes.
```
JonathanGiles: Polling is not a noun.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollResponse.java
g2vinay: updated

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: updated

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: valid point, 
added get prefix.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: updated

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: updated

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
g2vinay: the get prefix will take some features to become a habit.

PR #5826, opened 2019-10-14
Adds Poller support for BlobAsyncClient and BlobClient beginCopyFromUrl

### LINKED ISSUE #5643: [API] Using polling strategy for startCopyFromURL
Might need to merge startCopy and abortCopy in one API.

### LINKED ISSUE #5755: [QUERY] Copy blob from existing location to new location
**Query/Question**
Is there a method to perform a copy operation existing in blob storage to another location? It appears that there is not a function that does so.

http://azure.github.io/azure-sdk-for-java/track2reports/index.html

https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob

***Why is this not a Bug or a Feature Request?***
This appears to be a basic feature that exists in the blob storage library on Azure Functions.

As a reference, the REST API docs lists the operation I would like:

https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob

Looking at the Java Docs it doesn't appear there is anything similar to the REST API or the Blob library on azure functions `StartCopyAsync`

Or is the copy from url method smart enough to switch context?

For more information, I'm trying to move files quickly but I noticed that Databricks DBFS move is taking longer for larger filer - my hunch is it isn't performing a native azure blob move.

conniey:
* Changes startCopyFromURL to beginCopyFromUrl. This conforms with our Java design guidelines.
* Adds support for `Poller` API by exposing `BlobCopyOperation`.
  * Every polling interval, `getBlobInfo()` is called to see if the copy status is complete.
  * To cancel the copy operation, `abortCopyFromUrl` is called.

### TODOs
- [x] Update FileAsyncClient/FileClient
- [x] Add tests

Fixes #5643
Fixes #5755

### DIFF
--- a/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileAsyncClient.java
+++ b/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileAsyncClient.java
@@ -17,7 +17,6 @@ import com.azure.core.util.logging.ClientLogger;
 import com.azure.core.util.polling.PollResponse;
 import com.azure.core.util.polling.PollResponse.OperationStatus;
 import com.azure.core.util.polling.Poller;
-import com.azure.storage.common.Utility;
 import com.azure.storage.common.StorageSharedKeyCredential;
 import com.azure.storage.common.implementation.Constants;
 import com.azure.storage.common.implementation.StorageImplUtils;
--- a/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileClient.java
+++ b/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/FileClient.java
@@ -11,7 +11,6 @@ import com.azure.core.implementation.util.FluxUtil;
 import com.azure.core.util.Context;
 import com.azure.core.util.logging.ClientLogger;
 import com.azure.core.util.polling.Poller;
-import com.azure.storage.common.Utility;
 import com.azure.storage.common.StorageSharedKeyCredential;
 import com.azure.storage.common.Utility;
 import com.azure.storage.common.implementation.StorageImplUtils;
--- a/sdk/storage/azure-storage-file/src/samples/java/com/azure/storage/file/FileAsyncJavaDocCodeSamples.java
+++ b/sdk/storage/azure-storage-file/src/samples/java/com/azure/storage/file/FileAsyncJavaDocCodeSamples.java
@@ -138,9 +138,9 @@ public class FileAsyncJavaDocCodeSamples {
             Collections.singletonMap("file", "metadata"), Duration.ofSeconds(2));
 
         poller.getObserver().subscribe(response -> {
-                final FileCopyInfo value = response.getValue();
-                System.out.printf("Copy source: %s. Status: %s.%n", value.getCopySourceUrl(), value.getCopyStatus());
-            }, error -> System.err.println("Error: " + error),
+            final FileCopyInfo value = response.getValue();
+            System.out.printf("Copy source: %s. Status: %s.%n", value.getCopySourceUrl(), value.getCopyStatus());
+        }, error -> System.err.println("Error: " + error),
             () -> System.out.println("Complete copying the file."));
         // END: com.azure.storage.file.fileAsyncClient.beginCopy#string-map-duration
     }
--- a/sdk/storage/azure-storage-file/src/samples/java/com/azure/storage/file/FileJavaDocCodeSamples.java
+++ b/sdk/storage/azure-storage-file/src/samples/java/com/azure/storage/file/FileJavaDocCodeSamples.java
@@ -68,7 +68,6 @@ public class FileJavaDocCodeSamples {
         return fileClient;
     }
 
-
     /**
      * Generates code sample for creating a {@link FileClient} with SAS token.
      *
@@ -152,9 +151,9 @@ public class FileJavaDocCodeSamples {
             Collections.singletonMap("file", "metadata"), Duration.ofSeconds(2));
 
         poller.getObserver().subscribe(response -> {
-                final FileCopyInfo value = response.getValue();
-                System.out.printf("Copy source: %s. Status: %s.%n", value.getCopySourceUrl(), value.getCopyStatus());
-            });
+            final FileCopyInfo value = response.getValue();
+            System.out.printf("Copy source: %s. Status: %s.%n", value.getCopySourceUrl(), value.getCopyStatus());
+        });
         // END: com.azure.storage.file.fileClient.beginCopy#string-map-duration
     }
 


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobCopyOperation.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Objects;
+
+/**
+ * The status of a long running Azure Blob copy operation.
+ *
+ * @see BlobAsyncClient#beginCopyFromUrl(URL)
+ * @see BlobClient#beginCopyFromUrl(URL)
+ */
+@Immutable
+public class BlobCopyOperation {
```
conniey: Should I update this to BlobCopyInfo to match [FileCopyInfo](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/models/FileCopyInfo.java)?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobCopyInfo.java
```diff
@@ -0,0 +1,109 @@
...
+ * @see BlobClient#beginCopyFromUrl(URL)
+ */
+@Immutable
+public class BlobCopyInfo {
+    private final String copyId;
+    private final String copySource;
+    private final CopyStatusType copyStatus;
+    private final String eTag;
+    private final OffsetDateTime lastModified;
+    private final String error;
```
sima-zhu: Is error needed here? Storage will throw error to StorageException. It is not needed for hand craft model.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -234,65 +244,124 @@ public boolean isSnapshot() {
...
+     * @param sourceUrl The source URL to copy from. URLs outside of Azure may only be copied to block blobs.
      * @param metadata Metadata to associate with the destination blob.
      * @param tier {@link AccessTier} for the destination blob.
      * @param priority {@link RehydratePriority} for rehydrating the blob.
-     * @param sourceModifiedAccessConditions {@link ModifiedAccessConditions} against the source. Standard HTTP Access
-     * conditions related to the modification of data. ETag and LastModifiedTime are used to construct conditions
-     * related to when the blob was changed relative to the given request. The request will fail if the specified
-     * condition is not satisfied.
+     * @param sourceModifiedAccessConditions {@link ModifiedAccessConditions} against the source. Standard HTTP
+     *     Access conditions related to the modification of data. ETag and LastModifiedTime are used to construct
```
sima-zhu: No space here.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -234,65 +244,124 @@ public boolean isSnapshot() {
...
+        return new Poller<>(Duration.ofSeconds(1),
+            response -> onPoll(response),
+            () -> {
+                final Mono<BlobsStartCopyFromURLResponse> restResponse = withContext(
+                    context -> azureBlobStorage.blobs().startCopyFromURLWithRestResponseAsync(null, null,
+                        sourceUrl, null, metadata, tier, priority, null, sourceConditions,
+                        destinationAccessConditions.getModifiedAccessConditions(),
+                        destinationAccessConditions.getLeaseAccessConditions(), context));
+
+                return postProcessResponse(restResponse)
```
sima-zhu: FYI, we remove postProcessResponse in another PR.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -234,65 +244,124 @@ public boolean isSnapshot() {
...
-            .setSourceIfNoneMatch(sourceModifiedAccessConditions.getIfNoneMatch());
+        final SourceModifiedAccessConditions sourceConditions = new SourceModifiedAccessConditions()
+            .setSourceIfModifiedSince(sourceModifiedCondition.getIfModifiedSince())
+            .setSourceIfUnmodifiedSince(sourceModifiedCondition.getIfUnmodifiedSince())
+            .setSourceIfMatch(sourceModifiedCondition.getIfMatch())
+            .setSourceIfNoneMatch(sourceModifiedCondition.getIfNoneMatch());
+
+        final AtomicReference<String> copyId = new AtomicReference<>();
+
+        return new Poller<>(Duration.ofSeconds(1),
```
sima-zhu: Where is the duration of 1s from?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -234,65 +244,124 @@ public boolean isSnapshot() {
...
-            destAccessConditions.getModifiedAccessConditions(), destAccessConditions.getLeaseAccessConditions(),
-            context))
-            .map(rb -> new SimpleResponse<>(rb, rb.getDeserializedHeaders().getCopyId()));
+    private Mono<PollResponse<BlobCopyInfo>> onPoll(PollResponse<BlobCopyInfo> pollResponse) {
+        if (pollResponse.getStatus() == OperationStatus.SUCCESSFULLY_COMPLETED
+            || pollResponse.getStatus() == OperationStatus.FAILED) {
+            return Mono.just(pollResponse);
+        }
+
+        return getProperties().map(response -> {
```
sima-zhu: Is there a maximum number of retries for onPoll operations? Based on poller doc, if the getProperties failed, then the poller will continue to poll. We need to stop somehow in this case.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobCopyInfo.java
conniey: Ah yes. Its not an actual exception. The service call will succeed. The get blob info call returns a set of headers, and it can contain a copy error description, if the copy failed.

sima-zhu: In case of failure, we get StorageException before you construct this class. Then async emitted a Mono.error()
I did not see any use of the error.

conniey: If we started a copy operation, and 50% of the way it fails, then it'll set this status description. When we poll, getFileProperties() will return successfully, but have this header information populated. There won't be a StorageException thrown.

https://docs.microsoft.com/en-us/rest/api/storageservices/get-file-properties#response

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
conniey: Fixed. It'll end polling if it is unable to get the file properties.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
conniey: This is borrowed from Key Vault as their certificate creations take an average of 15 seconds to create. I have no strong opinions on the polling interval.

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/specialized/BlobClientBaseJavaDocCodeSnippets.java
```diff
@@ -59,12 +62,15 @@ public void existsCodeSnippet() {
...
 
     /**
-     * Code snippets for {@link BlobClientBase#startCopyFromURL(URL)}
+     * Code snippets for {@link BlobClientBase#beginCopyFromUrl(URL)}
      */
     public void startCopyFromURL() {
-        // BEGIN: com.azure.storage.blob.specialized.BlobClientBase.startCopyFromURL#URL
-        System.out.printf("Copy identifier: %s%n", client.startCopyFromURL(url));
-        // END: com.azure.storage.blob.specialized.BlobClientBase.startCopyFromURL#URL
+        // BEGIN: com.azure.storage.blob.specialized.BlobClientBase.beginCopyFromUrl#URL
```
alzimmermsft: We add code snippets for aborting/cancelling the copy operation.

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/specialized/BlobClientBaseJavaDocCodeSnippets.java
```diff
@@ -59,12 +62,15 @@ public void existsCodeSnippet() {
...
+        final Poller<BlobCopyInfo, Void> poller = client.beginCopyFromUrl(url);
+        poller.getObserver().subscribe(response -> {
+            System.out.printf("Copy identifier: %s%n", response.getValue().getCopyId());
```
alzimmermsft: We should also add the copy status into this.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -233,64 +242,142 @@ public boolean isSnapshot() {
...
+            final CopyStatusType status = response.getCopyStatus();
+            final BlobCopyInfo result = new BlobCopyInfo(response.getCopySource(), response.getCopyId(), status,
+                response.getETag(), response.getCopyCompletionTime(), response.getCopyStatusDescription());
+
+            OperationStatus operationStatus;
+            switch (status) {
+                case SUCCESS:
+                    operationStatus = OperationStatus.SUCCESSFULLY_COMPLETED;
+                    break;
+                case FAILED:
```
alzimmermsft: Does `FAILED` really correspond to `NOT_STARTED`?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -233,64 +242,142 @@ public boolean isSnapshot() {
...
+        }
+
+        return getProperties().map(response -> {
+            final CopyStatusType status = response.getCopyStatus();
+            final BlobCopyInfo result = new BlobCopyInfo(response.getCopySource(), response.getCopyId(), status,
+                response.getETag(), response.getCopyCompletionTime(), response.getCopyStatusDescription());
+
+            OperationStatus operationStatus;
+            switch (status) {
+                case SUCCESS:
```
alzimmermsft: Should these be `CopyStatusType.SUCCESS`?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -234,65 +244,124 @@ public boolean isSnapshot() {
...
+                            headers.getETag(), headers.getLastModified(), headers.getErrorCode());
+                    });
+            },
+            unused -> {
+                final String copyIdentifier = copyId.get();
+
+                if (!ImplUtils.isNullOrEmpty(copyIdentifier)) {
+                    logger.info("Cancelling copy operation for copy: {}", copyIdentifier);
+
+                    Utility.blockWithOptionalTimeout(abortCopyFromURL(copyIdentifier), null);
```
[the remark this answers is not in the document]
conniey: This should be changed to use a mono now.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
conniey: I added a static import for CopyStatusType, so it doesn't need it.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobCopyInfo.java
sima-zhu: Do you mean at the time of failure, we have the chance of getting status code of 2xx?
Then why don't you construct the class use `headers.getCopyStatusDescription` instead of `headers.getErrorCode()`?

conniey: No. The service call is "getBlobProperties". The call passes when it is able to get the blob information.

The error is inside the BlobProperties itself, where there is metadata about copy status, if there was something being copied. The REST API explains it.

Issue #276, opened 2013-03-01
Enable testing the validation of the priority of GetJob


guangyang:
Dev estimate: 0
Test estimate: 0

Moved from https://github.com/WindowsAzure/azure-sdk-for-java-pr/issues/464.
Hi nimbus team, 
This is an interesting behavior, I am wondering whether this is by design. After creating a job with priority 3, we run a get job, the priority become 0. Job entity returned from create job request is p3, but 100 ms later, the job entity’s priority becomes 0. This is blocking one of our unit test scenario. 

Here are the fiddler traces of both of them 

Returned from create job .

``` xml
[...]

Issue #275, opened 2013-03-01
Media Services How To


guangyang:
Moved from https://github.com/WindowsAzure/azure-sdk-for-java-pr/issues/466.
Dev Estimate: 1
Test Estimate: 5
Writer Estimate: 20
PM Estimate: 5

Acceptance Criteria:
A how to guide which documents not only all of the calls of the Media Services Java library, but also covers the logical groupings of Uploading a File, Processing an Asset, and Downloading a File.

PR #5829, opened 2019-10-14
Resolves issues 275 and 276

David-Noble-at-work:
This change resolves:

- #275 Direct TCP: RntbdClientChannelPool should not warn about channel closures due to inactivity needs-v3-port
- #276 Bump netty.version to 4.1.42.Final and netty-tcnative.version to 2.0.26.Final


mitchdenny:

PR #5833, opened 2019-10-14
 more API review  changes

hemanttanwar:
Based on review comment on : https://apiview.azurewebsites.net/Assemblies/Review/2a23558e091248958cbfd49125e58cb7
1. Change Uri -> Url
2. rename function name 
3. Use Duration instead of long.

### DIFF
--- a/sdk/identity/azure-identity/src/main/java/com/azure/identity/DeviceCodeChallenge.java
+++ b/sdk/identity/azure-identity/src/main/java/com/azure/identity/DeviceCodeChallenge.java
@@ -16,7 +16,7 @@ public class DeviceCodeChallenge {
     /**
      * Creates an instance of a device code challenge.
      *
-     * @param userCode code which user needs to provide when authenticating at the verification URI
+     * @param userCode code which user needs to provide when authenticating at the verification URL
      * @param deviceCode code which should be included in the request for the access token
      * @param verificationUrl URL where user can authenticate
      * @param expiresIn expiration time of device code in seconds
@@ -43,7 +43,7 @@ public class DeviceCodeChallenge {
 
 
     /**
-     * @return code which user needs to provide when authenticating at the verification URI.
+     * @return code which user needs to provide when authenticating at the verification URL.
      */
     public String getUserCode() {
         return userCode;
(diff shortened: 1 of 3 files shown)


sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/AadCredentialBuilderBase.java
```diff
@@ -13,7 +13,7 @@
 
     /**
      * Specifies the Azure Active Directory endpoint to acquire tokens.
-     * @param authorityHost the Azure Active Directory endpoint
+     * @param authorityHost the Azure Active Directory endpredirectUrioint
```
jianghaolu: typo?

sdk/identity/azure-identity/src/main/java/com/azure/identity/credential/AuthorizationCodeCredentialBuilder.java
```diff
@@ -39,11 +39,11 @@ public AuthorizationCodeCredentialBuilder authorizationCode(String authCode) {
      * to complete the {@link AuthorizationCodeCredential} construction from there.
      * This is also called Reply URLs in some contexts.
      *
-     * @param redirectUri the redirect URI to send the authorization code
+     * @param redirectUrl the redirect URL to send the authorization code
```
jianghaolu: The issue with this change is that this does not match the name in other places / libraries Microsoft publishes e.g. https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/src/main/java/com/microsoft/aad/msal4j/AuthorizationCodeParameters.java#L41.

@JonathanGiles

JonathanGiles: From the reviews we did 'offline', I still lean towards URL.

abagarwams:
Not sure what this is about?

sdk/identity/azure-identity/src/main/java/com/azure/identity/DeviceCodeChallenge.java
```diff
@@ -18,31 +18,27 @@
      *
      * @param userCode code which user needs to provide when authenticating at the verification URI
      * @param deviceCode code which should be included in the request for the access token
-     * @param verificationUri URI where user can authenticate
+     * @param verificationUrl URI where user can authenticate
```
conniey:
```suggestion
     * @param verificationUrl URL where user can authenticate
```

jianghaolu: Fixed here and everywhere

PR #5834, opened 2019-10-14
Removing test entities classes from azure-core and moving into test directory (and into azure-core-test)

JonathanGiles:


pom.client.xml
```diff
@@ -718,6 +718,9 @@
             <version>${maven-surefire-plugin.version}</version>
             <configuration>
               <argLine>
+                <!-- Azure-core needs to open the test entities to Jackson for unit tests to operate -->
+                --add-opens com.azure.core/com.azure.core.implementation.entities=com.fasterxml.jackson.databind
```
srnagar: nit: move `opens` to the bottom along with other `opens` args.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/implementation/entities/AccessPolicy.java
```diff
@@ -0,0 +1,92 @@
...
+package com.azure.core.test.implementation.entities;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.time.OffsetDateTime;
+
+/**
+ * An Access policy.
+ */
+public class AccessPolicy {
```
srnagar: I see this duplicated in azure-core src/test as well. We should look into moving any tests in azure-core that require this class to azure-core-test later.

Issue #5779, opened 2019-10-11
[Layering] Need for ProxyAuthenticationPolicy 


anuchandy:
Today azure-core has [ProxyAuthenticationPolicy](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/ProxyAuthenticationPolicy.java) type, I felt that we can remove this type for GA since HttpClients already have a way to specify proxy. In addition to that HttpClient support different type of proxies Http/Socks[4/5]. Socks[4/5] cannot support through ProxyAuthenticationPolicy, it is only for Http. 

@JonathanGiles if you think this can be removed then I can open PR, which also enables any configuration to be specified on httpClientBuilders, which open more proxy configuration.

JonathanGiles:
Yes, please remove it as it is unused as of now. We can always re-introduce it in the future. Thank you for tidying up.

JonathanGiles:
You might want to consider which other policies are currently unused and whether they should be removed for now. You might also want to consider moving these into a sample, so that we can point users to it if they want to know how to do it.

joshfree:
Unless this is already in PR, please barcheck this with adpship before starting any work.  This feels like something we can live with, given the strong desire to lock down churn in Azure.Core and Azure.Identity this week.

/cc @Petermarcu

anuchandy:
Addressed in the PR https://github.com/Azure/azure-sdk-for-java/issues/5779

joshfree:
Closing issue as the PR has been merged.

PR #5836, opened 2019-10-14
Getting rid of ProxyPolicy, Enabling proxy credentials to be specified in HttpClient level

anuchandy:


anuchandy:
Addresses https://github.com/Azure/azure-sdk-for-java/issues/5779

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
```diff
@@ -4,13 +4,17 @@
...
 import java.net.InetSocketAddress;
+import java.util.Objects;
 
 /**
  * proxy configuration.
  */
 public class ProxyOptions {
     private final InetSocketAddress address;
     private final Type type;
+    private String userName;
```
JonathanGiles: This should be `username` and `getUsername()`

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
```diff
@@ -23,6 +27,19 @@ public ProxyOptions(Type type, InetSocketAddress address) {
...
     }
 
+    /**
+     * Set the proxy credentials.
+     *
+     * @param username proxy user name
+     * @param password proxy password
+     * @return the updated ProxyOptions object
+     */
+    public ProxyOptions credentials(String username, String password) {
```
JonathanGiles: This should be `setCredentials`

PR #5837, opened 2019-10-14
Return errors in Publisher's error channel for ConfigurationAsyncClient

srnagar:
The intent of this change is to use Publisher's error channel to propagate exceptions instead of throwing Exceptions from methods that return Mono/Flux type in `ConfigurationAsyncClient` .

Other async clients will be addressed in follow-up PRs.

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
@@ -21,7 +21,6 @@ import com.azure.core.implementation.util.ImplUtils;
 import com.azure.core.util.Context;
 import com.azure.core.util.logging.ClientLogger;
 import org.reactivestreams.Publisher;
-import reactor.core.Exceptions;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
--- a/sdk/core/azure-core/src/main/java/com/azure/core/implementation/util/FluxUtil.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/implementation/util/FluxUtil.java
@@ -6,7 +6,6 @@ package com.azure.core.implementation.util;
 import com.azure.core.http.rest.Response;
 import com.azure.core.util.Context;
 import com.azure.core.util.logging.ClientLogger;
-import java.util.function.Supplier;
 import org.reactivestreams.Subscriber;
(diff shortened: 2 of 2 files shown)


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
```diff
@@ -112,8 +113,12 @@
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Mono<ConfigurationSetting> addSetting(ConfigurationSetting setting) {
-        return withContext(context -> addSetting(setting, context))
-            .flatMap(response -> Mono.justOrEmpty(response.getValue()));
+        try {
```
conniey: Where is the the RuntimeException being thrown from? I'm surprised that the `withContext(…)` or underlying RestProxy isn't bubbling this to downstream.

srnagar: The exception is thrown in one of the methods in this class `validateSetting(setting)`. Besides, the purpose of this change is to catch all runtime exceptions before the user code - sdk api handshake and return them in the error channel.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
```diff
@@ -526,8 +591,12 @@
      */
     @ServiceMethod(returns = ReturnType.COLLECTION)
     public PagedFlux<ConfigurationSetting> listSettings(SettingSelector selector) {
-        return new PagedFlux<>(() -> withContext(context -> listFirstPageSettings(selector, context)),
-            continuationToken -> withContext(context -> listNextPageSettings(context, continuationToken)));
+        try {
+            return new PagedFlux<>(() -> withContext(context -> listFirstPageSettings(selector, context)),
+                continuationToken -> withContext(context -> listNextPageSettings(context, continuationToken)));
+        } catch (RuntimeException ex) {
+            return new PagedFlux<>(() -> Mono.error(logger.logExceptionAsError(Exceptions.propagate(ex))));
```
JonathanGiles: What about a `monoError` here too?

srnagar: Updated these to `monoError` as well.

PR #5848, opened 2019-10-15
Mapping function to convert from PagedFlux<T> to PagedFlux<S>

### LINKED ISSUE #5825: Blocks Preview 5 ADLS: Missing API - PagedFlux map function to return a PagedFlux
Currently, PagedFlux extends Flux so the map operation returns a Flux. 

Storage needs a way to go from a PagedFlux of A to a PagedFlux of C to implement DataLake features.

Some datalake APIs (that are a separate namespace from blobs) call directly into blob APIs (ex: listBlobContainers) that returns a PagedFlux of BlobContainerItems.

To make the namespace for a user consistent, we need to be able to return a PagedFlux of SomeOtherItem. 

For example,
```
 public PagedFlux<FileSystemItem> listFileSystems(ListFileSystemsOptions options) {
        return blobServiceAsyncClient.listBlobContainers(Transforms.toListBlobContainersOptions(options))
            .map(blobContainerItem -> Transforms.toFileSystemItem(blobContainerItem));
    }
```
This breaks because the map call is returning a Flux, but we need to return to the user a PagedFlux.

srnagar:
This change is to introduce a mapping function in `PagedFlux` to convert from value of type T to value of type S in `PagedResponse`

Fixes #5825

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
@@ -86,7 +86,7 @@ public class PagedFlux<T> extends PagedFluxBase<T, PagedResponse<T>> {
             pagedResponse.getStatusCode(),
             pagedResponse.getHeaders(),
             pagedResponse.getValue().stream().map(mapper).collect(Collectors.toList()),
-            pagedResponse.getNextLink(),
+            pagedResponse.getContinuationToken(),
             null);
     }
 }


sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
```diff
@@ -63,4 +66,28 @@ public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever,
...
     }
+
+    /**
+     * Maps this PagedFlux instance of T to a PagedFlux instance of type S as per the provided mapper function.
+     *
+     * @param mapper The mapper function to convert from type T to type S.
+     * @param <S> The mapped type.
+     * @return A PagedFlux of type S.
+     */
+    public <S> PagedFlux<S> mapTo(Function<T, S> mapper) {
```
srnagar: Can't override `map` method in Flux as the return type is not `Flux`. Open to suggestions for a better method name.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
```diff
@@ -63,4 +66,28 @@ public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever,
...
+                pagedResponse.getValue().stream().map(mapper).collect(Collectors.toList()),
+                pagedResponse.getNextLink(),
+                null)),
+            continuationToken -> getNextPageRetriever().apply(continuationToken)
+                .map(pagedResponse -> new PagedResponseBase<HttpRequest, S>(pagedResponse.getRequest(),
+                    pagedResponse.getStatusCode(),
+                    pagedResponse.getHeaders(),
+                    pagedResponse.getValue().stream().map(mapper).collect(Collectors.toList()),
+                    pagedResponse.getNextLink(),
+                    null)));
```
JonathanGiles: You should consider extracting the function out of the map call so that you can reuse the same code without duplication.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
JonathanGiles: I'm not sure what a better name is. I wonder if somehow 'page' should be in it, because we will possibly need to add more page-specific functionality in the future. We already have `byPage`, maybe this should be `mapPage`. In the future we might have `zipPage` or `transformPage` if necessary.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
srnagar: Done.

srnagar: Done

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
```diff
@@ -63,4 +66,27 @@ public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever,
...
+                .map(mapPagedResponse(mapper)));
+    }
+
+    private <S> Function<PagedResponse<T>, PagedResponse<S>> mapPagedResponse(Function<T, S> mapper) {
+        return pagedResponse -> new PagedResponseBase<HttpRequest, S>(pagedResponse.getRequest(),
+            pagedResponse.getStatusCode(),
+            pagedResponse.getHeaders(),
+            pagedResponse.getValue().stream().map(mapper).collect(Collectors.toList()),
+            pagedResponse.getContinuationToken(),
+            null);
```
AlexGhiondea: What does this `null` correspond to?

alzimmermsft: The parameter here is for a second generic type on the PagedResponse class for deserialized headers which is a property mapped class of headers.

@JonathanGiles I've pondering why we have this as part of the PagedResponse interface/abstract class, I've only seen them in autogen SDKs.

JonathanGiles: So `PagedResponseBase`, which Srikanta is creating here, is not public API. It implements the `PagedResponse` API, which is what users should only ever see. This second parameter type is only on PagedResponseBase, and is not visible to end users. There is probably scope to clean this up, but it isn't user-facing, so it is low priority in my eyes.

joshfree:
Please be sure to squash and merge, once approved by shiproom

PR #5856, opened 2019-10-15
Prepend Common Blob Classes with Blob

### LINKED ISSUE #5791: [Swagger] Add Swagger Transforms to Change Common Class Names
For the classes that are generated with the same names across the services they should be prepended with the name of the service, for example StorageErrorCode -> BlobStorageErrorCode.

alzimmermsft:
Partially fixes #5791

Renames all the common classes, those generated in each of the Blobs, Files, and Queues SDKs, in Blobs to begin with Blob.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
@@ -113,14 +113,18 @@ class ServiceAPITest extends APISpec {
         setup:
         def NUM_CONTAINERS = 5
         def PAGE_RESULTS = 3
+        def containerName = generateContainerName()
+        def containerPrefix = containerName.substring(0, Math.min(60, containerName.length()))
 
         def containers = [] as Collection<BlobContainerClient>
         for (i in (1..NUM_CONTAINERS)) {
-            containers << primaryBlobServiceClient.createBlobContainer(generateContainerName())
+            containers << primaryBlobServiceClient.createBlobContainer(containerPrefix + i)
         }
 
         expect:
-        primaryBlobServiceClient.listBlobContainers(new ListBlobContainersOptions().setMaxResultsPerPage(PAGE_RESULTS), null)
+        primaryBlobServiceClient.listBlobContainers(new ListBlobContainersOptions()
+            .setPrefix(containerPrefix)
+            .setMaxResultsPerPage(PAGE_RESULTS), null)
             .iterableByPage().iterator().next().getValue().size() == PAGE_RESULTS
 
         cleanup:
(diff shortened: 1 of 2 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerAccessPolicies.java
```diff
@@ -25,9 +25,9 @@ public PublicAccessType getBlobAccessType() {
     }
 
     /**
-     * @return the {@link SignedIdentifier SignedIdentifiers} associates with the container.
+     * @return the {@link BlobSignedIdentifier SignedIdentifiers} associates with the container.
```
conniey: Update link name -> BlobSignedIdentifier

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
```diff
@@ -108,7 +108,7 @@ private AppendBlobOutputStream(final AppendBlobAsyncClient client,
 
             return client.appendBlockWithResponse(blockData, writeLength, appendBlobAccessConditions)
                 .then()
-                .onErrorResume(t -> t instanceof IOException || t instanceof StorageException, e -> {
+                .onErrorResume(t -> t instanceof IOException || t instanceof BlobStorageException, e -> {
                     this.lastError = new IOException(e);
                     return null;
```
conniey: Does this work? I noticed all these `null` returns. iirc, it'll throw an exception if a null is returned.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerAccessPolicies.java
alzimmermsft: Changed

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
alzimmermsft: I'll change these to return Mono.empty

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobMetrics.java
```diff
@@ -97,9 +97,9 @@ public Boolean isIncludeAPIs() {
      * summary statistics for called API operations.
      *
      * @param includeAPIs the includeAPIs value to set.
-     * @return the Metrics object itself.
+     * @return the BlobMetrics object itself.
      */
-    public Metrics setIncludeAPIs(Boolean includeAPIs) {
+    public BlobMetrics setIncludeAPIs(Boolean includeAPIs) {
```
conniey: Should this be setIncludeApis? And same with the getter?

alzimmermsft: Will document this issue, this will require another Swagger transform.

alzimmermsft:
All pipeline steps passed, GitHub isn't updating properly.

PR #5858, opened 2019-10-15
Use mono error in KV Keys async clients

srnagar:
The intent of this change is to use Publisher's error channel to propagate exceptions instead of throwing Exceptions from methods that return Mono/Flux type in:
- KeyAsyncClient
- CryptographyAsyncClient
- KeyEncryptionKeyAsyncClient
- SecreteAsyncClient

Other async clients will be addressed in follow-up PRs.

### DIFF
--- a/sdk/keyvault/azure-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/SecretAsyncClient.java
+++ b/sdk/keyvault/azure-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/SecretAsyncClient.java
@@ -29,6 +29,7 @@ import com.azure.core.exception.ResourceNotFoundException;
 import com.azure.core.exception.ResourceModifiedException;
 import com.azure.core.exception.HttpRequestException;
 
+import static com.azure.core.implementation.util.FluxUtil.monoError;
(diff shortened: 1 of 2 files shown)


sdk/keyvault/azure-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/KeyAsyncClient.java
```diff
@@ -563,9 +620,13 @@
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Mono<Response<Key>> getKeyWithResponse(KeyProperties keyProperties) {
-        Objects.requireNonNull(keyProperties, "The Key Properties parameter cannot be null.");
-        return withContext(context -> getKeyWithResponse(keyProperties.getName(), keyProperties.getVersion() == null ? ""
-            : keyProperties.getVersion(), context));
+        try {
+            Objects.requireNonNull(keyProperties, "The Key Properties parameter cannot be null.");
```
g2vinay: Just to confirm all exceptions including null pointer will flow through Mono.error ?

srnagar: Yes

PR #5860, opened 2019-10-15
Update artifact versions using the script

srnagar:
Update the artifact versions to prep for the next release.

### DIFF
--- a/sdk/core/azure-core-http-okhttp/pom.xml
+++ b/sdk/core/azure-core-http-okhttp/pom.xml
@@ -64,7 +64,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-test</artifactId>
-      <version>1.0.0-preview.6</version>
+      <version>1.0.0-preview.7</version>  <!-- {x-version-update;com.azure:azure-core-test;current} -->
     </dependency>
 
     <dependency>


sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueConfiguration.java
```diff
@@ -9,5 +9,5 @@
 class QueueConfiguration {
     //TODO: Eventually remove these hardcoded strings with https://github.com/Azure/azure-sdk-for-java/issues/3141
     static final String NAME = "azure-storage-queue";
-    static final String VERSION = "12.0.0-preview.4";
+    static final String VERSION = "12.0.0-preview.5";
```
weshaggard: Did you have to update this one manually? @JimSuplizio looks like we might need to add some support for the source code itself as well.

JimSuplizio: @weshaggard Yeah, you're right. It would be trivial to update the script to deal with .java files, my concern is how long it'll take to process the entire tree. There are far less .md and pom files than there are java files. I'll do some experimentation.

srnagar: Yeah, this had to be done manually and @JimSuplizio is looking into automating it. We also have a github issue to remove these hardcoded version numbers from source code. If we do that, then updating the script won't be necessary.

PR #5862, opened 2019-10-15
Fixed PR merge issues, and removed unused imports

kushagraThapar:
The PRs
https://github.com/Azure/azure-sdk-for-java/pull/5804
https://github.com/Azure/azure-sdk-for-java/pull/5185
 
PR 5804 added the following line (225) to GatewayAddressCache.java
logger.error("unexpected failure", ex);
where ex is defined 3 lines before that in .onErrorResume(ex ->
and then PR 5185 renamed ex to throwable but merge of the files left the logger.error line still referencing ex which no longer exists.


kushagraThapar:
> Why there is no CI running? that should capture such issues right?

Yes, it did, and that's how I came to know :)

Issue #5062, opened 2019-08-20
Management libraries should be included in DependencyManagement


bsiegel:
The client libraries currently manage their dependency versions via dependencyManagement in the parent POM. The management libraries should also share the same dependencyManagement section so that dependency versions are consistent across management & client libraries. The dependency checker should then analyze both sets of libraries.

yaohaizh:
@bsiegel, we already have the same dependencyManagement to the ./parent/pom.xml for management plane, so is there any additional work item that management plane need handle? Thanks.

bsiegel:
Hey @yaohaizh, the issue we have with dependencyManagement is that it doesn't enforce anything - individual libraries are free to override what's there with their own dependency versions. Next steps would be to run the dependency analyzer on the management plane libraries and get a report of anywhere they deviate from the dependencyManagement section.

yaohaizh:
Close the issue as it is fixed on https://github.com/Azure/azure-sdk-for-java/pull/5879

PR #5879, opened 2019-10-16
Update the management plane dependency version.

yaohaizh:


yaohaizh:
> There's also work going on to remove the properties and dependencyManagement section from the parent/pom.xml which would allow us to not have to publish it. 
@JimSuplizio remove the dependency to parent/pom.xml would be good solution. Do you have any details about the work regarding above?

yaohaizh:
Also, I found this issue logged for https://github.com/Azure/azure-sdk-for-java/issues/5062 logged for DependencyManagement. Is there any previous discussion about how to deal with DependencyManagement? Thanks.

JimSuplizio:
> Also, I found this issue logged for #5062 logged for DependencyManagement. Is there any previous discussion about how to deal with DependencyManagement? Thanks.
@yaohaizh  Basically, what I'd recommend for you and for the management libraries would be to copy the [properties section](https://github.com/Azure/azure-sdk-for-java/blob/46174253a3ff90e2752d86f9e22be6afd38fdd68/parent/pom.xml#L94) and [dependencyManagement section](https://github.com/Azure/azure-sdk-for-java/blob/46174253a3ff90e2752d86f9e22be6afd38fdd68/parent/pom.xml#L195) out of the parent and into pom.management.xml and remove the <parent> entry from pom.management.xml and see what happens when you compile. You may need to lift a few other pieces out. Basically, if you can get management compiling without the parent pom you're going to be far better off especially when you start moving things in the management over to the service pipelines.

yaohaizh:
@JimSuplizio sounds good to me for the pom.management.xml, but I'm not sure about the file pom.xml since it contains sub modules of data/management/client. 

@jianghaolu do you have any suggestion here for pom.management.xml?

pom.management.xml
```diff
@@ -21,7 +21,7 @@
   <parent>
     <groupId>com.azure</groupId>
     <artifactId>azure-sdk-parent</artifactId>
-    <version>1.1.0</version> <!-- {x-version-exempt;com.azure:azure-sdk-parent;all of management needs to be updated in order for this to be at the latest} -->
+    <version>1.4.0</version> <!-- {x-version-exempt;com.azure:azure-sdk-parent;all of management needs to be updated in order for this to be at the latest} -->
```
jianghaolu: It should be 1.5.0 now.

yaohaizh:
Updated the versions to 1.5 for parent and bump pox.management.xml to 1.2 and cascade the updating to all the management SDKs.  @JimSuplizio as talked with @jianghaolu , we would like to have  all the dependencies in parent/pom.xml is so that we can avoid version conflicts if customers are using any combination of our SDKs. From the customer perspective, I think we would like to provide unified set of SDKs rather diverged versions, and it would be important later when management plane SDK also adopt Azure core.

JimSuplizio:
> Updated the versions to 1.5 for parent and bump pox.management.xml to 1.2 and cascade the updating to all the management SDKs. @JimSuplizio as talked with @jianghaolu , we would like to have all the dependencies in parent/pom.xml is so that we can avoid version conflicts if customers are using any combination of our SDKs. From the customer perspective, I think we would like to provide unified set of SDKs rather diverged versions, and it would be important later when management plane SDK also adopt Azure core.

adding @weshaggard 
@yaohaizh I'm actively trying to remove the version strings in the properties as well as the dependencyManagement section of the parent POM file so we can stop publishing it. This is part of the external dependencies work that I'm doing. I can't touch management because it's not in the pipelines yet. My recommendation still stands, it would be better if you could break the dependency on parent\pom.xml for management before the work is done to move management into the pipelines you'd be far better off. If that's not done here, when I ultimately remove those sections form parent I'm just going to copy them as-is into the management pom. I recommended you guys do this work because, right now, I am unable to build management on my windows box and that's kind of a problem (won't be once your team has moved them into the pipelines but there's no ETA on that actually getting done).

yaohaizh:
@JimSuplizio Any update on this? The change was made as we agreed.

Issue #5890, opened 2019-10-16
White list default http headers for logging


JonathanGiles:
Refer to the headers added [here](https://github.com/Azure/azure-sdk-for-net/pull/8166), we need to white list the same set for Java logging.

Fixed by #5893

joshfree:
We should get this snuck in ahead of GA

https://github.com/Azure/azure-sdk-for-net/pull/8166/files

joshfree:
(tagging as blocking-release for Preview 5 solely for bug tracking on the query; not that I would actually block Preview 5 for this change if it doesn't make it in time)

samvaity:
Closed in #5893

PR #5893, opened 2019-10-17
Add default set of headers to be white listed for logging

samvaity:
Reference: https://github.com/Azure/azure-sdk-for-java/issues/5890

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
@@ -5,6 +5,7 @@ package com.azure.core.http.policy;
 
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
```diff
@@ -54,18 +78,23 @@ public HttpLogOptions setLogLevel(final HttpLogDetailLevel logLevel) {
...
+     * <p>
+     * If a set of allowedHeaderNames is not provided, the default header names will be used to be whitelisted.
      *
      * @param allowedHeaderNames The list of whitelisted header names from the user.
      * @return The updated HttpLogOptions object.
-     * @throws NullPointerException If {@code allowedHeaderNames} is {@code null}.
      */
     public HttpLogOptions setAllowedHeaderNames(final Set<String> allowedHeaderNames) {
-        this.allowedHeaderNames = allowedHeaderNames;
+        this.allowedHeaderNames = allowedHeaderNames == null ?  this.allowedHeaderNames : allowedHeaderNames;
```
JonathanGiles: If they pass in null, just have that clear out the whitelist.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
```diff
@@ -17,7 +18,30 @@
 
     public HttpLogOptions() {
         logLevel = HttpLogDetailLevel.NONE;
-        allowedHeaderNames = new HashSet<>();
+        allowedHeaderNames = new HashSet<>(Arrays.asList(
```
srnagar: Extract the `Arrays.asList()` into a static field and reuse. Don't have to create a list and a hashset each time an instance of HttpLogOptions is created.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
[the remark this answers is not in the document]
JonathanGiles: This can be done via `getAllowedHeaderNames().remove(..)`, so we don't need a convenience API.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
JonathanGiles: I partially agree. The Arrays.asList(...) call can be a private static final field, but there should be a new set per instance, so that it can be modified via the API.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -122,7 +122,7 @@ public HttpLoggingPolicy(HttpLogOptions httpLogOptions) {
 
     private void formatAllowableHeaders(Set<String> allowedHeaderNames, HttpHeaders requestResponseHeaders,
                                         ClientLogger logger) {
-        if (allowedHeaderNames != null && !allowedHeaderNames.isEmpty()) {
+        if (!allowedHeaderNames.isEmpty()) {
             StringBuilder sb = new StringBuilder();
             for (HttpHeader header : requestResponseHeaders) {
                 sb.append(header.getName()).append(":");
```
JonathanGiles: This should be case insensitive

samvaity: updated here - https://github.com/Azure/azure-sdk-for-java/pull/5893/files#diff-1bf1978289f15f56cf8db11581def1a9R129

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
samvaity: updated

PR #5898, opened 2019-10-17
Removed URL from public API

### LINKED ISSUE #5656: Investigate if storage methods taking URL input should change to taking String url input
Currently, both the .NET and Java team are using URL type input for methods like `copyFromUrlWithResponse` and such. 
Investigate if they should be taking  String input instead.

alzimmermsft:
Fixes #5656 

Removed URL and URI as parameters from public client API, parameters now use Strings.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/AppendBlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/AppendBlobAsyncClient.java
@@ -227,7 +227,7 @@ public final class AppendBlobAsyncClient extends BlobAsyncClientBase {
         try {
             url = new URL(sourceUrl);
         } catch (MalformedURLException ex) {
-            throw logger.logExceptionAsError(new IllegalArgumentException(ex));
+            throw logger.logExceptionAsError(new IllegalArgumentException("'sourceUrl' is not a valid url."));
         }
 
         return this.azureBlobStorage.appendBlobs().appendBlockFromUrlWithRestResponseAsync(null, null, url, 0,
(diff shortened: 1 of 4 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -395,8 +404,15 @@ public boolean isSnapshot() {
             .setSourceIfMatch(sourceModifiedAccessConditions.getIfMatch())
             .setSourceIfNoneMatch(sourceModifiedAccessConditions.getIfNoneMatch());
 
+        URL url;
+        try {
+            url = new URL(copySource);
+        } catch (MalformedURLException ex) {
+            throw logger.logExceptionAsError(new IllegalArgumentException(ex));
```
conniey: Should we add a message here with that malformed url as an inner exception.
("'copysource' is not a valid url.", ex)

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/AppendBlobAsyncClient.java
```diff
@@ -205,21 +209,28 @@
...
         sourceRange = sourceRange == null ? new BlobRange(0) : sourceRange;
         destAccessConditions = destAccessConditions == null
             ? new AppendBlobAccessConditions() : destAccessConditions;
 
-        return this.azureBlobStorage.appendBlobs().appendBlockFromUrlWithRestResponseAsync(null, null, sourceURL, 0,
+        URL url;
+        try {
+            url = new URL(sourceUrl);
+        } catch (MalformedURLException ex) {
+            throw logger.logExceptionAsError(new IllegalArgumentException(ex));
```
conniey: I'd prefer to add a message to this, too. "'sourceUrl' is not a valid url.", Ex)

alzimmermsft: Done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
alzimmermsft: Done

PR #5902, opened 2019-10-17
Properly Pass Service Version through Client Hierarchy

alzimmermsft:
Fixes an issue where the service version wasn't being passed when create clients through other clients. This is currently blocking other development.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/SpecializedBlobClientBuilder.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/SpecializedBlobClientBuilder.java
@@ -508,7 +508,7 @@ public final class SpecializedBlobClientBuilder {
         return this;
     }
 
-    private BlobServiceVersion fromClientVersion(String version) {
+    private static BlobServiceVersion fromClientVersion(String version) {
         return Arrays.stream(BlobServiceVersion.values())
             .filter(en -> Objects.equals(en.getVersion(), version))
             .findFirst()


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -158,6 +159,14 @@ public String getAccountName() {
         return this.accountName;
     }
 
+    /**
+     * Gets the service version the client is using.
+     *
+     * @return the service version the client is using.
+     */
+    public String getServiceVersion() {
```
conniey: Do we need to expose this as a public rather than package-private variable? 🤔 The reason is that we'll have to pass this public API change through the board.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/SpecializedBlobClientBuilder.java
```diff
@@ -502,4 +507,11 @@ public SpecializedBlobClientBuilder serviceVersion(BlobServiceVersion version) {
         this.version = version;
         return this;
     }
+
+    private BlobServiceVersion fromClientVersion(String version) {
```
conniey: this can be private static because it doesn't depend on any member variables from what I can see.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
alzimmermsft: Good question, we've generally erred on having the build of the client being more public in storage to allow jumping through the hierarchy more easily. The tough part is at some point or another this needs to surface through public API to be accessible in the specialized/root package divide.

@JonathanGiles mind chiming in on this?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/SpecializedBlobClientBuilder.java
alzimmermsft: Done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
JonathanGiles: I believe this has changed more recently to return BlobServiceVersion, however I would rather it not be public API. There are ways to enable this (with *Helper classes), so I would encourage some degree of investigation so that, ideally, our storage APIs are only public when the target user is the end-developer, and not our code (which indicates a failing on our part).

PR #5908, opened 2019-10-17
Sas generation

### LINKED ISSUE #5745: [ArchBoard] Standardize SAS Token Generation across .NET, Java, Python, and JS Azure Storage SDKs
Currently, SAS token generation has different stories for each language and it needs to be standardized.

### LINKED ISSUE #5852: [QUERY] [V12][Storage] replacement for generateSAS on blobClient
**Query/Question**
In the preview4 version of the V12 storage api, the blobClient.generateSAS(...) method was removed. The documentation is not clear how this should be replaced.
BlobServiceSasSignatureValues seems to have a generateSAS method - but requires credentials to be passed in. The beauty of the old API, was that these were hidden from the user when just using the blobClient, which allowed a clear separation of credential setup and actual use of the API. (e.g. injecting the fully configured client into a service which has no credential access).

Or am i missing something ?

***Why is this not a Bug or a feature Request?***
This was changed between preview.2 and preview.4

**Setup (please complete the following information if applicable):**
Example code in preview.2:
final String signature = blob.generateSAS(expiryTime, BlobSasPermission.parse(permissions)); 
 
**Information Checklist**
- [x] Query Added
- [x] Setup information Added

conniey:
This PR aligns the SAS query parameter generation to be more like the other languages. Aligns the behaviour/algorithm to the one used by [Go](https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/sas_service.go#L33) and [.NET](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Blobs/src/Sas/BlobSasBuilder.cs).

* Removes much of the client-side validation.
* Replaces get/set canonicalName with getters and setters for containerName and blobName.
  * We were the only language that did this.
* Hides "resource". This is inferred based on whether "blobName" and "snapshotId" are set.
* Adds codesnippets and updates documentation.

## TODO

- [x] Fix test breaks. (Once the ones in master start working again.)

Related to #5745 
Fixes #5852

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobServiceSasSignatureValues.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobServiceSasSignatureValues.java
@@ -17,10 +17,11 @@ import com.azure.storage.common.sas.SasProtocol;
 import java.time.OffsetDateTime;
 
 /**
- * BlobServiceSasSignatureValues is used to generate a Shared Access Signature (SAS) for an Azure Storage service. Once
- * all the values here are set appropriately, call {@link #generateSasQueryParameters(StorageSharedKeyCredential)} or
- * {@link #generateSasQueryParameters(UserDelegationKey, String)} to obtain a representation of the SAS which can be
- * applied to blob urls.
+ * Used to generate a Shared Access Signature (SAS) for an Azure Blob Storage service. Once all the values here are set,
+ * call {@link
+ * #generateSasQueryParameters(StorageSharedKeyCredential) generateSasQueryParameters(StorageSharedKeyCredential)} or
+ * {@link #generateSasQueryParameters(UserDelegationKey, String) generateSasQueryParameters(UserDelegationKey, String)}
+ * to obtain a representation of the SAS which can be applied to blob urls.
  *
  * <p><strong>Generating SAS query parameters with {@link StorageSharedKeyCredential}</strong></p>
  * The following code generates SAS query parameters for an Azure storage blob.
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobServiceSasSignatureValues.java
```diff
@@ -474,60 +478,111 @@ public BlobServiceSasQueryParameters generateSasQueryParameters(SharedKeyCredent
...
      * encoded string, or the UTF-8 charset isn't supported.
-     * @throws NullPointerException if {@code delegationKey} is null. Or if any of {@code version},
-     * {@code canonicalName}, {@code resource}, {@code expiryTime} or {@code permissions} are null.
+     * @throws IllegalArgumentException if {@link #getPermissions()} contains an invalid character for the SAS resource.
+     * @throws NullPointerException if {@code delegationKey} or {@code account} is null.
+     * @see <a href="https://docs.microsoft.com/rest/api/storageservices/create-user-delegation-sas">Create a user
+     * delegation SAS</a>
      */
-    public BlobServiceSasQueryParameters generateSasQueryParameters(UserDelegationKey delegationKey) {
+    public BlobServiceSasQueryParameters generateSasQueryParameters(UserDelegationKey delegationKey, String accountName) {
```
srnagar: This is a breaking change. Should there be an overload that takes just `delegationKey` too as that was supported earlier?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobServiceSasSignatureValues.java
```diff
@@ -474,60 +478,111 @@ public BlobServiceSasQueryParameters generateSasQueryParameters(SharedKeyCredent
...
+     * https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/sas_service.go#L33
+     * https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Blobs/src/Sas/BlobSasBuilder.cs
      */
-    private void assertGenerateOK(boolean usingUserDelegation) {
-        Utility.assertNotNull("version", this.version);
-        Utility.assertNotNull("canonicalName", this.canonicalName);
-        Utility.assertNotNull("resource", this.resource);
+    private void ensureState() {
+        if (version == null) {
+            version = Constants.HeaderConstants.TARGET_STORAGE_VERSION;
```
JonathanGiles: Should this be changed based on the service version work that @samvaity has been doing (which is now in the repo)?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobServiceSasSignatureValues.java
conniey: It only worked because the had the `setCanonicalName("")` which allowed users to specify account, container and blob name. This overload is used in all the other languages.

conniey: @alzimmermsft @rickle-msft thoughts?

alzimmermsft: I'm okay with the change if it means we are going to be consistent with the other languages. SAS token generation is hard and being consistent in API will reduce overhead.

PR #5911, opened 2019-10-18
Default block size

rickle-msft:
Adds an optimization that adjusts the default block size when blobs are above a certain size. Changing the default value in this case, will trigger server side optimizations.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
+++ b/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
@@ -206,8 +206,8 @@
   <suppress checks="LineLength" files=".*[/\\]storage[/\\]file[/\\]datalake[/\\]models[/\\].*"/>
 
   <!-- Suppress CRUD operation names as they won't feel fluent -->
-  <suppress checks="com.azure.tools.checkstyle.checks.ServiceClientCheck" files="com.azure.storage.blob.specialized.LeaseClient.java"/>
-  <suppress checks="com.azure.tools.checkstyle.checks.ServiceClientCheck" files="com.azure.storage.blob.specialized.LeaseAsyncClient.java"/>
+  <suppress checks="com.azure.tools.checkstyle.checks.ServiceClientCheck" files="com.azure.storage.blob.specialized.BlobLeaseClient.java"/>
+  <suppress checks="com.azure.tools.checkstyle.checks.ServiceClientCheck" files="com.azure.storage.blob.specialized.BlobLeaseAsyncClient.java"/>
   <suppress checks="com.azure.tools.checkstyle.checks.ServiceClientCheck" files="com.azure.storage.blob.batch.BlobBatchClient.java"/>
   <suppress checks="com.azure.tools.checkstyle.checks.ServiceClientCheck" files="com.azure.storage.blob.batch.BlobBatchAsyncClient.java"/>
   <suppress checks="com.azure.tools.checkstyle.checks.ServiceClientCheck" files="com.azure.data.appconfiguration.ConfigurationClient.java"/>
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedIterable.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedIterable.java
@@ -5,6 +5,7 @@ package com.azure.core.http.rest;
 
 import com.azure.core.util.IterableStream;
 
+import java.util.function.Function;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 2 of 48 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -199,7 +208,8 @@ private SpecializedBlobClientBuilder prepareBuilder() {
      * Docs for Put Block</a> and the <a href="https://docs.microsoft.com/rest/api/storageservices/put-block-list">Azure
      * Docs for Put Block List</a>.
      * <p>
-     * The data passed need not support multiple subscriptions/be replayable as is required in other upload methods when
+     * The data passed need not support multiple subscriptions/be replayable as is required in other upload methods
+     * when
```
conniey: odd wrapping with single word on line

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -208,16 +218,20 @@ private SpecializedBlobClientBuilder prepareBuilder() {
...
      * data. Larger buffers means we will have to stage fewer blocks and therefore require fewer IO operations. The
-     * trade-offs between these values are context-dependent, so some experimentation may be required to optimize inputs
+     * trade-offs between these values are context-dependent, so some experimentation may be required to optimize
+     * inputs
      * for a given scenario.
      *
      * <p><strong>Code Samples</strong></p>
      *
-     * {@codesnippet com.azure.storage.blob.BlobAsyncClient.uploadWithResponse#Flux-ParallelTransferOptions-BlobHttpHeaders-Map-AccessTier-BlobAccessConditions}
+     * {@codesnippet com.azure.storage.blob.BlobAsyncClient
```
conniey: this will break codesnippets. same with the one below. they have an exception tob e longer than 120 characters

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -208,16 +218,20 @@ private SpecializedBlobClientBuilder prepareBuilder() {
      * <p>
      * Typically, the greater the number of buffers used, the greater the possible parallelism when transferring the
      * data. Larger buffers means we will have to stage fewer blocks and therefore require fewer IO operations. The
-     * trade-offs between these values are context-dependent, so some experimentation may be required to optimize inputs
+     * trade-offs between these values are context-dependent, so some experimentation may be required to optimize
+     * inputs
```
conniey: odd line wrapping

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -329,27 +344,28 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
      * @param headers {@link BlobHttpHeaders}
      * @param metadata Metadata to associate with the blob.
      * @param tier {@link AccessTier} for the destination blob.
      * @param accessConditions {@link BlobAccessConditions}
+     *
      * @return An empty response
+     *
      * @throws IllegalArgumentException If {@code blockSize} is less than 0 or greater than 100MB
-     * @throws UncheckedIOException If an I/O error occurs
+     * @throws UncheckedIOException     If an I/O error occurs
```
conniey: we don't align Javadoc parameters.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -329,27 +344,28 @@ as we can guarantee we only need at most two buffers for any call to write (two
      *
      * <p><strong>Code Samples</strong></p>
      *
-     * {@codesnippet com.azure.storage.blob.BlobAsyncClient.uploadFromFile#String-ParallelTransferOptions-BlobHttpHeaders-Map-AccessTier-BlobAccessConditions}
+     * {@codesnippet com.azure.storage.blob.BlobAsyncClient
```
conniey: Would revert this line break for codesnippet

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/ParallelTransferOptions.java
```diff
@@ -63,30 +58,53 @@ public ProgressReceiver getProgressReceiver() {
...
+     */
     public ParallelTransferOptions setProgressReceiver(ProgressReceiver progressReceiver) {
         this.progressReceiver = progressReceiver;
         return this;
     }
+
+    /**
+     * RESERVED FOR INTERNAL USE.
+     */
+    public void populateAndApplyDefaults(ParallelTransferOptions other) {
```
conniey: This is a public API change. Does it need to be public? It would have to be reviewed with other language folks

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
rickle-msft: IDE autoformatting. Thanks for catching.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/ParallelTransferOptions.java
rickle-msft: It has to be public because it's in a different package from where it's used

alzimmermsft: Is this constructor used anywhere else other than BlobAsyncClient? If not could we continue to localize this there?

rickle-msft: It's BlobAsyncClient and BlobAsyncClientBase, which are in two different packages

amishra-dev: @conniey @alzimmermsft i am assuming this is resolved. I need this PR to go in.
/cc @pemari-msft fyi

JonathanGiles: I've had an offline discussion on this today. I do not want us to get into the habit of creating 'reserved for internal use' public APIs that only exist to serve our needs in working around Java package limitations. I've suggested an alternative approach that should be considered, and @hemanttanwar is working through this for other APIs that exist solely to serve our needs in #5962. One option may be to push this as-is and then to follow-up in #5962 with it being made private again, but we should not ship with any API outside of the 'implementation' packages that is visible to users and reserved for internal use.

amishra-dev: doing it in #5962 sounds like a good option @JonathanGiles 
@rickle-msft please merge this PR

JonathanGiles: Two comments:
1) This still needs to go through the escrow process.
2) I'm not convinced that this API will be able to be easily hidden using the *Helper approach, and would like to discuss in some more detail any other options that may exist to negate the need to introduce this method.

Issue #5845, opened 2019-10-15
[Storage] Support all valid connection strings


mikeharder:
All valid connection strings should be supported.  There is some documentation at https://docs.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string, though it doesn't specify exactly what is and isn't valid.

Here is a specific connection string which is valid in both .NET and Java `com.microsoft.azure.azure-storage@8.4.0`, but invalid in Java `com.azure.azure-storage-blob@12.0.0-preview.4`:

```
# Blob endpoint only
DefaultEndpointsProtocol=https;
BlobEndpoint=http://www.mydomain.com;
AccountName=storagesample;
AccountKey=<account-key>
```

In  `com.azure.azure-storage-blob@12.0.0-preview.4`, the following exception is thrown:

```
Exception in thread "main" java.lang.NullPointerException: 'endpoint' cannot be null.
at java.base/java.util.Objects.requireNonNull(Objects.java:246)
at com.azure.storage.common.BaseClientBuilder.buildPipeline(BaseClientBuilder.java:72)
at com.azure.storage.blob.BlobServiceClientBuilder.buildAsyncClient(BlobServiceClientBuilder.java:57)
at com.azure.storage.blob.BlobServiceClientBuilder.buildClient(BlobServiceClientBuilder.java:48)
```

alzimmermsft:
We should target a fix for this before GA but it is mainly an implementation detail

PR #5924, opened 2019-10-18
Adding support for all connection string formats

anuchandy:
Adding support for all supported connection string format. All the types used for connection string handling is in common implementation package, nothing affecting public API.

The behavior of the connection parsing is now align with .NET connection string parsing w.r.t formats and validations.

Addresses: https://github.com/Azure/azure-sdk-for-java/issues/5845

>> TODO: Something we want to look into post GA is whether validation errors can be more specific and consistent across languages. Majority of the users are going to copy the settings from portal or cli and use it with SDK, so chances of invalid combinations will be rare but still good to be specific.

### DIFF
--- a/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
@@ -27,8 +27,10 @@ import com.azure.storage.blob.BlobUrlParts;
 import com.azure.storage.blob.implementation.AzureBlobStorageBuilder;
 import com.azure.storage.blob.implementation.AzureBlobStorageImpl;
 import com.azure.storage.common.implementation.Constants;
-import com.azure.storage.common.Utility;
 import com.azure.storage.common.StorageSharedKeyCredential;
+import com.azure.storage.common.implementation.connectionstring.StorageAuthenticationSettings;
+import com.azure.storage.common.implementation.connectionstring.StorageConnectionString;
+import com.azure.storage.common.implementation.connectionstring.StorageEndpoint;
 import com.azure.storage.common.implementation.credentials.SasTokenCredential;
 import com.azure.storage.common.implementation.policy.SasTokenCredentialPolicy;
 import com.azure.storage.common.policy.RequestRetryOptions;
@@ -41,7 +43,6 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 
 /**
@@ -300,39 +301,33 @@ public final class EncryptedBlobClientBuilder {
     }
 
     /**
-     * Constructs a {@link StorageSharedKeyCredential} used to authorize requests sent to the service. Additionally,
-     * if the connection string contains `DefaultEndpointsProtocol` and `EndpointSuffix` it will set the {@link
-     * #endpoint(String) endpoint}.
+     * Sets the connection string to connect to the service.
      *
      * @param connectionString Connection string of the storage account.
      * @return the updated EncryptedBlobClientBuilder
-     * @throws IllegalArgumentException If {@code connectionString} doesn't contain `AccountName` or `AccountKey`.
-     * @throws NullPointerException If {@code connectionString} is {@code null}.
+     * @throws IllegalArgumentException If {@code connectionString} is invalid.
      */
     public EncryptedBlobClientBuilder connectionString(String connectionString) {
-        Objects.requireNonNull(connectionString, "'connectionString' cannot be null.");
-
-        Map<String, String> connectionStringPieces = Utility.parseConnectionString(connectionString);
-
-        String accountName = connectionStringPieces.get(Constants.ConnectionStringConstants.ACCOUNT_NAME_NAME);
-        String accountKey = connectionStringPieces.get(Constants.ConnectionStringConstants.ACCOUNT_KEY_NAME);
-
-        if (ImplUtils.isNullOrEmpty(accountName) || ImplUtils.isNullOrEmpty(accountKey)) {
-            throw logger.logExceptionAsError(
-                new IllegalArgumentException("'connectionString' must contain 'AccountName' and 'AccountKey'."));
+        StorageConnectionString storageConnectionString
+                = StorageConnectionString.create(connectionString, logger);
+        StorageEndpoint endpoint = storageConnectionString.getBlobEndpoint();
+        if (endpoint == null || endpoint.getPrimaryUri() == null) {
+            throw logger
+                    .logExceptionAsError(new IllegalArgumentException(
+                            "connectionString missing required settings to derive blob service endpoint."));
         }
-
-        String endpointProtocol
-                = connectionStringPieces.get(Constants.ConnectionStringConstants.DEFAULT_ENDPOINTS_PROTOCOL_NAME);
-        String endpointSuffix = connectionStringPieces.get(Constants.ConnectionStringConstants.ENDPOINT_SUFFIX_NAME);
-
-        if (!ImplUtils.isNullOrEmpty(endpointProtocol) && !ImplUtils.isNullOrEmpty(endpointSuffix)) {
-            endpoint(String.format("%s://%s.blob.%s", endpointProtocol, accountName,
-                endpointSuffix.replaceFirst("^\\.", "")));
+        this.endpoint(endpoint.getPrimaryUri());
+        if (storageConnectionString.getAccountName() != null) {
+            this.accountName = storageConnectionString.getAccountName();
         }
-
-        this.accountName = accountName;
-        return credential(new StorageSharedKeyCredential(accountName, accountKey));
+        StorageAuthenticationSettings authSettings = storageConnectionString.getStorageAuthSettings();
+        if (authSettings.getType() == StorageAuthenticationSettings.Type.ACCOUNT_NAME_KEY) {
+            this.credential(new StorageSharedKeyCredential(authSettings.getAccount().getName(),
+                    authSettings.getAccount().getAccessKey()));
+        } else if (authSettings.getType() == StorageAuthenticationSettings.Type.SAS_TOKEN) {
+            this.sasToken(authSettings.getSasToken());
+        }
+        return this;
     }
 
     /**
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
@@ -54,7 +54,7 @@ public final class Constants {
         /**
          * The setting name for the storage account name.
          */
-        public static final String ACCOUNT_NAME_NAME = "AccountName";
+        public static final String ACCOUNT_NAME = "AccountName";
 
         /**
          * The setting name for the storage account key.
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/ConnectionSettingsFilter.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/ConnectionSettingsFilter.java
@@ -35,9 +35,7 @@ interface ConnectionSettingsFilter {
         return (ConnectionSettings inputSettings) -> {
             ConnectionSettings outputSettings = inputSettings.clone();
             for (final String settingName : settingNames) {
-                if (outputSettings.hasSetting(settingName)) {
-                    outputSettings.removeSetting(settingName);
-                }
+                outputSettings.removeSetting(settingName);
             }
             return outputSettings;
         };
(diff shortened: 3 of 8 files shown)


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -314,16 +314,17 @@ public EncryptedBlobClientBuilder connectionString(String connectionString) {
 
         Map<String, String> connectionStringPieces = Utility.parseConnectionString(connectionString);
 
-        String accountName = connectionStringPieces.get(Constants.ConnectionStringConstants.ACCOUNT_NAME);
-        String accountKey = connectionStringPieces.get(Constants.ConnectionStringConstants.ACCOUNT_KEY);
+        String accountName = connectionStringPieces.get(Constants.ConnectionStringConstants.ACCOUNT_NAME_NAME);
```
samvaity: `ACCOUNT_NAME` ?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -314,16 +314,17 @@ public EncryptedBlobClientBuilder connectionString(String connectionString) {
...
+        String accountKey = connectionStringPieces.get(Constants.ConnectionStringConstants.ACCOUNT_KEY_NAME);
 
         if (ImplUtils.isNullOrEmpty(accountName) || ImplUtils.isNullOrEmpty(accountKey)) {
             throw logger.logExceptionAsError(
                 new IllegalArgumentException("'connectionString' must contain 'AccountName' and 'AccountKey'."));
         }
 
-        String endpointProtocol = connectionStringPieces.get(Constants.ConnectionStringConstants.ENDPOINT_PROTOCOL);
-        String endpointSuffix = connectionStringPieces.get(Constants.ConnectionStringConstants.ENDPOINT_SUFFIX);
+        String endpointProtocol
```
sima-zhu: Why do we use different way of constructing endpoint here compare to blob, file..?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
```diff
@@ -52,24 +52,105 @@ private Constants() {
...
+
+        /**
+         * The setting name for a custom queue endpoint.
+         */
+        public static final String QUEUE_ENDPOINT_NAME = "QueueEndpoint";
+
+        /**
+         * The setting name for a custom queue secondary endpoint.
+         */
+        public static final String QUEUE_SECONDARY_ENDPOINT_NAME = "QueueSecondaryEndpoint";
```
sima-zhu: Can we generate secondary endpoint from portal? Or from CLI?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
```diff
@@ -52,24 +52,105 @@ private Constants() {
...
+
+        /**
+         * The format string for the secondary endpoint in emulator.
+         */
+        public static final String EMULATOR_SECONDARY_ENDPOINT_FORMAT = "%s://%s:%s/devstoreaccount1-secondary";
+
+        /**
+         * The default account key for the development storage.
+         */
+        public static final String EMULATOR_ACCOUNT_KEY
```
sima-zhu: Is this used by all emulators? Is it better to read from config?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
```diff
@@ -52,24 +52,105 @@ private Constants() {
...
+                = "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==";
+
+        /**
+         * The default account name for the development storage.
+         */
+        public static final String EMULATOR_ACCOUNT_NAME = "devstoreaccount1";
```
sima-zhu: Same there

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/StorageAuthenticationSettings.java
```diff
@@ -0,0 +1,152 @@
...
+     * @param settings the connection settings.
+     * @return the StorageAuthenticationSettings.
+     */
+    public static StorageAuthenticationSettings fromConnectionSettings(final ConnectionSettings settings) {
+        final String accountName = settings.getSettingValue(Constants.ConnectionStringConstants.ACCOUNT_NAME_NAME);
+        final String accountKey = settings.getSettingValue(Constants.ConnectionStringConstants.ACCOUNT_KEY_NAME);
+        final String sasSignature =
+                settings.getSettingValue(Constants.ConnectionStringConstants.SHARED_ACCESS_SIGNATURE_NAME);
+
+        if (accountName != null && accountKey != null && sasSignature == null) {
```
sima-zhu: If account name and key not null, then get into the block. No need to check sas, right?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
```diff
@@ -52,24 +52,105 @@ private Constants() {
      */
     public static final class ConnectionStringConstants {
         /**
-         * The AccountName key.
+         * The setting name for the storage account name.
          */
-        public static final String ACCOUNT_NAME = "accountname";
+        public static final String ACCOUNT_NAME_NAME = "AccountName";
```
alzimmermsft: ACCOUNT_NAME_NAME? Why not ACCOUNT_NAME as it was?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/ConnectionSettings.java
```diff
@@ -0,0 +1,102 @@
...
+        final String[] settings = connString.split(";");
+        for (int i = 0; i < settings.length; i++) {
+            String setting = settings[i].trim();
+            if (setting.length() > 0) {
+                final int idx = setting.indexOf("=");
+                if (idx == -1
+                        || idx == 0
+                        || idx == settings[i].length() - 1) {
+                    // handle no_equal_symbol, "=Bar", "Foo="
+                    throw logger.logExceptionAsError(new IllegalArgumentException("Invalid connection string."));
```
alzimmermsft: Is it actually invalid for a connection string piece to have no value?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/ConnectionSettings.java
```diff
@@ -0,0 +1,102 @@
...
+                }
+                map.put(setting.substring(0, idx), setting.substring(idx + 1));
```
alzimmermsft: Are connection strings case sensitive or insensitive? If insensitive we'd want to support that.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/ConnectionSettingsFilter.java
```diff
@@ -0,0 +1,131 @@
...
+            }
+            return outputSettings;
+        };
+    }
+
+    static ConnectionSettingsFilter optional(final String... settingNames) {
+        return (ConnectionSettings inputSettings) -> {
+            ConnectionSettings outputSettings = inputSettings.clone();
+            for (final String settingName : settingNames) {
+                if (outputSettings.hasSetting(settingName)) {
```
alzimmermsft: Could we just call remove here? Do we need to check for it existing in the map before removing?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/StorageServiceConnectionString.java
```diff
@@ -0,0 +1,203 @@
...
+                        ConnectionSettingsFilter.matchesOne(
+                                ConnectionSettingsFilter.matchesAll(requireAccountKey()),
+                                requireSas()),
+                        requireAccountName(),
+                        optionalEndpoints(),
+                        optionalEndpointProtocolAndSuffix())
+        );
+    }
+
+    private static ConnectionSettingsFilter explicitEndpointsMatchSpec() {
```
sima-zhu: What's the purpose the this match? Any example fit in this one but failed the automaticEndpointsMatchSpec?

alzimmermsft:
Given that you added a new package in implementation you need to export it to all the modules that are using it.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/ConnectionSettings.java
anuchandy: we can have either strict check or relaxed check on absence of value, here we're being bit strict, if needed it's easy to relax it later than the other way. 

The rule at least now is: If user want a setting then specify it as a valid pair of non-null key-value, if setting is not needed then don't include it in the connection string.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/ConnectionSettings.java
anuchandy: The documentation is not clear about case-sensitivity but the storage SDKs already out there uses case-sensitive lookup, we're following the same pattern.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/ConnectionSettingsFilter.java
anuchandy: good catch, this can be removed.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
anuchandy: yes, I'm about to update the PR to use `StorageConnectionString` type here as well that also removes parseConnectionString from Utility.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
anuchandy: I couldn't find a way to generate it from portal. It seems by appending `-secondary` to the blob name should result in secondary endpoint. I guess it's also possible to define a custom domain for secondary.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
anuchandy: Its same for all emulator - https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator#authorize-with-shared-key-credentials so it can be a const.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
anuchandy: account name for emulator is also a standard name https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator#authorize-with-shared-key-credentials hence can be const.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/StorageAuthenticationSettings.java
anuchandy: Actually as per validation rules - {account name:key} and {Sas} are mutually exclusive, we do that validation already on connection string, here it is just reinforcing for readability.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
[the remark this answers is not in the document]
anuchandy: agree, will name it to `ACCOUNT_NAME`

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
anuchandy: agree, will revert this change.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/StorageServiceConnectionString.java
anuchandy: Yes, if there is no `AccountName` but `BlobEndpoint` then automatic match will fail and explicit will succeeded.

PR #5929, opened 2019-10-18
Removed the json files for user delegate key tests

sima-zhu:


sima-zhu:
Closed it as it is not needed.

PR #5931, opened 2019-10-18
rerv5 api review comments

hemanttanwar:


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/sas/AccountSasService.java
```diff
@@ -69,7 +69,7 @@ public static AccountSasService parse(String servicesString) {
     /**
      * @return the access status for blob resources.
      */
-    public boolean isBlob() {
+    public boolean hasBlobAccess() {
```
alzimmermsft: Could you make these changes for the other setters and getters in this class?

Issue #5938, opened 2019-10-18
Add AccessConditions Models to Azure Core


alzimmermsft:
To match what the other languages are doing we need to add a few model classes in Azure Core to handle HTTP Access Conditions:

MatchConditions
- IfMatch: String
- IfNoneMatch: String

RequestConditions extends MatchConditions
- IfModifiedSince: OffsetDateTime
- IfUnmodifiedSince: OffsetDateTime

alzimmermsft:
Fixed in #5941

PR #5941, opened 2019-10-18
Add access condition models to Azure Core

alzimmermsft:
Implements #5938 

Adds `MatchConditions` and `RequestConditions` model classes to Azure Core to support HTTP access conditions.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/RequestConditions.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/RequestConditions.java
@@ -3,18 +3,22 @@
 
 package com.azure.core.http;
 
+import com.azure.core.annotation.Fluent;
+
 import java.time.OffsetDateTime;
 
 /**
  * Specifies HTTP options for conditional requests based on modification time.
  */
+@Fluent
 public class RequestConditions extends MatchConditions {
     private OffsetDateTime ifModifiedSince;
     private OffsetDateTime ifUnmodifiedSince;
 
     /**
-     * {@inheritDoc}
+     * Optionally limit requests to resources that match the passed ETag.
      *
+     * @param ifMatch ETag that resources must match.
      * @return The updated ResourceConditions object.
      */
     @Override
@@ -24,8 +28,9 @@ public class RequestConditions extends MatchConditions {
     }
 
     /**
-     * {@inheritDoc}
+     * Optionally limit requests to resources that do not match the passed ETag.
      *
+     * @param ifNoneMatch ETag that resources must not match.
      * @return The updated ResourceConditions object.
      */
     @Override


sdk/core/azure-core/src/main/java/com/azure/core/http/RequestConditions.java
```diff
@@ -0,0 +1,78 @@
...
+// Licensed under the MIT License.
+
+package com.azure.core.http;
+
+import java.time.OffsetDateTime;
+
+/**
+ * Specifies HTTP options for conditional requests based on modification time.
+ */
+public class RequestConditions extends MatchConditions {
```
srnagar: Add @Fluent annotation.

sdk/core/azure-core/src/main/java/com/azure/core/http/RequestConditions.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    public RequestConditions setIfMatch(String ifMatch) {
+        super.setIfMatch(ifMatch);
+        return this;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return The updated ResourceConditions object.
```
srnagar: {@inheritDoc} + new doc doesn't generate the Javadocs correctly. Use `@inheritDoc` only or add new javadoc.

sdk/core/azure-core/src/main/java/com/azure/core/http/RequestConditions.java
alzimmermsft: Done

sdk/core/azure-core/src/main/java/com/azure/core/http/RequestConditions.java
alzimmermsft: Pulled the super class documentation class directly.

sdk/core/azure-core/src/main/java/com/azure/core/http/MatchConditions.java
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
conniey: @mssfang Would this make your azconfig code easier?

sdk/core/azure-core/src/main/java/com/azure/core/http/RequestConditions.java
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @return The datetime that resources must have remained unmodified since.
+     */
+    public OffsetDateTime getIfUnmodifiedSince() {
+        return ifUnmodifiedSince;
+    }
+
+    /**
+     * Optionally limit requests to resources that have remained unmodified since the passed
+     * {@link OffsetDateTime datetime}.
```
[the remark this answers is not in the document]
alzimmermsft: There is one :)

sdk/core/azure-core/src/main/java/com/azure/core/http/MatchConditions.java
mssfang: That makes AzConfig to have more public APIs to maintain. 🔢

Issue #5920, opened 2019-10-18
BlobClient.openInputStream() throws StorageException without ErrorCode if spaces are in blob name


daolis:
If you try to open a input stream on BlobClient a exception is thrown without ErrorCode

Clone test program from: https://github.com/daolis/azureCopBlobInputStreamError

1. Create Azure BlobStore
2. Create Container `testcontainer`
3. Create Blobs `mytestblob.txt` and `my test blob.txt` with content
4. Set Environment Variables: `TEST_ACCOUNT_NAME` and `TEST_ACCOUNT_KEY`

**Expected behavior**
Input stream should be opened and can be used to stream to a OutputStream

**Screenshots**
From example code:
![image](https://user-images.githubusercontent.com/2550938/67082262-153b2a00-f199-11e9-9b95-3a87adcab2ad.png)

**Setup:**
 - OS: Linux
 - IDE : IntelliJ
 - com.azure:azure-storage-blob:12.0.0-preview.4

rickle-msft:
@daolis Thank you for reporting this issue. It looks to be an issue with url encoding the blob name. We will look into the best way to solve this problem as we don't want to always encode a path given to us for fear of double encoding, but we also shouldn't leave customers with this vague error.

@vcolin7 Can you include this in your blobName encoding work? Of note here is that if your blob has a space in the name, and you go to get the url from the portal, the url will have an unencoded space in it, which is what seems to be breaking this api. (Technically what's breaking is the get properties call inside the BlobInputStream constructor. It fails with a 400 because of the unencoded space). It's unfortunate that this is what the portal gives because we have to be able to handle it since customers are inevitably going to grab urls from the portal and give it to us. Perhaps we need to scan the blobName for unencoded characters?

vcolin7:
No problem @rickle-msft, this is included in the blobName changes I've been working on and should be fixed once I merge them.

daolis:
@vcolin7 Thank you! :-)

Additional information from another test:
I tried to create the BlobClient with `testcontainer.getBlobClient("my%20test%20blob.txt");`. Then openInputStream was successful, but `blockBlobClient.upload(stream, length);` failed afterwards.

rickle-msft:
@daolis Thank you for the update. Can you share what the error message was, please?

daolis:
Hi @vcolin7 
Sorry, i also used the blob name with spaces for the target blob.
Then i get the following exception..

[...]

Issue #4324, opened 2019-07-08
[bug] URL Encoding to support Blob names with special characters


alzimmermsft:
Blobs are allowed to have special characters in their name which will need to be encoded when they are used as part of the service URL.

Questions stand on where the encoding should happen, there are a few places where a blob name is set
- Setting the endpoint in the builder (this would likely be a URL)
- Setting the blob name in the builder (this could be from a URL)
- Getting a blob client from the container client (this could be from a URL)

alzimmermsft:
A broader discussion will need to occur given the URL encoding issues found within Files.

alzimmermsft:
Plan is to add documentation to the `blobName` setter methods in our builders to indicate that blob names need to be encoded.

If the encoding process is more complicated add in a helper method to perform the encoding and use that in the documentation.

rickle-msft:
Documentation should also go in the ContainerClient methods that construct any form of blob client. 

We've used the internal Utility method in the past to special case <space>/+/%20 to do the right thing. It has caused problems in the past with SAS signatures in the query parameters, but I can't say off hand whether this will be strictly necessary for the blob names. I tend to think not because we didn't internally encode in v8 and just let the customer do it themselves.

vcolin7:
I tested a few options including the Utility method Rick mentioned and that one seemed to be the best one for UTF-8 encoding while converting spaces into %20.

For testing I used [all printable and extended ASCII characters](https://theasciicode.com.ar/). I also used a bunch of different Unicode characters for languages such as Chinese, Russian, Korean, Greek, Japanese and even more obscure ones like Esperanto. I found no issues when trying to upload or download blobs with names containing these characters while using `Utility.urlEncode()` to encode their names.

I also discovered a couple interesting things:

- Azure Storage does not seem to encode the following ASCII characters when manually generating a SAS URL `!'()*-./\_~`.
- Our utility method does not encode the following ASCII characters: `*-._`.
- Forward slashes `/` and backslashes `\` are both used for establishing virtual hierarchies by the Azure Storage service, which will finally parse them to `/`.
- ASCII characters not allowed on blob names while using the web UI are: `#` and `%`.
- Uploading files that contain a percentage symbol `%` is also not allowed through the Azure SDK.
- However, you can use the Azure SDK to upload blob names that start with a pound/hashtag (`#`) with no encoding. The Azure Service would create a blob with no name in such a case.

[...]

PR #5943, opened 2019-10-18
Added UTF-8 encoding for blob names in methods that are used to build…

vcolin7:
… blob clients or manipulate blobs. Also added documentation where blob names are part of the flow so that the user knows what is happening in case they are thinking of passing an already encoded name.


vcolin7:
This PR is for issue #4324 and also provides a fix for #5920

joshfree:
Thanks @vcolin7 for making this fix to Storage.  Once the tests are updated and you get signoff from at least 1 developer from Storage (@rickle-msft @jaschrep-msft and/or @gapra-msft) please proceed with squashing and merging this PR (hopefully today)

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -136,12 +139,15 @@
      * @throws UnsupportedOperationException If this batch has already added an operation of another type.
      */
     public Response<Void> deleteBlob(String containerName, String blobName) {
-        return deleteBlobHelper(String.format(PATH_TEMPLATE, containerName, blobName), null, null);
+        return deleteBlobHelper(String.format(PATH_TEMPLATE, containerName,
```
rickle-msft: This should just call into the next overload so we don't have to worry about missing formatting/encoding logic in one of them

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
```diff
@@ -217,12 +225,15 @@
      * @throws UnsupportedOperationException If this batch has already added an operation of another type.
      */
     public Response<Void> setBlobAccessTier(String containerName, String blobName, AccessTier accessTier) {
-        return setBlobAccessTierHelper(String.format(PATH_TEMPLATE, containerName, blobName), accessTier, null);
+        return setBlobAccessTierHelper(String.format(PATH_TEMPLATE, containerName,
```
rickle-msft: Same comment about calling the next overload.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -324,6 +329,8 @@ public EncryptedBlobClientBuilder connectionString(String connectionString) {
      * with blobs in the root container, it is best to set the endpoint to the account url and specify the blob name
      * separately using the {@link EncryptedBlobClientBuilder#blobName(String) blobName} method.</p>
      *
+     * <p>Blob name is encoded to UTF-8 using the {@link com.azure.storage.common.Utility#urlEncode(String)} method.</p>
+     *
```
rickle-msft: This seems like it's setting us up for double encoding since they're already handing us an endpoint. I thought we agreed to have an overload that accepts a doNotEncodeBlobName flag or whatever it should be called.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -76,6 +76,8 @@
     /**
      * Package-private constructor for use by {@link BlobClientBuilder}.
      *
+     * <p>Blob name is encoded to UTF-8 using the {@link com.azure.storage.common.Utility#urlEncode(String)} method.</p>
```
rickle-msft: I don't think these comments on the constructors are accurate because the encoding happens in the builder. It seems like it's more accurate to say these constructors expect that the blobName is already properly encoded.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
```diff
@@ -138,11 +140,13 @@ public String getBlobName() {
...
      * Sets the blob name that will be used as part of the URL path.
      *
+     * <p>Blob name is encoded to UTF-8 using the {@link com.azure.storage.common.Utility#urlEncode(String)} method.</p>
+     *
      * @param blobName The blob name.
      * @return the updated BlobUrlParts object.
      */
     public BlobUrlParts setBlobName(String blobName) {
-        this.blobName = blobName;
+        this.blobName = Utility.urlEncode(Utility.urlDecode(blobName));
```
rickle-msft: Overload to skip encoding?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
```diff
@@ -267,6 +273,8 @@ public URL toUrl() {
      * is no path element for the container, the name of this blob in the root container will be set as the
      * containerName field in the resulting {@code BlobURLParts}.</p>
      *
+     * <p>Blob name is encoded to UTF-8 using the {@link com.azure.storage.common.Utility#urlEncode(String)} method.</p>
```
rickle-msft: Another place that feels likely to give us double encoding since we are taking in a URL

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -110,7 +113,7 @@ protected BlobAsyncClientBase(HttpPipeline pipeline, String url, BlobServiceVers
 
         this.accountName = accountName;
         this.containerName = containerName;
-        this.blobName = blobName;
+        this.blobName = Utility.urlEncode(Utility.urlDecode(blobName));
```
rickle-msft: Don't we encode in the builder, so encoding in the constructor again would be double encoding?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -159,6 +162,8 @@ public final String getContainerName() {
     /**
      * Get the blob name.
      *
+     * <p>Blob name is encoded to UTF-8 using the {@link com.azure.storage.common.Utility#urlEncode(String)} method.</p>
```
rickle-msft: I also don't know that we should specify what method is doing the encoding since Utilities are supposed to be implementation details

rickle-msft: We should also probably be decoding on gets if we are encoding on sets. For one, we want it to round trip. For another, what if someone gets the name of one blob so they can set it as the name of a blob in a different container. Now we encode the name on setting it for the first blob, don't decode it in the getter, then encode it again in the second builder.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
```diff
@@ -1172,18 +1173,18 @@ class ContainerAPITest extends APISpec {
         def blobs = cc.listBlobs().iterator()
 
         then:
-        blobs.next().getName() == name
-        blobs.next().getName() == name + "2"
-        blobs.next().getName() == name + "3"
+        Utility.urlDecode(blobs.next().getName()) == name
```
rickle-msft: See above comment on decoding in getters. We should actually be validating that get without decoding is equal to the name.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
vcolin7: Sounds good, I'll change the comments on the constructors.

sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatch.java
vcolin7: You are right, let's make it call the next overload.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
vcolin7: `BlobUrlParts.parse()` takes care of parsing the endpoint and decoding its blob name from UTF-8 in case it is already encoded, to then finally encode it. The reason behind this is that since it's part of the public API, it is entirely possible that people might pass a non-encoded URL as an argument too.

I think there was also a communication mishap on my behalf when I told you about the last approach I discussed with Alan. There are several reasons why I think it's best to just implicitly do encoding for the user instead of letting the responsibility rest with them:

- It makes the API simple to use if they never have to worry about passing encoded or non-encoded blob names, since with the changes from my last commit we will decode first to avoid double encoding. From running some tests I noticed this added little to no overhead.
- It can be argued that having overloaded methods that add a flag is unnecessary work and would make the API look less simple and straightforward, since we could just modify the already existing methods to add an `encodeBlobName` flag as an argument.
- If we added a flag to either an overloaded method or the already existing ones, I don't think it would be much different from letting the users know we need the `blobName` to be encoded and that they can manually use `Utility.urlEncode()` to do so, as you and Alan first discussed in the original issue.

The bottomline is that I believe implicit encoding is the most user-friendly way as long as we make sure we don't double encode anywhere in the flow.

What do you think?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
vcolin7: See answer about overloads above.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
vcolin7: See answer about overloads and double encoding above.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
vcolin7: See answer about overloads and double encoding above.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
vcolin7: I agree we can leave out which method is used for encoding as long as we specify is UTF-8. As for the getters, even though we won't incur in any double encoding (see comment about that and overloads above), it sounds good to me so that users will always get to see and manipulate the original name of the blob.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
vcolin7: See above answer about this same issue.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
vcolin7: I thought of adding said comments wherever a `blobName` or URL/endpoint is passed as an argument to generate a blob client so that the user knew what would happen to it. In hindsight, I think we could omit these comments wherever the `blobName` is not directly altered. What do you think?

As a side note, there are some cases where the `blobName` is set in the constructor, such as `BlobAsyncClientBase`, which is used to create our specialized blob clients.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
rickle-msft: I didn't see that you were doing a urlDecode before urlEncode in the setters. In that case, I think I'm fine with the PR as long as you add the update to return the unencoded name since we're assuming that will be the most common case

 And can you actually also add to either the existing test or a new test that passes in an already encoded name and ensures it doesn't get double encoded. And also getting the name off of one client and passing it to another and again checking it doesn't get double encoded?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
```diff
@@ -1165,9 +1166,9 @@ class ContainerAPITest extends APISpec {
...
+        Utility.urlDecode(blobs.next().getName()) == name + "2"
+        Utility.urlDecode(blobs.next().getName()) == name + "3"
 
         where:
         name          | _
```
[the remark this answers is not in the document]
vcolin7: Uncommented them and ran the tests. They showed some unaccounted-for encoding issues that have now been resolved.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
vcolin7: Will add said getter and test changes.

PR #5948, opened 2019-10-18
Added ability to upload with no overwrite by default

### LINKED ISSUE #5881: [API] Add Simple Upload Overload That Takes an Overwrite Flag
Add an upload overload that takes an additional `overwrite` flag and when the flag is set to false pass an HTTP conditional of `If-None-Match=*` to prevent the current blob from being overwritten.

gapra-msft:
Fixes #5881

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
@@ -210,11 +210,11 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
      *
      * <p><strong>Code Samples</strong></p>
      *
-     * {@codesnippet com.azure.storage.blob.BlobAsyncClient.uploadWithResponse#Flux-ParallelTransferOptions-Boolean-BlobHttpHeaders-Map-AccessTier-BlobAccessConditions}
+     * {@codesnippet com.azure.storage.blob.BlobAsyncClient.uploadWithResponse#Flux-ParallelTransferOptions-boolean-BlobHttpHeaders-Map-AccessTier-BlobAccessConditions}
      *
      * <p><strong>Using Progress Reporting</strong></p>
      *
-     * {@codesnippet com.azure.storage.blob.BlobAsyncClient.uploadWithResponse#Flux-ParallelTransferOptions-Boolean-BlobHttpHeaders-Map-AccessTier-BlobAccessConditions.ProgressReporter}
+     * {@codesnippet com.azure.storage.blob.BlobAsyncClient.uploadWithResponse#Flux-ParallelTransferOptions-boolean-BlobHttpHeaders-Map-AccessTier-BlobAccessConditions.ProgressReporter}
      *
      * @param data The data to write to the blob. Unlike other upload methods, this method does not require that the
      * {@code Flux} be replayable. In other words, it does not have to support multiple subscribers and is not expected


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
```diff
@@ -138,7 +138,7 @@
      */
     public Mono<BlockBlobItem> upload(Flux<ByteBuffer> data, ParallelTransferOptions parallelTransferOptions) {
         try {
-            return this.uploadWithResponse(data, parallelTransferOptions, null, null, null, null)
+            return this.uploadWithResponse(data, parallelTransferOptions, false, null, null, null, null)
```
alzimmermsft: Shouldn't the flag for overwrite be on the minimal overload as that doesn't have the ability to pass HTTP conditionals (If-Match, If-None-Match)? The change is meant to be `overwrite=false` -> add `If-None-Match=*`, the maximum overload has the capability and more to do this.

jaschrep-msft: My understanding of this feature is for there to be a default behavior of no overwrite. If the flag is on the min overload, there is no default behavior, its only ever specified.

I'm not against providing the flag on the min, and then forcing people to actually specify the if-none-match on the max, but we should go into that intentionally deciding there is no default behavior for whether or not to overwrite.

alzimmermsft: From talking with @tg-msft on this change the goal was overwrite by default is false, Java doesn't have default parameters so we'd document what false and true mean, and if overwrite is false we inject an `If-None-Match=*`. From what I've been told this is supposed to change default functionality.

tg-msft: Yes, we actually have the simpler overrides with `Upload(Stream content)` or `Upload(string path)` just call the version with `overwrite` and pass `false`.  https://github.com/Azure/azure-sdk-for-net/blob/78b0c8056bf7984f50d81577faed8d6a11dcdca3/sdk/storage/Azure.Storage.Blobs/src/BlobClient.cs#L479

jaschrep-msft: Yes it is. But if a boolean is on the min overload, there is no default fucntionality, because the user is always forced to specify.

alzimmermsft: We should add another overload to all of the uploading methods which adds a boolean flag on top of the minimal overload, the minimal overload can be documented that it will pass false into the new overload.

tg-msft: For what it's worth, Upload/Download are the only things in .NET where we have multiple overloads so there's a spectrum between min/max.  These are the most important APIs so we're willing to have a few extra layers of convenience.  https://apiview.azurewebsites.net/Assemblies/Review/c52f059a-6d57-4d40-b7d2-a5816565a660#Azure.Storage.Blobs.BlobClient

rickle-msft: It feels strange to special case this one flag and add an overload just for this. Is that just so eventhubs doesn't break?

rickle-msft: If there's a conflict between the flag and the access conditions, we should do the safe option and use the one that doesn't overwrite. 

If the argument for a middle overload is that there may be an inexperienced customer that wants to overwrite and they should be able to just upload and overwrite without using the max overload, the same holds for a customer who wants to overwrite with metadata and properties. Just because they can use metadata, doesn't mean they can safely use accessConditions, so there should be the flag on the max overload, and there's no need for a middle option just for this flag.

PR #5951, opened 2019-10-19
Added @JsonIgnore to getLogger in JsonSerializable

kushagraThapar:
* Added @JsonIgnore to getLogger in JsonSerializable
* Updated version to 3.3.1
* This fix is intended for the issue faced in spring-data-cosmsodb SDK: https://github.com/microsoft/spring-data-cosmosdb/issues/423

### DIFF
--- a/sdk/cosmos/microsoft-azure-cosmos-benchmark/pom.xml
+++ b/sdk/cosmos/microsoft-azure-cosmos-benchmark/pom.xml
@@ -8,12 +8,12 @@ Licensed under the MIT License.
   <parent>
     <groupId>com.microsoft.azure</groupId>
     <artifactId>azure-cosmos-parent</artifactId>
-    <version>3.4.0</version> <!-- {x-version-update;com.microsoft.azure:azure-cosmos-parent;current} -->
+    <version>3.3.1</version> <!-- {x-version-update;com.microsoft.azure:azure-cosmos-parent;current} -->
   </parent>
 
   <groupId>com.microsoft.azure</groupId>
   <artifactId>azure-cosmos-benchmark</artifactId>
-  <version>3.4.0</version> <!-- {x-version-update;com.microsoft.azure:azure-cosmos-benchmark;current} -->
+  <version>3.3.1</version> <!-- {x-version-update;com.microsoft.azure:azure-cosmos-benchmark;current} -->
   <name>Microsoft Azure SDK for SQL API of Azure Cosmos DB Service - Benchmarking tool</name>
   <description>This package contains Benchmarking tool for Microsoft Azure SDK for SQL API of Azure Cosmos DB Service</description>
   <url>https://github.com/Azure/azure-sdk-for-java</url>
@@ -33,7 +33,7 @@ Licensed under the MIT License.
     <dependency>
       <groupId>com.microsoft.azure</groupId>
       <artifactId>azure-cosmos</artifactId>
-      <version>3.4.0</version> <!-- {x-version-update;com.microsoft.azure:azure-cosmos;current} -->
+      <version>3.3.1</version> <!-- {x-version-update;com.microsoft.azure:azure-cosmos;current} -->
     </dependency>
 
     <dependency>
--- a/sdk/cosmos/changelog/README.md
+++ b/sdk/cosmos/changelog/README.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### 3.3.1
+- Added @JsonIgnore on getLogger in JsonSerializable
+
 ### 3.3.0
 - **Version Upgrades**
     - Project Reactor version upgrade to 3.3.0.RELEASE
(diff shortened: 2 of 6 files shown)


sdk/cosmos/microsoft-azure-cosmos-benchmark/pom.xml
```diff
@@ -8,12 +8,12 @@ Licensed under the MIT License.
   <parent>
     <groupId>com.microsoft.azure</groupId>
     <artifactId>azure-cosmos-parent</artifactId>
-    <version>3.4.0</version> <!-- {x-version-update;com.microsoft.azure:azure-cosmos-parent;current} -->
```
mbhaskar: Just to double check 3.4.0 was just a snapshot right?

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/JsonSerializable.java
```diff
@@ -32,14 +33,14 @@
     private final static ObjectMapper OBJECT_MAPPER = Utils.getSimpleObjectMapper();
     private ObjectMapper om;
     transient ObjectNode propertyBag = null;
-
```
mbhaskar: Please refrain from making unrelated formatting changes, you can configure intellij to stop doing this automatically

kushagraThapar: I really wanted to, but couldn't stop it. :( 
You can show me a demo sometime :)

sdk/cosmos/microsoft-azure-cosmos-benchmark/pom.xml
kushagraThapar: Not anymore. 
The practice has been changed to update the versions to next to be released version.

JimSuplizio:
CC: @weshaggard 
@kushagraThapar Why is the version of this ticked back down from 3.4.0 to 3.3.1? Is this supposed to be a hotfix? For hotfixes we generally recommend branching from the last release tag (https://github.com/Azure/azure-sdk/blob/master/docs/policies/repobranching.md#hotfix-branches) and applying just the fix you want, otherwise you risk picking up other changes you didn't want for that hotfix.

Last but not least, when versioning files please take a look at the **Versions and versioning** section of https://github.com/Azure/azure-sdk-for-java/blob/master/CONTRIBUTING.md. The versions of Cosmos versions should be updated in [version_data.txt](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/version_data.txt) and then the [update_versions.py](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/update_versions.py) script should be run which will update the pom and readme files. There's a secondary PR for this to update the version in java files as well which is going to run into merge issues of these version changes.

kushagraThapar:
> CC: @weshaggard
> @kushagraThapar Why is the version of this ticked back down from 3.4.0 to 3.3.1? Is this supposed to be a hotfix? For hotfixes we generally recommend branching from the last release tag (https://github.com/Azure/azure-sdk/blob/master/docs/policies/repobranching.md#hotfix-branches) and applying just the fix you want, otherwise you risk picking up other changes you didn't want for that hotfix.
> 
> Last but not least, when versioning files please take a look at the **Versions and versioning** section of https://github.com/Azure/azure-sdk-for-java/blob/master/CONTRIBUTING.md. The versions of Cosmos versions should be updated in [version_data.txt](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/version_data.txt) and then the [update_versions.py](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/update_versions.py) script should be run which will update the pom and readme files. There's a secondary PR for this to update the version in java files as well which is going to run into merge issues of these version changes.

@JimSuplizio, this wasn't a hotfix, but more like a small bug fix (that was needed in spring-data-cosmosdb SDK) that wasn't related to any previous changes. 
Since the master wasn't updated since release 3.3.0, so released it from master itself. 
This PR: https://github.com/Azure/azure-sdk-for-java/pull/5959 changes the version back to 3.4.0 to avoid any confusion. 

Going forward, I will make sure to create a hotfix branch, and follow the hotfix pattern for any bug fixes, or hotfixes releases.

Regarding tags, I couldn't find any tags under azure-sdk-for-java that were related to cosmos SDK previous releases. Since cosmos SDK and azure-sdk-for-java follow different release cycles, how can we create tags for previous versions that got released, and for this release as well?

Also, do we need to update version_data.txt every time we release a new cosmos SDK version?

For now, I can create a PR that updates the version_data.txt file. Will that be okay?

weshaggard:
>Regarding tags, I couldn't find any tags under azure-sdk-for-java that were related to cosmos SDK previous releases. Since cosmos SDK and azure-sdk-for-java follow different release cycles, how can we create tags for previous versions that got released, and for this release as well?

The github UI doesn't do well with listing all the tags I suggest using the command line `git tag -l azure-cosmos*` should find you all the release tags for cosmos. Currently it only lists 3.2.2 but I suspect that is because the 3.3.0 package must not have been released with the correct release pipeline.

Issue #3241, opened 2019-03-28
[Discuss] Enable a way to choose between Exponential vs Fixed retry for RetryPolicy


anuchandy:
Explore the following operators:

1. https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#retry-long-
2. https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#retryBackoff-long-java.time.Duration-java.time.Duration-double-

mrm9084:
The App Configuration Service team would be interested in this being implemented. The App Configuration Provider currently has this being added, but when we switch over to using the SDK it would be lost.

Issue #5945, opened 2019-10-18
RetryPolicy should support custom retry headers for different Azure services 


srnagar:
Currently, RetryPolicy uses `retry-after-ms` to determine the delay duration. However, different Azure services return the delay duration in different response headers. 
So, we should look into supporting different headers for delay duration.

For e.g. AppConfig uses `retry-after-ms` while CosmosDB uses `x-ms-retry-after-ms`.

RetryPolicy should also allow a customized list of HTTP status codes that are retryable. 
https://azure.github.io/azure-sdk/general_azurecore.html#retry-policy

hemanttanwar:
PR :  https://github.com/Azure/azure-sdk-for-java/pull/6196

PR #5953, opened 2019-10-20
Enable exponential backoff for retry policy in http pipeline

srnagar:
Enable RetryPolicy to use various retry strategies including fixed delay and exponential backoff.

Fixes https://github.com/Azure/azure-sdk-for-java/issues/3241

@JonathanGiles should this go in before GA?

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
@@ -26,7 +26,7 @@ public class RetryPolicy implements HttpPipelinePolicy {
     private final RetryStrategy retryStrategy;
 
     /**
-     * Creates a {@link FixedDelay} RetryPolicy with a default {@link ExponentialBackoff} strategy.
+     * Creates a default {@link ExponentialBackoff} retry policy.
      */
     public RetryPolicy() {
         this(new ExponentialBackoff());


alzimmermsft:
Could the Storage retry policy and the enum in Storage Common be rolled up to using this?

Have you cross referenced the design spec to align with it? https://azure.github.io/azure-sdk/general_azurecore.html#retry-policy

srnagar:
This provides exponential and fixed delay options. The default policy was fixed earlier, so, didn't change it. As per the guidelines, the default should be exponential. 

I can change it if that's what we want to go with for GA but since this is a change in azure-core that may increase the risk of breaking tests in client libraries.

Note that there are more things to do for retry policy - https://github.com/Azure/azure-sdk-for-java/issues/5945

Yes, storage can utilize this change by either using the provided implementations for fixed/exponential delay or by providing a custom implementation specific to storage.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License.
+
+package com.azure.core.http.policy;
+
+import java.time.Duration;
+
+/**
+ * The interface for determining the retry strategy used in {@link RetryPolicy}.
+ */
+public interface RetryStrategy {
```
alzimmermsft: Could this interface have an apply retry/attempt retry added to it? This way the strategy implementation is the only thing the RetryPolicy needs to run.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -22,26 +25,32 @@
...
      *
      * @param maxRetries the maximum number of retries to attempt.
-     * @param delayDuration the delay between retries
+     * @param delayDuration the delay between retries.
      */
     public RetryPolicy(int maxRetries, Duration delayDuration) {
-        this.maxRetries = maxRetries;
-        this.delayDuration = delayDuration;
+        this(new FixedDelay(maxRetries, delayDuration));
+    }
```
JonathanGiles: I wonder if we bite the bullet and remove this API. We do not believe fixed retry policies are good, and if we removed this it would be easier to remove support for this in the future. Thoughts?

alzimmermsft: I'm not against it, fixed delay unless prompted by the service for the delay time are very rigid.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Max number of retry attempts to be make.
+     * @return The max number of retry attempts.
+     */
+    int getMaxRetries();
```
anuchandy: It would be nice to think about having a method `boolean shouldRetry(HttpResponse response)` so that user can control the retry behavior by inspecting the response. If we have such a method then we could remove `getMaxRetries()` method. (since max retry can also be controlled by managing necessary state inside `shouldRetry` implementation)

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
srnagar: Looks like most of the places used default retry policy and none of them used this overload. So, I have removed it and made the default use exponential backoff as per spec.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryStrategy.java
srnagar: As discussed with @anuchandy, I have added a default implementation for `shouldRetry()` in the interface. Implementations can override if necessary.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryStrategy.java
srnagar: I have added a default method called `shouldRetry()` to the interface which will make it more flexible while the RetryPolicy is still in control of actually attempting the retry.

PR #5955, opened 2019-10-20
Storage Blobs Extends and Uses RequestConditions

### LINKED ISSUE #5939: Update Blob Access Condition Classes
The Blob access condition classes need to be updated to extend the Azure Core access condition models, and at the same time flatten their structure for easier use.

BlobAccessConditions extends RequestAccessConditions
- String: leaseId

AppendBlobAccessConditions extends BlobAccessConditions
- Long: maxSize
- Long: appendPosition

PageBlobAccessConditions extends BlobAccessConditions
- Long: ifSequenceNumberLessThanOrEqualTo
- Long: ifSequenceNumberLessThan
- Long: ifSequenceNumberEqualTo

alzimmermsft:
Fixes #5939 

Updates the Storage Blob access conditions classes to extend those in Azure Core.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/SpecializedBlobClientBuilder.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/SpecializedBlobClientBuilder.java
@@ -125,8 +125,8 @@ public final class SpecializedBlobClientBuilder {
      * Creates a {@link BlockBlobAsyncClient} based on options set in the Builder. BlockBlobAsyncClients are used to
      * perform generic upload operations such as {@link BlockBlobAsyncClient#upload(Flux, long) upload from file}
      * and block blob specific operations such as {@link BlockBlobAsyncClient#stageBlockWithResponse(String, Flux, long,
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/AzureBlobStorageBuilder.java
```diff
@@ -91,7 +91,11 @@ public AzureBlobStorageImpl build() {
         if (this.url != null) {
             client.setUrl(this.url);
         }
-        client.setVersion(this.version);
+        if (this.version != null) {
+            client.setVersion(this.version);
+        } else {
+            client.setVersion("2019-02-02");
```
JonathanGiles: We should never let version be null

alzimmermsft: Codegen artifact, our client builders should never get into this state.

alzimmermsft:
Common transient test failure, pipelines has an issue where individual steps cannot be re-ran, merging.

PR #5961, opened 2019-10-21
Fixes Swagger Generation Naming and Moves Properties Classes

### LINKED ISSUE #5925: Use Swagger Transforms to Match API Spec
There is going to be another round of API reviews to find where the codegen classes are not inline with the naming guidelines, this issue is meant to track all found instances.

alzimmermsft:
Fixes #5925 

Updates Swagger generation to follow API naming specs and removed the `blob` prefix from the `BlobHttpHeaders` setters and getters.

Additionally moved the `BlobProperties` and `BlobContainerProperties` classes into the models package and updated their constructor to tag value parameters instead of the implementation class.


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerProperties.java
```diff
@@ -1,13 +1,7 @@
...
 // Licensed under the MIT License.
 
-package com.azure.storage.blob;
-
-import com.azure.storage.blob.implementation.models.ContainerGetPropertiesHeaders;
-import com.azure.storage.blob.models.LeaseDurationType;
-import com.azure.storage.blob.models.LeaseStateType;
-import com.azure.storage.blob.models.LeaseStatusType;
-import com.azure.storage.blob.models.PublicAccessType;
+package com.azure.storage.blob.models;
```
conniey: These classes can be annotated with `@Immutable`. I had forgot to mention that on a few of the non-generated models I saw.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
```diff
@@ -1,17 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
```
conniey: Immutable if possible.

PR #5964, opened 2019-10-21
Fix storage live tests and bug in Poller

conniey:
* Fixes bugs in the test code failing Storage live tests.
* Swapped from using the incorrect copyfromUrl method to Poller one.
* Poller was invoking activation operation twice, which would fail some of the tests.
* Updated the records for the tests that were fixed.
* 37 of the files are the record files. There are only 1 java file change and 2 groovy file changes.


sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -193,8 +193,8 @@ public Poller(Duration pollInterval, Function<PollResponse<T>, Mono<PollResponse
             .flux()
             .repeat()
             .takeUntil(pollResponse -> isComplete())
-            .share()
-            .delaySubscription(onActivation);
```
sima-zhu: What's the difference here?

conniey: Anu has a good description of the difference in Observation 1
https://github.com/Azure/azure-sdk-for-java/pull/5932#discussion_r336798299

PR #5979, opened 2019-10-22
Handle Empty Buffers in Size Agnostic Upload

### LINKED ISSUE #5975: [Storage] BlobClientAsync.upload() fails if ByteBuffer is empty
`BlobClientAsync.upload()` fails if the `Flux<ByteBuffer>` contains an empty `ByteBuffer`:

## Code
```java
BlobAsyncClient.upload(Flux.just(ByteBuffer.wrap(new byte[0])), new ParallelTransferOptions())
```

## Exception
```
Exception in thread "main" com.azure.storage.blob.models.StorageException: Status code 400, "﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidHeaderValue</Code><Message>The value for one of the HTTP headers is not in the correct format.
RequestId:55b592ae-001e-0094-1866-88ad4f000000
Time:2019-10-21T23:20:09.7166120Z</Message><HeaderName>Content-Length</HeaderName><HeaderValue>0</HeaderValue></Error>"
```

alzimmermsft:
Fixes #5975 

Adds support to handle empty buffers in the data stream or an empty data stream in the size agnostic upload method.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
@@ -252,7 +252,7 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
             preserve the ordering of the buffers, but we don't really care if one is split before another.
              */
             Flux<ByteBuffer> chunkedSource = data
-                .filter(buffer -> buffer.remaining() > 0)
+                .filter(ByteBuffer::hasRemaining)
                 .flatMapSequential(buffer -> {
                     if (buffer.remaining() <= blockSize) {
                         return Flux.just(buffer);


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -253,19 +251,21 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
-                return Flux.range(0, numSplits)
-                    .map(i -> {
-                        ByteBuffer duplicate = buffer.duplicate().asReadOnlyBuffer();
-                        duplicate.position(i * blockSize);
-                        duplicate.limit(Math.min(duplicate.limit(), (i + 1) * blockSize));
-                        return duplicate;
-                    });
-            });
+            Flux<ByteBuffer> chunkedSource = data
+                .filter(buffer -> buffer.remaining() > 0)
```
mikeharder: Would this be better?

```suggestion
                .filter(buffer -> buffer.hasRemaining())
```

Issue #5971, opened 2019-10-21
Blocks Preview 5 ADLS: Missing API - PagedIterable map function to return a PagedIterable 


gapra-msft:
Related to issue #5825
Storage needs a way to go from a PagedIterable of A to a PagedIterable of C to implement DataLake features.

Some datalake APIs (that are a separate namespace from blobs) call directly into blob APIs (ex: listBlobContainers) that returns a PagedIterable of BlobContainerItems.

To make the namespace for a user consistent, we need to be able to return a PagedIterable of SomeOtherItem.

For example,
```
    public PagedIterable<FileSystemItem> listFileSystems(ListFileSystemsOptions options, Duration timeout) {
        return blobServiceClient.listBlobContainers(Transforms.toListBlobContainersOptions(options), timeout).to(blobContainerItem -> Transforms.toFileSystemItem(blobContainerItem));
    }
```

joshfree:
@srnagar will own investigating this, similar to #5825

PR #5982, opened 2019-10-22
Support for mapping paged iterable from one type to another

srnagar:
The intent of this change is to support converting a PagedIterable of one type to another type. 

Fixes https://github.com/Azure/azure-sdk-for-java/issues/5971

### DIFF
--- a/sdk/core/azure-core/src/samples/java/com/azure/core/http/rest/PagedIterableBaseJavaDocCodeSnippets.java
+++ b/sdk/core/azure-core/src/samples/java/com/azure/core/http/rest/PagedIterableBaseJavaDocCodeSnippets.java
@@ -1,0 +1,111 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.http.rest;
+
+import java.util.Iterator;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import reactor.core.publisher.Mono;
+
+/**
+ * Code snippets for {@link PagedIterableBase}
+ */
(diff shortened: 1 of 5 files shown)


sdk/core/azure-core/src/samples/java/com/azure/core/http/rest/PagedIterableBaseJavaDocCodeSnippets.java
```diff
@@ -0,0 +1,111 @@
...
+     * Provides an example for iterate over each response using streamByPage function.
+     **/
+    public void streamByPageSnippet() {
+        // BEGIN: com.azure.core.http.rest.pagedIterableBase.streamByPage
+        // process the streamByPage
+        CustomPagedFlux<String> customPagedFlux = createCustomInstance();
+        PagedIterableBase<String, PagedResponse<String>> customPagedIterableResponse =
+            new PagedIterableBase<>(customPagedFlux);
+        customPagedIterableResponse.streamByPage().forEach(resp -> {
+            System.out.printf("Response headers are %s. Url %s  and status code %d %n", resp.getHeaders(),
```
joshfree: nit: anyway to avoid the same exact code being written 3x in this file?  otherwise lgtm

srnagar: These are code snippets that get inserted into our Javadocs. They all have unique codesnippet tags and get inserted for different methods. So, in this case, when viewed on javadocs will provide a clear self-contained sample at the cost of a bit of duplication.

Issue #5923, opened 2019-10-18
Azure Core - Poller  - Block / Block Until [State Checks missing]


g2vinay:
Block - check if the poller has not already competed before blocking.
Block Until - check poller is not already in the target status and is not already completed.

in azure-core/util/polling/Poller.java

joshfree:
Related https://github.com/Azure/azure-sdk-for-java/issues/5809 and https://github.com/Azure/azure-sdk-for-java/issues/5805

joshfree:
@anuchandy could you please coordinate with @g2vinay on the related Poller reliability issues #5805 #5809 #5923

anuchandy:
@joshfree - I'm reviewing the current implementation of poller, left some observations here - https://github.com/Azure/azure-sdk-for-java/pull/5932#pullrequestreview-304279959, still looking into the code.

anuchandy:
We're working updating Poller implementation, this issue is related to multi-casting and not applicable in the new design.

Issue #5809, opened 2019-10-13
[BUG] Cancelling Poller operation does not cancel observer.


conniey:
## Expect
After cancelling a poller operation, the observer should report a completion or error.

A side note is that calling cancellation does not stop polling for responses from the service until the service explicitly returns USER_CANCELLED.  I would expect that if a user calls Poller.cancelOperation(), the status is immediately set to USER_CANCELLED after the `cancelOperation` completes.

## Actual
It hangs
```java
@Test
    public void subscribeToAllPollEventStopPollingAfterNSecondsTest() throws Exception {
        // Assert
        Duration pollInterval = Duration.ofSeconds(1);
        Duration waitTime = Duration.ofSeconds(3);
        PollResponse<Response> success = new PollResponse<>(OperationStatus.SUCCESSFULLY_COMPLETED, new Response("Created: Cert A"));
        PollResponse<Response> inProgress = new PollResponse<>(PollResponse.OperationStatus.IN_PROGRESS,new Response("Starting : Cert A"));

        when(pollOperation.apply(any())).thenReturn(Mono.just(inProgress), Mono.just(inProgress), Mono.just(success));

        // Act
        Poller<Response> poller = new Poller<>(pollInterval, pollOperation, null,
            ignored -> new PollResponse<Response>(OperationStatus.USER_CANCELLED, null));

        // Assert
        StepVerifier.create(poller.getObserver())
            .expectNext(inProgress)
            .then(() -> poller.setAutoPollingEnabled(false))
            .expectNoEvent(waitTime)
            .then(() -> poller.cancelOperation())
            .verifyError();

        Assert.assertEquals(OperationStatus.IN_PROGRESS, poller.getStatus());
        Assert.assertFalse(poller.isAutoPollingEnabled());
    }
```

anuchandy:
We're reworking Poller and this will be automatically enabled by new publisher based Poller. See this issue https://github.com/Azure/azure-sdk-for-java/issues/6036 for details

Issue #5805, opened 2019-10-13
Polling does keeps polling even when setAutoPollingEnabled is false


conniey:
Expects that this test case passes. But it fails on: `.expectNoEvent`

```java
@Mock
private Function<PollResponse<Response>, Mono<PollResponse<Response>>> pollOperation;

@Test
public void subscribeToAllPollEventStopPollingAfterNSecondsAndRestartedTest() {
    // Arrange
    final PollResponse<Response> successPollResponse = new PollResponse<>(OperationStatus.SUCCESSFULLY_COMPLETED, new Response("Created : Cert A"), Duration.ofSeconds(1));
    final PollResponse<Response> inProgressPollResponse = new PollResponse<>(OperationStatus.IN_PROGRESS, new Response("Starting : Cert A"));
    final Duration pollInterval = Duration.ofMillis(100);

    when(pollOperation.apply(any())).thenReturn(Mono.just(inProgressPollResponse), Mono.just(successPollResponse));

    // Act
    final Poller<Response> poller = new Poller<>(pollInterval, pollOperation);

    // Assert
    StepVerifier.create(poller.getObserver())
        .expectNext(inProgressPollResponse)
        .then(() -> poller.setAutoPollingEnabled(false))
        .expectNoEvent(Duration.ofSeconds(3))
        .then(() -> poller.setAutoPollingEnabled(true))
        .expectNext(successPollResponse)
        .verifyComplete();

    Assert.assertEquals(OperationStatus.SUCCESSFULLY_COMPLETED, poller.getStatus());
    Assert.assertTrue(poller.isAutoPollingEnabled());
}
```

anuchandy:
We're working on new poller design and killing the idea of applying on-off on auto-polling, it will be achievable via regular reactor operator. See this issue https://github.com/Azure/azure-sdk-for-java/issues/6036 for details.

PR #5987, opened 2019-10-22
Poller - Sync poller and State less asynchronous poller (a.k.a PollerFlux)

anuchandy:
This PR is for redesigning the Poller. Highlights are:

1. Clear separation of sync and async polling.
2. The async poller follows reactor pattern and implementation ensure there is no concurrency/race issues.
3. The async poller enables customers to retrieve final result and performing LRO cancellation from within operator chain (the idiomatic reactor style).
5. The polling can be cancelled reactive way by simply unsubscribing (similar to .net cancellationToken signaling).
6. The poller is open for extensibility (including resumePolling/rehydration of polling)
7. While we use client SDK as the basic use case for poller, we also thought about advanced polling involved in management plane and this design open for that as well.

The design adhere to general guideline on Polling.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
@@ -3,6 +3,7 @@
 
 package com.azure.core.util.polling;
 
+import com.azure.core.implementation.util.FluxUtil;
 import com.azure.core.util.logging.ClientLogger;
 import reactor.core.CoreSubscriber;
 import reactor.core.publisher.Flux;
@@ -140,9 +141,10 @@ public final class PollerFlux<T, U> extends Flux<AsyncPollResponse<T, U>> {
             if (this.guardActivationCall.compareAndSet(this, 0, 1)) {
                 final Mono<T> activationMono;
                 try {
-                    activationMono = activationOperation.apply(this.rootContext);
-                } catch (Throwable throwable) {
-                    return Mono.error(throwable);
+                    activationMono = this.activationOperation.apply(this.rootContext);
+                } catch (RuntimeException e) {
+                    this.guardActivationCall.compareAndSet(this, 1, 0);
+                    return FluxUtil.monoError(logger, e);
                 }
                 //
                 return activationMono.map((T result) ->
@@ -154,7 +156,7 @@ public final class PollerFlux<T, U> extends Flux<AsyncPollResponse<T, U>> {
                     this.activated = true;
                     return true;
                 })
-                .doOnError(throwable -> guardActivationCall.compareAndSet(this, 1, 0));
+                .doOnError(throwable -> this.guardActivationCall.compareAndSet(this, 1, 0));
             } else {
                 return Mono.empty();
             }
@@ -171,8 +173,8 @@ public final class PollerFlux<T, U> extends Flux<AsyncPollResponse<T, U>> {
             // Create a Polling Context per subscription
             () -> this.rootContext.copy(),
             // Do polling
-            // set|read in state as needed, reactor guarantee thread-safety of state object.
-            cxt -> Mono.defer(() -> pollOperation.apply(cxt))
+            // set|read to|from context as needed, reactor guarantee thread-safety of cxt object.
+            cxt -> Mono.defer(() -> this.pollOperation.apply(cxt))
                 .delaySubscription(getDelay(cxt.getLatestResponse()))
                 .switchIfEmpty(Mono.error(new IllegalStateException("PollOperation returned Mono.empty().")))
                 .repeat()
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollingContext.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollingContext.java
@@ -10,19 +10,20 @@ import java.util.Map;
 import java.util.Objects;
 
 /**
- * The context type whose instance is used to share context across calls on
- * various polling related operations for a specific long-running operation.
+ * A key/value store that is propagated between various poll related operations associated with
+ * {@link PollerFlux} and {@link SyncPoller} poller. The context also expose activation and
+ * latest {@link PollResponse}.
  *
  * @param <T> the type of the poll response.
  */
 public final class PollingContext<T> {
     private final ClientLogger logger = new ClientLogger(PollingContext.class);
     private final Map<String, String> map;
-    private PollResponse<T> activationResponse;
-    private PollResponse<T> latestResponse;
+    private volatile PollResponse<T> activationResponse;
+    private volatile PollResponse<T> latestResponse;
 
     /**
-     * Get a data from the context with the specific key.
+     * Get a value from the context with the provided key.
      *
      * @param name the key to look for
      * @return the value of the key if exists, else null
@@ -32,10 +33,10 @@ public final class PollingContext<T> {
     }
 
     /**
-     * Set data in the context.
+     * Set a key-value pair in the context.
      *
-     * @param name the key for data
-     * @param value the value of data
+     * @param name the key
+     * @param value the value
      * @return an updated PollingContext
      */
     public PollingContext<T> setData(String name, String value) {
@@ -116,6 +117,7 @@ public final class PollingContext<T> {
                 "'activationResponse' cannot be null.");
         this.latestResponse = Objects.requireNonNull(latestResponse,
                 "'latestResponse' cannot be null.");
-        this.map = map;
+        this.map = Objects.requireNonNull(map,
+                "'map' cannot be null.");
     }
 }


sdk/core/azure-core/src/main/java/com/azure/core/util/polling/implementation/LongRunningOperation.java
```diff
@@ -0,0 +1,205 @@
...
+//
+//    pollFlux
+//       .takeUntil(pr -> pr.getStatus() == PollResponse.OperationStatus.IN_PROGRESS)
+//       .block();
+//
+//    pollFlux
+//       .timeout(Duration.ofMinutes(15))
+//       .flatmap(pr ->
+//           if(pr.getStatus() == PollResponse.OperationStatus.FAILED)) {
+//              return pr.cancel();
```
jianghaolu: do we need to cancel if failed? I expect to call cancel() in `.switchIfEmpty()`, i.e. if the poll operation times out after 15 minutes.

anuchandy: you're right, the sample is wrong, will fix it.

Draft is still in progress and I probably used bad names for types like AsyncPollResponse, it's mainly to communicate the concept.

anuchandy: updated sample to show case a possible use case of `cancel()`.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/implementation/LongRunningOperation.java
```diff
@@ -0,0 +1,212 @@
...
+                return lroStartMono.map((PollResultT response) -> {
+                    this.lroStartResponse = new PollResponse<>(PollResponse.OperationStatus.NOT_STARTED, response);
+                    return this.lroStartResponse;
+                })
+                .switchIfEmpty(Mono.error(new RuntimeException("lroStartMono should produce a Response.")))
+                .doOnError(throwable -> guardLroStartCall.compareAndSet(this, 1, 0));
+            } else {
+                return Mono.empty();
+            }
+        }).repeatWhenEmpty((Flux<Long> longFlux) -> longFlux.concatMap(ignored -> Flux.just(true)));
```
jianghaolu: Is the call to `repeatWhenEmpty` here for threads which failed to acquire the lock to wait for the 1st thread to complete the work?

anuchandy: Yes, that is correct. we do busy loop here instead of blocking to handle very rare but still possible concurrency bug

anuchandy: And we want only one successful call to happen on start lro

joshfree:
Associating PR with tracking issues:

Fixes #5805, Fixes #5809, Fixes #5923

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -0,0 +1,247 @@
...
+ *       .last()
+ *       .flatMap(asyncPollResponse -> {
+ *           // LRO completed, retrieving final result.
+ *           return asyncPollResponse.getFinalResult();
+ *       }).block();
+ *
+ * @param <T> The type of poll response value
+ * @param <U> The type of the final result of long-running operation
+ */
+public final class PollerFlux<T, U> extends Flux<AsyncPollResponse<T, U>> {
```
JonathanGiles: I'm not sure if this needs to be public API?

anuchandy: Update: Synced with @JonathanGiles offline about possible extensibility point on `PollerFlux` e.g. support for "rehydrating Polling from an ID at a later stage" (from a totally different process/machine)

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/Poller.java
```diff
@@ -1,499 +1,496 @@
...
-     * Current known status as a result of last poll event or last response from a manual polling.
-     *
-     * @return Current status or {@code null} if no status is available.
-     */
-    public OperationStatus getStatus() {
-        final PollResponse<T> current = this.pollResponse;
-        return current != null ? current.getStatus() : null;
-    }
-}
+//// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
anuchandy: my bad, It will be deleted.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
```diff
@@ -0,0 +1,122 @@
...
+        this.cancellationOperation = Objects.requireNonNull(cancellationOperation);
+        this.fetchResultOperation = Objects.requireNonNull(fetchResultOperation);
+        this.innerResponse = Objects.requireNonNull(innerResponse);
+    }
+
+    /**
+     * Represents the status of the long-running operation at the time the last polling operation finished successfully.
+     * @return A {@link LongRunningOperationStatus} representing the result of the poll operation.
+     */
+    public LongRunningOperationStatus getStatus() {
```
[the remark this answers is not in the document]
jianghaolu: I raised the same concern with Anu today - but I came to like it for 2 reasons - 1) It disambiguates itself from other `OperationStatus` classes in other services. 2) It's easy to find this class by searching globally for `LongRunningOperation`.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -0,0 +1,262 @@
...
+                })
+                .switchIfEmpty(Mono.defer(() -> {
+                    this.activationResponse = new PollResponse<>(LongRunningOperationStatus.NOT_STARTED, null);
+                    return Mono.just(this.activationResponse);
+                }))
+                .doOnError(throwable -> guardActivationCall.compareAndSet(this, 1, 0));
+            } else {
+                return Mono.empty();
+            }
+        }).repeatWhenEmpty((Flux<Long> longFlux) -> longFlux.concatMap(ignored -> Flux.just(true)));
```
srnagar: Should there be a limit to how many times/how long this repeat should run? If the `activationOperation.get()` takes a long time, this might spin for a bit which can cause high CPU usage.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -0,0 +1,262 @@
...
+        return Flux.using(
+            // Create a state per subscription
+            () -> new State<>(activationResponse),
+            // Do polling
+            // set|read in state as needed, reactor guarantee thread-safety of state object.
+            state -> Mono.defer(() -> pollOperation.apply(activationResponse, state.getLastResponse()))
+                .delaySubscription(getDelay(state.getLastResponse()))
+                .repeat()
+                .takeUntil(currentPollResponse -> currentPollResponse.getStatus().isComplete())
+                .onErrorResume(throwable -> {
```
srnagar: If the error is not recoverable, we continue to poll until the user cancels?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -0,0 +1,262 @@
...
+                    logger.warning("Received an error from pollOperation. Any error from pollOperation " +
+                        "will be ignored and polling will be continued.");
```
srnagar: Include the exception in the log message.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    PollResponse<T> poll();
+
+    /**
+     * Wait for polling to complete. The polling is considered complete based on status defined in
+     * {@link LongRunningOperationStatus}.
+     *
+     * @return the final poll response
+     */
+    PollResponse<T> waitForCompletion();
```
srnagar: In sync world, if this method is called, this will block indefinitely with no option for user to cancel. Should we always enforce a timeout for `waitUntil` methods?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
```diff
@@ -0,0 +1,72 @@
...
+     *         does not complete within the given {@link Duration}.
+     */
+    PollResponse<T> waitForCompletion(Duration timeout);
+
+    /**
+     * Wait until the given {@link LongRunningOperationStatus} is received.
+     *
+     * @param statusToWaitFor the desired {@link LongRunningOperationStatus} to block for.
+     * @return {@link PollResponse} whose {@link PollResponse#getStatus()} matches {@code statusToWaitFor}.
+     * @throws IllegalArgumentException if {@code timeout} is zero or negative and if {@code statusToWaitFor} is
```
srnagar: This method doesn't take a timeout. Doc needs to be updated.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Wait until the given {@link LongRunningOperationStatus} is received.
+     *
+     * @param statusToWaitFor the desired {@link LongRunningOperationStatus} to block for.
+     * @param timeout the duration to waits for the polling.
+     * @return {@link PollResponse} whose {@link PollResponse#getStatus()} matches {@code statusToWaitFor}.
+     * @throws java.util.concurrent.TimeoutException TimeoutException will be thrown if polling
+     *         does not find the matching status within the given {@link Duration}.
+     */
+    PollResponse<T> waitUntil(LongRunningOperationStatus statusToWaitFor, Duration timeout);
```
srnagar: Worth considering `waitUntilStatusOrCompletion(LongRunningOperationStatus statusToWaitFor)` overload which waits until the requested status is reached or the operation is completed, whichever is earlier.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
anuchandy: Discussed offline, we'll keep both methods `waitUntil()` and `waitUntil(Timeout)` to simplify the basic user experience. We also looked to see any other APIs has similar behavior - so far we identified the blocking calls in reactor [`Flux::blockFirst()`, `Flux::blockFirst(Timeout)`] and `Future.get()` `Future.get(Timeout)`.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
anuchandy: nice, let's add it.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
```diff
@@ -0,0 +1,122 @@
...
+        return innerResponse.getValue();
+    }
+
+    /**
+     * Returns the delay the service has requested until the next polling operation is performed. A null or negative
+     * value will be taken to mean that the Poller should determine on its own when the next poll operation is
+     * to occur.
+     * @return Duration How long to wait before next retry.
+     */
+    public Duration getRetryAfter() {
```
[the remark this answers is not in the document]
anuchandy: yes, I don't see any need to expose this. Let's make it package private

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @param activationResponse the response from activation operation
+     * @param innerResponse the {@link PollResponse} this type composes
+     * @param cancellationOperation the cancellation operation if supported by the service
+     * @param fetchResultOperation the operation to fetch final result of long-running operation, if supported
+     *                             by the service
+     */
+    AsyncPollResponse(PollResponse<T> activationResponse,
+                      PollResponse<T> innerResponse,
+                      BiFunction<PollResponse<T>, PollResponse<T>, Mono<T>> cancellationOperation,
```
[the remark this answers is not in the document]
anuchandy: yes, will do.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
anuchandy: the current assumption is - in such cases we expect the SDK implementer to return a `PollResponse` with failed status and its value as details of failure along with any service returned data. 

Looking at the current Poller's code comment - it says: "We expect Mono from pollOperation should never return Mono.error(). If any unexpected scenario happens in pollOperation, it should catch it and return a valid PollResponse."

I'm not sure this choice was based on some architecture review - @JonathanGiles @hemanttanwar do you remember?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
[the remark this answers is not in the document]
anuchandy: actually let's kill it, we're not using it anyway even internally.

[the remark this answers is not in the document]
anuchandy: I overlooked, actually this is needed for DefaultSyncPoller that is built-on PollerFlux.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
```diff
@@ -0,0 +1,72 @@
...
+ * @param <T> The type of poll response value.
+ * @param <U> The type of the final result of long-running operation.
+ */
+public interface SyncPoller<T, U> {
+    /**
+     * Poll once and return the poll response received.
+     *
+     * @return the poll response
+     */
+    PollResponse<T> poll();
```
JonathanGiles: Does this force another poll to the server, or does it just return the last poll response? If it forces a poll, this may lead to people overwhelming services with poll requests.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * A map of properties provided by the service that will be made available into the next poll operation.
+     *
+     * @return Map of properties that were returned from the service, and which will be made available into the next
+     *     poll operation.
+     */
+    public Map<Object, Object> getProperties() {
+        return innerResponse.getProperties();
+    }
```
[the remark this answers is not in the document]
g2vinay: What exactly is returned from the service here mean ?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/LongRunningOperationStatus.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * An enum to represent all possible states that a long-running operation may find itself in.
+ * The poll operation is considered complete when the status is one of {@code SUCCESSFULLY_COMPLETED},
+ * {@code USER_CANCELLED} or {@code FAILED}.
+ */
+public final class LongRunningOperationStatus extends ExpandableStringEnum<LongRunningOperationStatus> {
```
g2vinay: If the intention is to not use a generic name like OperationStatus,
Then PollOperationStatus is a good candidate

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
jianghaolu: I believe this forces another poll - there's no backend thread running constantly to poll the latest status.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
jianghaolu: Or `waitUtilStatuses(longRunningOperationStatus... statustoWaitFor)`?

Anyways, I'm joking 😆 But I think there's a limit to how many of these convenience methods to add. Users can always turn to the async API 😄

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
g2vinay: The current implementation of poll operation in KV Poller, 
handles the error and returns FAILED response to poller.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
jianghaolu: park for discussion tomorrow

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
jianghaolu: If the user thinks it's a force REST call, they probably won't wrap this method in a busy loop. On the other hand, if the user thinks this method just checks the status of the last poll (while some thread is polling in the background), they probably won't check the status in a busy loop too. So I'm not sure if your concern maps to a real world scenario.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -0,0 +1,247 @@
...
+ * e.g.: retrieving the final result of long-running operation after it's completion.
+ *
+ *   PollerFlux<BlobCopyStatus, BlobInfo> cpBlobPollerFlux = asyncClient.beginBlobCopy();
+ *
+ *   cpBlobPollerFlux
+ *       .last()
+ *       .flatMap(asyncPollResponse -> {
+ *           // LRO completed, retrieving final result.
+ *           return asyncPollResponse.getFinalResult();
+ *       }).block();
```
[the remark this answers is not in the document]
anuchandy: Yes, it's moved in the latest commit

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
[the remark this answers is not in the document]
g2vinay: PollOperationStatus can work too.

Poll/Polling are possibly more common search terms compared to Long Running Operation

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
```diff
@@ -0,0 +1,122 @@
...
+                return Mono.error(throwable);
+            }
+        });
+    }
+
+    /**
+     * @return a Mono, upon subscription it fetches the final result of long-running operation if it
+     * is supported by the service.
+     */
+    public Mono<U> getFinalResult() {
```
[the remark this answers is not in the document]
anuchandy: this is done

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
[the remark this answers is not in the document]
anuchandy: I managed to kill it and updated the PR

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
anuchandy: done

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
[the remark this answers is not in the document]
anuchandy: I killed this property with the latest update

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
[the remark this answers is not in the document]
anuchandy: guys thoughts on Vinay's proposal?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
anuchandy: busy loop runs only on the rare occurring concurrency cases, like user using the same PollerFlux instance in merge/flatmap operator concurrently. It should be fine to busy loop here, its very common pattern to handle such edge cases.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
anuchandy: We could just keep `waitUtil(longRunningOperationStatus)` and when needed we can make it var-args whiteout breaking any breaking change. This also means, the overload for timeout should take timeout as first arg so that we can turn second param var-args without breaking.

yes, let's discuss this.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/implementation/OperationRequirePollContext.java
```diff
@@ -0,0 +1,19 @@
...
+
+package com.azure.core.util.polling.implementation;
+
+/**
+ * INTERNAL CLASS.
+ *
+ * An exception indicating that an operation on long-running operation
+ * requires a {@link com.azure.core.util.polling.PollingContext}
+ */
+public class OperationRequirePollContext extends RuntimeException {
```
srnagar: `PollContextRequiredException`? Prefer to suffix exceptions with `*Exception`

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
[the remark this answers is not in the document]
anuchandy: actually I removed it, but seems its not reflecting in GitHub review section. If you choose the file `AsyncPollResponse` from "Files Changed" menu then you should see it.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -0,0 +1,276 @@
...
+
+    private Flux<AsyncPollResponse<T, U>> pollingLoop(PollingContext<T> pollingContext) {
+        return Flux.using(
+            // Create a Polling Context per subscription
+            () -> pollingContext,
+            // Do polling
+            // set|read in state as needed, reactor guarantee thread-safety of state object.
+            cxt -> Mono.defer(() -> pollOperation.apply(cxt))
+                    .delaySubscription(getDelay(cxt.getLatestResponse()))
+                    .switchIfEmpty(Mono.error(new IllegalStateException("PollOperation returned Mono.empty().")))
```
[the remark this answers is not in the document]
anuchandy: 1. If "pollOperation" emit an error, which is not allowed per contract, then we log and continue polling.
2. If "pollOperation" returns empty publisher, that also not allowed per contract, then we treat it as an error then follows 1 (i.e. log and continue polling).

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/implementation/OperationRequirePollContext.java
anuchandy: yes, will do

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollingContext.java
```diff
@@ -0,0 +1,122 @@
...
+                    .logExceptionAsError(new IllegalStateException(
+                            "setOnetimeActivationResponse can be called only once."));
+        } else {
+            this.activationResponse = activationResponse;
+            this.latestResponse = this.activationResponse;
+        }
+    }
+
+    @Override
+    public PollingContext<T> clone() {
```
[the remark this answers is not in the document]
anuchandy: Similar to `HttpRequest` type, I'm going to not override `clone()` but add a `copy()` method. We will make it package private so that it's not exposed to "SDK implementer"  from inside various LRO operation handlers.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollingContext.java
```diff
@@ -0,0 +1,122 @@
...
+     * Creates PollingContext.
+     *
+     * @param activationResponse activation poll response holding result of activation operation call.
+     * @param latestResponse latest poll response from pollOperation.
+     * @param map the map to store context
+     */
+    private PollingContext(PollResponse<T> activationResponse,
+                           PollResponse<T> latestResponse,
+                           Map<String, String> map) {
+        this.activationResponse = Objects.requireNonNull(activationResponse,
```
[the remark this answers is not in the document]
anuchandy: With package private `copy()` method chance of NPE is eliminated, if it happens then it will be only from inside `DefaultSyncPoller` or `FluxPoller` that's  considered as a bug.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -0,0 +1,276 @@
...
+
+    private static <T, U> PollResponse<T> toPollResponse(AsyncPollResponse<T, U> asyncPollResponse) {
+        return new PollResponse<>(asyncPollResponse.getStatus(),
+            asyncPollResponse.getValue(),
+            asyncPollResponse.getRetryAfter());
+    }
+
+    private boolean matchStatus(AsyncPollResponse<T, U> currentPollResponse,
+                                LongRunningOperationStatus statusToWaitFor) {
+        if (currentPollResponse == null || statusToWaitFor == null) {
```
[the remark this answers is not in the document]
anuchandy: I would rather be explicit here mainly for readability.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
anuchandy: User shouldn't use tight loop and must honor retry-after to handle any 429.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -0,0 +1,276 @@
...
+            if (response.getStatus().isComplete()) {
+                this.terminalPollContext = context;
+            }
+            return response;
+        }
+    }
+
+    @Override
+    public U getFinalResult() {
+        PollingContext<T> currentTerminalPollContext = this.terminalPollContext;
```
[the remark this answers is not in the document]
anuchandy: sure, we can reuse `waitUntilCompletion`, will do it.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollingContext.java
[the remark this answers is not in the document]
anuchandy: thanks for the thoughts

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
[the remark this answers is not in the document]
anuchandy: sorry actually no, only the `AsyncPollResponse` has ability to retrieve finalResult.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -0,0 +1,276 @@
...
+    public void cancelOperation() {
+        PollingContext<T> context1 = this.pollingContext.clone();
+        if (context1.getActivationResponse() == context1.getLatestResponse()) {
+            this.cancelOperation.apply(context1, context1.getActivationResponse())
+                .block();
+        } else {
+            try {
+                this.cancelOperation.apply(null, this.activationResponse).block();
+            } catch (OperationRequirePollContext crp) {
+                PollingContext<T> context2 = this.pollingContext.clone();
```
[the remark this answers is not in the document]
anuchandy: as we know generally SyncPoller is not thread safe so we're making an attempt to be safe at least in the cases we could. Will add doc.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/SyncPoller.java
anuchandy: Ok, I made Timeout first parameter so when needed we can support var-args without breaking.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+        });
+    }
+
+    /**
+     * Returns the delay the service has requested until the next polling operation is performed. A null or negative
+     * value will be taken to mean that the Poller should determine on its own when the next poll operation is
+     * to occur.
+     * Note: package private
```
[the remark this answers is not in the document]
anuchandy: Deleted comment.

 getProperties is already removed and seems GitHub glitch.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -0,0 +1,276 @@
...
+            PollResponse<T> response = toPollResponse(asyncPollResponse);
+            if (response.getStatus().isComplete()) {
+                this.terminalPollContext = context;
+            }
+            return response;
+        }
+    }
+
+    @Override
+    public U getFinalResult() {
```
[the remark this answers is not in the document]
anuchandy: I like it in general, it's nice to have, but I prefer to hold-on on making more user facing additions, rather wait and add it if needed in future, it wont be a breaking change. But if you really think it's needed  I'm fine with enabling it.

User can achieve the same via:

```java
poller.waitForCompletion(timeout):
poller.getFinalResult();
```

Or can use async Poller for all kind of filtering and timeout cases.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/AsyncPollResponse.java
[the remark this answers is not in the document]
anuchandy: Thinking a bit more - two "operations" here:

1. The long running "Operation"
2. The poll "Operation"

The "poll Operation" returns the status of the "long running operation" running in service side not status of the "poll Operation" itself, so I'm wondering the name "PollOperationStatus" is confusing?

PR #5990, opened 2019-10-22
Add Deserialized Headers to Download Returns

### LINKED ISSUE #5972: [API] Improve downloadWithResponse return type
Currently, it is only possible to get headers by indexing a map with a String value. This also means that any Storage specific headers have to be prefixed with "x-ms-" and all user-defined headers have to be indexed with the "x-ms-meta" prefix. This makes these values very difficult to work with.

We should check specifically for any other APIs that return metadata and ensure these are deserialized into a Map<String, String> that parses out the prefix.

alzimmermsft:
Fixes #5972 

Updates the response types for Blobs and Files download to include deserialized headers.


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobDownloadAsyncResponse.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.function.Function;
+
+/**
+ * This class contains the response information returned from the server when downloading a blob.
+ *
+ * <p>
+ * Failed downloads are automatically retied based on the {@link ReliableDownloadOptions download options}, the retry
```
alzimmermsft: retied -> retried

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobDownloadAsyncResponse.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param headers Raw headers returned in the response.
+     * @param value Stream of download data being returned by the service.
+     * @param deserializedHeaders Headers deserialized into an object.
+     * @param options {@link ReliableDownloadOptions} used to retry downloads is errors occur.
+     * @param info {@link HttpGetterInfo} used to retry a partially failed download from a specific offset.
+     * @param getter Function used to attempt to retry a partially failed download.
+     */
+    public BlobDownloadAsyncResponse(HttpRequest request, int statusCode, HttpHeaders headers, Flux<ByteBuffer> value,
+        BlobDownloadHeaders deserializedHeaders, ReliableDownloadOptions options, HttpGetterInfo info,
+        Function<HttpGetterInfo, Mono<BlobDownloadAsyncResponse>> getter) {
```
alzimmermsft: Can the getter method be made internal to this? This was previously a package private constructor which mean it should only be called by internal code.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobDownloadAsyncResponse.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private Flux<ByteBuffer> applyReliableDownload(Flux<ByteBuffer> data, int currentRetryCount,
+        ReliableDownloadOptions options) {
+        return data.doOnNext(buffer -> {
+            /*
+            Update how much data we have received in case we need to retry and propagate to the user the data we
+            have received.
+             */
+            this.info.setOffset(this.info.getOffset() + buffer.remaining());
```
conniey: It's possible to encounter a race condition here because we are not atomically doing this get and update set. And flatMapMany can be run asynchronously.

rickle-msft: It's my understanding that the contract of reactive streams states that calls to onNext must be strictly sequential; no two threads can enter the method at the same time. If that is the case, there is no danger of race conditions. Running asynchronously does not necessarily imply that anything is happening in parallel.

conniey: I thought having side effects would just wait for the subscriber to do its onNext computation but not necessarily the side effect.

Looks like this is isn't a problem.
https://stackoverflow.com/a/53748683

conniey: My understanding of .doOnNext were for operations that had no consequence like logging. But we are processing the element here, something that map or flatmap is better suited for.
https://stackoverflow.com/a/44372692

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobDownloadAsyncResponse.java
```diff
@@ -0,0 +1,112 @@
...
+        add 1 before calling into tryContinueFlux, we set the initial value to -1.
+         */
+        Flux<ByteBuffer> value = (options.maxRetryRequests() == 0)
+            ? super.getValue()
+            : applyReliableDownload(super.getValue(), -1, options);
+
+        return value.switchIfEmpty(Flux.just(ByteBuffer.wrap(new byte[0])));
+    }
+
+    private Flux<ByteBuffer> tryContinueFlux(Throwable t, int retryCount, ReliableDownloadOptions options) {
```
conniey: Could this be replaced with retryWhen? The companion publisher outputting the offset that we were last at or completion if we are out of retries.

alzimmermsft:
@rickle-msft @jaschrep-msft @gapra-msft Could I get a review on this

PR #6006, opened 2019-10-23
generate overview from readme for javadocs

JimSuplizio:
As part of the versioning story for docs we want the contents of the readme.md for any given artifact to be appended to the landing page. This generation step needs to be done before the javadoc.jar file is generated for a given artifact. 

The way I'd ended up having to deal with converting the readme.md contents to html was through a python script which utilizes markdown2. While the build machines all have python installed I was very cautious to ensure that the dependency on python for the build did not extend to the regular developer by creating a profile, generate-overview, and adding that to the build and package step of our build system. Locally, anyone can run this by just adding -Dgenerate-overview to the command line build. This profile also appends the overview configuration option to the maven-javadoc-plugin. This ensures that any other additional options defined in individual pom files are still honored.

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -900,7 +900,6 @@
             <artifactId>maven-javadoc-plugin</artifactId>
             <version>${maven-javadoc.version}</version>
             <configuration combine.children="append">
-              <additionalparam>-Xdoclint:all -Xdoclint:-missing</additionalparam>
               <overview>${project.basedir}/readme_overview.html</overview>
             </configuration>
           </plugin>
--- a/sdk/core/azure-core-test/README.md
+++ b/sdk/core/azure-core-test/README.md
@@ -39,6 +39,7 @@ Use [TestBase][TestBase.java] to easily create live and playback test cases. Ext
 // Set the AZURE_TEST_MODE environment variable to either PLAYBACK or RECORD to determine if tests are playback or
 // live. By default, tests are run in playback mode.
 public class SessionTests extends TestBase {
+    // Use your JUnit or TestNG annotation here for your testcase
     public void fooTest() {
         // Do some network calls.
     }


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -58,6 +58,7 @@ jobs:
     steps:
       - script: |
           echo "##vso[build.addbuildtag]Scheduled"
+          pip install markdown2
```
weshaggard: This isn't going to run all the time, see the condition below. 

The more I see external steps like this being required the more I think we should have one wrapper script step here  that we run in the pipeline instead of inline in maven, or have a wrapper script that we call maven. Otherwise I don't think people will easily be able to figure out how to successfully run these.

weshaggard: I believe the only thing we need to do in order to accomplish this is enumerate all the readme.md files under `sdk/${{parameters.ServiceDirectory}}`.  That should be pretty easy to do in your python script, once that is done then we just add one script step:
```
- script: |
     pip install markdown2
     python ./eng/pipelines/scripts/generate_overview_from_readme.py --rf sdk/${{parameters.ServiceDirectory}}
```

JimSuplizio: Good catch on the pip install of markdown2, I've moved that to an unconditional step. 

The purpose of the way I'd done the execution of the script is to produce it for artifact being built before the packaging, on the readme.md that exists in that artifact's directory. This would needlessly process other readme.md files are different levels (Cosmos is an example of this). If you feel strongly I can tear out the python exec calls out of the profile but, unfortunately, the -Dgenerate-overview is still going to be there because of conditional appended configuration to add the overview option to the javadoc command. Also, this is better for testing changes locally since it allows producing the overview as part of the build when that option is set instead of having to run a script on the source tree prior.

pom.client.xml
```diff
@@ -861,6 +861,53 @@
         <module>./sdk/template/azure-sdk-template</module>
       </modules>
     </profile>
+
+    <!-- Skip overview generation unless -Dgenerate-overview is passed as an argument.
```
weshaggard: Is there any reason this is in both client and data as opposed to in the shared parent? Not a huge deal but wondering if there was a reason.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
weshaggard: I'm not personally super worried about a little over processing of readme.md files under a given service folder but we can leave this how it is for now. My biggest concern is around injecting the need for python in the middle of the maven execution and having to know you need a particular version of python already installed on the path as well as a version of markdown2 already installed. Whenever there are extra dependencies like that it becomes hard to track down all the things that are needed to be setup in order to work with the tools.

pom.client.xml
JimSuplizio: The primary reason was that I don't want to inadvertently make any changes that'll affect Management builds.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
JimSuplizio: One thing that may mitigate, but not alleviate, your concern is that I ensured my script has no python 3 dependencies and markdown2 is documented to work with 2.6+ or 3.3+ versions. I do agree with the concern about adding things like this into maven execution.

sdk/core/azure-core-test/README.md
```diff
@@ -39,7 +39,7 @@ Use [TestBase][TestBase.java] to easily create live and playback test cases. Ext
 // Set the AZURE_TEST_MODE environment variable to either PLAYBACK or RECORD to determine if tests are playback or
 // live. By default, tests are run in playback mode.
 public class SessionTests extends TestBase {
-    @Test
+    // Use your JUnit or TestNG annotation here for your testcase
```
JimSuplizio: To be very clear why this changed. The README.md files are being rendered into html and added to the landing page for a given artifact. Javadocs does not like or recognize @Test notation and this would actually cause the javadoc build to fail. There is no way around this other than turning off doclint which, frankly, is a pretty bad idea. Also, trying to add a custom tag prevents the failure but javadoc stops processing a given document which means, quite literally, it cuts off adding the html to the landing page when it encounters one.

PR #6013, opened 2019-10-23
Added allowed headers and queries

sima-zhu:


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/StorageAllowedHeadersAndQueries.java
```diff
@@ -0,0 +1,104 @@
+package com.azure.storage.common.implementation;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class StorageAllowedHeadersAndQueries {
+    public static class BlobHeadersAndQueries {
```
JonathanGiles: Why are you introducing static inner classes? There doesn't appear to be any need for this.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/StorageAllowedHeadersAndQueries.java
```diff
@@ -0,0 +1,104 @@
...
+        private static Set<String> blobHeaders = Stream.of(
```
JonathanGiles: static fields should be all upper-case, e.g. BLOB_HEADERS

sdk/storage/azure-storage-file/src/main/java/com/azure/storage/file/BuilderHelper.java
```diff
@@ -85,6 +91,20 @@ static HttpPipeline buildPipeline(Supplier<HttpPipelinePolicy> credentialPolicyS
...
+     * Sets the allowed headers and queries to logOptions.
+     * @param logOptions the log options for headers and queries.
+     */
+    private static void loadLogOptions(final HttpLogOptions logOptions) {
+        Objects.requireNonNull(logOptions);
+
+        logOptions.setAllowedHeaderNames(StorageAllowedHeadersAndQueries.FileShareHeadersAndQueries
+            .getFileShareHeaders());
+
+        logOptions.setAllowedQueryParamNames(StorageAllowedHeadersAndQueries.FileShareHeadersAndQueries
```
JonathanGiles: Are you intending to overwrite the existing white list with your white list, or are you intended to supplement the existing white list with additional headers / query params? At present you're overwriting, which may not be what you're intending to do.

sima-zhu: Good point. I need to append to existing ones.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/StorageAllowedHeadersAndQueries.java
sima-zhu: Try to make the section clear.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/StorageAllowedHeadersAndQueries.java
sima-zhu: Done

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/StorageAllowedHeadersAndQueries.java
sima-zhu: Make the structure flat will be a little hard to read, as there might have more coming for queue-crypto, file-datalake. 
Make them into separate class is a little extra to me. 
Let me know if you have any other opinions.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -190,6 +194,18 @@ private HttpPipeline getHttpPipeline(BlobServiceVersion serviceVersion) {
...
     }
 
+    /**
+     * Sets the allowed headers and queries to logOptions.
+     * @param logOptions the log options for headers and queries.
+     */
+    private static void loadLogOptions(final HttpLogOptions logOptions) {
+        Objects.requireNonNull(logOptions);
+
+        logOptions.setAllowedHeaderNames(StorageAllowedHeadersAndQueries.BlobHeadersAndQueries.getBlobHeaders());
```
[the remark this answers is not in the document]
sima-zhu: It appends to the logOption instead of overwrite now.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/StorageAllowedHeadersAndQueries.java
```diff
@@ -0,0 +1,104 @@
+package com.azure.storage.common.implementation;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class StorageAllowedHeadersAndQueries {
```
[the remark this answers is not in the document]
sima-zhu: Done.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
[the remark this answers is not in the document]
tg-msft: We set our lists in the .ctor before customers have a chance to interact with it.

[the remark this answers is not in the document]
alzimmermsft: @tg-msft Are they immutable or would a customers choice override the default?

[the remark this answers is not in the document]
tg-msft: It's a `List<string>` property so a customer could do any list operation (i.e., add new values, replace existing values, clear the whole thing, etc.).

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/StorageAllowedHeadersAndQueries.java
```diff
@@ -0,0 +1,107 @@
...
+        private static final Set<String> BLOB_HEADERS = Stream.of(
+            "Access-Control-Allow-Origin", "Cache-Control", "Content-Length", "Content-Type", "Date", "Request-Id",
+            "traceparent", "Transfer-Encoding", "User-Agent", "x-ms-client-request-id", "x-ms-date", "x-ms-error-code",
+            "x-ms-request-id", "x-ms-return-client-request-id", "x-ms-version", "Accept-Ranges", "Content-Disposition",
+            "Content-Encoding", "Content-Language", "Content-MD5", "Content-Range", "ETag", "Last-Modified", "Server",
+            "Vary", "x-ms-content-crc64", "x-ms-copy-action", "x-ms-copy-completion-time", "x-ms-copy-id",
+            "x-ms-copy-progress", "x-ms-copy-status", "x-ms-has-immutability-policy", "x-ms-has-legal-hold",
+            "x-ms-lease-state", "x-ms-lease-status", "x-ms-range", "x-ms-request-server-encrypted",
+            "x-ms-server-encrypted", "x-ms-snapshot", "x-ms-source-range", "If-Match", "If-Modified-Since",
+            "If-None-Match", "If-Unmodified-Since", "x-ms-access-tier", "x-ms-access-tier-change-time",
```
rickle-msft: What about metadata headers, anything prefixed with "x-ms-meta" and any other user defined headers? For example, what if they set custom headers to give information to a proxy? It looks like there's no way to log those and they'll always get redacted.

tg-msft: Correct, this is a problem across all languages right now.  Users would have to add specific metadata headers they wanted logged.  I don't think we'll be able to add support for that before the next release, but I'm confident we could figure out a way to do that without breaking in the future.

sima-zhu: @rickle-msft I have changed my implementation which customer is able to remove, add headers on the top of the core and storage list.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobLogOptions.java
```diff
@@ -0,0 +1,163 @@
...
+        .collect(Collectors.toCollection(HashSet::new));
+
+    private Set<String> allowedHeaderNames;
+    private Set<String> allowedQueryParamNames;
+
+    /**
+     * Constructor for combining core and Storage Blob allowed whitelist headers and queries.
+     */
+    public BlobLogOptions() {
+        this.allowedHeaderNames = getHeaderSet();
```
alzimmermsft: Are we losing the default headers and query parameters the base HttpLogOptions is adding? I'm still not certain on having subclasses to handle this.

@JonathanGiles Input on subclasses?

JonathanGiles: I am not a fan of creating subclasses whenever we want to have a different set of default whitelisted headers. I would like to see a better approach than this.

joshfree:
Fixes #6009

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobLogOptions.java
conniey: I'm curious why we needed to subclass this rather than doing something like:
```java
options.addAllowedHeader(… headers);
options.addAllowedQueryParameter(… parameters);
```

Two notes about the HttpLogOptions:
* Do we need to suffix these methods with "Names"? I feel like it's unnecessarily verbose.
* I wish HttpLogOptions.addAllowedHeaderName(), I wish took an Iterable or varargs if we wanted to add more than one header.

sima-zhu: I was thinking to force customer to have the storage headers when do the setter.
Current design is to add a new default getter in builder helper.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -426,6 +429,16 @@ public EncryptedBlobClientBuilder httpLogOptions(HttpLogOptions logOptions) {
         return this;
     }
 
+    /**
+     * Gets the default Storage whitelist log headers and query parameters.
+     */
+    public static HttpLogOptions getBlobDefaultLogOptions(){
```
JonathanGiles: `getDefaultHttpLogOptions`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/SpecializedBlobClientBuilder.java
```diff
@@ -73,7 +74,7 @@
 
     private HttpClient httpClient;
     private final List<HttpPipelinePolicy> additionalPolicies = new ArrayList<>();
-    private HttpLogOptions logOptions = new HttpLogOptions();
+    private HttpLogOptions logOptions = getBlobDefaultLogOptions();
```
JonathanGiles: You should just leave this as null. Create it only in the build method if it is non-null.

sima-zhu: Static will bring race condition, which two threads access the logOption, then one might overwrite the other. Prefer to call the method for every initialization.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
sima-zhu: Done.

PR #6014, opened 2019-10-23
Added ability to not overwrite by default

gapra-msft:
#5881

### DIFF
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
@@ -991,7 +991,7 @@ class BlockBlobAPITest extends APISpec {
         when:
         ParallelTransferOptions parallelTransferOptions = new ParallelTransferOptions()
             .setBlockSize(4).setNumBuffers(4)
-        blobac.upload(null, parallelTransferOptions).block()
+        blobac.upload(null, parallelTransferOptions, true).block()
 
         then:
         thrown(NullPointerException)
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
```diff
@@ -138,8 +138,59 @@
...
+    public Mono<BlockBlobItem> upload(Flux<ByteBuffer> data, ParallelTransferOptions parallelTransferOptions,
+        boolean overwrite) {
+        try {
+            Mono<BlockBlobItem> uploadTask = this.uploadWithResponse(data, parallelTransferOptions, null, null, null,
+                null).flatMap(FluxUtil::toMono);
+
+            if (overwrite) {
+                return uploadTask;
+            } else {
+                return exists()
```
alzimmermsft: Making an exists call to the service then uploading introduces race conditions, instead of using this if/else block we should call into the with response overload passing `new BlobRequestConditions().setIfNoneMatch("*")` for the access conditions. That call will trigger the upload to fail with a 412 - Preconditions not met with the blob already exists. If we want to map the exception returned by the service to one that would be clearer to a customer we should do that with an `onErrorMap` step in the reactive stream.

joshfree: Please open up a bug to track this feedback given the feedback missed the merge;

/cc @amishra-dev as FYI of new storage bug

kurtzeborn: @alzimmermsft opened https://github.com/Azure/azure-sdk-for-java/issues/6017 to track this.

PR #6018, opened 2019-10-24
Improve shared token cache lookup

jianghaolu:
Now consistent with .NET library behavior

### DIFF
--- a/sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
+++ b/sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
 /**
  * A credential provider that provides token credentials from the MSAL shared token cache.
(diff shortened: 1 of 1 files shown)


sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
```diff
@@ -71,33 +73,25 @@
...
+                    if (username == null || username.equals(cached.username())) {
+                        if (!accounts.containsKey(cached.homeAccountId())) {
+                            accounts.put(cached.homeAccountId(), cached);
                         }
                     }
                 }
 
-                if (requestedAccount == null) {
+                if (accounts.size() == 0) {
                     return Mono.error(new RuntimeException("Requested account was not found"));
```
schaabs: Seems like this could also be hit if there are no accounts no?

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
```diff
@@ -71,33 +73,25 @@
...
+                } else if (accounts.size() > 1) {
+                    return Mono.error(new RuntimeException("Multiple entries for the user account " + username
+                        + " were found in the shared token cache. This is not currently supported by the"
+                        + " SharedTokenCacheCredential."));
```
schaabs: Same kind of issue here. If the username wasn't specified this message wouldn't read right.

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
jianghaolu: Added better message

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
jianghaolu: Added better message

schaabs:
The error messages look good.

PR #6030, opened 2019-10-24
Update the default block size for HTBB blobs

rickle-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/ParallelTransferOptions.java
```diff
@@ -5,48 +5,44 @@
...
-    private int numBuffers;
+    private Integer blockSize;
+    private Integer numBuffers;
     private ProgressReceiver progressReceiver;
 
     /**
      * Creates a new {@link ParallelTransferOptions} with default parameters applied.
-     * blockSize = 4MB
-     * numBuffers = 8
      */
```
alzimmermsft: Thoughts on changing this class from a Fluent to a constructor based design? Given its simplicity that might make it easier to use and could remove the need for the apply defaults API.

`public ParallelTransferOptions()` -> Uses base line defaults
`public ParallelTransferOptions(Integer, Integer)` -> Uses what is passed, otherwise defaults (most common scenario)
`public ParallelTransferOptions(Integer, Integer, ProgressReceiver)` -> Same as above but can take a progress reported

rickle-msft: I don't know what the guidelines are on when to use Fluent and when to use constructors, so I'm fine with this if you think it's best

srnagar: Yeah, I think this should be constructor based and immutable. For any operation, the set of options is consistent if this is immutable.

PR #6035, opened 2019-10-24
Issues/consistent name changes

chrischild:
These changes area a result of running the new checkstyle rule from this [PR](https://github.com/Azure/azure-sdk-for-java/pull/5698).

I did not want to pollute the other issue with all these name changes.


JonathanGiles:
Can you add suppression rules to ignore files in `implementation` packages or test packages? We can keep the PR small to start with and focus on public API only.

chrischild:
Sounds good. I'll close this one and re-run it.

PR #6037, opened 2019-10-24
Rename ReliableDownloadOptions and methods

### LINKED ISSUE #6032: ReliableDownloadOptions has incorrect method names
Missing the 'set' / 'get' prefixes here: https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/ReliableDownloadOptions.java

JonathanGiles:
Naming was incorrect for both the class and the methods within it. This PR cleans that up. I'm relying on the CI system to check correctness, as my local build env is outdated and not working - there may be follow-up commits to resolve newly-identified issues, but this already represents the entirety of the changes intended for public API.

Fixes #6032

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/DownloadRetryOptions.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/DownloadRetryOptions.java
@@ -9,7 +9,8 @@ import com.azure.core.util.logging.ClientLogger;
 import java.util.Locale;
 
 /**
- * This class contains the configuration options used to reliably download from the blob service.
+ * This class contains the configuration options used to download from the blob service using smart retries, starting
+ * from a checkpoint, rather than starting over from the beginning.
  */
 @Fluent
 public final class DownloadRetryOptions {


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/ReliableDownload.java
```diff
@@ -21,25 +21,25 @@
  * This class automatically retries failed reads from a blob download stream.
  *
  * <p>
- * Failed downloads are automatically retied based on the {@link ReliableDownloadOptions download options}, the retry
+ * Failed downloads are automatically retried based on the {@link DownloadRetryOptions download options}, the retry
  * will be resumed from the point where the download failed. This allows for the download to be consumed as one
  * continuous stream.
  * </p>
  */
 final class ReliableDownload {
```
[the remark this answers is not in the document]
JonathanGiles: I considered it but I decided to avoid for now, just because it is not public API.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/DownloadRetryOptions.java
```diff
@@ -12,10 +12,10 @@
  * This class contains the configuration options used to reliably download from the blob service.
```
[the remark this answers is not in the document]
JonathanGiles: I don't know the best terminology to use. Is something like this more accurate: "This class contains the configuration options used to download from the blob service using smart retries, starting from a checkpoint, rather than starting over from the beginning."

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/DownloadRetryOptions.java
```diff
@@ -30,7 +30,7 @@
      *
      * @return the maximum number of retries to attempt before the request finally fails
      */
-    public int maxRetryRequests() {
+    public int getMaxRetryRequests() {
```
srnagar: nit: `getMaxRetries` or `getMaxRetryAttempts`?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/DownloadRetryOptions.java
```diff
@@ -12,10 +12,10 @@
  * This class contains the configuration options used to reliably download from the blob service.
  */
 @Fluent
-public final class ReliableDownloadOptions {
+public final class DownloadRetryOptions {
     private static final String PARAMETER_NOT_IN_RANGE = "The value of the parameter '%s' should be between %s and %s.";
```
srnagar: Message can simply be `Max retries cannot be a negative integer.` 
Current message prints out `"The value of the parameter '%s' should be between 0 and 2147483647."`. Not very user-friendly.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/DownloadRetryOptions.java
JonathanGiles: I would leave the naming as-is. It presumably has been standardised across languages.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/DownloadRetryOptions.java
JonathanGiles: I'm happy to make this change if others approve

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/DownloadRetryOptions.java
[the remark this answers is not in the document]
alzimmermsft: I think that is a good change which help explains the use of the class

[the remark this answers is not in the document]
JonathanGiles: I've made the change

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/DownloadRetryOptions.java
JonathanGiles: @srnagar Lets do this in a follow-up issue.

PR #6042, opened 2019-10-24
Improve uploadFromFile Performance

### LINKED ISSUE #6028: [Storage] uploadFromFile() small blobs: throughput regression between v8 and v12
**Describe the bug**
Email thread about perf from Mike

10 KB | 32 | UploadFromFile | 3668 | 875 | 24%
-- | -- | -- | -- | -- | --

**Expected behavior**
atleast within 90% of track 1 since 10kb blobs are majority of the blobs uploaded

This can be a shared investigation between Rick/James and Mike
@joshfree @pemari-msft  fyi

alzimmermsft:
Fixes #6028 

Improves `uploadFromFile` performance for small files by uploading them using Put Blob instead of using Stage Block and Commit Block List. Additionally, updates `SwaggerMethodInterface` to use a static caching mechanism instead of an instance level cache as these cached values don't change between instantiations.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
@@ -24,7 +24,6 @@ import com.azure.storage.common.implementation.Constants;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.ByteBuffer;
@@ -433,11 +432,6 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
         BlobHttpHeaders headers, Map<String, String> metadata, AccessTier tier,
         BlobRequestConditions accessConditions) {
         try {
-            final ParallelTransferOptions finalParallelTransferOptions = parallelTransferOptions == null
-                ? new ParallelTransferOptions(null, null, null)
-                : new ParallelTransferOptions(parallelTransferOptions.getBlockSize(),
-                    parallelTransferOptions.getNumBuffers(), parallelTransferOptions.getProgressReceiver());
-
             return Mono.using(() -> uploadFileResourceSupplier(filePath),
                 channel -> {
                     try {
@@ -446,7 +440,7 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
 
                         // If the file is larger than 256MB chunk it and stage it as blocks.
                         if (fileSize > 256 * Constants.MB) {
-                            return uploadBlocks(fileSize, finalParallelTransferOptions, headers, metadata, tier,
+                            return uploadBlocks(fileSize, parallelTransferOptions, headers, metadata, tier,
                                 accessConditions, channel, blockBlobAsyncClient);
                         } else {
                             // Otherwise we know it can be sent in a single request reducing network overhead.
@@ -466,14 +460,21 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
     private Mono<Void> uploadBlocks(long fileSize, ParallelTransferOptions parallelTransferOptions,
         BlobHttpHeaders headers, Map<String, String> metadata, AccessTier tier,
         BlobRequestConditions accessConditions, AsynchronousFileChannel channel, BlockBlobAsyncClient client) {
-        ProgressReceiver progressReceiver = parallelTransferOptions.getProgressReceiver();
+        final ParallelTransferOptions finalParallelTransferOptions = (parallelTransferOptions == null)
+            ? new ParallelTransferOptions(null, null, null)
+            : new ParallelTransferOptions(parallelTransferOptions.getBlockSize(),
+            parallelTransferOptions.getNumBuffers(), parallelTransferOptions.getProgressReceiver());
+        final BlobRequestConditions finalRequestConditions = (accessConditions == null)
+            ? new BlobRequestConditions()
+            : accessConditions;
+        ProgressReceiver progressReceiver = finalParallelTransferOptions.getProgressReceiver();
 
         // See ProgressReporter for an explanation on why this lock is necessary and why we use AtomicLong.
         AtomicLong totalProgress = new AtomicLong();
         Lock progressLock = new ReentrantLock();
 
         final SortedMap<Long, String> blockIds = new TreeMap<>();
-        return Flux.fromIterable(sliceFile(fileSize, parallelTransferOptions.getBlockSize()))
+        return Flux.fromIterable(sliceFile(fileSize, finalParallelTransferOptions.getBlockSize()))
             .flatMap(chunk -> {
                 String blockId = getBlockID();
                 blockIds.put(chunk.getOffset(), blockId);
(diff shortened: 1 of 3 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -420,51 +417,71 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
-                                channel.close();
-                            } catch (IOException e) {
-                                throw logger.logExceptionAsError(new UncheckedIOException(e));
-                            }
-                        });
+                    try {
+                        long fileSize = channel.size();
+
+                        // If the file is larger than 256MB chunk it and stage it as blocks.
+                        if (fileSize > 256 * Constants.MB) {
```
alzimmermsft: Thoughts on this being 256MB exactly or should we back off to a lower amount such as 128 or 196? In the performance tests we were running upload vs staging blocks overhead becomes virtually non-existent after 10MB or so, this would be a cost and reliability optimization by requiring less service calls at this point.

rickle-msft: I'm not sure it matters a ton right now, as we are ultimately going to make this configurable in the near future. Customers were pretty quick to ask for this in v11, so we'll have to do it here as well. I think 256 is a good default because that's the service limit and anything else is just a guess.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -420,51 +417,71 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
+                new ArrayList<>(blockIds.values()), headers, metadata, tier, accessConditions)))
+            .then()
+            .doOnTerminate(() -> uploadFileCleanup(channel));
+    }
+
+    private Mono<Void> uploadFullBlob(BlobHttpHeaders headers, Map<String, String> metadata, AccessTier tier,
+        BlobRequestConditions accessConditions, AsynchronousFileChannel channel, long fileSize) {
+        return getBlockBlobAsyncClient().uploadWithResponse(FluxUtil.readFile(channel),
+            fileSize, headers, metadata, tier, accessConditions)
+            .then();
```
rickle-msft: I think you need uploadFileCleanup here, too.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: I'll leave it as is for now

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: This is being called from within the `Mono.using` and the `resourceCleanup` consumer is the equivalent of a final so this shouldn't need to do a cleanup here.

rickle-msft: In that case, is the doOnTerminate on line 475 in uploadBlocks unnecessary?

alzimmermsft: I don't think it is needed either.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -420,51 +417,66 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
+                            return uploadBlocks(fileSize, finalParallelTransferOptions, headers, metadata, tier,
+                                accessConditions, channel, blockBlobAsyncClient);
+                        } else {
+                            // Otherwise we know it can be sent in a single request reducing network overhead.
+                            return blockBlobAsyncClient.uploadWithResponse(FluxUtil.readFile(channel), fileSize,
+                                headers, metadata, tier, accessConditions)
+                                .then();
+                        }
+                    } catch (IOException ex) {
+                        return Mono.error(new UncheckedIOException(ex));
```
srnagar: Why is the `IOException` wrapped with `UncheckedIOException`? Can this just be `Mono.error(ex)`?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: Removed

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -420,51 +417,66 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
+    private Mono<Void> uploadBlocks(long fileSize, ParallelTransferOptions parallelTransferOptions,
+        BlobHttpHeaders headers, Map<String, String> metadata, AccessTier tier,
+        BlobRequestConditions accessConditions, AsynchronousFileChannel channel, BlockBlobAsyncClient client) {
+        ProgressReceiver progressReceiver = parallelTransferOptions.getProgressReceiver();
+
+        // See ProgressReporter for an explanation on why this lock is necessary and why we use AtomicLong.
+        AtomicLong totalProgress = new AtomicLong();
+        Lock progressLock = new ReentrantLock();
+
+        final SortedMap<Long, String> blockIds = new TreeMap<>();
```
srnagar: Does this have to be sorted? It's used to commit blocks. Is order important for committing blocks?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerInterfaceParser.java
```diff
@@ -20,7 +20,7 @@
 public class SwaggerInterfaceParser {
     private final String host;
     private final String serviceName;
-    private final Map<Method, SwaggerMethodParser> methodParsers = new HashMap<>();
+    private static final Map<Method, SwaggerMethodParser> METHOD_PARSERS = new ConcurrentHashMap<>();
```
srnagar: Is this still necessary to have after making the change to `getBlockBlobAsyncClient()`?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: Yes, the order of the block IDs determines how the blob is created, we want to make sure this is in the correct order.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/SwaggerInterfaceParser.java
srnagar: Let's keep it as it's going to improve performance.

alzimmermsft: Yup, that call is what triggers this code path, we may have reduced the number of calls but this was the hot spot.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: Changed

PR #6046, opened 2019-10-25
SSL Dependency update

g2vinay:
This dependency is expected to provide performance boost of 35-40% in RPS.
Running Storage Tests to verify and provide results.

### DIFF
--- a/sdk/keyvault/azure-security-keyvault-secrets/README.md
+++ b/sdk/keyvault/azure-security-keyvault-secrets/README.md
@@ -70,25 +70,7 @@ HttpClient client = new NettyAsyncHttpClientBuilder()
 ```
 
 ### Default SSL library
-All client libraries, by default, use Tomcat native Boring SSL. The Boring SSL library is an uber jar containing bits of linux/mac/windows and provides better performance compared to default jdk SSL.
-If you prefer to use default jdk SSL then exclude the Boring SSL in your pom.xml.
-
-[//]: # ({x-version-update-start;com.azure:azure-security-keyvault-secrets;current})
-```xml
-<!-- Add KeyVault Secrets dependency without Tomcat Native Boring SSL -->
-<dependency>
-    <groupId>com.azure</groupId>
-    <artifactId>azure-security-keyvault-secrets</artifactId>
-    <version>4.0.0</version>
-    <exclusions>
-      <exclusion>
-        <groupId>io.netty</groupId>
-        <artifactId>netty-tcnative-boringssl-static</artifactId>
-      </exclusion>
-    </exclusions>
-</dependency>
-```
-[//]: # ({x-version-update-end})
+All client libraries, by default, use the Tomcat-native Boring SSL library to enable native-level performance for SSL operations. The Boring SSL library is an uber jar containing native libraries for Linux / macOS / Windows, and provides better performance compared to the default SSL implementation within the JDK. For more information, including how to reduce the dependency size, refer to the [performance tuning][performance_tuning] section of the wiki.
 
 ### Prerequisites
 
@@ -378,5 +360,6 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 [azure_keyvault_cli_full]:https://docs.microsoft.com/cli/azure/keyvault?view=azure-cli-latest
 [secrets_samples]: src/samples/java/com/azure/security/keyvault/secrets
 [samples]: src/samples/README.md
+[performance_tuning]: https://github.com/Azure/azure-sdk-for-java/wiki/Performance-Tuning
 
 ![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-java%2Fsdk%2Fkeyvault%2Fazure-security-keyvault-secrets%2FREADME.png)
--- a/sdk/appconfiguration/azure-data-appconfiguration/README.md
+++ b/sdk/appconfiguration/azure-data-appconfiguration/README.md
@@ -84,6 +84,9 @@ HttpClient client = new NettyAsyncHttpClientBuilder()
     .build();
 ```
 
+### Default SSL library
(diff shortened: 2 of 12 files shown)


joshfree:
This is a great improvement and something that I would like to ship at some point very soon.  However we're hours away from locking down for the November SDK Update.  I think this change could use some bake time in master, and I would suggest that we merge this after November SDK Update is cut from master; so that it can bake and ship in the December SDK Update release.

/cc @Petermarcu as FYI

anuchandy:
nice @g2vinay! it might be good to mentioned in readme( or equivalent) about this dependency that azure-core is pulling transitively for the user application, so that users concerned about size can exclude it and fallback to JDK ssl.

anuchandy:
The tc-native seems a uber jar containing bits for linux/Mac/windows: https://repo1.maven.org/maven2/io/netty/netty-tcnative/2.0.25.Final/

We might want to think about applying `classifier` so that only the bits based on target OS will be pulled. Please see netty doc: https://netty.io/wiki/forked-tomcat-native.html#how-to-download-netty-tcnative-boringssl-static

sdk/core/azure-core/pom.xml
```diff
@@ -88,6 +88,11 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative-boringssl-static</artifactId>
```
srnagar: Is this verified to work with OkHttp too? I know OkHttp supports boringssl but we should verify if the versions are compatible.

JonathanGiles:
I agree with @anuchandy - please investigate the document he links to and explore using the operating system classifier so that only the bits for a single OS are used. Weigh up the pros and cons of this approach (particularly in terms of ease of deployment for end users), and let us know your thoughts.

JonathanGiles:
@g2vinay Can you give us a status update on this when you have further progress? Thanks!

g2vinay:
@JonathanGiles , I'll push updates soon.

g2vinay:
Storage Perf Numbers   (with this PR   vs GA version)
Windows VM 64bit, 2.30 GHz (WIndows 10 Pro)
Ram 7.00 GB

Processor: 
Storage Version         Parallel            API            FileSize             RPS
12.0.0                         32              download          1MB              42.3
12.1.0-beta.1             32               download          1MB             108.2     (155.79% increase from GA ver)

12.0.0                         32            uploadBlockBlob 1MB              50.7
12.1.0-beta.1             32             uploadBlockBlob 1MB             149.2     (194.28% increase from GA ver)

@JonathanGiles @joshfree @anuchandy 

@anuchandy , as disucssed we will not be using classifier due to reliability issues with maven os plugin, we wouldn't want our users to deal with.

sdk/core/azure-core/pom.xml
g2vinay: The versions are compatible,
but with okhttp one more dependency is needed and requires the user to manually setup the Security provider in the code.
Best to handle this in a future PR when okhttp performance becomes important.
and currently okhttp has some issues like resource leaks and exceptions thrown with storage SDK, Need to get it stable before focusing on its perf.
@JonathanGiles

parent/pom.xml
```diff
@@ -194,6 +195,7 @@
     <maven-eclipse-plugin.version>2.8</maven-eclipse-plugin.version>
     <maven-assembly-plugin.version>2.2</maven-assembly-plugin.version>
     <maven-surefire-plugin.version>3.0.0-M3</maven-surefire-plugin.version>
+    <os-maven-plugin.version>1.4.0.Final</os-maven-plugin.version>
```
g2vinay: remove this

sdk/core/azure-core/pom.xml
```diff
@@ -96,6 +100,10 @@
       <artifactId>junit-jupiter-engine</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
```
g2vinay: remove this, not intended

sdk/keyvault/azure-security-keyvault-secrets/README.md
```diff
@@ -69,6 +69,28 @@ HttpClient client = new NettyAsyncHttpClientBuilder()
...
+    <exclusions>
+      <exclusion>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-tcnative-boringssl-static</artifactId>
+      </exclusion>
+    </exclusions>
+</dependency>
+```
+[//]: # ({x-version-update-end})
+
```
g2vinay: @JonathanGiles @anuchandy , @srnagar 
Can you review this piece, 
Let me know if I should add something more, 
Will then add it to all track 2 SDKs

sdk/core/azure-core/pom.xml
JonathanGiles: That is fine - just file a tech debt issue to track enabling SSL for OkHttp

sdk/keyvault/azure-security-keyvault-secrets/README.md
```diff
@@ -69,6 +69,27 @@ HttpClient client = new NettyAsyncHttpClientBuilder()
     .build();
 ```
 
+### Default SSL library
+All client libraries, by default, use Tomcat native Boring SSL. The Boring SSL library is an uber jar containing bits of linux/mac/windows and provides better performance compared to default jdk SSL.
```
JonathanGiles:
```suggestion
All client libraries, by default, use the Tomcat-native Boring SSL library to enable native-level performance for SSL operations. The Boring SSL library is an uber jar containing native libraries for Linux / macOS / Windows, and provides better performance compared to the default SSL implementation within the JDK. For more information, including how to reduce the dependency size, refer to the [performance tuning](https://github.com/Azure/azure-sdk-for-java/wiki/Performance-Tuning) section of the wiki.
```

sdk/keyvault/azure-security-keyvault-secrets/README.md
JonathanGiles: With my text changes above, all of this section should be deleted.

g2vinay: done

sdk/core/azure-core/pom.xml
g2vinay: Opened the issue - #6266

PR #6060, opened 2019-10-28
Prefetch Upload To Determine Code Path

### LINKED ISSUE #6033: [Perf] BlobAsyncClient.upload() small blobs: throughput regression between v8 and v12
There is a significant throughput regression between v8 sync and v12 async when using the "primary" upload API and small (10 KB) blobs (with `parallel=32`):

```
v8 sync:   5285 uploads/second
v12 async:  462 uploads/second
```

## APIs
```
v8:  CloudBlockBlob.upload(InputStream, length)
v12: BlobAsyncClient.upload(Flux<ByteBuffer>, ParallelTransferOptions)
```

This is likely related to #6028, since both upload methods use similar code.

alzimmermsft:
Fixes #6033 

Adds functionality for `BlobAsyncClient`'s upload method to determine whether it should use Put Blob or Stage Blocks and Put Block List. Logic is performed by pre-fetching `ByteBuffer`s from the reactive stream and if there is greater than 4MB of data it will perform staging of blocks otherwise a put blob operation is performed.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
@@ -71,7 +71,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
  * Docs</a> for more information.
  */
 public class BlobAsyncClient extends BlobAsyncClientBase {
-    private static final int CHUNKED_UPLOAD_REQUIREMENT = 4* Constants.MB;
+    private static final int CHUNKED_UPLOAD_REQUIREMENT = 4 * Constants.MB;
 
     public static final int BLOB_DEFAULT_UPLOAD_BLOCK_SIZE = 4 * Constants.MB;
     public static final int BLOB_DEFAULT_NUMBER_OF_BUFFERS = 8;
@@ -292,19 +292,25 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
         AccessTier tier, BlobRequestConditions accessConditions) {
         try {
             Objects.requireNonNull(data, "'data' must not be null");
-            BlobRequestConditions accessConditionsFinal = accessConditions == null
+            BlobRequestConditions validatedAccessConditions = accessConditions == null
                 ? new BlobRequestConditions() : accessConditions;
-            final ParallelTransferOptions finalParallelTransferOptions = parallelTransferOptions == null
+            final ParallelTransferOptions validatedParallelTransferOptions = parallelTransferOptions == null
                 ? new ParallelTransferOptions(null, null, null)
                 : new ParallelTransferOptions(parallelTransferOptions.getBlockSize(),
                 parallelTransferOptions.getNumBuffers(), parallelTransferOptions.getProgressReceiver());
 
             BlockBlobAsyncClient blockBlobAsyncClient = getBlockBlobAsyncClient();
-            return determineUploadFullOrChunked(data, (stream) -> uploadInChunks(blockBlobAsyncClient, stream,
-                finalParallelTransferOptions, headers, metadata, tier, accessConditionsFinal),
-                (stream, length) -> blockBlobAsyncClient.uploadWithResponse(
-                    ProgressReporter.addProgressReporting(stream, finalParallelTransferOptions.getProgressReceiver()),
-                    length, headers, metadata, tier, null, accessConditionsFinal));
+
+            Function<Flux<ByteBuffer>, Mono<Response<BlockBlobItem>>> uploadInChunksFunction = (stream) ->
+                uploadInChunks(blockBlobAsyncClient, stream, validatedParallelTransferOptions, headers, metadata, tier,
+                    validatedAccessConditions);
+
+            BiFunction<Flux<ByteBuffer>, Long, Mono<Response<BlockBlobItem>>> uploadFullBlobMethod =
+                (stream, length) -> blockBlobAsyncClient.uploadWithResponse(ProgressReporter
+                    .addProgressReporting(stream, validatedParallelTransferOptions.getProgressReceiver()),
+                    length, headers, metadata, tier, null, validatedAccessConditions);
+
+            return determineUploadFullOrChunked(data, uploadInChunksFunction, uploadFullBlobMethod);
         } catch (RuntimeException ex) {
             return monoError(logger, ex);
         }


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -359,6 +368,27 @@ as we can guarantee we only need at most two buffers for any call to write (two
         }
     }
 
+    private Mono<Response<BlockBlobItem>> determineUploadFullOrChunked(final Flux<ByteBuffer> data,
+        final Function<Flux<ByteBuffer>, Mono<Response<BlockBlobItem>>> path1,
+        final BiFunction<Flux<ByteBuffer>, Long, Mono<Response<BlockBlobItem>>> path2) {
+        final int[] predicateCount = { 4 * Constants.MB };
```
srnagar: Should this use the blockSize specified in `ParallelTransferOptions` instead of hardcoding this value?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -284,6 +284,15 @@ private SpecializedBlobClientBuilder prepareBuilder() {
      * @return A reactive response containing the information of the uploaded block blob.
      */
     public Mono<Response<BlockBlobItem>> uploadWithResponse(Flux<ByteBuffer> data,
+        ParallelTransferOptions parallelTransferOptions, BlobHttpHeaders headers, Map<String, String> metadata,
+        AccessTier tier, BlobRequestConditions accessConditions) {
+        return determineUploadFullOrChunked(data, (stream) ->
```
srnagar: I would prefer to have the try/catch here in the public api instead of in a private method that gets called from here. We may cause some exceptions here that are thrown directly before calling the private method.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -295,10 +304,10 @@ private SpecializedBlobClientBuilder prepareBuilder() {
             final ParallelTransferOptions finalParallelTransferOptions = parallelTransferOptions == null
                 ? new ParallelTransferOptions(null, null, null)
                 : new ParallelTransferOptions(parallelTransferOptions.getBlockSize(),
-                    parallelTransferOptions.getNumBuffers(), parallelTransferOptions.getProgressReceiver());
+                parallelTransferOptions.getNumBuffers(), parallelTransferOptions.getProgressReceiver());
 
             // See ProgressReporter for an explanation on why this lock is necessary and why we use AtomicLong.
-            AtomicLong totalProgress = new AtomicLong(0);
+            AtomicLong totalProgress = new AtomicLong();
```
srnagar: L354 and L364 (path 1) call  `getBlockBlobAsyncClient()` twice which is a bit of an expensive operation. Do the `get` once and use it in path 1 and path 2.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -359,6 +368,27 @@ as we can guarantee we only need at most two buffers for any call to write (two
         }
     }
 
+    private Mono<Response<BlockBlobItem>> determineUploadFullOrChunked(final Flux<ByteBuffer> data,
+        final Function<Flux<ByteBuffer>, Mono<Response<BlockBlobItem>>> path1,
```
srnagar: nit: naming `path1` -> `stageAndUpload`, `path2` -> `upload` or something similar

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -359,6 +368,27 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
+            .bufferUntil(buffer -> {
+                predicateCount[0] -= buffer.remaining();
+                return predicateCount[0] < 0;
+            }, true)
+            .next()
+            .flatMap(bufferedData -> usePath1[0]
+                ? path1.apply(data)
+                : path2.apply(Flux.fromIterable(bufferedData), bufferedData.stream()
+                .map(buffer -> (long) buffer.remaining())
+                .reduce(0L, Long::sum)));
```
anuchandy: This is cool @alzimmermsft. :) .

I small question 

Do we still need this code?

```java
bufferedData.stream()
     .map(buffer -> (long) buffer.remaining())
     .reduce(0L, Long::sum)
```

if we do `4 * Constants.MB + predicateCount[0]` then that should give us total size of bytes in `bufferedData` right?

alzimmermsft: Correct, could just use `4 * Constants.MB - predicateCount[0]` to get the difference between the 4MB check and what data has been streamed through the Flux.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: Changed to `uploadInChunks` and `uploadFullBlob`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
anuchandy: actually we're skipping the last ByteBuffer from buffering but predicateCount[0] includes the last one as well so `4 * Constants.MB + predicateCount[0]` won't work always.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: I think that is a reasonable change, basically have the logic of if we cannot fill an entire buffer just send it as a single upload.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -359,6 +368,27 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
+        return data
+            .filter(ByteBuffer::hasRemaining)
+            .doOnEach(signal -> usePath1[0] = !signal.isOnComplete())
+            .bufferUntil(buffer -> {
+                predicateCount[0] -= buffer.remaining();
+                return predicateCount[0] < 0;
+            }, true)
+            .next()
+            .flatMap(bufferedData -> usePath1[0]
+                ? path1.apply(data)
```
mikeharder: I'm not an expert with Reactor, but I don't know if it's valid to pass `data` to `path1` after already reading some of the buffers.  This will work with some Fluxes (like `Flux.just()`), but it may not work with an arbitrary Flux which I believe is free to discard buffers after they have been read.

For example, I believe the default mode for a `Flux` received from `reactor-netty` will recycle memory as soon as a buffer has been published.  You would need to call `retain()` to change this behavior.

https://projectreactor.io/docs/netty/release/api/reactor/netty/ByteBufFlux.html#retain--

anuchandy: In our SDK, the contract in general when user provide `Flux<ByteBuffer>` is - the provided Flux has to be replayable means multiple subscriptions can be done.

Reactor Netty's design is based on pooled ByteBuf (ref counted) so as soon as each chunk is consumed and if ref count is zero, the chunk is returned to the pool.

mikeharder: This seems problematic to me for a few reasons:

1. To upload a large file (say 1 GB) using `upload(Flux<ByteBuffer>)`, the entire content needs to be stored in memory until the upload is complete.
2. If the provided `Flux<ByteBuffer>` is not replayable, the failure mode could be really bad (e.g. uploading the wrong content).

anuchandy:
@alzimmermsft @srnagar @mikeharder thoughts on something like below to solve this?


```java
public  Mono<Void> upload(Flux<ByteBuffer> source /*stream to upload*/) {
    final int[] groupingPredicate = { 4 * Constants.MB };
    //
    // Use groupBy operator to split the source flux into two sub-fluxes
    // "first-group" flux and "second-group" flux
    Flux<GroupedFlux<Integer, ByteBuffer>> splitFlux = source
            .groupBy(buffer -> {
                if (groupingPredicate[0] < 0) {
                    return 2;
                } else {
                    groupingPredicate[0] -= buffer.remaining();
                    if (groupingPredicate[0] >= 0) {
                        return 1;
                    } else {
                        return 2;
                    }
                }
            });

    return splitFlux
            .next()
            .flatMap((Function<GroupedFlux<Integer, ByteBuffer>, Mono<Void>>) groupedFlux -> {
                if (groupedFlux.key() == 2) {
                    // There is no "first-group" (because the first chunk itself greater than 4MB)
                    return chunkedUpload(groupedFlux);
                } else {
                    // "groupedFlux" is the "first-group".
                    return splitFlux
                            .next() // try get "second-group"
                            .flatMap(new Function<GroupedFlux<Integer, ByteBuffer>, Mono<Void>>() {
                                @Override
                                public Mono<Void> apply(GroupedFlux<Integer, ByteBuffer> secondGroup) {
                                    // there is "second-group", concat it with "first-group" and do chunked upload.
                                    return chunkedUpload(groupedFlux.concatWith(secondGroup));
                                }
                            })
                            .switchIfEmpty(Mono.defer(new Supplier<Mono<Void>>() {
                                // There is no "second-group" hence do single upload.
                                @Override
                                public Mono<Void> get() {
                                    return singleUpload(groupedFlux);
                                }
                            }));

                }
            });

}


private Mono<Void> singleUpload(Flux<ByteBuffer> toUpload) {
    // Make API call for single upload
    return Mono.empty();
}

private Mono<Void> chunkedUpload(Flux<ByteBuffer> toUpload) {
    // Do the chunking and upload via multiple calls
    return Mono.empty();
}
```

anuchandy:
Ok, the previous one has a bug, below one seems working as expected. @alzimmermsft would you mind validating this?

```java
public  Mono<Void> upload(Flux<ByteBuffer> source /*stream to upload*/) {
    final int MAX_SINGLE_UPLOAD_SIZE = 4 * 1024 * 1024;
    final int[] group1MaxSize = { 0 };
    final boolean[] group1Exhausted =  { false };
    //
    Flux<GroupedFlux<Integer, ByteBuffer>> splitFlux = source
            .groupBy(buffer -> {
                if (group1Exhausted[0]) {
                    return 2;
                } else {
                    group1Exhausted[0] = (group1MaxSize[0] + buffer.remaining()) > MAX_SINGLE_UPLOAD_SIZE;
                    if (group1Exhausted[0]) {
                        return 2;
                    } else {
                        group1MaxSize[0] += buffer.remaining();
                        return 1;
                    }
                }
            }).sort(Comparator.comparingInt(GroupedFlux::key));

    return splitFlux.collect(() -> new ArrayList<GroupedFlux<Integer, ByteBuffer>>(),
            (accumulator, groupedFlux) -> {
                accumulator.add(groupedFlux);
            }).flatMap(list -> {
                if (list.size() == 0) {
                    return Mono.empty();
                } else if (list.size() == 1) {
                    GroupedFlux<Integer, ByteBuffer> groupedFlux = list.get(0);
                    if (groupedFlux.key() == 1) {
                        return singleUpload(groupedFlux, group1MaxSize[0]);
                    } else {
                        return chunkedUpload(groupedFlux);
                    }
                } else {
                    GroupedFlux<Integer, ByteBuffer> groupedFlux1 = list.get(0);
                    GroupedFlux<Integer, ByteBuffer> groupedFlux2 = list.get(1);
                    return chunkedUpload(groupedFlux1.concatWith(groupedFlux2));
                }
            });
}


private Mono<Void> singleUpload(Flux<ByteBuffer> toUpload, int size) {
    System.out.println("Uploading as single blob, size:" + size);
    // Make API call for single upload
    return Mono.empty();
}

private Mono<Void> chunkedUpload(Flux<ByteBuffer> toUpload) {
    System.out.println("Uploading as chunked blob");
    // Do the chunking and upload via multiple calls
    return Mono.empty();
}
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: @rickle-msft Thoughts on this change, maybe hold off on it until the possible extension of `ParallelTransferOptions` to include a property that indicates limit to trigger Put Blob vs Stage Blocks and put Block List.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
rickle-msft: Just to clarify a couple points here: Anything passed to generated code has to be replayable. The method under discussion, though, is for the express purpose of uploading non-replayable Fluxes (also for Fluxes of unknown length, but not as important here).

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -284,6 +285,15 @@ private SpecializedBlobClientBuilder prepareBuilder() {
      * @return A reactive response containing the information of the uploaded block blob.
      */
     public Mono<Response<BlockBlobItem>> uploadWithResponse(Flux<ByteBuffer> data,
+        ParallelTransferOptions parallelTransferOptions, BlobHttpHeaders headers, Map<String, String> metadata,
+        AccessTier tier, BlobRequestConditions accessConditions) {
+        return determineUploadFullOrChunked(data, (stream) ->
+                uploadInChunks(stream, parallelTransferOptions, headers, metadata, tier, accessConditions),
+            (stream, length) -> getBlockBlobAsyncClient()
+                .uploadWithResponse(stream, length, headers, metadata, tier, accessConditions));
```
rickle-msft: We should pass ProgressReporter.addProgressReporting(stream) into upload()

mikeharder:
I'm uncertain if the latest code will hold all the `ByteBuffer` in memory during the upload (which we don't want).  Testing now.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
rickle-msft: I'm ok on holding off on this because we actually want this feature for uploadFromFile as well. We can do both at once.

mikeharder:
I've identified two problems with the current code:

1. It seems to have a functional bug in the `concatWith` codepath, where the second flux cannot be enumerated.
2. It does load all the `ByteBuffer` into memory before the upload starts which is not acceptable.

We might be able to fix both issues by using `windowUntil()` instead of `groupBy()`.

anuchandy:
Here is an approach using `GroupBy` but without loading the entire source into memory.

```java
public enum GroupState {
    NONE,
    WAIT_FOR_NEXT_GROUP,
    COMPLETED
}

public  Flux<Void> upload(Flux<ByteBuffer> source /*stream to upload*/) {
    final int MAX_SINGLE_UPLOAD_SIZE = 4 * 1024 * 1024;
    final int[] bufferedSize = { 0 };
    final boolean[] maxBuffered =  { false };
    //
    Flux<GroupedFlux<Integer, ByteBuffer>> splitFlux = source
            .groupBy(buffer -> {
                if (maxBuffered[0]) {
                    return 2;
                } else {
                    maxBuffered[0] = (bufferedSize[0] + buffer.remaining()) > MAX_SINGLE_UPLOAD_SIZE;
                    if (maxBuffered[0]) {
                        return 2;
                    } else {
                        bufferedSize[0] += buffer.remaining();
                        return 1;
                    }
                }
            });
    //
    GroupState[] state = new GroupState[1];
    state[0] = GroupState.NONE;
    //
    GroupedFlux<Integer, ByteBuffer>[] savedGroup = new GroupedFlux[1];
    //
    return splitFlux.flatMap((Function<GroupedFlux<Integer, ByteBuffer>, Flux<Void>>) group -> {
        if (state[0] == GroupState.NONE) {
            if (group.key() == 2) {
                state[0] = GroupState.COMPLETED;
                return chunkedUpload(group);
            } else {
                state[0] = GroupState.WAIT_FOR_NEXT_GROUP;
                savedGroup[0] = group;
                return Flux.empty();
            }
        } else if(state[0] == GroupState.WAIT_FOR_NEXT_GROUP) {
            state[0] = GroupState.COMPLETED;
            return chunkedUpload(savedGroup[0].concatWith(group));
        } else {
            return Flux.empty();
        }
    }).switchIfEmpty(Flux.defer(() -> {
        if (state[0] == GroupState.WAIT_FOR_NEXT_GROUP) {
            return singleUpload(savedGroup[0], bufferedSize[0]);
        } else {
            return Flux.empty();
        }
    }));
}

private Flux<Void> chunkedUpload(Flux<ByteBuffer> toUpload) {
    System.out.println("Uploading as chunked blob");
    return toUpload.map(b -> {
        System.out.println("writing [chunked]...");
        return b;
    }).thenMany(Flux.empty());
}

private Flux<Void> singleUpload(Flux<ByteBuffer> toUpload, int size) {
    System.out.println("Uploading as single blob, size:" + size);
    return toUpload.map(b -> {
        System.out.println("writing [single]...");
        return b;
    }).thenMany(Flux.empty());
}
```

mikeharder:
I think this is both functionally correct and it seems to only buffer as much memory as necessary (the first 4 MB plus the next `ByteBuffer`):

```java
private static Mono<String> UploadWindowBy(Flux<ByteBuffer> data) {
    final long chunkedUploadRequirement = 4 * 1024 * 1024;
    final long[] bufferedDataSize = {0};

    return data
            .windowUntil(buffer -> {
                if (bufferedDataSize[0] > chunkedUploadRequirement) {
                    return false;
                }
                else {
                    bufferedDataSize[0] += buffer.remaining();
                    return bufferedDataSize[0] > chunkedUploadRequirement;
                }
            }, true)
            .buffer(2)
            .next()
            .flatMap(fluxes -> {
                if (fluxes.size() == 1) {
                    return UploadSmall(fluxes.get(0), bufferedDataSize[0]);
                }
                else {
                    return UploadLarge(fluxes.get(0).concatWith(fluxes.get(1)));
                }
            });
}
```

mikeharder:
We may need one more fix.  We believe some `Flux<ByteBuffer>` might use pooled buffers, such that after you have consumed buffer with index `N`, the buffer with index `N - 1` is no longer valid.  In this case, we may need to copy the contents of the `ByteBuffer` up to 4 MB, rather than just keeping pointers to these `ByteBuffer`.  I am testing this now, and it should be a simple fix.

anuchandy:
Both `WindowUntil|GroupBy` solves the use case, `WindowUntil` is more clean. In both cases we have to do a deep copy of `ByteBuffer`s in first `window|group` the reason is upstream can invalidate a `ByteBuffer` as soon as it receives request for the next `ByteBuffer.`

mikeharder:
This performs a deep copy on the `ByteBuffer` in the first 4MB window.  There may be a better way to deep copy, this was just the first way I found.

It seems to be functionally correct, but it has a slight inefficiency in that the 4MB cannot be freed until the whole upload is done.  We should be able to fix this by using a slightly different data structure for the cached `ByteBuffers`.

```java
    private static Mono<String> UploadWindowBy(Flux<ByteBuffer> data) {
        final long chunkedUploadRequirement = 4 * 1024 * 1024;
        final long[] bufferedDataSize = {0};
        final List<ByteBuffer> cache = new ArrayList<ByteBuffer>();

        return data
                .windowUntil(buffer -> {
                    if (bufferedDataSize[0] <= chunkedUploadRequirement) {
                        bufferedDataSize[0] += buffer.remaining();

                        if (bufferedDataSize[0] <= chunkedUploadRequirement) {
                            cache.add(deepCopy(buffer));
                            return false;
                        }
                        else {
                            return true;
                        }
                    }
                    else {
                        return false;
                    }
                }, true)
                .buffer(2)
                .next()
                .flatMap(fluxes -> {
                    if (fluxes.size() == 1) {
                        return UploadSmall(Flux.fromIterable(cache), bufferedDataSize[0]);
                    }
                    else {
                        return UploadLarge(Flux.fromIterable(cache).concatWith(fluxes.get(1)));
                    }
                });
    }

    private static ByteBuffer deepCopy(ByteBuffer b)
    {
        return ByteBuffer.allocate(b.remaining()).put(b).position(0);
    }
```

anuchandy:
To ensure each subscription to the windowed Mono starts with a clean state for the local variables, It's good practice to put the entire Windowed Mono creation inside a `Mono.defer`, including `bufferredDataSize` and `cache list`. 

```java
    return Mono.defer(() -> {
        final long chunkedUploadRequirement = 4 * 1024 * 1024;
        final long[] bufferedDataSize = {0};
        final List<ByteBuffer> cache = new ArrayList<ByteBuffer>();
        //
        return data
                .windowUntil(buffer -> {
                    if (bufferedDataSize[0] <= chunkedUploadRequirement) {
                        bufferedDataSize[0] += buffer.remaining();

                        if (bufferedDataSize[0] <= chunkedUploadRequirement) {
                            cache.add(deepCopy(buffer));
                            return false;
                        }
                        else {
                            return true;
                        }
                    }
                    else {
                        return false;
                    }
                }, true)
                .buffer(2)
                .next()
                .flatMap(fluxes -> {
                    if (fluxes.size() == 1) {
                        return UploadSmall(Flux.fromIterable(cache), bufferedDataSize[0]);
                    }
                    else {
                        return UploadLarge(Flux.fromIterable(cache).concatWith(fluxes.get(1)));
                    }
                });
    });
```

mikeharder:
One more improvement to free the copied buffers sooner in the pipeline:

```java
private static Mono<String> UploadWindowBy(Flux<ByteBuffer> data) {
    final long chunkedUploadRequirement = 4 * 1024 * 1024;
    final long[] bufferedDataSize = {0};

    final LinkedList<ByteBuffer> queue = new LinkedList<ByteBuffer>();

    return data
            .windowUntil(buffer -> {
                if (bufferedDataSize[0] <= chunkedUploadRequirement) {
                    bufferedDataSize[0] += buffer.remaining();

                    if (bufferedDataSize[0] <= chunkedUploadRequirement) {
                        queue.add(deepCopy(buffer));
                        return false;
                    }
                    else {
                        return true;
                    }
                }
                else {
                    return false;
                }
            }, true)
            .buffer(2)
            .next()
            .flatMap(fluxes -> {
                if (fluxes.size() == 1) {
                    return UploadSmall(dequeuingFlux(queue), bufferedDataSize[0]);
                }
                else {
                    return UploadLarge(dequeuingFlux(queue).concatWith(fluxes.get(1)));
                }
            });
}

private static ByteBuffer deepCopy(ByteBuffer b)
{
    return ByteBuffer.allocate(b.remaining()).put(b).position(0);
}

private static Flux<ByteBuffer> dequeuingFlux(Queue<ByteBuffer> queue) {
    return Flux.generate(sink -> {
        ByteBuffer b = queue.poll();
        if (b != null) {
            sink.next(b);
        }
        else {
            sink.complete();
        }
    });
}
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: Done

joshfree:
@amishra-dev could you help get some more eyes on this final Storage PR?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -287,76 +289,145 @@ private SpecializedBlobClientBuilder prepareBuilder() {
...
+                        return true;
+                    } else {
+                        /*
+                         * Buffer until the first 4MB are emitted from the stream in case Put Blob is used. This is
+                         * done to support replayability which is required by the Put Blob code path, it doesn't buffer
+                         * the stream in a way that the Stage Blocks and Put Block List code path does, and this API
+                         * explicitly states that it supports non-replayable streams.
+                         */
+                        ByteBuffer cachedBuffer = ByteBuffer.allocate(buffer.remaining()).put(buffer);
+                        cachedBuffer.position(0);
```
srnagar: Consider using `cachedBuffer.flip()` instead. It updates the state of a few things underneath -  like limit, mark and position which may be relevant to something downstream.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -287,76 +289,145 @@ private SpecializedBlobClientBuilder prepareBuilder() {
...
+            }) // TODO: parallelism?
+            .collect(Collectors.toList())
+            .flatMap(ids ->
+                blockBlobAsyncClient.commitBlockListWithResponse(ids, headers, metadata, tier, accessConditions));
+    }
+
+    private Mono<Response<BlockBlobItem>> determineUploadFullOrChunked(final Flux<ByteBuffer> data,
+        final Function<Flux<ByteBuffer>, Mono<Response<BlockBlobItem>>> uploadInChunks,
+        final BiFunction<Flux<ByteBuffer>, Long, Mono<Response<BlockBlobItem>>> uploadFullBlob) {
+        final long chunkedUploadRequirement = 4 * Constants.MB;
```
srnagar: Are we going to make 4 MB the default size to decide whether to chunk or not? If so, make this a static constant. I guess this way, the caching of BB is controlled by us and the overhead is limited to a max of 8 MB.
The other option is to use the block size in `ParallelTransferOptions` which can be 1 GB and we'll have to cache it resulting in a noticeable increase in memory usage.

alzimmermsft: For now this will be made into a static constant.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -287,76 +289,145 @@ private SpecializedBlobClientBuilder prepareBuilder() {
...
-                        });
-                });
+                parallelTransferOptions.getNumBuffers(), parallelTransferOptions.getProgressReceiver());
+
+            BlockBlobAsyncClient blockBlobAsyncClient = getBlockBlobAsyncClient();
+            return determineUploadFullOrChunked(data, (stream) -> uploadInChunks(blockBlobAsyncClient, stream,
+                finalParallelTransferOptions, headers, metadata, tier, accessConditionsFinal),
+                (stream, length) -> blockBlobAsyncClient.uploadWithResponse(
+                    ProgressReporter.addProgressReporting(stream, finalParallelTransferOptions.getProgressReceiver()),
+                    length, headers, metadata, tier, null, accessConditionsFinal));
```
jaschrep-msft: Nit. Please use better indentation, as there's a lot of nested calls in here.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
jaschrep-msft: Even if we stick with this 4 megabyte value, what happens if block size in parallel transfer options is higher than this? This looks as if the following steps are possible

1. Try to upload 1023MB blob with 1GB block size
2. See that blob is larger than 4MB and then move to chunked
3. Upload the entire blob in a put block call and then commit block list

I don't know if it's the right decision to do all this work for smart branching only to wind up in this scenario. If we allow a block size as high as 1GB then we should support it throughout our decision making.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -287,76 +289,145 @@ private SpecializedBlobClientBuilder prepareBuilder() {
...
+                        cachedBuffers.add(cachedBuffer);
+                        return false;
+                    }
+                }
+            }, true)
+            .buffer(2)
+            .next()
+            .flatMap(fluxes -> {
+                if (fluxes.size() == 1) {
+                    return uploadFullBlob.apply(Flux.fromIterable(cachedBuffers), bufferedDataSize[0]);
```
anuchandy: should we use `dequeuingFlux` here as well instead of `Flux.fromIterable(cachedBuffers)`?

alzimmermsft: This was changed to using `fromIterable` as that is replayable whereas `dequeuingFlux` isn't, this code path needs the flux to be replayable as part of it's contract.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: Will use flip

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -287,76 +289,145 @@ private SpecializedBlobClientBuilder prepareBuilder() {
...
+            parallelTransferOptions.getBlockSize());
+
+        /*
+        Break the source Flux into chunks that are <= chunk size. This makes filling the pooled buffers much easier
+        as we can guarantee we only need at most two buffers for any call to write (two in the case of one pool
+        buffer filling up with more data to write). We use flatMapSequential because we need to guarantee we
+        preserve the ordering of the buffers, but we don't really care if one is split before another.
+         */
+        Flux<ByteBuffer> chunkedSource = data
+            .filter(ByteBuffer::hasRemaining)
```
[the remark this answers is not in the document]
alzimmermsft: I'll need to test it but I remember running into issues with having this on the `determine...` method but I'll test it out

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -287,76 +289,145 @@ private SpecializedBlobClientBuilder prepareBuilder() {
...
+                finalParallelTransferOptions, headers, metadata, tier, accessConditionsFinal),
+                (stream, length) -> blockBlobAsyncClient.uploadWithResponse(
+                    ProgressReporter.addProgressReporting(stream, finalParallelTransferOptions.getProgressReceiver()),
+                    length, headers, metadata, tier, null, accessConditionsFinal));
+        } catch (RuntimeException ex) {
+            return monoError(logger, ex);
+        }
+    }
+
+    private Mono<Response<BlockBlobItem>> uploadInChunks(BlockBlobAsyncClient blockBlobAsyncClient,
```
jaschrep-msft: It looks as if there were many perf optimizations made in this method. Are these not applicable to `uploadBlocks()`, the similar method used when we upload from a file and so know the length already? Or are all the optimizations based on the scenario of not having this information present? If they're truly incompatible then so be it.

mikeharder:
After this change, I am seeing the same performance from both `BlockBlobClient.Upload(Flux, size)` and `BlobClient.Upload(Flux)`.

  | Upload(Flux, size) | Upload(Flux)
-- | -- | --
10 KB | 4984 | 5252
10 MB | 31.22 | 31.4
256 MB | 1.24 | 1.24

I manually checked the resource usage of the 256 MB scenario using `top`, and CPU and memory are similar for both APIs.  This change looks good from a perf perspective.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: I think @srnagar slightly misspoke, the maximum allowed for a stage block call is 100MB.

It is a good point about using Put Blob up to the limit of a passed buffer size. I would have a slight concern if someone accidentally passed 10KB instead of 10MB and other order of magnitude mistakes, possibly doing a Math.max(4MB, bufferSize).

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: The methods are fairly similar in name and parameters aside from that they differ largely in their intended use. `uploadBlocks` (maybe rename it to something mentioning files) will read a file from disk and upload it using a cold (replayable) flux, `uploadInChunks` in meant to be used by hot (non-replayable) fluxes which need to have an intermediate buffering applied. Having `uploadFromFile` call into `uploadInChunks` would lead to additional buffering overhead that isn't needed. I think there are definitely pieces that could be take out of both methods and put somewhere common but it's out of scope for this change, luckily this would be an implementation change and could be done in the future without breaking API.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
alzimmermsft: Done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
rickle-msft: Perhaps some added inline docs to state the intent and effect of some of the less intuitive steps would help clarify this.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
jaschrep-msft: Right about the 100MB. My mind just went there because of the 1GB remark.

Anyway, I do agree with Rick in that we should be listening to the customer's configuration settings if they are provided, and using this constant here seems like we're ignoring those values for our decision making.

alzimmermsft: I'm good with making the change to allow the `blockSize` configuration to determine which code path is taken, will make sure to update the Javadocs for this method to indicate that its value will determine whether Put Blob or Stage Blocks is used.

alzimmermsft: Updated to use `blockSize` to determine the Put Blob threshold.

alzimmermsft:
@jaschrep-msft could you verify that the last commit meets your request and we can close on this PR.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
rickle-msft: Wait sorry for the churn. I think we want this to have the same putBlob vs. putBlock behavior as uploadFromFile for consistency between APIs, and uploadFromFile should have a separate value in the parallelTransferOptions, otherwise they can't make use of the 256mb limit on putBlob. My suggestion earlier of configuration was meant as a future work item to be alongside when we add this for uploadFromFile. I propose we always use the default for the blocksize now, and we'll add configuration in as a feature for both apis later.

alzimmermsft: Ah, I'll revert to the last commit.

alzimmermsft: Issue to add this option to `ParallelTransferOptions` https://github.com/Azure/azure-sdk-for-java/issues/6097

PR #6065, opened 2019-10-28
Move SAS classes into SAS package

### LINKED ISSUE #6064: [API] Move SAS Classes into SAS Packages
Move the SAS token generation related classes into a SAS package inside of each module.

com.azure.storage.blob.sas
com.azure.storage.file.sas
com.azure.storage.queue.sas

alzimmermsft:
Fixes #6064 

Moves the SAS token generation classes into a SAS package for each module.


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/package-info.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Package containing sas classes used by Azure Storage Blobs.
```
JonathanGiles: upper-case the 'sas' and expand it in brackets to help users understand what this means.

alzimmermsft: SAS [shared access signature] or SAS (shared access signature)?

alzimmermsft:
@rickle-msft This is the change discussed in the standup to move these classes into the `sas` package. Could you give this a review?

PR #6070, opened 2019-10-29
Added log message for missing playback jsons.

sima-zhu:
This is customer request changes. Failed a little earlier if the playback json is missing. Log more useful message.


sdk/core/azure-core-test/src/main/java/com/azure/core/test/InterceptorManager.java
```diff
@@ -191,8 +193,11 @@ private File getRecordFile(String testName) {
         }
 
         String filePath = folderFile.getPath() + "/" + testName + ".json";
```
alzimmermsft: Instead of generating the file path, checking existence, failing or get the file why not use

```java
File playbackFile = new File(folderFile, testName + ".json");
if (!playbackFile.exists()) {
    // throw exception
}

return playbackFile;
```

sima-zhu:
Addressed feedback from Alan

PR #6076, opened 2019-10-29
Set resourceResponse in CosmosResponse

kushagraThapar:
Sync responses were not setting resourceResponse in parent class `CosmosResponse`

This was causing sync responses to throw NPE in a bunch of methods like 
`getRequestLatency()`
`getRequestCharge()`
and almost every other method.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosContainerTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosContainerTest.java
@@ -8,7 +8,6 @@ package com.azure.cosmos;
 
 import com.azure.cosmos.rx.TestSuiteBase;
 import com.azure.cosmos.internal.HttpConstants;
-import com.azure.cosmos.rx.TestSuiteBase;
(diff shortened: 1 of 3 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainerResponse.java
```diff
@@ -12,7 +12,7 @@
     private final CosmosContainer container;
 
     CosmosContainerResponse(CosmosAsyncContainerResponse response, CosmosDatabase database, CosmosClient client) {
-        super(response.getProperties());
+        super(response.resourceResponseWrapper, response.getProperties());
```
kirankumarkolli: Won't sending wrapper out  break design abstractions?

kushagraThapar: wrapper is already being used in all the public API classes as to fetch the properties. This won't break out the abstraction design. 
This will just make sure to assign the wrapper in the parent class from the response object.

Issue #5344, opened 2019-09-11
Consider moving build tools into the azure-sdk-tools repo. 


weshaggard:
@JonathanGiles @jianghaolu it is generally a bad idea to build and install tools packages as a pre-step of building a repository. It adds an serialization point as well as step everyone must always remember todo which makes working in the repo less than ideal. For other repos and tools we generally build and publish the tools from another repo like azure-sdk-tools repo we should consider doing that for these tools as well. That way we can simply add a dependency to them in the parent somewhere and then folks don't have to manually go build and install them as a preset.

PR #6080, opened 2019-10-30
Build and module improvements

### LINKED ISSUE #5428: Fail the build if there are compiler warnings for java 9 plus builds
After adding support for modules in Java 9 plus builds, there are several compiler warnings related to visibility of modules. 

Fix these warnings and change `<failOnWarning>` flag to `true` in maven-compiler-plugin configuration of `java9plus` profile.

### LINKED ISSUE #5582: Simplify building libraries by using JDK 11
JDK 9 and later releases offers [JEP 247](http://openjdk.java.net/jeps/247), which enables compilation of source code that targets an older set of APIs. We could use this to build the Azure SDKs against the JDK 8 API set, whilst using JDK 11 to achieve this. We could therefore abandon JDK 8 on our machines and baseline on JDK 11, whilst still ensuring compatibility with JDK 8 users of our libraries.

This could also simplify our CI/CD system somewhat - we would still need to two multiple builds, but they would simply both use JDK 11 with a release flag that is switched from the default of '8' to instead be '11' to validate continued support for the latest 'LTS' version.

// cc @mitchdenny @JimSuplizio @srnagar

JonathanGiles:
This is an exploration PR that cleans up builds a little, in terms of the pom.client.xml file, as well as the pipelines and the module-info files. It is an ongoing exploration but I hope that by pushing changes to this PR, that the CI pipeline will begin to validate my changes and give feedback.

I will request review before anything of this nature gets merged, so please do not bother reviewing eagerly (unless you are interested).

Fixes #5582
Fixes #5428 (partially)

### DIFF
--- a/eng/spotbugs-aggregate-report/pom.xml
+++ b/eng/spotbugs-aggregate-report/pom.xml
@@ -55,9 +55,9 @@
 <!--                <source>..\..\sdk\keyvault\azure-security-keyvault-certificates\src\main\java\com</source>-->
 <!--                <source>..\..\sdk\keyvault\azure-security-keyvault-certificates\src\samples\java\com</source>-->
                 <source>..\..\sdk\keyvault\azure-security-keyvault-keys\src\main\java\com</source>
-                <source>..\..\sdk\keyvault\azure-security-keyvault-keys\src\samples\java\com</source>
+<!--                <source>..\..\sdk\keyvault\azure-security-keyvault-keys\src\samples\java\com</source>-->
                 <source>..\..\sdk\keyvault\azure-security-keyvault-secrets\src\main\java\com</source>
-                <source>..\..\sdk\keyvault\azure-security-keyvault-secrets\src\samples\java\com</source>
+<!--                <source>..\..\sdk\keyvault\azure-security-keyvault-secrets\src\samples\java\com</source>-->
                 <source>..\..\sdk\storage\azure-storage-common\src\main\java\com</source>
                 <source>..\..\sdk\storage\azure-storage-common\src\samples\java\com</source>
                 <source>..\..\sdk\storage\azure-storage-blob\src\main\java\com</source>


pom.client.xml
```diff
@@ -684,31 +712,69 @@
       </reporting>
     </profile>
 
-    <!-- Setup for Java 9+ -->
+    <!-- Setup for Java 11+ -->
     <profile>
       <id>java-lts</id>
       <activation>
-        <jdk>[9,)</jdk>
+        <jdk>[11,)</jdk>
```
srnagar: So, we will not support building on JDK 9 or 10? The profiles are now only for 8 and 11+.

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -33,7 +33,6 @@
   </scm>
 
   <properties>
```
srnagar: The whole `<properties>` section can be deleted as it's empty now.

eng/spotbugs-aggregate-report/pom.xml
```diff
@@ -49,16 +48,19 @@
...
                 <source>..\..\sdk\identity\azure-identity\src\main\java\com</source>
                 <source>..\..\sdk\identity\azure-identity\src\samples\java\com</source>
-                <!--                <source>..\..\sdk\keyvault\azure-keyvault-keys\src\main\java\com</source>-->
-                <!--                <source>..\..\sdk\keyvault\azure-keyvault-keys\src\samples\java\com</source>-->
-                <!--                <source>..\..\sdk\keyvault\azure-keyvault-secrets\src\main\java\com</source>-->
-                <!--                <source>..\..\sdk\keyvault\azure-keyvault-secrets\src\samples\java\com</source>-->
+<!--                <source>..\..\sdk\keyvault\azure-security-keyvault-certificates\src\main\java\com</source>-->
+<!--                <source>..\..\sdk\keyvault\azure-security-keyvault-certificates\src\samples\java\com</source>-->
+                <source>..\..\sdk\keyvault\azure-security-keyvault-keys\src\main\java\com</source>
+<!--                <source>..\..\sdk\keyvault\azure-security-keyvault-keys\src\samples\java\com</source>-->
```
srnagar: Can the samples be uncommented too for Keys and Secrets?

JonathanGiles: They currently have spotbugs issues. @g2vinay will be resolving this soon.

pom.client.xml
JonathanGiles: I will need to test again on 9. Previously I ran into bugs in 9 that prevented compilation.

eng/jacoco-test-coverage/pom.xml
```diff
@@ -55,11 +54,27 @@
       <artifactId>azure-core-http-netty</artifactId>
       <version>1.0.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;current} -->
     </dependency>
+    <dependency>
```
weshaggard: Do these tools have to have strong dependencies on all the libraries in the repo? Is there not a way to have the tools take the libraries as an argument when running them as opposed to being required to depend on them?

I filed issue https://github.com/Azure/azure-sdk-for-java/issues/5344 a little while ago asking questions about these tools and whether or not we had a way to remove them as a pre-step of the build and instead pull them from a package an run them instead of directly referencing everything.

JonathanGiles:
@mitchdenny @JimSuplizio @weshaggard This PR is now at the point where I have made all the build-related changes I want to make. There are additional changes I am going to make, but it relates to the azure-core modules which are probably not as interesting to you. If you could please review the build-related changes (filter by pom.xml and *.yml), that would be appreciated.

JonathanGiles:
This PR is now ready to merge pending review approval. I'm removing the labels and requesting interested parties to please work through this as soon as possible, so it bit-rots as little as possible.

JonathanGiles:
Test failures appear to be intermittent build failures.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponseBase.java
```diff
@@ -1,12 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-package com.azure.core.implementation.http;
+package com.azure.core.http.rest;
```
srnagar: Some of the classes in implementation packages were lean on javadocs (like this one is missing javadoc for both constructors).  Maybe it's hard to catch all of them here. We should add checkstyle rule to report missing javadocs  on public methods/classes in non-impl packages.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerAdapter.java
```diff
@@ -71,17 +73,18 @@
      * {@code
      *      class FooMetadataHeaders {
      *          String name;
-     *          @HeaderCollection("header-collection-prefix-")
+     *          {@literal @}HeaderCollection("header-collection-prefix-")
      *          Map<String,String> headerCollection;
```
srnagar: nit: candidate for codesnippet.

sdk/core/azure-core/src/main/java/module-info.java
```diff
@@ -25,109 +25,26 @@
...
     opens com.azure.core.util to com.fasterxml.jackson.databind;
     opens com.azure.core.util.logging to com.fasterxml.jackson.databind;
+    opens com.azure.core.util.serializer to com.fasterxml.jackson.databind;
     opens com.azure.core.implementation to com.fasterxml.jackson.databind;
     opens com.azure.core.implementation.serializer to com.fasterxml.jackson.databind;
-    opens com.azure.core.implementation.serializer.jackson to com.fasterxml.jackson.databind;
-    opens com.azure.core.implementation.util to com.fasterxml.jackson.databind;
     opens com.azure.core.implementation.serializer.jsonwrapper to com.fasterxml.jackson.databind;
 
-    // exporting some packages for internal use only
```
srnagar: Nice! 👍

eng/code-quality-reports/src/main/java/com/azure/tools/checkstyle/checks/ExternalDependencyExposedCheck.java
```diff
@@ -24,7 +24,7 @@
     private static final String EXTERNAL_DEPENDENCY_ERROR =
         "Class ''%s'', is a class from external dependency. You should not use it as a %s type.";
     private static final Set<String> VALID_DEPENDENCY_SET = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
-        "java", "com.azure", "reactor"
+        "java", "com.azure", "reactor", "org.reactivestreams"
```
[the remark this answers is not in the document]
JonathanGiles: FluxUtil exposes `Subscription` and CoreUtils exposes `Publisher`.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponseBase.java
JonathanGiles: I've added additional checkstyle rules and will add more javadocs as it identifies issues.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/AzureBlobStorageBuilder.java
```diff
@@ -85,7 +88,12 @@ public AzureBlobStorageBuilder pipeline(HttpPipeline pipeline) {
      */
     public AzureBlobStorageImpl build() {
         if (pipeline == null) {
-            this.pipeline = RestProxy.createDefaultPipeline();
+            this.pipeline = new HttpPipelineBuilder()
+                       .policies(
```
alzimmermsft: nit: Indentation

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/AzureBlobStorageBuilder.java
```diff
@@ -85,7 +88,12 @@ public AzureBlobStorageBuilder pipeline(HttpPipeline pipeline) {
      */
     public AzureBlobStorageImpl build() {
         if (pipeline == null) {
-            this.pipeline = RestProxy.createDefaultPipeline();
+            this.pipeline = new HttpPipelineBuilder()
```
alzimmermsft: This class and all others with a similar name are codegen, you'll need to work with @jianghaolu to update this appropriately.

For example Storage doesn't use the `RetryPolicy` in Azure Core, instead the `RequestRetryPolicy` is used.

alzimmermsft: I don't think this code path should be reach when building one of the public clients, but want to make sure we build the right thing if we do.

JonathanGiles: @alzimmermsft This code is copy / paste from the old `RestProxy.createDefaultPipeline()` code path that was called (which I have removed, hence the copy / paste). This means that there should be no change in behaviour before / after this code change. If the RetryPolicy should not be used, then that needs to be cleaned up consistently (but probably in a follow-up PR).

JonathanGiles:
@mitchdenny @JimSuplizio @weshaggard @danieljurek Could I please get some eyes on the engsys related changes in this PR please? I'm keen to get this merged as it bitrots a little bit every day. Thanks.

eng/jacoco-test-coverage/pom.xml
JonathanGiles: I don't know - this would be a good thing for us to investigate, as it is a nuisance.

pom.client.xml
```diff
@@ -684,31 +712,69 @@
       </reporting>
     </profile>
 
-    <!-- Setup for Java 9+ -->
+    <!-- Setup for Java 11+ -->
     <profile>
       <id>java-lts</id>
```
weshaggard: Should we change the profile id to java-11?

pom.client.xml
```diff
@@ -684,31 +712,69 @@
...
       </activation>
       <build>
         <plugins>
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-compiler-plugin</artifactId>
             <version>${maven-compiler-plugin.version}</version>
-            <!-- setup defaults for compile and testCompile -->
             <configuration>
-              <source>9</source>
```
weshaggard: Does this mean you we plan to still use the default target of java 8 for the compile and target now?

JonathanGiles: The intention is to compile with JDK 11 but run to target JDK 8 APIs. There is a secondary run (internally) that compiles just the module-info.java files at JDK 11 level. These are then combined into a single release jar. This means that we can drop the JDK 8 build pipeline (we still keep JDK 8 for the test pipelines though).

pom.client.xml
JonathanGiles: We could do. In a few years we would need to rename this to java-17 when that becomes the LTS version, which is why I've hesitated so far.

eng/pipelines/client.yml
```diff
@@ -212,21 +208,21 @@ jobs:
...
           JavaVersion: '1.8'
           ProfileFlag: ''
-        Linux - Java LTS:
+        Linux - Java 11:
           OSName: 'Linux'
           OSVmImage: 'ubuntu-16.04'
           JavaVersion: '1.11'
-          ProfileFlag: '-Djava-lts'
-        macOS - Java LTS:
+          ProfileFlag: ''
```
weshaggard: We should remove the `ProfileFlag` now that we don't plan to use it any longer.

JonathanGiles: I wondered if this should remain or not. I'll remove it.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/AzureBlobStorageBuilder.java
jianghaolu: This is an easy fix in the generator.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -46,13 +40,8 @@ jobs:
 
     strategy:
```
mitchdenny: We don't need the matrix here if we only have one thing in it, its just a unmatrixed job.

pom.client.xml
weshaggard: Given we aren't really using it by name any longer I guess it doesn't really matter.

pom.client.xml
weshaggard: Sounds good.

PR #6085, opened 2019-10-30
EventHubConsumer updates

### LINKED ISSUE #5994: Create EventHubConsumerClient
https://gist.github.com/srnagar/68f0d06b862eaa5ca94f6d3a1f7f2622

* Add receiveAll() to consumer
* EventHubClient.getEventhubPropderites() and EventHubClient.getPartitionsIds() duplicated on the producer/consumer

conniey:
* EventHubConsumer (sync and async) is no longer tied to a single partition.
* EventHubConsumerAsyncClient can receive from all partitions.
* Consumer can be created from EventHubClientBuilder
* **NOTE** Does not include making EHC package-private. 

Fixes #5994

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProperties.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProperties.java
@@ -4,7 +4,6 @@
 package com.azure.messaging.eventhubs;
 
 import com.azure.core.annotation.Immutable;
-import com.azure.messaging.eventhubs.models.EventPosition;
 
 import java.time.Instant;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
```diff
@@ -280,6 +290,71 @@ public EventHubClientBuilder retry(RetryOptions retryOptions) {
         return this;
     }
 
+    /**
+     * Sets the position within the partition where the consumer should begin reading events.
+     *
+     * @param eventPosition Position within an Event Hub partition to begin consuming events.
+     * @return The updated {@link EventHubClientBuilder} object.
+     */
+    public EventHubClientBuilder startingPosition(EventPosition eventPosition) {
```
srnagar: startingPosition or initialPosition? What are other languages using?

conniey: .net uses it. So I copied from them.

PR #6095, opened 2019-10-30
readme updates

JimSuplizio:
There are two different readme updates in this PR. The first is to change the doc links from https://azure.github.io/azure-sdk-for-java/track2reports/index.html to https://azure.github.io/azure-sdk-for-java to just link to the repo root. The following files are affected by this change:
- sdk\core\azure-core\README.md
- sdk\core\azure-core-tracing-opencensus\README.md
- sdk\identity\azure-identity\README.md
- sdk\core\README.md
- sdk\eventhubs\azure-messaging-eventhubs-checkpointstore-blob\README.md
- sdk\keyvault\azure-security-keyvault-certificates\README.md
- sdk\keyvault\azure-security-keyvault-keys\README.md
- sdk\keyvault\azure-security-keyvault-secrets\README.md

The second change is spacing. Code blocks in markdown need to have blank lines before and after the code block. In some cases not having these blank lines resulted in erroneous code blocks when the markdown was converted into html which seemed to happen when dealing with the maven dependency include.

### DIFF
--- a/sdk/core/README.md
+++ b/sdk/core/README.md
@@ -1,6 +1,6 @@
 # Azure Core shared library for Java
 
-[![Build Documentation](https://img.shields.io/badge/documentation-published-blue.svg)](https://azure.github.io/azure-sdk-for-java/track2reports/index.html)
+[![Build Documentation](https://img.shields.io/badge/documentation-published-blue.svg)](https://azure.github.io/azure-sdk-for-java)
 
 Azure Core provides shared primitives, abstractions, and helpers for modern Java Azure SDK client libraries. These libraries follow the [Azure SDK Design Guidelines for Java](https://azuresdkspecs.z5.web.core.windows.net/JavaSpec.html) and can be easily identified by package names starting with `com.azure` and module names starting with `azure-`, e.g. `com.azure.storage.blobs` would be found within the `/sdk/storage/azure-storage-blob` directory. A more complete list of client libraries using Azure Core can be found [here](https://github.com/Azure/azure-sdk-for-java).
 


sdk/core/README.md
```diff
@@ -1,6 +1,6 @@
 # Azure Core shared library for Java
 
-[![Build Documentation](https://img.shields.io/badge/documentation-published-blue.svg)](https://azure.github.io/azure-sdk-for-java/track2reports/index.html)
+[![Build Documentation](https://img.shields.io/badge/documentation-published-blue.svg)](https://azure.github.io/azure-sdk-for-java)
```
weshaggard: I was under the impression this link would be `https://azure.github.io/azure-sdk-for-java/core.html` although it doesn't look like that exists yet. What you do have seems to be a direct replacement so I'm good  with that for now but otherwise I do think we should try to be more specific to the service.

cc @scbedd

JimSuplizio: For right now we're going to leave this linking to https://azure.github.io/azure-sdk-for-java however ALL of the readme's are going to need to be changed to link to their respective areas (core to core, keyvault to keyvault etc)

sdk/core/azure-core/README.md
```diff
@@ -1,6 +1,6 @@
 # Azure Core client library for Java
 
-[![Build Documentation](https://img.shields.io/badge/documentation-published-blue.svg)](https://azure.github.io/azure-sdk-for-java/track2reports/index.html)
+[![Build Documentation](https://img.shields.io/badge/documentation-published-blue.svg)](https://azure.github.io/azure-sdk-for-java)
```
joshfree: quick comment --- the new landing page needs improvements - "Latest SDK" and "Previous SDK" for "Track 2" and "Track 1" isn't very clear for customers.  @kaerm is this a doc asset that you're working on?

kaerm: I am not. Also, unfortunately I do not think there is a better way for us to refer to them as we are not supposed to have a name or even use tracks. (I personally fully agree with the sentiment though) @joshfree

JimSuplizio: @joshfree what needs to get updated is that these, will ultimately, should probably point to the same library that the readme is for (see @weshaggard's comment above). The doc index hasn't been updated (it has to wait, there's a change that needs to go into pom.client.xml and pom.data.xml those need to wait until the release is finished) and that landing page is broken up by area (keyvault, core, cosmos etc) which contains the list of libraries in that area both tracks. I have my own github.io updated and would be happy to show you what the docs are going to look like.

PR #6098, opened 2019-10-30
Remove inkrecognizer from code base.

mitchdenny:
Removes ink recognizer from the code base.


minaltolpadi:
Confirming that it is okay to remove the InkRecognizer module from this repo.

JimSuplizio:
@mitchdenny - it would appear that the inkrecognizer also needs to get removed from pom.data.xml

a-p-o:
I found the Ink Recognizer code useful. Is there another repository with the code?

PR #6108, opened 2019-10-31
Updating staging profile id to match com.azure namespace

kushagraThapar:
Because of namespace change to com.azure from com.microsoft.azure we need to change the staging profile id as well.


mitchdenny:
Checked staging id and this is correct. Thanks.

PR #6125, opened 2019-10-31
Add EventHubConnection

conniey:
* Exposes a way to share connections using EventHubConnection
  * Producers and consumers only dispose of the connection if it is a dedicated connection
* EventHubClient and AsyncClient are package-private.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/README.md
+++ b/sdk/eventhubs/azure-messaging-eventhubs/README.md
@@ -69,23 +69,24 @@ documentation][event_hubs_product_docs] | [Samples][sample_examples]
 For the Event Hubs client library to interact with an Event Hub, it will need to understand how to connect and authorize
 with it.
 
-### Create an Event Hub client using a connection string
+### Create an Event Hub producer using a connection string
 
 The easiest means for doing so is to use a connection string, which is created automatically when creating an Event Hubs
 namespace. If you aren't familiar with shared access policies in Azure, you may wish to follow the step-by-step guide to
 [get an Event Hubs connection string][event_hubs_connection_string].
 
-Both the asynchronous and synchronous Event Hub clients can be created using `EventHubClientBuilder`. Invoking
-`.buildAsyncClient()` will create an `EventHubAsyncClient` and invoking `.buildClient()` will create an `EventHubClient`.
+Both the asynchronous and synchronous Event Hub producer and consumer clients can be created using 
+`EventHubClientBuilder`. Invoking `.buildAsyncProducer()` and `buildProducer` will build the asynchronous and 
+synchronous producers. Similarly, `.buildAsyncConsumer` and `.buildConsumer` will build the appropriate consumers.
 
-The snippet below creates an asynchronous Event Hub client.
+The snippet below creates an asynchronous Event Hub producer.
 
 ```java
 String connectionString = "<< CONNECTION STRING FOR THE EVENT HUBS NAMESPACE >>";
 String eventHubName = "<< NAME OF THE EVENT HUB >>";
-EventHubAsyncClient client = new EventHubClientBuilder()
+EventHubProducerAsyncClient client = new EventHubClientBuilder()
     .connectionString(connectionString, eventHubName)
-    .buildAsyncClient();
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClient.java
```diff
@@ -20,22 +19,11 @@
...
- *
- * {@codesnippet com.azure.messaging.eventhubs.eventhubclient.instantiation}
- *
  * @see EventHubClientBuilder
  * @see EventHubAsyncClient To communicate with Event Hub using an asynchronous client.
  * @see <a href="https://docs.microsoft.com/Azure/event-hubs/event-hubs-about">About Azure Event Hubs</a>
  */
-@ServiceClient(builder = EventHubClientBuilder.class)
-public class EventHubClient implements Closeable {
+class EventHubClient implements Closeable {
```
srnagar: Maybe we can remove the package-private sync client now. Since EventProcessor uses only async client.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
```diff
@@ -209,6 +210,18 @@ public EventHubClientBuilder configuration(Configuration configuration) {
...
     }
 
+    /**
+     * Sets the Event Hub connection to use when interacting with Event Hubs. If not set, a new connection will be
+     * constructed and used.
+     *
+     * @param eventHubConnection Event Hub connection to use.
+     * @return The updated {@link EventHubClientBuilder} object.
+     */
+    public EventHubClientBuilder connection(EventHubConnection eventHubConnection) {
```
srnagar: Check for null.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
```diff
@@ -445,25 +458,28 @@ public EventHubClient buildClient() {
         return new EventHubClient(client, connectionOptions);
     }
 
-    private static EventHubAsyncClient buildAsyncClient(ConnectionOptions connectionOptions) {
+    private EventHubAsyncClient buildAsyncClient(ConnectionOptions connectionOptions) {
```
srnagar: `buildAsyncClient()` overload should be package-private and `buildClient()` should be deleted.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClient.java
conniey: It's used to create a synchronous receiver and consumer, so I kept it. :)

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
conniey: In general, I think we do validation on build, so if they set this to null, we will just create a new connection for them.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
conniey: Oh yeah, I thought I had reverted this. I had used it to verify samples, so users couldn't copy code that was broken.

conniey: Oh, this is the private one used by the builder, there is a package-private no parameter one

PR #6129, opened 2019-10-31
Add toReactorContext to FluxUtil

### LINKED ISSUE #5838: FluxUtil method toReactorContext
A large number of the APIs in the storage file.datalake package are simply wrappers that call into APIs into the storage blob package. 
Currently a data lake async client is easily able to call into the corresponding async methods in blob, however a data lake sync client max overload (that has a parameter for Context) is not able to call the package private max overload in the blob async client. 

An ideal solution to this problem would be for the data lake sync clients to subscribe to the async max overload like follows using a FluxUtil.toReactorContext method

```
public Response<Void> setMetadataWithResponse(Map<String, String> metadata, PathAccessConditions accessConditions,
        Mono<Response<Void>> response = client.setMetadataWithResponse(metadata, accessConditions)
            .subscriberContext(FluxUtil.toReactorContext(context));
        return Utility.blockWithOptionalTimeout(response, timeout);
    }
```

alzimmermsft:
Fixes #5838 

Adds a `toReactorContext` method to `FluxUtil` which converts an Azure Core `Context` to a Reactor `Context`.

### DIFF
--- a/sdk/core/azure-core/src/test/java/com/azure/core/util/FluxUtilTest.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/util/FluxUtilTest.java
@@ -3,8 +3,7 @@
 
 package com.azure.core.util;
 
-import org.junit.Assert;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
@@ -12,8 +11,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class FluxUtilTest {
     @Test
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/implementation/util/FluxUtil.java
```diff
@@ -178,11 +178,31 @@ private static void accept(ByteArrayOutputStream byteOutputStream, ByteBuffer by
      * @return The azure context
      */
     private static Context toAzureContext(reactor.util.context.Context context) {
-        Map<Object, Object> keyValues = context.stream().collect(Collectors.toMap(Entry::getKey, Entry::getValue));
-        if (ImplUtils.isNullOrEmpty(keyValues)) {
+        if (context == null) {
```
srnagar: This check is not needed. Reactor guarantees that `subscriberContext()` will return a non-null context.

alzimmermsft: Removed

sdk/core/azure-core/src/test/java/com/azure/core/implementation/util/FluxUtilTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.azure.core.implementation.util;
+
+import com.azure.core.util.Context;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class FluxUtilTest {
```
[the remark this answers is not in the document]
alzimmermsft: Done

PR #6135, opened 2019-11-01
Auto update App config version and name for useragent string

### LINKED ISSUE #3141: Remove hardcoded SDK_NAME and SDK_VERSION from ConfigurationClient
We should be reading the metadata from the JAR file rather than having constants.

Since this is used in the UserAgentPolicy, it should conform to this standard. https://azuresdkspecs.z5.web.core.windows.net/JavaSpec.html#general-telemetry

srnagar:
The intent of this change is to read the version and artifact name from maven pom.xml instead of hard-coded strings in Java classes which can get out of sync with the real version.

Partially fixes #3141 

This is just a POC, will be extended to remaining client libraries in future PRs.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 package com.azure.core.util;
 
 import com.azure.core.util.logging.ClientLogger;
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/AppConfigUserAgentPropertiesTest.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/AppConfigUserAgentPropertiesTest.java
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 package com.azure.data.appconfiguration;
 
 import static org.junit.Assert.assertNotNull;
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/UserAgentProperties.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/UserAgentProperties.java
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 package com.azure.core.util;
 
 /**
--- a/sdk/core/azure-core/src/test/java/com/azure/core/util/AzureUserAgentUtilTest.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/util/AzureUserAgentUtilTest.java
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 package com.azure.core.util;
 
 import static org.junit.Assert.assertNotNull;


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/AzureConfiguration.java
```diff
@@ -3,13 +3,59 @@
...
+    private static final String NAME;
+    private static final String VERSION;
+    private static final String VERSION_NOT_DEFINED = "Version not defined";
+    private static final String ARTIFACTID_NOT_DEFINED = "ArtifactId not defined";
+    private static final String AZURE_APPCONFIG_PROPERTIES = "azure-appconfig.properties";
+    private static final String VERSION_PROPERTY_NAME = "version";
+    private static final String ARTIFACT_PROPERTY_NAME = "artifact";
+
+    static {
+        String version = null;
```
alzimmermsft: Do we need to define these as null? Instead could we set them to the default values you have.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/AzureConfiguration.java
```diff
@@ -3,13 +3,59 @@
...
+            .getResourceAsStream(AZURE_APPCONFIG_PROPERTIES)) {
+            Properties properties = new Properties();
+            properties.load(inputStream);
+            version = properties.getProperty(VERSION_PROPERTY_NAME);
+            artifactId = properties.getProperty(ARTIFACT_PROPERTY_NAME);
+        } catch (Exception ex) {
+            LOGGER.warning("Failed to get AppConfiguration version and name from properties.", ex);
+        }
+        VERSION = version != null ? version : VERSION_NOT_DEFINED;
+        NAME = artifactId != null ? artifactId : ARTIFACTID_NOT_DEFINED;
```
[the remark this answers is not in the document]
srnagar: Sure, I'll update this to "unknown"

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/AzureConfiguration.java
srnagar: The `getProperty()` method below might return `null` if the property is not set. So, the last 2 lines that do the null check and set to default is still required. So, I didn't set it to default here.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/AzureConfiguration.java
```diff
@@ -3,13 +3,59 @@
...
  * Gets the SDK information for this library component.
  */
 class AzureConfiguration {
-    //TODO: Eventually remove these hardcoded strings with https://github.com/Azure/azure-sdk-for-java/issues/3141
-    static final String NAME = "application-configuration";
-    // {x-version-update-start;com.azure:azure-data-appconfiguration;current}
-    static final String VERSION = "1.0.0-preview.6";
-    // {x-version-update-end}
+
+    private static final ClientLogger LOGGER = new ClientLogger(AzureConfiguration.class);
```
mssfang: interesting. Does this trigger checkstyle  error?

mssfang: It should be non-static and use `logger` as the name

conniey:
Excited we no longer have to do a Ctrl+Shift+R and for these anymore.

Is it possible to put this common get version logic into azure-core? I can see us copy and pasting this deserialisation logic in all the other libraries.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/AzureConfiguration.java
srnagar: Since it's used in a static code block, it has to be static and hence uppercase `LOGGER`.

sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
```diff
@@ -0,0 +1,55 @@
...
+        ClientLogger logger = new ClientLogger(AzureUserAgentUtil.class);
+        String version = null;
+        String name = null;
+        try (InputStream inputStream = AzureUserAgentUtil.class.getClassLoader()
+            .getResourceAsStream(propertiesFileName)) {
+            Properties properties = new Properties();
+            properties.load(inputStream);
+            version = properties.getProperty(VERSION_PROPERTY);
+            name = properties.getProperty(NAME_PROPERTY);
+            logger.info("User agent properties: name = " + name + ", version = " + version + ", properties file = "
```
alzimmermsft: Info logging level or verbose?

sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param propertiesFileName The file name defining the user agent properties.
+     * @return the {@link UserAgentProperties}.
+     */
+    public static UserAgentProperties getUserAgentProperties(String propertiesFileName) {
+        return USER_AGENT_PROPERTIES.computeIfAbsent(propertiesFileName, AzureUserAgentUtil::readFromPropertiesFile);
+    }
+
+    private static UserAgentProperties readFromPropertiesFile(String propertiesFileName) {
+        ClientLogger logger = new ClientLogger(AzureUserAgentUtil.class);
+        String version = null;
```
alzimmermsft: Could initialize this to `UNKNOWN_VERSION`, if the try/catch throws an exception before loading this will be set to the default and we won't need to null check afterwards, this will require an additional change to use the `getProperty(String key, String defaultValue)` within the try/catch in case the property doesn't exist.

sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
```diff
@@ -0,0 +1,55 @@
...
+            version = properties.getProperty(VERSION_PROPERTY);
+            name = properties.getProperty(NAME_PROPERTY);
+            logger.info("User agent properties: name = " + name + ", version = " + version + ", properties file = "
+                + propertiesFileName);
+        } catch (IOException ex) {
+            logger.warning("Failed to get user agent properties from " + propertiesFileName, ex);
+        }
+        version = version != null ? version : UNKNOWN_VERSION;
+        name = name != null ? name : UNKNOWN_NAME;
+        UserAgentProperties userAgentProperties = new UserAgentProperties(name, version);
```
alzimmermsft: Do we need the local variable, instead could we just return the new object?

sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param propertiesFileName The file name defining the user agent properties.
+     * @return the {@link UserAgentProperties}.
+     */
+    public static UserAgentProperties getUserAgentProperties(String propertiesFileName) {
+        return USER_AGENT_PROPERTIES.computeIfAbsent(propertiesFileName, AzureUserAgentUtil::readFromPropertiesFile);
+    }
+
+    private static UserAgentProperties readFromPropertiesFile(String propertiesFileName) {
+        ClientLogger logger = new ClientLogger(AzureUserAgentUtil.class);
```
alzimmermsft: I've seen `static` loggers used in other utility classes, should those be removed or could this be promoted to a class level static?

srnagar:
Closing this in favor of https://github.com/Azure/azure-sdk-for-java/pull/6213

PR #6139, opened 2019-11-01
DR fixes for java sdk

simplynaveen20:
This PR contains miscellaneous fixes for disaster recovery.

1. Force refresh database account on write forbidden exception
2. Force refresh when read request failed due to endpoint unavailable if none of preferred region mentioned in connection policy is available
3. If we hit scenario like region gone or write forbidden exception twice then code was not doing cache refresh, means it will only handle one disaster recovery .
4. One bug is still open where we are not doing background location refresh if account is multi-master or user doesn't provided useMultipleWriteLocations = true, need to discuss.

### DIFF
--- a/sdk/cosmos/microsoft-azure-cosmos/src/test/java/com/azure/data/cosmos/internal/directconnectivity/GlobalEndPointManagerTest.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/test/java/com/azure/data/cosmos/internal/directconnectivity/GlobalEndPointManagerTest.java
@@ -259,9 +259,7 @@ public class GlobalEndPointManagerTest {
         Assert.assertTrue(availableReadEndpointByLocation.keySet().iterator().next().equalsIgnoreCase("East Asia"));
 
         AtomicBoolean isRefreshing = getIsRefreshing(globalEndPointManager);
-        AtomicBoolean isRefreshInBackground = getRefreshInBackground(globalEndPointManager);
         Assert.assertFalse(isRefreshing.get());
-        Assert.assertTrue(isRefreshInBackground.get());
 
         databaseAccount = new DatabaseAccount(dbAccountJson3);
         Mockito.when(databaseAccountManagerInternal.getDatabaseAccountFromEndpoint(Matchers.any())).thenReturn(Flux.just(databaseAccount));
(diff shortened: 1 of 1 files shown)


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/GlobalEndpointManager.java
```diff
@@ -164,17 +181,24 @@ public void close() {
 
                     return databaseAccountObs.map(dbAccount -> {
                         this.locationCache.onDatabaseAccountRead(dbAccount);
+                        this.isRefreshing.set(false);
                         return dbAccount;
                     }).flatMap(dbAccount -> {
                         // trigger a startRefreshLocationTimerAsync don't wait on it.
-                        this.startRefreshLocationTimerAsync();
+                        if(!this.refreshInBackground.get())
+                        {
```
moderakh: code style: please "{" on the same line as if, here and in newly added other places.

simplynaveen20: Done

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/GlobalEndpointManager.java
```diff
@@ -41,6 +41,7 @@
     private final ConnectionPolicy connectionPolicy;
     private final DatabaseAccountManagerInternal owner;
     private final AtomicBoolean isRefreshing;
+    private final AtomicBoolean refreshInBackground;
```
moderakh: why can't we reuse isRefreshing? what's the difference between isRefreshing and refreshInBackground?

simplynaveen20: That's the main part of this PR , isRefreshing -> if there is region outage then all the requests should not start account refresh and only one do the work . refreshInBackground -> this is for the background periodic task. 
Initially both the activities were govern by single flag and causing overlaps and  issues.
Please connect with me offline  I will walk you through the code and logic

kirankumarkolli:
@moderakh thoughts on UT?

moderakh:
@kirankumarkolli for new customers we need to always validate the multi region fail over scenario.
I thought about how we can automate this before to save us time.

My thought for automated testing
We can have a test which uses java azure management sdk to add/remove region (the scenario above), and during that time we can validate that some basic tests work during fail over.
This should be automatable and will save us time. If we decide to add this automated test, I would estimate half a week work for the test setup.

PR #6143, opened 2019-11-01
Download to file refactor

rickle-msft:
Includes eTag locking and cutting out an unnecessary getProperties call.

Some small auxiliary changes to re-enable htbb optimizations that came up as a part of resolving merge conflicts. I also fixed a race condition in the scrubEtagPolicy that I hit when testing with high parallelism.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
@@ -541,9 +541,8 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
         final ParallelTransferOptions finalParallelTransferOptions =
             ModelHelper.populateAndApplyDefaults(parallelTransferOptions);
         final BlobRequestConditions finalRequestConditions = (accessConditions == null)
-            ? new BlobRequestConditions()
-            : accessConditions;
-        ProgressReceiver progressReceiver = finalParallelTransferOptions.getProgressReceiver();
+            ? new BlobRequestConditions() : accessConditions;
+
 
         // See ProgressReporter for an explanation on why this lock is necessary and why we use AtomicLong.
         AtomicLong totalProgress = new AtomicLong();
@@ -551,14 +550,14 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
 
         final SortedMap<Long, String> blockIds = new TreeMap<>();
         return Flux.fromIterable(sliceFile(fileSize, finalParallelTransferOptions.getBlockSize(),
-            parallelTransferOptions == null || parallelTransferOptions.getBlockSize() == null))
+            parallelTransferOptions))
             .flatMap(chunk -> {
                 String blockId = getBlockID();
                 blockIds.put(chunk.getOffset(), blockId);
 
                 Flux<ByteBuffer> progressData = ProgressReporter.addParallelProgressReporting(
                     FluxUtil.readFile(channel, chunk.getOffset(), chunk.getCount()),
-                    progressReceiver, progressLock, totalProgress);
+                    finalParallelTransferOptions.getProgressReceiver(), progressLock, totalProgress);
 
                 return client.stageBlockWithResponse(blockId, progressData, chunk.getCount(), null,
                     finalRequestConditions.getLeaseId());
@@ -595,7 +594,8 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
         return Base64.getEncoder().encodeToString(UUID.randomUUID().toString().getBytes(StandardCharsets.UTF_8));
     }
 
-    private List<BlobRange> sliceFile(long fileSize, int blockSize, boolean enableHtbbOptimization) {
+    private List<BlobRange> sliceFile(long fileSize, int blockSize, ParallelTransferOptions originalOptions) {
+        boolean enableHtbbOptimization = originalOptions == null || originalOptions.getBlockSize() == null;
         List<BlobRange> ranges = new ArrayList<>();
         if (fileSize > 100 * Constants.MB && enableHtbbOptimization) {
             blockSize = BLOB_DEFAULT_HTBB_UPLOAD_BLOCK_SIZE;
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
@@ -723,16 +723,16 @@ public class BlobAsyncClientBase {
     Mono<Response<BlobProperties>> downloadToFileWithResponse(String filePath, BlobRange range,
         ParallelTransferOptions parallelTransferOptions, DownloadRetryOptions downloadRetryOptions,
         BlobRequestConditions requestConditions, boolean rangeGetContentMd5, Context context) {
-        BlobRange rangeReal = range == null ? new BlobRange(0) : range;
+        BlobRange finalRange = range == null ? new BlobRange(0) : range;
         final ParallelTransferOptions finalParallelTransferOptions =
             ModelHelper.populateAndApplyDefaults(parallelTransferOptions);
-        BlobRequestConditions conditionsReal = requestConditions == null
+        BlobRequestConditions finalConditions = requestConditions == null
             ? new BlobRequestConditions() : requestConditions;
 
         AsynchronousFileChannel channel = downloadToFileResourceSupplier(filePath);
         return Mono.just(channel)
-            .flatMap(c -> this.downloadToFileImpl(c, rangeReal, finalParallelTransferOptions,
-                downloadRetryOptions, conditionsReal, rangeGetContentMd5, context))
+            .flatMap(c -> this.downloadToFileImpl(c, finalRange, finalParallelTransferOptions,
+                downloadRetryOptions, finalConditions, rangeGetContentMd5, context))
             .doFinally(signalType -> this.downloadToFileCleanup(channel, filePath, signalType));
     }
 
@@ -745,7 +745,7 @@ public class BlobAsyncClientBase {
         }
     }
 
-    private Mono<Response<BlobProperties>> downloadToFileImpl(AsynchronousFileChannel file, BlobRange rangeReal,
+    private Mono<Response<BlobProperties>> downloadToFileImpl(AsynchronousFileChannel file, BlobRange finalRange,
         ParallelTransferOptions finalParallelTransferOptions, DownloadRetryOptions downloadRetryOptions,
         BlobRequestConditions requestConditions, boolean rangeGetContentMd5, Context context) {
         // See ProgressReporter for an explanation on why this lock is necessary and why we use AtomicLong.
@@ -755,11 +755,11 @@ public class BlobAsyncClientBase {
         /*
          * Downloads the first chunk and gets the size of the data and etag if not specified by the user.
          */
-       return getSetupMono(rangeReal, finalParallelTransferOptions, downloadRetryOptions, requestConditions,
+        return getSetupMono(finalRange, finalParallelTransferOptions, downloadRetryOptions, requestConditions,
            rangeGetContentMd5, context)
         .flatMap(setupTuple3 -> {
             long newCount = setupTuple3.getT1();
-            BlobRequestConditions realConditions = setupTuple3.getT2();
+            BlobRequestConditions finalConditions = setupTuple3.getT2();
 
             int numChunks = calculateNumBlocks(newCount, finalParallelTransferOptions.getBlockSize());
 
@@ -779,12 +779,12 @@ public class BlobAsyncClientBase {
                     long chunkSizeActual = Math.min(finalParallelTransferOptions.getBlockSize(),
                         newCount - (chunkNum.longValue() * finalParallelTransferOptions.getBlockSize().longValue()));
                     BlobRange chunkRange = new BlobRange(
-                        rangeReal.getOffset() +
-                            (chunkNum.longValue() * finalParallelTransferOptions.getBlockSize().longValue()),
+                        finalRange.getOffset()
+                            + (chunkNum.longValue() * finalParallelTransferOptions.getBlockSize().longValue()),
                         chunkSizeActual);
 
                     // Make the download call.
-                    return this.downloadWithResponse(chunkRange, downloadRetryOptions, realConditions,
+                    return this.downloadWithResponse(chunkRange, downloadRetryOptions, finalConditions,
                         rangeGetContentMd5, null)
                         .subscribeOn(Schedulers.elastic())
                         .flatMap(response ->
@@ -885,8 +885,8 @@ public class BlobAsyncClientBase {
             .setLeaseId(accessConditions.getLeaseId());
     }
 
-    private static Mono<Void> writeBodyToFile(BlobDownloadAsyncResponse response,
-        AsynchronousFileChannel file, long chunkNum, ParallelTransferOptions optionsReal, Lock progressLock,
+    private static Mono<Void> writeBodyToFile(BlobDownloadAsyncResponse response, AsynchronousFileChannel file,
+        long chunkNum, ParallelTransferOptions finalParallelTransferOptions, Lock progressLock,
         AtomicLong totalProgress) {
 
         // Extract the body.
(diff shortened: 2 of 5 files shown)


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
```diff
@@ -49,15 +46,18 @@ private HttpResponse scrubETagHeader(HttpResponse unprocessedResponse) {
...
+        HttpHeaders headers = unprocessedResponse.getHeaders();
         headers.put(eTagHeader.getName(), eTag);
-        innerHttpResponse = unprocessedResponse;
-        return new InnerHttpResponse(unprocessedResponse.getRequest());
+        return new InnerHttpResponse(unprocessedResponse, headers, unprocessedResponse.getRequest());
     }
 
     private final class InnerHttpResponse extends HttpResponse {
-        protected InnerHttpResponse(HttpRequest request) {
+        private final HttpResponse innerHttpResponse;
```
rickle-msft: @sima-zhu These are the changes I was discussing with you in Teams. Please let me know if you have concerns about changing the scope of these variables.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -767,29 +745,201 @@ private AsynchronousFileChannel downloadToFileResourceSupplier(String filePath)
...
+        return this.downloadWithResponse(new BlobRange(r.getOffset(), initialChunkSize), downloadRetryOptions,
+            requestConditions, rangeGetContentMd5, context)
+            .subscribeOn(Schedulers.elastic())
+            .flatMap(response -> {
+                /*
+                Either the etag was set and it matches because the download succeeded, so this is a no-op, or there
+                was no etag, so we set it here. ETag locking is vital to ensure we download one, consistent view
+                of the file.
+                 */
+                BlobRequestConditions newConditions = setEtag(requestConditions, response.getHeaders()
```
alzimmermsft: Could get the ETag through the deserializedHeaders on the response object instead of accessing the raw header map, Same goes for the other places the raw headers map is being used.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -767,29 +745,201 @@ private AsynchronousFileChannel downloadToFileResourceSupplier(String filePath)
...
-            if (pos + count > offset + length) {
-                count = offset + length - pos;
+        return numBlocks;
+    }
+
+    /*
+    Download the first chunk. Construct a Mono which will emit the total count for calculating the number of chunks,
+    access conditions containing the etag to lock on, and the response from downloading the first chunk.
+     */
+    private Mono<Tuple3<Long, BlobRequestConditions, BlobDownloadAsyncResponse>> getSetupMono(BlobRange r,
```
alzimmermsft: Could we use full names for parameters, `range` instead of `r` and `options` instead of `o`? It'll make reading the code later a little less cryptic.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -767,29 +745,201 @@ private AsynchronousFileChannel downloadToFileResourceSupplier(String filePath)
...
+       return getSetupMono(rangeReal, finalParallelTransferOptions, downloadRetryOptions, requestConditions,
+           rangeGetContentMd5, context)
+        .flatMap(setupTuple3 -> {
+            long newCount = setupTuple3.getT1();
+            BlobRequestConditions realConditions = setupTuple3.getT2();
+
+            int numChunks = calculateNumBlocks(newCount, finalParallelTransferOptions.getBlockSize());
+
+            // In case it is an empty blob, this ensures we still actually perform a download operation.
+            numChunks = numChunks == 0 ? 1 : numChunks;
```
alzimmermsft: For the case when the blob is empty could we return `Mono.just(writeBodyToFile(initialResponse, file, 0, options, lock, progess)`?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -767,29 +745,201 @@ private AsynchronousFileChannel downloadToFileResourceSupplier(String filePath)
...
+                accessConditions.getIfModifiedSince())
+            .setIfUnmodifiedSince(
+                accessConditions.getIfModifiedSince())
+            .setIfMatch(etag)
+            .setIfNoneMatch(
+                accessConditions.getIfNoneMatch())
+            .setLeaseId(accessConditions.getLeaseId());
+    }
+
+    private static Mono<Response<BlobProperties>> writeBodyToFile(BlobDownloadAsyncResponse response,
```
alzimmermsft: Could we change this to be `Mono<Void>` and change the `response` to `Flux<ByteBuffer> data` and then in the stream that calls into this instead of using `.last()` we use `thenReturn(<new method that creates the response>)`, that way we aren't constructing a new response object during every write being performed.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -767,29 +745,201 @@ private AsynchronousFileChannel downloadToFileResourceSupplier(String filePath)
...
+            BlobRequestConditions realConditions = setupTuple3.getT2();
+
+            int numChunks = calculateNumBlocks(newCount, finalParallelTransferOptions.getBlockSize());
+
+            // In case it is an empty blob, this ensures we still actually perform a download operation.
+            numChunks = numChunks == 0 ? 1 : numChunks;
+
+            BlobDownloadAsyncResponse initialResponse = setupTuple3.getT3();
+            return Flux.range(0, numChunks)
+                .flatMap(chunkNum -> {
```
alzimmermsft: In the future this is a spot where we could use a `ParallelFlux` to parallelize the operation, for now this is good.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -767,29 +745,201 @@ private AsynchronousFileChannel downloadToFileResourceSupplier(String filePath)
...
+    }
+
+    private static long extractTotalBlobLength(String contentRange) {
+        return Long.parseLong(contentRange.split("/")[1]);
+    }
+
+    private void downloadToFileCleanup(AsynchronousFileChannel channel, String filePath, SignalType signalType) {
+        try {
+            channel.close();
+            if (!signalType.equals(SignalType.ON_COMPLETE)) {
```
alzimmermsft: This is a safeguard in case the operation fails and the user won't have a potentially corrupted file, correct? If so, would we want to log a message here stating that the operation wasn't able to successfully complete, cleaning up resources.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -767,29 +745,201 @@ private AsynchronousFileChannel downloadToFileResourceSupplier(String filePath)
...
+                If the user either didn't specify a count or they specified a count greater than the size of the
+                remaining data, take the size of the remaining data. This is to prevent the case where the count
+                is much much larger than the size of the blob and we could try to download at an invalid offset.
+                 */
+                long newCount = r.getCount() == null || r.getCount() > (totalLength - r.getOffset())
+                    ? totalLength - r.getOffset() : r.getCount();
+
+                return Mono.zip(Mono.just(newCount), Mono.just(newConditions), Mono.just(response));
+            })
+            .onErrorResume(throwable -> {
```
alzimmermsft: Give that we should only be resuming if the exception is a `BlobStorageException` could we use the `onErrorResume` that takes a `Class<E>` to simplify the check logic so we don't need to verify instance and perform casting?

https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#onErrorResume-java.lang.Class-java.util.function.Function-

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
rickle-msft: The one place I can't do this is when I'm checking the range after catching a StorageException. That one just has a generic response, so I have to go into the raw headers.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
rickle-msft: I think that should be functionally equivalent. Is this a code style suggestion you think will make the logic clearer?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
rickle-msft: I didn't use that overload because I need additional logic on top of just validating the type anyway, but if you think it'll be cleaner, I can change it

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ScrubEtagPolicy.java
rickle-msft: Sima mentioned offline she is fine with these changes.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
rickle-msft: Could you explain this a bit more for my own edification? It's been my understanding in the past that we haven't needed to use operators like .parallel(), etc. because core (and what was then called the runtime) sits on top of an asynchronous event loop, so the level of the clients gets that for free when using flatMap and other async operators. We also don't seem to use ParallelFlux anywhere else in the codebase?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
rickle-msft: But I don't actually think this will clean anything up. I still need the special case for the first element to write the body to the file without calling download. I think this is the cleanest way to get the empty file case into the rest of the workflow.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
alzimmermsft: Under the hood the method is the same as the predicate version where the predicate is `Class.isInstance(exception)` then the exception passed in is the type of the class.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -701,97 +703,242 @@ public boolean isSnapshot() {
...
+        data = ProgressReporter.addParallelProgressReporting(data,
+            optionsReal.getProgressReceiver(), progressLock, totalProgress);
+
+        // Write to the file.
+        return FluxUtil.writeFile(data, file,
+            chunkNum * optionsReal.getBlockSize());
+    }
+
+    private static Response<BlobProperties> buildBlobPropertiesResponse(BlobDownloadAsyncResponse response) {
+        BlobProperties properties = new BlobProperties(null, response.getDeserializedHeaders().getLastModified(),
```
gapra-msft: nit: this could be slightly less verbose by making this method return just BlobProperties and only accept the headers as an argument. and wherever you call it you can essentially just say  new SimpleResponse<>(response, buildBlobProperties(response.getDeserializedHeaders()));

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -701,97 +703,242 @@ public boolean isSnapshot() {
...
-        final ParallelTransferOptions finalParallelTransferOptions = parallelTransferOptions == null
-            ? new ParallelTransferOptions(null, null, null)
-            : new ParallelTransferOptions(parallelTransferOptions.getBlockSize(),
-                parallelTransferOptions.getNumBuffers(), parallelTransferOptions.getProgressReceiver());
+        ParallelTransferOptions parallelTransferOptions, DownloadRetryOptions downloadRetryOptions,
+        BlobRequestConditions requestConditions, boolean rangeGetContentMd5, Context context) {
+        BlobRange rangeReal = range == null ? new BlobRange(0) : range;
+        final ParallelTransferOptions finalParallelTransferOptions =
+            ModelHelper.populateAndApplyDefaults(parallelTransferOptions);
+        BlobRequestConditions conditionsReal = requestConditions == null
```
gapra-msft: Could we go with the name finalRequestConditions or something? conditionsReal sounds a little odd

gapra-msft: same with rangeReal

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -597,9 +595,9 @@ private String getBlockID() {
         return Base64.getEncoder().encodeToString(UUID.randomUUID().toString().getBytes(StandardCharsets.UTF_8));
     }
 
-    private List<BlobRange> sliceFile(long fileSize, int blockSize) {
+    private List<BlobRange> sliceFile(long fileSize, int blockSize, boolean enableHtbbOptimization) {
```
jaschrep-msft: It looks as if this boolean is only ever decided based on the original `parallelTransferOptions.getBlockSize()` existing. But then we pass in the blocksize that is either the original or the default if there was no original, and then this boolean to double check whether the number was originally specified. It sounds confusing just trying to describe it.

I'd have populate/apply defaults contain the htbb optimizations if possible. That way, we know the block size we have was already determined correctly.

If that's not possible, then we should at least pass in the old block size without this extra boolean so it's at least all in one place how we're determining this. Perhaps slicing the file BEFORE applying defaults would help communicate that logic better.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BuilderHelper.java
```diff
@@ -119,4 +119,6 @@ private static HttpPipelinePolicy getResponseValidationPolicy() {
             .addOptionalEcho(Constants.HeaderConstants.ENCRYPTION_KEY_SHA256)
             .build();
     }
+
+
```
jaschrep-msft: Revert

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
rickle-msft: I only ever call this once. It's just to hide this ugly constructor block. I think it's more concise constructing the response in the method cus then I can hide the response construction, too.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
rickle-msft: I think this is actually the best place for checking the optimization. Not every method that needs to call populateAndApplyDefaults can enable this optimization (buffered upload, for example), so I think it should be local to uploadFromFile at least until there's a good reason to move it. 

We can't slice the file before applying the defaults because the slicing is based on the chunk size, be that the default or the htbb optimization size or the value passed by the customer. 

The best compromise I could come up with is passing the original options. Let me know what you think.

jaschrep-msft: I think this is a fine way to do it. I was mostly concerned about readablilty. But if this is a condition we don't always want to check, I think this is fine.

PR #6144, opened 2019-11-01
New DocIndex

JimSuplizio:
Add the code to create the new DocIndex. An example of what this looks like can be found [here](https://jimsuplizio.github.io/azure-sdk-for-java/). This new DocIndex handles versioned docs. The doc, themselves, are stored in azure blob storage (I'm still doing the initial population so only the index is there right now, it'll be about 30 minutes or so). 

It's also worth noting that while there are changes to pom.client.xml and pom.data.xml these changes will NOT cause us to have to republish them. The reason for this change was to pipe the version into the script that generates the additional overview from the readme so we can have the version on the landing page with a link back to the index.

### DIFF
--- a/eng/pipelines/docindex.yml
+++ b/eng/pipelines/docindex.yml
@@ -48,6 +48,14 @@ jobs:
           # the service name in all caps
           $serviceMapHash = Get-Content -Path "$(Build.SourcesDirectory)/eng/docgeneration/service-mapper.json" | ConvertFrom-Json -AsHashtable
 
+          # There are some artifact that show up, due to the way we do discovery, that are never shipped.
+          # Keep a list of those here and anything we don't want to ship can be added to here which will
+          # cause them to get skipped when generating the DocIndex
+          $ArtifactsToSkip = (
+            'azure-cosmos-benchmark',
+            'azure-sdk-template'
+          )
+
           # The list of services is being constructed from the directory list under the sdk folder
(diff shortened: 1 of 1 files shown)


eng/pipelines/docindex.yml
```diff
@@ -0,0 +1,189 @@
...
+          overWrite: true
+      - pwsh: ls
+        workingDirectory: $(Build.SourcesDirectory)/docfx_project/
+      - pwsh: ls
+        workingDirectory: $(Build.SourcesDirectory)/docfx_project/templates
+      - pwsh: |
+          Copy-Item "$(Build.SourcesDirectory)/eng/docgeneration/docfx.json" -Destination "$(Build.SourcesDirectory)/docfx_project/" -Force
+        displayName: Copy over docfx.json
+      - pwsh: |
+
```
scbedd: I feel like this should be an external powershell hosted in the `eng/` directory or the like. Maybe `eng/docgeneration`. 

However, I think that's the case for this entire thing, and both `JS` and `.NET` have the same problem. So I don't want to block this checkin based on this minor issue.

JimSuplizio: @scbedd That's a fair enough request. I'll rework this when you file the issue.

JimSuplizio:
Because the azp run commands weren't working here I'd ended up kicking off a run through the pipeline, itself. Since that passed I'm going to check this in.

PR #6147, opened 2019-11-02
Update GA Versions

JimSuplizio:
Updated the post release versions. 
- GA versions had their minor incremented and -preview.1 added to their versions
- Preview versions were incremented
- com.azure:azure-storage-file-datalake was not incremented as it has not been released at it's current preview version. This may be released before this PR is approved, if so this PR will be updated.

Made one minor tweak to the update_versions.py script. When updating versions in java source files, verify the files exists and raise a FileNotFoundError if it doesn't.


eng/versioning/version_client.txt
```diff
@@ -3,26 +3,26 @@
...
-com.azure:azure-core-management;1.0.0-preview.7;1.0.0-preview.7
-com.azure:azure-core-test;1.0.0;1.0.0
-com.azure:azure-data-appconfiguration;1.0.0-preview.6;1.0.0-preview.6
-com.azure:azure-identity;1.0.0;1.0.0
-com.azure:azure-security-keyvault-certificates;4.0.0-preview.5;4.0.0-preview.5
-com.azure:azure-security-keyvault-keys;4.0.0;4.0.0
-com.azure:azure-security-keyvault-secrets;4.0.0;4.0.0
-com.azure:azure-messaging-eventhubs;5.0.0-preview.5;5.0.0-preview.5
-com.azure:azure-messaging-eventhubs-checkpointstore-blob;1.0.0-preview.3;1.0.0-preview.3
+com.azure:azure-core;1.1.0-preview.1;1.1.0-preview.1
```
weshaggard: Ideally we wouldn't bump the dependency version here and keep it at GA. I release we still need to module references for now but if we can avoid bumping the dependency versions we should do so.

PR #6151, opened 2019-11-04
Clean-up unified pipelines.

mitchdenny:
This PR removes the client.yml pipeline that was in use before unified pipelines was rolled out. In preparation for this PR I have disabled the ```java - client - ci``` pipeline and ```java - client``` pipelines and moved them into a ```legacy``` folder in Azure Pipelines.

I've also moved the old release pipeline into a legacy folder as well. With the change to build all of track 2 libraries via the core pipeline I believe that the old java pipelines are now fully redundant. Merging this would put us in a consistent state.

/cc @JonathanGiles

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -268,3 +268,28 @@ jobs:
         inputs:
           mergeTestResults: true
           testRunTitle: '$(OSName) on Java $(JavaVersion)'
+
+      - task: PublishTestResults@2
+        condition: succeededOrFailed()
+        inputs:
+          mergeTestResults: true
+          testRunTitle: '$(OSName) on Java $(JavaVersion)'
+
+      # Azure DevOps only seems to respect the last code coverage result published, so only do this for Linux + Java LTS.
+      # Code coverage reporting is setup only for Track 2 modules.
+      - task: PublishCodeCoverageResults@1
+        condition: and(eq(variables['OsName'], 'Linux'), eq(variables['JavaVersion'], '1.11'))
+        inputs:
+          codeCoverageTool: JaCoCo
+          summaryFileLocation: eng/jacoco-test-coverage/target/site/test-coverage/jacoco.xml
+          reportDirectory: eng/jacoco-test-coverage/target/site/test-coverage/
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -268,3 +268,28 @@ jobs:
         inputs:
           mergeTestResults: true
           testRunTitle: '$(OSName) on Java $(JavaVersion)'
+
+      - task: PublishTestResults@2
```
srnagar: With the latest changes to JaCoCo configuration, we need one additional step - `Generate aggregate code coverage report` that's added in `client.yml`. We don't need the coverage report generated for each PR, so, when you add this step, include a condition to run it only in a non-PR build.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -268,3 +268,28 @@ jobs:
...
+        condition: and(eq(variables['OsName'], 'Linux'), eq(variables['JavaVersion'], '1.11'))
+        inputs:
+          codeCoverageTool: JaCoCo
+          summaryFileLocation: eng/jacoco-test-coverage/target/site/test-coverage/jacoco.xml
+          reportDirectory: eng/jacoco-test-coverage/target/site/test-coverage/
+          failIfCoverageEmpty: true
+
+      # Publish code coverage to an artifact so it can be uploaded to the Maven site.
+      # Do this for track 2 modules only. Code coverage from previous step is only published for Linux + Java LTS.
+      - task: PublishPipelineArtifact@0
```
srnagar: There's another release pipeline that looks for successful runs of `java - client` pipeline to copy the artifacts to `azuresdkartifacts` blob storage. Can you please update that pipeline to check for successful completion of this pipeline instead?

mitchdenny: We will need to bring the functionality from that pipeline and put it into the unified pipeline rather than have it trigger off the completion of the unified pipeline. Pretty much everything that used to run across the whole repo needs to evolve to just work against the subset of artifacts produced by the pipeline. So if we are looking at the eventhubs pipeline, we only publish stuff related to the eventhubs pipeline.

PR #6155, opened 2019-11-04
Update GA versions to prep for hotfix

JimSuplizio:
Update GA versions to prep for hotfix


eng/versioning/version_client.txt
```diff
@@ -3,26 +3,26 @@
 
 com.azure:azure-sdk-parent;1.5.0;1.5.0
 com.azure:azure-client-sdk-parent;1.6.0;1.6.0
-com.azure:azure-core;1.0.0;1.0.0
```
weshaggard: Lets try to avoid bumping the dependency versions here as that will force us to ship everything which I'm not sure if we want to do that.

JimSuplizio: Sounds good. The second commit to this PR does what you requested.

PR #6156, opened 2019-11-04
Support Azurite and Test Simple Scenarios

### LINKED ISSUE #6140: [BUG] Can not run dev tests with Azurite
**Describe the bug**
Can not use Azurite for local development because the request urls are constructed in the wrong way. 

local: 
```
Exception in thread "main" com.azure.storage.blob.models.BlobStorageException: Status code 400, (empty body)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
```

azurite: 
```
172.17.0.1 - - [01/Nov/2019:16:05:52 +0000] "GET /mycontainer?restype=container HTTP/1.1" 400 -

info: BlobStorageContextMiddleware: RequestMethod=GET RequestURL=http://127.0.0.1/mycontainer?restype=container RequestHeaders:{"host":"127.0.0.1:10000","accept":"*/*","date":"Fri, 01 Nov 2019 16:05:52 GMT","authorization":"SharedKey devstoreaccount1:RPl/IHhGATnQupcOZ2wI6JmI1CJpmnmCp/XY49Amx5M=","x-ms-version":"2019-02-02","x-ms-client-request-id":"c3b4c250-e4cb-4534-8856-e9928acf07f8","user-agent":"azsdk-java-azure-storage-blob/12.0.0 11.0.4; Mac OS X 10.15","content-length":"0"} ClientIP=172.17.0.1 Protocol=http HTTPVersion=1.1
info: BlobStorageContextMiddleware: Account=mycontainer Container=undefined Blob=
verbose: DispatchMiddleware: Dispatching request...
error: DispatchMiddleware: Incoming URL doesn't match any of swagger defined request patterns.
```

**To Reproduce**
```
docker run -p 10000:10000 mcr.microsoft.com/azure-storage/azurite azurite-blob  -l /data --blobHost 0.0.0.0 -d debug.log
```
pom.xml: 
```
<dependency>
            <groupId>com.azure</groupId>
            <version>12.0.0</version>
        </dependency>
```
(`azure-storage-blob` was built locally (commit `bbf1cbd`))

***Code Snippet***
```
import com.azure.storage.blob.BlobContainerClient;
import com.azure.storage.blob.BlobServiceClientBuilder;

public class LocalAzurite {
    public static void main(String[] args) throws NoSuchFieldException {

        BlobContainerClient container = new BlobServiceClientBuilder()
                .connectionString("UseDevelopmentStorage=true;")
                .buildClient()
                .getBlobContainerClient("mycontainer");

        if (!container.exists()) container.create();

        System.out.println("exists: " + container.exists());
        System.out.println(container.getBlobClient("myblob").getBlobUrl());
    }
}
```

**Expected behavior**
To create the blob without error. 

 - OS:  macOS 10.15, Oracle JDK 11
 - IDE : IntelliJ
 - Version of the Library used: current master branch (commit `bbf1cbd`)

**Additional context**
For Azurite to work the urls have to be in the format: 
    ```http://127.0.0.1:10000/<account-name>/<resource-path>```

But the produced url is 
    http://127.0.0.1:10000/mycontainer
instead of 
    http://127.0.0.1:10000/devstoreaccount1/mycontainer

I noticed there is support for the emulator: StorageConnectionString#create:  https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/connectionstring/StorageConnectionString.java#L106

However the endpoint is later shortened, so that the account-name is not in the path any more: 
SpecializedBlobClientBuilder#getUrl:  https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/SpecializedBlobClientBuilder.java#L213

1. Is there support to use Azurite with `azure-storage-blob`?
2. How can they be used together?

alzimmermsft:
Fixes #6140 

Fixes issue when constructing a client using an IP based URL (http://127.0.0.1:10000/devstoreaccount1) where the account name was lost from the URL path. Unit tests were added to verify that the endpoint is constructed properly when an IP-based URL is passed and in addition to those a few tests verifying basic service calls to Azurite.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AzuriteTest.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AzuriteTest.groovy
@@ -224,10 +224,7 @@ class AzuriteTest extends APISpec {
 
     def "Connect to Azurite"() {
         setup:
-        def serviceClient = new BlobServiceClientBuilder()
-            .endpoint(azuriteEndpoint)
-            .credential(azuriteCredential)
-            .buildClient()
+        def serviceClient = getAzuriteServiceClient()
 
         when:
         serviceClient.getProperties()
--- a/sdk/storage/azure-storage-blob/src/test/resources/session-records/AzuriteTestazuriteurlconstructblobclient.json
+++ b/sdk/storage/azure-storage-blob/src/test/resources/session-records/AzuriteTestazuriteurlconstructblobclient.json
@@ -1,23 +1,23 @@
 {
   "networkCallRecords" : [ {
     "Method" : "PUT",
-    "Uri" : "https://azstoragesdkaccount.blob.core.windows.net/jtcazuriteurlconstructblobclient085463f573f96cf5a4?restype=container",
+    "Uri" : "https://azstoragesdkaccount.blob.core.windows.net/jtcazuriteurlconstructblobclient089849f8c7b6bb14f9?restype=container",
     "Headers" : {
       "x-ms-version" : "2019-02-02",
       "User-Agent" : "azsdk-java-azure-storage-blob/12.1.0-preview.1 11.0.4; Windows 10 10.0",
-      "x-ms-client-request-id" : "c305280e-3a42-4124-aafc-43ac8a83043a"
+      "x-ms-client-request-id" : "9e8a99b1-a66e-4033-834e-b9384e5f1b08"
     },
     "Response" : {
       "x-ms-version" : "2019-02-02",
       "Server" : "Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0",
-      "ETag" : "0x8D7618EB883294B",
-      "Last-Modified" : "Tue, 05 Nov 2019 01:23:10 GMT",
+      "ETag" : "0x8D761B5AEB42980",
+      "Last-Modified" : "Tue, 05 Nov 2019 06:02:04 GMT",
       "retry-after" : "0",
       "Content-Length" : "0",
       "StatusCode" : "201",
-      "x-ms-request-id" : "b3cab20f-901e-0077-7877-93b537000000",
-      "Date" : "Tue, 05 Nov 2019 01:23:09 GMT",
-      "x-ms-client-request-id" : "c305280e-3a42-4124-aafc-43ac8a83043a"
+      "x-ms-request-id" : "bf176165-101e-0029-7c9e-934634000000",
+      "Date" : "Tue, 05 Nov 2019 06:02:03 GMT",
+      "x-ms-client-request-id" : "9e8a99b1-a66e-4033-834e-b9384e5f1b08"
     },
     "Exception" : null
   }, {
@@ -26,7 +26,7 @@
     "Headers" : {
       "x-ms-version" : "2019-02-02",
       "User-Agent" : "azsdk-java-azure-storage-blob/12.1.0-preview.1 11.0.4; Windows 10 10.0",
-      "x-ms-client-request-id" : "0083a3b4-6593-44ad-82d3-f78f76be42cf"
+      "x-ms-client-request-id" : "9a693b2b-8d9c-4564-b454-2c0eb95e437f"
     },
     "Response" : {
       "Transfer-Encoding" : "chunked",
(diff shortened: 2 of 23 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
```diff
@@ -296,7 +293,7 @@ public static BlobUrlParts parse(String url) {
     public static BlobUrlParts parse(URL url) {
         BlobUrlParts parts = new BlobUrlParts().setScheme(url.getProtocol());
 
-        if (IP_V4_URL_PATTERN.matcher(url.getHost()).find()) {
+        if (ModelHelper.IP_V4_URL_PATTERN.matcher(url.getHost()).find()) {
```
rickle-msft: I think the toUrl method may also need to be updated. It looks like it always sets the container right after the host and never considers if it needs to set the account name?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
```diff
@@ -17,17 +18,13 @@
...
  * existing URL into its parts with the {@link #parse(URL)} class. You may construct a URL from parts by calling {@link
  * #toUrl()}.
  */
 public final class BlobUrlParts {
-    private static final Pattern IP_V4_URL_PATTERN = Pattern
-        .compile("(?:\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3})|(?:localhost)");
-
     private final ClientLogger logger = new ClientLogger(BlobUrlParts.class);
 
     private String scheme;
```
rickle-msft: I think the docs for setHost() should be updated to give an example of ip style

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
alzimmermsft: Updated

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
alzimmermsft: Added an example of the IP style URL Azurite expects.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/AzuriteTest.groovy
```diff
@@ -0,0 +1,226 @@
...
+import com.azure.storage.blob.models.BlobStorageException
+import com.azure.storage.blob.specialized.BlobClientBase
+import com.azure.storage.blob.specialized.BlobLeaseClientBuilder
+import com.azure.storage.blob.specialized.SpecializedBlobClientBuilder
+import com.azure.storage.common.StorageSharedKeyCredential
+import spock.lang.Unroll
+
+class AzuriteTest extends APISpec {
+    String azuriteEndpoint = "http://127.0.0.1:10000/devstoreaccount1"
+    StorageSharedKeyCredential azuriteCredential = new StorageSharedKeyCredential("devstoreaccount1", "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==")
```
rickle-msft: If this is a real key, it needs to be in an environment variable and should be rotated. If it's some fake key value, can you please add a comment saying so.

alzimmermsft: It's a fake key, will add documentation that it is.

PR #6168, opened 2019-11-05
Test Enhancements and General Fixes

### LINKED ISSUE #6192: Investigate Live Test Timeouts
The nightly live tests have been timing out for a few days, determine the root cause issue.

alzimmermsft:
Fixes #6192, #6193, and #6194

This PR performs a few general fixes and adds stability to a few tests.

- Updated the handling of SAS tokens to strip any leading `?` as it wasn't needed when being applied to the request URL. (#6193)
- Changed the prefetch limit on a windowing function to reduce the likelihood of the window hitting a terminal state before a path was determined. (#6192)

- Cleaned up duplicate testing methods into a centralized method.
- Updated file comparison method to use `AsynchronousFileChannel` as it has shown to be more consistent when running the tests. (#6914 - more robust checking)
- Removed Content-MD5 from an encryption test as the MD5 represented the non-encrypted data and was leading to the service returning an invalid Content-MD5 exception. (#6194 - Content-MD5 calculated from unencrypted blob)
- Fixed test that was missing a parameter. (#6194 - Missing parameter in call)
- Updating batching test to no longer create resources that weren't required. (#6194 - Unnecessary resource allocation)


amishra-dev:
@alzimmermsft is this PR in response to some issue that was filed? If not, can you please file an issue for this. 
Also if there are more such PR's that you intend to send out can you please make sure that the issues are filed now so that we can triage it?
/cc @joshfree @rickle-msft

alzimmermsft:
> 
> 
> @alzimmermsft is this PR in response to some issue that was filed? If not, can you please file an issue for this.
> Also if there are more such PR's that you intend to send out can you please make sure that the issues are filed now so that we can triage it?
> /cc @joshfree @rickle-msft

Added issues for the problems fixed.

Issue #6025, opened 2019-10-24
[BUG] User Agent Missing Parentheses around Comment Section


mrm9084:
User Agent shows as 
azsdk-java-application-configuration/1.0.0-preview.4 1.8.0_212; Linux 4.4.0-128-generic

where it should be
azsdk-java-application-configuration/1.0.0-preview.4 (1.8.0_212; Linux 4.4.0-128-generic)

See RFC: https://tools.ietf.org/html/rfc7231#section-5.5.3
I find the example here a bit clearer: https://www-archive.mozilla.org/build/user-agent-strings.html

PR #6172, opened 2019-11-05
fix the missing parethese: <platform_info>: information about the cur…

mssfang:
From the given General Guidelines for User-Agent Header:

<platform_info>: information about the currently executing language runtime and OS, e.g. “(NODE-VERSION v4.5.0; Windows_NT 10.0.14393)”

https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy

fixes: #6025


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -21,7 +21,7 @@
 
     // From the design guidelines, the default user agent header format is:
     // azsdk-java-<client_lib>/<sdk_version> <platform_info>
-    private static final String USER_AGENT_FORMAT = DEFAULT_USER_AGENT_HEADER + "-%s/%s %s";
+    private static final String USER_AGENT_FORMAT = DEFAULT_USER_AGENT_HEADER + "-%s/%s (%s)";
```
srnagar: Can you please add a unit test to ensure that the generated string is as expected and future changes don't accidentally break this format?

mssfang: updated

Issue #3169, opened 2019-03-21
Fix Dependencies Missing from DependencyManagement


bsiegel:
Based on the report, we seem to have many cases where we have dependencies in use that are not managed by the Parent POM dependencyManagement section. We should resolve these, and then modify the dependency checker tool to emit a nonzero exit code when any inconsistency is detected.

kurtzeborn:
Considering the work that @mitchdenny and @JimSuplizio have done on this, I think this issue is obsolete.

JimSuplizio:
@kurtzeborn @bsiegel @mitchdenny 
Yes, with the centralization of external dependencies and removal/disallowing of dependencyManagement sections in POM files, the dependencies are in a much better state.

PR #6176, opened 2019-11-06
Reduce analyse CI time

### LINKED ISSUE #6128: Remove site:site site:stage From PR Validation
Look into replacing this step with modular versions of the step.

- Checkstyle linting
- SpotBugs linting
- Javadoc generation

This step is a common failure but it takes upwards of 20-30 minutes to run and generally when it fails there is no explanation of why it failed just what failed (ex. 10 checkstyle errors). This not only slows down PR validation but slows down the development cycle.

JonathanGiles:
Making the site:site site:stage step conditional on not being a nightly build, same with the spotbugs aggregate report. Adding JavaDoc to the analyse step.

Will do a few updates to this PR to test that the new setup catches issues, so it may be a bit noisy to start with.

Fixes #6128

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -157,6 +157,19 @@ jobs:
           publishJUnitResults: false
           goals: 'spotbugs:check checkstyle:checkstyle-aggregate javadoc:aggregate'
 
+      # We run a separate SpotBugs aggregate report step here to roll-up all the issues identified per-module
+      - task: Maven@3
+        displayName: 'Run SpotBugs aggregate analysis'
+        inputs:
+          mavenPomFile: eng/spotbugs-aggregate-report/pom.xml
+          options: '--batch-mode -DskipTests -Dgpg.skip'
+          mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
+          javaHomeOption: 'JDKVersion'
+          jdkVersionOption: '1.11'
+          jdkArchitectureOption: 'x64'
+          publishJUnitResults: false
+          goals: 'install'
+
       # Generate the Maven site report, including SpotBugs, JavaDoc, CheckStyle, Jacoco, etc only when we are not
       # running as a PR check.
       - task: Maven@3
@@ -171,20 +184,6 @@ jobs:
           jdkArchitectureOption: 'x64'
           publishJUnitResults: false
           goals: 'install site:site site:stage'
-
-      # We run a separate SpotBugs aggregate report step here to roll-up all the issues identified per-module in the
-      # previous step.
-      - task: Maven@3
-        displayName: 'Run SpotBugs aggregate analysis'
-        inputs:
-          mavenPomFile: eng/spotbugs-aggregate-report/pom.xml
-          options: '--batch-mode -DskipTests -Dgpg.skip'
-          mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
-          javaHomeOption: 'JDKVersion'
-          jdkVersionOption: '1.11'
-          jdkArchitectureOption: 'x64'
-          publishJUnitResults: false
-          goals: 'install'
 
       - script: |
           git clone https://github.com/JonathanGiles/DependencyChecker.git
--- a/sdk/core/azure-core/src/samples/java/com/azure/core/http/rest/PagedFluxJavaDocCodeSnippets.java
+++ b/sdk/core/azure-core/src/samples/java/com/azure/core/http/rest/PagedFluxJavaDocCodeSnippets.java
@@ -15,8 +15,7 @@ import java.util.function.Supplier;
  * Code snippets for {@link PagedFlux}
  */
 public final class PagedFluxJavaDocCodeSnippets {
-
-
+    
     /**
      * Code snippets for showing usage of {@link PagedFlux} in class docs
      */


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -184,6 +185,7 @@ jobs:
           jdkArchitectureOption: 'x64'
           publishJUnitResults: false
           goals: 'install'
+          condition: ne(variables['Build.Reason'], 'PullRequest')
```
alzimmermsft: I think we should continue running the SpotBugs Aggregate step, I've generally seen it catch things that the SpotBugs and Checkstyle step doesn't (usually the same issues that the site:site site:stage step catches).

srnagar: The reason for this is because ` 'Run SpotBugs, CheckStyle, and JavaDoc'` step doesn't include non-shipping-modules. If we add `-Dinclude-non-shipping-modules` to `options` in that step, then this step won't be required.

This step uses `eng/spotbugs-aggregate-reports/pom.xml` which is not included if the above flag is not set.

JonathanGiles: Does this force us to also have to run 'install' on the earlier step?

alzimmermsft: Before installing the clients the shipping modules are installed, should be a safe change to make.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -155,12 +155,13 @@ jobs:
           jdkVersionOption: '1.11'
           jdkArchitectureOption: 'x64'
           publishJUnitResults: false
-          goals: 'spotbugs:check checkstyle:checkstyle-aggregate'
+          goals: 'spotbugs:check checkstyle:checkstyle-aggregate javadoc:aggregate'
```
JimSuplizio: You'll have to excuse my naivete but what does this javadoc aggregate actually buy us here? Wouldn't any javadoc problems with a given module be found during the javadoc step in the build?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
alzimmermsft: Analyze runs these steps in the following order:

1. Install Reporting Tools
2. Install SDKs
3. Checkstyle, SpotBugs, and Javadocs (changed in this PR)

It should be safe to add the `-Dinclude-non-shipping-modules` as an option in this step and be able to remove the `SpotBugs Aggregate Analysis` step that runs further down the line. If we opt to keep both and them being their own unique steps I would highly recommend moving the SptBugs Aggregate step before the `Generate Maven Report` step as both generally will fail with the same issues but the SpotBugs step is more verbose on the problem and will fail faster leading to better turnaround times.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
JonathanGiles: This is validating that all JavaDocs for all client libraries remain valid. This helps to prevent broken references across libraries. The build step is only building the libraries for one service, so we may miss some changes.

JonathanGiles:
This PR reduces the Analyze step from approximately 26m to 11m in pipelines.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -155,43 +155,45 @@ jobs:
...
-          options: '--batch-mode -DskipTests -Dgpg.skip'
+          mavenPomFile: pom.client.xml
+          options: '$(DefaultOptions) -DskipTests -Dgpg.skip'
           mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
           javaHomeOption: 'JDKVersion'
           jdkVersionOption: '1.11'
           jdkArchitectureOption: 'x64'
           publishJUnitResults: false
-          goals: 'install'
+          goals: 'install site:site site:stage'
```
JonathanGiles: Note for reviewers: I switched the order of these two tasks, so that the more useful failures will be presented before the Maven site failures, which lack a lot of useful information.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mitchdenny: In a world where we have a pipeline per service, is validating all java docs still the correct thing though? For example, we might actually be building KeyVault, with a binary dependency on Core 1.0.0, but the source contains Core 1.1.0 (substitute versions as necessary to consider a future where the focus has gone off core and is on something else).

Sure ... its validating that the JavaDocs are consistent with what is in the repository, but thats not actually the that it is being built against.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -155,43 +155,45 @@ jobs:
...
 
       - script: |
           git clone https://github.com/JonathanGiles/DependencyChecker.git
           mkdir input && cp dependencies.json input/
         displayName: 'Download dependency checker'
+        condition: ne(variables['Build.Reason'], 'PullRequest')
 
       - task: Maven@3
         displayName: 'Analyze dependencies'
+        condition: ne(variables['Build.Reason'], 'PullRequest')
```
mitchdenny: We need to be a little bit careful about shifting things that should be detected out of the PR process.

JonathanGiles: At this point in time the analyse dependencies step only creates a report. When it can detect violations and fail the build, we should definitely enable this to run all the time.

weshaggard: Looks like we have issue https://github.com/Azure/azure-sdk-for-java/issues/3169 tracking that work

PR #6179, opened 2019-11-06
Adding support for creating reactor-netty-http client from an existing client

anuchandy:


anuchandy:
This enables users to share resources such as connection pool across clients. Also open a way to configure any settings in reactor-netty or netty those are not exposed in builder.

PR #6187, opened 2019-11-06
Update README.md

jenniferdai:
In the public maven repository, I can't find the version stated above - the latest is 12.0.0-preview.4. Please let me know if I'm doing something wrong.

### DIFF
--- a/sdk/storage/azure-storage-blob/README.md
+++ b/sdk/storage/azure-storage-blob/README.md
@@ -22,7 +22,7 @@ definition, such as text or binary data.
 <dependency>
     <groupId>com.azure</groupId>
     <artifactId>azure-storage-blob</artifactId>
-    <version>12.1.0-preview.1</version>
+    <version>12.0.0-preview.4</version>
 </dependency>
 ```
 [//]: # ({x-version-update-end})


JimSuplizio:
The latest released version on Maven is [12.0.0](https://search.maven.org/search?q=g:com.azure%20AND%20a:azure-storage-blob&core=gav). Maven versioning isn't true Semantic Versioning 2.0.0 and because of this it doesn't sort correctly. This is something that's currently being discussed and when things are agreed upon an update will be made in short order to correct that (on our side).

I do not want this change checked in and here's why.

The idea of what people should expect to find when they seek out our repos has been a regular topic of conversation over the past few months.  This is what we’ve landed on at a high level:
•	When customers/developers come to any of our repos, they should expect to find an active developer community, not a snapshot of a specific point in time.
•	The master branch in the repo will always be “latest” meaning that it will often represent a state that has not yet been released.  
•	Documentation (on github.io and docs.microsoft.com) will be published with each release.  We’ve done some work on github.io to preserve documentation for each version of a library that is released going forward.  Customers will be able to find the exact documentation matching the version of the library they are using (which may not be the latest released) on github.io.  
•	The readme contents for a given library's version are being converted to html and their contents added to the landing page and this is the snapshot of that moment in time. Here's the landing page for [azure-storage-blob](https://azuresdkdocs.blob.core.windows.net/$web/java/azure-storage-blob/12.0.0/index.html)

Shortly after we've released we're going to be setting the versions to the next version and using the tooling in place to update all of the respective pom and readme.md files. Hand edits like this should no longer be being done in the java repo.

jenniferdai:
Thank you for your response. I will close this pull request. While I am here, I was wondering if I could ask a clarifying question. I was directed here while looking for Java ADL Gen2 APIs for storage, essentially to implement filesystem operations. Are these the recommended ones? Thank you for your help.

JimSuplizio:
Adding @srnagar and @JonathanGiles who may be better suited to answer your question.

joshfree:
Thank you for your help @jenniferdai and thanks @JimSuplizio for updating the readmes to make it clearer for customers who want to find the General Availability readme, samples, and instructions.

While I do agree the master branch shows the active development of the product, we also want to delight our customers and make it very easy for them to find the supported SDK install point and instructions too.  I believe we can do a better job here and still not interfere with vNext branch activity.  Here's an example of what I mean from the Roslyn repo which blends the latest rolling build results with the latest supported install points - https://github.com/dotnet/roslyn#download-c-and-visual-basic

joshfree:
Just another comment, we're updating the README to make the Master branch and Release branch (release tagging) more clear as well as adding links to additional details on the 12.0.0 (GA) libraries including installation instructions: https://github.com/Azure/azure-sdk-for-java/pull/6248

/cc @hemanttanwar

jenniferdai:
@srnagar and @JonathanGiles wanted to follow up and see if you had an answer regarding the above q / or if there is someone else you would direct me to. thank you!

PR #6191, opened 2019-11-06
Adding collection routing map refresh flag on partition migration

simplynaveen20:
This PR will add x-ms-collectionroutingmap-refresh in header with value true in case of partition migration on backend , which will notify gateway to refresh collection routing map cache on address call

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/internal/directconnectivity/GatewayAddressCache.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/internal/directconnectivity/GatewayAddressCache.java
@@ -155,7 +155,7 @@ public class GatewayAddressCache implements IAddressCache {
 
         if (suboptimalServerPartitionTimestamp != null) {
             logger.debug("suboptimalServerPartitionTimestamp is {}", suboptimalServerPartitionTimestamp);
-            
+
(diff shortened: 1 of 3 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/internal/directconnectivity/GatewayAddressCache.java
```diff
@@ -258,6 +258,11 @@ private URL getServiceEndpoint() {
             headers.put(HttpConstants.HttpHeaders.FORCE_REFRESH, Boolean.TRUE.toString());
         }
 
+        if(request.forceCollectionRoutingMapRefresh)
+        {
+            headers.put(HttpConstants.HttpHeaders.FORCE_COLLECTION_ROUTING_MAP_REFRESH, Boolean.TRUE.toString());
```
kirankumarkolli: TRUE.ToString() materialization everytime.

simplynaveen20: We are using Boolean.TRUE.toString() in above FORCE_REFRESH too , and any way this is not major scenario in fact a rare one in which we can that tostring cost

moderakh:
@Kushagra, on direct connectivity related changes, did we run ReadMyWrites test? I don't see it in the above CIs

PR #6196, opened 2019-11-06
5945 retrypolicy response headers

### LINKED ISSUE #5945: RetryPolicy should support custom retry headers for different Azure services

hemanttanwar:
Currently, RetryPolicy uses retry-after-ms to determine the delay duration. However, different Azure services return the delay duration in different response headers.
So, we should look into supporting different headers for delay duration.
For e.g. AppConfig uses retry-after-ms while CosmosDB uses x-ms-retry-after-ms.
RetryPolicy should also allow a customized list of HTTP status codes that are retryable.
https://github.com/Azure/azure-sdk-for-java/issues/5945

Fixes #5945

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
@@ -46,7 +46,8 @@ public class RetryPolicyOptions {
      * @param retryStrategy The {@link RetryStrategy} used for retries. It will default to {@link ExponentialBackoff}
      * if provided value is {@code null}.
      * @param retryAfterHeader The HTTP header, such as 'Retry-After' or 'x-ms-retry-after-ms', to lookup for the
-     * retry delay.The value {@code null} is valid.
+     * retry delay. If the value is {@code null}, {@link RetryPolicy} will use the retry strategy to compute the delay
+     * and ignore the delay provided in response header.
      * @param retryAfterTimeUnit The time unit to use when applying the retry delay. {@code null} is valid if, and only
      * if, {@code retryAfterHeader} is {@code null}.
      * @throws NullPointerException When {@code retryAfterTimeUnit} is {@code null} and {@code retryAfterHeader} is


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -20,25 +24,49 @@
...
+        this(new ExponentialBackoff(), HttpHeaderName.AZURE_RETRY_AFTER_MS_HEADER);
+    }
+
+    /**
+     * Creates a default {@link ExponentialBackoff} retry policy.
+     * @param retryAfterHeader The retry after http header name to be used get retry after value from
+     * {@link HttpResponse}.
+     * @throws NullPointerException if {@code retryAfterHeader} is {@code null}.
+     */
+    public RetryPolicy(HttpHeaderName retryAfterHeader) {
```
JonathanGiles: Why did you choose an enum rather than a string?

hemanttanwar: I started with String, I already have two string to define, We will have more to use. Enum allow to ensure valid string always and not worried about passing invalid String values, thus defined as type .  We will need more values later https://github.com/google/guava/blob/master/guava/src/com/google/common/net/HttpHeaders.java

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpHeaderName.java
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the MIT License.
+
+package com.azure.core.http;
+
+import java.util.Locale;
+
+/**
+ * Declare various headers which can be used to interact with Azure services.
+ */
+public enum HttpHeaderName {
```
srnagar: It would be simpler to just have a string arg to `RetryPolicy` constructor.  This won't be necessary if you do that.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -20,25 +24,49 @@
...
+     * Creates a {@link RetryPolicy} with the provided {@link RetryStrategy} and {@link HttpHeaderName}.
+     *
+     * @param retryStrategy The {@link RetryStrategy} used for retries.
+     * @param retryAfterHeader The retry after http header name to be used get retry after value from
+     * {@link HttpResponse}.
+     * @throws NullPointerException if {@code retryStrategy} or {@code retryAfterHeader} is {@code null}.
+     */
+    public RetryPolicy(RetryStrategy retryStrategy, HttpHeaderName retryAfterHeader) {
         this.retryStrategy = Objects.requireNonNull(retryStrategy, "'retryStrategy' cannot be null");
+        this.retryAfterHeader = Objects.requireNonNull(retryAfterHeader, "'retryAfterHeader' cannot be null");
```
srnagar: This can be `null`. If the header is not provided, then the retry strategy will be used to determine the delay. If provided, the delay will be determined based on the header value. Also, you will need the TimeUnit as well. Can't assume the unit is always milliseconds. For e.g. the default HTTP header `Retry-After` returns delay in seconds.

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpHeaderName.java
hemanttanwar: See my answer on why I did not use String. But if  we all feel that  it is overengineering, we can just use String. I feel that header names are so common in rest  that somewhere in core we should have list of  all the  header names.. instead of using String in many libraries.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
JonathanGiles: That link is to a series of strings, which would be a better approach than an enum.

hemanttanwar: Okay I will go with String.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -94,14 +131,18 @@ private Duration determineDelayDuration(HttpResponse response, int tryCount) {
             return retryStrategy.calculateRetryDelay(tryCount);
         }
 
-        String retryHeader = response.getHeaderValue(RETRY_AFTER_MS_HEADER);
+        String retryHeaderValue = null;
+
+        if (!isNullOrEmpty(retryAfterHeader)) {
+            retryHeaderValue = response.getHeaderValue(retryAfterHeader);
```
hemanttanwar: retryAfterHeader : This could be null  when user do not want to use any response header for retry.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -20,25 +25,57 @@
...
+     * {@code retryAfterHeader} in {@link HttpResponse}.
+     * @throws NullPointerException if {@code retryStrategy} is {@code null}.
+     */
+    public RetryPolicy(RetryStrategy retryStrategy, String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
+        this.retryStrategy = Objects.requireNonNull(retryStrategy, "'retryStrategy' cannot be null.");
+        this.retryAfterHeader = retryAfterHeader;
+        this.retryAfterTimeUnit = retryAfterTimeUnit;
+        if (!isNullOrEmpty(retryAfterHeader)) {
+            Objects.requireNonNull(retryAfterTimeUnit, "'retryAfterTimeUnit' cannot be null.");
+        }
```
hemanttanwar: retryAfterHeader : This could be null  when user do not want to use any response header for retry.

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpHeaders.java
```diff
@@ -15,6 +15,17 @@
...
+    /**
+     * The header which will have retry after value in milliseconds.
+     */
+    public static final String AZURE_RETRY_AFTER_MS_HEADER = "retry-after-ms";
+
+    /**
+     * The header which will have retry after value in milliseconds.
+     */
+    public static final String AZURE_X_MS_RETRY_AFTER_MS_HEADER = "x-ms-retry-after-ms";
+
```
JonathanGiles: I wouldn't bother adding these strings as public API yet. Leave it up to client libraries to pass it in for now.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -20,25 +25,58 @@
...
      * Creates a default {@link ExponentialBackoff} retry policy.
      */
     public RetryPolicy() {
-        this(new ExponentialBackoff());
+        this(new ExponentialBackoff(), null, null);
+    }
+
+    /**
+     * Creates a default {@link ExponentialBackoff} retry policy.
+     * @param retryAfterHeader The retry after http header name to be used get retry after value from
```
JonathanGiles:
```suggestion
     * @param retryAfterHeader The 'retry-after' HTTP header name to lookup for the retry duration.
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -20,25 +25,58 @@
...
+     * @param retryAfterTimeUnit The time unit to use while applying retry based on value specified in
+     *      * {@code retryAfterHeader} in {@link HttpResponse}.
+     * {@link HttpResponse}.
```
JonathanGiles: This line should be removed.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -20,25 +25,58 @@
...
+    /**
+     * Creates a {@link RetryPolicy} with the provided {@link RetryStrategy} and {@code retryAfterHeader}.
+     *
+     * @param retryStrategy The {@link RetryStrategy} used for retries.
+     * @param retryAfterHeader The retry after http header name to be used get retry after value from
+     * {@link HttpResponse}. The value {@code null} is valid.
+     * @param retryAfterTimeUnit The time unit to use while applying retry based on value specified in
+     * {@code retryAfterHeader} in {@link HttpResponse}.The value {@code null} is valid only in case when
+     * {@code retryAfterHeader} is empty or {@code null}.
+     * @throws NullPointerException if {@code retryStrategy} is {@code null}.
```
alzimmermsft: Update to include statement about the new case where this could throw the NPE

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpHeaders.java
```diff
@@ -15,6 +15,17 @@
...
+    /**
+     * The header which will have retry after value in milliseconds.
+     */
+    public static final String AZURE_RETRY_AFTER_MS_HEADER = "retry-after-ms";
+
+    /**
+     * The header which will have retry after value in milliseconds.
+     */
+    public static final String AZURE_X_MS_RETRY_AFTER_MS_HEADER = "x-ms-retry-after-ms";
```
alzimmermsft: Add in Retry-After as an option, that is the standard HTTP header

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpHeaders.java
```diff
@@ -15,6 +15,17 @@
  * A collection of headers on an HTTP request or response.
  */
 public class HttpHeaders implements Iterable<HttpHeader> {
+
+    /**
+     * The header which will have retry after value in milliseconds.
+     */
+    public static final String AZURE_RETRY_AFTER_MS_HEADER = "retry-after-ms";
```
alzimmermsft: I think these should be moved to the `RetryPolicy` itself, that would make them more discoverable.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -20,25 +25,58 @@
...
+     */
+    public RetryPolicy(RetryStrategy retryStrategy, String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
```
alzimmermsft: Do we know of any service that would support multiple retry headers?

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpHeaders.java
hemanttanwar: removing it based on previous comment from Jonathan.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
hemanttanwar: I have not researched it. I could not answer his.

anuchandy:
just brain storming: what do you guys think about offloading the entire retry delay computation logic to strategies? Instead of adding more Ctrs overloads to RetryPolicy.

Something like:

```java
public interface RetryStrategy {

    default Duration Duration calculateRetryDelay(HttpResponse httpResponse, int retryAttempts) {
        // Look up RFC defined and Azure well known heades such as:
        // retry-after-ms, x-ms-retry-after-ms in a specific order that is agreed across SDKS
        //
        // case1: there is retry header
        Duration delay = httpResponse.headers.get('retry-after-ms') | httpResponse.headers.get('x-ms-retry-after-ms')
        return delay;
        // case 2: there is no retry header
        return calculateRetryDelay(retryAttempts);
    }

    Duration calculateRetryDelay(int retryAttempts);
}
```

This way, if someone want customization, they could override `calculateRetryDelay(HttpResponse httpResponse, int retryAttempts)` and look at any header, multiple header or any other fields of the request/response to decide the delay. This also address the case that @alzimmermsft of pointed - cases where service uses different headers to communicate the delay.

itye-msft:
Hi guys, 
Based on @JonathanGiles request, I am adding my recommendation to also modify the RequestIdPolicy.

Currently `RequestIdPolicy` adds `x-ms-client-request-id` header which is a legacy header.
The service seems to ignore this header, because it expects the new format header `client-request-id`.

The proposed change appends the new header in addition to the legacy one in order to maintain a backwards compatibility. 

This way tracing will start working.

```java

package com.azure.core.http.policy;

import com.azure.core.http.HttpPipelineCallContext;
import com.azure.core.http.HttpPipelineNextPolicy;

import java.util.UUID;

/**
 * The pipeline policy that puts a UUID in the request header. Azure uses the request id as
 */
    private static final String REQUEST_ID_HEADER = "client-request-id";
    private static final String LEGACY_REQUEST_ID_HEADER = "x-ms-client-request-id";

    @Override
        String requestId = context.getHttpRequest().getHeaders().getValue(REQUEST_ID_HEADER);
        if (requestId == null) {
            String randomUUID = UUID.randomUUID().toString();
            context.getHttpRequest().getHeaders().put(REQUEST_ID_HEADER, randomUUID);

            //also set the legacy header for backwards compatibility.
            context.getHttpRequest().getHeaders().put(LEGACY_REQUEST_ID_HEADER, randomUUID);
        }
        return next.process();
    }
}
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -20,25 +25,67 @@
...
+     *
+     * @param retryAfterHeader The 'retry-after' HTTP header name to lookup for the retry duration.The value
+     * {@code null} is valid.
+     * @param retryAfterTimeUnit The time unit to use while applying retry based on value specified in
+     * {@code retryAfterHeader} in {@link HttpResponse}.The value {@code null} is valid only in case when
+     * {@code retryAfterHeader} is empty or {@code null}.
+     * @throws NullPointerException Only if {@code retryAfterTimeUnit} is {@code null} and {@code retryAfterHeader}
+     * is not {@code null}.
+     */
+    public RetryPolicy(String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
```
alzimmermsft: As @srnagar, you, and I discussed let's use a properties class to pass these values instead, so we'll have three constructors for RetryPolicy

1. `RetryPolicy()` defaults to exponential back off which may or may not respect retry after headers
2. `RetryPolicy(RetryStrategy)`
3. `RetryPolicy(RetryOptions)`

`RetryOptions` will contain the options to configure how retrying will occur such as the `RetryStrategy` which will default to ExponentialBackoff and headers to look for to determine retry delay which may or may not be ignored by default.

hemanttanwar: Since we have identified 4-5 more Policies which will need `Options` , I would generalize that all the policy have two constructor similar to following
1. `RetryPolicy()`  
2. `RetryPolicy(RetryOptions)`  

And Options is where we will configure strategy, retry, header  etc. All the variations.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -5,40 +5,45 @@
...
-     * Creates a RetryPolicy with the provided {@link RetryStrategy}.
+     * Creates a {@link RetryPolicy} with the provided {@link RetryPolicyOptions}.
      *
-     * @param retryStrategy The {@link RetryStrategy} used for retries.
+     * @param retryPolicyOptions with given {@link RetryPolicyOptions}.
+     * @throws NullPointerException if {@code retryPolicyOptions} is {@code null}.
      */
-    public RetryPolicy(RetryStrategy retryStrategy) {
-        this.retryStrategy = Objects.requireNonNull(retryStrategy, "'retryStrategy' cannot be null");
+    public RetryPolicy(RetryPolicyOptions retryPolicyOptions) {
```
alzimmermsft: This is an API breaking change.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Objects;
+
+import static com.azure.core.util.CoreUtils.isNullOrEmpty;
+
+/**
+ * Configuration options for {@link RetryPolicy}.
+ */
+public class RetryPolicyOptions {
+
+    private final RetryStrategy retryStrategy;
```
alzimmermsft: Would we want to make this a Fluent model instead? This is likely something that would get extended in the future and that should handle extensions more easily.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
hemanttanwar: yes we have to change any Client library using this, If we are fine with overall approach for all the Policy to have following constructor 

RetryPolicy()
RetryPolicy(RetryOptions)

hemanttanwar: Adding the constructor back.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param retryAfterHeader The 'retry-after' HTTP header name to lookup for the retry duration.The value
+     * {@code null} is valid.
+     * @param retryAfterTimeUnit The time unit to use while applying retry based on value specified in
+     * {@code retryAfterHeader} in {@link HttpResponse}.The value {@code null} is valid only in case when
+     * {@code retryAfterHeader} is empty or {@code null}.
+     * @throws NullPointerException Only if {@code retryAfterTimeUnit} is {@code null} and {@code retryAfterHeader}
+     * is not {@code null}.
+     */
+    public RetryPolicyOptions setRetryAfterHeader(String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
+        return setRetryOptions(new ExponentialBackoff(), retryAfterHeader, retryAfterTimeUnit);
```
alzimmermsft: I don't think this should set the `RetryStrategy`, in addition to this wouldn't applying the response retry header make more sense for `FixedDelay`?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public RetryPolicyOptions setRetryStrategy(RetryStrategy retryStrategy) {
+        return setRetryOptions(retryStrategy, null, null);
+    }
+
+    /**
+     * Sets default {@link ExponentialBackoff} retry policy along with provided {@code retryAfterHeader} and
+     * {@code retryAfterTimeUnit}.
+     *
+     * @param retryAfterHeader The 'retry-after' HTTP header name to lookup for the retry duration.The value
```
alzimmermsft: Let's either remove the specific retry header from the Javadoc or change the Javadoc to something such as:

```suggestion
     * @param retryAfterHeader The HTTP header, such as 'Retry-After' or 'x-ms-retry-after-ms', to lookup for the retry delay
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+     * {@code null} is valid.
+     * @param retryAfterTimeUnit The time unit to use while applying retry based on value specified in
+     * {@code retryAfterHeader} in {@link HttpResponse}.The value {@code null} is valid only in case when
+     * {@code retryAfterHeader} is empty or {@code null}.
```
alzimmermsft: Let's simplify this comment a bit.

``` Java
The time unit to use when applying the retry delay. {@code null} is valid if, and only if, {@code retryAfterHeader} is {@code null}.
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+        this.retryAfterHeader = retryAfterHeader;
+        this.retryAfterTimeUnit = retryAfterTimeUnit;
+        if (!isNullOrEmpty(retryAfterHeader)) {
+            Objects.requireNonNull(retryAfterTimeUnit, "'retryAfterTimeUnit' cannot be null.");
+        }
+        return this;
+    }
+
+    /**
+     * @return {@link RetryStrategy} to be used  in this {@link RetryPolicyOptions}.
```
alzimmermsft:
```suggestion
     * @return The {@link RetryStrategy} used when retrying requests.
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public RetryStrategy getRetryStrategy() {
+        return retryStrategy;
+    }
+
+    /**
+     * @return {@code retryAfterHeader} to be used  in this {@link RetryPolicyOptions}.
```
alzimmermsft:
```suggestion
     * @return The HTTP header which contains the retry delay returned by the service.
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public String getRetryAfterHeader() {
+        return retryAfterHeader;
+    }
+
+    /**
+     * @return {@link ChronoUnit} to be used  for {@code retryAfterHeader}.
```
alzimmermsft:
```suggestion
     * @return The {@link ChronoUnit} used when applying request retry delays.
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+     * @throws NullPointerException Only if {@code retryAfterTimeUnit} is {@code null} and {@code retryAfterHeader}
+     * is not {@code null}.
+     */
+    public RetryPolicyOptions setRetryAfterHeader(String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
+        return setRetryOptions(new ExponentialBackoff(), retryAfterHeader, retryAfterTimeUnit);
+    }
+
+    private RetryPolicyOptions setRetryOptions(RetryStrategy retryStrategy, String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
+        this.retryStrategy = Objects.requireNonNull(retryStrategy, "'retryStrategy' cannot be null.");
+        this.retryAfterHeader = retryAfterHeader;
```
alzimmermsft: I don't think it makes sense to set the value only to throw if the prerequisites are invalid, this will lead to the object getting into an invalid state if the exception is caught and swallowed.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+@Fluent
+public class RetryPolicyOptions {
+
+    private RetryStrategy retryStrategy;
+    private String retryAfterHeader;
+    private ChronoUnit retryAfterTimeUnit;
+
+    /**
+     * Creates a default {@link ExponentialBackoff} for retry policy. It will not use any {@code retryAfterHeader}
```
alzimmermsft:
```suggestion
     * Creates a default {@link RetryPolicyOptions} used by a {@link RetryPolicy}. This will use {@link ExponentialBackoff} as the {@link #getRetryStrategy retry strategy} and will ignore retry delay headers.
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -17,19 +17,19 @@
 
 /**
  * A pipeline policy that retries when a recoverable HTTP error occurs.
+ * @see RetryPolicyOptions
  */
 public class RetryPolicy implements HttpPipelinePolicy {
 
-    private static final String RETRY_AFTER_MS_HEADER = "retry-after-ms";
```
alzimmermsft: You need to update the AppConfiguration RetryPolicy creation to use the `retry-after-ms` header with `ChronoUnit.MILLIS` to maintain current functionality after this change.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -38,7 +38,19 @@ public RetryPolicy() {
...
+    }
+
+    /**
+     * Creates a {@link RetryPolicy} with the provided {@link RetryPolicyOptions}.
+     *
+     * @param retryPolicyOptions with given {@link RetryPolicyOptions}.
+     * @throws NullPointerException if {@code retryPolicyOptions} is {@code null}.
+     */
+    public RetryPolicy(RetryPolicyOptions retryPolicyOptions) {
+        this.retryPolicyOptions = Objects.requireNonNull(retryPolicyOptions,
```
srnagar: This should also check that `retryPolicyOptions.getRetryStrategy()` is also non-null. We need that to be set.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.time.temporal.ChronoUnit;
+import java.util.Objects;
+
+import static com.azure.core.util.CoreUtils.isNullOrEmpty;
+
+/**
+ * Configuration options for {@link RetryPolicy}.
+ */
+@Fluent
```
srnagar: This should be `Immutable` not `Fluent`. We don't want these options changing in the middle of a retry.

hemanttanwar: We did Fluent, to allow client to chain the setters. So we do not need to create many variations of Constructor to set the options.  
If  we do Immutable, They could not be updated which make sense in lot of cases.  But is that true for all 100% of cases ?
We have to choose between @Fluent or @Immutable  , we can not do  both ?

hemanttanwar: After speaking to Allan and Srikanta , we are going with immutable approach.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
hemanttanwar: removing setters

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
hemanttanwar: adding old constructor back `RetryPolicy(RetryStrategy)`

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -17,19 +19,22 @@
...
+    private final RetryPolicyOptions retryPolicyOptions;
 
     /**
-     * Creates a default {@link ExponentialBackoff} retry policy.
+     * Creates {@link RetryPolicy} with default {@link ExponentialBackoff} as {@link RetryStrategy}and use
+     * 'retry-after-ms' in {@link HttpResponse} header for calculating retry delay.
      */
     public RetryPolicy() {
-        this(new ExponentialBackoff());
+        this(new RetryPolicyOptions(new ExponentialBackoff(), RETRY_AFTER_MS_HEADER, ChronoUnit.MILLIS));
```
alzimmermsft: I don't think the default should use `retry-after-ms` for the header, as far as I know this is only used by AppConfiguration.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,101 @@
...
+     *  {@link #getRetryStrategy retry strategy}
+     *
+     * @param retryAfterHeader The HTTP header, such as 'Retry-After' or 'x-ms-retry-after-ms', to lookup for the
+     * retry delay. The value {@code null} is valid.
+     * @param retryAfterTimeUnit The time unit to use when applying the retry delay. {@code null} is valid if, and only
+     * if, {@code retryAfterHeader} is {@code null}.
+     * @throws NullPointerException Only if {@code retryAfterTimeUnit} is {@code null} and {@code retryAfterHeader}
+     * is not {@code null}.
+     */
+    public RetryPolicyOptions(String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
```
alzimmermsft: Should this be a separate constructor from the overload that also takes in the `RetryStrategy`?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param retryStrategy The {@link RetryStrategy} used for retries.
+     * @param retryAfterHeader The HTTP header, such as 'Retry-After' or 'x-ms-retry-after-ms', to lookup for the
+     * retry delay.The value {@code null} is valid.
+     * @param retryAfterTimeUnit The time unit to use when applying the retry delay. {@code null} is valid if, and only
+     *      * if, {@code retryAfterHeader} is {@code null}.
+     * @throws NullPointerException if {@code retryStrategy} is {@code null}. Also when {@code retryAfterTimeUnit}
+     * is {@code null} and {@code retryAfterHeader} is not {@code null}.
+     */
+    public RetryPolicyOptions(RetryStrategy retryStrategy, String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
+        this.retryStrategy = Objects.requireNonNull(retryStrategy, "'retryStrategy' cannot be null.");
```
alzimmermsft: Should this be using `requireNonNull` or just use `ExponentialBackoff` if it is null?

hemanttanwar: will use use ExponentialBackoff

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
hemanttanwar: removing this constructor

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
hemanttanwar: Updated AppConfiguration with this header

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
hemanttanwar: removing this.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -5,7 +5,7 @@
 
 import com.azure.core.http.HttpPipelineBuilder;
 import com.azure.core.annotation.ServiceClientBuilder;
-import com.azure.core.http.policy.AddDatePolicy;
+import com.azure.core.http.policy.*;
```
conniey: You'll need to update IntelliJ to not turn it into a star import after like 99 instead of what it has set by default.. 10?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -76,7 +93,8 @@ public RetryPolicy(RetryStrategy retryStrategy) {
     }
 
     private boolean shouldRetry(HttpResponse response, int tryCount) {
-        return tryCount < retryStrategy.getMaxRetries() && retryStrategy.shouldRetry(response);
+        return tryCount < retryPolicyOptions.getRetryStrategy().getMaxRetries() && retryPolicyOptions.getRetryStrategy()
```
conniey: nit: this is an odd line break. Would expect it to be before `&&`.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -175,7 +168,8 @@ public ConfigurationAsyncClient buildAsyncClient() {
         policies.add(new ConfigurationCredentialsPolicy(buildCredential));
         HttpPolicyProviders.addBeforeRetryPolicies(policies);
 
-        policies.add(retryPolicy == null ? new RetryPolicy() : retryPolicy);
+        policies.add(retryPolicy == null ? new RetryPolicy( new RetryPolicyOptions(new ExponentialBackoff(),
```
conniey: Since RetryPolicyOptions is immutable, to save allocations, I'd have a default RetryPolicyOptions that is static and immutable.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
hemanttanwar: yes

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Creates the {@link RetryPolicyOptions} with provided {@link RetryStrategy}, {@code retryAfterHeader} and
+     * {@code retryAfterTimeUnit} that will be used when a request is retried.
+     *
+     * @param retryStrategy The {@link RetryStrategy} used for retries. It will default to {@link ExponentialBackoff}
+     * if provided value is {@code null}.
+     * @param retryAfterHeader The HTTP header, such as 'Retry-After' or 'x-ms-retry-after-ms', to lookup for the
+     * retry delay.The value {@code null} is valid.
```
srnagar: Instead of saying `null` is valid, add a comment to mention what happens when the value is `null`. For e.g. `If this value is null, retry policy will use the retry strategy to compute the delay and ignore the delay provided in response header`.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicyOptions.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param retryAfterTimeUnit The time unit to use when applying the retry delay. {@code null} is valid if, and only
+     * if, {@code retryAfterHeader} is {@code null}.
+     * @throws NullPointerException When {@code retryAfterTimeUnit} is {@code null} and {@code retryAfterHeader} is
+     * not {@code null}.
+     */
+    public RetryPolicyOptions(RetryStrategy retryStrategy, String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
+
+        if (Objects.isNull(retryStrategy)) {
+            this.retryStrategy = new ExponentialBackoff();
```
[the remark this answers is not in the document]
hemanttanwar: We tried to avoid many variations on constructor. Strategy cannot be null, Thus  defaulting to  `ExponentialBackoff`  . Following constructor was removed.

public RetryPolicyOptions( String retryAfterHeader, ChronoUnit retryAfterTimeUnit) ;
trying to manage between many variations in constructor or few constructor.

JonathanGiles:
I realise this has already been merged, but I'm not convinced of the value of the RetryPolicyOptions type. It seems to present more API complexity for not much more gain - unless I'm missing something?

JonathanGiles:
@hemanttanwar I'm keen to discuss the comment I raised yesterday. Is there an argument for the RetryPolicyOptions type?

anuchandy:
bringing up: https://github.com/Azure/azure-sdk-for-java/pull/6196#issuecomment-551237355

anuchandy:
My proposal earlier was: Offload the entire retry delay computation logic to strategies Instead of adding more Ctrs overloads to RetryPolicy.

Something like:

```java
public interface RetryStrategy {

    default Duration Duration calculateRetryDelay(HttpResponse httpResponse, int retryAttempts) {
        // Look up RFC defined and Azure well known heades such as:
        // retry-after-ms, x-ms-retry-after-ms in a specific order that is agreed across SDKS
        //
        // case1: there is retry header
        Duration delay = httpResponse.headers.get('retry-after-ms') | httpResponse.headers.get('x-ms-retry-after-ms')
        return delay;
        // case 2: there is no retry header
        return calculateRetryDelay(retryAttempts);
    }

    Duration calculateRetryDelay(int retryAttempts);
}
```

This way, if someone want customization, they could override `calculateRetryDelay(HttpResponse httpResponse, int retryAttempts)` and look at any header, multiple header or any other fields of the request/response to decide the delay. This also address the case that @alzimmermsft of pointed - cases where service uses different headers to communicate the delay.

  \\cc @srnagar

anuchandy:
basically any method in `RetryStrategy` that takes a dynamic decision will have access to `HttpResponse` (and indirectly has access to `HttpRequest` as well). So AzConfig customization will looks like:

```java
RetryPolicy DEFAULT_RETRY_POLICY = new RetryPolicy(new ExponentialBackoff() {
    @Override
    public  Duration calculateRetryDelay(HttpResponse httpResponse, int retryAttempts)) {
        String delay = httpResponse.getHeaderValue("retry-after-ms");
        if (delay != null) {
            return Duration.ofMillis(Long.parseLong(delay));
        } else {
            return calculateRetryDelay(retryAttempts);
        }
    }
});

policies.add(retryPolicy == null ? DEFAULT_RETRY_POLICY : retryPolicy);
policies.addAll(this.policies);
HttpPolicyProviders.addAfterRetryPolicies(policies);
```

conniey:
I like @anuchandy 's proposed suggestion. Maybe opening up an issue so this issue doesn't get lost in this merged PR.

anuchandy:
sure, created an issue https://github.com/Azure/azure-sdk-for-java/issues/6363 to triage.

hemanttanwar:
I have Closed  many related issue and moving all the discussion into one #6217 

I will have a another PR  soon do implement suggested changes by @anuchandy

PR #6199, opened 2019-11-06
Change 'preview' to 'beta' in versions

JimSuplizio:
Change "preview" to "beta" in version strings. The reason this fix was necessary is because the version comparison on Maven isn't a semver2 comparison (which is what we were previously trying to confirm to) which results in preview versions showing as higher than the released versions (eg. 1.0.0-preview.2 > 1.0.0). An example of this can be found [here](https://search.maven.org/search?q=g:com.azure%20AND%20a:azure-storage-blob). Maven sorts show the latest version as 12.0.0-preview.4 instead of 12.0.0.


JimSuplizio:
The failures in appconfiguration and keyvault are not the result of this PR, they're the result of the JUnit 5 upgrade. An [issue ](https://github.com/Azure/azure-sdk-for-java/issues/6208) was filed for that.

JimSuplizio:
I'm squashing and merging this, the test failures are known (and commented above).

Issue #5247, opened 2019-09-04
Direct TCP: address read-my-writes test failure


David-Noble-at-work:
After 1-5 minutes the `RntbdTransportClient` stops processing request right after warning message is logged:
```
WARN  com.azure.data.cosmos.internal.directconnectivity.rntbd.RntbdRequestManager - ChannelHandlerContext(RntbdRequestManager#0, [id: 0x88ff2701, L:/10.105.157.160:61595 - R:cdb-ms-prod-westus2-fd21.documents.azure.com/40.78.243.196:16774]) response ignored because there is no matching pending request: RntbdResponse(...
```
A full log for a recent run is attached: [tests.log.zip](https://github.com/Azure/azure-sdk-for-java/files/3576694/tests.log.zip)

kushagraThapar:
This is fixed.

PR #6201, opened 2019-11-07
Cosmos Issue #5247: Address read-my-writes test failure

David-Noble-at-work:
Addresses issue #5247 and improves Direct TCP diagnostics. The fix to the read-my-writes test failure is a single-line change to `RntbdTransportClient`. A number of other changes were made to:

- Improve Direct TCP diagnostics
  One of the more notable improvements is the addition of `RntbdRequestRecord.Stage` which is useful for tracking a request through its lifetime in the RNTBD request pipeline. This addition ripples through the code in a few places. 

- An RntbdTransportClient.Options addition: requestExpiryInverval.
  This option will eventually be used to cancel "lost" requests. A lost request is one which was successfully sent but for which we get no response. The default value is 5 seconds, the time within which the server guarantees a response. Rationale: requests are sometimes lost when testing under load. When these requests are lost we currently wait the full request timeout interval; 65 seconds by default. Waiting just 5 seconds to fail a lost request once sent reduces our latency significantly.

- Modest azure-cosmos-benchmark package changes that I found useful in troubleshooting the read-my-writes test failure.

- Import optimization, code tidying/method sorting in a couple of places.

Most of the latter two changes were ported from my working branch on the v2.6 code path. The port to v2.6 and v3.0 will follow completion of this PR.

### DIFF
--- a/sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/ReadMyWriteWorkflow.java
+++ b/sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/ReadMyWriteWorkflow.java
@@ -133,7 +133,7 @@ class ReadMyWriteWorkflow extends AsyncBenchmark<Document> {
 
         concurrencyControlSemaphore.acquire();
         logger.debug("concurrencyControlSemaphore: {}", concurrencyControlSemaphore);
-        
+
         try {
             obs.subscribeOn(Schedulers.parallel()).subscribe(baseSubscriber);
         } catch (Throwable error) {


sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/Main.java
```diff
@@ -73,6 +74,8 @@ public static void main(String[] args) throws Exception {
             System.err.println("INVALID Usage: " + e.getMessage());
             System.err.println("Try '-help' for more information.");
             throw e;
+        } finally {
+            System.exit(0);
```
moderakh: why do we need this?
This will cause problem for tests directly invoking Main:main()

PR #6202, opened 2019-11-07
Add a tool to convert exported dependency data for graphing

bsiegel:
The data is exported in a format that will be consumed by the client side dependency graph (currently available in Azure/azure-sdk-tools#278)

### DIFF
--- a/eng/dump-deps/dump-deps.js
+++ b/eng/dump-deps/dump-deps.js
@@ -46,7 +46,7 @@ const convertDependencyFile = async (inputFile) => {
           };
         }
 
-        if (depChains.some(chain => !chain["dependencyChain"])) {
+        if (depChains.length === 1 && !depChains[0].dependencyChain) {
           // dep is a direct dependency of lib
           dumpData[libId].deps.push({
             name: depName,


bsiegel:
Currently there is a bug with the DependencyChecker tool, the JSON reporter is cutting off output in the middle of the export leading to an incomplete document. That also needs to be fixed before we can start using this tool.

JonathanGiles:
FYI - #6176 moves the dependency work to only occur in the nightly builds. This was done to reduce build time, but also it was based on the current situation (I believe) that no validation of dependencies currently occurs, just reporting. If this changes, this dependency check should happen in PRs again.

bsiegel:
The bug in the DependencyChecker is fixed in JonathanGiles/DependencyChecker#4

weshaggard:
@bsiegel what is the latest status on this? Is there somewhere I can play with the output report?

bsiegel:
> @bsiegel Brandon Siegel FTE what is the latest status on this? Is there somewhere I can play with the output report?

@weshaggard I don't have them uploaded anywhere but you can run the tool locally to generate the "data.js" file, and then simply drop it along size the files from the renderer and it will work.

I've been working on some Mobile team related stuff the last couple of days, but I believe the last few things I needed to do were a few more tweaks to the renderer (just nice to haves, really) and then wanted to load a graph for each language and take a final visual pass to make sure everything looked correct. Then we can merge all of these.

I also looked into adding them to the build results in Azure Pipelines but that unfortunately only supports markdown, javascript is not available there. So I would plan to have another task to publish the graph alongside the HTML report for each pipeline that publishes a report (I can add a hyperlink in the HTML report to take you to the graph).

bsiegel:
Sample: https://gist.githack.com/bsiegel/89047bb3b914073fdb62026eacd5038b/raw/a5d28c16cdaee7a61b1995135a0e78152f651f06/index.html?data=java.js

JonathanGiles:
Are there any further plans for this PR? Do we plan to merge and automate this?

JimSuplizio:
> Are there any further plans for this PR? Do we plan to merge and automate this?

Yes, the various language EngSys owners are meeting up to discuss this.

JimSuplizio:
I'm going to close this PR. The dependency report was changed to be an inter-dependency report and I can do this with a simple powershell script which will not require the output from the dependency checker.

PR #6205, opened 2019-11-07
Cosmos CFP: fix for race condition issue

milismsft:
ChangeFeedProcessor: fix a race condition issue where the lease document has been updated while we are trying to write the current state like when check-pointing the continuation token.

### DIFF
--- a/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
@@ -53,9 +53,8 @@ class AutoCheckpointer implements ChangeFeedObserver {
     @Override
     public Mono<Void> processChanges(ChangeFeedObserverContext context, List<CosmosItemProperties> docs) {
         return this.observer.processChanges(context, docs)
-            .onErrorResume(throwable -> {
+            .doOnError(throwable -> {
                 logger.warn("Unexpected exception from thread {}", Thread.currentThread().getId(), throwable);
-                return Mono.error(throwable);
             })
             .then(this.afterProcessChanges(context));
     }
@@ -65,9 +64,8 @@ class AutoCheckpointer implements ChangeFeedObserver {
 
         if (this.isCheckpointNeeded()) {
             return context.checkpoint()
-                .onErrorResume(throwable -> {
+                .doOnError(throwable -> {
                     logger.warn("Checkpoint failed; this worker will be killed", throwable);
-                    return Mono.error(throwable);
                 })
                 .doOnSuccess( (Void) -> {
                     this.processedDocCount = 0;
--- a/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/LeaseStoreManagerImpl.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/LeaseStoreManagerImpl.java
@@ -380,10 +380,9 @@ public class LeaseStoreManagerImpl implements LeaseStoreManager, LeaseStoreManag
 
                     return serverLease;
                 }))
-            .onErrorResume(throwable -> {
+            .doOnError(throwable -> {
                 logger.info("Partition {} lease with token '{}' failed to checkpoint for owner '{}' with continuation token '{}'",
                     lease.getLeaseToken(), lease.getConcurrencyToken(), lease.getOwner(), lease.getContinuationToken());
-                return Mono.error(throwable);
             });
     }
 
--- a/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
@@ -97,9 +97,8 @@ class PartitionProcessorImpl implements PartitionProcessor {
                 this.lastContinuation = documentFeedResponse.continuationToken();
                 if (documentFeedResponse.results() != null && documentFeedResponse.results().size() > 0) {
                     return this.dispatchChanges(documentFeedResponse)
-                        .onErrorResume(throwable -> {
+                        .doOnError(throwable -> {
                             logger.debug("Exception was thrown from thread {}", Thread.currentThread().getId(), throwable);
-                            return Mono.error(throwable);
                         })
                         .doOnSuccess((Void) -> {
                             this.options.requestContinuation(this.lastContinuation);
--- a/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/LeaseRenewerImpl.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/LeaseRenewerImpl.java
@@ -62,13 +62,12 @@ class LeaseRenewerImpl implements LeaseRenewer {
                 return !cancellationToken.isCancellationRequested();
             })
             .then()
-            .onErrorResume(throwable -> {
+            .doOnError(throwable -> {
                 if (throwable instanceof LeaseLostException) {
                     logger.info("Partition {}: renew lease loop failed.", this.lease.getLeaseToken(), throwable);
                 } else {
                     logger.error("Partition {}: renew lease loop failed.", this.lease.getLeaseToken(), throwable);
                 }
-                return Mono.error(throwable);
             });
     }
 
--- a/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/ObserverExceptionWrappingChangeFeedObserverDecorator.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/ObserverExceptionWrappingChangeFeedObserverDecorator.java
@@ -52,9 +52,8 @@ class ObserverExceptionWrappingChangeFeedObserverDecorator implements ChangeFeed
     @Override
     public Mono<Void> processChanges(ChangeFeedObserverContext context, List<CosmosItemProperties> docs) {
         return this.changeFeedObserver.processChanges(context, docs)
-            .onErrorResume(throwable -> {
+            .doOnError(throwable -> {
                 this.logger.warn("Exception thrown during ChangeFeedObserver.processChanges from thread {}", Thread.currentThread().getId(), throwable);
-                return Mono.error(new ObserverException(throwable));
             });
     }
 }


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
```diff
@@ -50,6 +53,10 @@ public void close(ChangeFeedObserverContext context, ChangeFeedObserverCloseReas
     @Override
     public Mono<Void> processChanges(ChangeFeedObserverContext context, List<CosmosItemProperties> docs) {
         return this.observer.processChanges(context, docs)
+            .onErrorResume(throwable -> {
+                logger.warn("Unexpected exception from thread {}", Thread.currentThread().getId(), throwable);
```
kushagraThapar: logger.error ?
logger.debug ?

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
```diff
@@ -58,10 +65,15 @@ public void close(ChangeFeedObserverContext context, ChangeFeedObserverCloseReas
 
         if (this.isCheckpointNeeded()) {
             return context.checkpoint()
-                .doOnSuccess((Void) -> {
+                .onErrorResume(throwable -> {
+                    logger.warn("Checkpoint failed; this worker will be killed", throwable);
```
kushagraThapar: logger.error ?
logger.debug ?

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/DefaultObserver.java
```diff
@@ -34,8 +34,13 @@ public void close(ChangeFeedObserverContext context, ChangeFeedObserverCloseReas
     @Override
     public Mono<Void> processChanges(ChangeFeedObserverContext context, List<CosmosItemProperties> docs) {
         log.info("Start processing from thread {}", Thread.currentThread().getId());
-        consumer.accept(docs);
-        log.info("Done processing from thread {}", Thread.currentThread().getId());
+        try {
+            consumer.accept(docs);
+            log.info("Done processing from thread {}", Thread.currentThread().getId());
+        } catch (Exception ex) {
+            log.warn("Unexpected exception thrown from thread {}", Thread.currentThread().getId(), ex);
```
kushagraThapar: logger.error ?
logger.debug ?

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
```diff
@@ -50,6 +53,10 @@ public void close(ChangeFeedObserverContext context, ChangeFeedObserverCloseReas
     @Override
     public Mono<Void> processChanges(ChangeFeedObserverContext context, List<CosmosItemProperties> docs) {
         return this.observer.processChanges(context, docs)
+            .onErrorResume(throwable -> {
```
moderakh: `onErrorResume` is for when you are changing the return value, which is not the case here.
`doOnError` is more suitable for these cases
https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#doOnError-java.util.function.Consumer-

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
```diff
@@ -58,10 +65,15 @@ public void close(ChangeFeedObserverContext context, ChangeFeedObserverCloseReas
 
         if (this.isCheckpointNeeded()) {
             return context.checkpoint()
-                .doOnSuccess((Void) -> {
+                .onErrorResume(throwable -> {
```
moderakh: `onErrorResume` is for when you are changing the return value, which is not the case here.
`doOnError` is more suitable for these cases
https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#doOnError-java.util.function.Consumer-

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/LeaseStoreManagerImpl.java
```diff
@@ -379,7 +379,12 @@ public LeaseStoreManagerBuilderDefinition hostName(String hostName) {
                     serverLease.setContinuationToken(continuationToken);
 
                     return serverLease;
-                }));
+                }))
+            .onErrorResume(throwable -> {
```
moderakh: ditto

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
```diff
@@ -96,11 +97,15 @@ public PartitionProcessorImpl(ChangeFeedObserver observer, ChangeFeedContextClie
                 this.lastContinuation = documentFeedResponse.continuationToken();
                 if (documentFeedResponse.results() != null && documentFeedResponse.results().size() > 0) {
                     return this.dispatchChanges(documentFeedResponse)
-                        .doFinally( (Void) -> {
+                        .onErrorResume(throwable -> {
+                            logger.debug("Exception was thrown from thread {}", Thread.currentThread().getId(), throwable);
+                            return Mono.error(throwable);
```
moderakh: ditto

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
```diff
@@ -157,13 +163,21 @@ public PartitionProcessorImpl(ChangeFeedObserver observer, ChangeFeedContextClie
...
+                        this.options.partitionKey().getInternalPartitionKey(), Thread.currentThread().getId());
+                    this.resultException = (LeaseLostException) throwable;
                 } else if (throwable instanceof TaskCancelledException) {
-                    this.logger.debug("Exception: partition {}", this.settings.getPartitionKeyRangeId(), throwable);
+                    logger.debug("Task cancelled exception: partition {} from {}",
+                        this.settings.getPartitionKeyRangeId(), Thread.currentThread().getId(), throwable);
                     this.resultException = (TaskCancelledException) throwable;
+                } else {
+                    logger.warn("Unexpected exception from thread {}", Thread.currentThread().getId(), throwable);
+                    this.resultException = new RuntimeException(throwable);
```
moderakh: just please be aware that this doesn't record exception stacktrace. If that matters. this will maintain the original exception stacktrace though.

Is that ok?

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
```diff
@@ -175,8 +189,13 @@ public PartitionProcessorImpl(ChangeFeedObserver observer, ChangeFeedContextClie
 
                 return true;
             })
-            .onErrorResume(throwable -> Flux.empty())
-            .then();
+            .onErrorResume(throwable -> {
+                if (this.resultException == null) {
+                    this.resultException = new RuntimeException(throwable);
```
moderakh: just please be aware that this doesn't record exception stacktrace. If that matters. this will maintain the original exception stacktrace though.

Is that ok?

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
milismsft: logger.warn is the right thing here; the exception could be thrown from the user code which handles the feeds. In any case the side effect of this is to kill the current worker processing the feeds and start a new one from where we left (some documents might be processed multiple times).

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
milismsft: Same as above, the exception could be thrown from the user code which handles the feeds. This is more or less for tracking purposes, as we want to capture as part of the default logs any case where we might have failed to check-point.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/DefaultObserver.java
milismsft: Same as above, the exception could be thrown from the user code which handles the feeds. This is more or less for tracking purposes, as we want to capture as part of the default logs any case where we might have failed to check-point. And it is not an error because the case is recoverable by killing the current worker and create a new one.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
milismsft: Fixed

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/AutoCheckpointer.java
milismsft: Fixed

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/LeaseStoreManagerImpl.java
milismsft: Fixed

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
milismsft: Fixed

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
milismsft: yes; the reason we create a RuntimeException here and assign it to the resultException is to signal to the "parent" that the processing worker should be killed along with the corresponding renewal worker.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
milismsft: just like above, yes, the reason we create a RuntimeException here and assign it to the resultException is to signal to the "parent" that the processing worker should be killed along with the corresponding renewal worker.

PR #6210, opened 2019-11-07
Updated Datalake versioning

gapra-msft:


sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileAsyncClient.java
```diff
@@ -26,9 +26,7 @@
 
 import java.nio.ByteBuffer;
 
-import static com.azure.core.util.FluxUtil.monoError;
-import static com.azure.core.util.FluxUtil.fluxError;
-import static com.azure.core.util.FluxUtil.withContext;
+import static com.azure.core.implementation.util.FluxUtil.*;
```
alzimmermsft: `*` import failing Checkstyles

alzimmermsft:
> 
> 
> The versioning for these was done incorrectly, please look at the [Versions and versioning](https://github.com/Azure/azure-sdk-for-java/blob/master/CONTRIBUTING.md) section of the contributing.md.
> 
> The versions you'd wanted to change for this should have been updated using the update_versions script. Further, there are a number of "<!-- {x-version-update..." tags that were removed from pom files that should not have been. Also, you're reverting the changes in projects under the eng directory to the version that you're about to release, which you'll need to do in order for things to pass but then they're going to have to be set back to the "next" version after this is released.

@JimSuplizio 

This PR is meant to change the Maven dependencies to those in Maven central, Datalake is being released a little later than the other libraries and this is meant to fix its version references.

gapra-msft:
@JimSuplizio and I synced offline.

PR #6213, opened 2019-11-07
User agent util and JUnit 5 fix for app config

### LINKED ISSUE #3141: Remove hardcoded SDK_NAME and SDK_VERSION from ConfigurationClient

### LINKED ISSUE #6208: Moving to JUnit 5 for unit test has caused test failures in the appconfiguration and keyvault piplines
With the update to JUnit 5 both the appconfiguration and keyvault ci pipelines are seeing a bunch of NullPointerExceptions in their test runs. I'd ended up running CI runs targeting this SHA, and the previous SHA, to verify.
https://dev.azure.com/azure-sdk/public/_build/results?buildId=182476&view=results
https://dev.azure.com/azure-sdk/public/_build/results?buildId=182478&view=results

srnagar:
The intent of this change is to read the version and artifact name from maven pom.xml instead of hard-coded strings in Java classes which can get out of sync with the real version.

This change also fixes the issues caused by upgrading to JUnit 5 in azure-core.

Partially fixes #3141
Partially fixes #6208 

This PR is for AppConfig only. The other client libraries will be updated in a follow-up PR.

Closing the related PR - https://github.com/Azure/azure-sdk-for-java/pull/6135 as this required updating JUnit as well.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
+++ b/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
@@ -68,11 +68,18 @@
   KeyVaultCredentials.getAuthenticationCredentials to return null. -->
   <Match>
     <Or>
-      <Class name="com.azure.core.util.CoreUtils" /> <!-- false positive -->
+      <Class name="com.azure.core.util.CoreUtils"/> <!-- false positive -->
       <Class name="com.microsoft.azure.keyvault.authentication.KeyVaultCredentials"/>
       <Class name="com.azure.storage.blob.FileTransferExample"/>
     </Or>
     <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE"/>
+  </Match>
+
+  <Match>
+    <Or>
+      <Class name="com.azure.core.util.CoreUtils"/> <!-- false positive -->
+    </Or>
+    <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NULL_VALUE"/>
   </Match>
 
   <!-- These KeyVaultClientBase methods are made synchronous by blocking and waiting for a result. They do not return
@@ -117,6 +124,7 @@
     <Or>
       <Class name="~com\.microsoft\.azure\.(.+)Impl"/>
       <Class name="~com\.azure\.(.+)Impl"/>
+      <Class name="com.azure.core.util.CoreUtils"/>
     </Or>
     <Bug pattern="NP_LOAD_OF_KNOWN_NULL_VALUE"/>
   </Match>
--- a/eng/pipelines/client.yml
+++ b/eng/pipelines/client.yml
@@ -92,7 +92,7 @@ jobs:
         displayName: 'Install all client libraries (for SpotBugs analysis)'
         inputs:
           mavenPomFile: pom.client.xml
-          options: '$(DefaultOptions) -DskipTests -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Dinclude-non-shipping-modules'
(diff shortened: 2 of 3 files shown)


sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -56,10 +56,24 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-library</artifactId>
```
JonathanGiles: test scope?

sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
```diff
@@ -0,0 +1,57 @@
...
+     * Helper method that returns the {@link UserAgentProperties} defined in the {@code propertiesFileName}.
+     *
+     * @param propertiesFileName The file name defining the user agent properties.
+     * @return the {@link UserAgentProperties}.
+     */
+    public static UserAgentProperties getUserAgentProperties(String propertiesFileName) {
+        return USER_AGENT_PROPERTIES.computeIfAbsent(propertiesFileName, AzureUserAgentUtil::readFromPropertiesFile);
+    }
+
+    private static UserAgentProperties readFromPropertiesFile(String propertiesFileName) {
```
JonathanGiles: I would be inclined to just return a Properties or Map, rather than have a new type.

sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Utility class for fetching the user agent properties like client library name and version.
+ */
+public class AzureUserAgentUtil {
```
JonathanGiles: We should iterate on the name a little...I think we should make it less specific so that we can load in other properties too.

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
srnagar: Updated to test scope.

sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
srnagar: I thought about returning a Map/Properties but that would require each of the client libraries to know/define the keys ("name", "version" etc). 

`Properties` type is mutable, so, Map may be a better option here if a new type is not preferable.

sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
srnagar: AzureApplicationPropertiesUtil, AzurePropertiesUtil, AzureSdkPropertiesUtil, PropertiesUtil and the other option is to move this method to CoreUtil which already exists.

srnagar: Moved this method to `CoreUtils`.

sdk/core/azure-core/src/main/java/com/azure/core/util/AzureUserAgentUtil.java
srnagar: Returning a `Map<String, String>` now.

Issue #5711, opened 2019-10-07
Align Logging with the Updated Logging Specification


alzimmermsft:
https://azure.github.io/azure-sdk/general_implementation.html#logging

- [x] Review places where exceptions are being logged and verify they are using the warning or exception level
- [x] Update logger to log exception messages during the warning or error level and the entire stack trace during the verbose level

PR #6218, opened 2019-11-07
Updates the ClientLogger Logging Logic

### LINKED ISSUE #5446: ClientLogger does not log exceptions
When logging exceptions with ClientLogger, it filters out exceptions before forwarding it to slf4j.

> I think we should consider updating ClientLogger to provide as much information about an exception (including stacktrace) as possible. For some reason, ClientLogger filters out exceptions before sending it to slf4j logger. If we don't do that, then slf4j will log the entire stacktrace.

https://github.com/Azure/azure-sdk-for-java/pull/5346#discussion_r326474853

alzimmermsft:
Fixes #5446 and #5711 

Updates `ClientLogger` to log exception messages when the environment logging level is set to a log level less than `VERBOSE`. Fixes an issue that the stack trace logging was checking the log level used to log the exception and not the environment logging level.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
@@ -270,8 +270,8 @@ public class ClientLogger {
      * @return Flag indicating if the environment and logger are configured to support logging at the given log level.
      */
     private boolean canLogAtLevel(int logLevel, int environmentLoggingLevel) {
-        // Attempting to log at a level not supported by the environment or logger is null.
-        if (logLevel < environmentLoggingLevel || logger == null) {
+        // Attempting to log at a level not supported by the environment.
+        if (logLevel < environmentLoggingLevel) {
             return false;
         }
 


sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -219,48 +246,99 @@ private void performLogging(int logLevel, String format, Object... args) {
...
+     * @return The arguments with the last element removed.
+     */
+    private Object[] removeThrowable(Object... args) {
+        return Arrays.copyOf(args, args.length - 1);
+    }
+
+    /*
+     * Helper class which contains logging configuration.
+     */
+    private static class LoggingConfiguration {
```
srnagar: I think that too many of these `LoggingConfiguration` objects will be created even when logging is not enabled. Is there a way to check `canLog` condition first and then proceed with creating this object?

srnagar: For e.g. in Event Hubs, where we have millions of events being processed per minute, if there's a `verbose` or `info` level log statement in code, it's going to create those many `LoggingConfiguration` objects each time that code is executed only to find out that the log level set by user is `warning` or above.

alzimmermsft: Removed `LoggingConfiguration` class.

alzimmermsft: Removed `LoggingConfiguration` class.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -219,48 +251,67 @@ private void performLogging(int logLevel, String format, Object... args) {
...
+    /*
+     * Determines if the environment and logger support logging at the given log level.
+     *
+     * @param logLevel Logging level for the log message.
+     * @param environmentLoggingLevel Logging level the environment is set to support.
+     * @return Flag indicating if the environment and logger are configured to support logging at the given log level.
+     */
+    private boolean canLogAtLevel(int logLevel, int environmentLoggingLevel) {
+        // Attempting to log at a level not supported by the environment or logger is null.
+        if (logLevel < environmentLoggingLevel || logger == null) {
```
srnagar: `logger` will never be `null`. Don't need the null check here.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -193,10 +201,28 @@ private RuntimeException logException(RuntimeException runtimeException, int log
      * @param format formattable message.
      * @param args Arguments for the message, if an exception is being logged last argument is the throwable.
      */
-    private void performLogging(int logLevel, String format, Object... args) {
+    private void performLogging(int logLevel, int environmentLogLevel, boolean isExceptionLogging, String format,
```
srnagar: `logLevel` should be sufficient to check whether throwable should be stripped out or not. Don't need the `environmentLogLevel` here.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
alzimmermsft: This may have been done in the past for SpotBugs, will investigate.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
alzimmermsft: Based on the specification `DO log exceptions thrown as a Warning level message. If the log level set to Verbose, append stack trace information to the message.` I took it to mean if an exception is being logged log it as a warning and if the environment logging level is set to verbose additionally include the stack trace.

PR #6226, opened 2019-11-08
Update Datalake versions after release

JimSuplizio:


eng/versioning/version_client_java_files.txt
```diff
@@ -3,7 +3,6 @@
 # has been fixed this file and the update logic in update_versions.py should no longer be necessary.
 
 # The file format here should be the relative path from the root of the azure-sdk-for-java
-sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/AzureConfiguration.java
```
JimSuplizio: This was an opportunistic fix. Apparently earlier today this file was removed from source (not renamed, deleted) and this caused the update script to fail.

sdk/storage/azure-storage-file-datalake/pom.xml
```diff
@@ -76,13 +76,13 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-test</artifactId>
-      <version>1.0.0</version>  <!-- {x-version-update;com.azure:azure-core-test;dependency} -->
+      <version>1.1.0-beta.1</version>  <!-- {x-version-update;com.azure:azure-core-test;dependency} -->
```
mitchdenny: I'm assuming this means we are shipping this version of Core to get Data Lake out?

eng/spotbugs-aggregate-report/pom.xml
```diff
@@ -68,8 +68,8 @@
                 <source>..\..\sdk\storage\azure-storage-blob-cryptography\src\samples\java\com</source>
                 <source>..\..\sdk\storage\azure-storage-file-share\src\main\java\com</source>
                 <source>..\..\sdk\storage\azure-storage-file-share\src\samples\java\com</source>
-<!--                <source>..\..\sdk\storage\azure-storage-file-datalake\src\main\java\com</source>-->
-<!--                <source>..\..\sdk\storage\azure-storage-file-datalake\src\samples\java\com</source>-->
+                <source>..\..\sdk\storage\azure-storage-file-datalake\src\main\java\com</source>
```
alzimmermsft: This is going to hit an issue with SpotBugs if we don't bump the version of Azure Core Datalake is depending on.

sdk/storage/azure-storage-file-datalake/pom.xml
JimSuplizio: @mitchdenny , Datalake had to revert this version temporarily like 2 hours go in order to ship. I'm just changing it back to what it was previously since they shipped. The updating post GA snafu'd their original release. This is the value that it was prior to them swapping so they could build and target the released version.

JimSuplizio: @mitchdenny the changed things back to build and release, that was done in this [PR](https://github.com/Azure/azure-sdk-for-java/pull/6210) . I'm putting the versions back to the next version.

eng/spotbugs-aggregate-report/pom.xml
JimSuplizio: @alzimmermsft  You're right. All of the changes made have to be reverted including the source changes for the core changes post GA which were changed back in the original PR.

PR #6230, opened 2019-11-08
Renamed internal package to implementation

kushagraThapar:
* Renamed `internal` package to `implementation`
** The reason to do that is to enable checkstyle and spotBugs on the project and suppress implementation from the package list because of the nomenclature followed by azure-core team.  

Other than package name change, there is one more change which reflects in pom.xml
* Skip implementation package from generating docs.


kushagraThapar:
Because of merge conflicts, need to close this PR. 
I have opened the same PR here: https://github.com/Azure/azure-sdk-for-java/pull/6234

PR #6246, opened 2019-11-08
Minor readme update

JimSuplizio:
There is no content change, this is only a spacing change. Code blocks in markdown should have an blank line before them. While the code block is valid without the blank line it trips up some parsers and in our case it trips up markdown2's conversion when we're generating the html for the documentation.

### DIFF
--- a/sdk/storage/azure-storage-file-datalake/README.md
+++ b/sdk/storage/azure-storage-file-datalake/README.md
@@ -19,6 +19,7 @@ efficient data access.
 ### Adding the package to your product
 
 Add a dependency on Azure Storage Blob
+
 [//]: # ({x-version-update-start;com.azure:azure-storage-blob;current})
 ```xml
 <dependency>


sdk/storage/azure-storage-file-datalake/README.md
```diff
@@ -19,6 +19,7 @@ efficient data access.
 ### Adding the package to your product
 
 Add a dependency on Azure Storage Blob
+
```
srnagar: Is it possible to update `markdown2` parser or pre-process the `README.md` file before running it through `markdown2` to insert new lines as necessary? This can break again the next time someone else modifies this file and removes the new line and they wouldn't even know it. Also, new content will be added and there's no way anyone would know/remember about this new line requirement.

JimSuplizio: @srnagar so if you're using a markdown editor (like the markdownlint in VS Code) it'll actually give warnings about the spacing. It's funny because it literally only seems to have a problem "add a dependency" XML code block and I'd gone through the readme files a while back and thought I'd gotten them all. I've no intention, right now, on touching the markdown2 code to fix formatting errors that a linter would have caught.

PR #6249, opened 2019-11-08
Support custom application id in user agent telemetry policy 

### LINKED ISSUE #5842: Support custom application id in user agent  telemetry policy
Support custom applicationId that users can set in user agent, as per new guidelines for `UserAgentPolicy` defined here - https://github.com/Azure/azure-sdk/pull/614/files

Include length check for user-provided applicationId and verify no spaces are included in the appId.

samvaity:
Support custom `applicationId` that users can be set in the user agent, as per new guidelines for UserAgentPolicy defined here - https://github.com/Azure/azure-sdk/pull/614/files

Fixes #5842

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
@@ -10,6 +10,8 @@ import com.azure.core.http.HttpPipelineNextPolicy;
 import com.azure.core.http.HttpResponse;
 import reactor.core.publisher.Mono;
 
+import java.util.Objects;
+
 /**
  * Pipeline policy that adds "User-Agent" header to a request.
  *
@@ -68,9 +70,11 @@ public class UserAgentPolicy implements HttpPipelinePolicy {
      * @param sdkVersion Version of the client library.
      * @param version {@link ServiceVersion} of the service to be used when making requests.
      * @param configuration Configuration store that will be checked for the AZURE_TELEMETRY_DISABLED.
+     * @throws NullPointerException if {@code configuration} is {@code null}.
      */
     public UserAgentPolicy(String applicationId, String sdkName, String sdkVersion, Configuration configuration,
                            ServiceVersion version) {
+        Objects.requireNonNull(configuration, "'configuration' cannot be null.");
         boolean telemetryDisabled = configuration.get(Configuration.PROPERTY_AZURE_TELEMETRY_DISABLED, false);
         if (telemetryDisabled) {
             this.userAgent = String.format(DISABLED_TELEMETRY_USER_AGENT_FORMAT, sdkName, sdkVersion,
@@ -96,8 +100,10 @@ public class UserAgentPolicy implements HttpPipelinePolicy {
      * @param sdkVersion Version of the client library.
      * @param version {@link ServiceVersion} of the service to be used when making requests.
      * @param configuration Configuration store that will be checked for the AZURE_TELEMETRY_DISABLED.
+     * @throws NullPointerException if {@code configuration} is {@code null}.
      */
     public UserAgentPolicy(String sdkName, String sdkVersion, Configuration configuration, ServiceVersion version) {
+        Objects.requireNonNull(configuration, "'configuration' cannot be null.");
         boolean telemetryDisabled = configuration.get(Configuration.PROPERTY_AZURE_TELEMETRY_DISABLED, false);
         if (telemetryDisabled) {
             this.userAgent = String.format(DISABLED_TELEMETRY_USER_AGENT_FORMAT, sdkName, sdkVersion,


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
```diff
@@ -13,9 +15,13 @@
  * The log configurations for HTTP messages.
  */
 public class HttpLogOptions {
+    private String applicationId;
```
samvaity: Should we rename this file to `HttpClientOptions`?

alzimmermsft: That would be an API breaking change

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -59,19 +63,26 @@ public UserAgentPolicy(String userAgent) {
...
      * in the user agent.
      *
+     * @param applicationId User specified application Id.
      * @param sdkName Name of the client library.
      * @param sdkVersion Version of the client library.
      * @param version {@link ServiceVersion} of the service to be used when making requests.
      * @param configuration Configuration store that will be checked for the AZURE_TELEMETRY_DISABLED.
      */
-    public UserAgentPolicy(String sdkName, String sdkVersion, Configuration configuration, ServiceVersion version) {
+    public UserAgentPolicy(String applicationId, String sdkName, String sdkVersion, Configuration configuration,
```
alzimmermsft: This needs to be a new method overload

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -59,19 +63,26 @@ public UserAgentPolicy(String userAgent) {
...
             this.userAgent = String.format(DISABLED_TELEMETRY_USER_AGENT_FORMAT, sdkName, sdkVersion,
                 version.getVersion());
         } else {
-            this.userAgent = String.format(USER_AGENT_FORMAT, sdkName, sdkVersion, getPlatformInfo(),
-                version.getVersion());
+            if (applicationId == null) {
+                this.userAgent = String.format(USER_AGENT_FORMAT, sdkName, sdkVersion, getPlatformInfo(),
+                    version.getVersion());
+            } else {
+                this.userAgent = String.format(APPLICATION_ID_USER_AGENT_FORMAT, applicationId, sdkName, sdkVersion,
```
alzimmermsft: @mssfang Does this resolve the issue that the Spring Integration team wanted where they wanted to append their custom application ID to the User-Agent string

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -59,19 +63,26 @@ public UserAgentPolicy(String userAgent) {
...
+                    getPlatformInfo(), version.getVersion());
+            }
```
alzimmermsft: You'll need to update how the UserAgentPolicy's process method works, right now it is expecting that the header would begin with `azsdk`. If the header policy is incorrectly applied after a retry policy the customer could run into an issue where the User-Agent header eventually exceeds the allowed limit for a header.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
```diff
@@ -72,6 +79,38 @@ public HttpLogOptions setLogLevel(final HttpLogDetailLevel logLevel) {
...
+    }
+
+    /**
+     * Sets the custom application specific id supplied by the user of the client library.
+     *
+     * @param applicationId The user specified application id.
+     * @return The updated HttpLogOptions object.
+     */
+    public HttpLogOptions setApplicationId(final String applicationId) {
+        if (applicationId != null
```
alzimmermsft: I think the logic here could be simplified, right now this is going to perform the same check multiple times.

``` java
if (!CoreUtils.isNullOrEmpty(applicationId)) {  // Would an issue arise if the application ID is an empty string? Should we disallow that?
    if (applicationId.contains(" ")) {
        // throw error
    } else if (applicationId.length() > 24) {
        // throw error
    } else {
        this,applicationId = applicationId;
    }
}

return this;
```

sdk/core/azure-core/src/test/java/com/azure/core/UserAgentTests.java
```diff
@@ -57,4 +58,52 @@ public void customUserAgentTests() throws Exception {
...
+
+    @Test
+    public void customApplicationIdUserAgentTest() throws Exception {
+        final HttpPipeline pipeline = new HttpPipelineBuilder()
+            .httpClient(new NoOpHttpClient() {
+                @Override
+                public Mono<HttpResponse> send(HttpRequest request) {
+                    String header = request.getHeaders().getValue("User-Agent");
+                    String expectedHeaderPrefix = "user_specified_appId";
+                    Assertions.assertTrue(header.startsWith(expectedHeaderPrefix));
```
alzimmermsft: Could we validate the whole User-Agent string

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
mssfang: <platform_info> itself don't contains parentheses in it. It should be either make sure <platform_info> value contains parentheses at begin and end of itself or hard code parentheses around <platform_info>, 
```java
private static final String USER_AGENT_FORMAT = DEFAULT_USER_AGENT_HEADER + "-%s/%s (%s)";
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
samvaity: Updated it to check for if it `contains` `azsdk`.
And filed an issue for the header append issue on retry - https://github.com/Azure/azure-sdk-for-java/issues/6269

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogOptions.java
samvaity: Simplified the logic.
Currently, we are allowing empty application Id.
@JonathanGiles do we want to disallow empty application Id?

sdk/core/azure-core/src/test/java/com/azure/core/UserAgentTests.java
samvaity: added

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
samvaity: added as a new overload.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -14,7 +14,7 @@
  * Pipeline policy that adds "User-Agent" header to a request.
  *
  * The format for the "User-Agent" string is outlined in
- * <a href="https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy>Azure Core: Telemetry policy</a>.
+ * <a href="https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy>Azure Core: Telemetry policy </a>.
```
conniey: This space before the closing `</a>` isn't necessary

sdk/core/azure-core/src/test/java/com/azure/core/UserAgentTests.java
```diff
@@ -57,4 +58,61 @@ public void customUserAgentTests() throws Exception {
...
+                }
+            })
+            .policies(new UserAgentPolicy(
+                null,
+                "package.name",
+                "package_version",
+                Configuration.NONE,
+                () -> "1.0"))
+            .build();
+        HttpResponse response = pipeline.send(new HttpRequest(HttpMethod.GET,
```
conniey: Use StepVerifier rather than .block()

sdk/core/azure-core/src/test/java/com/azure/core/UserAgentTests.java
```diff
@@ -57,4 +58,61 @@ public void customUserAgentTests() throws Exception {
...
+                }
+            })
+            .policies(new UserAgentPolicy(
+                testAppId,
+                testSdkName,
+                testPackageVersion,
+                Configuration.NONE,
+                () -> "1.0"))
+            .build();
+        HttpResponse response = pipeline.send(new HttpRequest(HttpMethod.GET,
```
conniey: Use step verifier.

sdk/core/azure-core/src/test/java/com/azure/core/UserAgentTests.java
```diff
@@ -57,4 +58,61 @@ public void customUserAgentTests() throws Exception {
...
+            .policies(new UserAgentPolicy(
+                null,
+                "package.name",
+                "package_version",
+                Configuration.NONE,
+                () -> "1.0"))
+            .build();
+        HttpResponse response = pipeline.send(new HttpRequest(HttpMethod.GET,
+            new URL("http://localhost"))).block();
+        Assertions.assertEquals(200, response.getStatusCode());
```
conniey: I would also assert that the `response` contains the headers you expect to exist?

sdk/core/azure-core/src/test/java/com/azure/core/http/policy/HttpLogOptionsTest.java
```diff
@@ -0,0 +1,31 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class HttpLogOptionsTest {
+
+    @Test
+    public void testMaxApplicationId() {
+        IllegalArgumentException thrown = assertThrows(IllegalArgumentException.class,
+            () -> new HttpLogOptions().setApplicationId("AppId-0123456789012345678912345"));
+        assertEquals(thrown.getMessage(), "'applicationId' length cannot be greater than 24");
```
conniey: Asserting the actual message isn't a good idea because this could be localised or message changed which would break the test.

sdk/core/azure-core/src/test/java/com/azure/core/http/policy/HttpLogOptionsTest.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Test
+    public void testApplicationIdContainsSpace() {
+        IllegalArgumentException thrown = assertThrows(IllegalArgumentException.class,
+            () -> new HttpLogOptions().setApplicationId("AppId 78912345"));
+        assertEquals(thrown.getMessage(), "'applicationId' must not contain a space.");
```
conniey: Asserting the exception type is sufficient.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -23,6 +23,10 @@
     // azsdk-java-<client_lib>/<sdk_version> <platform_info>
     private static final String USER_AGENT_FORMAT = DEFAULT_USER_AGENT_HEADER + "-%s/%s %s";
 
+    // From the design guidelines, the application id user agent header format is:
```
JonathanGiles: The link in the JavaDoc for the class is missing its closing " at the end of the href. Can you please fix that.

sdk/keyvault/azure-security-keyvault-secrets/src/test/java/com/azure/security/keyvault/secrets/SecretClientTestBase.java
```diff
@@ -67,7 +67,7 @@ void beforeTestSetup() {
         HttpClient httpClient;
         // Closest to API goes first, closest to wire goes last.
         final List<HttpPipelinePolicy> policies = new ArrayList<>();
-        policies.add(new UserAgentPolicy(AzureKeyVaultConfiguration.SDK_NAME, AzureKeyVaultConfiguration.SDK_VERSION,  Configuration.getGlobalConfiguration().clone(), SecretServiceVersion.getLatest()));
+        policies.add(new UserAgentPolicy(null, AzureKeyVaultConfiguration.SDK_NAME, AzureKeyVaultConfiguration.SDK_VERSION,  Configuration.getGlobalConfiguration().clone(), SecretServiceVersion.getLatest()));
```
JonathanGiles: You could leave this code as it was, couldn't you?

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTestBase.java
```diff
@@ -74,7 +74,7 @@ void beforeTestSetup() {
 
         // Closest to API goes first, closest to wire goes last.
         final List<HttpPipelinePolicy> policies = new ArrayList<>();
-        policies.add(new UserAgentPolicy(AzureKeyVaultConfiguration.SDK_NAME, AzureKeyVaultConfiguration.SDK_VERSION,  Configuration.getGlobalConfiguration().clone(), CryptographyServiceVersion.getLatest()));
+        policies.add(new UserAgentPolicy(null, AzureKeyVaultConfiguration.SDK_NAME, AzureKeyVaultConfiguration.SDK_VERSION,  Configuration.getGlobalConfiguration().clone(), CryptographyServiceVersion.getLatest()));
```
JonathanGiles: Leave it as it was

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyClientTestBase.java
```diff
@@ -69,7 +69,7 @@ void beforeTestSetup() {
         HttpClient httpClient;
         // Closest to API goes first, closest to wire goes last.
         final List<HttpPipelinePolicy> policies = new ArrayList<>();
-        policies.add(new UserAgentPolicy(AzureKeyVaultConfiguration.SDK_NAME, AzureKeyVaultConfiguration.SDK_VERSION,  Configuration.getGlobalConfiguration().clone(), KeyServiceVersion.getLatest()));
+        policies.add(new UserAgentPolicy(null, AzureKeyVaultConfiguration.SDK_NAME, AzureKeyVaultConfiguration.SDK_VERSION,  Configuration.getGlobalConfiguration().clone(), KeyServiceVersion.getLatest()));
```
JonathanGiles: Leave it as it was.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTestBase.java
samvaity: updated

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
samvaity: updated

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -59,11 +63,46 @@ public UserAgentPolicy(String userAgent) {
...
+
+    /**
+     * Creates a UserAgentPolicy with the {@code sdkName} and {@code sdkVersion} in the User-Agent header value.
+     *
+     * If the passed configuration contains true for AZURE_TELEMETRY_DISABLED the platform information won't be included
+     * in the user agent.
+     *
+     * @deprecated This method is deprecated as it does not set the custom application id sent by user on the
+     * "User-Agent" header when sending a request.
+     * <p> Use {@link UserAgentPolicy#UserAgentPolicy(String, String, String, com.azure.core.util.Configuration,
```
JonathanGiles: Get rid of the <p> and move the explanation text directly after the last sentence finishes.

samvaity: Removed the whole text as we decided to not deprecate the older constructor.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -53,6 +57,35 @@ public UserAgentPolicy(String userAgent) {
...
+     *
+     * @param applicationId User specified application Id.
+     * @param sdkName Name of the client library.
+     * @param sdkVersion Version of the client library.
+     * @param version {@link ServiceVersion} of the service to be used when making requests.
+     * @param configuration Configuration store that will be checked for the AZURE_TELEMETRY_DISABLED.
+     */
+    public UserAgentPolicy(String applicationId, String sdkName, String sdkVersion, Configuration configuration,
+                           ServiceVersion version) {
+        boolean telemetryDisabled = configuration.get(Configuration.PROPERTY_AZURE_TELEMETRY_DISABLED, false);
```
alzimmermsft: Should `null` check the passed configuration.

samvaity: done

Issue #4955, opened 2019-08-12
Switch to JUnit version 5 for unit tests in azure-core


srnagar:
Migrate to JUnit 5. Jonathan has a draft PR that can be used as starting point.
https://github.com/Azure/azure-sdk-for-java/pull/4483

mclay:
hi @JonathanGiles, could you pls explain briefly what's left to be done here so i can finish it - pls. assign me this task

JonathanGiles:
@mclay This is a relatively big task. For all `/sdk/<service>/azure-*` libraries, we remove junit 4 and replace it with junit 5. This will require some changes to the tests in some places as the JUnit 4 and JUnit 5 APIs differ somewhat. This could be decomposed into a number of sub-tasks, e.g. firstly focusing on `/sdk/core/azure-core` and the other core modules. Separate PRs could then be developed for Key Vault, Storage, etc. Because we only focus on the 'next-gen' client libraries, we only care about modules which begin with `azure-`.

If you're keen to take this on, we would love to accept your pull requests! I've assigned the issue to you.

joshfree:
@mclay just a quick note that this issue doesn't apply to /sdk/storage as storage is using groovy tests and not JUnit

JonathanGiles:
@mclay Let me know if you plan to pick this up and work on it, otherwise I'll reassign to myself or someone else so that we may begin to make this move. Thanks!

mclay:
Hi @JonathanGiles i am on it and provide the first Bits of it until the End of this Week

JonathanGiles:
Excellent news! Thanks for the update.

PR #6255, opened 2019-11-10
Switch to JUnit version 5 for unit tests #4955

mclay:
- migrate unit tests in sdk/eventhubs/*

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubMessageSerializerTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubMessageSerializerTest.java
@@ -190,22 +190,22 @@ public class EventHubMessageSerializerTest {
      */
     @Test
     public void throwsWhenIncorrectTypeInResponse() {
+        // Arrange
+        final Long eventHubName = 100L;
+        final Date createdAtAsDate = new Date(1569275540L);
+        final String[] partitionIds = new String[]{"1", "foo", "bar", "baz"};
+
+        final Map<String, Object> values = new HashMap<>();
+        values.put(ManagementChannel.MANAGEMENT_ENTITY_NAME_KEY, eventHubName);
+        values.put(ManagementChannel.MANAGEMENT_RESULT_CREATED_AT, createdAtAsDate);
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubMessageSerializerTest.java
```diff
@@ -181,56 +179,60 @@ public void deserializeEventHubProperties() {
...
 
-        final Message message = Proton.message();
-        message.setBody(amqpValue);
+            final Message message = Proton.message();
+            message.setBody(amqpValue);
 
-        // Act
-        serializer.deserialize(message, EventHubProperties.class);
+            // Act
+            serializer.deserialize(message, EventHubProperties.class);
```
conniey: Only this line should be in the `assertThrows(() -> {})`, to scope down where we expect the error to be.

sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubMessageSerializerTest.java
```diff
@@ -181,56 +179,60 @@ public void deserializeEventHubProperties() {
...
```
conniey: Only this line should be in the `assertThrows(() -> {})`, to scope down where we expect the error to be.

sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubMessageSerializerTest.java
mclay: thanks for the review. changes are incorporated

sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubMessageSerializerTest.java
mclay: thanks for the review. changes are incorporated

JonathanGiles:
Thanks again @mclay!

PR #6258, opened 2019-11-11
Adding synchronous receive all and tests

### LINKED ISSUE #6236: Implement method in EventHubConsumerClient to receive from all partitions
Implement method in EventHubConsumerClient to receive from all partitions with two overloads - one with default "earliest" event position and second with a user-provided event position (same position will be used for all partitions).

conniey:
* Adding `EventHubConsumerClient.receive(int)` and `EventHubConsumerClient.receive(int, Duration)`.
* Adding integration and unit tests for EventHubConsumer and EventHubProducer

Fixes #6236

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubConsumerAsyncClientIntegrationTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubConsumerAsyncClientIntegrationTest.java
@@ -11,6 +11,7 @@ import com.azure.messaging.eventhubs.models.PartitionContext;
 import com.azure.messaging.eventhubs.models.PartitionEvent;
 import com.azure.messaging.eventhubs.models.SendOptions;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
 import reactor.core.Disposable;
 import reactor.core.Disposables;
@@ -20,6 +21,8 @@ import reactor.test.StepVerifier;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -41,13 +44,10 @@ import static com.azure.messaging.eventhubs.EventHubClientBuilder.DEFAULT_CONSUM
  * @see EventPositionIntegrationTest
  */
 public class EventHubConsumerAsyncClientIntegrationTest extends IntegrationTestBase {
-    private static final String PARTITION_ID = "0";
     private static final String PARTITION_ID_HEADER = "SENT_PARTITION_ID";
 
     private final String[] expectedPartitionIds = new String[]{"0", "1"};
 
-    // The maximum number of receivers on a partition + consumer group is 5.
-    private static final int MAX_NUMBER_OF_CONSUMERS = 5;
     private static final String MESSAGE_TRACKING_ID = UUID.randomUUID().toString();
 
     private EventHubAsyncClient client;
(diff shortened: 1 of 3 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConsumerClient.java
```diff
@@ -175,6 +183,60 @@ public PartitionProperties getPartitionProperties(String partitionId) {
...
+        Objects.requireNonNull(maximumWaitTime, "'maximumWaitTime' cannot be null.");
+
+        if (maximumMessageCount < 1) {
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("'maximumMessageCount' cannot be less than 1."));
+        } else if (maximumWaitTime.isNegative() || maximumWaitTime.isZero()) {
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("'maximumWaitTime' cannot be zero or less."));
+        }
+
```
srnagar: nit: remove extra empty line

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConsumerClient.java
```diff
@@ -175,6 +183,60 @@ public PartitionProperties getPartitionProperties(String partitionId) {
...
+        final Flux<PartitionEvent> events = Flux.create(emitter -> {
+            queueWork(RECEIVE_ALL_KEY, maximumMessageCount, maximumWaitTime, emitter);
+        });
+
+        final Flux<PartitionEvent> map = events.collectList().map(x -> {
+            logger.info("Number of events received: {}", x.size());
+            return Flux.fromIterable(x);
+        }).block();
+        return new IterableStream<>(map);
```
srnagar: Any reason why we can't just `return new IterableStream<>(events.block())` here?

sdk/eventhubs/pom.service.xml
```diff
@@ -31,8 +31,6 @@
         </property>
       </activation>
       <modules>
-        <!-- NOTE: Commenting out because we should now be binary composed.
```
srnagar: 👍

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConsumerClient.java
```diff
@@ -119,41 +122,46 @@ public EventHubProperties getProperties() {
...
+     * @return A set of {@link PartitionEvent} that was received. The iterable contains up to
+     *     {@code maximumMessageCount} events.
+     *
      * @throws NullPointerException if {@code maximumWaitTime} is null.
      * @throws IllegalArgumentException if {@code maximumMessageCount} is less than 1 or {@code maximumWaitTime} is
      *     zero or a negative duration.
      */
     public IterableStream<PartitionEvent> receive(String partitionId, int maximumMessageCount,
-            Duration maximumWaitTime) {
+        Duration maximumWaitTime) {
```
srnagar: .NET uses `ReadFromPartitionAsync`. While I don't like `read`, should we use `receiveFromPartition` to be closer to what .NET is using?

conniey: I'm going to defer these naming changes until we've finished our review because it results in a lot of changed code. I'll create an issue for it though.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConsumerClient.java
conniey: Good catch, I used it before for debugging.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConsumerClient.java
```diff
@@ -175,6 +183,55 @@ public PartitionProperties getPartitionProperties(String partitionId) {
...
+        } else if (maximumWaitTime.isNegative() || maximumWaitTime.isZero()) {
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("'maximumWaitTime' cannot be zero or less."));
+        }
+
+        final Flux<PartitionEvent> events = Flux.create(emitter -> {
+            queueWork(RECEIVE_ALL_KEY, maximumMessageCount, maximumWaitTime, emitter);
+        });
+
+        return new IterableStream<>(events);
```
srnagar: The `map` operation can be removed in `receive(String partitionId, int maximumMessageCount, Duration maximumWaitTime)` overload as well.

PR #6275, opened 2019-11-12
Fixes authorization when using client credentials.

### LINKED ISSUE #5447: The client created by using host, enventhubname and Credential doesn't work
We have a customer reporting this issue. Alex forwarded this to us.

image

Error:          Note: Use ConnectString to    create EventhubAsyncClient works well

conniey:
* Uses correct audience and scope when authorizing with Azure Event Hubs.
* Adds tests to ensure that it works.

**TODO**
- [ ] Add Integration test.

Fixes #5447


sdk/eventhubs/azure-messaging-eventhubs/pom.xml
```diff
@@ -46,6 +46,12 @@
     </dependency>
 
     <!-- Test dependencies -->
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-identity</artifactId>
+      <version>1.1.0-beta.1</version> <!-- {x-version-update;com.azure:azure-identity;dependency} -->
```
srnagar: Prefer to take the released version dependency instead of `beta`

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/CBSNode.java
```diff
@@ -18,10 +18,11 @@
     /**
      * Authorizes the caller with the CBS node to access resources for the {@code audience}.
      *
-     * @param audience Resource that the callee needs access to.
+     * @param audience The audience to which the token applies. This can be the path within the AMQP message broker.
+     * @param scopes The requested scopes for the {@link com.azure.core.credential.TokenCredential}.
```
srnagar: nit: use short class name instead.

sdk/eventhubs/azure-messaging-eventhubs/pom.xml
conniey: I copied it from storage. . . Should we be using the released one there too?
https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/pom.xml#L92-L97

PR #6277, opened 2019-11-12
Getting rid of the azure-core-test module.

### LINKED ISSUE #6268: Investigate Removal of Azure Core Test Module
Investigate the removal of the Azure Core Test module and move the code into the test directory of Azure Core. Having Azure Core Test as a separate module introduces additional risk for dependency conflicts between the SDKs we release and it should always be pulled in as a testing dependency. 

It would be better for dependency management, release overhead, and API maintenance to have Azure Core Test merged into Azure Core. Azure Core is already the common denominator between the SDKs we release and this would allow for it to be pulled in as a testing dependency where needed to provide the functionality that Azure Core Test currently provides.

JonathanGiles:
Fixes #6268


sdk/core/azure-core/src/test/java/com/azure/core/implementation/InterceptorManager.java
```diff
@@ -1,13 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-package com.azure.core.test;
+package com.azure.core.implementation;
```
alzimmermsft: Do we want this under implementation? Doesn't this Force us to export this in the module-info again to every other module that will use it?

alzimmermsft: Same goes for other classes accessed outside of Azure Core. I could see a refactoring to have TestBase have wrapper method around this and other classes such as TestResourceNamer.

JonathanGiles: I'm not as worried about this given these are test classes and not part of the public API or module system.

pom.client.xml
```diff
@@ -967,6 +974,6 @@
     <module>sdk/storage/azure-storage-file-share</module>
     <module>sdk/storage/azure-storage-file-datalake</module>
     <module>sdk/storage/azure-storage-queue</module>
-    <module>sdk/storage/azure-storage-queue-cryptography</module>
+<!--    <module>sdk/storage/azure-storage-queue-cryptography</module>-->
```
conniey: Did you mean to comment this out?

sdk/appconfiguration/pom.service.xml
```diff
@@ -9,10 +9,9 @@
   <packaging>pom</packaging>
   <version>1.0.0</version><!-- Need not change for every release-->
   <modules>
-    <!-- NOTE: Commenting out because we should now be binary composed. 
+    <!-- NOTE: Commenting out because we should now be binary composed.
```
conniey: Does this comment apply? We're still compiling from modules.

pom.client.xml
JonathanGiles: Yep. There is no code in this module. Alan said it would likely be removed soon, so I'm avoiding build errors by commenting this out.

sdk/appconfiguration/pom.service.xml
JonathanGiles: Engsys have aspirations here, so I'm leaving it in for now.

weshaggard:
Just calling out the obvious but this change now requires that all the track 2 libraries stick with a live dependency on Azure.Core and thus we cannot reference the 1.0.0 GA of Azure.Core any longer and thus we cannot ship any of the track 2 libraries again without shipping Azure.Core. While that might be OK for this change it is something we need to keep in mind for future changes.

sdk/core/azure-core/src/test/java/com/azure/core/implementation/InterceptorManager.java
alzimmermsft: Could we move them to `implementation.test`?

pom.client.xml
```diff
@@ -815,8 +823,8 @@
                 --add-opens com.azure.core/com.azure.core.implementation.entities=com.fasterxml.jackson.databind
                 --add-opens com.azure.core/com.azure.core.implementation.entities=ALL-UNNAMED
               </argLine>
-              <useSystemClassLoader>false</useSystemClassLoader>
-              <forkCount>1</forkCount>
+              <useSystemClassLoader>true</useSystemClassLoader>
+<!--              <forkCount>1</forkCount>-->
```
srnagar: This used to break some of the storage tests if the forkCount is not set to 1. Will this work for storage tests now?

sdk/core/azure-core/src/test/java/com/azure/core/implementation/InterceptorManager.java
srnagar: Since these are in `src/test` directory, `exports <packagename>` will not include these classes. So, `implementation` or `implementation.test` doesn't make a difference.

pom.client.xml
```diff
@@ -951,7 +959,6 @@
 <!--    <module>sdk/core/azure-core-management</module>--> <!-- Removed for now, as this library is unused -->
     <module>sdk/core/azure-core-http-netty</module>
     <module>sdk/core/azure-core-http-okhttp</module>
-    <module>sdk/core/azure-core-test</module>
```
srnagar: This will require a bump in version number before the next release.

sdk/storage/azure-storage-common/pom.xml
```diff
@@ -58,12 +58,6 @@
       <scope>provided</scope>
     </dependency>
 
-    <dependency>
```
srnagar: Don't we need the azure-core `test-jar` dependency here now that `azure-core-test` is removed? Or was this dependency added eagerly before it was required?

sdk/storage/azure-storage-queue-cryptography/pom.xml
```diff
@@ -40,12 +40,6 @@
     </dependency>
 <!--    <dependency>-->
```
srnagar: nit: remove these commented dependencies as well.

JonathanGiles:
Currently the build fails on JDK 9+, because the test configuration can't find the classes in the test-jar that is generated. Further exploration is required.

PR #6283, opened 2019-11-12
Fix Build.Reason for several steps

JimSuplizio:
The Build.Reason in the file was "Scheduled" however according to the predefined build variables in the [pipeline doc](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables), it should be "Schedule"


JimSuplizio:
I verified this fix by creating a private pipeline and scheduling a build. Everything looked good so I'm going to merge this.

PR #6303, opened 2019-11-12
Azure ARM Poller implementation for Azure.core PollerFlux

anuchandy:


anuchandy:
@JonathanGiles - PR contains `PollerFlux` implementation for Azure ARM polling algorithm. 

PR also demonstrate how to dehydrate the polling from a string.

Note: I worked only the parts specific to Polling and cleanup related to that. It looks like we can almost remove AzureProxy, mot of the code in was related to polling. The module is still excluded from the main project.

anuchandy:
@joshfree since azure-mgmt is in our rader, generally its good for all languages to analyze how they're going to adapt azure-core poller concepts in ARM poller, basically validating whether the current azure-core poller language specific abstraction is sufficient to enable it.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/PagedList.java
```diff
@@ -1,415 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
yaohaizh: If this class is removed, what is the suggested replacement or usage for the paged list?

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/Error.java
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
yaohaizh: Need default constructor for serialize/deserialize of pollstate object.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/FinalResult.java
```diff
@@ -0,0 +1,50 @@
...
+    @JsonIgnore
+    private static final ClientLogger logger = new ClientLogger(FinalResult.class);
+
+    @JsonProperty(value = "resultUri")
+    private URL resultUri;
+    @JsonProperty(value = "result")
+    private String result;
+
+    /**
+     * Creates FinalResult.
```
yaohaizh: Default pollstate constructor

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/LocationData.java
```diff
@@ -0,0 +1,94 @@
...
+    @JsonProperty(value = "pollUrl", required = true)
+    private URL pollUrl;
+    @JsonProperty(value = "provisioningState", required = true)
+    private String provisioningState;
+    @JsonProperty(value = "pollError")
+    private Error pollError;
+    @JsonProperty(value = "finalResult")
+    private FinalResult finalResult;
+
+    /**
```
yaohaizh: Default constructor for serialize/deserialization

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/PollerFactory.java
```diff
@@ -0,0 +1,269 @@
...
+        Duration defaultPollInterval,
+        Function<PollingContext<PollResult<T>>, Mono<PollResult<T>>> lroInitOperation) {
+        Objects.requireNonNull(serializerAdapter, "'serializerAdapter' cannot be null.");
+        Objects.requireNonNull(pipeline, "'pipeline' cannot be null.");
+        Objects.requireNonNull(pollResultType, "'pollResultType' cannot be null.");
+        Objects.requireNonNull(finalResultType, "'finalResultType' cannot be null.");
+        Objects.requireNonNull(defaultPollInterval, "'defaultPollInterval' cannot be null.");
+        Objects.requireNonNull(lroInitOperation, "'lroInitOperation' cannot be null.");
+
+        return new PollerFlux<>(defaultPollInterval,
```
yaohaizh: The most struggle me when I tried this PR is about lroInitOperation that will succeed without polling operation.  

1. Current the PollerFlux/PollerFlux doesn't provide the short circuit when the init operation doesn't need polling operation. It will run the poll function anyway with delayed retry time.
2. Secondly, the Response<T> object from RestProxy will lost the string of the body, in some case, the some information will lost during pollingstate serialization and deserialization because of the properties are marked with WRITE_ONLY.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/AzureProxy.java
```diff
@@ -1,421 +0,0 @@
...
-            new SwaggerInterfaceParser(swaggerInterface, createDefaultSerializer(), baseUrl);
-        final AzureProxy azureProxy = new AzureProxy(httpPipeline, interfaceParser);
-        return (A) Proxy
-            .newProxyInstance(swaggerInterface.getClassLoader(), new Class<?>[]{swaggerInterface}, azureProxy);
-    }
-
-    @Override
-    protected Object handleHttpResponse(final HttpRequest httpRequest, Mono<HttpDecodedResponse> asyncHttpResponse,
-        final SwaggerMethodParser methodParser, Type returnType, Context context) {
-        if (TypeUtil.isTypeOrSubTypeOf(returnType, Flux.class)) {
```
yaohaizh: Without this method, the client will lost the flexibility to access the original response body.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/AzureProxy.java
```diff
@@ -1,421 +0,0 @@
...
-        String baseUrl = null;
-
-        if (azureEnvironment != null) {
-            final AzureHost azureHost = swaggerInterface.getAnnotation(AzureHost.class);
-            if (azureHost != null) {
-                baseUrl = azureEnvironment.url(azureHost.endpoint());
-            }
-        }
-
-        final SwaggerInterfaceParser interfaceParser =
```
yaohaizh: This is another convenient that the AzureProxy provide for host URL. Since the other libraries cannot access the SwaggerInterfaceParser, each RestService method have to add the @HostParam("url") parameter, which is not developer friendly on the service client side.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/PagedList.java
anuchandy: Now the equivalent type is in azure-core - `PagedFlux` (async), `PagedIterable` (sync), hence this is no longer required in core mgmt.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/AzureProxy.java
anuchandy: My thought here was - The generated LRO methods in "private" service interface will be returning `Mono<ByteBuffer>`, so azure-core won't touch the body. The "LRO initialization Function" will be using this method and will have access to the raw-bytes.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/PollerFactory.java
anuchandy: 1. Agree, we're missing a short circuit for synchronously completed LRO and imposes delay for the first poll that won't make any network call. Two options I could thinking of addressing with without breaking public surface: (1). Not to apply delay before the first poll and immediately call into poll function, it can check for completion state and return (2). Use PollerContext to signal synchronous completion, we check for that flag before starting polling loop.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/AzureProxy.java
anuchandy: To add some context: the only reason we were overriding `handleHttpResponse` was to handle LRO return types. We no longer need it in the RestProxy level because (1). as mentioned in the previous comment we will return `Flux<ByteBuffer>` from service interface LRO proxy methods (2). then in the generated code, we will use `PollerFactory` to create and return `PollerFlux`.

Please let me know if you see an issue with this approach.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/Error.java
```diff
@@ -0,0 +1,47 @@
...
+package com.azure.core.management.implementation.polling;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+/**
+ * Type holding error details of long-running-operation (LRO) or polling-operation of an LRO.
+ */
+class Error {
```
weidongxu-microsoft: Should we use a different class name? This might be a bit too easy to be mistaken with `java.lang.Error`.

Not a big issue since it is not public.

weidongxu-microsoft:
I've tried with proxy method return raw `Mono<SimpleResponse<Flux<ByteBuffer>>>`. It works somehow. But I still have a few questions.

1
`lroInitOperation` in `PollerFactory.create` is quite painful to write (given the double async in Mono response + Flux bytes). I think this method could be provided by mgmt module (at least provide a default implementation).

2
I am not sure about which Type to provide for `pollResultType`.
Some LRO would have different polling result based on resource state, which we do not know beforehand.
E.g. `/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}/providers/Microsoft.Resources/deployments/{deploymentName}` would have code 200 + complete resouce JSON when resource exists and needs update, alternatively code 201 + `{"status":"Running"}` when resource needs to be created.

I think usually (or at least as a fallback) the raw response would be enough for the intermittent poll result.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/PollingState.java
```diff
@@ -170,7 +170,7 @@ public static PollingState from(SerializerAdapter serializerAdapter, String valu
     /**
      * @return the current status of the long-running-operation.
      */
-    LongRunningOperationStatus getOperationStatus() {
+    public LongRunningOperationStatus getOperationStatus() {
```
anuchandy: @yaohaizh - consumers are suppose to call getStatus() on AsyncPollRrsponse,PollRrsponse which returns LongRunningStatus.  Are we making LongRunningOperation::getOperationStatus() public for any internal (within mgmt sdk) use?

yaohaizh: There is the first factory method in this file that PollState is not set internally in the PollerFactory file, thus, our MGMT code is such as https://github.com/Azure/azure-libraries-for-java/commit/09e3cfad5fa3800ab7b746ffed1e5513df7d25e9#diff-2cfa33e57f52c488063ba39769edcac8R131

yaohaizh: I try to put above code into PollerFactory to totally handle PollState, but it seems not easy given the current method signature.

anuchandy: Ok got it, so to enable synchronous completion scenario: along with the [feature](https://github.com/Azure/azure-sdk-for-java/pull/7599) you requested, we also need `PollingState:: getOperationStatus ()` to be accessible. This should be fine - given `PollingState` is in impl package and customers are not suppose to use it directly. 

After couple of lroInitOperation use cases for various services, if we find we can generalize init op then we could have a version of factory with default initOp baked-in.

yaohaizh: Agree, current the design is not mature since it manipulates the PollState in different places. Ideally, the acess level change should be avoided. 

The https://github.com/Azure/azure-libraries-for-java/blob/fix_tests/azure-mgmt-resources/src/main/java/com/azure/management/AzureServiceClient.java `getLroResultAsync` & `activationOperation` has been tested aganist ARM/storage services and works well. 

Please help generalizes these methods into the azure-core-management if you think it is more reasonable. Thanks!

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/PollerFactory.java
```diff
@@ -0,0 +1,321 @@
...
+        Type finalResultType,
+        Duration defaultPollInterval,
+        Mono<Response<Flux<ByteBuffer>>> lroInitMono) {
+        Objects.requireNonNull(serializerAdapter, "'serializerAdapter' cannot be null.");
+        Objects.requireNonNull(pipeline, "'pipeline' cannot be null.");
+        Objects.requireNonNull(pollResultType, "'pollResultType' cannot be null.");
+        Objects.requireNonNull(finalResultType, "'finalResultType' cannot be null.");
+        Objects.requireNonNull(defaultPollInterval, "'defaultPollInterval' cannot be null.");
+        Objects.requireNonNull(lroInitMono, "'lroInitMono' cannot be null.");
+        Function<PollingContext<PollResult<T>>, Mono<PollResult<T>>> defaultLroInitOperation = context -> lroInitMono
```
anuchandy: @yaohaizh what's your thought on this version of factory method that provide default lro init?

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/PollingState.java
anuchandy: thanks for the example, how about [this](https://github.com/Azure/azure-sdk-for-java/pull/6303/commits/07f909cb446aac55c5ce4f59ec36635a2436d859)

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/PollerFactory.java
weidongxu-microsoft: My wish is that we could re-use the code in core for the deserialization (https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/RestProxy.java#L496). It is quite complicated logic there (handle wired type, handle HEAD method, etc.). We probably not use that often, but it could be a hidden deviation of implementation.

I know it is really really hard to reuse core code... Just asking for advice.

yaohaizh: @anuchandy  sorry for delay. For this change, it is not resolve the operation status problem yet and the client have no way to determine the case of 200/201 cases.

yaohaizh: @anuchandy It seems we still need more change after the PR of https://github.com/Azure/azure-sdk-for-java/pull/7599 I'm okay with this change. @weidongxu-microsoft , could you help double check this PR from your side. Thanks.

anuchandy: @yaohaizh you're right, https://github.com/Azure/azure-sdk-for-java/pull/7599 needs to be in for fully support 200/201 . At the moment team is doing some critical fixes in azure-core, I will discuss this with Jonathan and get back on merge plan and release of azure-core with initOp.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/Error.java
JonathanGiles: I agree - this name should be more specific. `PollingError` or similar.

sdk/core/azure-core-management/pom.xml
```diff
@@ -44,6 +44,8 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <legal><![CDATA[[INFO] Any downloads listed may be third party software.  Microsoft grants you no rights for third party software.]]></legal>
+    <jacoco.min.branchcoverage>0.21</jacoco.min.branchcoverage>
+    <jacoco.min.linecoverage>0.24</jacoco.min.linecoverage>
```
JonathanGiles: What is the expected timeframe to remove these two lines?

anuchandy:
@yaohaizh with LROInit in place, should we make PollingState::getOperationStatus() package private?

yaohaizh:
> @yaohaizh with LROInit in place, should we make PollingState::getOperationStatus() package private?

right, just reverted.

anuchandy:
@yaohaizh PR lgtm. (I can't approve it since it was originally opened by me)

yaohaizh:
@anuchandy the PR should be ready for merge.But the master azure-core  ci is failing because of other changs. Let me know if this is okay to merge this PR.

anuchandy:
@yaohaizh, the core test failure is fixed now. Requesting rerun.

anuchandy:
@yaohaizh all checks are green, should be good to merge.

yaohaizh:
@anuchandy, thanks!

Issue #6307, opened 2019-11-12
Consistency Layer and collection cache optional null value change


kushagraThapar:
**Is your feature request related to a problem? Please describe.**
Flux / Mono doesn't support null values. In order to keep the consistency and caching logic same between different versions of SDK, we need to update the consistency layer and caching layer with value holder objects, which can contain null values, and then can be passed around API calls. 

**Describe the solution you'd like**
Use value holder objects which can contain null values. 

**Describe alternatives you've considered**
Current alternate is to use switchIfEmpty() or Mono.optional or Flux.optional (which is empty() internally)

But this has some issues.

kushagraThapar:
This has been fixed on both v2 and v3, closing it now.

PR #6312, opened 2019-11-13
Null Value holder change across collection cache and address cache layer

kushagraThapar:
* Updated collection layer logic with same as V2 SDK
* Removed Mono.empty() and Flux.empty() where V2 SDK was returning null values, instead used a ValueHolder<>(null)
* Addresses the issue: https://github.com/Azure/azure-sdk-for-java/issues/6307

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosResponse.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosResponse.java
@@ -107,7 +107,7 @@ public class CosmosResponse<T extends Resource> {
      *
      * @return diagnostics information for the current request to Azure Cosmos DB service.
      */
-    public CosmosResponseDiagnostics getCosmosResponseDiagnosticsString() {
+    public CosmosResponseDiagnostics getCosmosResponseDiagnostics() {
         return resourceResponseWrapper.getCosmosResponseDiagnostics();
     }
 
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/CollectionCrudTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/CollectionCrudTest.java
@@ -265,12 +265,12 @@ public class CollectionCrudTest extends TestSuiteBase {
             CosmosItemRequestOptions options = new CosmosItemRequestOptions();
             options.setPartitionKey(new PartitionKey("mypkValue"));
             CosmosAsyncItemResponse readDocumentResponse = item.read(options).block();
-            logger.info("Client 1 READ Document Client Side Request Statistics {}", readDocumentResponse.getCosmosResponseDiagnosticsString());
+            logger.info("Client 1 READ Document Client Side Request Statistics {}", readDocumentResponse.getCosmosResponseDiagnostics());
             logger.info("Client 1 READ Document Latency {}", readDocumentResponse.getRequestLatency());
 
             BridgeInternal.setProperty(document, "name", "New Updated Document");
             CosmosAsyncItemResponse upsertDocumentResponse = collection.upsertItem(document).block();
-            logger.info("Client 1 Upsert Document Client Side Request Statistics {}", upsertDocumentResponse.getCosmosResponseDiagnosticsString());
+            logger.info("Client 1 Upsert Document Client Side Request Statistics {}", upsertDocumentResponse.getCosmosResponseDiagnostics());
             logger.info("Client 1 Upsert Document Latency {}", upsertDocumentResponse.getRequestLatency());
 
             //  DELETE the existing collection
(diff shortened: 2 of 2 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/caches/RxCollectionCache.java
```diff
@@ -117,19 +128,20 @@ public void refresh(String resourceAddress, Map<String, Object> properties) {
 
                     });
         }
-        return Mono.empty();
+        return Mono.just(new Utils.ValueHolder<>(null));
```
simplynaveen20: I think ValueHolder has initialize method to create new object

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/caches/RxPartitionKeyRangeCache.java
```diff
@@ -52,89 +53,100 @@ public RxPartitionKeyRangeCache(AsyncDocumentClient client, RxCollectionCache co
...
-                    }
-
-                    return Mono.error(err);
-                });
+                              .map(Utils.ValueHolder::new)
+                              .onErrorResume(err -> {
+                                  logger.debug("tryLookupAsync on collectionRid {} encountered failure", collectionRid, err);
+                                  CosmosClientException dce = Utils.as(err, CosmosClientException.class);
+                                  if (dce != null && Exceptions.isStatusCode(dce, HttpConstants.StatusCodes.NOTFOUND)) {
+                                      return Mono.just(new Utils.ValueHolder<>(null));
```
simplynaveen20: Same here and other similar places ,  we could use initialize

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/caches/RxCollectionCache.java
kushagraThapar: Yes it does, in fact I used that initially, but it creates more confusion as if it was initializing something big. I just wanted to make it clear and concise, so used constructor :)

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/caches/RxCollectionCache.java
```diff
@@ -117,19 +128,20 @@ public void refresh(String resourceAddress, Map<String, Object> properties) {
...
 
-        return this.collectionInfoByIdCache.getAsync(
-                collectionResourceId,
-                null,
-                () -> this.getByRidAsync(collectionResourceId, properties));
+        Mono<DocumentCollection> async = this.collectionInfoByIdCache.getAsync(
+            collectionResourceId,
+            null,
+            () -> this.getByRidAsync(collectionResourceId, properties));
+        return async.map(Utils.ValueHolder::new);
```
moderakh: `async.map(Utils.ValueHolder::new)` seems odd.

If `async` Mono may return a null value, then by the time the follow has reached `map(Utils.ValueHolder::new);` Mono will explode.

if `async` Mono can never return a null value, then why doing `async.map(Utils.ValueHolder::new);` at all?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/AddressResolver.java
```diff
@@ -232,20 +233,25 @@ private static void ensureRoutingMapPresent(
...
                 new PartitionKeyRangeIdentity(collection.getResourceId(), range.getId()),
                 forceRefreshPartitionAddresses);
 
-            return addressesObs.flatMap(addresses -> Mono.just(new ResolutionResult(range, addresses))).switchIfEmpty(Mono.defer(() -> {
-                logger.info(
+            return addressesObs.flatMap(addressesValueHolder -> {
+
+                if (addressesValueHolder.v == null) {
+                    logger.info(
                         "Could not resolve addresses for identity {}/{}. Potentially collection cache or routing map cache is outdated. Return empty - upper logic will refresh and retry. ",
```
moderakh: update the log message accordingly. "Return null" instead of "return empty"

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/caches/RxCollectionCache.java
kushagraThapar: Good point, but `async` Mono can never return a null value. 
The reason for transforming the the `async` value to Utils.ValueHolder is because of the fact that whoever calls this method also returns null values in some other cases. 
If we don't do the transformation here, we will have to do the transformation at the caller level, and there are multiple callers, so that will be code duplication. 
So in my opinion it is better to do the transformation at one place than everywhere else. 
The same comment applies to below comments as well.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/AddressResolver.java
kushagraThapar: Will fix this in next iteration. Thanks for catching this.

PR #6320, opened 2019-11-13
Update addLinks support for eventhubs in Java tracing package

### LINKED ISSUE #6188: Fix `Links.create`  for opentelemetry
Currently, `addLink` feature for opentelemetry package is not working as expected. It might need additional fixes on the opensource opentelemetry package.
Fix and update the `azure-core-tracing-opentelemetry` package accordingly.

samvaity:
`AddLinks` is needed when event processor is exporting batch messages to the user to create a link between the messages in the same batch.
But in Java SDK since the SDK supports event processing loop and passes a single message to user, it should just create span for message processing and make it a child of context from message (or orphan if there is no context).

Fixes #6188

### DIFF
--- a/sdk/core/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/OpenCensusTracer.java
+++ b/sdk/core/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/OpenCensusTracer.java
@@ -96,7 +96,7 @@ public class OpenCensusTracer implements com.azure.core.util.tracing.Tracer {
     @Override
     public void end(int responseCode, Throwable throwable, Context context) {
         Objects.requireNonNull(context, "'context' cannot be null.");
-        final Span span = getOrDefault(PARENT_SPAN_KEY, null, Span.class, context);
+        final Span span = getOrDefault(context, PARENT_SPAN_KEY, null, Span.class);
         if (span == null) {
             return;
         }
@@ -118,7 +118,7 @@ public class OpenCensusTracer implements com.azure.core.util.tracing.Tracer {
             return;
         }
 
-        final Span span = getOrDefault(PARENT_SPAN_KEY, null, Span.class, context);
+        final Span span = getOrDefault(context, PARENT_SPAN_KEY, null, Span.class);
         if (span != null) {
             span.putAttribute(key, AttributeValue.stringAttributeValue(value));
         } else {
@@ -139,7 +139,7 @@ public class OpenCensusTracer implements com.azure.core.util.tracing.Tracer {
      */
     @Override
     public void end(String statusMessage, Throwable throwable, Context context) {
-        final Span span = getOrDefault(PARENT_SPAN_KEY, null, Span.class, context);
+        final Span span = getOrDefault(context, PARENT_SPAN_KEY, null, Span.class);
         if (span == null) {
             logger.warning("Failed to find span to end it.");
             return;
@@ -157,13 +157,13 @@ public class OpenCensusTracer implements com.azure.core.util.tracing.Tracer {
      */
     @Override
     public void addLink(Context context) {
-        final Span span = getOrDefault(PARENT_SPAN_KEY, null, Span.class, context);
+        final Span span = getOrDefault(context, PARENT_SPAN_KEY, null, Span.class);
         if (span == null) {
             logger.warning("Failed to find span to link it.");
             return;
         }
 
-        final SpanContext spanContext = getOrDefault(SPAN_CONTEXT_KEY, null, SpanContext.class, context);
+        final SpanContext spanContext = getOrDefault(context, SPAN_CONTEXT_KEY, null, SpanContext.class);
         if (spanContext == null) {
             logger.warning("Failed to find span context to link it.");
             return;
@@ -193,7 +193,7 @@ public class OpenCensusTracer implements com.azure.core.util.tracing.Tracer {
     private Context startScopedSpan(String spanName, Context context) {
         Objects.requireNonNull(context, "'context' cannot be null.");
         Span span;
-        SpanContext spanContext = getOrDefault(SPAN_CONTEXT_KEY, null, SpanContext.class, context);
+        SpanContext spanContext = getOrDefault(context, SPAN_CONTEXT_KEY, null, SpanContext.class);
         if (spanContext != null) {
             span = startSpanWithRemoteParent(spanName, spanContext);
         } else {
@@ -242,10 +242,10 @@ public class OpenCensusTracer implements com.azure.core.util.tracing.Tracer {
         span.putAttribute(COMPONENT, AttributeValue.stringAttributeValue(parseComponentValue(spanName)));
         span.putAttribute(
             MESSAGE_BUS_DESTINATION,
-            AttributeValue.stringAttributeValue(getOrDefault(ENTITY_PATH_KEY, "", String.class, context)));
+            AttributeValue.stringAttributeValue(getOrDefault(context, ENTITY_PATH_KEY, "", String.class)));
         span.putAttribute(
             PEER_ENDPOINT,
-            AttributeValue.stringAttributeValue(getOrDefault(HOST_NAME_KEY, "", String.class, context)));
+            AttributeValue.stringAttributeValue(getOrDefault(context, HOST_NAME_KEY, "", String.class)));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/Tracer.java
```diff
@@ -170,20 +170,6 @@
      */
     Context setSpanName(String spanName, Context context);
```
conniey: Since we've GA'd azure-core, wouldn't removing this be a breaking API change? @JonathanGiles

JonathanGiles: Connie is right - this is breaking API.

JonathanGiles: If this API is no longer required, and will never be required, we should consider deprecating it - but we can't just go straight to removing it.

conniey:
I can see addLink being useful for tracers in general. I think it's better to remove the implementation from Event Hubs rather than removing the entire feature.

samvaity:
> I can see addLink being useful for tracers in general. I think it's better to remove the implementation from Event Hubs rather than removing the entire feature.

At the moment, `addLink` works differently in the two opensource package dependencies `opencensus` and `opentelemetry`. So keeping addLinks would then require us to maintain both these versions in our plugins. Rather than doing that we can deprecate this feature for now and later with more stable opentelemetry package can think about having this feature back.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -481,33 +479,17 @@ public String getEventHubName() {
...
-                    tracerProvider.endSpan(eventSpanContext, Signal.complete());
-                    event.addContext(SPAN_CONTEXT_KEY, eventSpanContext);
-                }
+        // Starting the span makes the sampling decision (nothing is logged at this time)
+        Context eventSpanContext = tracerProvider.startSpan(parentContext, ProcessKind.MESSAGE);
+        if (eventSpanContext != null) {
+            Optional<Object> eventDiagnosticIdOptional = eventSpanContext.getData(DIAGNOSTIC_ID_KEY);
+
+            if (eventDiagnosticIdOptional.isPresent()) {
+                event.addProperty(DIAGNOSTIC_ID_KEY, eventDiagnosticIdOptional.get().toString());
```
lmolkova: What if message already has diagnostic-id? We should not create message span or stamp context on the message

samvaity:
azp run java - eventhubs - ci

conniey:
> At the moment, addLink works differently in the two opensource package dependencies opencensus and opentelemetry. So keeping addLinks would then require us to maintain both these versions in the Tracer class when we don't even use it in our SDK's. Rather, we can deprecate this feature for now and later with a more stable opentelemetry package can think about having this feature back.

Deprecating and then undeprecating when it is convenient doesn't seem like good API design. When I see `Deprecated`, it is usually because this is "bad practise" or some "security flaw" and here is a better way to perform this operation. I don't see the problem in keeping two implementations of "addLink".

Opinions, @JonathanGiles?

JonathanGiles:
It feels too early to deprecate the addLinks Tracer API - as we don't yet have a replacement proposed. That means people who were using this API (which is almost undoubtedly no one) will have nowhere to go if they want to achieve the same functionality without using deprecated API. I would consider leaving this as-is until there is a viable replacement.

samvaity:
> It feels too early to deprecate the addLinks Tracer API - as we don't yet have a replacement proposed. That means people who were using this API (which is almost undoubtedly no one) will have nowhere to go if they want to achieve the same functionality without using deprecated API. I would consider leaving this as-is until there is a viable replacement.

@JonathanGiles, @conniey I think a good way of supporting `addLinks` would be to have it working (keeping it as is) with the original (supported) tracing package, `azure-core-tracing-opencensus` plugin and no mention in the newer(updated) `opentelemetry` plugin. Does that work?

sdk/core/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
```diff
@@ -199,8 +199,6 @@ public void startSpanOverloadNullPointerException() {
         assertThrows(NullPointerException.class, () -> openCensusTracer.start("", Context.NONE, null));
     }
```
conniey: There are no tests added?

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -350,6 +349,23 @@ private Builder getSpanBuilder(String spanName, Context context) {
...
     }
 
+    /**
+     * Extracts a {@link Span.Builder} from the given {@code context}.
+     *
+     * @param context The context containing the Span.Builder.
+     * @return The {@link Span.Builder} contained in the context, and {@code null} if it does not.
+     */
+    private Span.Builder getSharedSpanBuilder(Context context) {
+        final Object value = getOptionalObject(context, SPAN_BUILDER_KEY);
```
conniey: I think `getOptionalObject` is clunky. You have to perform the same check over for an instance every time you get from it. I'm not sure what work it is actually saving you.

```java
    @SuppressWarnings("unchecked")
    public <T> T getOrDefault(Object key, T defaultValue, Class<T> clazz) {
        final Optional<Object> optional = getData(key);
        final Object result = optional.filter(value -> value.getClass().equals(clazz)).orElse(() -> {
           logger.warning("Could not extract key '{}' of type '{}' from context.", key, clazz);
           return defaultValue;
        });

        return (T) result;
    }
```

sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClientTest.java
```diff
@@ -256,52 +255,6 @@ public void sendStartSpanSingleMessage() {
         verify(tracer1, times(3)).end(eq("success"), isNull(), any());
     }
 
-    /**
```
conniey: This test can be kept. Just ensure that no links are added.

sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubProducerClientTest.java
```diff
@@ -179,45 +177,6 @@ public void sendStartSpanSingleMessage() {
         verify(tracer1, times(2)).end(eq("success"), isNull(), any());
     }
 
-    /**
```
conniey: Same. This test can be kept. just ensure no links are added so that code path is tested.

sdk/core/azure-core-tracing-opencensus/src/test/java/com/azure/core/tracing/opencensus/OpenCensusTracerTest.java
samvaity: I think that was a leftover comment and addLinks was already tested in OpenCensus.

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java
samvaity: I think this method helps to clear out a lot of redundant methods, thanks @conniey 🎆

sdk/core/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/OpenCensusTracer.java
```diff
@@ -242,10 +242,10 @@ private void addSpanRequestAttributes(Span span, Context context, String spanNam
         span.putAttribute(COMPONENT, AttributeValue.stringAttributeValue(parseComponentValue(spanName)));
         span.putAttribute(
             MESSAGE_BUS_DESTINATION,
-            AttributeValue.stringAttributeValue(getRequestKeyAttribute(context, ENTITY_PATH_KEY)));
+            AttributeValue.stringAttributeValue(getOrDefault(ENTITY_PATH_KEY, "", String.class, context)));
         span.putAttribute(
             PEER_ENDPOINT,
-            AttributeValue.stringAttributeValue(getRequestKeyAttribute(context, HOST_NAME_KEY)));
+            AttributeValue.stringAttributeValue(getOrDefault(HOST_NAME_KEY, "", String.class, context)));
```
conniey: Thinking about the arguments... it should be (Context context, String key, T defaultValue, Class<T> clazz).

PR #6324, opened 2019-11-13
Rename BatchOptions -> CreateBatchOptions. And getRetryMode.

conniey:
* From API review, BatchOptions -> CreateBatchOptions.
* Rename: RetryOptions.getRetryMode()/setRetryMode(RetryMode mode) to get/set Mode because it is obvious that it is the mode being set.
* EventHubConnection has properties that are exposed other than being an empty class.


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/RetryOptions.java
```diff
@@ -33,7 +33,7 @@ public RetryOptions() {
      * @param retryMode The retry approach to use for calculating delays.
      * @return The updated {@link RetryOptions} object.
      */
-    public RetryOptions setRetryMode(RetryMode retryMode) {
+    public RetryOptions setMode(RetryMode retryMode) {
```
alzimmermsft: @srnagar @hemanttanwar are there configuration options in this `RetryOptions` class that could be promoted into the Azure Core RetryOptions class?

srnagar: azure-core retry options have HTTP specifics like headers that don't apply to amqp. We'll leave this as-is for GA. We can look into the feasibility of having one retry policy for both in the future.

PR #6328, opened 2019-11-13
Cpk https enforce

### LINKED ISSUE #5486: CPK should enforce HTTPS
CPK sends a secret over the wire. Currently, if your URL uses HTTP instead of HTTPS, there is no check to stop this and a user can accidentally broadcast their own secret over an insecure channel.

Clients with a CPK automatically switch to HTTPS when building the client if they happen to be on HTTP.

1. Throw when constructing a client with CPK and HTTP
2. Throw in pipeline before sending if request is in HTTP and has a CPK
3. Convert individual message in pipeline from HTTP to HTTPS if it has a CPK

**Information Checklist**
- [x] Description Added

rickle-msft:
Resolves #5486


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BuilderHelper.java
```diff
@@ -109,6 +131,18 @@ public static String getEndpoint(BlobUrlParts parts) {
...
 
+    /**
+     * Validates that the client is properly configured for using cpk.
+     *
+     * @param customerProvidedKey The cpk object.
+     * @param endpoint The endpoint for the client.
+     */
+    public static void validateCpk(CpkInfo customerProvidedKey, String endpoint) {
+        if (customerProvidedKey != null && !BlobUrlParts.parse(endpoint).getScheme().equals(Constants.HTTPS)) {
+            throw new IllegalArgumentException("Using a customer provided key requires https");
```
alzimmermsft: Should this throw an `IllegalStateException` instead?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClientBuilder.java
```diff
@@ -86,19 +83,12 @@ public BlobServiceClient buildClient() {
...
-            } else if (tokenCredential != null) {
-                return new BearerTokenAuthenticationPolicy(tokenCredential, String.format("%s/.default", endpoint));
-            } else if (sasTokenCredential != null) {
-                return new SasTokenCredentialPolicy(sasTokenCredential);
-            } else {
-                throw logger.logExceptionAsError(
-                    new IllegalArgumentException("Authorization credentials must be set."));
-            }
-        }, retryOptions, logOptions, httpClient, additionalPolicies, configuration, serviceVersion);
+        HttpPipeline pipeline = (httpPipeline != null) ? httpPipeline : BuilderHelper.buildPipeline(
```
alzimmermsft: Could you add a check before this call to ensure there is a credential set in the builder? The changes made here no longer having it throw an `IllegalArgumentException` when no credential has been set.

rickle-msft: Good catch!

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BuilderHelper.java
alzimmermsft: I'm not strong on this change.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BuilderHelper.java
```diff
@@ -109,6 +130,18 @@ public static String getEndpoint(BlobUrlParts parts) {
         }
     }
 
+    /**
+     * Validates that the client is properly configured for using cpk.
+     *
+     * @param customerProvidedKey The cpk object.
+     * @param endpoint The endpoint for the client.
+     */
+    public static void validateCpk(CpkInfo customerProvidedKey, String endpoint) {
```
gapra-msft: Just a thought, instead of failing why dont we change their url to https and log a warning or something?

rickle-msft:
@gapra-msft Containers and Blobs can have anonymous access, so we only throw on Service Client when there are no credentials. As for changing the scheme, I think this is a good idea, but I think this could cause problems with a proxy. Think Fiddler being so difficult with https. If we change it and then they start getting weird proxy failures, it'll be hard for them to figure out what went wrong. But if we fail fast and give them a clear message about what to fix, they can take immediate action.

Issue #5682, opened 2019-10-03
Java Repo doesn't follow Azure SDK team directory structure


joshfree:
Related https://github.com/Azure/azure-sdk-for-java/issues/2847

The azure/azure-sdk-for-java repo has a large number of management plane SDKs in the top level directory.  These all need to be moved under sdk/ with their supporting build pipeline, auto-pr-bot, and CODEOWNERS entries updated accordingly.

I believe Java is the last repo which is yet to be updated to match this directory structure convention under the azure/azure-sdk-for-* repos.

joshfree:
cc: @weshaggard @kurtzeborn @mitchdenny @JimSuplizio

weshaggard:
FYI this is something that @weidongxu-microsoft is working on in https://github.com/Azure/azure-sdk-for-java/pull/5378

weidongxu-microsoft:
github will respond with 502 Gateway Error if I make a huge PR with all the changes. I will have to break it up to smaller ones, starting from https://github.com/Azure/azure-sdk-for-java/pull/6332

weidongxu-microsoft:
There is "data-plane" inside "eventhubs". So this folder under project root is not cleaned.

weidongxu-microsoft:
PR on SDK and PR on swagger spec is all merged.

PR on autorest.java is yet to merge, but since now mgmt SDK is at same level under project root as before, relativePath of pom.xml no longer need to be changed, so this PR is no longer mandatory for this issue.

PR #6333, opened 2019-11-14
mgmt sdk restructure

weidongxu-microsoft:


weidongxu-microsoft:
https://github.com/Azure/azure-sdk-for-java/issues/5682

weidongxu-microsoft:
Batch done via script https://github.com/weidongxu-microsoft/misc-automation/blob/master/update_spec_java_sdk.py

mitchdenny:
I noticed that your ci.mgmt.yml files are all duplicates, this might make maintenance a bit challenging for you in the future, so you might like to consider refactoring into a template, but in terms of the substance of the changes I don't think that this will negatively impact the client SDK builds.

weshaggard:
@weidongxu-microsoft after trying to pull these changes locally I'm hitting file long path issues on windows. My root repo path is `D:\adp\azure-sdk-for-java` which is about as small a root we can get. I think we are going to have to figure out a way to shorten these file paths.

weshaggard:
@weidongxu-microsoft can you please work with @mitchdenny and @JimSuplizio to come up with a strategy to reduce the file length. As it stands after your change if you have a root path greater then 9 chars in length you will hit long path. The longest path I found was 251 chars in length without the root. @chidozieononiwu  correct me if I'm wrong but for other repo's we have been trying to keep about 50 chars available for root paths and I think we should strive for that here as well.

weidongxu-microsoft:
@weshaggard 
I think it mostly due to new folder structure. Previously it is something like
`/recoveryservices.siterecovery/resource-manager/`
and now it is
`/sdk/recoveryservices.siterecovery/azure-mgmt-recoveryservices.siterecovery/`

The service name get repeated twice.

`azure-mgmt-recoveryservices.siterecovery` part of the folder is probably not of significance since that already under the folder of this service name. So one option for those service with long name, we could just use say `mgmt` instead. E.g.
`/sdk/recoveryservices.siterecovery/mgmt/`

Another option is try to get with a shorter service name, but that might not be easy.

Issue #5298, opened 2019-09-09
[BUG] Core Playback does not parse/record body for GetPermission. 


gapra-msft:
The GetPermission Storage API https://docs.microsoft.com/en-us/rest/api/storageservices/get-permission does not return a content-type header indicating JSON so the permission does not get deserialized and stored in playback files.

sima-zhu:
The change has pushed into master. Closed the issue. Please feel free to reopen if trigger any other issues.

PR #6361, opened 2019-11-15
Fixed the getPermission playback test

sima-zhu:
The PR is to fix the case when content type has not set but there is response body available. 

https://github.com/Azure/azure-sdk-for-java/issues/5298


sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
```diff
@@ -145,7 +145,16 @@ private void captureRequestHeaders(HttpHeaders requestHeaders, Map<String, Strin
...
         String contentType = response.getHeaderValue(CONTENT_TYPE);
         if (contentType == null) {
-            return Mono.just(responseData);
+            return response.getBodyAsByteArray().switchIfEmpty(Mono.just(new byte[0])).map(bytes -> {
+                if (bytes.length == 0) {
+                    return responseData;
+                }
+
+                String content = new String(bytes, StandardCharsets.UTF_8);
+                responseData.put("Content-Length", Integer.toString(content.length()));
```
srnagar: Define a constant for `Content-Length` just like you have for `CONTENT_TYPE` and `BODY`.

sima-zhu: Fixed

PR #6364, opened 2019-11-15
Tracing Instrumentation in Batch send operations Eventhubs

### LINKED ISSUE #6319: Standardize Message and Send spans on Send operation
Currently, there seems to be an inconsistency between the languages and span relationship between the multiple overloads of the send operation.
Work items:
-	Always make Message span a sibling of Send:
-	Instrument message on creation in EventDataBatch (and friends) in all languages

samvaity:
- Standardizes Message and Send spans relation as siblings in eventhubs send operation.
- Add tracing instrumentation in batch send operation.

Fixes #6319

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventDataBatch.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventDataBatch.java
@@ -136,7 +136,7 @@ public final class EventDataBatch {
             Context eventSpanContext = tracerProvider.startSpan(eventData.getContext(), ProcessKind.MESSAGE);
             Optional<Object> eventDiagnosticIdOptional = eventSpanContext.getData(DIAGNOSTIC_ID_KEY);
             if (eventDiagnosticIdOptional.isPresent()) {
-                eventData.addProperty(DIAGNOSTIC_ID_KEY, eventDiagnosticIdOptional.get().toString());
+                eventData.getProperties().put(DIAGNOSTIC_ID_KEY, eventDiagnosticIdOptional.get().toString());
                 tracerProvider.endSpan(eventSpanContext, Signal.complete());
                 eventData.addContext(SPAN_CONTEXT_KEY, eventSpanContext);
             }
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
@@ -250,8 +250,8 @@ public class EventHubProducerAsyncClient implements Closeable {
                         ? batchMaxSize
                         : maximumLinkSize;
 
-                    return Mono.just(new EventDataBatch(batchSize, clone.getPartitionId(), clone.getPartitionKey(),
-                        link::getErrorContext, tracerProvider));
+                    return Mono.just(new EventDataBatch(batchSize, partitionId, partitionKey, link::getErrorContext,
+                        tracerProvider));
                 }));
     }
 
@@ -446,9 +446,11 @@ public class EventHubProducerAsyncClient implements Closeable {
     }
 
     private Mono<Void> sendInternal(Flux<EventData> events, SendOptions options) {
-        final SendOptions clone = options.clone();
-        if (!CoreUtils.isNullOrEmpty(clone.getPartitionKey())
-                && !CoreUtils.isNullOrEmpty(clone.getPartitionId())) {
+        final String partitionKey = options.getPartitionKey();
+        final String partitionId = options.getPartitionId();
+
+        if (!CoreUtils.isNullOrEmpty(partitionKey)
+                && !CoreUtils.isNullOrEmpty(partitionId)) {
             return monoError(logger, new IllegalArgumentException(String.format(Locale.US,
                 "SendOptions.getPartitionKey() and SendOptions.getPartitionId() are both set. Only one or the"
                     + " other can be used. partitionKey: '%s'. partitionId: '%s'",
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClientTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClientTest.java
@@ -44,6 +44,7 @@ import java.util.List;
 
 import static com.azure.core.util.tracing.Tracer.DIAGNOSTIC_ID_KEY;
 import static com.azure.core.util.tracing.Tracer.PARENT_SPAN_KEY;
+import static com.azure.core.util.tracing.Tracer.SPAN_CONTEXT_KEY;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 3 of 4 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventDataBatch.java
```diff
@@ -105,10 +117,38 @@ public boolean tryAdd(final EventData eventData) {
...
+     */
+    private EventData traceMessageSpan(EventData eventData) {
+        Optional<Object> eventContextData = eventData.getContext().getData(SPAN_CONTEXT_KEY);
+        if (eventContextData.isPresent()) {
+            // if message has context (in case of retries), don't start a message span or add a new context
+            return eventData;
+        } else {
+            // Starting the span makes the sampling decision (nothing is logged at this time)
+            Context eventSpanContext = tracerProvider.startSpan(eventData.getContext(), ProcessKind.MESSAGE);
+            if (eventSpanContext != null) {
```
conniey: Is there a scenario where tracing is enabled `eventSpanContext` is null? Seems like an incorrectly written tracer.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -423,15 +424,38 @@ public String getEventHubName() {
         }).collect(Collectors.toList());
 
         return getSendLink(batch.getPartitionId())
-            .flatMap(link -> messages.size() == 1
-                ? link.send(messages.get(0))
-                : link.send(messages));
+            .flatMap(link -> {
+
+                Context userSpanContext = parentContext.get();
```
conniey: if tracing is not enabled, this call result in a null reference exception.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -423,15 +424,38 @@ public String getEventHubName() {
...
+    /**
+     * Method to start a "Azure.EventHubs.send" span and return the created span in context object.
+     *
+     * @param userSpanContext context containing the optional parent span from the user.
+     * @param link the {@link com.azure.core.amqp.implementation.AmqpSendLink}.
+     *
+     * @return the updated {@link com.azure.core.util.Context} object containing the created send span
+     */
+    private Context traceSendSpan(Context userSpanContext, AmqpSendLink link) {
+        Context entityContext = userSpanContext.addData(ENTITY_PATH_KEY, link.getEntityPath());
```
conniey: Does this need to be a method? it's only called once

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -578,20 +545,23 @@ public void close() {
...
         private final Integer maxNumberOfBatches;
         private final ErrorContextProvider contextProvider;
+        private final TracerProvider tracerProvider;
 
         private volatile EventDataBatch currentBatch;
 
-        EventDataCollector(CreateBatchOptions options, Integer maxNumberOfBatches,
-            ErrorContextProvider contextProvider) {
+        EventDataCollector(CreateBatchOptions options, Integer maxNumberOfBatches, ErrorContextProvider contextProvider,
+            final TracerProvider tracerProvider) {
```
conniey: To be consistent, this parameter does not need to be `final`

sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClientTest.java
```diff
@@ -256,52 +256,6 @@ public void sendStartSpanSingleMessage() {
         verify(tracer1, times(3)).end(eq("success"), isNull(), any());
     }
```
conniey: In lieu of this, should add a test case for the batch.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
samvaity: Yes, it is only called once. But adding the tracing stuff in a method also helps to explain what happens here and keep it neat?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventDataBatch.java
```diff
@@ -105,10 +117,35 @@ public boolean tryAdd(final EventData eventData) {
...
+        Optional<Object> eventContextData = eventData.getContext().getData(SPAN_CONTEXT_KEY);
+        if (eventContextData.isPresent()) {
+            // if message has context (in case of retries), don't start a message span or add a new context
+            return eventData;
+        } else {
+            // Starting the span makes the sampling decision (nothing is logged at this time)
+            Context eventSpanContext = tracerProvider.startSpan(eventData.getContext(), ProcessKind.MESSAGE);
+            Optional<Object> eventDiagnosticIdOptional = eventSpanContext.getData(DIAGNOSTIC_ID_KEY);
+            if (eventDiagnosticIdOptional.isPresent()) {
+                eventData.addProperty(DIAGNOSTIC_ID_KEY, eventDiagnosticIdOptional.get().toString());
```
lmolkova: what happens if eventData already have the same property? Will it throw/override?

samvaity: I think the `properties` on eventData is a hashmap so it would just override. 
@conniey ^^

conniey: Override

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
conniey: Its easy enough to explain in a comment instead of having to create another call frame for it.

sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventDataBatchIntegrationTest.java
```diff
@@ -79,7 +79,7 @@ public void sendSmallEventsFullBatch() {
     @Test
     public void sendSmallEventsFullBatchPartitionKey() {
         // Arrange
-        final EventDataBatch batch = new EventDataBatch(ClientConstants.MAX_MESSAGE_LENGTH_BYTES, null, PARTITION_KEY, contextProvider);
+        final EventDataBatch batch = new EventDataBatch(ClientConstants.MAX_MESSAGE_LENGTH_BYTES, null, PARTITION_KEY, contextProvider, null);
```
conniey: is this going to result in a null reference exception? I think you can mock something for this.

PR #6365, opened 2019-11-15
Delete old stale files that we don't need any longer

weshaggard:
cc @mitchdenny @JimSuplizio


weshaggard:
@weidongxu-microsoft  @iscai-msft @jianghaolu can you guys confirm or deny whether or not the package.json or gulpfile.js is needed in the root of this repo? I assume they are related to codegen in some way but I'm not sure how or if they are consumed.

weidongxu-microsoft:
@weshaggard 

IFAIW, package.json is dependencies configure for npm/nodejs. npm will install gulp, which uses gulpfile.js as configure.

Then, gulpfile.js will consume api-specs.json which contains configuration for autorest.java source readme.md and other input parameters.

E.g.
```
  "advisor/resource-manager": {
    "source": "specification/advisor/resource-manager/readme.md",
    "args": "--multiapi --fluent"
  },
```

weshaggard:
@weidongxu-microsoft thanks for letting know these are in use. I will remove them from my PR.

I don't know how much work it is but it would be idea if we could move these files out of the root and somewhere under the eng directory to try and clean-up the root directory further.

weshaggard:
@weidongxu-microsoft as an aside it doesn't look the other language repo's have these files so it would be good if we can figure out how to either eliminate or move them.

weidongxu-microsoft:
@weshaggard We will have some brainstorm within our team to see whether what are the alternatives, and how other teams do for this. (one alternative would be let swagger spec raise PR automatically based on readme.md there)

weidongxu-microsoft:
@weshaggard Alternatively, could we just move them into somewhere under `eng` folder and leave them there? It will not cluster in project root, while we do not need to spend time to figure out another approach (though that should be relatively simple, some python script might just works).

What I heard from swagger team is that long term, the publish of mgmt SDK could be integrated into swagger CI. Basically when modification (by service team) on spec JSON file is merged, swagger CI will use README.java.md to generate mgmt SDK, and automatically put a PR into azure-sdk-for-java. Then service team could use the release request site to ask us to merge that PR. Ideally we could get that PR through our CI and merge it without touching the code at all.
This approach would probably also do away with the gulp as well. Though it will be long term and I do not get a date from swagger team.

weshaggard:
Yes I agree ideally this should be coming from the swagger api specs repo but if it is easy to update the paths and move these scripts under `eng\mgmt` that would be a nice easy fix to help clean-up the root repo for now.

Issue #6217, opened 2019-11-07
[Azure-core ] Requirement Gathering- Customizing azure-core Http Policy


hemanttanwar:
Various Azure client library need to have many variations in creating HttpPolicy. Current implementation is lacking and does not support these variations. 
This is an attempt to look into future and see is this scalable system to support variations in http policies. 

## What are current  known requirement to customize policies ?
  List of requirement which we know of today.

  1. Generate request ID and add it to request with different HttpHeaders.
       https://github.com/Azure/azure-sdk-for-java/pull/6214/files 
     • "x-ms-client-request-id
     • "client-request-id
  2. Retry-delay header name in RetryPolicy : https://github.com/Azure/azure-sdk-for-java/pull/6196
  3. Allow customers set the header : https://github.com/Azure/azure-sdk-for-java/issues/6171
     • "x-ms-client-request-id"
     • "x-ms-correlation-request-id"
     • "correlation-context"
 4. Storage wants to change endpoint uri during retry.
  

We need to look at Azure Core http policies and current client libraries is doing their own thing due to missing capabilities and weigh that against the likelihood of it being common.

## Possible Solutions:
List possible solutions.

### Option 1 
RetryPolicy(Function<> handlerCustomImplementation): Each policy can take in a function provided by Client to override its functionality completely. 

### Option 2 
Take Custom Options in each Http Policy to customize. 
for example:  RetryPolicy(RetryPolicyCustomOptions options)

These options can very a lot for example

             boolean values : Example  generate request id?
             Map of Key ,value : add  something in header.
             List of names : response retry-delay header  name.

itye-msft:
Based on Azure Search docs (https://docs.microsoft.com/en-us/rest/api/searchservice/common-http-request-and-response-headers-used-in-azure-search) and Swagger spec, Search uses `client-request-id` and `return-client-request-id`. 
Basically the service ignores the "x-ms-client-request-id". If a user wants to turn on tracing it's not possible unless a new header is added.
Hopefully done internally by default with the option to change that later.

hemanttanwar:
Here are few suggestions ..

Option 1 : 
   Use Options to collect variations in several Policies. This is already done in storage 
`
	public final class RequestRetryPolicy implements HttpPipelinePolicy {
		public RequestRetryPolicy(RequestRetryOptions requestRetryOptions) () { ... }
	}
 `    
[...]

Issue #6171, opened 2019-11-05
[Feature] Add support for setting RequestID, Correlation Id and Correlation Context


mssfang:
In brief: our AppConfig libraries should support letting customers set the following headers:
•	"x-ms-client-request-id"
•	"x-ms-correlation-request-id"
•	"correlation-context"

x-ms-client-request-id is set by Azure.Core, but in AppConfig, callers should be able to override the default request id.

You can see the .NET implementation of these requirements here:
https://github.com/Azure/azure-sdk-for-net/blob/7c2ff45b0f4c60fc9ed2420969b556851e74cd4d/sdk/appconfiguration/Azure.Data.AppConfiguration/src/CustomHeadersPolicy.cs

You can see how this is called in the Provider for correlation-context here:
https://github.com/Azure/Azconfig-DotnetProvider/blob/a06e3b1aa7f2b03e4a87059ea6897f4768f96598/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs#L117

hemanttanwar:
Combining Multiple issues to bring discussion at one place
Moving this discussion on to #6217

PR #6370, opened 2019-11-15
Adding Options for Request Id Policy.

hemanttanwar:
Allow customers set the value in RequestId header with following headers names
1. "x-ms-client-request-id"
2. "x-ms-correlation-request-id"
3. "client-request-id"

Should use something like RequestIDOptions POJO to collect information from user. Reference PR https://github.com/Azure/azure-sdk-for-java/pull/6196/files

Coordinate with PR, since it is waiting for this change. : #6214
Related Issue : #6171

Additional context

This change is inline with overall strategy discussed here #6217

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
@@ -23,8 +23,7 @@ public class RequestIdPolicy implements HttpPipelinePolicy {
     private final RequestIdPolicyOptions requestIdPolicyOptions;
 
     /**
-     * Creates {@link RequestIdPolicy} with default {@link ExponentialBackoff} as {@link RetryStrategy}and use
-     * 'retry-after-ms' in {@link HttpResponse} header for calculating retry delay.
+     * Creates {@link RequestIdPolicy}  with default behaviour.
      */
     public RequestIdPolicy() {
         this.requestIdPolicyOptions = null;
@@ -35,25 +34,33 @@ public class RequestIdPolicy implements HttpPipelinePolicy {
      * User can specify {@link java.util.function.Function} to dynamically generated id for various id headers.
      *
      * @param requestIdPolicyOptions with given {@link RetryPolicyOptions}.
-     * @throws NullPointerException if {@code RequestIdPolicyOptions} is {@code null}.
      */
     public RequestIdPolicy(RequestIdPolicyOptions requestIdPolicyOptions) {
-        this.requestIdPolicyOptions = Objects.requireNonNull(requestIdPolicyOptions,
-            "'retryPolicyOptions' cannot be null.");
+        this.requestIdPolicyOptions = requestIdPolicyOptions;
     }
 
 
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        String requestId = context.getHttpRequest().getHeaders().getValue(REQUEST_ID_HEADER);
         if (!Objects.isNull(requestIdPolicyOptions)) {
             HttpHeaders httpHeaders =  requestIdPolicyOptions.getIdHeaderSupplier().get();
             if (!Objects.isNull(httpHeaders)) {
-                httpHeaders.stream().forEach(httpHeader ->
-                    context.getHttpRequest().getHeaders().put(httpHeader.getName(), httpHeader.getValue())
-                );
+                httpHeaders.stream().forEach(httpHeader -> {
+                    String requestIdHeaderValue = context.getHttpRequest().getHeaders().getValue(httpHeader.getName());
+                    if (requestIdHeaderValue == null) {
+                        context.getHttpRequest().getHeaders().put(httpHeader.getName(), httpHeader.getValue());
+                    }
+            });
+            }else {
+                if (requestId == null) {
+                    context.getHttpRequest().getHeaders().put(REQUEST_ID_HEADER, UUID.randomUUID().toString());
+                }
             }
         } else {
-            context.getHttpRequest().getHeaders().put(REQUEST_ID_HEADER, UUID.randomUUID().toString());
+            if (requestId == null) {
+                context.getHttpRequest().getHeaders().put(REQUEST_ID_HEADER, UUID.randomUUID().toString());
+            }
         }
         return next.process();
     }
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicyOptions.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicyOptions.java
@@ -18,7 +18,7 @@ import java.util.function.Supplier;
 @Immutable
 public class RequestIdPolicyOptions {
 
-    private volatile Supplier<HttpHeaders> idHeaderSupplier;
+    private final Supplier<HttpHeaders> idHeaderSupplier;
 
     /**
      * Creates {@link RequestIdPolicyOptions} with provided {@code idHeaderSupplier }.
@@ -32,6 +32,7 @@ public class RequestIdPolicyOptions {
     }
 
     /**
+     * Get the provided {@link Supplier} which dynamically generate id for various request id headers.
      *
      * @return {@link Supplier} to provide dynamically generated id for various request id headers
      */


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,24 +3,56 @@
...
  * the unique identifier for the request.
+ * @see RequestIdPolicyOptions
  */
 public class RequestIdPolicy implements HttpPipelinePolicy {
     private static final String REQUEST_ID_HEADER = "x-ms-client-request-id";
 
+    private final RequestIdPolicyOptions requestIdPolicyOptions;
+
+    /**
+     * Creates {@link RequestIdPolicy} with default {@link ExponentialBackoff} as {@link RetryStrategy}and use
```
srnagar: Update javadoc as this is not related to retry policy.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,24 +3,56 @@
...
+
+    /**
+     * Creates a {@link RequestIdPolicy} with the provided {@link RequestIdPolicyOptions}.
+     * User can specify {@link java.util.function.Function} to dynamically generated id for various id headers.
+     *
+     * @param requestIdPolicyOptions with given {@link RetryPolicyOptions}.
+     * @throws NullPointerException if {@code RequestIdPolicyOptions} is {@code null}.
+     */
+    public RequestIdPolicy(RequestIdPolicyOptions requestIdPolicyOptions) {
+        this.requestIdPolicyOptions = Objects.requireNonNull(requestIdPolicyOptions,
```
srnagar: Null check is not necessary as the no-args constructor sets the `requestIdPolicyOptions` to null anyway.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicyOptions.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Immutable Configuration options for {@link RequestIdPolicy}.
+ * User can provide a {@link Supplier} to generated id for various id headers to be used in
+ * {@link com.azure.core.http.HttpRequest}. Examples of such headers are 'x-ms-client-request-id
+ * or 'x-ms-correlation-request-id' etc.
+ */
+@Immutable
+public class RequestIdPolicyOptions {
+
+    private volatile Supplier<HttpHeaders> idHeaderSupplier;
```
srnagar: Why `volatile`?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,24 +3,56 @@
...
+    }
+
+
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
-        String requestId = context.getHttpRequest().getHeaders().getValue(REQUEST_ID_HEADER);
-        if (requestId == null) {
+        if (!Objects.isNull(requestIdPolicyOptions)) {
+            HttpHeaders httpHeaders =  requestIdPolicyOptions.getIdHeaderSupplier().get();
+            if (!Objects.isNull(httpHeaders)) {
```
srnagar: A random header id should be added if `httpHeaders` is null.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicyOptions.java
```diff
@@ -0,0 +1,41 @@
...
+     * @param idHeaderSupplier {@link Supplier} to provide dynamically generated id for various request id headers.
+     * Examples of such headers are 'x-ms-client-request-id or 'x-ms-correlation-request-id' etc.
+     */
+    public RequestIdPolicyOptions(Supplier<HttpHeaders> idHeaderSupplier) {
+        this.idHeaderSupplier = Objects.requireNonNull(idHeaderSupplier,
+            "'idHeaderSupplier' cannot be null.");
+    }
+
+    /**
+     *
```
srnagar: Missing javadoc

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicyOptions.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private volatile Supplier<HttpHeaders> idHeaderSupplier;
+
+    /**
+     * Creates {@link RequestIdPolicyOptions} with provided {@code idHeaderSupplier }.
+     *
+     * @param idHeaderSupplier {@link Supplier} to provide dynamically generated id for various request id headers.
+     * Examples of such headers are 'x-ms-client-request-id or 'x-ms-correlation-request-id' etc.
+     */
+    public RequestIdPolicyOptions(Supplier<HttpHeaders> idHeaderSupplier) {
```
srnagar: name suggestion: `requestIdSupplier`

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicyOptions.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.function.Supplier;
+
+/**
+ * Immutable Configuration options for {@link RequestIdPolicy}.
+ * User can provide a {@link Supplier} to generated id for various id headers to be used in
+ * {@link com.azure.core.http.HttpRequest}. Examples of such headers are 'x-ms-client-request-id
+ * or 'x-ms-correlation-request-id' etc.
+ */
+@Immutable
+public class RequestIdPolicyOptions {
```
srnagar: Do we need a separate type for this? Can the `Supplier` instead be passed to `RequestIdPolicy` constructor directly?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,24 +3,56 @@
...
 import com.azure.core.http.HttpPipelineNextPolicy;
 import com.azure.core.http.HttpResponse;
 import reactor.core.publisher.Mono;
 
+import java.util.Objects;
 import java.util.UUID;
 
 /**
- * The pipeline policy that puts a UUID in the request header. Azure uses the request id as
+ * The pipeline policy that puts a UUID or user provided id in the request header. Azure uses the request id as
```
srnagar: With this change, we can't guarantee that the request-id is a unique identifier. User-provided request ids may not be unique.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,24 +3,56 @@
...
+     */
+    public RequestIdPolicy() {
+        this.requestIdPolicyOptions = null;
+    }
+
+    /**
+     * Creates a {@link RequestIdPolicy} with the provided {@link RequestIdPolicyOptions}.
+     * User can specify {@link java.util.function.Function} to dynamically generated id for various id headers.
+     *
+     * @param requestIdPolicyOptions with given {@link RetryPolicyOptions}.
```
srnagar: Fix javadoc - `RequestIdPolicyOptions`

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,24 +3,56 @@
...
+     * Creates {@link RequestIdPolicy} with default {@link ExponentialBackoff} as {@link RetryStrategy}and use
+     * 'retry-after-ms' in {@link HttpResponse} header for calculating retry delay.
+     */
+    public RequestIdPolicy() {
+        this.requestIdPolicyOptions = null;
+    }
+
+    /**
+     * Creates a {@link RequestIdPolicy} with the provided {@link RequestIdPolicyOptions}.
+     * User can specify {@link java.util.function.Function} to dynamically generated id for various id headers.
```
srnagar: This should be `RequestIdPolicyOptions` instead of `Function`. Also, import packages and use class names instead of fully qualified names.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,24 +3,56 @@
...
-        if (requestId == null) {
+        if (!Objects.isNull(requestIdPolicyOptions)) {
+            HttpHeaders httpHeaders =  requestIdPolicyOptions.getIdHeaderSupplier().get();
+            if (!Objects.isNull(httpHeaders)) {
+                httpHeaders.stream().forEach(httpHeader ->
+                    context.getHttpRequest().getHeaders().put(httpHeader.getName(), httpHeader.getValue())
+                );
+            }
+        } else {
             context.getHttpRequest().getHeaders().put(REQUEST_ID_HEADER, UUID.randomUUID().toString());
```
srnagar: This will now overwrite the requestId that was already in `context`. Existing requestId should be preserved. Random id should be used only when the requestId in context is null.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,24 +3,56 @@
...
+
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
-        String requestId = context.getHttpRequest().getHeaders().getValue(REQUEST_ID_HEADER);
-        if (requestId == null) {
+        if (!Objects.isNull(requestIdPolicyOptions)) {
+            HttpHeaders httpHeaders =  requestIdPolicyOptions.getIdHeaderSupplier().get();
+            if (!Objects.isNull(httpHeaders)) {
+                httpHeaders.stream().forEach(httpHeader ->
+                    context.getHttpRequest().getHeaders().put(httpHeader.getName(), httpHeader.getValue())
```
srnagar: This should check for existing headers with the same name in the context. If it already exists, it should not be overwritten.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicyOptions.java
alzimmermsft: Agreed on this comment, I don't think this policy will require additional configuration outside of being able to set Request ID headers.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
alzimmermsft: I think this is a good place to have a discussion, can take it offline if needed.

@JonathanGiles is there any guidelines around whether a pipeline policy should overwrite or only add if not exists?

hemanttanwar: I am also unsure about guideline on this. I did hear that every new request  ( including retry ) to Azure service should use new new Request ID.  But at the same time it is "optional" in Azure Docs
https://docs.microsoft.com/en-us/rest/api/azure/devops/symbol/requests/create%20requests?view=azure-devops-rest-5.1

hemanttanwar: I just spoke to Jeffery on this  and his comments are "Every retry should have the same ID. This way, we can tell how many retries occurred for a single logical operation."

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicyOptions.java
hemanttanwar: agree

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
itye-msft: While it's true, it may also be possible that a customer generates their own UUID and may want to correlate it with the ID generated by their system.
I have no idea how the service implements it, but as a user I would expect the service to save the provided ID with the rest of the raw data I sent, so I can analyze it later. Therefore the SDK cannot guarantee uniqueness if user provides the ID, which is fine, if the persistence layer can support it. 
I want to suggest that the code will generate a default ID, and allow the customer to override it based on their need.

hemanttanwar:
Closing this PR this  work was done in a separate PR by adding a constructor `public RequestIdPolicy(String requestIdHeaderName)` . So closing this PR.

PR #6381, opened 2019-11-17
Introducing Revapi into build system

JonathanGiles:
Currently it is configured to not fail the build (because it is comparing the 1.0.0-preview.N releases rather than 1.0.0). Once this versioning issue is fixed, we should flip this to fail the build.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/revapi/revapi.json
+++ b/eng/code-quality-reports/src/main/resources/revapi/revapi.json
@@ -26,18 +26,22 @@
     "extension": "revapi.ignore",
     "configuration": [
       {
-        "code": "java.missing.newClass",
-        "new": "missing-class org.reactivestreams.Publisher",
+        "regex": true,
+        "code": "java\\.missing\\.(oldClass|newClass)",
+        "new": "missing\\-class org\\.reactivestreams\\.Publisher",
         "justification": "This class is part of the reactive streams initiative in Java."
       },
       {
-        "code": "java.missing.newClass",
-        "new": "missing-class reactor.util.context.Context",
+        "regex": true,
+        "code": "java\\.missing\\.(oldClass|newClass)",
+        "new": "missing\\-class reactor\\.util\\.context\\.Context",
         "justification": "Reactor is allowed in the public API. Since we don't analyze dependencies, these show up as being unresolved."
       },
       {
-        "code": "java.missing.newClass",
-        "new": "missing-class com.fasterxml.jackson.databind.ObjectMapper",
+        "regex": true,
+        "code": "java\\.missing\\.(oldClass|newClass)",
+        "new": "missing\\-class com\\.fasterxml\\.jackson\\.databind\\.ObjectMapper",
+        "exampleUseChainInNewApi": ".*com\\.azure\\.core\\.util\\.serializer\\.JacksonAdapter.*",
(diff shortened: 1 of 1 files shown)


weshaggard:
@mitchdenny have changes been made to the triggering of the internal pipelines? I thought we were overriding those so we could always trigger them without path filters getting in the way?

weshaggard:
@JonathanGiles can you give a little information on this tool and how it works? Does it pull down the latest version of a package and compare the API's with the current version? Does it only do a diff or does it detect breaking changes?

JonathanGiles:
The way Revapi is currently configured to work is to pull down the latest Maven-released artifact, and to compare that against the locally-built artifact. It doesn't diff, it does a deeper analysis where it identifies source and binary breaking changes, as well as potentially breaking and non-breaking changes. We can configure what checks are run, what results in a build failure, and we can whitelist allowed breakages. At present the build is configured to not fail due to the versioning issue we currently have, but once we are past that we should apply the build failure config so builds will fail early when changes are introduced.

weshaggard:
>The way Revapi is currently configured to work is to pull down the latest Maven-released artifact

Is there a way to configure it to only pull GA releases or is that what it already does?

JonathanGiles:
It compares against the latest non-snapshot release. You can specify a format to work around that: https://revapi.org/modules/revapi-maven-plugin/#leaving_out_pre_releases

pom.client.xml
```diff
@@ -645,6 +675,32 @@
           </reportSet>
         </reportSets>
       </plugin>
+
+      <plugin>
+        <groupId>org.revapi</groupId>
+        <artifactId>revapi-maven-plugin</artifactId>
+        <version>0.11.2</version> <!-- {x-version-update;org.revapi:revapi-maven-plugin;external_dependency} -->
```
JimSuplizio: @conniey is the revapi-java plugin dependency not needed here?

conniey: No. I checked. It's not needed.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/AccessTier.java
```diff
@@ -55,7 +55,7 @@
     /**
      * Static value P60 for AccessTier.
      */
-    public static final AccessTier P60 = fromString("P60");
+    public static final AccessTier P65 = fromString("P60");
```
alzimmermsft: This is just a validation that revapi work correctly right?

conniey: Yup. It's With the "(REMOVE)" commit. Lol. Trying to see if it'll fail the storage build.

eng/revapi.json
```diff
@@ -0,0 +1,25 @@
...
+        "configuration": {
+            "filter": {
+                "packages": {
+                    "regex": true,
+                    "exclude" : ["com\\.azure\\..+\\.implementation(\\..+)?"]
+                }
+            }
+        }
+    },
+    {
```
hemanttanwar: We might use reclassify extension in future.   More information here https://revapi.org/modules/revapi-java/differences.html

conniey: This file can be updated in the future. I didn't add this because we don't have any breaking changes.

PR #6384, opened 2019-11-18
Rename ProxyConfiguration -> ProxyOptions

conniey:
* ProxyConfiguration is renamed to ProxyOptions
* Moving proxy options out of model class.
* Update EventHubClientBuilder.proxyConfiguration


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/ProxyOptions.java
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-package com.azure.core.amqp.models;
+package com.azure.core.amqp;
```
srnagar: Any reason why this can't be in `models` package?

conniey: It was odd that we had a package for one class (and its dependency) while RetryOptions lived in the main package. There was also feedback from Jonathan's peer as to why we had created this and put it in models.

PR #6395, opened 2019-11-18
Add Wildcard Support for Header Logging

### LINKED ISSUE #6031: Add method of header prefix in Storage logOptions
**What we have:**
We currently provide a default list of headers for logging. Customer can provide their list on the top of the default ones. 

**Future Request:**
However, we are lack of the ability to whitelist the header name with some prefix like "x-ms-*". Customers need to type in what they want to have one by one. Nice to have a method in LogOption or Storage to give the customer such options.

alzimmermsft:
Fixes #6031 

This PR adds the ability for the `HttpLogOptions` class to have patterns which are used to match headers during logging. This is done through an additional set of options to add pattern strings which will be used and concatenated to create a single `Pattern` which matches case-insensitively against the headers sent and returned during request and response.

Additional minor changes were performed the the `HttpLoggingPolicy` to clone the values of the passed `HttpLogOptions` instead of accessing them during log time to reduce duplicative work that only needs to be performed during construction of the policy.


alzimmermsft:
Closing PR as implementation needs additional discussion.

PR #6397, opened 2019-11-18
Install parent pom in client.yml

srnagar:
This change is to check if installing the parent pom in client.yml pipeline will help fix the spotbugs issue we are seeing in this build - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=190039&view=logs&j=7d609b72-0ac3-58c2-b077-8bb479e271dd&t=80cd1f54-207a-5d88-3d75-e97b6929c595

### DIFF
--- a/eng/pipelines/client.yml
+++ b/eng/pipelines/client.yml
@@ -14,6 +14,7 @@ resources:
 variables:
   DefaultOptions: '--batch-mode -Dmaven.wagon.http.pool=false --settings eng/settings.xml'
   LoggingOptions: '-Dorg.slf4j.simpleLogger.defaultLogLevel=error -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn'
+  MemoryOptions: '-Xmx3072m'
 
 jobs:
   - job: 'Build'


eng/pipelines/client.yml
```diff
@@ -88,6 +89,18 @@ jobs:
           publishJUnitResults: false
           goals: 'install'
 
+      - task: Maven@3
+        displayName: 'Install azure-sdk-parent'
+        inputs:
+          mavenPomFile: parent/pom.xml
```
weshaggard: Is this necessary because we have the same version of the parent pom in the repo as on maven but we require new changes that aren't on maven?

srnagar: It's not necessary but I have seen `Run SpotBugs` task failing because it couldn't find some dependency packages. There's no logical explanation to this as the previous `install` steps worked but the only diff in this pipeline and the PR pipeline is the `install parent pom` task. So, I added it to check if that makes a difference.

srnagar:
I am not going to merge this change as @mitchdenny is in the process of deleting `client.yml` and moving all the test coverage related steps to `archetype-sdk-client.yml`. https://github.com/Azure/azure-sdk-for-java/pull/6151

eng/pipelines/client.yml
JimSuplizio: @weshaggard I'd recommended that @srnagar try this. The spotbugs compilation failures were utterly baffling since the only thing different in the steps prior to it running was the installation of the parent pom.

PR #6400, opened 2019-11-18
Customizing azure-core Http Header in RetryPolicy  #6217

hemanttanwar:
1. Removing `RetryPolicyOptions`  used by `RetryPolicy` . It was not present in GA release as well.

2. Supporting Client to overwrite Custom 'Retry-After' Header in `RetryPolicy` .

Using Option 4  from discussion here -> https://github.com/Azure/azure-sdk-for-java/issues/6217#issuecomment-555732555
More discussion here https://github.com/Azure/azure-sdk-for-java/issues/6217

3.  Supporting ability to set multiple id header in RequestIdPolicy : 
Search team need these headers
     "x-ms-client-request-id
     "client-request-id"

Azure-App Config need these headers
     "x-ms-client-request-id"
     "x-ms-correlation-request-id"

More detail discussion : https://github.com/Azure/azure-sdk-for-java/issues/6217
Relevant PR from search team for reference:   https://github.com/Azure/azure-sdk-for-java/pull/6214/files

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
@@ -29,8 +29,8 @@ public class RetryPolicy implements HttpPipelinePolicy {
     private final ChronoUnit retryAfterTimeUnit;
 
     /**
-     * Creates {@link RetryPolicy} with default {@link ExponentialBackoff} as {@link RetryStrategy} and use
-     * 'retry-after-ms' in {@link HttpResponse} header for calculating retry delay.
+     * Creates {@link RetryPolicy} with default {@link ExponentialBackoff} as {@link RetryStrategy} and ignore the
+     * delay provided in response header.
      */
     public RetryPolicy() {
         this(new ExponentialBackoff(), null, null);
@@ -63,11 +63,12 @@ public class RetryPolicy implements HttpPipelinePolicy {
      * and ignore the delay provided in response header.
      * @param  retryAfterTimeUnit The time unit to use when applying the retry delay. {@code null} is valid if, and only
      * if, {@code retryAfterHeader} is {@code null}.
+     *
      * @throws NullPointerException When {@code retryStrategy} is {@code null}. Also when {@code retryAfterTimeUnit}
      * is {@code null} and {@code retryAfterHeader} is not {@code null}.
      */
     public RetryPolicy(RetryStrategy retryStrategy, String retryAfterHeader, ChronoUnit retryAfterTimeUnit) {
-        this.retryStrategy = Objects.requireNonNull(retryStrategy, "'retryStrategy' cannot be null");
+        this.retryStrategy = Objects.requireNonNull(retryStrategy, "'retryStrategy' cannot be null.");
         this.retryAfterHeader = retryAfterHeader;
         this.retryAfterTimeUnit = retryAfterTimeUnit;
         if (!isNullOrEmpty(retryAfterHeader)) {
@@ -76,10 +77,11 @@ public class RetryPolicy implements HttpPipelinePolicy {
     }
 
     /**
-     * Creates a {@link RetryPolicy} with the provided {@link RetryStrategy} and use 'retry-after-ms' in
-     * {@link HttpResponse} header for calculating retry delay.
+     * Creates a {@link RetryPolicy} with the provided {@link RetryStrategy} and ignore the delay provided in
+     * response header.
      *
      * @param retryStrategy The {@link RetryStrategy} used for retries.
+     *
      * @throws NullPointerException When {@code retryStrategy} is {@code null}.
      */
     public RetryPolicy(RetryStrategy retryStrategy) {


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -109,21 +91,9 @@ private Duration determineDelayDuration(HttpResponse response, int tryCount) {
         // Response will not have a retry-after-ms header.
         if (code != 429        // too many requests
             && code != 503) {  // service unavailable
```
anuchandy: thoughts on moving 429 and 503 checks to `RetryStrategy::calculateRetryDelay(HttpResponse, int)` as well? this way we ensure `RetryStrategy::calculateRetryDelay(HttpResponse, int)` is always gets called

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -180,7 +177,18 @@ public ConfigurationAsyncClient buildAsyncClient() {
         policies.add(new ConfigurationCredentialsPolicy(buildCredential));
         HttpPolicyProviders.addBeforeRetryPolicies(policies);
 
-        policies.add(retryPolicy == null ? DEFAULT_RETRY_POLICY : retryPolicy);
+        RetryPolicy defaultRetryPolicy = new RetryPolicy(new ExponentialBackoff() {
+            @Override
+            public Duration calculateRetryDelay(HttpResponse httpResponse, int retryAttempts) {
+                String delay = httpResponse.getHeaderValue(RETRY_AFTER_MS_HEADER);
```
anuchandy: Looks like by default we handle following headers in `RetryStrategy` default method:

```
retry-after-ms
 x-ms-retry-after-ms
```

so I guess we don't need this customization in AzConfig?

hemanttanwar: based on what we spoke today, we will only use one response header to avoid any complexity in core. Header name x-ms-retry-after-ms

JonathanGiles:
I'm not understanding this PR. The initial design requirement was to support custom headers. This was introduced in the previously-merged PR that added the RetryPolicyOptions type, which I have suggested isn't the right approach. This PR takes this out, which is good, but instead appears to focus on customising the retry delay (which I don't think has been a feature request yet), whilst making the requested ability to support custom headers much more complicated. Have you worked through the requirements you've outlined in #6217 and determined how they will be implemented? Why is this custom header requirement being conflated with the unrelated work on custom retry duration?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -3,59 +3,88 @@
...
 import java.time.Duration;
 
 /**
  * A pipeline policy that retries when a recoverable HTTP error occurs.
- * @see RetryPolicyOptions
  */
 public class RetryPolicy implements HttpPipelinePolicy {
 
     private final ClientLogger logger = new ClientLogger(RetryPolicy.class);
+    private static final String RETRY_AFTER_MS_HEADER = "retry-after-ms";
```
hemanttanwar: GA code have this as default retry after header.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,61 @@
...
+    /**
+     * Creates {@link RequestIdPolicy} with provided {@link Supplier} to dynamically generate request id for each
+     * {@link HttpRequest}.
+     *
+     * @param requestIdSupplier to dynamically generate to request id for each {@link HttpRequest}. {@code null} is
+     * valid value. It is suggested that this {@link Supplier} should provide unique value every time
+     * it is called. Example of these headers are 'x-ms-client-request-id', 'x-ms-correlation-request-id'.
+     */
+    public RequestIdPolicy(Supplier<HttpHeaders> requestIdSupplier) {
+        this.requestIdSupplier = requestIdSupplier;
```
hemanttanwar: We are using HttpHeaders because, There can be multiple headers which are of ID nature. We already have been asked for multiple id related headers by app config
"x-ms-client-request-id"
"x-ms-correlation-request-id"

Thus `HttpHeaders` represent multiple Headers which client can set.

sdk/core/azure-core/src/test/java/com/azure/core/http/policy/RequestIdPolicyTests.java
```diff
@@ -108,7 +108,7 @@ public void sameRequestIdForRetry() throws Exception {
                     return Mono.just(mockResponse);
                 }
             })
-            .policies(new RequestIdPolicy(), new RetryPolicy(new RetryPolicyOptions(new FixedDelay(1, Duration.of(0, ChronoUnit.SECONDS)))))
+            .policies(new RequestIdPolicy(), new RetryPolicy(new FixedDelay(1, Duration.of(0, ChronoUnit.SECONDS))))
```
mssfang: Do you have a test for customized request-id per request?

hemanttanwar: I will add one.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,61 @@
...
  */
 public class RequestIdPolicy implements HttpPipelinePolicy {
     private static final String REQUEST_ID_HEADER = "x-ms-client-request-id";
+    private final Supplier<HttpHeaders> requestIdSupplier;
+
+    /**
+     * Creates default {@link RequestIdPolicy}.
+     */
+    public RequestIdPolicy() {
+        requestIdSupplier = null;
```
alzimmermsft: Instead of setting this to `null` here could we just insert a default supplier which replicates current functionality? That way we don't need to do the initial `null` check when applying the policy.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
alzimmermsft: Do we want this to be nullable?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,61 @@
...
 import com.azure.core.http.HttpResponse;
 import reactor.core.publisher.Mono;
 
+import java.util.Objects;
 import java.util.UUID;
+import java.util.function.Supplier;
 
 /**
- * The pipeline policy that puts a UUID in the request header. Azure uses the request id as
+ * The pipeline policy that puts a UUID or user provided id in the request header. Azure uses the request id as
```
alzimmermsft: Maybe remove the comment about what will be put into the header but discuss that it inserts request ID headers.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,61 @@
...
 
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+
+        if (Objects.nonNull(requestIdSupplier)) {
+            HttpHeaders httpHeaders = requestIdSupplier.get();
+            if (Objects.nonNull(httpHeaders) && httpHeaders.getSize() > 0) {
+                httpHeaders.stream().forEach(httpHeader -> {
+                    String requestIdHeaderValue = context.getHttpRequest().getHeaders().getValue(httpHeader.getName());
+                    if (requestIdHeaderValue == null) {
```
alzimmermsft: This is a general question for all policies which mutate the request, should this just insert the header value?

@JonathanGiles @srnagar

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,61 @@
...
+        this.requestIdSupplier = requestIdSupplier;
+    }
 
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+
+        if (Objects.nonNull(requestIdSupplier)) {
+            HttpHeaders httpHeaders = requestIdSupplier.get();
+            if (Objects.nonNull(httpHeaders) && httpHeaders.getSize() > 0) {
+                httpHeaders.stream().forEach(httpHeader -> {
```
alzimmermsft: Could we use a `for` loop as `HttpHeaders` implements `Iterable<HttpHeader>`

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,61 @@
...
+    public RequestIdPolicy(Supplier<HttpHeaders> requestIdSupplier) {
+        this.requestIdSupplier = requestIdSupplier;
+    }
 
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+
+        if (Objects.nonNull(requestIdSupplier)) {
+            HttpHeaders httpHeaders = requestIdSupplier.get();
+            if (Objects.nonNull(httpHeaders) && httpHeaders.getSize() > 0) {
```
alzimmermsft: Continuing from a previous comment, if the supplier is made non-nullable and we add a default header in the base constructor (and possible default to this is null is passed into the overloaded constructor), should we do nothing if the supplier returns nothing. `x-ms-client-request-id` is optional by many services, so this would be allowed, but it would hurt tracking issues.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
alzimmermsft: While this was the default retry after header during GA the only service that I know of that uses it is AppConfiguration, so this was service specific. @JonathanGiles based on this could this be removed safely as the constructor in AppConfiguration has been updated.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
```diff
@@ -3,59 +3,88 @@
...
+    private final ChronoUnit retryAfterTimeUnit;
 
     /**
-     * Creates {@link RetryPolicy} with default {@link ExponentialBackoff} as {@link RetryStrategy}and use
+     * Creates {@link RetryPolicy} with default {@link ExponentialBackoff} as {@link RetryStrategy} and use
      * 'retry-after-ms' in {@link HttpResponse} header for calculating retry delay.
      */
     public RetryPolicy() {
-        this(new RetryPolicyOptions(new ExponentialBackoff()));
+        this(new ExponentialBackoff(), RETRY_AFTER_MS_HEADER, ChronoUnit.MILLIS);
```
alzimmermsft: Based on the comment about removing RETRY_AFTER_MS_HEADER this may need to be updated to pass null, null.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java
JonathanGiles: Yes. It looks like it is unnecessary.

hemanttanwar: okay removing this.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
hemanttanwar: based on above suggestion, I am making it not nullable.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
hemanttanwar: You answered , I think, In order to track request, we do need request id. And if  client does not provide or give us empty .. we should provide it.

hemanttanwar:
@itye-msft Please review RequestIDPolicy.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,63 @@
...
+    public RequestIdPolicy(Supplier<HttpHeaders> requestIdSupplier) {
+        this.requestIdSupplier = Objects.requireNonNull(requestIdSupplier, "'requestIdSupplier' must not be null");
+    }
 
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+
+        HttpHeaders httpHeaders = requestIdSupplier.get();
+        if (Objects.nonNull(httpHeaders) && httpHeaders.getSize() > 0) {
+            for (HttpHeader header:httpHeaders) {
```
mssfang:
```suggestion
            for (HttpHeader header : httpHeaders) {
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,63 @@
...
+     * @throws NullPointerException when {@code requestIdSupplier} is {@code null}.
+     */
+    public RequestIdPolicy(Supplier<HttpHeaders> requestIdSupplier) {
+        this.requestIdSupplier = Objects.requireNonNull(requestIdSupplier, "'requestIdSupplier' must not be null");
+    }
 
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+
+        HttpHeaders httpHeaders = requestIdSupplier.get();
```
srnagar: Wouldn't this allow users to provide a random set of HTTP headers and not just the request id headers? 

For e.g. the supplier can return 
```java
new HttpHeaders().put("foo", "bar").put(REQUEST_ID_HEADER, UUID.randomUUID().toString());
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,63 @@
...
+    public RequestIdPolicy() {
+        requestIdSupplier = () -> new HttpHeaders().put(REQUEST_ID_HEADER, UUID.randomUUID().toString());
+    }
+
+    /**
+     * Creates {@link RequestIdPolicy} with provided {@link Supplier} to dynamically generate request id for each
+     * {@link HttpRequest}.
+     *
+     * @param requestIdSupplier to dynamically generate to request id for each {@link HttpRequest}. It is suggested
+     * that this {@link Supplier} provides unique value every time it is called.
```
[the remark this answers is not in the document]
hemanttanwar: When I spoke  to Jeffrey Richter about this, I understood, request-id  is for telemetry purpose. For example Cognitive Search request-id is optional. https://docs.microsoft.com/en-us/rest/api/searchservice/common-http-request-and-response-headers-used-in-azure-search

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
hemanttanwar: yes . But unless we keep a bag of valid request id header,  we can not validate. And whenever a new request-id header is introduced, we need to update the bag.

PR #6404, opened 2019-11-19
EventHubConnection is internal

### LINKED ISSUE #6393: Making EventHubConnection internal
Making EventHubConnection internal so that it can be toggled and managed by EventHubClientBuilder.

conniey:
* Making EventHubConnection internal.
* Exposing `shareConnection()` toggle to share connection instead.

Fixes #6393


conniey:
Something is up with the mac OS builds. They're still queued. The others have passed though.

PR #6405, opened 2019-11-19
Adding memory/cpu statistics and gateway information

simplynaveen20:
This PR will add memory and cpu information in the diagnostics .
Also currently we are not capturing diagnostics in case of gateway mode , this PR will add those.
This PR will be followed immediately by  at least two PR on diagnostics which will add retry and query pipeline latency information.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
@@ -76,7 +76,6 @@ public class CosmosResponseDiagnosticsTest extends TestSuiteBase {
             assertThat(diagnostics).contains("Status Code : 404");
             assertThat(diagnostics).contains("Operation Type : " + OperationType.Read);
             assertThat(exception.getCosmosResponseDiagnostics().getRequestLatency()).isNotNull();
-            System.out.println(diagnostics);
         }
     }
 


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
```diff
@@ -210,6 +264,29 @@ void setRegionsContacted(Set<URI> regionsContacted) {
...
 
+    private void printSystemInformation(StringBuilder stringBuilder) {
+        try {
+            long totalMemory = Runtime.getRuntime().totalMemory() / 1024;
+            long freeMemory = Runtime.getRuntime().freeMemory() / 1024;
+            long maxMemory = Runtime.getRuntime().maxMemory() / 1024;
+            String used_Memory = totalMemory - freeMemory + " KB";
+            String available_Memory = (maxMemory - (totalMemory - freeMemory)) + " KB";
+
+            OperatingSystemMXBean mbean = (com.sun.management.OperatingSystemMXBean)
```
simplynaveen20: OperatingSystemMXBean from java.lang.management does not have precise information of CPU usage in percentage , therefore using com.sun.management. 
Java.lang have https://docs.oracle.com/javase/7/docs/api/java/lang/management/OperatingSystemMXBean.html#getSystemLoadAverage(), which after running on VMs unable to get the the correlation with usage properly

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
```diff
@@ -0,0 +1,149 @@
...
+        assertThat(this.gatewayClient).isNotNull();
+        createdDatabase.delete();
+        this.gatewayClient.close();
+        if (this.directClient != null) {
+            this.directClient.close();
+        }
+    }
+
+    @Test(groups = {"simple"})
+    public void testGatewayDiagnostics() {
```
moderakh: we don't use "test" as prefix in async sdk. please remove here and for other tests.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
```diff
@@ -0,0 +1,149 @@
...
+        cosmosItemProperties.setId(UUID.randomUUID().toString());
+        try {
+            CosmosItemResponse createResponse = cosmosContainer.createItem(cosmosItemProperties);
+            String diagnostics = createResponse.getCosmosResponseDiagnostics().toString();
+            assertThat(diagnostics).contains("Connection Mode : " + ConnectionMode.GATEWAY);
+            assertThat(diagnostics).contains("Gateway statistics");
+            assertThat(diagnostics).contains("Operation Type : " + OperationType.Create);
+            assertThat(createResponse.getCosmosResponseDiagnostics().getRequestLatency()).isNotNull();
+        } catch (CosmosClientException e) {
+            fail(e.getMessage());
```
moderakh: don't catch and fail, if this the test method throws, testng will automatically fail. drop try/catch

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
```diff
@@ -0,0 +1,149 @@
...
+        try {
+            CosmosItemResponse createResponse = cosmosContainer.createItem(cosmosItemProperties);
+            String diagnostics = createResponse.getCosmosResponseDiagnostics().toString();
+            assertThat(diagnostics).contains("Connection Mode : " + ConnectionMode.DIRECT);
+            assertThat(diagnostics).contains("StoreResponseStatistics");
+            assertThat(diagnostics).doesNotContain("Gateway request URI :");
+            assertThat(diagnostics).contains("AddressResolutionStatistics");
+            assertThat(createResponse.getCosmosResponseDiagnostics().getRequestLatency()).isNotNull();
+        } catch (CosmosClientException e) {
+            fail(e.getMessage());
```
moderakh: don't catch and fail, if this the test method throws, testng will automatically fail. drop try/catch , here too and anywhere else applicable

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
```diff
@@ -0,0 +1,149 @@
...
+import static org.assertj.core.api.Assertions.fail;
+
+public class CosmosResponseDiagnosticsTest extends TestSuiteBase {
+    private CosmosClient gatewayClient;
+    private CosmosDatabase createdDatabase;
+    private CosmosContainer cosmosContainer;
+    private CosmosClient directClient;
+
+    private final static String DATABASE_NAME = "TestDiagnosticDB";
+    private final static String CONTAINER_NAME = "TestDiagnosticContainer";
```
moderakh: please try to use the shared database and collection to avoid master resource throttling.

See DocumentCrudTest as example

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
moderakh: This is oracle specific implementation. There are other JVMs too: IBM, Azul, Openjdk, etc, if the JVM is not oracle based implementation is com.sun.management  available?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
```diff
@@ -210,6 +264,29 @@ void setRegionsContacted(Set<URI> regionsContacted) {
...
+            long totalMemory = Runtime.getRuntime().totalMemory() / 1024;
+            long freeMemory = Runtime.getRuntime().freeMemory() / 1024;
+            long maxMemory = Runtime.getRuntime().maxMemory() / 1024;
+            String usedMemory = totalMemory - freeMemory + " KB";
+            String availableMemory = (maxMemory - (totalMemory - freeMemory)) + " KB";
+
+            OperatingSystemMXBean mbean = (com.sun.management.OperatingSystemMXBean)
+                ManagementFactory.getOperatingSystemMXBean();
+            String processCpuLoad = Double.toString(mbean.getProcessCpuLoad());
+            String systemCpuLoad = Double.toString(mbean.getSystemCpuLoad());
```
moderakh: see this comment  from stackoverflow around OperatingSystemMXBean
> Thanks for this answer but i have observed that running this code increases CPU utilization by 30-40%. This means that we get increased CPU Utilization whenever we run this code

from:
https://stackoverflow.com/questions/19781087/using-operatingsystemmxbean-to-get-cpu-usage

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
simplynaveen20: Openjdk has  com.sun.management available , and is giving numbers. I will try with another jdk maybe IBM

simplynaveen20: Tested on AdoptSDk,IBM SDK , ZULU  they all have support for com.sun.management

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
simplynaveen20: Thanks that was helpful

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
simplynaveen20: Done

simplynaveen20:
> 2\. stigated the p

We are calculating this on toString() method and non in the our regular api.
I can spend some time on reading other libraries , do have any suggestion ?

simplynaveen20:
> Naveen,
> 
> Two questions around using `com.sun.management.OperatingSystemMXBean`:
> 
> 1. as this is not a default java lib (it is specific to oracle/sun) I am not sure if we should rely on that or not, Could you please check what other popular metrics libraries (i.e, dropwizard (https://www.dropwizard.io/en/stable/), micrometer (https://micrometer.io/)) do in regard to cpu measuring? Do they use this?
> 2. as per the discussion on the stackoverflow thread using this may have 30-40% perf impact. Have we investigated the perf impact?
> 
> > have observed that running this code increases CPU utilization by 30-40%. This means that we get increased CPU Utilization whenever we run this code
> 
> from:
> https://stackoverflow.com/questions/19781087/using-operatingsystemmxbean-to-get-cpu-usage

We are calculating this on toString() method and non in the our regular api.
I can spend some time on reading other libraries , do have any suggestion ?

simplynaveen20:
> > Naveen,
> > Two questions around using `com.sun.management.OperatingSystemMXBean`:
> > 
> > 1. as this is not a default java lib (it is specific to oracle/sun) I am not sure if we should rely on that or not, Could you please check what other popular metrics libraries (i.e, dropwizard (https://www.dropwizard.io/en/stable/), micrometer (https://micrometer.io/)) do in regard to cpu measuring? Do they use this?
> > 2. as per the discussion on the stackoverflow thread using this may have 30-40% perf impact. Have we investigated the perf impact?
> > 
> > > have observed that running this code increases CPU utilization by 30-40%. This means that we get increased CPU Utilization whenever we run this code
> > 
> > 
> > from:
> > https://stackoverflow.com/questions/19781087/using-operatingsystemmxbean-to-get-cpu-usage
> 
> We are calculating this on toString() method and not in the our regular api flow.However I tested the latency for these method, it was in fraction of ms.
> I can spend some time on reading other libraries , do have any suggestion ?

Micrometer is using sun and ibm package, which ever first will give the result, and doesn't able to find any thing dropwizard
https://github.com/micrometer-metrics/micrometer/blob/72924823327e63f4d93f6bf2b88d3fffe1523af6/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java#L50

https://github.com/micrometer-metrics/micrometer/blob/72924823327e63f4d93f6bf2b88d3fffe1523af6/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java#L97

simplynaveen20:
> > > Naveen,
> > > Two questions around using `com.sun.management.OperatingSystemMXBean`:
> > > 
> > > 1. as this is not a default java lib (it is specific to oracle/sun) I am not sure if we should rely on that or not, Could you please check what other popular metrics libraries (i.e, dropwizard (https://www.dropwizard.io/en/stable/), micrometer (https://micrometer.io/)) do in regard to cpu measuring? Do they use this?
> > > 2. as per the discussion on the stackoverflow thread using this may have 30-40% perf impact. Have we investigated the perf impact?
> > > 
> > > > have observed that running this code increases CPU utilization by 30-40%. This means that we get increased CPU Utilization whenever we run this code
> > > 
> > > 
> > > from:
> > > https://stackoverflow.com/questions/19781087/using-operatingsystemmxbean-to-get-cpu-usage
> > 
> > 
> > We are calculating this on toString() method and not in the our regular api flow.However I tested the latency for these method, it was in fraction of ms.
> > I can spend some time on reading other libraries , do have any suggestion ?
> 
> Micrometer is using sun and ibm package, which ever first will give the result
> https://github.com/micrometer-metrics/micrometer/blob/72924823327e63f4d93f6bf2b88d3fffe1523af6/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java#L50
> 
> https://github.com/micrometer-metrics/micrometer/blob/72924823327e63f4d93f6bf2b88d3fffe1523af6/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java#L97

DropwijardMetrics have sun package dependency https://github.com/dropwizard/metrics/blob/3336d1aae5d94cf94a73db52526fc437a19b9287/metrics-jvm/src/main/java/com/codahale/metrics/jvm/FileDescriptorRatioGauge.java#L34

simplynaveen20:
> Naveen Could you please run the ReadThroughput workflow of the benchmark and see if this change has any impact on the throughput?
> 
> this is to verify this claim:
> 
> `have observed that running this code increases CPU utilization by 30-40%. This means that we get increased CPU Utilization whenever we run this code`
> 
> from:
> https://stackoverflow.com/questions/19781087/using-operatingsystemmxbean-to-get-cpu-usage

Below numbers look good
#Successful Operations
             count = 1000000
         mean rate = 101.00 events/second
     1-minute rate = 100.73 events/second
     5-minute rate = 100.85 events/second
    15-minute rate = 101.13 events/second
#Unsuccessful Operations
             count = 0

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Factory;
+import org.testng.annotations.Test;
+
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+public class CosmosResponseDiagnosticsTest extends TestSuiteBase {
```
moderakh: please address what we discussed regarding tests and the fact that because the config from parent is overwriten here we may run the same test multiple times.

simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
```diff
@@ -210,6 +265,27 @@ void setRegionsContacted(Set<URI> regionsContacted) {
...
+
+            String processCpuLoad = Double.toString(this.mbean.getProcessCpuLoad());
+            String systemCpuLoad = Double.toString(this.mbean.getSystemCpuLoad());
+
+            stringBuilder.append("System State Information -------").append(System.lineSeparator());
+            stringBuilder.append("Used Memory : " + usedMemory).append(System.lineSeparator());
+            stringBuilder.append("Available Memory : " + availableMemory).append(System.lineSeparator());
+            stringBuilder.append("CPU Process Load : " + processCpuLoad).append(System.lineSeparator());
+            stringBuilder.append("CPU System Load : " + systemCpuLoad).append(System.lineSeparator());
+        } catch (Exception e) {
```
moderakh: can anything in the try/catch throw?

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
```diff
@@ -0,0 +1,147 @@
...
+    @Test(groups = {"simple"})
+    public void gatewayDiagnostics() throws CosmosClientException {
+        CosmosItemProperties cosmosItemProperties = getCosmosItemProperties();
+        CosmosItemResponse createResponse = container.createItem(cosmosItemProperties);
+        String diagnostics = createResponse.getCosmosResponseDiagnostics().toString();
+        assertThat(diagnostics).contains("Connection Mode : " + ConnectionMode.GATEWAY);
+        assertThat(diagnostics).contains("Gateway statistics");
+        assertThat(diagnostics).contains("Operation Type : " + OperationType.Create);
+        assertThat(createResponse.getCosmosResponseDiagnostics().getRequestLatency()).isNotNull();
+        deleteItem(createResponse);
```
moderakh: as long as you are using a shared collection from TestSuiteBase, you don't need to delete documents, the test framework will do that when needed.

please remove deleteItem here and elsewhere

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
```diff
@@ -89,14 +103,48 @@ void recordResponse(RxDocumentServiceRequest request, StoreResult storeResult) {
...
+            this.gatewayStatistic = new GatewayStatistic();
+            this.gatewayStatistic.operationType = rxDocumentServiceRequest.getOperationType();
+            if (storeResponse != null) {
+                this.gatewayStatistic.statusCode = storeResponse.getStatus();
+                this.gatewayStatistic.subStatusCode = DirectBridgeInternal.getSubStatusCode(storeResponse);
+
+                ISessionToken sessionToken = null;
+                String headerValue;
+                if ((headerValue = storeResponse.getHeaderValue(HttpConstants.HttpHeaders.SESSION_TOKEN)) != null) {
+                    sessionToken = SessionTokenHelper.parse(headerValue);
```
moderakh: We are encoding SessionToken to a string and then decode it back here to SessionToken object. and then at the end we encode it back to string in `ClientStatistics.toString()` 

can we avoid parsing SessionToken and just rely on its string value?

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
```diff
@@ -0,0 +1,147 @@
...
+            String diagnostics = exception.getCosmosResponseDiagnostics().toString();
+            assertThat(exception.getStatusCode()).isEqualTo(HttpConstants.StatusCodes.NOTFOUND);
+            assertThat(diagnostics).contains("Connection Mode : " + ConnectionMode.GATEWAY);
+            assertThat(diagnostics).contains("Gateway statistics");
+            assertThat(diagnostics).contains("Status Code : 404");
+            assertThat(diagnostics).contains("Operation Type : " + OperationType.Read);
+            assertThat(exception.getCosmosResponseDiagnostics().getRequestLatency()).isNotNull();
+            System.out.println(diagnostics);
+        } finally {
+            deleteItem(createResponse);
```
moderakh: as long as you are using a shared collection from TestSuiteBase, you don't need to delete documents, the test framework will do that when needed.

please remove deleteItem here and elsewhere

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
simplynaveen20: Intention is be safe and not affect the other diagnostics, please suggest if you think otherwise

moderakh: it is good to be safe. looks good. I was just curious if you have seen any specific exception getting thrown from here.

in general diagnostics code should be written in a way that never throws. you did the right thing. thank you

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
```diff
@@ -0,0 +1,134 @@
...
+            fail("request should fail as partition key is wrong");
+        } catch (CosmosClientException exception) {
+            String diagnostics = exception.getCosmosResponseDiagnostics().toString();
+            assertThat(exception.getStatusCode()).isEqualTo(HttpConstants.StatusCodes.NOTFOUND);
+            assertThat(diagnostics).contains("Connection Mode : " + ConnectionMode.GATEWAY);
+            assertThat(diagnostics).contains("Gateway statistics");
+            assertThat(diagnostics).contains("Status Code : 404");
+            assertThat(diagnostics).contains("Operation Type : " + OperationType.Read);
+            assertThat(exception.getCosmosResponseDiagnostics().getRequestLatency()).isNotNull();
+            System.out.println(diagnostics);
```
moderakh: please use logger in the test no system.out.

simplynaveen20: removed it , it was not intentional

kushagraThapar:
GatewayAddressCacheTest ran locally, passes fine. Other failed tests are queryLargeDocument known issues.

Issue #6197, opened 2019-11-06
Add Capabilities to Modify the User-Agent Header More Easily


alzimmermsft:
The current implementation of the `UserAgentPolicy` checks the `HttpRequest` for an existing `User-Agent` header. If the existing header is present it will perform simple verification that the header isn't the same as the default User-Agent header, if it is it'll just use the current header otherwise it will prepend the stored User-Agent header to the current header. This implementation makes it difficult to modify the applied header in a dynamic way without needing to build a custom `HttpPipeline` or `HttpPipelinePolicy`.

When applying the `UserAgentPolicy` it has access to the user passed `Context`, we should leverage this to allow the user to pass specific context values which will mutate the `User-Agent` header.

Proposition is adding two keys which the `UserAgentPolicy` will attemp to find, `Custom-User-Agent` and `Additional-User-Agent`. If the custom user agent is found this will overwrite any value the policy would apply and the current User-Agent header, if additional user agent is found this will be appended to the user agent the policy would apply, and if neither are found the user agent will simply be the user agent the policy would apply.

**Current Logic**
- Retrieve current User-Agent header
- If header doesn't exist or begins with the SDK User-Agent indicator
  - Set the User-Agent to the value stored in the policy
- Otherwise
  - Prepend the value stored in the policy to the current User-Agent

**Proposed Logic**
- Check the passed context for values using keys `Custom-User-Agnet` and `Additional-User-Agent`
- If `Custom-User-Agent` is found use it as the `User-Agent`
- If `Additional-User-Agent` is found append it to the value stored in the policy
- Otherwise use the value stored in the policy

JonathanGiles:
Works for me! +1

PR #6407, opened 2019-11-19
Changes for UserAgentPolicy

### LINKED ISSUE #6269: User agent header exceeds allowed limit on multiple Retries applied.
Retry policy keeps on appending to existing User Agent header causing it to exceed the header allowed limit.

Ref: 
` If the header policy is incorrectly applied after a retry policy the customer could run into an issue where the User-Agent header eventually exceeds the allowed limit for a header.`
https://github.com/Azure/azure-sdk-for-java/pull/6249#discussion_r344401938

alzimmermsft:
Fixes #6269, Implements #6197 

**Functional Breaking Change, Needs Discussion**

Currently, when the `UserAgentPolicy` is being applied it will check the passed `HttpRequest` for having an existing User-Agent header, if it does it will check the value of the header for containing the default User-Agent (`azsdk-java`) or being and just apply the value stored in the policy. If the existing header is using a value that doesn't contain the default User-Agent or isn't null the value stored in the policy will be prepended to the existing header and that will be used. This PR changes this functionality to simply use the value stored in the policy no matter what is stored in the passed request.

New reserved keys have been added to `Context`, `Override-User-Agent` and `Append-User-Agent`, which are used to store values which are used to modify the `User-Agent` header that will be applied by the `UserAgentPolicy` in an ad-hoc manner. `Override-User-Agent` will override the value stored in the policy and `Append-User-Agent` will be appended to the value stored in the policy.

The constructor in the `UserAgentPolicy` which takes an applicationId had the ServiceVersion parameter removed as it is unused and there is no plans to use it. This constructor was introduced during the 1.1.0 preview and hasn't been released. The previously released constructor which takes a ServiceVersion has been marked as deprecated as the new constructor has a super set of functionality and should be used instead.

Internally in the `UserAgentPolicy` how the User-Agent header is constructed has been changed to use a `StringBuilder` which only has a requirement of a base User-Agent header in the format of `azsdk-java-<sdk_name>/<sdk_version>` as applicationId is an optional prepend value and platform telemetry is an optional append value.


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -62,82 +71,113 @@ public UserAgentPolicy(String userAgent) {
...
-        } else {
-            if (applicationId == null) {
-                this.userAgent = String.format(USER_AGENT_FORMAT, sdkName, sdkVersion, getPlatformInfo(),
-                    version.getVersion());
-            } else {
-                this.userAgent = String.format(APPLICATION_ID_USER_AGENT_FORMAT, applicationId, sdkName, sdkVersion,
-                    getPlatformInfo(), version.getVersion());
-            }
-        }
+    public UserAgentPolicy(String applicationId, String sdkName, String sdkVersion, Configuration configuration) {
```
srnagar: So, are we not including the service version in user agent now?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
```diff
@@ -62,82 +71,113 @@ public UserAgentPolicy(String userAgent) {
...
+     * @param context request context
+     * @param next The next policy to invoke.
+     * @return A publisher that initiates the request upon subscription and emits a response on completion.
      */
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
-        String header = context.getHttpRequest().getHeaders().getValue("User-Agent");
-        if (header == null || header.contains(DEFAULT_USER_AGENT_HEADER)) {
-            header = userAgent;
+        String overrideUserAgent = (String) context.getData(OVERRIDE_USER_AGENT_CONTEXT_KEY).orElse(null);
```
srnagar: We should also consider what happens to the user agent strings that were generated with the released version of the API and how this change will impact the user data analytics. Do we have other languages also making these changes to the user agent string?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
alzimmermsft: This was never being used, while it was used in the String.format calls there weren't enough placeholders for it to be applied.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
alzimmermsft: I'll search the other languages handling on this, if need be we can revert this portion of the change.

PR #6409, opened 2019-11-19
blob Sas

gapra-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobSasImplUtil.java
```diff
@@ -0,0 +1,300 @@
...
+     * Creates a new {@link BlobSasImplUtil} with the specified parameters
+     *
+     * @param sasValues {@link BlobServiceSasSignatureValues}
+     * @param containerName The container name
+     * @param blobName The blob name
+     * @param snapshotId The snapshot id
+     */
+    public BlobSasImplUtil(BlobServiceSasSignatureValues sasValues, String containerName, String blobName,
+        String snapshotId) {
+        Objects.requireNonNull(sasValues);
```
gapra-msft: add objects require non null to container

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/BlobServiceSasSignatureValues.java
```diff
@@ -87,9 +88,12 @@
 
     private String contentType;
 
+    private BlobSasImplUtil blobSasImplUtil;
```
gapra-msft: remove

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/BlobServiceSasSignatureValues.java
```diff
@@ -358,6 +370,7 @@ public String getIdentifier() {
      * @param identifier Name of the access policy
      * @return the updated BlobServiceSASSignatureValues object
      */
+    @Deprecated
```
rickle-msft: We don't want to deprecate this because a customer may use the expiry + permissions overload but still want to set an identifier to coordinate start times or for revoke purposes.

gapra-msft: That sounds reasonable, will undo. 

Though why couldnt a customer just use the constructor for identifier and then set the expiryTime and permissions with the setters?

rickle-msft: They could, but you could use that argument in reverse, right? You could deprecate expiry and permissions, and then customers could always set those in the constructor and do everything else via setters. The point of the constructors is to hopefully lead people to understand a minimal set of required parameters, but we also need to enable them to use setters because there's actually not a single parameter that strictly *has* to be in the query string

rickle-msft:
Changelog

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
```diff
@@ -184,4 +186,39 @@ public void uploadFromFile(String filePath, ParallelTransferOptions parallelTran
...
+     * Generates a user delegation sas for the blob using the specified
+     * {@link BlobServiceSasSignatureValues}.
+     * @see BlobServiceSasSignatureValues for more information on how to construct a user delegation SAS.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.BlobClient.generateUserDelegationSas#BlobServiceSasSignatureValues-UserDelegationKey}
+     *
+     * @param blobServiceSasSignatureValues {@link BlobServiceSasSignatureValues}
+     * @param userDelegationKey {@link UserDelegationKey}
```
rickle-msft: >link [](start = 34, length = 4)

This is a generated type and doesn't have great docs on it, so we should put an actual doc string and link to the method that they can use to retrieve one

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/BlobServiceSasQueryParameters.java
```diff
@@ -240,7 +240,10 @@ UserDelegationKey userDelegationKey() {
      * Encodes all SAS query parameters into a string that can be appended to a URL.
      *
      * @return A {@code String} representing all SAS query parameters.
+     * @deprecated Please use the generate*Sas method on the desired container/blob client after initializing
```
rickle-msft: >@deprecated [](start = 7, length = 11)

I thought we couldn't deprecate this because it's used in BlobUrlParts

rickle-msft:
>     public String getContainerName() {

Can we deprecate these, too? Or do we need them across package boundaries?

---
Refers to: sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/BlobServiceSasSignatureValues.java:302 in cd2bf81. [](commit_id = cd2bf81fbc084477dde9e40abb16cea3ae7d2221, deletion_comment = False)

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -564,6 +568,43 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.BlobAsyncClient.generateSas#BlobServiceSasSignatureValues}
+     *
+     * @param blobServiceSasSignatureValues {@link BlobServiceSasSignatureValues}
+     *
+     * @return A {@code String} representing all SAS query parameters.
+     */
+    public String generateSas(BlobServiceSasSignatureValues blobServiceSasSignatureValues) {
```
rickle-msft: >generateSas [](start = 18, length = 11)

I think these should be marked as override? Is there a reason why don't just put this logic on the base client and let them inherit normally?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
gapra-msft: I'll replace this with the following 

     * @param userDelegationKey A {@link UserDelegationKey} object used to sign the SAS values.
     * @see BlobServiceClient#getUserDelegationKey(OffsetDateTime, OffsetDateTime) for more information on how to get a
     * user delegation key.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
gapra-msft: Whoops, I forgot these extend BlobBase. I'll remove from the types in the blob package

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/BlobServiceSasQueryParameters.java
gapra-msft: For some reason I thought the BlobUrlParts was just initializing this class, but youre right, its using this method. Need to undeprecate (I can also add a work item to eventually get rid of this class entirely and port everything to the Util class)

gapra-msft:
> Changelog

Added changelog for common and blob

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
gapra-msft: Changed in BlobAsyncClientBase, BlobClientBase, BlobContainerClient, BlobContainerAsyncClient

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1384,4 +1404,44 @@ private void downloadToFileCleanup(AsynchronousFileChannel channel, String fileP
...
+     * {@codesnippet com.azure.storage.blob.specialized.BlobAsyncClientBase.generateUserDelegationSas#BlobServiceSasSignatureValues-UserDelegationKey}
+     *
+     * @param blobServiceSasSignatureValues {@link BlobServiceSasSignatureValues}
+     * @param userDelegationKey {@link UserDelegationKey}
+     *
+     * @return A {@code String} representing all SAS query parameters.
+     */
+    public String generateUserDelegationSas(BlobServiceSasSignatureValues blobServiceSasSignatureValues,
+        UserDelegationKey userDelegationKey) {
+        return getBlobAsyncClient().generateUserDelegationSas(blobServiceSasSignatureValues, userDelegationKey);
```
alzimmermsft: Shouldn't the implementation of this method and `generateSas` live in this class instead of `BlobAsyncClient` given that `BlobAsyncClientBase` is the super class to `BlobAsyncClient`? This goes against common practice by having the super class using the sub-classes implementation.

Additionally moving the implementation here means `AppendBlobAsyncClient`, `BlockBlobAsyncClient`, and `PageBlobAsyncClient` get the implementation for free.

alzimmermsft: Additionally this will allow for the removal of `prepareBuilder` and `getBlobAsyncClient`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
```diff
@@ -799,4 +801,44 @@ public StorageAccountInfo getAccountInfo() {
...
+     * {@codesnippet com.azure.storage.blob.specialized.BlobClientBase.generateUserDelegationSas#BlobServiceSasSignatureValues-UserDelegationKey}
+     *
+     * @param blobServiceSasSignatureValues {@link BlobServiceSasSignatureValues}
+     * @param userDelegationKey {@link UserDelegationKey}
+     *
+     * @return A {@code String} representing all SAS query parameters.
+     */
+    public String generateUserDelegationSas(BlobServiceSasSignatureValues blobServiceSasSignatureValues,
+        UserDelegationKey userDelegationKey) {
+        return getBlobClient().generateUserDelegationSas(blobServiceSasSignatureValues, userDelegationKey);
```
alzimmermsft: Same comment as `BlobAsyncClientBase`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/BlobServiceSasQueryParameters.java
gapra-msft: Undeprecated. Created an issue here https://github.com/Azure/azure-sdk-for-java/issues/6541

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -900,6 +904,42 @@ public CpkInfo getCustomerProvidedKey() {
             });
     }
 
+    /**
+     * Generates a user delegation sas for the container using the specified
+     * {@link BlobServiceSasSignatureValues}.
+     * @see BlobServiceSasSignatureValues for more information on how to construct a user delegation SAS.
```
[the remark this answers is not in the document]
gapra-msft: Ill use this cleaner way

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -564,6 +568,43 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
+     */
+    public String generateUserDelegationSas(BlobServiceSasSignatureValues blobServiceSasSignatureValues,
+        UserDelegationKey userDelegationKey) {
+        return new BlobSasImplUtil(blobServiceSasSignatureValues, getContainerName(), getBlobName(), getSnapshotId())
+            .generateUserDelegationSas(userDelegationKey, getAccountName());
+    }
+
+    /**
+     * Generates a service sas for the blob using the specified {@link BlobServiceSasSignatureValues}
+     * @see BlobServiceSasSignatureValues for more information on how to construct a service SAS.
```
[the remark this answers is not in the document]
gapra-msft: Will use other method

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -564,6 +568,43 @@ as we can guarantee we only need at most two buffers for any call to write (two
         }
     }
 
+    /**
+     * Generates a user delegation sas for the blob using the specified
+     * {@link BlobServiceSasSignatureValues}.
+     * @see BlobServiceSasSignatureValues for more information on how to construct a user delegation SAS.
```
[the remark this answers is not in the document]
gapra-msft: Will use other method

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
gapra-msft: Yep, I just moved this in an earlier commit since Rick mentioned it. I forgot that BlobAsyncClient inherited from BlobAsyncClientBase

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
gapra-msft: Same response as above

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/BlobServiceSasSignatureValues.java
```diff
@@ -129,7 +143,12 @@ public BlobServiceSasSignatureValues() {
      * @param contentEncoding The content-encoding header for the SAS.
      * @param contentLanguage The content-language header for the SAS.
      * @param contentType The content-type header for the SAS.
+     * @deprecated Please use {@link #BlobServiceSasSignatureValues(String)},
+     * {@link #BlobServiceSasSignatureValues(OffsetDateTime, BlobSasPermission)}, or
+     * {@link #BlobServiceSasSignatureValues(OffsetDateTime, BlobContainerSasPermission)}
+     * followed by calls to the desired setters.
      */
+    @Deprecated
```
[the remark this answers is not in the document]
gapra-msft: No it would not be possible currently unless the user were to use the BlobSasImplUtil class in the implementation.util package (which is not a thing we should expect our customers to do)

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/BlobServiceSasSignatureValues.java
```diff
@@ -89,27 +82,48 @@
 
     /**
      * Creates an object with empty values for all fields.
+     * @deprecated Please use {@link #BlobServiceSasSignatureValues(String)},
+     * {@link #BlobServiceSasSignatureValues(OffsetDateTime, BlobSasPermission)}, or
+     * {@link #BlobServiceSasSignatureValues(OffsetDateTime, BlobContainerSasPermission)}
      */
+    @Deprecated
```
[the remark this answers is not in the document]
gapra-msft: The decision was made to deprecate so that users dont get confused by having multiple ways to create a SAS. If we did not deprecate them, users may feel like they need to set those fields when they dont need to.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/sas/BlobServiceSasSignatureValues.java
```diff
@@ -15,28 +15,21 @@
...
- * The following code generates SAS query parameters for an Azure storage blob.
- * <p>
- * {@codesnippet com.azure.storage.blob.specialized.BlobServiceSasSignatureValues.generateSasQueryParameters#StorageSharedKeyCredential}
- *
- * <p><strong>Generating SAS query parameters with {@link UserDelegationKey}</strong></p>
- * The following sample generates SAS query parameters for an Azure storage container.
- * <p>
- * {@codesnippet com.azure.storage.blob.specialized.BlobServiceSasSignatureValues.generateSasQueryParameters#UserDelegationKey-String}
+ * Used to initialize parameters for a Shared Access Signature (SAS) for an Azure Blob Storage service. Once all the
+ * values here are set, use the appropriate generate*Sas method on the desired container/blob client to obtain a
```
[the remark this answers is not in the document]
gapra-msft: will change

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -564,6 +564,8 @@ as we can guarantee we only need at most two buffers for any call to write (two
         }
     }
 
+
+
```
alzimmermsft: nit: random newline changes

PR #6417, opened 2019-11-19
Fixed file share live and playback tests.

sima-zhu:
The PR is to fix the file share failed live tests.


sima-zhu:
Involved some unnessary changes.

PR #6419, opened 2019-11-20
Added abililty to set/get access policy for datalake

gapra-msft:


sdk/storage/azure-storage-file-datalake/src/test/java/com/azure/storage/file/datalake/APISpec.groovy
```diff
@@ -90,7 +90,7 @@ class APISpec extends Specification {
 
     static final String garbageLeaseID = UUID.randomUUID().toString()
 
-    public static final String defaultEndpointTemplate = "https://%s.dfs.core.windows.net/"
+    public static final String defaultEndpointTemplate = "http://%s.dfs.core.windows.net/"
```
rickle-msft: Revert

gapra-msft: done

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/Transforms.java
```diff
@@ -297,4 +305,73 @@ private static FileReadHeaders toPathReadHeaders(BlobDownloadHeaders h) {
...
+        }
+        List<BlobSignedIdentifier> blobIdentifiers = new ArrayList<>();
+        for (DataLakeSignedIdentifier identifier : identifiers) {
+            blobIdentifiers.add(Transforms.toBlobIdentifier(identifier));
+        }
+        return blobIdentifiers;
+    }
+
+    private static BlobSignedIdentifier toBlobIdentifier(DataLakeSignedIdentifier identifier) {
+        if (identifier == null) {
```
alzimmermsft: If this is null do we want to add a null value into the converted list?

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/Transforms.java
```diff
@@ -297,4 +305,73 @@ private static FileReadHeaders toPathReadHeaders(BlobDownloadHeaders h) {
...
+            return null;
+        }
+        List<BlobSignedIdentifier> blobIdentifiers = new ArrayList<>();
+        for (DataLakeSignedIdentifier identifier : identifiers) {
+            blobIdentifiers.add(Transforms.toBlobIdentifier(identifier));
+        }
+        return blobIdentifiers;
+    }
+
+    private static BlobSignedIdentifier toBlobIdentifier(DataLakeSignedIdentifier identifier) {
```
alzimmermsft: Given that this is only called internally would we want to just inline this call into the list conversion?

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/Transforms.java
```diff
@@ -297,4 +305,73 @@ private static FileReadHeaders toPathReadHeaders(BlobDownloadHeaders h) {
...
+            return null;
+        }
+        List<DataLakeSignedIdentifier> dataLakeIdentifiers = new ArrayList<>();
+        for (BlobSignedIdentifier identifier : identifiers) {
+            dataLakeIdentifiers.add(Transforms.toDataLakeIdentifier(identifier));
+        }
+        return dataLakeIdentifiers;
+    }
+
+    private static DataLakeSignedIdentifier toDataLakeIdentifier(BlobSignedIdentifier identifier) {
```
alzimmermsft: Same comments as the other Identifier conversion

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/Transforms.java
gapra-msft: Since the datalake API just calls into the blob API, we're basically just spitting out exactly what was given to us to pipe it to the corresponding type, so I think its fine to have  that null check

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/Transforms.java
gapra-msft: I think we can keep this as a separate method since it could end up being needed elsewhere in the future, you never know.

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/Transforms.java
gapra-msft: Same as above

PR #6422, opened 2019-11-20
move gulp inside eng/mgmt

weidongxu-microsoft:


weidongxu-microsoft:
https://github.com/Azure/azure-sdk-for-java/pull/6365

eng/mgmt/gulpfile.js
```diff
@@ -55,6 +55,7 @@ gulp.task('default', function() {
 
 const maxParallelism = parseInt(args['parallel'], 10) || os.cpus().length;
 var specRoot = args['spec-root'] || defaultSpecRoot;
+var sdkRoot = path.join(process.cwd(), '../..')
```
yaohaizh: use const value for root folder

eng/mgmt/gulpfile.js
```diff
@@ -151,6 +152,17 @@ var codegen = function(project, cb) {
     return execa(cmd, [], { shell: true, stdio: "inherit" });
 };
 
+var deleteMgmtFolders = function(project) {
+    project = project.split('/')[0]
+    var projectRoot = path.join(sdkRoot, 'sdk', project);
+    fs.readdirSync(projectRoot).forEach(function(file, index) {
+        projectDir = path.join(projectRoot, file)
+        if(fs.lstatSync(projectDir).isDirectory() && file.startsWith('mgmt-')) {
```
JimSuplizio: @weidongxu-microsoft, I'm not a huge fan of startsWith on the directory name to determine management folders. If the management project pom files are all parented correctly they should have a parent artifact of azure-arm-parent. Couldn't we use that to more reliably determine if they're management?

weshaggard:
@weidongxu-microsoft thanks for cleaning this up. Is there any way to also include https://github.com/Azure/azure-sdk-for-java/blob/master/swagger_to_sdk_config.json as part of this move or is it part of another process?

weidongxu-microsoft:
@wesback swagger_to_sdk_config.json is used by swagger team for automation. I will involve Phoenix @NullMDR in the discussion. If we move it their script need modification (and need to be done in sync).

Hello Phoenix, is it OK that we move swagger_to_sdk_config.json to some inner folder?

eng/mgmt/gulpfile.js
weidongxu-microsoft: Good idea. We should have a "pom.mgmt.xml" in sdk/service root, and I could use that to find all mgmt folders. Let me try it.

I probably will not try with parent project pom. This is JS code (and we actually not want to modify it too much) and I probably do not want to involve too much dependency on XML parsing and POM handling.

mitchdenny: The POM restructuring that I am doing might end up addressing this ;)

NullMDR:
@weshaggard @weidongxu-microsoft Currently SDK Automation only support swagger_to_sdk_config.json in root folder. If you have a strong need on this feature we could discuss on it.

weshaggard:
>@weshaggard @weidongxu-microsoft Currently SDK Automation only support swagger_to_sdk_config.json in root folder. If you have a strong need on this feature we could discuss on it.

Thanks for the info. No strong reason just trying to clean-up the root of the repo. It would be nice if you guys could have an issue on your backlog to allow repo's to set this as a configuration option so folks aren't stuck with it being only in the root of the repo.

PR #6424, opened 2019-11-20
Parent POM and aggregator POM restructuring.

mitchdenny:
Experimental restructuring of parent POMs now that management plane has moved. The intention is for all cross service parent POMs to sit within the ```sdk/parents``` folder in their own sub-folder. For cases where we still have a parent POM on a per service basis, it will sit in a sub-folder in that services directory (e.g. ```sdk/cosmos/azure-cosmos-parent```) ... although this should be an exception and ideally we can eliminate them.

All other POMs will become pure aggregator POMs (including the root ```/pom.xml```). Also considering renaming the ```pom.service.xml``` files to be just ```pom.xml``` so they work with Maven defaults.

As part of this change we would probably move to a model where we build projects using the project list feature ```-pl``` and the ```-am``` switch. But that needs more exploration.

### DIFF
--- a/sdk/core/pom.xml
+++ b/sdk/core/pom.xml
@@ -14,7 +14,6 @@
     <module>azure-core-http-netty</module>
     <module>azure-core-http-okhttp</module>
     <module>azure-core-test</module>
-    <module>azure-core-tracing-opencensus</module>
     <module>azure-core-tracing-opentelemetry</module>
   </modules>
 </project>


JimSuplizio:
/CC: @JonathanGiles 
@mitchdenny there are a couple of things that you're going to need to update that have references to the parent/pom.xml, pom.client.xml or pom.service.xml files for building.
- https://github.com/Azure/azure-sdk-for-java/blob/master/CONTRIBUTING.md 
- https://github.com/Azure/azure-sdk-for-java/wiki/Building

pom.xml
```diff
@@ -19,8 +19,8 @@
   <parent>
     <groupId>com.azure</groupId>
     <artifactId>azure-sdk-parent</artifactId>
-    <version>1.6.0</version> <version>1.6.0</version> <!-- {x-version-update;com.azure:azure-sdk-parent;current} -->
-    <relativePath>./parent/pom.xml</relativePath>
+    <version>1.6.0</version> <!-- {x-version-update;com.azure:azure-sdk-parent;current} -->
+    <relativePath>sdk/parents/azure-sdk-parent</relativePath>
```
JimSuplizio: This is a general question about the relative paths across the pom files for their parent references. Is there any reason why the previous relative path tags included the xml file and the new relative path tags do not? I realize with your restructuring there's only one pom.xml in each of those parent directories but I'm just curious.

mitchdenny: Maven will by default append ```pom.xml``` when processing the relative path, so its redundant. It was redundant before, but with the restructuring now to have client, management and data poms in separate directories it is more redundant since there is no opportunity for confusion.

mitchdenny:
> Other than a minor question and the request for the contributing/how to build documents needing to be updated I'm OK with the changes. I do want to see the docs updated before this is submitted since this will changes the build commands.

Yep ... I'll definitely update the docs before pushing. I'm going to wait until later in the day my time so all you fine folks aren't using the build pools, otherwise I'll saturate it with updates to this PR ;) I still have an outstanding question on the Java channel about the use of the ```./profiles``` directory. I want to get that resolved too.

mitchdenny:
.... oh and the search build is meant to be failing at this point in case anyone is wondering :)

mitchdenny:
@weidongxu-microsoft this PR is a fairly dramatic change to the locations of the parent POMs within the Java repository. My intention here is to drive increased consistency with how we deal with some of the loose POM files/modules that we have sitting in the ```root```, ```./parents``` and ```./profiles``` directory.

At the moment this change works for the client and data plane libraries, but management plane is a false positive because I'm not really building it because of an issue in the ```./profiles``` directory. I'd like to resolve this.

My proposal is that I lift what is in the ```./profiles``` directory into the ```./sdk/profiles``` directory and deal with any issues that fall out of that. Would that cause any issues for you?

In the root of the repo, the only folders we would have after I am done are:

- ```.azure-pipelines``` (to be moved/deleted)
- ```.github```
- ```eng```
- ```samples```
- ```sdk```

There would only be one POM in the root (```pom.xml```) which would reference the management, client and data POMs in ```sdk/parents/*``` and anyone could simply go to the root and build any module using something like the following:

```$ mvn -pl com.azure:azure-core -am clean install```

The CI build would build the following way:

```$ mvn -f sdk/core/pom.xml package```

This is to cover some of our plugin scenarios (like azure-core-http-*). But overall I think that we are heading to a cleaner more consistent usage of Maven.

**Comment by weidongxu-microsoft** [comment]
@mitchdenny I searched the POMs under sdk and I didn't find any referring to there hybrid folders. So I think it is safe to move.

Just a thought, is it more proper to move ./profiles folder into ./eng folder (or somewhere else), instead of ./sdk folder? Reason is thatI assume we'd prefer ./sdk mostly contains core + identity + all services. Putting a profiles there might be a bit off.

**Comment by mitchdenny** [comment]
> Just a thought, is it more proper to move ./profiles folder into ./eng folder (or somewhere else), instead of ./sdk folder? Reason is that I assume we'd prefer ./sdk mostly contains core + identity + all services. Putting a `profiles` there might be a bit off.

I'll loop you into a chat in Teams that has some more context. There are actual shipping libraries in the /profiles folder. It is a module that groups together a set of APIs which corresponds to a particular service profile. So it probably belongs in sdk even though its not a pure service, its a bit of a special case like our parent POMs.

Now that I know more I'll do some experimentation and see how it comes together.

File: pom.xml [hunk 2]
```diff
@@ -7,398 +7,24 @@
...
-        <module>sdk/storage</module>
-        <module>sdk/template</module>
-        <!--
-        <module>sdk/search</module>
-        -->
-      </modules>
-    </profile>
-
-  </profiles>
+  <modules>
```
JimSuplizio: @mitchdenny isn't this going to cause us to have to publish this pom file?

mitchdenny: I've removed this file, but no anyway since its just an aggregator.

weidongxu-microsoft: Just a note, these profiles pom seems corresponding to this in spec https://github.com/Azure/azure-rest-api-specs/tree/master/profile

mitchdenny: I'm not familiar with how these spec files are processed does me moving the profiles directories cause a problem with code generation?

sdk/profiles/profile-parent-v2018-03-01-hybrid/pom.xml
```diff
@@ -10,7 +10,11 @@
     <groupId>com.microsoft.azure</groupId>
     <artifactId>azure-arm-parent</artifactId>
     <version>1.2.0</version>
+<<<<<<< HEAD
     <relativePath>../../pom.management.xml</relativePath>
+=======
+    <relativePath>../../parents/azure-arm-parent</relativePath>
+>>>>>>> Repoint profile parents.
```
weidongxu-microsoft: Need resolve conflict.

pom.xml
weidongxu-microsoft: Personally I think moving pom around would not affect published artifacts, so there should be no need to re-publish anything.

However there could be problem on `./pom.management.xml`, on swagger automation.

Current released autorest.java would generate service code with pom.xml refering to `<relativePath>../../../pom.management.xml</relativePath>`. And swagger automation uses that autorest.java.
If pom.management.xml moves without sync update/release on autorest.java, it could break swagger automation (though I think that automation on Java SDK is not a mandatory one for merge).

I am not 100% sure about this. Maybe you can just be prepared to at least temporary put back `./pom.management.xml` if we see swagger automation issue after this get merged.

sdk/profiles/profile-parent-v2018-03-01-hybrid/pom.xml
mitchdenny: Fixed

pom.xml
mitchdenny: Perhaps we can just tweak the automation and merge the changes simultaneously. Where is the file that controls this relative path? (I'm not really across our code-gen at the moment).

mitchdenny:
@conniey and @alzimmermsft interested in your input on something that has come up with this parent POM restructuring.

As part of my changes I've removed the ```<module />``` references from the ```pom.client.xml``` parent POM (now renamed/moved). But that impacts some of the work that we do in the Analyze job in the build stage.

A few of the Maven steps there use the parent POM to do things like checkstyles, revapi etc.

I believe that the idiomatic way to do this with Maven would have it such that each module has its own checkstyle/revapi/etc configuration, so that if you went into say ```sdk/core/azure-core``` and executed ```mvn checkstyle:blah``` it would run the appropriate goals with the appropriate configuration scoped to that build.

We've taken a different approach due to the who mono-repo thing which works today, but not with the POM file restructuring I'm doing to get back to a more idiomatic workflow.

Trying to plot a path forward here, any ideas? Ideally we wouldn't need to drop plugin configuration into every single module.

mitchdenny:
So you know, part of my thinking here is that some of these things we do in Analyze could probably just disappear if we do them automatically (scoped to the modules we are building) in the normal Maven lifecycle when we build the libraries.

pom.xml
weidongxu-microsoft: It is here https://github.com/Azure/autorest.java/blob/master/src/azurefluent/Model/CodeModelJvaf.cs#L145

Synced change might be hard. autorest.java need to get published to npm (not just code merged). Jianghao should know the details.

JonathanGiles:
Mitch - if this happens

> @conniey and @alzimmermsft interested in your input on something that has come up with this parent POM restructuring.
> 
> As part of my changes I've removed the `<module />` references from the `pom.client.xml` parent POM (now renamed/moved). But that impacts some of the work that we do in the Analyze job in the build stage.
> 
> A few of the Maven steps there use the parent POM to do things like checkstyles, revapi etc.

@mitchdenny Please be sure that wherever you end up it is still possible to run an aggregate command over the whole track two client library set for checkstyle, spotbugs, revapi, javadoc, jacoco, etc. Thanks.

mitchdenny:
> @mitchdenny Please be sure that wherever you end up it is still possible to run an aggregate command over the whole track two client library set for checkstyle, spotbugs, revapi, javadoc, jacoco, etc. Thanks.

I think that we'll end up needing to setup a seperate process for these aggregate reports. It doesn't make sense to have to build the universe to produce these reports in Analyze every single time we submit a PR. Would a separate uber analysis pipeline suffice here?

PR #6425, opened 2019-11-20
Update opentelemetry plugin package to use released package version

### LINKED ISSUE #6301: azure-core-tracing-opentelemetry should not be taking dependencies on SNAPSHOT versions
We should not be taking dependencies on SNAPSHOT versions, there are two in here io.opentelemetry:opentelemetry-api and io.opentelemetry:opentelemetry-sdk. We could not release with these dependencies. Further, I believe that unless azure-core-tracing-opentelemetry's version was also -SNAPSHOT, that Maven would reject the package.

We should not be adding tags to the individual component files. This overrides the repository settings defined further up in the hierarchy and bypasses our cache.

There's also one other big issue with the SNAPSHOT version dependencies is that Maven is going to try and query to pull these every time which will cause transient errors. I'm actually running into a transient error on my local machine because it this which built earlier and is failing now.

The PR That introduced this is https://github.com/Azure/azure-sdk-for-java/pull/6131

samvaity:
Fixes #6301

### DIFF
--- a/sdk/core/azure-core-tracing-opentelemetry/pom.xml
+++ b/sdk/core/azure-core-tracing-opentelemetry/pom.xml
@@ -35,6 +35,7 @@
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-api</artifactId>
+      <version>0.2.0</version>
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core-tracing-opentelemetry/pom.xml
```diff
@@ -54,7 +46,7 @@
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-sdk</artifactId>
-      <version>0.2.0-SNAPSHOT</version>
+      <version>0.2.0</version>
```
alzimmermsft: Does this need to list the dependency version as it should be determined transitively from the parent POM.

samvaity: I tried it without providing one but doesn't pick it up from the parent pom. Would it be because it is a test dependency?

PR #6427, opened 2019-11-20
Change event processor handlers from function to consumer

srnagar:
This change is to update the event handlers in EventProcessorBuilder to use `Consumer` instead of `Function`.  This is based on the API review feedback. Other feedback from API review will be addressed in follow-up PRs.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/README.md
+++ b/sdk/eventhubs/azure-messaging-eventhubs/README.md
@@ -310,7 +310,7 @@ class Program {
                     + "sequence number of event = " + partitionEvent.getEventData().getSequenceNumber());
             })
             .processError(errorContext -> {
-                System.out.println("Error occured while processing events " + errorContext.getThrowable().getMessage);
+                System.out.println("Error occurred while processing events " + errorContext.getThrowable().getMessage());
             })
             .buildEventProcessor();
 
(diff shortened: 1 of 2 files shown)


sdk/eventhubs/azure-messaging-eventhubs/README.md
```diff
@@ -308,7 +308,9 @@ class Program {
             .processEvent(partitionEvent -> {
                 System.out.println("Partition id = " + partitionEvent.getPartitionContext().getPartitionId() + " and "
                     + "sequence number of event = " + partitionEvent.getEventData().getSequenceNumber());
-                return Mono.empty();
+            })
+            .processError(errorContext -> {
+                System.out.println("Error occured while processing events " + errorContext.getThrowable().getMessage);
```
conniey: Does getMessage work? Surprised it's not a function.

sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/EventProcessorSample.java
```diff
@@ -4,12 +4,11 @@
 package com.azure.messaging.eventhubs;
```
conniey: One of the feedbacks was to have a reference sample of how we expect users implement handle errors (ie. if it is transient, or an AmqpException or etc). Is this in another PR?

sdk/eventhubs/azure-messaging-eventhubs/README.md
srnagar: Oh it's a typo. Fixed it.

sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/EventProcessorSample.java
srnagar: I will add that in another PR.

PR #6428, opened 2019-11-20
Cosmos: ChangeFeedProcessor fix for load-balancing leases

milismsft:
It addresses a bug which terminates the processing earlier than it should when an unexpected exception occurs in the thread which is load-balancing the lease documents.
A "catch all" exception handler has been added that will keep the loop which checks and acquires available leases alive until the change feed processor instance is stopped by calling the respective "stop()" method.


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionLoadBalancerImpl.java
```diff
@@ -131,6 +131,11 @@ public boolean isRunning() {
                         })
                     );
             })
+            .onErrorResume(throwable -> {
+                // "catch all" exception handler to keep the loop going until the user stops the change feed processor
+                logger.warn("Unexpected exception thown while trying to acquire available leases", throwable);
```
mkolt: minor spelling: thown -> thrown

PR #6429, opened 2019-11-20
Updating credentials and EventData from review

conniey:
* Updates to EventSharedKeyCredential -> EventSharedAccessKeyCredential
  * Removes tokenValidity duration from constructor (this is package-private now)
  * No longer throws checked exceptions.
* Removes EventData implements Comparable
  * Makes getContext() and addContext() package-private
  * Removes addProperty
  * Updates getBody() to return byte[]

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventDataBatch.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventDataBatch.java
@@ -221,9 +221,7 @@ public final class EventDataBatch {
             message.setMessageAnnotations(messageAnnotations);
         }
 
-        if (event.getBody() != null) {
-            message.setBody(new Data(new Binary(event.getBody())));
-        }
+        message.setBody(new Data(new Binary(event.getBody())));
 
         return message;
     }
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
@@ -112,9 +112,7 @@ class EventHubMessageSerializer implements MessageSerializer {
 
         setSystemProperties(eventData, message);
 
-        if (eventData.getBody() != null) {
-            message.setBody(new Data(new Binary(eventData.getBody())));
-        }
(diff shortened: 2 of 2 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
```diff
@@ -93,24 +86,10 @@ public EventData(byte[] body, Context context) {
...
-     * @param body The data to set for this event.
-     * @param context A specified key-value pair of type {@link Context}.
-     * @throws NullPointerException if {@code body} or if {@code context} is {@code null}.
-     */
-    public EventData(ByteBuffer body, Context context) {
-        Objects.requireNonNull(body, "'body' cannot be null.");
-        Objects.requireNonNull(body, "'context' cannot be null.");
-
-        this.body = body;
+        this.body = Objects.requireNonNull(body, "'body' cannot be null.").array();
```
alzimmermsft: Should this call into the `byte[]` constructor instead?

```java
this(Objects.requireNonNull(body, "'body' cannot be null.").array());
```

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubSharedKeyCredential.java
```diff
@@ -82,11 +98,21 @@ public EventHubSharedAccessKeyCredential(String policyName, String sharedAccessK
             throw new IllegalArgumentException("'tokenTimeToLive' has to positive and in the order-of seconds");
         }
 
-        hmac = Mac.getInstance(HASH_ALGORITHM);
+        try {
+            hmac = Mac.getInstance(HASH_ALGORITHM);
+        } catch (NoSuchAlgorithmException e) {
+            throw logger.logExceptionAsError(new AzureException(
```
alzimmermsft: Is `AzureException` the appropriate exception to be thrown here?

conniey: Any suggestions? Not sure what would apply.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
conniey: good catch. Fixed

PR #6434, opened 2019-11-20
Centralize versioning for external dependencies

JimSuplizio:
This is part 2 of centralizing version management which primarily deals with external dependencies and pom file cleanup.

Before anything else it's worth answering the question of "why" and here are the reasons:
- Right now we're in a state where we have to publish parent/pom.xml mainly due to the version strings in the properties and the dependencyManagement sections. 
- Changes are being made to external dependencies in the parent pom and incrementing its version with changes only trickling down through the client track with data/management tracks being left behind at earlier versions. Since some of these version updates are security related, they should be being made everywhere.
- There's no continuity among the pom files in the repo. Some pom files use the dependency fallback, some use the version strings defined in the parent and others just declare their own versions which are often the same. This includes plugin and external dependency versions.

eng\versioning\external_dependencies.txt - Similar to the version_*.txt files, this file contains the list of external dependencies. An external dependency is anything that we do not build/own OR any dependency that is not part of a given pipeline (ex. if a client library has a dependency on a data or management artifact, that is considered an external dependency.) Similar to previous version tags, every external dependency version will have a tag similar to the following:

The update_versions.py script has been slightly modified to deal with external dependency versions. The biggest change in here was to deal with the external dependency versions which don't follow semver and had to be handled slightly differently.

Every dependency and every plugin now has groupId, artifactId and versionId elements and <version> elements lines will need to be tagged as mentioned above. This is already the case with library versions. _**Within the next week or so automation will be in place to enforce this. Further, with these changes dependencyManagement sections in poms are going to be disallowed.**_

Because of the nature of changes to parent/pom, its version needed to be updated and this had to trickle down through the tree. 

It's worth noting that management libraries have not had these changes yet, only client and data tracks. Management is aware of these changes those are going to be incoming for management within the next few weeks.

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/pom.xml
+++ b/sdk/appconfiguration/azure-data-appconfiguration/pom.xml
@@ -95,7 +95,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-identity</artifactId>
-      <version>1.0.1</version>
+      <version>1.0.1</version> <!-- {x-version-update;com.azure:azure-identity;dependency} -->
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/sdk/storage/azure-storage-blob-cryptography/pom.xml
+++ b/sdk/storage/azure-storage-blob-cryptography/pom.xml
@@ -70,12 +70,24 @@
       <artifactId>azure-storage</artifactId>
       <version>8.4.0</version> <!-- {x-version-update;storage_com.microsoft.azure:azure-storage;external_dependency} -->
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-core</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.microsoft.azure</groupId>
       <artifactId>azure-keyvault-cryptography</artifactId>
       <version>1.2.2</version> <!-- {x-version-update;com.microsoft.azure:azure-keyvault-cryptography;external_dependency} -->
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-core</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>


eng/versioning/external_dependencies.txt
```diff
@@ -0,0 +1,221 @@
...
+io.micrometer:micrometer-registry-graphite;1.2.0
+io.netty:netty-buffer;4.1.42.Final
+io.netty:netty-codec-http;4.1.42.Final
+io.netty:netty-handler;4.1.42.Final
+io.netty:netty-handler-proxy;4.1.42.Final
+io.netty:netty-tcnative;2.0.26.Final
+io.netty:netty-tcnative-boringssl-static;2.0.26.Final
+io.opencensus:opencensus-api;0.24.0
+io.opencensus:opencensus-impl;0.24.0
+# SNAPSHOTS should not be allowed
```
samvaity: Should it be good to remove this comment now?

eng/versioning/external_dependencies.txt
```diff
@@ -0,0 +1,221 @@
...
+io.micrometer:micrometer-registry-azure-monitor;1.2.0
+io.micrometer:micrometer-registry-graphite;1.2.0
+io.netty:netty-buffer;4.1.42.Final
+io.netty:netty-codec-http;4.1.42.Final
+io.netty:netty-handler;4.1.42.Final
+io.netty:netty-handler-proxy;4.1.42.Final
+io.netty:netty-tcnative;2.0.26.Final
+io.netty:netty-tcnative-boringssl-static;2.0.26.Final
+io.opencensus:opencensus-api;0.24.0
+io.opencensus:opencensus-impl;0.24.0
```
samvaity: Not sure if we need the `opencensus-impl` and `opentelemetry-sdk` in here, those are test dependencies.

eng/versioning/external_dependencies.txt
JimSuplizio: Doh! Yes @samvaity, you are correct. I've removed the comment locally and it'll be updated when I push the next commit.

eng/versioning/external_dependencies.txt
JimSuplizio: I'll move them down under test with the next commit.

eng/versioning/external_dependencies.txt
```diff
@@ -0,0 +1,200 @@
...
+org.codehaus.mojo:properties-maven-plugin;1.0.0
+org.codehaus.mojo:xml-maven-plugin;1.0
+org.eclipse.jetty:jetty-maven-plugin;9.3.22.v20171030
+org.eclipse.m2e:lifecycle-mapping;1.0.0
+org.jacoco:jacoco-maven-plugin;0.8.4
+org.jacoco:org.jacoco.agent;0.8.4
+org.mockito:mockito-core;3.0.0
+org.nanohttpd:nanohttpd;2.3.1
+
+# External Dependency Exceptions
```
JimSuplizio: It's worth noting that these dependency exceptions, for those versions that are lower, are going to have written up issues for the specific area to look at updating to the common version. For those versions that are higher issues will be written up for us to investigate moving the common version to the higher version.

JimSuplizio:
It's worth noting that the intermittent Cosmos Emulator test failures are known and there's a [PR ](https://github.com/Azure/azure-sdk-for-java/pull/6519) in progress to fix them.

parent/pom.xml
```diff
@@ -16,13 +16,14 @@
     <url>http://microsoft.com</url>
   </organization>
 
+<!--
   <parent>
```
weshaggard: If we decide to stick without this we should just remove it instead of commenting it out.

eng/versioning/external_dependencies.txt
```diff
@@ -0,0 +1,200 @@
...
+org.slf4j:slf4j-simple;1.7.25
+
+## Test dependency versions
+cglib:cglib-nodep;3.2.7
+com.github.tomakehurst:wiremock-standalone;2.24.1
+com.microsoft.azure:adal4j;1.6.4
+com.microsoft.azure:azure-mgmt-graph-rbac;1.3.0
+com.microsoft.azure:azure-mgmt-keyvault;1.11.1
+com.microsoft.azure:azure-mgmt-resources;1.3.0
+test_jar_com.microsoft.azure:azure-mgmt-resources;1.3.1-SNAPSHOT
```
JonathanGiles: This seems like a weird naming convention and prone to people updating the wrong azure-mgmt-resources line.

JimSuplizio: @JonathanGiles this is the only one named like that. As soon as this is checked in there are going to be issues files for every POM version exception including this one. This is the only entry in the file like this and it's using a 2+ year old Jar.

parent/pom.xml
JimSuplizio: Agreed. I'll push another PR where I remove this.

JimSuplizio: Done. Latest commit has this removed.

eng/versioning/version_data.txt
```diff
@@ -23,21 +23,21 @@ com.microsoft.azure.cognitiveservices:azure-cognitiveservices-contentmoderator;1
 com.microsoft.azure.cognitiveservices:azure-cognitiveservices-customvision-prediction;1.1.0-beta.3;1.1.0-beta.3
 com.microsoft.azure.cognitiveservices:azure-cognitiveservices-customvision-training;1.1.0-beta.3;1.1.0-beta.3
 com.microsoft.azure.cognitiveservices:azure-cognitiveservices-faceapi;1.1.0-beta.1;1.1.0-beta.1
-com.microsoft.azure:azure-cosmos-parent;3.3.1;3.4.0
-com.microsoft.azure:azure-cosmos;3.3.1;3.4.0
-com.microsoft.azure:azure-cosmos-benchmark;3.3.1;3.4.0
+com.microsoft.azure:azure-cosmos-parent;3.4.0;3.5.0
+com.microsoft.azure:azure-cosmos;3.4.0;3.5.0
+com.microsoft.azure:azure-cosmos-benchmark;3.4.0;3.5.0
 com.microsoft.azure:azure-cosmos-examples;3.3.1;3.4.0
```
kushagraThapar: Why is this left untouched ?
Shouldn't azure-cosmos-examples be changed to 3.4.0 and 3.5.0?

sdk/cosmos/microsoft-azure-cosmos-benchmark/pom.xml
```diff
@@ -8,12 +8,12 @@ Licensed under the MIT License.
   <parent>
     <groupId>com.microsoft.azure</groupId>
     <artifactId>azure-cosmos-parent</artifactId>
-    <version>3.4.0</version> <!-- {x-version-update;com.microsoft.azure:azure-cosmos-parent;current} -->
+    <version>3.5.0</version> <!-- {x-version-update;com.microsoft.azure:azure-cosmos-parent;current} -->
```
kushagraThapar: Why are we skipping one version here to go to 3.5.0 from 3.4.0 ?

eng/versioning/external_dependencies.txt
```diff
@@ -0,0 +1,209 @@
...
+applicationinsights_com.microsoft.azure:azure-client-runtime;1.6.3
+applicationinsights_com.microsoft.azure:azure-client-authentication;1.6.3
+
+# everything under sdk\cosmos
+cosmos_org.mockito:mockito-core;1.10.19
+cosmos_com.google.guava:guava;27.0.1-jre
+cosmos_io.dropwizard.metrics:metrics-core;4.1.0
+cosmos_io.dropwizard.metrics:metrics-jvm;4.1.0
+cosmos_io.dropwizard.metrics:metrics-graphite;4.1.0
+
```
kushagraThapar: everything under sdk\cosmos ?

Is it everything that only cosmos SDK uses, and no one else uses (just cosmos set of external dependencies) ?

eng/versioning/version_data.txt
JimSuplizio: @kushagraThapar com.microsoft.azure:azure-cosmos-examples only has a 3.3.1 release, 3.4.0 has not been released so this version did not need to be changed.

sdk/cosmos/microsoft-azure-cosmos-benchmark/pom.xml
JimSuplizio: com.microsoft.azure:azure-cosmos-parent already has a 3.4.0 release, I just ticked this up to the next major.

JimSuplizio: This is related to the previous question and will be answered above.

eng/versioning/version_data.txt
JimSuplizio: As per our conversation I'll change the versions of benchmark/examples to the following with the next commit to this PR.
com.microsoft.azure:azure-cosmos-benchmark;3.4.0;3.4.0
com.microsoft.azure:azure-cosmos-examples;3.4.0;3.4.0

eng/versioning/external_dependencies.txt
JimSuplizio: It's not that nobody else uses these, it's that cosmos is using a different version (in these cases you've highlight the version is greater than the common version). Every dependency exception is going to receive an issue. If the version is greater than the common version then it'll be for the java team to look at upgrading the common version. If the version is less than the common version then the library owner will receive it for the purposes to look at upgrading their version to the common version. Ultimately, I'd like to see that entire section vanish.

eng/versioning/external_dependencies.txt
```diff
@@ -0,0 +1,209 @@
...
+org.apache.logging.log4j:log4j-core;2.11.1
+org.apache.qpid:proton-j;0.33.2
+org.asynchttpclient:async-http-client;2.10.4
+org.codehaus.groovy:groovy-eclipse-batch;2.5.8-01
+org.codehaus.groovy:groovy-eclipse-compiler;3.4.0-01
+org.powermock:powermock-api-mockito2;2.0.2
+org.powermock:powermock-module-junit4;2.0.2
+org.slf4j:slf4j-api;1.7.28
+org.slf4j:slf4j-log4j12;1.7.0
+org.slf4j:slf4j-simple;1.7.25
```
alzimmermsft: Where is this used as a non-test dependency? Generally we've only used the API module as a compile dependency as SLF4J is meant to be pluggable.

eng/versioning/external_dependencies.txt
```diff
@@ -0,0 +1,209 @@
...
+io.reactivex:rxjava;1.2.4
+org.apache.httpcomponents:httpclient;4.3.6
+org.apache.logging.log4j:log4j-api;2.11.1
+org.apache.logging.log4j:log4j-core;2.11.1
+org.apache.qpid:proton-j;0.33.2
+org.asynchttpclient:async-http-client;2.10.4
+org.codehaus.groovy:groovy-eclipse-batch;2.5.8-01
+org.codehaus.groovy:groovy-eclipse-compiler;3.4.0-01
+org.powermock:powermock-api-mockito2;2.0.2
+org.powermock:powermock-module-junit4;2.0.2
```
alzimmermsft: These should be test dependencies correct? At least the powermock modules should be test dependencies.

sdk/core/azure-core-test/pom.xml
```diff
@@ -44,37 +44,45 @@
...
       <groupId>io.projectreactor</groupId>
       <artifactId>reactor-test</artifactId>
+      <version>3.3.0.RELEASE</version> <!-- {x-version-update;io.projectreactor:reactor-test;external_dependency} -->
     </dependency>
 
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-api</artifactId>
+      <version>5.4.2</version> <!-- {x-version-update;org.junit.jupiter:junit-jupiter-api;external_dependency} -->
       <scope>compile</scope>
```
alzimmermsft: @srnagar @hemanttanwar @conniey Should this be a test dependency?

sdk/core/azure-core-test/pom.xml
```diff
@@ -44,37 +44,45 @@
     <dependency>
       <groupId>io.projectreactor</groupId>
       <artifactId>reactor-test</artifactId>
+      <version>3.3.0.RELEASE</version> <!-- {x-version-update;io.projectreactor:reactor-test;external_dependency} -->
     </dependency>
```
alzimmermsft: @srnagar @hemanttanwar @conniey Should this be a test dependency?

alzimmermsft: @JonathanGiles This may be rehashing an old discussion around whether Azure Core Test should exist as a standalone package or be part or Azure Core and generated as a test JAR.

eng/versioning/external_dependencies.txt
JimSuplizio: @alzimmermsft every place that slf4j-api is used it's a non-test dependency look in any of the following files
- sdk\appconfiguration\azure-data-appconfiguration\pom.xml
- sdk\core\azure-core\pom.xml
- sdk\core\azure-core-test\pom.xml
- sdk\cosmos\microsoft-azure-cosmos\pom.xml
- sdk\cosmos\microsoft-azure-cosmos-benchmark\pom.xml
- sdk\cosmos\microsoft-azure-cosmos-examples\pom.xml
- sdk\eventhubs\microsoft-azure-eventhubs\pom.xml
- sdk\identity\azure-identity\pom.xml
- sdk\servicebus\microsoft-azure-servicebus\pom.xml
- sdk\storage\azure-storage-blob\pom.xml
- sdk\storage\azure-storage-blob-batch\pom.xml
- sdk\storage\azure-storage-common\pom.xml
- sdk\storage\microsoft-azure-storage-blob\pom.xml

eng/versioning/external_dependencies.txt
JimSuplizio: I mostly left things under the same categories they were in under parent pom. If this need to be shifted in the file that can be done with a separate check in since, right now, I'm going to squash and merge before there are merge conflicts.

sdk/core/azure-core-test/pom.xml
JonathanGiles: I would like azure-core-test to be removed, but it is easier said than done. It is difficult at present due to JDK9+ modules and the visibility of some classes that we couldn't quite resolved.

PR #6449, opened 2019-11-20
Naming and documentation updates EventData, EventDataBatch, Builder

conniey:
* Suffixes buildProducer(), buildConsumer(), etc. with -Client()
* Update EventDataBatch.getCount()
* Fix documentation in EventData
* Fix documentation in package-info.
* SendOptions, and other model classes no longer use Cloneable.


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventData.java
```diff
@@ -256,6 +233,31 @@ public int hashCode() {
...
+
+    /**
+     * Adds a new key value pair to the existing context on Event Data.
+     *
+     * @param key The key for this context object
+     * @param value The value for this context object.
+     * @throws NullPointerException if {@code key} or {@code value} is null.
+     * @return The updated {@link EventData}.
+     */
+    EventData addContext(String key, Object value) {
```
srnagar: `getContext()` should be sufficient as the users can just do `getContext().addData(key, value).addData(key, value)`. Do we need this `addContext()` method in `EventData`?

conniey: The unfortunate thing is that addContext returns a new object, so they wouldn't be able to do the assignment back to `this.context`

PR #6451, opened 2019-11-21
Hotfix/storage 12.0.1

rickle-msft:


sdk/storage/azure-storage-blob/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log azure-storage-blob
 
+## Version XX.X.X (XXXX-XX-XX)
```
jaschrep-msft: Should this be updated?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -539,28 +538,26 @@ as we can guarantee we only need at most two buffers for any call to write (two
...
+
 
         // See ProgressReporter for an explanation on why this lock is necessary and why we use AtomicLong.
         AtomicLong totalProgress = new AtomicLong();
         Lock progressLock = new ReentrantLock();
 
         final SortedMap<Long, String> blockIds = new TreeMap<>();
-        return Flux.fromIterable(sliceFile(fileSize, finalParallelTransferOptions.getBlockSize()))
+        return Flux.fromIterable(sliceFile(fileSize, finalParallelTransferOptions.getBlockSize(),
+            parallelTransferOptions))
```
jaschrep-msft: Are file size and block size not part of parallel transfer options?

rickle-msft: fileSize is not. We get it from the file. We take in the parallelTransferOptions (not the final ones) because we want to be able to check if the customer explicitly set a block size so we don't overwrite it if they did.

sdk/storage/azure-storage-blob/CHANGELOG.md
rickle-msft: This won't get merged into master, so I'm not sure it matters. If it does, I'll do a separate PR on top of this so I don't have to restart CI for a readme change.

weshaggard: Right now this doesn't matter but we are working on better improvements to pull the changelog information into the release process as well as validating it so it will matter more in the future.

PR #6452, opened 2019-11-21
Renames for azure-core-amqp

conniey:
* Prefix RetryOptions, RetryPolicy, RetryMode, TransportType with Amqp.


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ReactorConnection.java
```diff
@@ -45,7 +45,7 @@
     private final ConnectionOptions connectionOptions;
     private final ReactorProvider reactorProvider;
     private final Disposable.Composite subscriptions;
-    private final RetryPolicy retryPolicy;
+    private final AmqpRetryPolicy retryPolicy;
 
     private ReactorExecutor executor;
     //TODO (conniey): handle failures and recreating the Reactor. Resubscribing the handlers, etc.
```
srnagar: Easier to track in github issues instead of having this todo here (if you haven't created one already)

PR #6453, opened 2019-11-21
Remove cloneable

conniey:
* RetryOptions and *RetryPolicy no longer implement Cloneable.
* ErrorContext and ErrorCondition are prefixed with Amqp.
* ExceptionUtil is moved into implementation class.

### DIFF
--- a/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/RetryOptions.java
+++ b/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/RetryOptions.java
@@ -28,7 +28,7 @@ public class RetryOptions {
     }
 
     /**
-     * Creates an instance configured with {@code retryOptions}.
+     * Creates an instance configured with {@code retryOptions}. This is not thread-safe.
      *
      * @param retryOptions Retry options to configure new instance with.
      * @throws NullPointerException if {@code retryOptions} is null.
--- a/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/exception/AmqpException.java
+++ b/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/exception/AmqpException.java
@@ -129,9 +129,9 @@ public class AmqpException extends AzureException {
     }
 
     /**
-     * Gets the ErrorCondition for this exception.
+     * Gets the {@link AmqpErrorCondition} for this exception.
      *
-     * @return The ErrorCondition for this exception, or {@code null} if nothing was set.
+     * @return The {@link AmqpErrorCondition} for this exception, or {@code null} if nothing was set.
      */
     public AmqpErrorCondition getErrorCondition() {
         return this.errorCondition;


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/RetryOptions.java
```diff
@@ -27,6 +27,20 @@ public RetryOptions() {
         retryMode = RetryMode.EXPONENTIAL;
     }
 
+    /**
+     * Creates an instance configured with {@code retryOptions}.
+     *
+     * @param retryOptions Retry options to configure new instance with.
+     * @throws NullPointerException if {@code retryOptions} is null.
+     */
+    public RetryOptions(RetryOptions retryOptions) {
```
srnagar: This is not thread-safe. While making it thread-safe may require adding synchronize to setter and getter methods and then synchronize on the input `retryOptions`, it's unnecessary overhead for this copy operation. So, it would just be better to document that the copy constructor doesn't guarantee thread-safety and users are responsible for synchronizing if they choose to use this constructor.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/exception/AmqpException.java
```diff
@@ -133,7 +133,7 @@ public boolean isTransient() {
      *
      * @return The ErrorCondition for this exception, or {@code null} if nothing was set.
```
srnagar: Update javadoc too.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/exception/AmqpException.java
```diff
@@ -98,7 +98,7 @@ public AmqpException(boolean isTransient, ErrorCondition errorCondition, Throwab
      * @param cause The Throwable which caused the creation of this AmqpException.
      * @param errorContext The context that caused this AMQP error.
      */
-    public AmqpException(boolean isTransient, String message, Throwable cause, ErrorContext errorContext) {
+    public AmqpException(boolean isTransient, String message, Throwable cause, AmqpErrorContext errorContext) {
```
srnagar: `errorContext` -> `amqpErrorContext`?

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/exception/AmqpException.java
```diff
@@ -133,7 +133,7 @@ public boolean isTransient() {
      *
      * @return The ErrorCondition for this exception, or {@code null} if nothing was set.
      */
-    public ErrorCondition getErrorCondition() {
+    public AmqpErrorCondition getErrorCondition() {
```
srnagar: `getAmqpErrorCondition()`

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/exception/AmqpException.java
```diff
@@ -142,7 +142,7 @@ public ErrorCondition getErrorCondition() {
      *
      * @return The context for this exception.
      */
-    public ErrorContext getContext() {
+    public AmqpErrorContext getContext() {
```
srnagar: Same here `getAmqpErrorContext()` and in other places that have getters and setters for the renamed types.

PR #6465, opened 2019-11-21
Paged Flux abstractions

anuchandy:


anuchandy:
PagedFluxCore type has following properties:

1. PagedFluxCore is not tied to protocol (http/tcp etc), it assumes each page is of any type that Extends from Page contract. Though the type Page is generic, it's exists under http package, if needed we can have a new PageCore then Page extends from it.
2. Each Subscription to PagedFluxCore will have an internal java.util.Map associated with it, this is where page-retriever can store state and access it across multiple calls. When SDK implementer creates PagedFluxCore, he can pass an initial map (with some state), this map will be used as the base for state-per-subscription.
3. PagedFluxCore Ctr takes only one Function, PageRetriever, we don't explicitly ask for two retrievers (first and next). This function will be called initially with null-continuation-token and all subsequent calls will have a non-null-continuation-token 
4. The PageRetriever returns Flux<T>, so PagedFluxCore don't care it emits one item or multiple items, all it care is if the last item (last Page) of the Flux has a continuation-token then that will be used for the next invocation of PageRetriever

anuchandy:
The generic abstraction don't assume there is a 1:1 mapping between a Page and a network-call, hence the pageRetriever returns Flux<T:Page>, if there is a 1:1 mapping then flux can emit one item, but if for whatever reason you want to emit multiple items (e.g. you retrieved multiple pages internally e.g. a range of pages) then you can do that.

anuchandy:
The only  assumptions generic abstraction make are: 

(1). Each Page has a pointer to the next page
(2). The last page will have this pointer as null

sdk/core/azure-core/src/main/java/com/azure/core/paging/PagedFluxCore.java
```diff
@@ -0,0 +1,91 @@
...
+                .doOnNext(page -> lastPageContinuationToken[0] = page.getContinuationToken()))
+            .takeUntil(page -> page.getContinuationToken() == null);
+    }
+
+    /**
+     * Subscribe to consume all items of type {@code T} in the sequence respectively.
+     * This is recommended for most common scenarios. This will seamlessly fetch next
+     * page when required and provide with a {@link Flux} of items.
+     *
+     * @param coreSubscriber The subscriber for this {@link PagedFluxBase}
```
JonathanGiles: Change this link to be PagedFluxCore

anuchandy:
A quick overview of types:

1. PagedFluxCore (PFC) = no public continuation token (can be internal via state), supports state management
2. ContinuableFluxCore (CFC) = supports public generic continuation token, supports state management
3. SimplePagedFlux (SPF) = supports public string continuation token, no state management

#### PagedFluxCore (PFC):

The "PagedFluxCore" is the core base type and all it require is Flux<P:PageCore<T>>. It doesn't enforce user facing continuation token constrain, hence "PageCore" contract contains only page items.

"PagedFluxCore" optionally support state through capture. The SDK implementor is free to use any number of objects for state, "PagedFluxCore" don't have to aware of type used for state. To support state capture, PagedFluxCore takes a "Provider (supplier) of Page retrieval Supplier".

#### ContinuablePagedFlux:

The "ContinuablePagedFlux" extends from "PagedFluxCore" and add support for continuation token. The continuation token is not tied to String, it can be anything that implements ContinuationToken contract.

The page type that "ContinuablePagedFlux" use is "ContinuablePage" which extends from "PageCore" and expose continuation token.

"ContinuablePagedFlux" also support optional capture based state management, hence it takes "Provider (Supplier) of Page Retriever Function that take continuation token". 

#### SimplePagedFlux:

"SimplePagedFlux" extends from "PagedFluxCore" and specialized to String based continuation token, it does not support state, it's for most of the pagination scenarios in Az SDKs.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
```diff
@@ -67,18 +66,31 @@ public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever,
         super(firstPageRetriever, nextPageRetriever);
     }
 
+    /**
+     * Creates an instance of {@link PagedFlux}. The constructor takes a provider, that when called should
+     * provides Page Retriever Function which accepts continuation token. The provider will be called for
```
alzimmermsft:
```suggestion
     * provide Page Retriever Function which accepts continuation token. The provider will be called for
```

anuchandy:
The `mapPage` method (which is marked as `@Deprecated`  in this PR) is used only by Data Lake SDK https://github.com/Azure/azure-sdk-for-java/blob/d6a8381fce62ea74c2658234004427ea0432748f/sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeServiceAsyncClient.java#L259 and

anuchandy:
@JonathanGiles  we talked about identifying the places where `mapPage` is used: It is used only by Data Lake SDK as the moment https://github.com/Azure/azure-sdk-for-java/blob/d6a8381fce62ea74c2658234004427ea0432748f/sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeServiceAsyncClient.java#L259

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
```diff
@@ -0,0 +1,156 @@
...
+
+    private final Supplier<Function<C, Flux<P>>> pageRetrieverProvider;
+    private final int defaultPrefetch;
+
+    /**
+     * Creates an instance of {@link ContinuablePagedFluxCore}.
+     *
+     * @param pageRetrieverProvider a provider that returns Page Retriever Function.
+     */
+    protected ContinuablePagedFluxCore(Supplier<Function<C, Flux<P>>> pageRetrieverProvider) {
```
conniey: It may be all the generic type arguments that are throwing me off, but why provide a supplier that returns a function rather than a reference to the function itself? (I was trying to see a concrete use case in the code snippets/samples but couldn't see when it would change.)

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param prefetch the number of Pages to be pre-fetched from the Page Retriever Function upon
+     *                 subscription
+     * @return a Flux of {@link ContinuablePage} identified by the given continuation token
+     */
+    public Flux<P> byPage(C continuationToken, int prefetch) {
+        if (continuationToken == null) {
+            return Flux.empty();
+        }
+        if (prefetch <= 0) {
+            throw new IllegalArgumentException("prefetch > 0 required but provided: " + prefetch);
```
conniey: Instead of throwing, should we be returning a Flux.error?

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuationState.java
```diff
@@ -0,0 +1,50 @@
...
+        this.lastContinuationToken = token;
+    }
+
+    /**
+     * Store the last seen continuation token.
+     *
+     * @param token the token
+     */
+    void setLastContinuationToken(C token) {
+        this.isDone = token == null ? true : false;
```
alzimmermsft:
```suggestion
        this.isDone = (token == null);
```

alzimmermsft:
Would this handle only requesting 1 page at a time and not trigger subsequent page requests? https://github.com/Azure/azure-sdk-for-java/issues/6518

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
anuchandy: This is to support optional state-management. 90% cases the string continuation-token is the state that flows between Page retrieval Function calls.  hence it will be

```java
() -> {
   return new Function<String, Flux<Page<Foo>>() {
     // page retrieval api call
   }
} 
```

When exploring Cosmos pagination we learned that some client SDK might need to manage internal state that is accessible between multiple call to Page Retrieval Function. Below is a made up example showcasing this use case.

```java
() -> {
    State state = new State();

    return new Function<String, Flux<Page<Foo>>() {
         if (firstCall) {
           result = make api call
           state.store(result.metadata)
           return Flux.just(new Page(result.items, result.link));
        } else {
           metadata = state.get();
           // use the metadata to prepare arguments for next call
           // result  = make api call
           return Flux.just(new Page(result.items, result.link));
        }
   }
}
```

This Supplier will be called for each Subscription to the Paged Flux instance, which means state per Subscription, the Page retrieval Function instance will capture this state, so throughout the life time of a Subscription the Page retrieval Function instance uses the same state.

anuchandy:
@alzimmermsft _Would this handle only requesting 1 page at a time and not trigger subsequent page requests_: its configurable in the `ContinuablePagedFluxCore` Ctr through prefetch argument. If not specified we default to the reactor prefetch.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
```diff
@@ -73,12 +97,12 @@ public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever,
      * @param mapper The mapper function to convert from type T to type S.
      * @param <S> The mapped type.
      * @return A PagedFlux of type S.
+     * @Deprecated
```
JonathanGiles: Use the Javadoc tag `@deprecated` and follow it with text on how users should respond to this deprecation.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFluxBase.java
```diff
@@ -38,15 +37,11 @@
  * @see Page
  * @see Flux
  */
-public class PagedFluxBase<T, P extends PagedResponse<T>> extends Flux<T> {
-
-    private final Supplier<Mono<P>> firstPageRetriever;
-
-    private final Function<String, Mono<P>> nextPageRetriever;
-
+public class PagedFluxBase<T, P extends PagedResponse<T>> extends ContinuablePagedFluxCore<String, T, P> {
```
JonathanGiles: It might be worth considering if this entire class should be deprecated now. I understand it is still the parent class of PagedFlux, but that is only while we don't make breaking changes.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePage.java
```diff
@@ -0,0 +1,20 @@
...
+ * Represents Page from service that has reference to next set of one or more pages.
+ *
+ * @param <C> Type of the continuation token
+ * @param <T> Type of the items in the page
+ */
+public interface ContinuablePage<C, T> {
+    /**
+     * @return list of items in the page.
+     */
+    List<T> getItems();
```
JonathanGiles: We had a discussion with Cosmos about making this be an Iterable or IterableStream. Did you have any reason to keep it as a List?

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFlux.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public abstract class ContinuablePagedFlux<O, C, T, P extends ContinuablePage<C, T>> extends Flux<T> {
+    /**
+     * @return a Flux of {@link ContinuablePage} that this Paged Flux represents.
+     */
+    public abstract Flux<P> byPage();
+    /**
+     * @return a Flux of {@link ContinuablePage} identified by the given continuation option.
+     */
+    public abstract Flux<P> byPage(O continuationOption);
```
JonathanGiles: For me this is the main area of contention. The fact that we have the byPage method take `O`, rather than `C`, and that we call this a `continuationOption` rather than a `continuationToken` (because it `O` and not `C`). My preference is to change this class to remove the `O` type, and to instead revert this back to being `C continuationToken`. Subclasses may choose to introduce the `O` type and a different byPage 'overload'. Barring objections from other reviewers (@conniey, @alzimmermsft, etc), this would be my recommendation.

sdk/core/azure-core/src/main/java/com/azure/core/paging/ContinuablePagedFlux.java
```diff
@@ -0,0 +1,119 @@
...
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * This class is a Flux that can operate on a type that extends {@link ContinuablePage} and also
+ * provides the ability to operate on individual items. This type support user providing continuation
+ * token and retrieve pages using such token.
```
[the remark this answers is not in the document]
anuchandy: thanks! Applied this suggestion. Agree we should do a full doc review.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
anuchandy: We originally introduced `mapPage` for addressing storage (I think search as well) client SDK use case, afaik there was no customer ask for it. I couldn't think of a scenario where customer need to transform a PagedFlux to another. 

Recently there is a discussion around mapping error for internal use. I'm bit worried adding more and more Flux operators in PagedFlux for such decorations for internal use-cases. Though provider pattern in this PR was originally designed for state mgmt, it can also extended to solve this, it keep the bar high for SDK implementor to do the decoration but avoid leaking more operators to public surface.

Added samples for decoration use case. I'm not sure how to link those from `@deprecated` tag.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFluxBase.java
anuchandy: yes, both `PagedFlux` and `PagedFluxBase` has exact same code in it. As you pointed the reason I didn't add deprecation tag is `PagedFluxBase` is base type for the "Stable" type `PagedFlux`.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFlux.java
anuchandy: totally agree, as you receommended removed the option  generic arg.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
anuchandy: right, I've updated the code.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFluxBase.java
anuchandy: @JonathanGiles is it still ok to have deprecated tag for PagedFluxBase? If so we can add it and recommend using `ContinuablePagedFluxCore`.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePage.java
anuchandy: The only reason I didn't replace `List` with  Iterable/IterableStream was to avoid breaking change. Should we add a getter that returns IterableStream and mark `List<T> getItems()` deprecated? how about the name `getElements()` for the new getter? thoughts?

sdk/core/azure-core/src/test/java/com/azure/core/http/rest/PagedIterableTest.java
```diff
@@ -139,16 +139,8 @@ public void streamFirstPage() {
...
-         * works it needs to begin the next publisher to determine whether onNext or onComplete should trigger. This
-         * results in 2 next page retrieval calls for the following reason:
-         *
-         * - Makes the initial get first page call, then needs to validate that get next page emits. 1 call made.
-         * - Retrieving the first page in verification moves the stream iterator to the initial next page, Reactor then
-         *   needs to verify that the page after it emits. 2 calls made.
-         */
-        assertEquals(2, pagedFlux.getNextPageRetrievals());
+        // The goal for this test is that 0 next page retrieval calls are made.
+        assertEquals(0, pagedFlux.getNextPageRetrievals());
```
anuchandy: @alzimmermsft - I've been thinking a bit to see how we can achieve the correct behavior here, i.e. avoiding 2 extra network calls consumer didn't ask us to perform, the new base type `ContinuablePagedFluxCore` will take care of this. I've updated the test to validate the expected next call count (i.e. 0). Please take a look and let me know if you've any concerns.

alzimmermsft: I have no concerns, this is a great side affect of this change!

alzimmermsft:
Any chance the a corresponding change is made to how `PagedIterable` works currently? https://github.com/Azure/azure-sdk-for-java/issues/6750 and https://github.com/Azure/azure-sdk-for-java/issues/6751 are the two sides of the sync and async paging concepts, this PR should resolve https://github.com/Azure/azure-sdk-for-java/issues/6750 but doesn't affect https://github.com/Azure/azure-sdk-for-java/issues/6751, would there be a follow-up PR to enable this concept for `PagedIterable`?

sdk/core/azure-core/src/test/java/com/azure/core/http/rest/PagedIterableTest.java
anuchandy: Cool, thanks for the review Alan.

anuchandy:
@alzimmermsft I will look into PagedIterable #6751 today.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePage.java
anuchandy: PR is updated by adding `ContinuablePage::getElements()`. In `Page` (that extends from `ContinuablePage`) the `getItems()` method is marked as deprecated with default implementation based `getElements()`.

anuchandy:
@alzimmermsft, regarding mapping error in `PagedIterable` (#6751): it looks like `PagedIterable` is backed by a `PagedFlux` instance, the Ctr signature is:

```java
public PagedIterable(PagedFlux<T> pagedFlux) {
```

So if the provided pagedFlux already has various error types mapped to one error type via decoration pattern I outlined previously then we don't need any special handing on `PagedIterable`. Right? Please let me know your thoughts.

sdk/core/azure-core/src/test/java/com/azure/core/http/rest/PagedIterableTest.java
anuchandy: and let's remove this FAQ entry once the PR is in: https://github.com/Azure/azure-sdk-for-java/wiki/Frequently-Asked-Questions#-why-does-synchronous-paged-iterator-fetches-2-pages-eagerly-when-iterating-by-page

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFluxBase.java
anuchandy: Marked as deprecated with suggestion to use ContinuablePagedFlux instead.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/ConfigurationSettingPage.java
```diff
@@ -30,12 +31,14 @@ public String getContinuationToken() {
...
-     * @return The list of {@link ConfigurationSetting ConfigurationSettings}.
+     * @return The iterable stream of {@link ConfigurationSetting ConfigurationSettings}.
      */
     @Override
-    public List<ConfigurationSetting> getItems() {
-        return items;
+    public IterableStream<ConfigurationSetting> getElements() {
+        return new IterableStream<>(items);
     }
+
```
conniey: nit; extra blank lines

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
```diff
@@ -64,21 +65,71 @@ public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever) {
...
+     * @param provider the Page Retrieval Provider
+     * @param ignored param is ignored, exists in signature only to avoid conflict with first ctr
+     */
+    // NOTES: Proposal for azure-core-v2:
+    //
+    // 1. Remove the first ctr "PagedFlux(Supplier<Mono<PagedResponse<T>>>)".
+    // 2. Add a new ctr        "PagedFlux(Supplier<PageRetriever<String, PagedResponse<T>>>)".
+    // 3. Remove the factory method "PagedFlux::create" and this PRIVATE ctr in favour of #2.
+    //
+    private PagedFlux(Supplier<PageRetriever<String, PagedResponse<T>>> provider, boolean ignored) {
```
conniey: It's too bad that we have to use a factory to create this. I think it makes it less discoverable since you can do `new Paged(...`, and if you want to use the cooler PageRetriever, figure out that instead of `new`, use `.create`

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/Page.java
```diff
@@ -3,26 +3,28 @@
...
      *
-     * @return A link to the next page, or {@code null} if there are no more results.
+     * @deprecated use {@link this#getElements()}.
      */
-    String getContinuationToken();
+    @Deprecated
+    default List<T> getItems() {
+        IterableStream<T> iterableStream = this.getElements();
+        return iterableStream == null
+            ? null
```
conniey: @JonathanGiles Do we have a guideline about when to return an empty collection and when to return null? I noticed this also in: https://github.com/Azure/azure-sdk-for-java/pull/6465/files#diff-f9bbb7ca67fb43df109d0ca82503fb7fR29 and a few other places.

I personally prefer returning an empty collection, because you'd know that there are no items, whereas, you'd have to do a null check, then check for empty if we return `null`.

sdk/core/azure-core/src/samples/java/com/azure/core/util/paging/PagedFluxCoreJavaDocCodeSnippets.java
```diff
@@ -0,0 +1,89 @@
...
+        class ContinuationState<C> {
+            private C lastContinuationToken;
+            private boolean isDone;
+
+            ContinuationState(C token) {
+                this.lastContinuationToken = token;
+            }
+
+            void setLastContinuationToken(C token) {
+                this.isDone = token == null ? true : false;
```
conniey: This can be simplified to: `this.isDone = token == null;`

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/PageRetriever.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+@FunctionalInterface
+public interface PageRetriever<C, P> {
+    /**
+     * Retrieve a set of one or more pages starting from the page identified by
+     * the given continuation token.
+     *
+     * @param continuationToken the token identifying the page set, a {@code null}
+     *                          value indicate that retrieve pages from the beginning
+     * @param pageSize the preferred number of items per page, a {@code null} value
```
conniey: Should we add some docs about what the behaviour will be if the pageSize is `< 1`?

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/Page.java
JonathanGiles: I don't think I have anything in the guidelines, but I strongly prefer to always return an empty collection rather than null.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/ItemPage.java
```diff
@@ -30,8 +31,10 @@
     private String continuationToken;
 
     @Override
-    public List<T> getItems() {
-        return items;
+    public IterableStream<T> getElements() {
+        return items == null
```
alzimmermsft: Same question about returning a null instead of an empty collection type.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponse.java
```diff
@@ -21,6 +24,9 @@
      * @return The items in the page.
      */
     default List<T> getValue() {
```
alzimmermsft: Should this be marked as deprecated as it is in Page?

srnagar: Maybe we can add @Deprecated annotation to this method too? This is redundant with the `getItems()` method  in `Page` this interface extends.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePage.java
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.util.paging;
```
srnagar: Not sure if this belongs in `util` - maybe just `com.azure.core.paging`?

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePage.java
```diff
@@ -0,0 +1,27 @@
...
+ * such a reference is known as continuation token.
+ *
+ * @param <C> Type of the continuation token
+ * @param <T> Type of the elements in the page
+ *
+ * @see ContinuablePagedFlux
+ */
+public interface ContinuablePage<C, T> {
+    /**
+     * @return an iterable stream of elements in the page.
```
srnagar: nit: add javadoc for method (it just has @return doc)

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFlux.java
```diff
@@ -0,0 +1,53 @@
...
+ * continuation tokens, allowing for restarting from a previously-retrieved continuation token.
+ *
+ * @param <C> the type of continuation token
+ * @param <T> the type of elements in the page
+ * @param <P> the type of page
+ *
+ * @see Flux
+ * @see ContinuablePage
+ */
+public abstract class ContinuablePagedFlux<C, T, P extends ContinuablePage<C, T>> extends Flux<T> {
```
srnagar: Since all methods in this class are `abstract` change this to an interface instead.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * Creates an instance of {@link ContinuablePagedFluxCore}.
+     *
+     * @param pageRetrieverProvider a provider that returns {@link PageRetriever}.
+     * @param defaultPageSize the default preferred page size
+     * @throws IllegalArgumentException if defaultPageSize is not greater than zero
+     */
+    protected ContinuablePagedFluxCore(Supplier<PageRetriever<C, P>> pageRetrieverProvider, int defaultPageSize) {
```
srnagar: nit: change name to `pageSize` instead of `defaultPageSize`

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
```diff
@@ -0,0 +1,228 @@
...
+        this.pageRetrieverProvider = Objects.requireNonNull(pageRetrieverProvider,
+            "'pageRetrieverProvider' function cannot be null.");
+        if (defaultPageSize <= 0) {
+            throw new IllegalArgumentException("defaultPageSize > 0 required but provided: " + defaultPageSize);
+        }
+        this.defaultPageSize = defaultPageSize;
+    }
+
+    /**
+     * @return the page size configured for this {@link ContinuablePagedFluxCore}, -1 if unspecified.
```
srnagar: Can we instead use `Integer` instead to indicate unspecified by just setting it to `null`?

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
anuchandy: I wish Java has a way to work around  type erasure which is causing this pain. As pointed in the code comments - proposal is to update the current ctr taking `Mono Supplier` with `PageRetriever Supplier`, but we've to wait for major version bump up for doing it.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/Page.java
anuchandy: agreed, it is good practice that our APIs exposing collection type to return EMPTY collection instead of NULL. The SDK author implementing `Page` or `ContinuablePage` MUST return an EMPTY IterableStream from `getElements()`, not NULL. The API we're discussing here is the default impl of Deprecated `getItems()` which uses `getElements()` underneath, so as long as SDK author implementing `getElements()` returns EMPTY stream, the `getItems()` will also return EMPTY list.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponse.java
anuchandy: `PagedResponse<T>` is both `Page<T>` and `Response<List<T>>`. When user consume it as `Response<List<T>>` type, say passing `PagedResponse<T>` instance to a method that takes `Response<List<T>>`, the `getItems()/getElements()` are not available in this context, so in this context if we deprecate `getValue()` then there is no alternate way to access the content.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/Page.java
```diff
@@ -3,26 +3,28 @@
...
-public interface Page<T> {
-
-    /**
-     * Gets a list of items returned from the service.
-     *
-     * @return A list of items from the service.
-     */
-    List<T> getItems();
-
+public interface Page<T> extends ContinuablePage<String, T> {
```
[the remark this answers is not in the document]
anuchandy: Two things here:
 
1. We need a base type to represent the general page contract, we don’t want that contract to be in `http` package since non-http based client libs might want to use it. A genuine concern from cosmos team was  - inheritance hierarchy of TCP based cosmos type containing core types from `http` package, the `Page` type is in `http` package.
2. Ideally we could merge `Page` and `ContinuablePage` and call it `Page` then move it out of `http` package but that is a breaking change.
 
So we’ve above two constraints - hence the type ContinuablePage and inheritance hierarchy as proposed in this PR.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
```diff
@@ -64,21 +65,71 @@ public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever) {
      */
     public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever,
                      Function<String, Mono<PagedResponse<T>>> nextPageRetriever) {
-        super(firstPageRetriever, nextPageRetriever);
+        this(() -> (continuationToken, pageSize) -> continuationToken == null
```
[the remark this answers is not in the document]
anuchandy: Storage  (e.g. in blob list) and many of the ARM services supports client driven pagination. Agree that a version of PageRetriever without pageSize can simplify implementation of client APIs that does not support page-size, but it brings a new type that is not used by customers but only by SDK implementors. I was thinking it is ok to keep the bar high for SDK implementor if that save us from adding a new type (and provide the type when customer ask for it)

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponse.java
anuchandy: Please see the comment: https://github.com/Azure/azure-sdk-for-java/pull/6465#discussion_r363119218

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePage.java
anuchandy: The poller abstractions are in `util` namespace so thought of being consistent with it, hence used the same package for paging abstractions. @JonathanGiles thoughts?

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFlux.java
anuchandy: It's an abstract class because it needs to extend from reactor Flux class.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponse.java
anuchandy: So my concern is - Is it legit to mark  `PagedResponse<T>::getValue()` "deprecated" when the contract `Response<T>::getValue()` it implements is "not deprecated". @JonathanGiles thoughts?

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePage.java
JonathanGiles: My main line of thinking is that I don't particularly want too many top-level packages. I am personally fine with it under `util`.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponse.java
JonathanGiles: Anu's final point is the right perspective on this. We can't say that `getValue` is deprecated in `PagedResponse` without first deprecating `getValue` in Response. Before this PR is merged though we should very carefully review all deprecations to ensure that they are correct and appropriate.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
anuchandy: PR is updated to use `pageSize`.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
anuchandy: agree, updated to use null instead of -1.

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePage.java
anuchandy: Updated the java doc

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/PageRetriever.java
anuchandy: As Srikant suggested, we now use `null` instead of -1, so no more -1.

sdk/core/azure-core/src/samples/java/com/azure/core/util/paging/PagedFluxCoreJavaDocCodeSnippets.java
anuchandy: yes, updated the PR.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/ConfigurationSettingPage.java
anuchandy: PR is updated to revert this.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponse.java
anuchandy: thanks Jonathan for the clarification, learned something new about deprecation.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
srnagar: Agree. Maybe this can be tracked in a github issue with appropriate labels so we don't forget to make this change when working on a major version release.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
```diff
@@ -64,21 +65,71 @@ public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever) {
...
+    /**
+     * Creates an instance of {@link PagedFlux} backed by a Page Retriever Supplier (provider).
+     * When invoked provider should return {@link PageRetriever}. The provider will be called for each
+     * Subscription to the PagedFlux instance. The Page Retriever can get called multiple times in serial
+     * fashion, each time after the completion of the Flux returned from the previous invocation.
+     * The final completion signal will be send to the Subscriber when the last Page emitted by the Flux
+     * returned by Page Retriever has {@code null} continuation token.
+     *
+     * The provider is useful mainly in two scenarios:
+     * 1. To manage state across multiple call to Page Retrieval within the same Subscription
```
srnagar: Use `<li>` instead so that the javadoc is properly formatted on html.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFluxBase.java
```diff
@@ -70,34 +67,38 @@ public PagedFluxBase(Supplier<Mono<P>> firstPageRetriever) {
...
+    /**
+     * PACKAGE INTERNAL CONSTRUCTOR, exists only to support the PRIVATE PagedFlux.ctr(Supplier, boolean)
+     * use case.
+     *
+     * Create PagedFlux backed by Page Retriever Function Supplier.
+     *
+     * @param provider the Page Retrieval Provider
+     * @param ignored ignored
+     */
+    public PagedFluxBase(Supplier<PageRetriever<String, P>> provider, boolean ignored) {
```
srnagar: Looks like you intended to make this a package-private ctr but it's marked as `public`.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedResponseBase.java
```diff
@@ -62,8 +63,10 @@ public PagedResponseBase(HttpRequest request, int statusCode, HttpHeaders header
      * {@inheritDoc}
      */
     @Override
-    public List<T> getItems() {
-        return items;
+    public IterableStream<T> getElements() {
+        return items == null
+            ? null
```
srnagar: Prefer to return empty IterableStream as [described here](https://github.com/Azure/azure-sdk-for-java/pull/6465/files?file-filters%5B%5D=.java&file-filters%5B%5D=.xml#r363119004)

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
```diff
@@ -0,0 +1,228 @@
...
+ * This indirectly associate the state objects to the Subscription. The Page Retriever can get called
+ * multiple times in serial fashion, each time after the completion of the Flux returned by the previous
+ * invocation. The final completion signal will be send to the Subscriber when the last Page emitted by
+ * the Flux returned by the Page Retriever has {@code null} continuation token.
+ *
+ * <p><strong>Extending PagedFluxCore for Custom Continuation Token support</strong></p>
+ * {@codesnippet com.azure.core.util.paging.pagedfluxcore.continuationtoken}
+ *
+ * @param <T> The type of elements in a {@link ContinuablePage}
+ * @param <P> The {@link ContinuablePage} holding items of type {@code T}.
```
srnagar: Add `@param <C>` as well here

sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
```diff
@@ -0,0 +1,228 @@
...
+        this.pageRetrieverProvider = Objects.requireNonNull(pageRetrieverProvider,
+            "'pageRetrieverProvider' function cannot be null.");
+        if (pageSize <= 0) {
+            throw new IllegalArgumentException("pageSize > 0 required but provided: " + pageSize);
+        }
+        this.defaultPageSize = pageSize;
+    }
+
+    /**
+     * @return the page size configured for this {@link ContinuablePagedFluxCore}, {@code null} if unspecified.
```
srnagar: Include method javadoc too.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFluxBase.java
JonathanGiles: This JavaDoc says package internal but the constructor is public?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/ItemPage.java
JonathanGiles: I agree with Alan

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/Page.java
anuchandy: updated to return empty list instead of null

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/ItemPage.java
anuchandy: yes, updated to return empty iterable-stream.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedFlux.java
```diff
@@ -64,21 +65,71 @@ public PagedFlux(Supplier<Mono<PagedResponse<T>>> firstPageRetriever) {
...
+     * Create PagedFlux backed by Page Retriever Function Supplier.
+     *
+     * @param provider the Page Retrieval Provider
+     * @param ignored param is ignored, exists in signature only to avoid conflict with first ctr
+     */
+    // NOTES: Proposal for azure-core-v2:
+    //
+    // 1. Remove the first ctr "PagedFlux(Supplier<Mono<PagedResponse<T>>>)".
+    // 2. Add a new ctr        "PagedFlux(Supplier<PageRetriever<String, PagedResponse<T>>>)".
+    // 3. Remove the factory method "PagedFlux::create" and this PRIVATE ctr in favour of #2.
```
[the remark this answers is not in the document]
anuchandy: done: https://github.com/Azure/azure-sdk-for-java/issues/6610#issuecomment-575419690

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/ConfigurationSettingPage.java
```diff
@@ -30,12 +31,12 @@ public String getContinuationToken() {
...
+     * Gets the iterable stream of {@link ConfigurationSetting ConfigurationSettings} on this page.
      *
-     * @return The list of {@link ConfigurationSetting ConfigurationSettings}.
+     * @return The iterable stream of {@link ConfigurationSetting ConfigurationSettings}.
      */
     @Override
-    public List<ConfigurationSetting> getItems() {
-        return items;
+    public IterableStream<ConfigurationSetting> getElements() {
+        return items == null ? IterableStream.empty() : new IterableStream<>(items);
```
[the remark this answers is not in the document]
anuchandy: added `IterableStream.of(Iterable<T>)` factory method:  if parameter to this factory is `null` then it return cached static empty stream otherwise it create a new `IterableStream` from the parameter.

anuchandy:
@hemanttanwar - Couple of client-lib pom are updated to refer "unreleased version of azure_core" which has pagination improvements. Could you please take a look at those and if all good let's get this in.

anuchandy:
@hemanttanwar thanks, as discussed I ran the following command on this changes:

```
python eng/versioning/update_versions.py --ut library --bt client --sr
```

and it didn't update any pom files which indicates that the current changes are correct.

PR #6467, opened 2019-11-21
Rename contentMD5 to contentMD5 in Datalake and Shares

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/6141


alzimmermsft:
> 
> 
> No major issue. It is better to disable the import ordering. Unnecessary to have.

Codegen applies parameter order, the reason why it is different is someone manually changed the files in the past.

PR #6470, opened 2019-11-21
Removed Unsafe Buffer Duplication

### LINKED ISSUE #5912: [BUG] Requires live mode messes with data in certain scenarios.
Given the following test helper method: 
```
def compareDataToFile(Flux<ByteBuffer> data, File file) {
        FileInputStream fis = new FileInputStream(file)

        for (ByteBuffer received : data.toIterable()) {
            byte[] readBuffer = new byte[received.remaining()]
            fis.read(readBuffer)
            for (int i = 0; i < received.remaining(); i++) {
                if (readBuffer[i] != received.get(i)) {
                    return false
                }
            }
        }

        fis.close()
        return true
    }
```
being called from a test that downloads a blob and passes the returned Flux in along with the sourceFile that was uploaded produces different results depending on the presence of the @Requires( {liveMode()} ) annotation. 

For dataSizes over a few MB, the test will consistently fail. More specifically, what appears to happen is that the presence of the annotation causes the items coming from the result of toIterable to be out of order. Small data sizes do not fail because there is only one emission, but anything with more than one emission will fail. The data downloaded is correct (which can be observed by downloading to a file and comparing those results instead). The test behavior changes reliably by toggling the presence of this annotation, but it is not clear why.

alzimmermsft:
Fixes #5912 

Fixes an issue where testing Blob Cryptography could fail when interacting with a `DirectByteBuffer`. `BlobDecryptionPolicy` was using an optimization where it duplicated a `ByteBuffer`, duplication is a shallow reference copy. The copied buffer could be mutated, such as the case of buffered pools in Netty Reactor, which would result in the content having different state than expected. The optimization has been removed and the decrypted data is being stored in a newly allocated buffer every time.

### DIFF
--- a/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
@@ -363,8 +363,8 @@ public class BlobDecryptionPolicy implements HttpPipelinePolicy {
                 if (encryptionData.getWrappedContentKey().getKeyId().equals(keyId)) {
                     return Mono.just(this.keyWrapper);
                 } else {
-                    throw logger.logExceptionAsError(Exceptions.propagate(new IllegalArgumentException("Key mismatch. " +
-                        "The key id stored on the service does not match the specified key.")));
+                    throw logger.logExceptionAsError(Exceptions.propagate(new IllegalArgumentException("Key mismatch. "
+                        + "The key id stored on the service does not match the specified key.")));
                 }
             });
         }


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
```diff
@@ -215,11 +205,10 @@
                             }
                         }
                         totalInputBytes.addAndGet(bytesToInput);
-                        plaintextByteBuffer.position(0); // Reset the position after writing.
 
                         // This is to get the decryptedBytes value into the next map to adjust the position of the
                         // output buffer because the data range was increased for decryption.
-                        plaintextByteBuffer.limit(decryptedBytes);
```
sima-zhu: The decryptedBytes is useless in current impl

sima-zhu: Is it better to set limit to decryptedBytes, and set pos to 0? Why leave some padding?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
```diff
@@ -215,11 +205,10 @@
...
+                        plaintextByteBuffer.flip();
```
rickle-msft: I'm not confident this change preserves the behavior. DecryptedBytes is the number of bytes stored in the output buffer whereas flip sets the limit to the position of plaintextByteBuffer, which is how many bytes were read. These might not be equivalent in the cases where plaintextByteBuffer didn't align to a block boundary, so some of the data is still held in the cipher, or where we called doFinal and there was padding that was read but was stripped by the cipher and therefore not output.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobDecryptionPolicy.java
alzimmermsft: Updated comment.

alzimmermsft:
@rickle-msft Should I close or merge this PR?

rickle-msft:
If we still think the original is unsafe, let's merge the fix.

PR #6471, opened 2019-11-21
Prepare Identity to use December release version.

hemanttanwar:
December version update for identity library to 1.0.1

### DIFF
--- a/sdk/template/azure-sdk-template/pom.xml
+++ b/sdk/template/azure-sdk-template/pom.xml
@@ -11,7 +11,7 @@
 
   <groupId>com.azure</groupId>
   <artifactId>azure-sdk-template</artifactId>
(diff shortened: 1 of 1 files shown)


eng/jacoco-test-coverage/pom.xml
```diff
@@ -80,7 +80,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-identity</artifactId>
-      <version>1.1.0-beta.1</version> <!-- {x-version-update;com.azure:azure-identity;current} -->
+      <version>1.0.1</version> <!-- {x-version-update;com.azure:azure-identity;current} -->
```
alzimmermsft: Is this meant to be a patch or minor version bump?

hemanttanwar: We spoke in standup yesterday  and we are good with 1.0.1 version.

PR #6472, opened 2019-11-21
More azure-core-amqp, Event Hubs renames

conniey:
* Rename MessageConstant -> AmqpMessageConstant
* Rename CBS -> ClaimsBasedSecurity
* EventDataBatch: Expose getMaxSize
* EventPosition: Remove fromOffset(long, bool) method
* Rename: getProperties() -> getEventHubProperties().
* Remove final from RetryPolicy.

### DIFF
--- a/sdk/eventhubs/pom.service.xml
+++ b/sdk/eventhubs/pom.service.xml
@@ -36,10 +36,10 @@
         <module>../core/azure-core-http-netty</module>
         <module>../identity/azure-identity</module>
         <module>../core/azure-core-amqp</module>
-<!--        <module>../storage/azure-storage-common</module>-->
-<!--        <module>../storage/azure-storage-blob</module>-->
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/pom.service.xml
```diff
@@ -36,10 +36,10 @@
         <module>../core/azure-core-http-netty</module>
         <module>../identity/azure-identity</module>
         <module>../core/azure-core-amqp</module>
-        <module>../storage/azure-storage-common</module>
-        <module>../storage/azure-storage-blob</module>
+<!--        <module>../storage/azure-storage-common</module>-->
```
srnagar: These should be uncommented.

conniey: Oh shoot. I actually committed that. good catch.

PR #6487, opened 2019-11-21
Remove secret env loading from system

sima-zhu:


sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
```diff
@@ -128,7 +128,7 @@
     private static final Configuration GLOBAL_CONFIGURATION = new Configuration();
 
     private static final String LOADED_FROM_RUNTIME = "Loaded {} from runtime parameters with value {}.";
```
alzimmermsft: Need to update this one as well.

sima-zhu: Done.

Issue #6489, opened 2019-11-21
AppConfig tests throw error log for listing APIs in afterTest


sima-zhu:
**Reproduce:**
Set the AZURE_LOG_LEVEL=2. Run any test under app config.

**Error Received:**
When we log the error out from AppCofig tests. 
It print error at the afterTest step:
```
[main] INFO com.azure.data.appconfiguration.ConfigurationClientTest - Cleaning up created key values.
SLF4J: Failed toString() invocation on an object of type [com.azure.data.appconfiguration.models.SettingSelector]
Reported exception:
java.util.MissingFormatArgumentException: Format specifier '%s'
	at java.util.Formatter.format(Formatter.java:2519)
	at java.util.Formatter.format(Formatter.java:2455)
	at java.lang.String.format(String.java:2940)
	at com.azure.data.appconfiguration.models.SettingSelector.toString(SettingSelector.java:190)
```

**Excepted:**
No error outputs.

PR #6491, opened 2019-11-21
MissingFormatArgumentException

mssfang:
The String format that doesn't match with argument. 

fixes: #6489


mssfang:
After offline discussion. The change stays the same.

PR #6520, opened 2019-11-22
Added https enforce for bearer token

gapra-msft:


sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/implementation/util/BuilderHelper.java
```diff
@@ -195,12 +217,36 @@ private static HttpPipelinePolicy getResponseValidationPolicy() {
...
+     * Validates that the client is properly configured to use https.
+     *
+     * @param objectToCheck The object to check for.
+     * @param objectName The name of the object.
+     * @param endpoint The endpoint for the client.
+     * @param logger {@link ClientLogger} used to log any exception.
+     */
+    public static void httpsValidation(Object objectToCheck, String objectName, String endpoint, ClientLogger logger) {
+        if (objectToCheck != null && !parseEndpoint(endpoint, logger).getScheme().equals(Constants.HTTPS)) {
+            throw new IllegalArgumentException("Using a(n) " + objectName + " requires https");
```
sima-zhu: Use logger.logAndThrow. Same to other places.

gapra-msft:
> I remember generateSas and Azurites has some other way of getting credential.
> Azurites can take http as endpoint. Might need to double check if this is backward compatible.

This is just adding a https check for the bearer token form of authentication

sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/implementation/util/BuilderHelper.java
gapra-msft: Updated to use logger

PR #6522, opened 2019-11-22
Properly Redact Query String

### LINKED ISSUE #6521: HttpLoggingPolicy May Log Sensitive Query Parameters
`HttpLoggingPolicy` has a code path which redacts illegal query parameters from being logged but another log statement simply logs the `URL` which will lead to these query parameters being logged anyway. The `URL` logging should be updated to redact the query parameters and then use them as the query string in the `URL` after that is should log the URL.

alzimmermsft:
Fixes #6521 

Updates the `HttpLoggingPolicy`'s request `URL` logging to create a cloned version of the `URL` and update the query string with the redacted version. Before this change two sets of query strings would be logged, the query string from `URL` `toString` which would include all query strings and the redacted variant, these code paths have been merged.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
@@ -183,8 +183,7 @@ public class HttpLoggingPolicy implements HttpPipelinePolicy {
         return queryStringBuilder.toString();
     }
 
-    private Function<HttpResponse, Mono<HttpResponse>> logResponseDelegate(final ClientLogger logger,
-        final URL url,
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -143,36 +151,41 @@ private void formatAllowableHeaders(Set<String> allowedHeaderNames, HttpHeaders
...
             }
         }
+
+        return queryStringBuilder.toString();
     }
 
     private Function<HttpResponse, Mono<HttpResponse>> logResponseDelegate(final ClientLogger logger,
-                                                                           final URL url,
-                                                                           final long startNs) {
+        final URL url,
```
srnagar: fix formatting. `url` and `startNs` will fit in the same line too

PR #6523, opened 2019-11-22
azure-core-amqp: Consolidate getEndpointStates and naming changes.

conniey:
This is the last of the azure-core-amqp API review changes.
* getEndpointStates() is the way we figure out the connection state of an AMQP link, session or connection. If the Flux closes or errors, that implies that the link itself is closed.
  * We had duplicated this work using getErrors() and getShutdownSignals().
* Classes now use AutoCloseable

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConnection.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubConnection.java
@@ -73,8 +73,8 @@ class EventHubConnection implements Closeable {
     }
 
     /**
-     * Creates or gets a send link. The same link is returned if there is an existing send link with the same {@code
-     * linkName}. Otherwise, a new link is created and returned.
+     * Creates or gets a send link. The same link is returned if there is an existing send link with the same
+     * {@code linkName}. Otherwise, a new link is created and returned.
      *
      * @param linkName The name of the link.
      * @param entityPath The remote address to connect to for the message broker.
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/CbsAuthorizationType.java
```diff
@@ -8,7 +8,7 @@
 /**
  * An enumeration of supported authorization methods with the {@link ClaimsBasedSecurityNode}.
  */
-public enum CBSAuthorizationType {
+public enum CbsAuthorizationType {
```
srnagar: Do we need to expand CBS here just like we have done for other types?

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ReactorSender.java
```diff
@@ -129,6 +143,11 @@
         );
     }
 
+    @Override
+    public Flux<AmqpEndpointState> getEndpointStates() {
+        return connectionStates;
```
srnagar: rename this to `endpointStates`

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/CbsAuthorizationType.java
conniey: CbsAuthorizationType is used in the ClaimsBasedSecurityNode class, I thought it would be overkill.

PR #6529, opened 2019-11-24
[Event Hubs] Replace constant strings with resource strings #3923

mclay:
- proposal implementation for azure-messaging-eventhubs

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
@@ -155,7 +155,7 @@ class EventHubMessageSerializer implements MessageSerializer {
             return (T) toEventHubProperties(amqpBody);
         } else {
             throw logger.logExceptionAsError(new IllegalArgumentException(String.format(
-                "Class '%s' is not a supported deserializable type.", deserializedType)));
+                Messages.CLASS_NOT_A_SUPPORTED_TYPE, deserializedType)));
         }
     }
 
@@ -423,7 +423,7 @@ class EventHubMessageSerializer implements MessageSerializer {
             return Double.BYTES;
         }
 
-        throw new IllegalArgumentException(String.format("Encoding Type: %s is not supported",
+        throw new IllegalArgumentException(String.format(Messages.ENCODING_TYPE_NOT_SUPPORTED,
             obj.getClass()));
     }
 
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/Messages.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/Messages.java
@@ -17,6 +17,8 @@ public enum Messages {
     private static final ClientLogger LOGGER = new ClientLogger(Messages.class);
     private static Properties properties;
     private static final String MESSAGES_PROPERTIES_PATH = "com/azure/messaging/eventhubs/messages.properties";
+    public static final String CLASS_NOT_A_SUPPORTED_TYPE = getMessage("CLASS_NOT_A_SUPPORTED_TYPE");
+    public static final String ENCODING_TYPE_NOT_SUPPORTED = getMessage("ENCODING_TYPE_NOT_SUPPORTED");
     public static final String PROCESS_SPAN_SCOPE_TYPE_ERROR = getMessage("PROCESS_SPAN_SCOPE_TYPE_ERROR");
     public static final String MESSAGE_NOT_OF_TYPE = getMessage("MESSAGE_NOT_OF_TYPE");
     public static final String REQUEST_VALUE_NOT_VALID = getMessage("REQUEST_VALUE_NOT_VALID");
@@ -39,9 +41,13 @@ public enum Messages {
         properties = new Properties();
         try (InputStream inputStream =
                  Thread.currentThread().getContextClassLoader().getResourceAsStream(MESSAGES_PROPERTIES_PATH)) {
-            properties.load(inputStream);
+            if (inputStream != null) {
+                properties.load(inputStream);
+            } else {
+                LOGGER.error("Message properties [{}] not found", MESSAGES_PROPERTIES_PATH); //NON-NLS
+            }
         } catch (IOException exception) {
-            LOGGER.error("Error loading message properties {}", Messages.class, exception); //NON-NLS
+            LOGGER.error("Error loading message properties [{}]", MESSAGES_PROPERTIES_PATH, exception); //NON-NLS
         }
         return properties;
     }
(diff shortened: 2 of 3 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/Messages.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * I18n messages loaded from the messages.properties file located within the same package.
+ */
+public enum Messages {
+    ;
```
JonathanGiles: :-)

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
```diff
@@ -215,7 +215,7 @@ private EventData deserializeEventData(Message message) {
             body = bodyData.getValue().getArray();
         } else {
             logger.warning(String.format(Locale.US,
```
conniey: @JonathanGiles @mclay Do you know how the string.format works with localisation? Is there something similar to CultureInfo.CurrentUICulture that C# has, which knows what the locale of the user is?

JonathanGiles: `Locale.getDefault()`

conniey: Ahh gotcha. Should remove the `String.format(Locale)` usage in favour of the overload without? from the docs, it looks like it uses Locale.getDefault(), by default with that overload.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/Messages.java
conniey: Is this why the checkstyle suppression exists?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/Messages.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final ClientLogger LOGGER = new ClientLogger(Messages.class);
```
conniey: @JonathanGiles Did we have a checkstyle for not having static final loggers? I see why it's used here though in that .getProperties() method below. 🤔

JonathanGiles: I think we did have a check for it....not sure why it isn't triggering.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/Messages.java
mclay: Yes

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/Messages.java
mclay: i excluded the check for messages to report errors during static initialisation.  


  <suppress checks="com.azure.tools.checkstyle.checks.GoodLoggingCheck" files="Messages.java"/>

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
mclay: resolved

mclay:
Hi @JonathanGiles @conniey,

could you please review the resolved issues again?

many thanks
Michael

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
```diff
@@ -155,7 +154,7 @@ public int getSize(Message amqpMessage) {
         } else if (deserializedType == EventHubProperties.class) {
             return (T) toEventHubProperties(amqpBody);
         } else {
-            throw logger.logExceptionAsError(new IllegalArgumentException(String.format(Locale.US,
+            throw logger.logExceptionAsError(new IllegalArgumentException(String.format(
                 "Class '%s' is not a supported deserializable type.", deserializedType)));
```
conniey: Should this be part of the messages.properties?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubMessageSerializer.java
```diff
@@ -426,7 +423,7 @@ private static int sizeof(Object obj) {
             return Double.BYTES;
         }
 
-        throw new IllegalArgumentException(String.format(Locale.US, "Encoding Type: %s is not supported",
+        throw new IllegalArgumentException(String.format("Encoding Type: %s is not supported",
```
conniey: Should this be in messages.properties?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/Messages.java
```diff
@@ -0,0 +1,56 @@
...
+    private static synchronized Properties getProperties() {
+        if (properties != null) {
+            return properties;
+        }
+        properties = new Properties();
+        try (InputStream inputStream =
+                 Thread.currentThread().getContextClassLoader().getResourceAsStream(MESSAGES_PROPERTIES_PATH)) {
+            properties.load(inputStream);
+        } catch (IOException exception) {
+            LOGGER.error("Error loading message properties {}", Messages.class, exception); //NON-NLS
```
conniey: What happens if we are unable to load the inputStream.. Do the messages still show? Should we be rethrowing this IOException? @JonathanGiles ?

mclay: If the messages couldn't be loaded ...

- there will be a error log indicating the cause of the failure
- the message text will not be shown but the _key_ of the message will be and the application continues to run...

I would not recommend re-throwing because since this called within a static initializer it will cause NoClassDefFoundError or ExceptionInInitializerError.


@conniey could you please review again

conniey:
Thanks!

Issue #6515, opened 2019-11-22
Cannot use IterableStream with Schedulers.parallel() or Schedulers.single()


conniey:
**Repo**:
1. Copy the snippet below
1. Execute test.

**Expected**: It passes
**Actual**:
```
java.lang.AssertionError: expectation "assertNext" failed
(expected: onNext(); actual: onError(java.lang.IllegalStateException: 
Iterating over a toIterable() / toStream() is blocking, which is not supported in thread parallel-1))
```

**Snippet**
```java
import com.azure.core.util.IterableStream;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.Test;
import reactor.core.publisher.Flux;
import reactor.core.publisher.Mono;
import reactor.core.scheduler.Scheduler;
import reactor.core.scheduler.Schedulers;
import reactor.test.StepVerifier;

import java.util.Arrays;

class IterableStreamTest {
    @Test
    public void repro() {
        Foo foo = new Foo();

        // Also fails with Schedulers.single().
        Scheduler scheduler = Schedulers.parallel();

        System.out.println("Using array.");
        StepVerifier.create(foo.getPartitionIdsFromArray().subscribeOn(scheduler))
            .assertNext(id -> Assertions.assertEquals("0", id))
            .assertNext(id -> Assertions.assertEquals("1", id))
            .assertNext(id -> Assertions.assertEquals("2", id))
            .expectComplete()
            .verify(Duration.ofSeconds(5));

        System.out.println("Using IterableStream.");
        StepVerifier.create(foo.getPartitionIds().subscribeOn(scheduler))
            .assertNext(id -> Assertions.assertEquals("0", id))
            .assertNext(id -> Assertions.assertEquals("1", id))
            .assertNext(id -> Assertions.assertEquals("2", id))
            .expectComplete()
            .verify(Duration.ofSeconds(5));
    }

    static class Foo {
        Mono<Bar> getProperties() {
            return Mono.fromCallable(() -> new Bar(new String[]{"0", "1", "2"}));
        }

        Flux<String> getPartitionIds() {
            return getProperties().flatMapMany(x -> Flux.fromIterable(x.getPartitionIds()));
        }

        Flux<String> getPartitionIdsFromArray() {
            return getProperties().flatMapMany(x -> Flux.fromArray(x.getPartitionIdsArray()));
        }
    }

    static class Bar {
        private final IterableStream<String> partitionIds;
        private final String[] partitionArrays;

[...]

PR #6534, opened 2019-11-25
Add Iterable<T> overload for IterableStream<T>

conniey:
* Adds an overload to create an IterableStream using Iterable.
* This considers the case where users already have a synchronous collection, and when they transform it to a Flux, then back again, may result in an IllegalStateException.
* Adds tests

See #6515 and https://github.com/reactor/reactor-core/issues/1972

### DIFF
--- a/sdk/core/azure-core/src/test/java/com/azure/core/util/IterableStreamTests.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/util/IterableStreamTests.java
@@ -1,0 +1,254 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.util;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Flux;
+import reactor.core.scheduler.Schedulers;
+
+import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/util/IterableStream.java
```diff
@@ -28,15 +32,30 @@
...
+    }
+
+    /**
+     * Creates an instance with the given {@link Iterable}.
+     *
+     * @param iterable Collection of items to iterate over.
+     * @throws NullPointerException if {@code iterable} is {@code null}.
+     */
+    public IterableStream(Iterable<T> iterable) {
+        this.iterable = Objects.requireNonNull(iterable, "'iterable' cannot be null.");
```
srnagar: Instead of having `flux` and `iterable` fields in this class, if an instance of IterableStream is created using an iterable, can this just be converted to `this.flux = Flux.fromIterable(Objects.requireNonNull(iterable, "'iterable' cannot be null."));`. Simplifies code in other methods too where you don't have to check if flux is null or iterable is null.

conniey: Unfortunately this will result in that illegal state exception again because we are moving from sync to async world then back again.

PR #6537, opened 2019-11-25
Version update for december release

hemanttanwar:
Planning to release following two for azure-core

com.azure:azure-core 1.0.0 (GA)--> 1.1.0           
com.azure:azure-core-http-netty 1.0.0(GA)  - > 1.1.0


weshaggard:
What about the versions in the readmes? Don't we need to update those as well?

Issue #6518, opened 2019-11-22
[BUG] PagedIterable.iterableByPage(token) processes listing operations eagerly


rickle-msft:
**Describe the bug**
Calling pagedIterable.iterableByPage(token) results in a network call to fetch the next page using the given token. Calling .iterator().next() on this iterable results in an identical network call. 

I believe it's because the iterableByPage(token) method eagerly calls the nextPageFunction.apply() instead of storing the token and waiting for a call to next. 

**Code Snippet**
```
        int NUM_BLOBS = 10
        int PAGE_SIZE = 6
        for (int i = 0; i < NUM_BLOBS; i++) {
            PageBlobClient bc = cc.getBlobClient(generateBlobName()).getPageBlobClient()
            bc.create(512)
        }

        PagedIterable<BlobItem> pagedIterable = cc.listBlobs(new ListBlobsOptions().setMaxResultsPerPage(PAGE_SIZE), null)
        PagedResponse pagedSyncResponse1 = pagedIterable.iterableByPage().iterator().next()
        PagedIterable<BlobItem> pagedIterable2 = cc.listBlobs(new ListBlobsOptions().setMaxResultsPerPage(PAGE_SIZE), null)
        PagedIterable byPageFromToken = pagedIterable2.iterableByPage(pagedSyncResponse1.getContinuationToken())
        def pageAfterToken = byPageFromToken.iterator().next()
```

**Expected behavior**
Only two list calls should be made here, one for each of the lines that has .iterator().next(). Instead, there is a third network call coming from the line that calls iterableByPage(token), and it is identical to the request on the next line.

rickle-msft:
I've tested this further, and I am seeing that each call to next on the paged iterable is making two service calls instead of 1. It also seems that requests are being made eagerly.

rickle-msft:
It seems that the number of requests is actually correct and the apparent doubles were caused by the first iterator continuing to eagerly process the list.

alzimmermsft:
@anuchandy could you verify whether the new `PagedFlux`/`PagedIterable` enhancements have resolved this issue?

alzimmermsft:
PR #6465 should have resolved this issue.

PR #6538, opened 2019-11-25
Reduce Prefetch Limit for PagedIterable and IterableStream

alzimmermsft:
Reduces scope of #6518 

`PagedIterable` and `IterableStream` were updated to use the prefetch overload of `toIterable` and `toStream`, a prefetch of `1` is used instead of the default of `256`. This change reduces the scope of issue #6518 which saw eager retrieval of the next `256` pages or number of paged responses, whichever is less.

### DIFF
--- a/sdk/core/azure-core/src/test/java/com/azure/core/http/rest/PagedIterableTest.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/http/rest/PagedIterableTest.java
@@ -248,7 +248,7 @@ public class PagedIterableTest {
     private static class TestPagedFlux<T> extends PagedFlux<T> {
         private int nextPageRetrievals = 0;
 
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/rest/PagedIterableBase.java
```diff
@@ -46,7 +46,7 @@ public PagedIterableBase(PagedFluxBase<T, P> pagedFluxBase) {
      * @return {@link Stream} of a Response that extends {@link PagedResponse}
      */
     public Stream<P> streamByPage() {
-        return pagedFluxBase.byPage().toStream();
+        return pagedFluxBase.byPage().toStream(1);
```
srnagar: nit: define a constant instead

PR #6539, opened 2019-11-25
Use `IterableStream<T>(Iterable)` ctor in EventHubProperties

### LINKED ISSUE #6515: Cannot use IterableStream with Schedulers.parallel() or Schedulers.single()

conniey:
* Updates EventHubProperties to use `IterableStream<T>(Iterable<T>)` overload. This fixes the IllegalStateException because we are no longer moving from the Flux universe, into synchronous then back into the Flux world.
* Adds back a package-private scheduler override on EventHubClientBuilder and forces our integration tests to use Schedulers.parallel, so we can double check that we don't have blocking calls in our code.

Fixes #6515

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubPropertiesTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubPropertiesTest.java
@@ -36,22 +36,41 @@ public class EventHubPropertiesTest {
     }
 
     /**
-     * Verifies that the {@link EventHubProperties#getPartitionIds()} array is not {@code null} when we pass {@code null}
-     * to the constructor.
+     * Throws when we try to set null partitionIds.
      */
     @Test
-    public void setsPropertiesNoPartitions() {
+    public void requiresPartitions() {
         // Arrange
         final String name = "Some-event-hub-name";
         final Instant instant = Instant.ofEpochSecond(145620);
 
-        // Act
-        final EventHubProperties eventHubProperties = new EventHubProperties(name, instant, null);
-
-        // Assert
-        Assertions.assertEquals(name, eventHubProperties.getName());
-        Assertions.assertEquals(instant, eventHubProperties.getCreatedAt());
-        Assertions.assertNotNull(eventHubProperties.getPartitionIds());
-        Assertions.assertEquals(0, eventHubProperties.getPartitionIds().stream().count());
+        // Act & Assert
+        Assertions.assertThrows(NullPointerException.class, () -> new EventHubProperties(name, instant, null));
     }
+
+    /**
+     * Throws when we try to set null createdAt.
+     */
+    @Test
+    public void requiresCreatedAt() {
+        // Arrange
+        final String name = "Some-event-hub-name";
+        final String[] partitionIds = new String[]{"one-partition", "two-partition", "three-partition"};
+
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProperties.java
```diff
@@ -31,9 +31,10 @@
         final String[] partitionIds) {
         this.name = name;
         this.createdAt = createdAt;
+
         this.partitionIds = partitionIds != null
```
srnagar: partitionIds cannot be null and if it is, then should we bubble up an error?

conniey: kk. will do

PR #6544, opened 2019-11-25
AzConfig: User provided Customized Headers Policy 

mssfang:
Support Customized Headers in AzConfig. Each request can have a new header name and value. But in azconfig, we only need to support three headers:
• "x-ms-client-request-id"
• "x-ms-correlation-request-id"
• "correlation-context"

fixes: #6171

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationAsyncClientTest.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationAsyncClientTest.java
@@ -22,6 +22,7 @@ import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.test.StepVerifier;
+import reactor.util.context.Context;
 
 import java.net.HttpURLConnection;
 import java.time.Duration;
@@ -877,7 +878,7 @@ public class ConfigurationAsyncClientTest extends ConfigurationClientTestBase {
         final HttpHeaders headers = getCustomizedHeaders();
         addHeadersFromContextPolicyRunner(expected ->
             StepVerifier.create(client.addConfigurationSettingWithResponse(expected)
-                .subscriberContext(reactor.util.context.Context.of(AddHeadersFromContextPolicy.AZURE_REQUEST_HTTP_HEADERS_KEY, headers)))
+                .subscriberContext(Context.of(AddHeadersFromContextPolicy.AZURE_REQUEST_HTTP_HEADERS_KEY, headers)))
                 .assertNext(response -> {
                     final HttpHeaders requestHeaders = response.getRequest().getHeaders();
                     assertContainsHeaders(headers, requestHeaders);


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/CustomHeadersPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+     * @param context The request context
+     * @param next The next HTTP pipeline policy to process the {@code context's} request after this policy
+     *     completes.
+     * @return A {@link Mono} representing the HTTP response that will arrive asynchronously.
+     */
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        final HttpHeaders httpHeaders = customHeaderSupplier.get();
+
+        if (Objects.nonNull(httpHeaders) && httpHeaders.getSize() > 0) {
```
alzimmermsft: How is this different from the PR @hemanttanwar reverted a few days ago?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -377,6 +379,18 @@ public ConfigurationClientBuilder serviceVersion(ConfigurationServiceVersion ver
...
     }
 
+    /**
+     * Sets the customized headers that is used when each request is sent. It is possible to use new header value for
+     * each request.
+     *
+     * @param headersSupplier user's headers supplier applied to each request.
+     * @return The updated ConfigurationClientBuilder object.
+     */
+    public ConfigurationClientBuilder headersSupplier(Supplier<HttpHeaders> headersSupplier) {
```
alzimmermsft: How is this expected to be used? This is a very different API than we have seen in other builders.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/CustomHeadersPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+import com.azure.core.http.HttpPipelineCallContext;
+import com.azure.core.http.HttpPipelineNextPolicy;
+import com.azure.core.http.HttpResponse;
+import com.azure.core.http.policy.HttpPipelinePolicy;
+import reactor.core.publisher.Mono;
+
+import java.util.Objects;
+import java.util.function.Supplier;
+
+public class CustomHeadersPolicy implements HttpPipelinePolicy {
```
alzimmermsft: Nothing uses this class

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/CustomHeadersPolicy.java
mssfang: almost same. But different in policy name and accept any headers

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
mssfang: User provides supplier that we call every time for each request.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/CustomizedHeadersTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+
+    @Test
+    public void newRequestIdForEachCall() throws Exception {
+
+        String customRequestIdHeaderName = "x-ms-client-custom-request-id";
+        String clientProvidedRequestId = UUID.randomUUID().toString();
+
+        Supplier<HttpHeaders> requestIdSupplier = () ->
+            new HttpHeaders().put(customRequestIdHeaderName, clientProvidedRequestId);
```
alzimmermsft: Is this meant to generate a new UUID per request?

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/CustomizedHeadersTest.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+        @Override
+        public Mono<String> getBodyAsString() {
+            return Mono.empty();
+        }
+        @Override
+        public Mono<String> getBodyAsString(Charset charset) {
+            return Mono.empty();
+        }
+    };
```
alzimmermsft: There is a `MockHttpResponse` class that you should be able to use instead of creating this class.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/CustomizedHeadersTest.java
```diff
@@ -0,0 +1,133 @@
...
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.UUID;
+import java.util.function.Supplier;
+
+public class CustomizedHeadersTest {
+
+    private static final String REQUEST_ID_HEADER = "x-ms-client-request-id";
+    private static final String CORRELATION_ID_HEADER = "x-ms-correlation-request-id";
+    private static final String CORRELATION_CONTEXT_HEADER = "correlation-context";
```
alzimmermsft: These aren't being used.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/CustomizedHeadersTest.java
```diff
@@ -0,0 +1,133 @@
...
+                    }
+                    firstRequestId = request.getHeaders().getValue(customRequestIdHeaderName);
+                    if (firstRequestId == null) {
+                        Assertions.fail();
+                    }
+                    return Mono.just(mockResponse);
+                }
+            })
+            .policies(new CustomHeadersPolicy(requestIdSupplier),
+                new RetryPolicy(new FixedDelay(1, Duration.of(0, ChronoUnit.SECONDS))))
```
alzimmermsft: Nothing is going to be retried as you haven't setup the client to return an error.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -377,6 +379,18 @@ public ConfigurationClientBuilder serviceVersion(ConfigurationServiceVersion ver
...
+        this.headersSupplier = headersSupplier;
```
alzimmermsft: The `HttpPipelinePolicy` is configured to throw an exception if the supplier is null, if you are going to add a `CustomHeaderPolicy` into the pipeline within the builder you'll need to ensure this is not null.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/CustomizedHeadersTest.java
```diff
@@ -0,0 +1,133 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.UUID;
+import java.util.function.Supplier;
+
+public class CustomizedHeadersTest {
+
+    private static final String REQUEST_ID_HEADER = "x-ms-client-request-id";
+    private static final String CORRELATION_ID_HEADER = "x-ms-correlation-request-id";
+    private static final String CORRELATION_CONTEXT_HEADER = "correlation-context";
+
```
hemanttanwar: My earlier PR has exactly same solution in https://github.com/Azure/azure-sdk-for-java/pull/6400   

@mssfang  We talked about this today morning,  Since you need these three Header and you did not feel comfortable putting these headers in MessageIDPolicy because they are not all related to MessageID.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/CustomHeadersPolicy.java
hemanttanwar: We have gone back and forth on this 2-3 times. 
Because of three headers , AppConfig need are beyond MessageID,  it probably not appropriate to add these three headers in MessageIDPolicy  even if technically you can do it.
    "x-ms-client-request-id";
    "x-ms-correlation-request-id";
    "correlation-context";

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/CustomizedHeadersTest.java
mssfang: Is that possible to add the Supplier to the AddHeaderPolicy. It seems odd to me by supporting other customized headers in the RequestIDPolicy

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -97,6 +98,7 @@
     private HttpPipelinePolicy retryPolicy;
     private Configuration configuration;
     private ConfigurationServiceVersion version;
+    private Supplier<HttpHeaders> headersSupplier;
```
alzimmermsft: Nothing uses this private field, it is only ever set

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/CustomHeadersPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+    private final Supplier<HttpHeaders> customHeaderSupplier;
+
+    public CustomHeadersPolicy(Supplier<HttpHeaders> customHeaderSupplier) {
```
alzimmermsft: @JonathanGiles could this constructor be moved into `HttpHeadersPolicy` to allow for a dynamic header policy and a static header policy? This way we could resolve this requirement without introducing additional policies.

mssfang: Or might be AddHeadersPolicy, but not in RequestIDPolicy

mssfang:
PR will be updated. I am trying to use it to show the approach without modify current RequestIDPolicy. Adding two other headers that not related to Request ID could be supported in other Policy.  Is it possible to support overwrite headers functionality in AddHeadersPolicy by introducing Supplier to AddHeadersPolicy?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/CustomHeadersPolicy.java
JonathanGiles: I'm all for reducing the number of policies. Please do consider this.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/CustomHeadersPolicy.java
srnagar: Add javadoc. Don't see this class used anywhere. Also, this is not specific to AppConfig. Consider moving this to core. @hemanttanwar is working on supporting this feature. I would like to see how this is used as the builder is not making use of this and the supplier.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/CustomHeadersPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+     *     completes.
+     * @return A {@link Mono} representing the HTTP response that will arrive asynchronously.
+     */
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        final HttpHeaders httpHeaders = customHeaderSupplier.get();
+
+        if (Objects.nonNull(httpHeaders) && httpHeaders.getSize() > 0) {
+            for (HttpHeader header : httpHeaders) {
+                context.getHttpRequest().getHeaders().put(header.getName(), header.getValue());
```
srnagar: Can `header.getName()` and `header.getValue()` be null? Will the service accept such headers?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
srnagar: I am not sure if this is the right approach. How will this ensure thread-safety? If the same client is making several concurrent service calls, how are the headers generated and tracked by this supplier?

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -36,7 +36,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.2.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
```
alzimmermsft: @JimSuplizio This may be changed back to 1.1.0 depending on what requirements AppConfiguration has on Azure Core, but how are we planning to handle versioning as dependency versions begin to skew?

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -46,7 +46,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-http-netty</artifactId>
-      <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
+      <version>1.2.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
```
hemanttanwar: This version change will be done by updating version_client.txt an version update script. Like we had discussion , I will create a PR for this change now.

mssfang: Thanks. Will revert this version changes

mssfang:
Added tests for the new Policy

Issue #5662, opened 2019-10-02
azure-core exception types should present a more user-friendly console output


JonathanGiles:
When an exception is encountered in azure-core and returned to the user, the result is a flat error message containing useful information such as response code, and the json payload. We should format this in a more user parsable way such that they can more easily act on the error being presented.

sima-zhu:
Closed it as it finished first round of cosmetic.

JonathanGiles:
What you did and what this issue requested are different things. This issue was to ensure that the formatting of an exception, when printed to the console, is as readable as possible. Try catching some exceptions in test code and observe that they do not format nicely on the console. I've reopened this issue so that it may be correctly reviewed and resolved.

sima-zhu:
A good example to reference: github.com/Azure/azure-sdk-for-java/issues/6439

sima-zhu:
One of the exception from RetryPolicy:

```
java.lang.RuntimeException: Max retries 5 times exceeded. Error Details: some error here

	at com.azure.core.http.policy.RetryPolicy.lambda$attemptAsync$1(RetryPolicy.java:89)
	at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onError(FluxOnErrorResume.java:88)
	at reactor.core.publisher.Operators.error(Operators.java:182)
	at reactor.core.publisher.FluxFlatMap.trySubscribeScalarMap(FluxFlatMap.java:134)
	at reactor.core.publisher.MonoFlatMap.subscribeOrReturn(MonoFlatMap.java:53)
	at reactor.core.publisher.InternalMonoOperator.subscribe(InternalMonoOperator.java:48)
	at reactor.core.publisher.MonoDelaySubscription.accept(MonoDelaySubscription.java:52)
	at reactor.core.publisher.MonoDelaySubscription.accept(MonoDelaySubscription.java:33)
	at reactor.core.publisher.FluxDelaySubscription$DelaySubscriptionOtherSubscriber.onNext(FluxDelaySubscription.java:123)
	at reactor.core.publisher.MonoDelay$MonoDelayRunnable.run(MonoDelay.java:117)
	at reactor.core.scheduler.SchedulerTask.call(SchedulerTask.java:68)
	at reactor.core.scheduler.SchedulerTask.call(SchedulerTask.java:28)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #6545, opened 2019-11-26
Added some useful text into exception.

sima-zhu:
https://github.com/Azure/azure-sdk-for-java/issues/5662

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
@@ -88,7 +88,8 @@ public class HttpLoggingPolicy implements HttpPipelinePolicy {
                 requestUrl.setQuery(getAllowedQueryString(request.getUrl().getQuery()));
                 logger.info("--> {} {}", request.getHttpMethod(), requestUrl.toUrl());
             } catch (MalformedURLException ex) {
-                return Mono.error(logger.logExceptionAsWarning(new IllegalStateException("Invalid request URL.")));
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -88,7 +88,8 @@ public HttpLoggingPolicy(HttpLogOptions httpLogOptions) {
                 requestUrl.setQuery(getAllowedQueryString(request.getUrl().getQuery()));
                 logger.info("--> {} {}", request.getHttpMethod(), requestUrl.toUrl());
             } catch (MalformedURLException ex) {
-                return Mono.error(logger.logExceptionAsWarning(new IllegalStateException("Invalid request URL.")));
+                return Mono.error(logger.logExceptionAsWarning(new IllegalStateException("Invalid request URL. URL: "
+                    + request.getUrl())));
```
alzimmermsft: Please view how request sending logging logs the URL, right now this has the potential to log query parameters that should be redacted such a SAS token `sig`.

sima-zhu: Logging happens after credential policy. Usually it doesn't include sas or other auth. The loggin here especially for response. So it will not expose the secret info

sima-zhu:
Close the PR, as the exception will log secret out.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: Given that is happens after the credential policy wouldn't that mean the SAS would be applied to the request URL?

PR #6546, opened 2019-11-26
Update samples for new Event Hubs core classes

conniey:
* Updates a bit of the javadocs in main classes (like the first sentence), but there will be another PR with a closer look at docs.
* Updates all the samples for publishing, consuming, adds sample for azure-identity

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/GetEventHubMetadata.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/GetEventHubMetadata.java
@@ -21,25 +21,27 @@ public class GetEventHubMetadata {
 
         // Instantiate a client that will be used to call the service. Using a try-resource block, so it disposes of
         // the client when we are done.
-        try (EventHubProducerClient client = new EventHubClientBuilder()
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/PublishEventsWithAzureIdentity.java
```diff
@@ -0,0 +1,89 @@
...
+
+        // The sample Flux contains three events, but it could be an infinite stream of telemetry events.
+        telemetryEvents.subscribe(event -> {
+            final EventDataBatch batch = currentBatch.get();
+            if (!batch.tryAdd(event)) {
+                producer.createBatch().map(newBatch -> {
+                    currentBatch.set(newBatch);
+
+                    // Adding that event we couldn't add before.
+                    newBatch.tryAdd(event);
```
srnagar: What happens if this `tryAdd` returns `false` i.e. if an event is too big for any batch? We have no logging or error to report it back to the user. We silently drop the event.

conniey: Yeah. I should propagate an error here.

PR #6554, opened 2019-11-26
Upgrade netty depedencies to latest version

srnagar:
Upgrading some of the internal netty dependencies used by reactor-netty to the latest version.

### DIFF
--- a/sdk/core/azure-core-http-netty/pom.xml
+++ b/sdk/core/azure-core-http-netty/pom.xml
@@ -79,6 +79,23 @@
     </dependency>
 
     <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-codec-http2</artifactId>
+      <version>4.1.42.Final</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-transport-native-unix-common</artifactId>
+      <version>4.1.42.Final</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-transport-native-epoll</artifactId>
+      <version>4.1.42.Final</version>
+      <classifier>linux-x86_64</classifier>
+    </dependency>
+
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core-http-netty/pom.xml
```diff
@@ -78,6 +78,23 @@
...
 
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-codec-http2</artifactId>
+      <version>4.1.42.Final</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-transport-native-unix-common</artifactId>
+      <version>4.1.42.Final</version>
```
anuchandy: LGTM, one question, so we no longer inherits version from the parent?

srnagar: @JimSuplizio is working on removing parent pom completely in this PR - https://github.com/Azure/azure-sdk-for-java/pull/6434

So, all versions will be explicitly added to every dependency in every pom file.

hemanttanwar: https://mvnrepository.com/artifact/io.netty/netty-transport-native-unix-common/4.1.43.Final
Could we go to more latest version ?

hemanttanwar: 4.1.43.Final ? Can we use this

srnagar: We are using 4.1.42.Final for all netty dependencies. So, I used the same version for these two as well. I will look into 4.1.43.Final for the next release.

JonathanGiles: Sharing my general opinion on dependencies: we want to be fast followers rather than jumping to the newest version immediately. If we can understand where the community is (and I have tooling that can help here), we can move in step with the pack to try to minimise dependency conflicts.

PR #6560, opened 2019-11-26
Update changelog for upcoming beta release

srnagar:


sdk/eventhubs/azure-messaging-eventhubs/CHANGELOG.md
```diff
@@ -1,4 +1,17 @@
 # Release History
+## 5.0.0-beta.6 (2019-12-02)
+- Artifact name changed from `preview` to `beta`.
+- Producer clients (both sync and async) support sending events only using `EventDataBatch`. All other send overloads are removed.
```
conniey: * We should say that we removed EventHubClient and EventHubAsyncClient and exposed the Consumer*Client and Producer*Clients.

conniey: BatchOptions is now called CreateBatchOptions.

conniey: Consuming events returns PartitionEvent containing PartitionContext and the EventData rather than EventData.

PR #6571, opened 2019-11-26
Event Hubs doc updates

conniey:
Fixing documentation and samples in Event Hubs.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClientJavaDocCodeSamples.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClientJavaDocCodeSamples.java
@@ -4,6 +4,7 @@
 package com.azure.messaging.eventhubs;
 
 import com.azure.messaging.eventhubs.models.CreateBatchOptions;
+import reactor.core.Exceptions;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
@@ -52,7 +53,10 @@ public class EventHubProducerAsyncClientJavaDocCodeSamples {
             batch.tryAdd(new EventData("test-event-1"));
             batch.tryAdd(new EventData("test-event-2"));
             return producer.send(batch);
-        }).block();
+        }).subscribe(unused -> {
+        },
+            error -> System.err.println("Error occurred while sending batch:" + error),
+            () -> System.out.println("Send complete."));
         // END: com.azure.messaging.eventhubs.eventhubasyncproducerclient.createBatch
 
         producer.close();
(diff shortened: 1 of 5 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/ReceiveOptions.java
```diff
@@ -30,20 +30,18 @@ public Long getOwnerLevel() {
     }
 
     /**
-     * Sets the {@code ownerLevel} value on this consumer. When populated, the level indicates that a consumer is
-     * intended to be the only reader of events for the requested partition and an associated consumer group. To do so,
-     * this consumer will attempt to assert ownership over the partition; in the case where more than one exclusive
-     * consumer attempts to assert ownership for the same partition/consumer group pair, the one having a larger {@link
-     * ReceiveOptions#getOwnerLevel()} value will "win".
+     * Sets the {@code ownerLevel} value on this recieve operation. When populated, the level indicates that the receive
```
srnagar:
```suggestion
     * Sets the {@code ownerLevel} value on this receive operation. When populated, the level indicates that the receive
```

sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/EventHubConsumerAsyncClientJavaDocCodeSamples.java
```diff
@@ -18,6 +18,10 @@
  * Code snippets demonstrating various {@link EventHubConsumerAsyncClient} scenarios.
  */
 public class EventHubConsumerAsyncClientJavaDocCodeSamples {
+    private final EventHubConsumerAsyncClient consumer = new EventHubClientBuilder()
+        .connectionString("fake-string")
+        .consumerGroup(EventHubClientBuilder.DEFAULT_CONSUMER_GROUP_NAME)
```
srnagar: We should just have  `"consumer-group"` here since the service team doesn't want to promote the use of `DEFAULT_CONSUMER_GROUP_NAME`.

sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClientJavaDocCodeSamples.java
```diff
@@ -37,84 +35,108 @@ public void instantiate() {
...
+                "Endpoint={fully-qualified-namespace};SharedAccessKeyName={policy-name};SharedAccessKey={key}",
+                "event-hub-name")
+            .buildAsyncProducerClient();
+
+        // Creating a batch without options set, will allow for automatic routing of events to any partition.
+        producer.createBatch().flatMap(batch -> {
+            batch.tryAdd(new EventData("test-event-1"));
+            batch.tryAdd(new EventData("test-event-2"));
+            return producer.send(batch);
+        }).block();
```
srnagar: Should our samples have `block()` or `subscribe()` for async clients?

conniey: Hmm... The bad thing is if I use .subscribe, since it's a non-blocking call, we'd have to add some sort of "Thread.sleep()" is that what we want instead?

srnagar: Yeah, it's tricky. Users usually copy samples and they'll block which defeats the purpose of async. If you add sleep, it's not ideal but at least users will realize they want to do something else here instead of sleep. Hopefully, adding a comment before sleep will help them understand what is expected.

PR #6576, opened 2019-11-27
AutoRest: fix cmd to use autorest.java with preview version

xseeseesee:


weidongxu-microsoft:
Please check with this argument. I think "autorest-java" and the new "use" is mutually exclusive and you need to verify that.

```
    const generatorPath = args['autorest-java']
        ? `--use=${path.resolve(args['autorest-java'])} `
        : '';
```

gulpfile.js
```diff
@@ -131,7 +137,7 @@ var codegen = function(project, cb) {
                         ' --java ' +
                         ' --azure-arm ' +
                         ' --azure-libraries-for-java-folder=' + process.cwd() + ' ' +
-                        ` --license-header=MICROSOFT_MIT_NO_CODEGEN ` +
+                        ` --license-header=MICROSOFT_MIT_NO_CODEGEN ` +
```
weidongxu-microsoft: better to clean up a bit here.

PR #6587, opened 2019-11-29
Enable dev package publishing for Java repo.

mitchdenny:
This PR adds the logic necessary to publish daily dev builds to a public Azure Artifacts feed. There are a couple of changes (high level summary):

1. Update ```set_versions.py``` to support taking an artifact id.
2. Update ```set_versions.py``` to correctly append build number against GA versions.
3. Update build to call ```set_versions.py``` for each published artifact.

This PR depends on an additional template in ```azure-sdk-tools``` which is currently in PR. That will need to be approved first and then this PR can be updated to point to the master ref of that repository.

https://github.com/Azure/azure-sdk-tools/pull/301

On the release side there is also a PR updating our release process on our internal repository.

### DIFF
--- a/sdk/core/ci.yml
+++ b/sdk/core/ci.yml
@@ -6,7 +6,6 @@ resources:
     - repository: azure-sdk-build-tools
       type: git
       name: internal/azure-sdk-build-tools
-      ref: refs/heads/feature/dev-package-builds
     - repository: azure-sdk-tools
       type: github
       name: Azure/azure-sdk-tools


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -55,6 +55,21 @@ jobs:
...
           versionSpec: '3.6'
 
+      - template: tools/daily-dev-build-variable/daily-dev-build-variable.yml@azure-sdk-tools
+
+      - ${{ each artifact in parameters.Artifacts }}:
+        - script: |
+            python3 --version 
+            python3 eng/versioning/set_versions.py --update-type all --build-type client --build-qualifier dev.$(Build.BuildNumber) --artifact-id ${{artifact.name}}
+          condition: eq(variables['SetDevVersion'],'true')
+          displayName: Apped dev package version suffix for ${{artifact.name}}
```
weshaggard: "Apped" -> "Append" .

Any reason to not do both of these steps in the same script block?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -73,7 +88,7 @@ jobs:
         inputs:
           mavenPomFile: sdk/${{parameters.ServiceDirectory}}/pom.service.xml
           goals: 'package'
-          options: '$(DefaultOptions) "-DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -DskipTests -Dinclude-template -Dgenerate-overview' # We include template-module so we ensure it always builds in CI
```
weshaggard: Did you intend to remove the "include-template" as part of this PR?

mitchdenny: Yeah ... I think it is redundant now that we have a separate pipeline for it (which we frequently use for testing).

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mitchdenny: The first script block is executed multiple times, once per artifact specified in ci.yml, the second block is executed once. Will fix the typo.

mitchdenny:
> The changes look good @mitchdenny, thank you! All in all the changes to set_versions were pretty minimal, not bad considering it was a best guess at the time versions were centralized.

Yeah ... actually I thought editing those scripts would be harder than it was. Was a very easy base to work from.

PR #6593, opened 2019-12-01
Add workaround for Checkstyle error, closes #6216

### LINKED ISSUE #6216: Update Contributing.md build environment instructions
Tracking issue to improve the contributing.md

> The `sdk-build-tools` doesn't get released to Maven, to install it you'll need to have the [eng](https://github.com/Azure/azure-sdk-for-java/tree/master/eng) folder copied locally and install `eng/code-quality-reports/pom.xml`. This should unblock you on this issue, please let us know if you run into additional issues and we'll be happy to help.
> 
> _Originally posted by @alzimmermsft in https://github.com/Azure/azure-sdk-for-java/issues/6181#issuecomment-551218225_

karianna:


JonathanGiles:
There are two separate issues at play here:

1) The maven-java-parent project has a DTD reference that has moved since it was first created. This means that validating the XML now fails. We had to do a similar change in azure-sdk-for-java too. maven-java-parent refers to http://www.puppycrawl.com/dtds/suppressions_1_0.dtd, which is not a valid URL, whereas azure-sdk-for-java refers to https://checkstyle.org/dtds/suppressions_1_2.dtd, which is valid. This can be fixed by editing the maven-java-parent file [here](https://github.com/microsoft/maven-java-parent/blob/master/build-tools/src/main/resources/checkstyle/checkstyle-suppressions.xml) to have the updated dtd url referenced.

2) There is, at present, a need to install the code quality check tooling into a local m2 cache before the build will run. In the relatively near term, azure-sdk-for-java will be getting updated to hopefully make this more automatic and less painful for users by @alzimmermsft.

Given this, I would suggest we don't merge this PR right now, or that we edit the text to cover the requirements of installing the code quality reports, without including the dtd issue that is separate.

joshfree:
@sima-zhu since you worked on https://github.com/Azure/azure-sdk-for-java/issues/8110, can you please review this PR as part of getting contributing.md into a current state (no pits of failures for new contributors)?

joshfree:
Followed up with @alzimmermsft offline

PR #6596, opened 2019-12-02
Directory per artifact

mitchdenny:
This PR changes the way that we structure the build outputs so that each artifact (maven) (e.g. ```azure-core```) lives within its own directory within the ```packages``` artifact (pipelines).

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -1126,7 +1126,7 @@
         </property>
       </activation>
       <properties>
-        <packageOutputDirectory>${project.build.directory}</packageOutputDirectory>
+        <packageOutputDirectory>${artifact.staging.directory}/${project.artifactId}</packageOutputDirectory>
       </properties>
     </profile>
 
--- a/pom.data.xml
+++ b/pom.data.xml
@@ -609,7 +609,7 @@
         </property>
       </activation>
       <properties>
-        <packageOutputDirectory>${project.build.directory}</packageOutputDirectory>
+        <packageOutputDirectory>${artifact.staging.directory}/${project.artifactId}</packageOutputDirectory>
       </properties>
     </profile>
 


pom.client.xml
```diff
@@ -94,7 +94,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <packageOutputDirectory>${project.build.directory}</packageOutputDirectory>
+    <packageOutputDirectory>${artifactStagingDirectory}/${project.artifactId}</packageOutputDirectory>
```
weshaggard: Don't we need to default "artifactStagingDirectory"? If we don't then I expect local builds might fail.

mitchdenny: Yeah good idea. At this point I'm just experimenting. I'm not certain this is the best approach. It might be better, and more generic to have a reusable PowerShell fragment that groups by regex match and then reuse it across all languages. Thoughts?

The other idea I am toying with is having a seperate artifact per file group.

weshaggard: I'd like to eliminate the regex matching if we can it is just a source of errors. It would certainly better to do have that only in this one place as opposed to the way we have it now but I think we can do better. I believe overriding the output path from the individual package build steps for the given language is better. That way the package output from the real build is the one ensuring we get the correct set of files into the correct output directory instead of a magic regex.

mitchdenny: That's fair, in that case this is probably the right approach then. So I'll add a default property value here that we can then override to repoint things, and people can of course still override the packageOuputDirectory value if they want to squash them together.

Obviously this is just the first in a bunch of changes that would need to be made to have this work across the ES.

mitchdenny: I've adjusted my approach a bit. In my latest commit I've added a profile to ```pom.client.xml``` to enable partitioning of the artifacts. My intention here is that we'll only do this as part of the Azure Pipelines builds.

I will probably need to do the same thing to pom.data.xml as well. After my POM restructuring changes I can probably just move it into pom.xml and it'll cover both scenarios.

weshaggard: Ok if you do it with the profile activation then you should revert back to the old packageOutputDirectory property value by default then correct?

mitchdenny: Reverted

mitchdenny:
@weshaggard my plan with this one is to merge it onto ```feature/artifact-partitioning``` so that I can keep testing/adjusting the later stages of the pipeline. I'll do that now and then I'll start another PR once things are stabilized there.

weshaggard:
Unless I'm missing something I see you didn't revert the packageOutputDirectory property https://github.com/Azure/azure-sdk-for-java/blob/feature/artifact-partitioning/pom.client.xml#L97.

mitchdenny:
> Unless I'm missing something I see you didn't revert the packageOutputDirectory property https://github.com/Azure/azure-sdk-for-java/blob/feature/artifact-partitioning/pom.client.xml#L97.

My bad ... I've fixed it up on the branch. I fixed up ```pom.data.xml``` but not ```pom.client.xml```. See the ```feature/artifact-partitioning```.

Issue #6542, opened 2019-11-25
[Bug] Account SAS does not work on Object level client auth


sima-zhu:
**Summary:**
When we use endpoint with SAS token to build the client, there is an exception thrown indicating authentication failure. 

**Reproduce:**
Run FileSamples, ShareSamples etc,

**Exceptions:**
```
Failed to create snapshot on share. Reasons: Status code 403, "﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:62117598-401a-004e-3ddf-a30a77000000
Time:2019-11-25T22:26:27.2670708Z</Message><AuthenticationErrorDetail>sr is mandatory. Cannot be empty</AuthenticationErrorDetail></Error>"
```

**Expect:**
No exception thrown and add test coverage on this.

PR #6601, opened 2019-12-02
Fixed issue where account SAS does not work on clients

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/6542

### DIFF
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/sas/CommonSasQueryParameters.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/sas/CommonSasQueryParameters.java
@@ -127,4 +127,112 @@ public class CommonSasQueryParameters extends BaseSasQueryParameters {
 
         return sb.toString();
     }
+
+    /**
+     * @return The signed identifier. Please see <a href="https://docs.microsoft.com/en-us/rest/api/storageservices/establishing-a-stored-access-policy">here</a>
+     * for more information.
+     */
+    public String getIdentifier() {
+        return identifier;
+    }
+
+    /**
+     * @return The storage resource.
+     */
+    public String getResource() {
+        return resource;
+    }
+
+    /**
+     * @return The Cache-Control header value when a client accesses the resource with this sas token.
+     */
+    public String getCacheControl() {
+        return cacheControl;
+    }
+
+    /**
+     * @return The Content-Disposition header value when a client accesses the resource with this sas token.
+     */
+    public String getContentDisposition() {
+        return contentDisposition;
+    }
+
+    /**
+     * @return The Content-Encoding header value when a client accesses the resource with this sas token.
+     */
+    public String getContentEncoding() {
+        return contentEncoding;
+    }
+
+    /**
+     * @return The Content-Language header value when a client accesses the resource with this sas token.
+     */
+    public String getContentLanguage() {
+        return contentLanguage;
+    }
+
+    /**
+     * @return The Content-Type header value when a client accesses the resource with this sas token.
+     */
+    public String getContentType() {
+        return contentType;
+    }
+
+    /**
+     * @return the object ID of the key.
+     */
+    public String getKeyObjectId() {
+        return keyObjectId;
+    }
+
+    /**
+     * @return the tenant ID of the key.
+     */
+    public String getKeyTenantId() {
+        return keyTenantId;
+    }
+
+    /**
+     * @return the datetime when the key becomes active.
+     */
+    public OffsetDateTime getKeyStart() {
+        return keyStart;
+    }
+
+    /**
+     * @return the datetime when the key expires.
+     */
+    public OffsetDateTime getKeyExpiry() {
+        return keyExpiry;
+    }
+
+    /**
+     * @return the services that are permitted by the key.
+     */
+    public String getKeyService() {
+        return keyService;
+    }
+
+    /**
+     * @return the service version that created the key.
+     */
+    public String getKeyVersion() {
+        return keyVersion;
+    }
+
+    /**
+     * @return The storage services being accessed (only for Account SAS). Please refer to {@link AccountSasService} for
+     * more details.
+     */
+    public String getServices() {
+        return services;
+    }
+
+    /**
+     * @return The storage resource types being accessed (only for Account SAS). Please refer to {@link
+     * AccountSasResourceType} for more details.
+     */
+    public String getResourceTypes() {
+        return resourceTypes;
+    }
 }


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/CommonSasQueryParameters.java
```diff
@@ -0,0 +1,123 @@
+package com.azure.storage.common.implementation;
```
alzimmermsft: Missing licensing header.

This class will need to be outside of an implementation package given that it is used in public API, starting with JDK9+ consumers of the SDK won't be able to interact with it directly.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
```diff
@@ -6,6 +6,9 @@
 import com.azure.storage.common.sas.SasProtocol;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
```
alzimmermsft: Unused imports? Not seeing any code changes other than this in the class but that could be from return type changes?

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareClientBuilder.java
```diff
@@ -189,8 +190,9 @@ public ShareClientBuilder endpoint(String endpoint) {
             }
             this.shareName = length >= 2 ? pathSegments[1] : this.shareName;
 
+            // TODO : What happens if a user has custom queries?
```
alzimmermsft: This is a good question that has gone unanswered for now, @rickle-msft thoughts? These could just be thrown back into the query string when the SAS token is generated, that way we are able to clean the SAS token using the SAS query parameters classes but don't lose custom query parameters unrelated to SAS tokens.

sdk/storage/tests.yml
```diff
@@ -24,5 +24,9 @@ jobs:
         AZURE_STORAGE_QUEUE_CONNECTION_STRING: $(java-storage-test-queue-connection-string)
         AZURE_STORAGE_BLOB_CONNECTION_STRING: $(java-storage-test-blob-connection-string)
         STORAGE_DATA_LAKE_ACCOUNT_NAME: $(java-storage-data-lake-account-name)
-        STORAGE_DATA_LAKE_ACCOUNT_KEY: $(java-storage-data-lake-account-key)
+        STORAGE_DATA_LAKE_ACCOUNT_KEY: $(java-storage-data-lake-account-key)[3:00 PM] Daniel Jurek
+        AZURE_STORAGE_FILE_ACCOUNT_NAME: $(java-storage-account-name)
+        AZURE_STORAGE_FILE_ACCOUNT_KEY: $(java-storage-account-key)
+        AZURE_STORAGE_QUEUE_ACCOUNT_NAME: $(java-storage-account-name)
+        AZURE_STORAGE_QUEUE_ACCOUNT_KEY: $(java-storage-account-key)
```
alzimmermsft: Did the changes which added these remove any of the older testing environment values such as File and Queue connection string and endpoint?

Good change to make Files and Queues more common with Blobs in terms of testing configurations and setup.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
```diff
@@ -166,24 +167,24 @@ public BlobUrlParts setSnapshot(String snapshot) {
...
 
     /**
-     * Gets the {@link BlobServiceSasQueryParameters} representing the SAS query parameters that will be used to
+     * Gets the {@link CommonSasQueryParameters} representing the SAS query parameters that will be used to
      * generate the SAS token for this URL.
      *
-     * @return the {@link BlobServiceSasQueryParameters} of the URL
+     * @return the {@link CommonSasQueryParameters} of the URL
      */
-    public BlobServiceSasQueryParameters getSasQueryParameters() {
```
alzimmermsft: Was this API in existence before 12.0.0 or 12.0.1? If so we will want to mark this one as deprecated and add the `CommonSasQueryParameters` variants as new APIs.

alzimmermsft: If this is needed to be done creating an internal method to convert from `BlobServiceSasQueryParameters` to `CommonSasQueryParameters` and vice-versa will make the internal implementation simpler.

gapra-msft: I dont think it would be possible to convert between the two since CommonSASQueryParameters is a super set of the BlobServiceSasQueryParameters (if its an account sas being used the blob service sas would be invalid) maybe we can discuss the right way to deal with this.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/CommonSasQueryParameters.java
gapra-msft: Just added the header

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/Constants.java
gapra-msft: Fixed

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareClientBuilder.java
gapra-msft: Created an issue here https://github.com/Azure/azure-sdk-for-java/issues/6604

sdk/storage/tests.yml
gapra-msft: nope, I havent removed them for now since I didnt want to scope creep, I'll create an issue for this.

gapra-msft: https://github.com/Azure/azure-sdk-for-java/issues/6605

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
gapra-msft: I guess at the very least we can pretend like its a BlobServiceSasQueryParameter that way we can mark it as deprecated and remove it in the future

gapra-msft: @alzimmermsft Can you check to see if this is better? I brought back the old methods and deprecated them.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
```diff
@@ -239,8 +266,8 @@ public URL toUrl() {
         if (this.snapshot != null) {
             url.setQueryParameter(Constants.UrlConstants.SNAPSHOT_QUERY_PARAMETER, this.snapshot);
         }
-        if (this.blobServiceSasQueryParameters != null) {
```
alzimmermsft: Does anything use the `blobServiceSasQueryParameters` anymore? If someone upgraded to a new version of the Blobs SDK without updating to using `commonSasQueryParameters` would this continue to function?

Should this become an if/else block where `CommonSasQueryParameters` takes precedence over the Blob service variant? Such as:

```java
String encodedSas;
if (this.commonSasQueryParameters != null) {
    encodedSas = this.commonSasQueryParameters.encode();
} else if (this.blobServiceQueryParameters != null) {
    encodedSas = this.blobServiceQueryParameters.encode();
}

if (!CoreUtils.isNullOrEmpty(encodedSas)) {
    url.setQuery(encodedSas);
}
```

gapra-msft: With the other blob SAS PR the idea is that all the ServiceSasQueryParams classes are going away (once this PR merges in I can go in and deprecate the whole class essentially). Using BlobServiceSasSignatureValues will only work with Service SASs so it would technically only work in certain cases. 

I think the if/else block would just make it redundant since CommonSasQueryParameters is just a super set of BlobServiceSasQueryParameters

alzimmermsft: Agreed with the change, my concern is the current production code could potentially begin failing upon upgrade without continuing to support `BlobServiceSasQueryParameters` to generate the query string. If I had the following code upon upgrade I could get a warning about using deprecated code but it would continue to compile, I would no longer have a URL which contains SAS query parameters given `CommonSasQueryParameters` is the only thing checked.

If we make sure to explicitly call out that they are still accepted in the API but won't trigger functionality I'm more okay with it.

```java
URL url = new BlobUrlParts()
  .setScheme("https://")
  .setHost("account.blob.core.windows.net")
  .setSasQueryParameters(blobServiceSasQueryParameters)
  .toUrl();
```

PR #6602, opened 2019-12-02
Add Custom Header - In AddHeaderPolicy and RequestIDPolicy 

hemanttanwar:
Search team require : Able to add custom request id header with a value.
App config require: To be able to add three custom header and provide values:
   These headers are  for message id, correlation id and context.

Shawn's Related PR https://github.com/Azure/azure-sdk-for-java/pull/6544/files This change will reduce need to create a Custom class in AppConfig  and he can use AddHeaderPolicy instead.

More discussion https://github.com/Azure/azure-sdk-for-java/issues/6217#issuecomment-557320672

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientTest.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientTest.java
@@ -7,14 +7,8 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import com.azure.core.exception.HttpResponseException;
 import com.azure.core.exception.ResourceExistsException;
-import com.azure.core.http.HttpHeaders;
-import com.azure.core.http.HttpPipeline;
-import com.azure.core.http.HttpPipelineBuilder;
 import com.azure.core.http.netty.NettyAsyncHttpClientBuilder;
-import com.azure.core.http.policy.AddHeadersFromContextPolicy;
 import com.azure.core.http.policy.HttpLogOptions;
-import com.azure.core.http.policy.HttpLoggingPolicy;
-import com.azure.core.http.policy.HttpPipelinePolicy;
 import com.azure.data.appconfiguration.models.ConfigurationSetting;
 import com.azure.data.appconfiguration.models.SettingFields;
 import com.azure.data.appconfiguration.models.SettingSelector;
@@ -63,22 +57,7 @@ public class ConfigurationClientTest extends ConfigurationClientTestBase {
                 .buildClient());
         }
     }
-    private void test(){
-        HttpPipeline pipeline = new HttpPipelineBuilder()
-            .policies(new AddHeadersFromContextPolicy())
-            .build();
-        ConfigurationClient configurationClient = new ConfigurationClientBuilder()
-            .pipeline(pipeline)
-            .endpoint("https://myconfig.azure.net/")
-            .connectionString("<Your connection String>")
-            .buildClient();
-        // Add your headers
-        final HttpHeaders headers = new HttpHeaders();
-        headers.put("my-header1", "my-header1-value");
-        headers.put("my-header2", "my-header2-value");
-        headers.put("my-header3", "my-header3-value");
-        configurationClient.addConfigurationSettingWithResponse(  new ConfigurationSetting().setKey("key").setValue("value"), new Context(AddHeadersFromContextPolicy.AZURE_REQUEST_HTTP_HEADERS_KEY, headers));
-    }
+
     @Override
     protected void afterTest() {
         logger.info("Cleaning up created key values.");


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,53 @@
...
 import java.util.UUID;
+import java.util.function.Supplier;
 
 /**
  * The pipeline policy that puts a UUID in the request header. Azure uses the request id as
  * the unique identifier for the request.
  */
 public class RequestIdPolicy implements HttpPipelinePolicy {
     private static final String REQUEST_ID_HEADER = "x-ms-client-request-id";
+    private final Supplier<HttpHeader> requestIdSupplier;
```
alzimmermsft: Would this be better off being a `Supplier<String>` where the string is used to set the header values that this policy will set?

hemanttanwar: Search Team and App Config team both need to pass in header name and header value both. Thus I am using HttpHeader.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,22 +3,53 @@
...
+
+
+    /**
+     * Creates a RequestIdPolicy provided {@link Supplier}.
+     *
+     * @param headerSupplier The {@link Supplier} to provide header to add to outgoing requests.
+     */
+    public RequestIdPolicy(Supplier<HttpHeader> headerSupplier) {
```
anuchandy: should we make it clear in the java-doc about assumptions?

By default RequestIdPolicy sets `x-ms-client-request-id` header if it is not already set. A different header can be treated as requestId header by returning a `non-null` Header from `requestIdSupplier`. Header returned by the `requestIdSupplier` will be ignored if it is `null` or there exists a header with same name, if `null` then `x-ms-client-request-id` will be set.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AddHeadersPolicy.java
```diff
@@ -10,23 +10,36 @@
...
 
+import java.util.Objects;
+import java.util.function.Supplier;
+
 /**
  * The pipeline policy that adds a particular set of headers to HTTP requests.
  */
 public class AddHeadersPolicy implements HttpPipelinePolicy {
-    private final HttpHeaders headers;
+    private final Supplier<HttpHeaders> headersSupplier;
```
srnagar: As discussed, using suppliers may not be thread-safe here. So, consider looking into using `Context`.

anuchandy: May be we should clarify this in our pipeline guidelines: An instance of a policy is shared across multiple calls, hence concurrently called by the http-client-threads sharing the same pipeline. This implicitly means, any business logic externalized from the policy, through Function, Supplier, are not thread-safe as well. Any state management should be happen through HttpPipelineCallContext object.

hemanttanwar: Will add POC for this approach in same PR.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -3,56 +3,46 @@
...
     /**
-     * Creates a RequestIdPolicy provided {@link Supplier}.
-     *
-     * @param headerSupplier The {@link Supplier} to provide header to add to outgoing requests.
+     * Creates  {@link RequestIdPolicy} with provided {@code requestIdHeaderName}.
+     * @param requestIdHeaderName to be used to set in {@link HttpRequest}.
      */
-    public RequestIdPolicy(Supplier<HttpHeader> headerSupplier) {
-        this.requestIdSupplier = Objects.requireNonNull(headerSupplier, "headerSupplier not be null.");
+    public RequestIdPolicy(String requestIdHeaderName) {
```
hemanttanwar: The search team only need to override header name. This constructor will satisfy their ask.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/OverrideHeaderPolicy.java
```diff
@@ -0,0 +1,33 @@
...
+import static com.azure.core.util.Context.AZURE_REQUEST_HTTP_HEADERS_KEY;
+import reactor.core.publisher.Mono;
+
+import java.util.Objects;
+
+/**
+ * The pipeline policy that override or add  {@link HttpHeader} in {@link HttpRequest} by reading values from
+ * {@link Context} with key  'azure-request-override-http-headers-key'.
+ */
+public class OverrideHeaderPolicy implements HttpPipelinePolicy {
```
hemanttanwar: This is to satisfy the ask by App Config to be able to provide values for following http headers
• "x-ms-client-request-id"
• "x-ms-correlation-request-id"
• "correlation-context"

End user need to add `HttpHeaders`  in Context object with a specific key to achieve this.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/OverrideHeaderPolicy.java
```diff
@@ -0,0 +1,39 @@
+package com.azure.core.http.policy;
```
alzimmermsft: Missing licensing information

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/OverrideHeaderPolicy.java
```diff
@@ -0,0 +1,39 @@
...
+import com.azure.core.util.Context;
+import com.azure.core.http.HttpRequest;
+import static com.azure.core.util.Context.AZURE_REQUEST_HTTP_HEADERS_KEY;
+import reactor.core.publisher.Mono;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * The pipeline policy that override or add  {@link HttpHeader} in {@link HttpRequest} by reading values from
```
alzimmermsft:
```suggestion
 * The pipeline policy that override or add {@link HttpHeaders} in {@link HttpRequest} by reading values from
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/OverrideHeaderPolicy.java
```diff
@@ -0,0 +1,39 @@
...
+ * {@link Context} with key  'azure-request-override-http-headers-key'. The value for this key should be of type
```
alzimmermsft:
```suggestion
 * {@link Context} with key 'azure-request-override-http-headers-key'. The value for this key should be of type
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/OverrideHeaderPolicy.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+
+        Optional<Object> customHttpHeadersObject = context.getData(AZURE_REQUEST_HTTP_HEADERS_KEY);
+        if (!Objects.isNull(customHttpHeadersObject) && customHttpHeadersObject.get() instanceof HttpHeaders) {
+            HttpHeaders customHttpHeaders = (HttpHeaders) customHttpHeadersObject.get();
+            // loop through customHttpHeaders and add header in HttpRequest
+            for (HttpHeader httpHeader : customHttpHeaders) {
+                if (!Objects.isNull(httpHeader.getName()) && !Objects.isNull(httpHeader.getValue())) {
+                    context.getHttpRequest().getHeaders().put(httpHeader.getName(), httpHeader.getValue());
```
alzimmermsft: Should we add a put overload into `HttpHeaders` which accepts a `HttpHeader`?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/OverrideHeaderPolicy.java
anuchandy: Can we have the policy to name as `AddHeadersFromContextPolicy` or something that indicates that headers are coming from the Context?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/OverrideHeaderPolicy.java
```diff
@@ -0,0 +1,39 @@
...
+ * {@link Context} with key  'azure-request-override-http-headers-key'. The value for this key should be of type
+ * {@link HttpHeaders} for it to be added in {@link HttpRequest}.
+ */
+public class OverrideHeaderPolicy implements HttpPipelinePolicy {
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+
+        Optional<Object> customHttpHeadersObject = context.getData(AZURE_REQUEST_HTTP_HEADERS_KEY);
+        if (!Objects.isNull(customHttpHeadersObject) && customHttpHeadersObject.get() instanceof HttpHeaders) {
```
anuchandy: Context will return an `empty` Optional if key does not exists, so we can check for the presence and if so check for the type, I think no need to check for nullability as `null instanceof HttpHeaders ` evaluated to false

```java
if (customHttpHeadersObject.isPresent() 
    && customHttpHeadersObject.get() instanceof HttpHeaders) {
```

sdk/core/azure-core/src/main/java/com/azure/core/util/Context.java
```diff
@@ -23,6 +24,9 @@
 public class Context {
     private final ClientLogger logger = new ClientLogger(Context.class);
 
+    /**Key used to override headers in HttpRequest. The Value for this key should be {@link HttpHeaders}.*/
+    public static String AZURE_REQUEST_HTTP_HEADERS_KEY = "azure-request-override-http-headers-key";
```
alzimmermsft: Does this constant make more sense in `Context` or the policy which uses it? Recently I made changes to `UserAgentPolicy` to look for `Context` values and I put the constants in the policy itself.

@srnagar @JonathanGiles Thoughts on location?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -6,22 +6,43 @@
...
+import com.azure.core.http.HttpHeader;
 import reactor.core.publisher.Mono;
 
+import java.util.Objects;
 import java.util.UUID;
 
 /**
- * The pipeline policy that puts a UUID in the request header. Azure uses the request id as
- * the unique identifier for the request.
+ * The pipeline policy that puts a UUID in the request header. The default {@link HttpHeader} name can be overwritten.
```
alzimmermsft: I would leave the original Javadoc as is and add a new paragraph to explain how the header name could be overwritten.

sdk/core/azure-core/src/main/java/com/azure/core/util/Context.java
hemanttanwar: I thought from end user point of view, Since in public API we are asking end user to add context.addData(key , value ) , Use will find the Key in Context class. 
But I can change if  everyone suggest another location.

alzimmermsft: I think Context works for now, if anything better can be thought of we can circle back on this.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AddHeadersFromContextPolicy.java
```diff
@@ -0,0 +1,44 @@
...
+import com.azure.core.http.HttpRequest;
+import static com.azure.core.util.Context.AZURE_REQUEST_HTTP_HEADERS_KEY;
+import reactor.core.publisher.Mono;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * The pipeline policy that override or add  {@link HttpHeaders} in {@link HttpRequest} by reading values from
+ * {@link Context} with key 'azure-request-override-http-headers-key'. The value for this key should be of type
```
srnagar: 'azure-request-override-http-headers-key' - can this just be `azure-http-headers-key`? I don't think all user-provided headers override existing headers. There may be new headers too.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AddHeadersFromContextPolicy.java
```diff
@@ -0,0 +1,44 @@
...
+ * {@link HttpHeaders} for it to be added in {@link HttpRequest}.
+ */
+public class AddHeadersFromContextPolicy implements HttpPipelinePolicy {
```
srnagar: Add unit tests for this policy.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AddHeadersFromContextPolicy.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.azure.core.http.HttpHeader;
+import com.azure.core.http.HttpHeaders;
+import com.azure.core.http.HttpPipelineCallContext;
+import com.azure.core.http.HttpPipelineNextPolicy;
+import com.azure.core.http.HttpResponse;
+import com.azure.core.util.Context;
+import com.azure.core.http.HttpRequest;
+import static com.azure.core.util.Context.AZURE_REQUEST_HTTP_HEADERS_KEY;
+import reactor.core.publisher.Mono;
```
srnagar: nit: Group static and non-static imports separately.

sdk/core/azure-core/src/test/java/com/azure/core/http/policy/AddHeadersFromContextPolicyTest.java
```diff
@@ -0,0 +1,88 @@
...
+                    Assertions.assertEquals(request.getHeaders().getValue("my-header1"), "my-header1-value");
+                    Assertions.assertEquals(request.getHeaders().getValue("my-header2"), "my-header2-value");
+                    return Mono.just(mockResponse);
+                }
+            })
+            .policies(new RequestIdPolicy())
+            .policies(new AddHeadersFromContextPolicy())
+            .build();
+
+        pipeline.send(new HttpRequest(HttpMethod.GET, new URL("http://localhost/")), new Context(Context.AZURE_REQUEST_HTTP_HEADERS_KEY, headers)).block();
```
srnagar: Can you add some assertions here to test the hypothesis?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AddHeadersFromContextPolicy.java
```diff
@@ -0,0 +1,45 @@
...
+import static com.azure.core.util.Context.AZURE_REQUEST_HTTP_HEADERS_KEY;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * The pipeline policy that override or add  {@link HttpHeaders} in {@link HttpRequest} by reading values from
+ * {@link Context} with key 'azure-http-headers-key'. The value for this key should be of type
+ * {@link HttpHeaders} for it to be added in {@link HttpRequest}.
+ */
```
JonathanGiles: Include code samples to demonstrate how this would work from an end-users perspective. They don't need to be code snippets because you won't be able to reference the  relevant code in app config or elsewhere, so you will need to add a checkstyle suppression.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AddHeadersFromContextPolicy.java
```diff
@@ -0,0 +1,45 @@
...
+public class AddHeadersFromContextPolicy implements HttpPipelinePolicy {
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+
+        Optional<Object> customHttpHeadersObject = context.getData(AZURE_REQUEST_HTTP_HEADERS_KEY);
+        if (customHttpHeadersObject.isPresent()
+            && customHttpHeadersObject.get() instanceof HttpHeaders) {
+
+            HttpHeaders customHttpHeaders = (HttpHeaders) customHttpHeadersObject.get();
```
JonathanGiles: This could be replaced with 

```java
context.getData(AZURE_REQUEST_HTTP_HEADERS_KEY).ifPresent(headers -> {
    // for loop over headers
});
```

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java
```diff
@@ -6,22 +6,47 @@
...
     private static final String REQUEST_ID_HEADER = "x-ms-client-request-id";
+    private final String requestIdHeaderName;
+
+    /**
+     * Creates  {@link RequestIdPolicy} with provided {@code requestIdHeaderName}.
+     * @param requestIdHeaderName to be used to set in {@link HttpRequest}.
+     */
+    public RequestIdPolicy(String requestIdHeaderName) {
+        this.requestIdHeaderName = Objects.requireNonNull(requestIdHeaderName,
+            "messageIdHeaderName can not be null.");
```
JonathanGiles: The variable is not named `messageIdHeaderName`.

sdk/core/azure-core/src/main/java/com/azure/core/util/Context.java
```diff
@@ -23,6 +24,9 @@
 public class Context {
     private final ClientLogger logger = new ClientLogger(Context.class);
 
+    /**Key used to override headers in HttpRequest. The Value for this key should be {@link HttpHeaders}.*/
+    public static final String AZURE_REQUEST_HTTP_HEADERS_KEY = "azure-http-headers-key";
```
JonathanGiles: This isn't the right place for this field. It feels like it should be in the policy, not here.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AddHeadersFromContextPolicy.java
hemanttanwar: included sample. Also discussed with team members here about this sample.

hemanttanwar: @JonathanGiles   Please look into the code sample. This would be last thing before I merge PR. Need your approval.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AddHeadersFromContextPolicy.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public class AddHeadersFromContextPolicy implements HttpPipelinePolicy {
+
+    /**Key used to override headers in HttpRequest. The Value for this key should be {@link HttpHeaders}.*/
+    public static final String AZURE_REQUEST_HTTP_HEADERS_KEY = "azure-http-headers-key";
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        context.getData(AZURE_REQUEST_HTTP_HEADERS_KEY).ifPresent(headers -> {
+            if (headers instanceof HttpHeaders) {
```
hemanttanwar: I am checking for `headers instanceof HttpHeaders` to make sure we do not have ClassCastException if end user did not use `HttpHeaders` .

conniey:
If you add some keywords (ie. Fixes) to your PRs, GitHub will automatically close the associated issue when the PR is merged. So you don't have to do that step manually.
https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

hemanttanwar:
This closes #6217

PR #6606, opened 2019-12-02
Verify pom file correctness in regards to versions and tags

JimSuplizio:
A great deal of work was done to centralize version management. As part of that work all dependency and plugins in POM files now require a version and a version update tag. This set of changes adds a verification script which will be run first thing as part of the Analyze step. The script will process the entire enlistment and verify the following:
1. There are no duplicate entries in any of the version_*.txt files
2. There are no duplicate entries in the external_dependencies.txt file
3. POM file verification across the enlistment which includes the following:
   a. There are no <dependencyManagement> sections
   a. Every <dependency> and <plugin> has a <groupId>, <artifactId> and <version>
   b. Every <version> has the appropriate x-version-update tag
   c. The <version>'s value is the same as the value in the version_*txt file or external_dependency

Output:

- If any errors are encountered, will report them at time they are encountered and continue processing. 
- Errors for any duplicate entries in the various version files will be reported before any POM files
- are processed.
- Errors for a given pom file will appear after the "processing pomFile=<fullPathAndFileName" line for that file and before the processing line for the next file. 
- Missing version x-version-update tag errors will cause the output of what the tag should be, in the cases where it is known (ie. pom file's version or parent's version) but in the cases of dependencies it'll output a tag and the appropriate type (current, dependency or external_dependency) will have to be selected. For example:
{x-version-update;<groupId>:<artifactId>;current|dependency|external_dependency<select one>}

This script can be run locally from the root of an enlistment. .\eng\versioning\pom_file_version_scanner.ps1

### DIFF
--- a/eng/versioning/pom_file_version_scanner.ps1
+++ b/eng/versioning/pom_file_version_scanner.ps1
@@ -4,25 +4,25 @@
 # Use case: This script verifies the following:
 # 1. There are no duplicate entries in any of the version_*.txt files
 # 2. There are no duplicate entries in the external_dependencies.txt file
-# 3. POM file verification across the enlistment which includes the following:
+# 3. POM file verification across the repo which includes the following:
 #    a. There are no <dependencyManagement> sections
 #    a. Every <dependency> and <plugin> has a <groupId>, <artifactId> and <version>
 #    b. Every <version> has the appropriate x-version-update tag
 #    c. The <version>'s value is the same as the value in the version_*txt file or external_dependency
 #
 # Output:
-# This script will process the entire enlistment. If any errors are encountered, will report them at
-# time they are encountered and continue processing.
+# This script will process the entire repo. If any errors are encountered, it will report them at
+# the time they are encountered and continue processing.
 # Errors for any duplicate entries in the various version files will be reported before any POM files
 # are processed.
-# Errors for a given pom file will appear after the "processing pomFile=<fullPathAndFileName" line for
+# Errors for a given pom file will appear after the "processing pomFile=<fullPathAndFileName>" line for
 # that file and before the processing line for the next file. Missing version x-version-update tag
 # errors will cause the output of what the tag should be in the cases where it is known (ie. pom file's
 # version or parent's version). In the cases of dependencies it'll output a tag and the appropriate
 # type (current, dependency or external_dependency) will have to be selected. For example:
 # <!-- {x-version-update;<groupId>:<artifactId>;current|dependency|external_dependency<select one>} -->"
 #
-# This script can be run locally from the root of an enlisment. .\eng\versioning\pom_file_version_scanner.ps1
+# This script can be run locally from the root of the repo. .\eng\versioning\pom_file_version_scanner.ps1
 
 # Since we're skipping Management for the moment, only look for files with certain parents. These
 # limitations will vanish once Management track is updated.
@@ -432,7 +432,7 @@ Write-Output "Total run time=$($TotalRunTime)"
 if ($script:FoundError)
 {
     Write-Output "There were errors encountered during execution. Please fix any errors and run the script again."
-    Write-Output "This script can be run locally from the root of an enlisment. .\eng\pom_file_version_scanner.ps1"
+    Write-Output "This script can be run locally from the root of the repo. .\eng\pom_file_version_scanner.ps1"
     exit(1)
 }
 


eng/versioning/external_dependencies.txt
```diff
@@ -16,7 +16,6 @@ com.microsoft.azure:azure-client-authentication;1.6.15
 com.microsoft.azure:azure-client-runtime;1.6.15
 com.microsoft.azure:azure-core;0.9.8
 com.microsoft.azure:azure-keyvault-cryptography;1.2.2
-com.microsoft.azure:azure-mgmt-keyvault;1.11.1
```
JimSuplizio: This was a duplicate entry (both were the same). Since this is a test dependency I removed this one and left the one there. Yes, the verification script found this..

sdk/eventgrid/microsoft-azure-eventgrid/pom.xml
```diff
@@ -110,7 +110,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
-                <version>2.8</version> <!-- {x-version-update;eventgrid_org.apache.maven.plugins:maven-javadoc-plugin;external_dependency} -->
+                <version>3.1.1</version> <!-- {x-version-update;org.apache.maven.plugins:maven-javadoc-plugin;external_dependency} -->
```
JimSuplizio: The POM file exception for eventgrid_org.apache.maven.plugins:maven-javadoc-plugin was removed and all tooling should have been at the same version. The script detected and report this.

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,370 @@
...
+    [string]$id
+    [string]$depVer
+    [string]$curVer
+    Dependency(
+        [string]$inputString
+    ){
+        $split = $inputString.Split(";")
+        $this.id = $split[0]
+        $this.depVer = $split[1]
+        $this.curVer = $split[2]
```
alzimmermsft: Should this do any validation on the number of substrings? What happens if there is an improper amount of substrings?

Should this do a pattern match on the depVer or curVer to ensure they are proper Maven dependencies?

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,370 @@
...
+
+class ExternalDependency {
+    [string]$id
+    [string]$ver
+    ExternalDependency(
+        [string]$inputString
+    ){
+        $split = $inputString.Split(";")
+        $this.id = $split[0]
+        $this.ver = $split[1]
```
alzimmermsft: Same question as the `Dependency` class.

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,370 @@
...
+# Not all POM files have a parent entry
+$PomFilesIgnoreParent = ("$($Path)\parent\pom.xml")
+$script:FoundError = $false
+$StartTime = $(get-date)
+
+class Dependency {
+    [string]$id
+    [string]$depVer
+    [string]$curVer
+    Dependency(
```
alzimmermsft: Would it be possible to add a comment here and in `ExternalDependency` on the expected `inputString` format?

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,370 @@
...
+            }
+            else
+            {
+                # verify the version tag and version are correct
+                $retVal = Test-Dependency-Tag-And-Version $libHash $extDepHash $versionNode.InnerText $versionNode.NextSibling.Value
+                if ($retVal)
+                {
+                    $script:FoundError = $true
+                    Write-Output "$($retVal)"
+                }
```
alzimmermsft: This block of code appears to be common, any chance this could become a function?

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,370 @@
...
+            }
+        }
+        else 
+        {   
+            $script:FoundError = $true
+            # <!-- {x-version-update;<groupId>:<artifactId>;current} -->
+            Write-Output "Error: Missing dependency version update tag for groupId=$($groupId), artifactId=$($artifactId). The tag should be <!-- {x-version-update;$($groupId):$($artifactId);current|dependency|external_dependency<select one>} -->"
+        }
+    }
+    # Verify every plugin as a group, artifact and version
```
alzimmermsft:
```suggestion
    # Verify every plugin has a group, artifact and version
```

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,370 @@
...
+        }
+        else 
+        {
+            # output an error for missing version element
+            $script:FoundError = $true
+            Write-Output "Error: Could not find project/parent/version node for $($pomFile)"
+        }    
+    }
+
+    # Verify every dependency as a group, artifact and version
```
alzimmermsft:
```suggestion
    # Verify every dependency has a group, artifact and version
```

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,370 @@
...
+    # GetElementsByTagName should get all dependencies including dependencies under plugins
+    foreach($dependencyNode in $xmlPomFile.GetElementsByTagName("dependency"))
+    {
+        $artifactId = $dependencyNode.artifactId
+        $groupId = $dependencyNode.groupId
```
alzimmermsft: Does `groupId` need to be validated here as it does during the plugin validation?

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,370 @@
...
+        {   
+            $script:FoundError = $true
+            # <!-- {x-version-update;<groupId>:<artifactId>;current} -->
+            Write-Output "Error: Missing dependency version update tag for groupId=$($groupId), artifactId=$($artifactId). The tag should be <!-- {x-version-update;$($groupId):$($artifactId);current|dependency|external_dependency<select one>} -->"
+        }
+    }
+    # Verify every plugin as a group, artifact and version
+    # Verify every dependency as a group, artifact and version
+    # GetElementsByTagName should get all dependencies including dependencies under plugins
+    foreach($pluginNode in $xmlPomFile.GetElementsByTagName("plugin"))
```
alzimmermsft: Dependency and plugin validation look very similar, any chance they could be merged into a common function which takes in error formatting strings to print error output?

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: @alzimmermsft it's trivial to add verification for the number of substrings and I can add that for correctness (though malformed entries would, in theory, get hit when the update_versions script is run). 

I really don't want to add anything that verifies everything is being used just yet only because Management is still being excluded until its centralized versioning work is done.

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: Yup, I'll do that.

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: No. For dependencies you could have a groupId and artifactId without a version but for plugins only an artifactId was necessary which is why both had to be verified.

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: No, unfortunately not. Write-Output will not work from inside a function (not in the way any sane person would think it would) which is why I'd opted to return the strings. Write-Error would work but the formatting of Write-Error is such where I'm only trying to use that when there are really dire (like malformed version/dependency files). For missing tags or versions I'm intentionally using Write-Output to list everything that's wrong for a given pom file without all the embellishments that Write-Error gives.

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: Fixed locally, will be updated with the next commit.

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: Fixed locally, will be updated with the next commit.

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: They're similar and slightly different but due to the way powershell deals with Write-Output combing these is far more trouble than what it's worth to condense a few lines of code that, honestly, aren't going to change (unless Maven suddenly changes its format but that's seriously doubtful).

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: Same answer as above. I'll add verification for the number of substrings.

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,438 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Use case: This script verifies the following:
+# 1. There are no duplicate entries in any of the version_*.txt files
+# 2. There are no duplicate entries in the external_dependencies.txt file
+# 3. POM file verification across the enlistment which includes the following:
```
weshaggard: nit: enlistment is old terminology in the git world we generally refer to it as a repo.

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,438 @@
...
+#    a. There are no <dependencyManagement> sections
+#    a. Every <dependency> and <plugin> has a <groupId>, <artifactId> and <version>
+#    b. Every <version> has the appropriate x-version-update tag
+#    c. The <version>'s value is the same as the value in the version_*txt file or external_dependency
+#
+# Output:
+# This script will process the entire enlistment. If any errors are encountered, will report them at
+# time they are encountered and continue processing.
```
weshaggard: enlistment ->repo

"will report them at time they are encountered and continue processing." -> "it will report them at the time they are encountered and continue processing"

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,438 @@
...
+# Errors for any duplicate entries in the various version files will be reported before any POM files
+# are processed.
+# Errors for a given pom file will appear after the "processing pomFile=<fullPathAndFileName" line for
```
weshaggard: `<fullPathAndFileName` -> `<fullPathAndFileName>`

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: Already changed, will be updated with the next commit.

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: Already changed and will be updated with the next commit.

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: Already changed and will be updated with the next commit.

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,438 @@
...
+    Write-Output "processing pomFile=$($pomFile)"
+    $dependencyManagement = $xmlPomFile.GetElementsByTagName("dependencyManagement")[0]
+    if ($dependencyManagement)
+    {
+        Write-Output "Error: <dependencyManagement> is not allowed. Every dependency must have its own version and version update tag"
+    }
+
+    # Ensure that the project has a version tag with the exception of projects under the eng directory which
+    # aren't releasing libraries but still need to have their dependencies checked
+    if ($pomFile.Split([IO.Path]::DirectorySeparatorChar) -notcontains "eng")
```
weshaggard: Don't you still want to verify dependencies under eng folder? If the only thing missing from the eng projects is a version then we should probably just add that version and not treat them special.

weshaggard: Nevermind I see you are still verifying the other stuff below, but it still might be interesting to add a version to the eng pom files to avoid this special case.

eng/versioning/pom_file_version_scanner.ps1
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    # Verify every dependency as a group, artifact and version
+    # GetElementsByTagName should get all dependencies including dependencies under plugins
+    foreach($dependencyNode in $xmlPomFile.GetElementsByTagName("dependency"))
+    {
+        $artifactId = $dependencyNode.artifactId
+        $groupId = $dependencyNode.groupId
+        $versionNode = $dependencyNode.GetElementsByTagName("version")[0]
+        if (!$versionNode)
```
weshaggard: Looks like the version and version tag logic might be worth a helper function since it seems to be duplicated in a few places.

JimSuplizio: This ends up being done twice, once for plugins and once for dependencies and the difference is that I'd wanted very tailed error messages. Without having line numbers I'd wanted to be more specific about whether something is coming from a plugin or dependency.

eng/versioning/pom_file_version_scanner.ps1
JimSuplizio: @weshaggard The main reason I'd opted to skip the version check for the tools under the eng directory is that we're not shipping them and I didn't see the need to additional bookkeeping to one of the tracking files for them.

Issue #6608, opened 2019-12-02
Compilable code snippets in README files


srnagar:
READMEs are the first point of interaction for users looking to use our SDKs. It contains several useful code snippets for users to get started quickly. It is important to keep these code samples correct at all times. However, when code is modified README gets outdated and hard to figure out which README is referencing the modified piece of code. 

Use https://github.com/zakhenry/embedme to embed code snippets from compilable Java files.

PR #6609, opened 2019-12-02
README - compilable code snippets 

srnagar:
READMEs are the first point of interaction for users looking to use our SDKs. It contains several useful code snippets for users to get started quickly. It is important to keep these code samples correct at all times. However, when code is modified README gets outdated and hard to figure out which README is referencing the modified piece of code. 

So, this change will introduce a mechanism to inject code from a compiled Java class to ensure freshness of samples. 

Tool used - https://github.com/zakhenry/embedme

How to use:
1. Install NodeJS if you don't already have it.
2. In README files, use the following language tag for all code snippets that you want to provide:
<pre>
```java
```
</pre>
3. Within the tag, add a comment (should be the only line) providing the path (relative to the README file) to the Java file containing the code snippet that you want to replace with. 
```
// path/to/class/containing/Samples.java
```
4. A sub-section of the Java file can also be used by providing line numbers.
```
// path/to/class/containing/Samples.java#L20-L30
```
5. Next run the `embedme` tool as shown below:
```node
npx embedme path/to/README.md
```
6. The code should now be inserted into the README file from the java class. 
7. Verify the changes to the README file. 

This PR is using this tool for 
- Core libraries
- Identity

Partially fixes https://github.com/Azure/azure-sdk-for-java/issues/6608


JonathanGiles:
I like this a lot. It will help to ensure our samples are always accurate and compiling. I think introducing a new sample file (ReadmeSamples.java) is a good idea and should be used consistently, especially given the line referencing approach used in the markdown. I would recommend you introduce some warning text at the top of these files to warn users against making any changes to the code that may result in line number differences.

alzimmermsft:
Would it be possible to update the Blob Batch README to use these? It already has a `ReadmeCodeSamples` class.

alzimmermsft:
Need to update the exclusion regex here to include ReademeSamples classes.

https://github.com/Azure/azure-sdk-for-java/blob/master/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml#L558

sdk/core/azure-core-http-netty/README.md
```diff
@@ -34,6 +34,8 @@ The following sections provide several code snippets covering some of the most c
 Create a Netty Http client that uses port 80 and has no proxy.
 
 ```java
+// ./src/samples/java/com/azure/core/http/netty/ReadmeSamples.java#L19-L19
```
alzimmermsft: Is there a way to hide this in the README?

srnagar:
@JonathanGiles Added a warning message to ReadmeSamples.java to ensure line numbers are correct.
@alzimmermsft - Updated storage blob batch README to use the code snippets in java file.

sdk/core/azure-core-http-netty/README.md
srnagar: There is a way to hide but I like to leave it as it is so that we'll know where the snippets are coming from and leaving it here will allow for future updates to automatically pick up the changes.

eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
```diff
@@ -558,7 +558,8 @@
   <Match>
     <Or>
       <Class name="~.*JavaDoc(CodeSnippets|CodeSamples|Samples)"/>
-      <Class name="com.azure.storage.blob.batch.ReadmeCodeSamples"/>
+      <Class name="~.*ReadmeSamples"/>
+      <Class name="com.azure.storage.blob.batch.ReadmeSamples"/>
```
JonathanGiles: You shouldn't need to explicitly call out the batch file should you?

sdk/core/azure-core-http-netty/README.md
```diff
@@ -34,6 +34,8 @@ The following sections provide several code snippets covering some of the most c
 Create a Netty Http client that uses port 80 and has no proxy.
 
 ```java
+// ./src/samples/java/com/azure/core/http/netty/ReadmeSamples.java#L23-L23
```
JonathanGiles: I agree with Alan, I'm not convinced that we want these to show up in all of the code samples. If we can hide it, whilst still retaining the ability for it to be updated when the sample code is changed, that would be best IMHO.

JonathanGiles:
Should we introduce tooling into /eng that would run this over all readme files?

sdk/core/azure-core-http-netty/README.md
srnagar: Using alternate embedding syntax that will hide the comment in README and will continue to be updated when sample code changes.

eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
srnagar: Removed it.

srnagar:
@JonathanGiles Yes, I will have a follow-up task to introduce tooling that will automatically run this to update READMEs. This PR is a first step to get everyone's thoughts on the tool and usage. If we want to go ahead with this, I will look into integrating with eng tools.

Issue #6611, opened 2019-12-02
Clone HttpLogOptions instead of Object Reference in HttpLoggingPolicy


alzimmermsft:
`HttpLoggingPolicy` current maintains a object reference to the `HttpLogOptions` which is used to construct it. Given that it is an object reference the underlying `HttpLogOptions` could be mutated to change how logging happens.
 `HttpLoggingPolicy` is meant to be immutable once constructed so it may make more sense for the `HttpLogOptions` to be cloned instead to ensure that logging is consistent while a given policy is being used.

Issue #6506, opened 2019-11-22
Only Log Headers When Log Level is Verbose


alzimmermsft:
Only log the `HttpHeaders` in a request or a response when the log level is set to VERBOSE.

PR #6614, opened 2019-12-02
Updates to HttpLoggingPolicy

### LINKED ISSUE #6504: Atomically Log in HttpLoggingPolicy
`HttpLoggingPolicy` will log the request in pieces, this results in potential race conditions where the request or response could be logged intersperse with other requests and response. The policy's logging should be updated to generate a single log string and perform a single logging operation. 

An additional question which should be answer by having the entirety of the response and request being logged as WARNING if the response is an error state.

alzimmermsft:
Fixes #6504, #6506, and #6611 

Updates `HttpLoggingPolicy` to clone the passed `HttpLogOptions` instead of maintaining a reference to the passed log options. `HttpHeaders` are now only logged when the environment logging level is set to `VERBOSE`. Logging is now done atomically by having the entire log message logged in a single logging request instead of numerous small logging requests.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
@@ -39,7 +39,7 @@ public class HttpLoggingPolicy implements HttpPipelinePolicy {
     private final HttpLogDetailLevel httpLogDetailLevel;
     private final Set<String> allowedHeaderNames;
     private final Set<String> allowedQueryParameterNames;
-    private final boolean prettyPrintJSON;
+    private final boolean prettyPrintJson;
 
     /**
      * Creates an HttpLoggingPolicy with the given log configurations.
@@ -58,7 +58,7 @@ public class HttpLoggingPolicy implements HttpPipelinePolicy {
      * include body logging, this flag does nothing.
      */
     private HttpLoggingPolicy(HttpLogOptions httpLogOptions, boolean prettyPrintJson) {
-        this.prettyPrintJSON = prettyPrintJson;
+        this.prettyPrintJson = prettyPrintJson;
 
         if (httpLogOptions == null) {
             this.httpLogDetailLevel = HttpLogDetailLevel.NONE;
@@ -182,9 +182,13 @@ public class HttpLoggingPolicy implements HttpPipelinePolicy {
         if (httpLogDetailLevel.shouldLogUrl()) {
             responseLogMessage.append("<-- ")
                 .append(response.getStatusCode())
-                .append(" ").append(getRedactedUrl(response.getRequest().getUrl()))
-                .append(" (").append(tookMs).append(" ms, ")
-                .append(bodySize).append(")")
+                .append(" ")
+                .append(getRedactedUrl(response.getRequest().getUrl()))
+                .append(" (")
+                .append(tookMs)
+                .append(" ms, ")
+                .append(bodySize)
+                .append(")")
                 .append(System.lineSeparator());
         }
 
@@ -228,7 +232,7 @@ public class HttpLoggingPolicy implements HttpPipelinePolicy {
      * @return A flag indicating if logging should be skipped.
      */
     private boolean shouldLoggingBeSkipped(int environmentLogLevel) {
-        return environmentLogLevel <= LogLevel.INFORMATIONAL.toNumeric();
+        return environmentLogLevel > LogLevel.INFORMATIONAL.toNumeric();
     }
 
     /*
@@ -314,7 +318,7 @@ public class HttpLoggingPolicy implements HttpPipelinePolicy {
      */
     private String prettyPrintIfNeeded(ClientLogger logger, String contentType, String body) {
         String result = body;
-        if (prettyPrintJSON && contentType != null
+        if (prettyPrintJson && contentType != null
             && (contentType.startsWith(ContentType.APPLICATION_JSON) || contentType.startsWith("text/json"))) {
             try {
                 final Object deserialized = PRETTY_PRINTER.readTree(body);


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -52,105 +57,142 @@ public HttpLoggingPolicy(HttpLogOptions httpLogOptions) {
...
+
+        if (httpLogOptions == null) {
+            this.httpLogDetailLevel = HttpLogDetailLevel.NONE;
+            this.allowedHeaderNames = Collections.emptySet();
+            this.allowedQueryParameterNames = Collections.emptySet();
+        } else {
+            this.httpLogDetailLevel = httpLogOptions.getLogLevel();
+            this.allowedHeaderNames = httpLogOptions.getAllowedHeaderNames()
+                .stream()
+                .map(String::toLowerCase)
```
srnagar: Wouldn't this result in a spotbugs issue for not using `Locale.ROOT`?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -52,105 +57,142 @@ public HttpLoggingPolicy(HttpLogOptions httpLogOptions) {
...
-                        });
-                    } catch (Exception e) {
-                        reqBodyLoggingMono = Mono.error(e);
-                    }
+                String contentType = request.getHeaders().getValue("Content-Type");
+                long contentLength = getContentLength(logger, request.getHeaders());
+
+                if (bodyIsPrintable(contentType, contentLength)) {
+                    requestLoggingMono = FluxUtil.collectBytesInByteBufferStream(request.getBody()).flatMap(bytes ->
+                        Mono.just(requestLogMessage.append(String.format("%d-byte body:%n%s%n--> END %s%n",
```
srnagar: Avoid using `String.format()` as this code gets executed for every request (if logging is enabled) and the performance of `String.format()` is bad.

Also, the environment variable for logging might be set to a lower numeric value but the slf4j log level might be higher which means that all these strings will be created but not really logged.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -52,105 +57,142 @@ public HttpLoggingPolicy(HttpLogOptions httpLogOptions) {
...
-            requestUrl.setQuery(getAllowedQueryString(request.getUrl().getQuery()));
-            logger.info("--> {} {}", request.getHttpMethod(), requestUrl.toString());
+    private Mono<String> logRequest(final ClientLogger logger, final HttpRequest request) {
+        /*
+         * Logging is either disabled or the logging level is above information (warning or error), this will result
+         * in nothing being logged so perform a no-op.
+         */
+        int numericLogLevel = LoggingUtil.getEnvironmentLoggingLevel().toNumeric();
+        if (numericLogLevel == LogLevel.DISABLED.toNumeric()
+            || numericLogLevel > LogLevel.INFORMATIONAL.toNumeric()) {
```
srnagar: Can this condition be simplified by just using `numericLogLevel > LogLevel.INFORMATIONAL.toNumeric()` since `LogLevel.DISABLED` is 5?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -52,105 +57,142 @@ public HttpLoggingPolicy(HttpLogOptions httpLogOptions) {
...
+    private Mono<HttpResponse> logResponse(final ClientLogger logger, final HttpResponse response, long startNs) {
+        /*
+         * Logging is either disabled or the logging level is above information (warning or error), this will result
+         * in nothing being logged so perform a no-op.
+         */
+        int numericLogLevel = LoggingUtil.getEnvironmentLoggingLevel().toNumeric();
+        if (numericLogLevel == LogLevel.DISABLED.toNumeric()
+            || numericLogLevel > LogLevel.INFORMATIONAL.toNumeric()) {
+            return Mono.just(response);
+        }
```
srnagar: This is common code between logRequest() and logResponse(). This can be extracted as a method to reduce duplication.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/LoggingUtil.java
```diff
@@ -0,0 +1,36 @@
...
+package com.azure.core.implementation;
+
+import com.azure.core.util.Configuration;
+import com.azure.core.util.CoreUtils;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public final class LoggingUtil {
```
srnagar: javadoc.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -19,22 +20,25 @@
...
 /**
  * The pipeline policy that handles logging of HTTP requests and responses.
  */
 public class HttpLoggingPolicy implements HttpPipelinePolicy {
     private static final ObjectMapper PRETTY_PRINTER = new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);
-    private final HttpLogOptions httpLogOptions;
-    private final boolean prettyPrintJSON;
     private static final int MAX_BODY_LOG_SIZE = 1024 * 16;
     private static final String REDACTED_PLACEHOLDER = "REDACTED";
+    private static final String APPLICATION_OCTET_STREAM = "application/octet-stream";
```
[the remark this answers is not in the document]
alzimmermsft: Updated

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -52,105 +56,142 @@ public HttpLoggingPolicy(HttpLogOptions httpLogOptions) {
      * @param prettyPrintJSON If true, pretty prints JSON message bodies when logging. If the detailLevel does not
      * include body logging, this flag does nothing.
      */
-    HttpLoggingPolicy(HttpLogOptions httpLogOptions, boolean prettyPrintJSON) {
-        this.httpLogOptions = httpLogOptions;
+    private HttpLoggingPolicy(HttpLogOptions httpLogOptions, boolean prettyPrintJSON) {
```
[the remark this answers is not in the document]
alzimmermsft: Done

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: Apparently is doesn't, don't know if this is a method reference issue. Will update to use Locale.ROOT.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: Removed all `String.format()`s, now using `StringBuilder.append` for all additions to the log message.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: Good catch, changed.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: Created a method for these.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/LoggingUtil.java
alzimmermsft: Added

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -19,23 +22,25 @@
...
     private static final ObjectMapper PRETTY_PRINTER = new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);
-    private final HttpLogOptions httpLogOptions;
-    private final boolean prettyPrintJSON;
     private static final int MAX_BODY_LOG_SIZE = 1024 * 16;
     private static final String REDACTED_PLACEHOLDER = "REDACTED";
 
+    private final HttpLogDetailLevel httpLogDetailLevel;
+    private final Set<String> allowedHeaderNames;
+    private final Set<String> allowedQueryParameterNames;
+    private final boolean prettyPrintJSON;
```
srnagar: nit: prettyPrintJSON -> prettyPrintJson

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -49,108 +54,202 @@ public HttpLoggingPolicy(HttpLogOptions httpLogOptions) {
...
+    /*
+     * Logs thr HTTP response.
+     *
+     * @param logger Logger used to log the response.
+     * @param response HTTP response returned from Azure.
+     * @param startNs Nanosecond representation of when the request was sent.
+     * @return A Mono containing the HTTP response.
+     */
+    private Mono<HttpResponse> logResponse(final ClientLogger logger, final HttpResponse response, long startNs) {
+        int numericLogLevel = LoggingUtil.getEnvironmentLoggingLevel().toNumeric();
```
srnagar: This line can be moved inside `shouldLoggingBeSkipped()` as well and the method can be parameter-less.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: Done

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: `numericLogLevel` is a parameter that is passed into adding headers to the log message, this is done to maintain a consistent logging state while the message is being generated. This will stay for now.

PR #6621, opened 2019-12-03
comment out the possibly test caused timeout

sima-zhu:


sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -169,20 +170,20 @@ class APISpec extends Specification {
...
+//    def cleanup() {
+//        def options = new ListBlobContainersOptions().setPrefix(containerPrefix + testName)
+//        for (BlobContainerItem container : primaryBlobServiceClient.listBlobContainers(options, Duration.ofSeconds(120))) {
+//            BlobContainerClient containerClient = primaryBlobServiceClient.getBlobContainerClient(container.getName())
+//
+//            if (container.getProperties().getLeaseState() == LeaseStateType.LEASED) {
+//                createLeaseClient(containerClient).breakLeaseWithResponse(0, null, null, null)
+//            }
+//
+//            containerClient.delete()
```
hemanttanwar: Wondering how these resources will be cleaned up  after test are run ?

sima-zhu: The PR is for testing the live tests for storage. It is not for real issue.

sima-zhu:
Close the one and move to alan's PR

PR #6622, opened 2019-12-03
Return an object for force close handles

gapra-msft:


sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareFileClient.java
```diff
@@ -896,11 +899,12 @@ public void forceCloseHandle(String handleId) {
...
      * concludes a {@link RuntimeException} will be thrown.
      * @param context Additional context that is passed through the Http pipeline during the service call.
-     * @return The number of handles closed.
+     * @return Information about the closed handles
      */
-    public int forceCloseAllHandles(Duration timeout, Context context) {
+    public CloseHandlesInfo forceCloseAllHandles(Duration timeout, Context context) {
         return new PagedIterable<>(shareFileAsyncClient.forceCloseAllHandlesWithOptionalTimeout(timeout, context))
-            .stream().reduce(0, Integer::sum);
+            .stream().reduce(new CloseHandlesInfo(0),
```
alzimmermsft: Instead of instantiating the `CloseHandlesInfo` object here should the reduce instead just begin with an integer and then map to `CloseHandlesInfo`. That way the object only needs to be created once instead of on each item emitted in the stream.

```java
.stream.reduce(0, Integer::sum).map(CloseHandlesInfo::new)
```

gapra-msft: discussed offline about this

gapra-msft:
> Can you update the changelog as well?

Done!

PR #6623, opened 2019-12-04
Hide secret info from log info (#6555)

alzimmermsft:
* Hide secret info from log info

### DIFF
--- a/eng/versioning/version_client.txt
+++ b/eng/versioning/version_client.txt
@@ -3,7 +3,7 @@
 
 com.azure:azure-sdk-parent;1.6.0;1.6.0
 com.azure:azure-client-sdk-parent;1.7.0;1.7.0
-com.azure:azure-core;1.1.1;1.1.1
+com.azure:azure-core;1.1.0;1.1.1
 com.azure:azure-core-amqp;1.0.0-beta.8;1.0.0-beta.8
 com.azure:azure-core-http-netty;1.1.0;1.1.0
 com.azure:azure-core-http-okhttp;1.1.0-beta.1;1.1.0-beta.1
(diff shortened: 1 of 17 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -196,9 +191,11 @@ private String getAllowedQueryString(String queryString) {
                 bodySize = contentLengthString + "-byte";
             }
             HttpLogDetailLevel httpLogLevel = httpLogOptions.getLogLevel();
-            // HttpResponseStatus responseStatus = HttpResponseStatus.valueOf(response.statusCode());
             if (httpLogLevel.shouldLogUrl()) {
-                logger.info("<-- {} {} ({} ms, {} body)", response.getStatusCode(), url, tookMs, bodySize);
+                UrlBuilder requestUrl = UrlBuilder.parse(url);
+                requestUrl.setQuery(getAllowedQueryString(url.getQuery()));
+                logger.info("<-- {} {} ({} ms, {} body)", response.getStatusCode(), requestUrl.toString(), tookMs,
+                    bodySize);
```
srnagar: L85-L87 and this segment can be de-duped.

eng/versioning/version_client.txt
```diff
@@ -3,7 +3,7 @@
 
 com.azure:azure-sdk-parent;1.6.0;1.6.0
 com.azure:azure-client-sdk-parent;1.7.0;1.7.0
-com.azure:azure-core;1.1.0;1.1.0
+com.azure:azure-core;1.1.1;1.1.1
```
JimSuplizio: I'm going to approve the changes but I do have a question. Did we need to update both the _dependency_ and _current_ versions or should the _dependency_ version still be 1.1.0?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: Leaving as is to match the commit.

PR #6624, opened 2019-12-04
Common Base Class For Storage Tests

### LINKED ISSUE #5259: Add a Test Common
There is common logic duplicated across the Storage modules with regards to testing, these commonalities should be promoted to Storage Common and standardized. The key areas are:

- [ ] Naming playback JSON files
- [ ] Generating and capturing random data
  - [x] Byte arrays
  - [x] Resource names
  - [x] UUIDs
  - [x] Datetimes
- [x] Creating the `HttpClient` used in clients
- [x] Determining which tests should have playback JSON files records. Not all tests are replayable due to APIs using random resources (datetimes, UUIDs, etc) or due to the playback file size generated (downloading large data sets, numerous service calls).

alzimmermsft:
Fixes #5259 

`StorageTestBase` is added into Azure Storage Common and a new build step was added for Storage Common to generate a test-jar. The Storage Common test jar is added as a test dependency in Storage Blob, Blob Batch, Blob Cryptography, File Datalake, File Share, and Queue. `StorageTestBase` contains methods which were common across all Storage modules to remove code duplication and standardize the Storage testing framework.

File Share and Queue unit tests were updated to use the naming convention that was used across Blob, Blob Batch, Blob Cryptography, and File Datalake, playback JSON files have been regenerated accordingly.


alzimmermsft:
Closing as this runs into access issues with JDK 11.

PR #6625, opened 2019-12-04
Add maven profile to verify readme

srnagar:
The intent of this PR is to add maven tools to update README files whenever sample code in Java file is updated. This also verifies that README is up-to-date and fails the build if the README file does not match the code sample in Java file.

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -974,6 +974,43 @@
     </profile>
 
     <profile>
+      <!-- separate profile for windows as the executable on windows is named npx.cmd and npx on other os families -->
+      <id>readme-codesnippet-windows</id>
+      <activation>
+        <property>
+          <name>readme-codesnippet-windows</name>
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>exec-maven-plugin</artifactId>
+            <version>1.2.1</version> <!-- {x-version-update;org.codehaus.mojo:exec-maven-plugin;external_dependency} -->
+            <executions>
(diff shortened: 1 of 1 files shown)


JimSuplizio:
@srnagar 
I'm curious as to what the process is here. Does the README get updated with the code sample during a local build, if the developer changes the sample, and then the developer submits any changed README with their changes?

srnagar:
@JimSuplizio Yes, here's the process:
- the developers will locally edit ReadmeSamples.java to update/add code snippets
- run `mvn package -Preadme-codesnippets -f pom.client.xml` to automatically update the README.md files with the new code samples
- check-in the java and readme files for PR
- pipeline then runs `verify` on this PR and
- if the readme doesn't match the java files, an error is thrown and the build fails. 

This is to catch cases where developers updated the `ReadmeSamples.java` file only and did not update the readme files.

pom.client.xml
```diff
@@ -972,6 +972,150 @@
...
+                <phase>prepare-package</phase>
+                <goals>
+                  <goal>exec</goal>
+                </goals>
+                <configuration>
+                  <executable>npx</executable>
+                  <arguments>
+                    <argument>embedme</argument>
+                    <argument>sdk/*/azure-*/README.md</argument>
+                  </arguments>
```
hemanttanwar: Would  this also cover the README in modules. For example 

sdk/eventhubs/README.md
sdk/eventhubs/azure-messaging-eventhubs/README.md

srnagar: As discussed, the code samples should only be at the module level and shouldn't be at sdk/*/ level. So, not including those READMEs here.

PR #6637, opened 2019-12-04
Version and changelog updates

rickle-msft:


sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/implementation/util/BuilderHelper.java
```diff
@@ -39,7 +39,7 @@
  */
 public final class BuilderHelper {
     private static final String DEFAULT_USER_AGENT_NAME = "azure-storage-file-datalake";
-    private static final String DEFAULT_USER_AGENT_VERSION = "12.0.0-preview.5";
+    private static final String DEFAULT_USER_AGENT_VERSION = "12.0.0-preview.7";
```
gapra-msft: Would you mind adding the version update text to something like this? I missed it when I put datalake out

```
    private static final String DEFAULT_USER_AGENT_NAME = "azure-storage-blob";
    // {x-version-update-start;com.azure:azure-storage-blob;current}
    private static final String DEFAULT_USER_AGENT_VERSION = "12.1.0-beta.1";
    private static final String DEFAULT_USER_AGENT_VERSION = "12.1.0";
    // {x-version-update-end}
```

rickle-msft: Done

eng/jacoco-test-coverage/pom.xml
```diff
@@ -110,37 +110,37 @@
...
       <groupId>com.azure</groupId>
       <artifactId>azure-storage-file-share</artifactId>
-      <version>12.0.0-beta.6</version> <!-- {x-version-update;com.azure:azure-storage-file-share;current} -->
+      <version>12.0.0</version> <!-- {x-version-update;com.azure:azure-storage-file-share;current} -->
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-storage-file-datalake</artifactId>
-      <version>12.0.0-beta.7</version> <!-- {x-version-update;com.azure:azure-storage-file-datalake;current} -->
+      <version>12.0.0-preview.7</version> <!-- {x-version-update;com.azure:azure-storage-file-datalake;current} -->
```
alzimmermsft: Should this be `12.0.0-beta.7`? Preview messes up Maven's most recent release structure. Or is this being done to ensure that preview.7 shows up as the most recent?

sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/AzureKeyVaultConfiguration.java
```diff
@@ -7,6 +7,6 @@
     //TODO: Eventually remove these hardcoded strings with https://github.com/Azure/azure-sdk-for-java/issues/3141
     static final String SDK_NAME = "Azure-Keyvault";
     // {x-version-update-start;com.azure:azure-security-keyvault-secrets;current}
-    static final String SDK_VERSION = "4.1.0-beta.1";
+    static final String SDK_VERSION = "4.0.1";
```
alzimmermsft: @g2vinay was this missed during the KeyVault release?

rickle-msft: I don't know keyvault details, but I did look into this, and this is what is in client_version.txt and is unmodified by my changes. So the version script just picked this up and it looks like maybe it was somehow skipped earlier. I can revert it if that's preferable.

eng/jacoco-test-coverage/pom.xml
rickle-msft: I'm not really sure of our policy surrounding beta and preview. We've been releasing everything pre-GA as preview, so I thought it would be confusing to suddenly switch to beta, but if there's some engineering systems reason why it needs to be beta now, I don't really feel strongly about it.

sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/AzureKeyVaultConfiguration.java
g2vinay: yeah, it looks like it was missed.
The release hasn't happened yet.
@hemanttanwar , lets restart the release pipeline when this merges.

eng/jacoco-test-coverage/pom.xml
hemanttanwar: After GA release we decided to use beta instead of using preview. here is the discussion on this  https://github.com/Azure/azure-sdk/pull/766

alzimmermsft: Plan moving forward is to use `beta` instead of `preview` since Maven doesn't understand `preview` and switches over to a string sort algorithm to determine the most recent release, this is why we saw cases where 12.0.0-preview.5 of Storage Blobs said it was more recent than 12.0.0. Given that this is another preview using preview again will have this show over that but in the future definitely should be using beta.

eng/jacoco-test-coverage/pom.xml
```diff
@@ -110,37 +110,37 @@
...
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-storage-queue</artifactId>
-      <version>12.1.0-beta.1</version> <!-- {x-version-update;com.azure:azure-storage-queue;current} -->
```
samvaity: @alzimmermsft This should be updated in the tracing sample files too.

sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/AzureKeyVaultConfiguration.java
JimSuplizio: Thanks for having this in your PR @rickle-msft. @g2vinay, the update script updates versions that are hard coded in source files as well, the PR that updated azure-security-keyvault-secrets should have had this file included. In the future, when you run the update script, any source files with version changes also need to be in the PR, not just the POM and README files.

PR #6638, opened 2019-12-04
Prevent HttpLoggingPolicy Consuming Body

### LINKED ISSUE #6618: HttpLoggingPolicy Eagerly Completes Hot Fluxes
`HttpLoggingPolicy` uses `FluxUtil,collectBytesInByteBufferStream` which uses [Flux.collect](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#collect-java.util.function.Supplier-java.util.function.BiConsumer-) to convert a stream of `ByteBuffer`s into a single emission `byte[]`. This will request an unbounded amount of items from the upstream upon subscription and the way `HttpLoggingPolicy` works is that it creates a logging `Mono` and then uses `.then` to continue the requests operation. This could result in a hot `Flux` being eagerly completed, potentially skewing or breaking the request being made.

Instead of using `FluxUtil,collectBytesInByteBufferStream` the logging policy should use `doOnNext` and either `doFinally` or `doOnComplete` (given the potential that we may not want to log cancelled request) which don't apply side effects onto the stream.

alzimmermsft:
Fixes #6618 

Updated `HttpLoggingPolicy` to use side-effect free operators to collect the request or response body when they will be logged. Additionally updated `BufferedHttpResponse` to buffer the `Flux<ByteBuffer>` data stream instead of collecting it as a single emission `byte[]` to replicate the stream more correctly.

### DIFF
--- a/sdk/core/azure-core/src/test/java/com/azure/core/http/policy/HttpLoggingPolicyTests.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/http/policy/HttpLoggingPolicyTests.java
@@ -44,7 +44,7 @@ import static org.junit.jupiter.api.Assertions.assertArrayEquals;
  */
 public class HttpLoggingPolicyTests {
     private static final String REDACTED = "REDACTED";
-    private static final Context context = new Context("caller-method", "HttpLoggingPolicyTests");
+    private static final Context CONTEXT = new Context("caller-method", "HttpLoggingPolicyTests");
 
     private String originalLogLevel;
     private PrintStream originalErr;
@@ -98,8 +98,8 @@ public class HttpLoggingPolicyTests {
             .httpClient(new NoOpHttpClient())
             .build();
 
-            StepVerifier.create(pipeline.send(new HttpRequest(HttpMethod.POST, requestUrl), context))
-                .verifyComplete();
+        StepVerifier.create(pipeline.send(new HttpRequest(HttpMethod.POST, requestUrl), CONTEXT))
+            .verifyComplete();
 
         String logString = new String(logCaptureStream.toByteArray(), StandardCharsets.UTF_8);
         Assertions.assertTrue(logString.contains(expectedQueryString));
@@ -149,7 +149,7 @@ public class HttpLoggingPolicyTests {
             .build();
 
         StepVerifier.create(pipeline.send(new HttpRequest(HttpMethod.POST, requestUrl, requestHeaders, stream),
-            context))
+            CONTEXT))
             .verifyComplete();
 
         String logString = new String(logCaptureStream.toByteArray(), StandardCharsets.UTF_8);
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -117,44 +117,59 @@ private HttpLoggingPolicy(HttpLogOptions httpLogOptions, boolean prettyPrintJson
...
+        long contentLength = getContentLength(logger, request.getHeaders());
+
+        if (shouldBodyBeLogged(contentType, contentLength)) {
+            ByteArrayOutputStream outputStream = new ByteArrayOutputStream((int) contentLength);
+
+            // Add non-mutating operators to the data stream.
+            request.setBody(
+                request.getBody()
+                    .doOnNext(byteBuffer -> {
+                        for (int i = byteBuffer.position(); i < byteBuffer.limit(); i++) {
```
srnagar: I am not sure how expensive this operation is going to be. Would iterating through the bytebuffer slow down the application just for logging? Would it be good to also have support for truncated logs for body - for e.g. if a flag is set to log only first few bytes of the body just to show what the payload's looking like but not the entire body.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
```diff
@@ -194,32 +209,45 @@ private HttpLoggingPolicy(HttpLogOptions httpLogOptions, boolean prettyPrintJson
...
-                responseLoggingMono = bufferedResponse.getBodyAsString().flatMap(body -> {
+        String contentTypeHeader = response.getHeaderValue("Content-Type");
+        long contentLength = getContentLength(logger, response.getHeaders());
+
+        if (shouldBodyBeLogged(contentTypeHeader, contentLength)) {
+            HttpResponse bufferedResponse = response.buffer();
+            ByteArrayOutputStream outputStream = new ByteArrayOutputStream((int) contentLength);
+            return bufferedResponse.getBody()
+                .doOnNext(byteBuffer -> {
+                    for (int i = byteBuffer.position(); i < byteBuffer.limit(); i++) {
```
srnagar: Same here. Should we have support for truncating body to log only the first few bytes?

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: I don't think there should be a large performance issue by iterating over the `ByteBuffer` using `get(int index)` given how the bulk `get(byte[] dst, int offset, int length)` method works internally. The bulk get method uses `get()` internally and has the side effect of mutating the `ByteBuffer`'s position. In terms of getting a smaller portion of the body I think that could make sense, right now body logging hard caps at 16KB and anything larger the body is ignored during logging, we could instead shrink that to 1KB or 4KB (just something small) and always log it.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: Response in your other question.

sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java
alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/6670

PR #6639, opened 2019-12-04
Returns a close handles info object for directory close handles APIs

gapra-msft:


gapra-msft:
> Changelog

Changelog should be good - the log says updated forceCloseHandles/ forceCloseAllHandles APIs without mentioning File/Directory

PR #6644, opened 2019-12-04
Changed Datalake to beta instead of preview

alzimmermsft:
Changed Datalake versioning from using preview to beta.


alzimmermsft:
Skipped waiting on management plane validation steps, this PR doesn't affect them.

PR #6655, opened 2019-12-05
Track 1 data plane dev packages

mitchdenny:
In my initial implementation of the dev packages creation I only covered client libraries. However track 1 needs this treatment as well. This doesn't cover all scenarios though as we have some repos that use the release side of our pipelines that don't participate.

So this is just a partial fix at this point. I'm going to look at the best way of servicing those libraries coming from other repos. Hypothetically they would need to have all of the versioning stuff that Jim has put together.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -240,6 +240,7 @@ jobs:
         artifact: reports
 
   - job: 'Test'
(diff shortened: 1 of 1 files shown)


mitchdenny:
Note the servicebus build here was just triggered against the branch head ref to simulate the scheduled scenario which is why it is showing up in the checks here.

mitchdenny:
Also executing core build to verify that we don't break anything on the track 2 side.

PR #6659, opened 2019-12-05
Added support for failed closed handles

gapra-msft:


sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/CloseHandlesInfo.java
```diff
@@ -22,6 +23,18 @@
      */
     public CloseHandlesInfo(Integer closedHandles) {
         this.closedHandles = closedHandles;
+        this.failedHandles = 0; // What should default be??
```
gapra-msft: Should we deprecate this?

alzimmermsft: Given that this class is generally only used by internal code but needs to be public to support it being returned I think having `failedHandles` being instantiated to 0 is a safe choice, it'll just need to be documented.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareErrorCode.java
```diff
@@ -155,7 +155,8 @@
     /**
      * Static value MultipleConditionHeadersNotSupported for ShareErrorCode.
      */
-    public static final ShareErrorCode MULTIPLE_CONDITION_HEADERS_NOT_SUPPORTED = fromString("MultipleConditionHeadersNotSupported");
+    public static final ShareErrorCode MULTIPLE_CONDITION_HEADERS_NOT_SUPPORTED
+        = fromString("MultipleConditionHeadersNotSupported");
```
alzimmermsft: Was this change due to Checkstyle issues? Given that this is codegen code I would just have the check ignore this file.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/CloseHandlesInfo.java
alzimmermsft: That is definitely another viable options to what I said in another comment.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareErrorCode.java
gapra-msft: I'll update the base branch to just ignore so I dont need to do it in every PR

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/CloseHandlesInfo.java
alzimmermsft: Pinging @JonathanGiles as this will be a common occurrence moving forward as service update and change model types, what is the strategy for handle these scenarios.

gapra-msft: Added a Javadoc detailing this

JonathanGiles: I don't entirely understand the question. Can someone please provide a bit more context?

gapra-msft: @JonathanGiles This constructor is technically just there to be used by internal code but needs to be public (since it is used in the base package). The failedHandles parameter was recently added since GA and this is the new constructor we use in internal code. 

Should we deprecate the old constructor?

hemanttanwar: One option is: We could use as a general approach for such common occurrence moving forward, is to use Fluent API approach ( using setters returns Model itself ) , ref: https://dev.to/awwsmm/build-a-fluent-interface-in-java-in-less-than-5-minutes-m7e

PR #6661, opened 2019-12-05
Move SpotBugs Analysis to Module References

### LINKED ISSUE #6660: Switch SpotBugs to Module Reference
To help support potential version skew and changes to deployment strategies SpotBugs reporting needs to be switch to module references instead of using version or source code references.

A potential solution to this issue is removing the SpotBugs aggregate report from PR validation and move SpotBugs analysis to the pom.service.xml or pom.xml level of PR validation.

alzimmermsft:
Fixes #6660 

Moves SpotBugs reporting to use module references instead of package or source code references. Update PR validation to work at the `pom.service.xml` level instead of `pom.client.xml`.


eng/pipelines/client.yml
```diff
@@ -89,9 +89,9 @@ jobs:
           goals: 'install'
 
       - task: Maven@3
-        displayName: 'Install all client libraries (for SpotBugs analysis)'
+        displayName: 'Install Module'
```
JimSuplizio: nit: I'd leave the "(for SpotBugs analysis)" in the display name

alzimmermsft: Added `(for SpotBugs analysis)` back to the name.

mitchdenny:
This is good stuff, I think the more we can scope what we are running things against the better!

eng/pipelines/client.yml
```diff
@@ -89,30 +89,17 @@ jobs:
           goals: 'install'
 
       - task: Maven@3
-        displayName: 'Install all client libraries (for SpotBugs analysis)'
+        displayName: 'Install Module and Run SpotBugs, Checkstyle, and Javadoc'
         inputs:
-          mavenPomFile: pom.client.xml
-          options: '$(DefaultOptions) -DskipTests -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true'
+          mavenPomFile: $(pomFile)
+          options: '$(DefaultOptions) -DskipTests -Dgpg.skip -Dverify-readme
```
srnagar: I am curious if `-Dverify-readme` worked because core's `pom.service.xml` just uses `pom.client.xml`. It uses relative path and it would be good to test this out with a different `pom.service.xml` like appconfig and see the logs to confirm verify-readme is looking at the right files.

alzimmermsft: Great question, I will run these changes with adding a Checkstyle, SpotBugs, Javadoc, and README verification error to validate that it will catch all of them.

alzimmermsft: Updated to work with running `pom.client.xml` and `pom.service.xml`.

alzimmermsft:
Verified that running the aggregate commands on `pom.client.xml` continue to function as they did before these changes.

alzimmermsft:
@mitchdenny @JimSuplizio @JonathanGiles @srnagar Could I get a review on the direction and changes made in this PR to validate that they are safe and move us in the direction we want to go with regards to our POM structuring and build.

If need be I will spin off a PR which changes what maven logs during build (removes logging of maven downloading files and Javadoc parsing files) so that we can allow more verbose logging during our PR validation.

eng/pipelines/client.yml
```diff
@@ -89,29 +89,16 @@ jobs:
           goals: 'install'
 
       - task: Maven@3
-        displayName: 'Install all client libraries (for SpotBugs analysis)'
+        displayName: 'Install Module and Run SpotBugs, Checkstyle, and Javadoc'
```
mitchdenny: Assume that ```eng/pipelines/client.yml``` doesn't exist (I'm actively trying to delete it :P). That means that all analysis type processes must be scoped to the libraries being build by a particular pipeline. Ideally they plug into the usual maven lifecycle.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -170,37 +170,11 @@ jobs:
...
-          goals: 'spotbugs:check checkstyle:checkstyle-aggregate revapi:check javadoc:aggregate'
-
-      # We run a separate SpotBugs aggregate report step here to roll-up all the issues identified per-module
-      - task: Maven@3
-        displayName: 'Run SpotBugs aggregate analysis'
-        inputs:
-          mavenPomFile: eng/spotbugs-aggregate-report/pom.xml
-          options: '--batch-mode -DskipTests -Dgpg.skip'
-          mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
+          mavenPomFile: sdk/${{parameters.ServiceDirectory}}/pom.service.xml
```
mitchdenny: This looks promising. Are you sure that running the install goal on the aggregate runs what you are expecting here?

alzimmermsft:
> 
> 
> This looks like it is heading in the right direction. If we can make these processes run off the aggregates then I think we are winning and it should also reduce our build times. What is the purpose of `spotbugs-aggregate-report/pom.xml` moving forward?

Moving forward `spotbugs-aggregate-report` would be removed for client SDKs, I'm leaving it around as I am uncertain on how the data SDKs leverage it. All that this POM file did was run install and given that the POM itself didn't have any additional build steps added it just ran all install steps of its parent `pom.client.xml`.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
alzimmermsft: Install runs slightly modified steps in checkstyle:check and javadoc:javadoc, these don't generate aggregates. Given that this step is ran during all PR validations I think this would be fine given that the nightly run will still run the site:site site:stage steps in reporting which are aggregates.

Any concerns with PR validation slightly changing by not running aggregates?

eng/pipelines/client.yml
alzimmermsft: Is this the change for using `-pl` and listing out the projects to install and verify instead of using `pom.service.xml`? If that is the case this should continue to function the same as all the modules listed in `pom.service.xml` are leveraging their parent POMs build and validation configurations.

alzimmermsft:
Current issue still needing to be addressed is that previously the verification of READMEs using the `embedme` tool used a regex search which ran from the root `sdk` directory in search of README.md files. This now fails when a module being built and validated doesn't have a README.md file, this is different than previous behavior. This could either be resolved by failing build due to a missing README or having a guard to prevent the check when the README doesn't exist. 

@srnagar @hemanttanwar @JimSuplizio Thoughts on the fix for this issue? I am leaning towards failing the build due to a README being missed.

srnagar:
I would suggest making README mandatory for every module. It's a good practice to have even if the module is for internal use only - the README can just document that it's internal.

alzimmermsft:
@srnagar @JonathanGiles @mitchdenny Could I get a review?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mitchdenny: Just coming back from being on leave. I don't have a concern about not running aggregate reports during PR validation. Ideally I'd like to eliminate aggregate report generation out of all pipelines.

pom.client.xml
```diff
@@ -259,6 +259,7 @@
               <failOnError>true</failOnError>
               <failOnWarnings>true</failOnWarnings>
               <doclint>all</doclint>
+              <quiet>true</quiet>
```
JonathanGiles: For other's [reference](https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#quiet), quiet "Shuts off non-error and non-warning messages, leaving only the warnings and errors appear, making them easier to view."

PR #6663, opened 2019-12-05
Added support for premium file properties

gapra-msft:


gapra-msft:
> I'm guessing another change added the additional fields into the model class?

Yep! The base branch has a new generation off the new swagger, easier to review that way

Issue #5174, opened 2019-08-29
Extend AZURE_TEST_MODE


alzimmermsft:
Currently, the AZURE_TEST_MODE environment variable only has two states which are RECORD and PLAYBACK. This is great for running locally where we will either want to RECORD new playback records or test using the records we have but this causes the nightly live tests to record records when it runs. We should investigate adding another option such as LIVE which sends requests to the service but doesn't record the requests and responses.

PR #6671, opened 2019-12-05
Add Live TestMode and Skip Recording Concepts

alzimmermsft:
Implements #5174

This PR introduces two new concepts to the Azure Core testing framework.

A new value `LIVE` was added to the `TestMode` enum. This state is used to indicate that tests should be ran against a live service without recording the network calls, this is meant to help reduce the overhead of testing by removing the recording process. The use case for this change is to remove recording tests during the nightly validation and in cases where sensitive information that hasn't been masked is being sent during testing.

A new test annotation, `DoNotRecord`, was added. This annotation is meant to be used with tests that don't want their network calls record or they make no network calls but surrounding testing infrastructure does. This annotation has an optional parameter of `skipInPlayback` which is meant to be used with tests that make network calls but won't be recorded due to the inability to redact sensitive data or they call into code paths that cannot be mocked.

### DIFF
--- a/sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestContextManager.java
+++ b/sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestContextManager.java
@@ -23,6 +23,7 @@ public class TestContextManager {
      * Constructs a {@link TestContextManager} based on the test method.
      *
      * @param testMethod Test method being ran.
+     * @param testMode The {@link TestMode} the test is running in.
      */
     public TestContextManager(Method testMethod, TestMode testMode) {
         this.testName = testMethod.getName();
--- a/sdk/core/azure-core-test/src/test/java/com/azure/core/test/FakeTestClass.java
+++ b/sdk/core/azure-core-test/src/test/java/com/azure/core/test/FakeTestClass.java
@@ -1,3 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
 package com.azure.core.test;
 
 import com.azure.core.test.annotation.DoNotRecord;


sdk/core/azure-core-test/src/main/java/com/azure/core/test/InterceptorManager.java
```diff
@@ -60,44 +65,100 @@
...
      * @param testMode The {@link TestMode} for this interceptor.
-     * @throws IOException If {@code testMode} is {@link TestMode#PLAYBACK} and an existing test session record could
-     * not be located or the data could not be deserialized into an instance of {@link RecordedData}.
+     * @throws UncheckedIOException If {@code testMode} is {@link TestMode#PLAYBACK} and an existing test session record
+     * could not be located or the data could not be deserialized into an instance of {@link RecordedData}.
      * @throws NullPointerException If {@code testName} is {@code null}.
+     * @deprecated Use {@link #InterceptorManager(String, TestMode, boolean)} instead.
      */
-    public InterceptorManager(String testName, TestMode testMode) throws IOException {
+    @Deprecated
```
srnagar: Don't need to deprecate this. `false` for `doNotRecord` is a good default.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/InterceptorManager.java
```diff
@@ -60,44 +65,100 @@
...
-     * @throws IOException An existing test session record could not be located or the data could not be deserialized
-     * into an instance of {@link RecordedData}.
+     * @param textReplacementRules A set of rules to replace text in {@link NetworkCallRecord#getResponse()} when
+     * playing back network calls.
+     * @param doNotRecord Flag indicating whether network calls should be record or played back.
+     * @throws UncheckedIOException An existing test session record could not be located or the data could not be
+     * deserialized into an instance of {@link RecordedData}.
      * @throws NullPointerException If {@code testName} or {@code textReplacementRules} is {@code null}.
      */
-    public InterceptorManager(String testName, Map<String, String> textReplacementRules) throws IOException {
```
srnagar: Removing `throws IOException` is a breaking change but this is core-test. I am not sure what is our policy for breaking changes in `core-test`.
cc @JonathanGiles

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestContextManager.java
```diff
@@ -0,0 +1,64 @@
...
+    public boolean doNotRecordTest() {
+        return doNotRecord;
+    }
+
+    /**
+     * Returns whether the current test was ran.
+     *
+     * @return Flag indicating whether the current test was ran.
+     */
+    public boolean wasTestRan() {
```
srnagar: nit: `wasTestRan` sounds odd. `didTestRun` ?

sdk/core/azure-core-test/src/main/java/com/azure/core/test/InterceptorManager.java
alzimmermsft: Is removing a throw a breaking change? I agree changing it from throwing an IOException to UncheckedIOException is breaking.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestContextManager.java
alzimmermsft: Yeah, that sounds better.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/InterceptorManager.java
alzimmermsft: I was thinking this morning whether the new overload should take this boolean or a TestContextManager object. The former is more direct on the change but the latter will allow extensibility in the future without changing the API, thoughts?

sdk/core/azure-core-test/src/main/java/com/azure/core/test/InterceptorManager.java
srnagar: Yes, removing `throws CheckedException` from a method signature is a breaking change because you can `catch` a checked exception only if something in the `try` block throws it.

sdk/core/azure-core-test/pom.xml
```diff
@@ -59,6 +59,12 @@
       <version>5.4.2</version> <!-- {x-version-update;org.junit.jupiter:junit-jupiter-engine;external_dependency} -->
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <version>5.4.2</version> <!-- {x-version-update;org.junit.jupiter:junit-jupiter-params;external_dependency} -->
```
hemanttanwar: There is more latest version `5.5.2`  , should we use that ?
https://mvnrepository.com/artifact/org.junit.jupiter/junit-jupiter-params/5.5.2

alzimmermsft: I believe there is an issue with using JUnit 5.5.* which is why we are using 5.4.2. I'm going to use 5.4.2 here since that is what we have in the external dependencies version file, if you think we should update it file an issue for discussion.

PR #6673, opened 2019-12-06
Added support for additional start copy parameters

gapra-msft:


sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/APISpec.groovy
```diff
@@ -39,7 +39,7 @@ class APISpec extends Specification {
     String connectionString
 
     // If debugging is enabled, recordings cannot run as there can only be one proxy at a time.
-    static boolean enableDebugging = false
+    static boolean enableDebugging = true
```
rickle-msft: revert

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileAPITests.groovy
```diff
@@ -424,6 +425,59 @@ class FileAPITests extends APISpec {
...
+        when:
+        SyncPoller<ShareFileCopyInfo, Void> poller = primaryFileClient.beginCopy(sourceURL, null, null,
+            PermissionCopyModeType.SOURCE, true, false, null, null)
+
+        def pollResponse = poller.poll()
+
+        then:
+        assert pollResponse.getValue().getCopyId() != null
+    }
+
```
rickle-msft: I think these tests can be parameterized and collapsed into one test. You also shouldn't need an assert statement in a then block

rickle-msft: The async tests can also be collapsed

gapra-msft: Collapsed

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareFileAsyncClient.java
```diff
@@ -241,26 +243,83 @@ public ShareServiceVersion getServiceVersion() {
...
+
+        FileSmbProperties finalSmbProperties = smbProperties == null ? new FileSmbProperties() : smbProperties;
+
+        String filePermissionKey = finalSmbProperties.getFilePermissionKey();
+
+        String fileAttributes = finalSmbProperties.setNtfsFileAttributes(FileConstants.FILE_ATTRIBUTES_NONE);
+        String fileCreationTime = FileSmbProperties.parseFileSMBDate(finalSmbProperties.getFileCreationTime());
+        String fileLastWriteTime = FileSmbProperties.parseFileSMBDate(finalSmbProperties.getFileLastWriteTime());
+
+        if (filePermissionCopyMode == PermissionCopyModeType.SOURCE) {
```
xiafu-msft: do we want to discuss when filePermissionCopyMode is null?
if if permission_key or permission_key is set and filePermissionCopyMode is null, probably we want to throw exception?

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileAPITests.groovy
```diff
@@ -424,6 +425,37 @@ class FileAPITests extends APISpec {
...
+        when:
+        SyncPoller<ShareFileCopyInfo, Void> poller = primaryFileClient.beginCopy(sourceURL, smbProperties,
+            setFilePermission ? filePermission : null, permissionType, ignoreReadOnly,
+            setArchiveAttribute, null, null)
+
+        def pollResponse = poller.poll()
+
+        then:
+        pollResponse.getValue().getCopyId() != null
```
xiafu-msft: Is it convenient to validate the SMB attributes are the same as the set ones also the acl is copied from source?
It is to validate the headers are exactly sent out from client side.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareFileAsyncClient.java
gapra-msft: In the doc it says it's an optional param

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileAPITests.groovy
gapra-msft: I can add those

Issue #6164, opened 2019-11-05
[Feature] Sync-Token HTTP implementation


mssfang:
Current App Config SDK doesn't consider Sync-Token parameter in the HTTP request.

https://msazure.visualstudio.com/Azure%20AppConfig/_git/AppConfigService?path=%2Fdocs%2Fprotocol%2Fconsistency.md&version=GBmaster

PR #6675, opened 2019-12-06
Sync token Policy implementation.

mssfang:
adding Sync-Token Policy to support eventual consistency for each request.

fixes: #6164


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * SyncToken class requires id, value and sequence number
+ */
+public final class SyncToken {
+    private final ClientLogger logger = new ClientLogger(SyncToken.class);
+
+    private String id, value;
+    private long sequenceNumber;
+    private String syncTokenString;
```
alzimmermsft: Should make these final as they shouldn't change during the lifetime of a `SyncToken`.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+        return value;
+    }
+
+    public long getSequenceNumber() {
+        return sequenceNumber;
+    }
+
+    public String getSyncTokenStringInRequest() {
+        return new StringBuilder().append(id).append("=").append(value).toString();
+    }
```
alzimmermsft: Should add Javadocs even though they are in implementation

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
alzimmermsft: These still can be made `final`

sdk/appconfiguration/azure-data-appconfiguration/src/test/resources/session-records/getConfigurationSettingNotFound.json
```diff
@@ -1,10 +1,10 @@
 {
```
conniey: If the test case itself hasn't changed, you don't need to update the records, it adds unnecessary files to the PR.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
```diff
@@ -214,7 +214,7 @@
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Mono<Response<ConfigurationSetting>> setConfigurationSettingWithResponse(ConfigurationSetting setting,
-                                                                       boolean ifUnchanged) {
+                                                                                    boolean ifUnchanged) {
```
conniey: This is squashed to the right margin, guidance would be to do an 8 space indentation rather than aligning parameters.

conniey: The same applies to the other methods below. It makes it really hard to read.

https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-136091.html#248

> If the above rules lead to confusing code or to code that's squished up against the right margin, just indent 8 spaces instead.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.data.appconfiguration.implementation;
+
+import com.azure.core.util.CoreUtils;
+import com.azure.core.util.logging.ClientLogger;
+
+/**
+ * SyncToken class requires id, value and sequence number
```
conniey: This doesn't tell me what the class does exactly or why we'd use it?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+public final class SyncToken {
+    private String id, value;
```
conniey: Why multiple variable declarations in a line?

conniey: https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141270.html#2991
> One declaration per line is recommended since it encourages commenting.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+     * @param sequenceNumber Token sequence number (version). Higher value means newer version of the same token.
+     */
+    private SyncToken(String id, String value, long sequenceNumber, String syncTokenString) {
+        this.id = id;
+        this.value = value;
+        this.sequenceNumber = sequenceNumber;
+        this.syncTokenString = syncTokenString;
+    }
+
+    static SyncToken fromSyncTokenString(String syncToken) {
```
conniey: Javadocs? What are the returns we expect?
From other static method names that parse values (ie. `Integer.parseInt`), `parseSyncToken` is a better name.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+
+package com.azure.data.appconfiguration.implementation;
+
+import com.azure.core.util.CoreUtils;
+import com.azure.core.util.logging.ClientLogger;
+
+/**
+ * SyncToken class requires id, value and sequence number
+ */
+public final class SyncToken {
```
conniey: This should have the `@Immutable` annotation.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+    public String getValue() {
+        return value;
+    }
+
+    public long getSequenceNumber() {
+        return sequenceNumber;
+    }
+
+    public String getSyncTokenStringInRequest() {
+        return new StringBuilder().append(id).append("=").append(value).toString();
```
conniey: It's more efficient to do `return id + "=" + value;` rather than creating a new object that is simply disposed of afterwards. https://stackoverflow.com/a/1532499

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public String getId() {
+        return id;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    public long getSequenceNumber() {
```
conniey: Documentation?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+import com.azure.core.util.CoreUtils;
+import com.azure.core.util.logging.ClientLogger;
+
+/**
+ * SyncToken class requires id, value and sequence number
+ */
+public final class SyncToken {
+    private String id, value;
+    private long sequenceNumber;
+    private String syncTokenString;
```
conniey: Why does this member variable exist `syncTokenString`?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+
+    static SyncToken fromSyncTokenString(String syncToken) {
+        if (CoreUtils.isNullOrEmpty(syncToken)) {
+            return null;
+        }
+        try {
+            final String[] syncTokenParts = syncToken.split(";", 2);
+            // Not a fully formed Sync-Token
+            if (syncTokenParts.length != 2) {
+                return null;
```
conniey: Log message would be nice to understand why it wasn't successfully parsed.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+
+            final String[] snParts = syncTokenParts[1].split("=", 2);
+            if (snParts.length != 2) {
+                return null;
+            }
+
+            final long sequenceNumber = Long.parseLong(snParts[1]);
+            return new SyncToken(idParts[0], idParts[1], sequenceNumber, syncToken);
+        } catch (NumberFormatException e) {
```
conniey: Generally, you want the try/catch to be limited to the method you expect this exception to be thrown. In this case, the Long.parseLong, not the entire parsing method.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,74 @@
...
+import com.azure.core.http.HttpPipelineCallContext;
+import com.azure.core.http.HttpPipelineNextPolicy;
+import com.azure.core.http.HttpResponse;
+import com.azure.core.http.policy.HttpPipelinePolicy;
+import reactor.core.publisher.Mono;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+public final class SyncTokenPolicy implements HttpPipelinePolicy {
```
conniey: Javadocs?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,74 @@
...
+    public SyncTokenPolicy() {
+        syncTokenMap = new ConcurrentHashMap<>();
+    }
+
+    /**
+     * Add or update the sync token id and value to a thread safe map.
+     *
+     * @param context request context
+     * @param next The next policy to invoke.
+     * @return
```
conniey: What does this return?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,74 @@
...
+import com.azure.core.http.HttpResponse;
+import com.azure.core.http.policy.HttpPipelinePolicy;
+import reactor.core.publisher.Mono;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+public final class SyncTokenPolicy implements HttpPipelinePolicy {
+    private static final String SYNC_TOKEN = "Sync-Token";
+    private final ConcurrentHashMap<String, SyncToken> syncTokenMap; // key is sync-token id
```
conniey: The constructor can be removed and you can simplify this by using:
```
private final ConcurrentHashMap<String, SyncToken> syncTokenMap = new ConcurrentHashMap<>();
```

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public String getValue() {
+        return value;
+    }
+
+    public long getSequenceNumber() {
+        return sequenceNumber;
+    }
+
+    public String getSyncTokenStringInRequest() {
```
conniey: I think this adds additional responsibilities to this class that is not related to a SyncToken. This should be moved to your usage in HttpPipelinePolicy.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,74 @@
...
+     * Add or update the sync token id and value to a thread safe map.
+     *
+     * @param context request context
+     * @param next The next policy to invoke.
+     * @return
+     */
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+
+        // attempt to add header to the request
```
conniey: Attempting? Are there scenarios where the header value is not added?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,74 @@
...
+            if (syncTokenValue == null) {
+                return Mono.just(httpResponse);
+            }
+
+            // Sync-Token header could have more than one value
+            final String[] syncTokens = syncTokenValue.split(",");
+            for (final String syncTokenStr : syncTokens) {
+                final SyncToken syncToken = SyncToken.fromSyncTokenString(syncTokenStr);
+                if (syncToken == null) {
+                    continue;
```
conniey: If the service returned something that could not be parsed as a sync token, we should log something.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,74 @@
...
+                }
+
+                final String tokenId = syncToken.getId();
+                final long tokenSequenceNumber = syncToken.getSequenceNumber();
```
conniey: This variable is used once, is there a reason we declared it rather than calling syncToken.getSequenceNumber()?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,74 @@
...
+
+                if (syncTokenMap.containsKey(tokenId)) {
```
conniey: You're not leveraging the advantages of using a concurrent hashmap. Between the time you check for containsKey and get(id), the item could have been removed and `existSyncToken` would return null.

sdk/appconfiguration/azure-data-appconfiguration/src/test/resources/session-records/aadAuthenticationAzConfigClient.json
```diff
@@ -4,7 +4,7 @@
     "Uri" : "https://appconfigtestshawnfang.azconfig.io/kv/newKey?api-version=1.0",
```
conniey: Remove `*.json` files where test cases have not changed. This adds unnecessary diffs to the PR.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
conniey: I would actually expect an InvalidArgumentException if something we expected to be a syncToken could not be parsed.

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -252,10 +252,9 @@
   <!-- Suppress external dependency check for jacksonwrapper overriding deserialize method -->
   <suppress checks="com.azure.tools.checkstyle.checks.ExternalDependencyExposedCheck" files="com.azure.core.implementation.serializer.jsonwrapper.jacksonwrapper.JacksonDeserializer.java"/>
 
-  <!-- Suppress ConfigurationCredentialsPolicy for HttpPipelinePolicyCheck.
-   This suppression need to be deleted after AzConfig using AAD credential.
-   Issue Link: https://github.com/Azure/azure-sdk-for-java/issues/5131 -->
+  <!-- Allow ConfigurationCredentialsPolicy and SyncTokenPolicy in Implementation folder.-->
```
conniey: Why do we allow these in the implementation folder? Since that issue is closed, I'm not sure why we have these suppressions?

conniey: This suppression could be removed by making these classes package-private in the root package. Classes in `com.azure.data.appconfiguration` do not have to be public.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,77 @@
...
+    private String syncTokenString;
+
+    /**
+     * Create an instance of SyncToken class
+     *
+     * @param id ID of token
+     * @param value Value of Token
+     * @param sequenceNumber Token sequence number (version). Higher value means newer version of the same token.
+     */
+    private SyncToken(String id, String value, long sequenceNumber, String syncTokenString) {
```
conniey: Why is this constructor private? It looks like you are jumping through hoops in your tests by creating a sync token string and parsing it when you could as easily do `new SyncToken(id, value, sequenceNumber)`. This is an implementation class, we don't expect people to new this up.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
```diff
@@ -0,0 +1,122 @@
...
+        assertEquals(secondRequestHeaderExpected, secondRequestHeader);
+
+        // verify the updated cached sync-token value
+        final HttpResponse response3 = pipeline2.send(new HttpRequest(HttpMethod.GET, new URL("http://localhost/"))).block();
+        final String thirdRequestHeader = response3.getRequest().getHeaders().getValue(SYNC_TOKEN);
+        final String thirdRequestHeaderExpected = constructSyncTokenStringWithoutSeqNumber(ID, UPDATED_VALUE) + "," + constructSyncTokenStringWithoutSeqNumber(NEW_ID, UPDATED_VALUE);
+        assertEquals(thirdRequestHeaderExpected, thirdRequestHeader);
+    }
+
+    private boolean syncTokenEquals(SyncToken syncToken, String id, String value, long sn) {
```
conniey: Why does this return boolean? If the assertions fail, it'll never hit `return true;`

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
```diff
@@ -0,0 +1,122 @@
...
+        final String secondSyncToken = constructSyncTokenString(ID, UPDATED_VALUE, UPDATED_SEQUENCE_NUMBER);
+        final HttpPipeline pipeline2 = customizedPipeline(secondSyncToken, syncTokenPolicy);
+        final HttpResponse response2 = pipeline2.send(new HttpRequest(HttpMethod.GET, new URL("http://localhost/"))).block();
+        // verify the new sync token value from the concurrent map
+        final String secondRequestHeader = response2.getRequest().getHeaders().getValue(SYNC_TOKEN);
+        final String secondRequestHeaderExpected = constructSyncTokenStringWithoutSeqNumber(ID, VALUE) + "," + constructSyncTokenStringWithoutSeqNumber(NEW_ID, UPDATED_VALUE);
+        assertEquals(secondRequestHeaderExpected, secondRequestHeader);
+
+        // verify the updated cached sync-token value
+        final HttpResponse response3 = pipeline2.send(new HttpRequest(HttpMethod.GET, new URL("http://localhost/"))).block();
```
conniey: Use StepVerifier. It was created to verify Monos and fluxes.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
```diff
@@ -0,0 +1,122 @@
...
+    @Test
+    public void parseSyncToken() {
+        String syncTokenString = constructSyncTokenString(ID, VALUE, SEQUENCE_NUMBER);
+        final SyncToken syncToken = SyncToken.fromSyncTokenString(syncTokenString);
+        syncTokenEquals(syncToken, ID, VALUE, SEQUENCE_NUMBER);
+    }
+
+    @Test
+    public void singleSyncTokenTest() throws Exception {
+        final SyncTokenPolicy syncTokenPolicy = new SyncTokenPolicy();
```
conniey: A unit test should be testing a single scenario. I can't understand what you are trying to verify because it doesn't have documentation.

Arrange, Act, and Assert was created to make unit tests more succinct and understandable.
https://docs.microsoft.com/en-us/visualstudio/test/unit-test-basics?view=vs-2019#write-your-tests

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+        final String firstSyncToken = constructSyncTokenString(ID, VALUE, SEQUENCE_NUMBER);
+        final HttpPipeline pipeline = customizedPipeline(firstSyncToken, syncTokenPolicy);
+        final HttpResponse response = pipeline.send(new HttpRequest(HttpMethod.GET, new URL("http://localhost/"))).block();
```
conniey: Step Verifier should be used... I'm not sure why you don't use Mockito to mock an HttpResponse and instead.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
```diff
@@ -0,0 +1,122 @@
...
+                    return Mono.just(new MockHttpResponse(request, 200, new HttpHeaders().put(SYNC_TOKEN, syncToken)));
+                }
+            })
+            .policies(syncTokenPolicy)
+            .build();
+    }
+
+
+    private String constructSyncTokenString(String id, String value, long sn) {
+        return String.format(id + "=" + value + ";sn=" + sn);
```
conniey: There is nothing to format. You should use simple string concatenation.

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
mssfang: Class 'ConfigurationCredentialsPolicy' implementing 'HttpPipelinePolicy' and should be a public class [HttpPipelinePolicy]

mssfang: Either way will suppress the one checkstyle rule.

sdk/appconfiguration/azure-data-appconfiguration/src/test/resources/session-records/getConfigurationSettingNotFound.json
mssfang: updated

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
mssfang: updated

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
mssfang: updated

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
mssfang: updated

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
mssfang: updated

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
mssfang: updated

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
mssfang: updated

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
```diff
@@ -0,0 +1,133 @@
...
+        String syncTokenString = constructSyncTokenString(ID, VALUE, SEQUENCE_NUMBER);
+        final SyncToken syncToken = SyncToken.parseSyncToken(syncTokenString);
+        syncTokenEquals(syncToken, ID, VALUE, SEQUENCE_NUMBER);
+    }
+
+    @Test
+    public void singleSyncTokenTest() throws Exception {
+        final SyncTokenPolicy syncTokenPolicy = new SyncTokenPolicy();
+
+        final String firstSyncToken = constructSyncTokenString(ID, VALUE, SEQUENCE_NUMBER);
```
conniey: You don't need to generate an entire pipeline to test the logic in the policy. You can mock the behaviour of things.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
mssfang: No other place uses this constructor but only the class itself. Is that making this class singleton would be more approachable?

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
mssfang: I found Mock can really test what I want to test. Maybe I am wrong. I want to test if the concurrent hashmap has updated properly. Mocking the pipeline doesn't really trigger the pipeline. When trigging the pipeline, the SyncTokenPolicy#process method runs and updates the concurrent map.

conniey: That's annoying that HttpPipelineCallContext is a final class. I'm going to ask about this.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
mssfang: I tried to make this a factory pattern to generate a new sync token.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            // Sync-Token header could have more than one value
+            final String[] syncTokens = syncTokenValue.split(",");
+            for (final String syncTokenStr : syncTokens) {
+                final SyncToken syncToken = SyncToken.parseSyncToken(syncTokenStr);
+
+                // Cannot parse sync-token skip update this sync-token to concurrent map
+                if (syncToken == null) {
+                    logger.logExceptionAsWarning(
```
alzimmermsft: Isn't this log message duplicative since any scenario where the returning sync token is null an exception would be logged by the parser? Additionally, should this be getting logged as a warning since there is nothing a customer could do to remedy this issue?

mssfang:
Please pause on reviewing this PR. 
Currently, the **async** tests of `listRevisionsWithPagination()`, `listRevisionsWithPaginationAndRepeatStream()`, `listRevisionsWithPaginationAndRepeatIterator()`,  and `listConfigurationSettingsWithPagination()`, and **sync** tests of `listRevisionsWithPagination()` and `listRevisionsWithPaginationAndRepeatStream()` are not stable and failed on my local live-run

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,118 @@
...
+ * Synchronization tokens are optional.
+ *
+ * Uses optional Sync-Token request/response headers will guarantee real-time consistency between different client
+ * instances and requests.
+ *
+ * @see <a href="https://github.com/Azure/AppConfiguration/blob/master/docs/REST/consistency.md"/>
+ */
+@Immutable
+public final class SyncToken {
+    private static final ClientLogger LOGGER = new ClientLogger(SyncToken.class);
```
conniey: Why is this a static logger?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,118 @@
...
+        // Not a fully formatted sync-token
+        if (syncTokenParts.length != 2) {
+            LOGGER.warning("Failed to parse sync token, it cannot split to two parts by delimiter ';'.");
+            return null;
+        }
+
+        final String[] idParts = syncTokenParts[0].split("=", 2);
+        // Identifier is missing a section.
+        if (idParts.length != 2) {
+            LOGGER.warning("Failed to parse sync token, it cannot split 'id=value' into two parts.");
```
conniey: Logging the actual syncToken would be useful for diagnosing problems.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Create an {@code SyncToken} by parsing a given sync-token string with only one sync-token.
+     *
+     * @param syncToken only one raw sync-token string from HTTP response header, ex.,
+     * <p>Sync-Token: {@code <id>=<value>;sn=<sn></p>}. But not
+     * <p>Sync-Token: {@code <id>=<value>;sn=<sn>,<id>=<value>;sn=<sn>}</p>
+     * @return {@code SyncToken} instance
+     */
+    static SyncToken parseSyncToken(String syncToken) {
```
conniey: Why a static factory method versus a constructor that does the parsing? I don't see any value that this adds.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+
+            // Skip sync-token updates of concurrent map if no 'Sync-Token' header
+            if (syncTokenValue == null) {
+                return Mono.just(httpResponse);
+            }
+
+            // Sync-Token header could have more than one value
+            final String[] syncTokens = syncTokenValue.split(",");
+            for (final String syncTokenStr : syncTokens) {
+                final SyncToken syncToken = SyncToken.parseSyncToken(syncTokenStr);
```
conniey: In your static method, you  return null if it's not a synctoken. you are potentially getting a null reference exception because you use it later on without checking it is not null.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
conniey: Srikanta pointed out during the testing brown bag that you can mock final classes.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
```diff
@@ -0,0 +1,137 @@
...
+import com.azure.core.test.http.NoOpHttpClient;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
+
+import java.net.URL;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class SyncTokenPolicyTest extends TestBase {
```
conniey: This doesn't need to inherit from TestBase. You aren't actually making any playback/record tests. It'll decrease the need for those empty *.json files.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
mssfang: From my point of view, it brings a meaningful method name for parsing syncToken string. So the constructor will not be performing further tasks besides just initializing fields. Functionally, it has no difference to constructor but just easier to test.

I am fine to change to just use constructor if it doesn't look good to you.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicyTest.java
mssfang: I remembered when I have no *.json file for new tests, it failed on PR pipeline complaining there is no json file for it. So a unit test will be fine without .json files? I will check it later but want to have a hint from you as well. Thanks

conniey: The Event Hubs tests don't have *.json files. We also don't extend from TestBase.

conniey: The pipeline fails because the setup step in TestBase looks for a .json file.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
conniey: How is it easier to test? If you want to use a static parse method, it should be throwing an exception if parsing a sync token is not possible rather than returning null.

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -33,20 +33,10 @@
...
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
-      <version>1.7.28</version> <!-- {x-version-update;org.slf4j:slf4j-api;external_dependency} -->
-    </dependency>
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-http-netty</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core-http-netty;current} -->
```
conniey: Why are we using current? It should be a dependency.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Due to the nature of some distributed systems real-time consistency between requests can't (or it's very hard) to be
+ * enforced implicitly. A solution is to allow protocol support in the form of multiple Synchronization Tokens.
+ * Synchronization tokens are optional.
+ * <p>
+ * Uses optional Sync-Token request/response headers will guarantee real-time consistency between different client
+ * instances and requests.
+ *
+ * @see <a href="https://github.com/Azure/AppConfiguration/blob/master/docs/REST/consistency.md"/>
```
conniey:
```suggestion
 * @see <a href="https://github.com/Azure/AppConfiguration/blob/master/docs/REST/consistency.md">Real-time Consistency</a>
```

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,96 @@
...
+     * <p>Sync-Token: {@code <id>=<value>;sn=<sn></p>}. But not
+     * <p>Sync-Token: {@code <id>=<value>;sn=<sn>,<id>=<value>;sn=<sn>}</p>
+     * Allows for better concurrency and client cache ability. The client may choose to use only token's last version,
+     * since token versions are inclusive. Not required for requests.
+     */
+    SyncToken(String syncToken) {
+        String localId = null;
+        String localValue = null;
+        Long localSequenceNumber = null;
+        if (!CoreUtils.isNullOrEmpty(syncToken)) {
```
conniey: Shouldn't we throw if it is empty rather than constructing an invalid object.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,96 @@
...
+ * Synchronization tokens are optional.
+ * <p>
+ * Uses optional Sync-Token request/response headers will guarantee real-time consistency between different client
+ * instances and requests.
+ *
+ * @see <a href="https://github.com/Azure/AppConfiguration/blob/master/docs/REST/consistency.md"/>
+ */
+@Immutable
+public final class SyncToken {
+    private final ClientLogger logger = new ClientLogger(SyncToken.class);
```
srnagar: ClientLogger is never used in this class.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,96 @@
...
+                        localSequenceNumber = Long.parseLong(snParts[1]);
+                        localId = idParts[0];
+                        localValue = idParts[1];
+                    } catch (NumberFormatException ex) {
+                    }
+                }
+            }
+        }
+
+        this.sequenceNumber = localSequenceNumber;
```
srnagar: Adding a log in the constructor when you are able to successfully parse the syncToken will help debugging. Log when parsing fails too (ClientLogger will then be used )

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,96 @@
...
+        this.value = localValue;
+        this.id = localId;
+    }
+
+    /**
+     * Get Token ID (opaque).
```
srnagar: What does opaque mean? Does it have to be part of the javadoc?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    SyncToken(String syncToken) {
+        String localId = null;
+        String localValue = null;
+        Long localSequenceNumber = null;
+        if (!CoreUtils.isNullOrEmpty(syncToken)) {
+            final String[] syncTokenParts = syncToken.split(";", 2);
+            if (syncTokenParts.length == 2) {
+                String[] idParts = syncTokenParts[0].split("=", 2);
+                String[] snParts = syncTokenParts[1].split("=", 2);
```
srnagar: Use string constants for token separators.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * A policy uses a concurrent map to maintain the latest sync-tokens. When this HTTP pipeline policy is triggered, the
+ * policy will retrieve all sync-tokens without sequence number segment from the concurrent map and use it in the HTTP
+ * request. Also after received the HTTP response, update the latest sync-tokens to the map.
+ */
+public final class SyncTokenPolicy implements HttpPipelinePolicy {
+    private static final String COMMA = ",";
+    private static final String EQUAL = "=";
+    private static final String SYNC_TOKEN = "Sync-Token";
+    private final ConcurrentHashMap<String, SyncToken> syncTokenMap = new ConcurrentHashMap<>(); // key is sync-token id
```
srnagar:
```suggestion
    private final Map<String, SyncToken> syncTokenMap = new ConcurrentHashMap<>(); // key is sync-token id
```

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,77 @@
...
+            // Sync-Token header could have more than one value
+            final String[] syncTokens = syncTokenValue.split(COMMA);
+            for (final String syncTokenStr : syncTokens) {
+                final SyncToken syncToken = new SyncToken(syncTokenStr);
+                final String tokenId = syncToken.getId();
+                // If the value is not thread safe and must be updated inside the method with a remapping function
+                // to ensure the entire operation is atomic.
+                syncTokenMap.compute(tokenId, (key, existingSyncToken) -> {
+                    if (existingSyncToken == null
+                        || syncToken.getSequenceNumber() > existingSyncToken.getSequenceNumber()) {
```
srnagar: This can throw NPE if `syncToken.getSequenceNumber()` returns null (when the syncToken string is invalid and can't be parsed)

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
mssfang: Used to match the Tracing changes. Will revert back to 1.2.0

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncTokenPolicy.java
```diff
@@ -0,0 +1,77 @@
...
+
+            // Skip sync-token updates of concurrent map if no 'Sync-Token' header
+            if (syncTokenValue == null) {
+                return Mono.just(httpResponse);
+            }
+
+            // Sync-Token header could have more than one value
+            final String[] syncTokens = syncTokenValue.split(COMMA);
+            for (final String syncTokenStr : syncTokens) {
+                final SyncToken syncToken = new SyncToken(syncTokenStr);
```
conniey: I would be more defensive about generating this...
```java
if (String.isNullOrEmpty(syncTokenString)) {
    continue;
}

final SyncToken syncToken;
try {
    syncToken = new SyncToken(syncTokenString);
} catch (IllegalArgumentException e) {
    logger.warning("Skipping invalid sync token '{}'.", syncTokenString);
    continue;
}
```

conniey: And since this token is generated from the service, we'd expect an IllegalArgumentException to be a sign that either the service has given us something wrong, or a spec has changed.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
mssfang: if the `syncToken` is invalid, we don't want to throw because the user has no control of sync token do anything to correct it. The invalid 'syncToken' will be ignored. I can log the message as a warning but still, the user can't control it. What do you think? logging for every invalid case of 'syncToken'. @conniey @srnagar

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
```diff
@@ -97,85 +81,6 @@ public void nullAADCredential() {
         });
     }
```
conniey: Is there a reason we removed these test cases? You can still test it, but you wouldn't be generating an actual client.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
conniey: https://github.com/Azure/azure-sdk-for-java/pull/6675/files#r373652294

conniey: I think you're adding additional responsibilities to this class. A syncToken class only cares about itself. It doesn't know that it lives in a pipeline policy. So if it gets invalid data, it should throw.

The pipeline policy, because it understands the response comes from the service, can handle it.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
mssfang: https://github.com/Azure/azure-sdk-for-java/pull/6675/files/ede705f56a3d69c3eed3765ea1769556f44874e2#diff-2705d6620e99b076f6e0a81319031b5bR671

Abstract the test into Client test base.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/SyncToken.java
mssfang: It means the user doesn't require to know what token id will be. https://msazure.visualstudio.com/Azure%20AppConfig/_git/AppConfigService?path=%2Fdocs%2Fprotocol%2Fconsistency.md&version=GBmaster&_a=preview

PR #6683, opened 2019-12-06
AsyncReadBenchmark should use v4 API e2e

moderakh:
Our benchmark tool has been wired up against the internal API of `AsyncDocumentClient.readDocument(.)` rather than the v4 API `cosmosContainer.getItem().read()`. Therefore the benchmark tool doesn't provide the true end 2 end perf numbers.

This PR changes AsyncReadBenchmark to use v4 API e2e (i.e, `cosmosContainer.getItem().read()`) to provide a correct perf number of read throughput e2e

### DIFF
--- a/sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncBenchmark.java
+++ b/sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncBenchmark.java
@@ -4,7 +4,6 @@
 package com.azure.cosmos.benchmark;
 
 import com.azure.cosmos.BridgeInternal;
-import com.azure.cosmos.BridgeInternalBenchmark;
 import com.azure.cosmos.CosmosAsyncClient;
 import com.azure.cosmos.CosmosBridgeInternal;
 import com.azure.cosmos.CosmosClientBuilder;


sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncBenchmark.java
```diff
@@ -56,15 +60,17 @@
...
+        v4Client = new CosmosClientBuilder()
+            .setEndpoint(cfg.getServiceEndpoint())
+            .setKey(cfg.getMasterKey())
+            .setConnectionPolicy(cfg.getConnectionPolicy())
+            .setConsistencyLevel(cfg.getConsistencyLevel())
+            .buildAsyncClient();
 
         logger = LoggerFactory.getLogger(this.getClass());
 
+        client =  BridgeInternalBenchmark.getInternalDocumentClient(v4Client);
```
kushagraThapar: If we want to use CosmosClient, then why bother getting internal `AsyncDocumentClient` ?
Shouldn't we change all the operations to use `v4Client` and remove `AsyncDocumentClient` completely from benchmark. I thought that is the goal for this PR.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncReadBenchmark.java
```diff
@@ -48,24 +48,25 @@ protected void hookOnError(Throwable throwable) {
 
     AsyncReadBenchmark(Configuration cfg) {
         super(cfg);
+        cosmosAsyncContainer = v4Client.getDatabase(cfg.getDatabaseId()).getContainer(cfg.getCollectionId()).read().block().getContainer();
```
kushagraThapar: Why are we reading the container again here ?
When calling `super(cfg)`, this will read the container in `AsyncBenchmark` constructor on line 75. 
If we change AsyncBenchmark to completely use `v4Client` then we can reuse the container read in the parent class. Is this PR only reflects partial changes for changing the clients, are there going to be more PRs which will handle the complete set of changes?

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncBenchmark.java
moderakh: there are many different benchmarks ReadMyWrites, Write, etc. The goal of this PR is to just make the change for the `AsyncReadBenchmark`. If we change client vs v4Client this will become a cascading change to all other benchmarks which we don't want. as the scope of this PR is only `AsyncReadBenchmark` and `AsyncReadBenchmark` internally now uses v4Client for reads.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/BridgeInternalBenchmark.java
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.cosmos;
+
+import com.azure.cosmos.implementation.AsyncDocumentClient;
+
+public class BridgeInternalBenchmark {
```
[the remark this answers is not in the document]
moderakh: Good point, didn't know we have that. will address in the next commit.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncReadBenchmark.java
moderakh: the parent class and other benchmarks are using documentClient. This PR will only change the AsyncReadBenchmark. The other benchmarks will be updated later, specifically we also have query api changes coming. So all those other benchmarks will be updated later once the surface API is frozen.

and on line 75 of the parent it reads the collection not container. here we need container.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncBenchmark.java
kushagraThapar: Makes sense.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncReadBenchmark.java
kushagraThapar: Sounds good for now. 
Can we put a TODO before this container read statement specifying `Get rid of this container read once all the benchmark uses v4Client`
Because once we change all the benchmarks, we should make sure not to re-read the container, since at that point, we have had already read the container through v4 client in parent class. 
Makes sense ?

moderakh: good to have. comment added.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncReadBenchmark.java
```diff
@@ -48,24 +48,26 @@ protected void hookOnError(Throwable throwable) {
...
         int index = (int) (i % docsToRead.size());
-        RequestOptions options = new RequestOptions();
-        options.setPartitionKey(new PartitionKey(docsToRead.get(index).getId()));
+        Document doc = docsToRead.get(index);
 
-        Flux<ResourceResponse<Document>> obs = client.readDocument(getDocumentLink(docsToRead.get(index)), options);
+        String partitionKeyValue = doc.getId();
+        Mono<CosmosAsyncItemResponse> result = cosmosAsyncContainer.getItem(doc.getId(), partitionKeyValue).read();
 
         concurrencyControlSemaphore.acquire();
```
kirankumarkolli: Sync vs async we need to baseline its impact.

moderakh: sure on my list.

PR #6685, opened 2019-12-06
Update Versions of Released Libraries

alzimmermsft:
Updates the current versions of released Azure SDKs.


sdk/core/azure-core-management/pom.xml
```diff
@@ -57,7 +57,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core;current} -->
+      <version>1.2.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;current} -->
```
alzimmermsft: @anuchandy Should this have a current version dependency on Azure Core?

sdk/core/azure-core-amqp/pom.xml
```diff
@@ -57,7 +57,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core;current} -->
+      <version>1.2.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;current} -->
```
alzimmermsft: @srnagar Does this need a current version dependency on Azure Core?

sdk/keyvault/azure-security-keyvault-secrets/pom.xml
```diff
@@ -11,7 +11,7 @@
 
   <groupId>com.azure</groupId>
   <artifactId>azure-security-keyvault-secrets</artifactId>
-  <version>4.0.1</version> <!-- {x-version-update;com.azure:azure-security-keyvault-secrets;current} -->
+  <version>4.1.0-beta.1</version> <!-- {x-version-update;com.azure:azure-security-keyvault-secrets;current} -->
```
g2vinay: why is the core dependency nit updated to the 1.2.0-beta.1 for SDK poms everywhere ?

alzimmermsft: They are marked as using `dependency` instead of `current`. I only updated `current` to bump the versions of the pom files.

sdk/core/azure-core-amqp/pom.xml
weshaggard: Hopefully it doesn't.

sdk/keyvault/azure-security-keyvault-secrets/pom.xml
weshaggard: In generally they shouldn't be updated automatically. We should make a decision to bump them if we want to rebaseline to the higher version of core.

JimSuplizio:
@alzimmermsft you missed com.azure:azure-security-keyvault-keys which released 4.0.1 on Friday. Also, why are none of the other java files with versions updated? The following files have versions hard coded in them that should have been updated with this PR. I've created another PR this morning to remedy this.
- sdk/keyvault/azure-security-keyvault-certificates/src/main/java/com/azure/security/keyvault/certificates/AzureKeyVaultConfiguration.java
- sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/AzureKeyVaultConfiguration.java
- sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/BlobCryptographyConfiguration.java
- sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BuilderHelper.java
- sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/FileConfiguration.java
- sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/implementation/util/BuilderHelper.java
- sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/QueueConfiguration.java
- sdk/storage/azure-storage-queue/src/main/java/com/azure/storage/queue/implementation/util/BuilderHelper.java

PR #6687, opened 2019-12-06
Support setting tenant ID on shared token cache credential

jianghaolu:


sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredential.java
```diff
@@ -36,7 +36,7 @@
     DefaultAzureCredential(IdentityClientOptions identityClientOptions) {
         super(new ArrayDeque<>(Arrays.asList(new EnvironmentCredential(identityClientOptions),
             new ManagedIdentityCredential(null, identityClientOptions),
-            new SharedTokenCacheCredential(null, "04b07795-8ddb-461a-bbee-02f9e1bf7b46",
+            new SharedTokenCacheCredential(null, null, "04b07795-8ddb-461a-bbee-02f9e1bf7b46",
```
conniey: Is 04b07795-8ddb-461a-bbee-02f9e1bf7b46 some root tenantId?

jianghaolu: It's the tenant of Azure CLI - the only first party app we (kind of) own. Without a first party app, users will not be able to authenticate from an arbitrary tenant.

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredentialBuilder.java
```diff
@@ -8,21 +8,9 @@
  *
  * @see SharedTokenCacheCredential
  */
-public class SharedTokenCacheCredentialBuilder extends CredentialBuilderBase<ManagedIdentityCredentialBuilder> {
-    private String clientId;
+public class SharedTokenCacheCredentialBuilder extends AadCredentialBuilderBase<SharedTokenCacheCredentialBuilder> {
```
srnagar: Change in parent class can be backward incompatible but in this case since AadCredentialBuilderBase is also a type of CredentialBuilderBase, it might be okay but good to get @JonathanGiles to take a look.

jianghaolu: Yes this is to add `tenantId()` method to this builder through AadCredentialBuilderBase.

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
```diff
@@ -41,31 +43,44 @@
...
 
         if (username == null) {
             this.username = configuration.get(Configuration.PROPERTY_AZURE_USERNAME);
         } else {
             this.username = username;
         }
-
-        this.clientId = clientId;
+        if (clientId == null) {
+            this.clientId = configuration.get(Configuration.PROPERTY_AZURE_CLIENT_ID);
```
g2vinay: We're giving priority to env config over the values passed through the builder ?

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
```diff
@@ -41,31 +43,44 @@
      * @param clientId the client ID of the application
      * @param identityClientOptions the options for configuring the identity client
      */
-    SharedTokenCacheCredential(String username, String clientId, IdentityClientOptions identityClientOptions) {
+    SharedTokenCacheCredential(String username, String clientId, String tenantId,
```
chlowell: I don't see `tenantId` used to filter accounts, does it magically happen elsewhere?

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredentialBuilder.java
```diff
@@ -8,21 +8,9 @@
...
     private String username;
 
-    /**
-     * Sets the client ID for the application.
-     *
-     * @param clientId The client ID for the application.
-     *
-     * @return The updated SharedTokenCacheCredentialBuilder object.
-     * */
```
g2vinay: I am assuming this comes from AadCredentialBuilderBase now.

jianghaolu: Yes

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
jianghaolu: Correct. This tenant ID only appears on the authority URL. The tenant ID info in the homeAccountId field is always going to be the home tenant ID for the account. You can get an access token for this tenant even if the refresh token is acquired for a different tenant (home tenant) - this is not a bug. The bug that we reported to MSAL is that you can get an access token for this tenant *even if* you don't have access to this tenant.

Since the tenant ID in the homeAccountId in the token cache is always going to be the home tenant, there's no filtering to do there.

Issue #6069, opened 2019-10-29
[BUG] Exception when Cosmos DB HTTP response header is larger than 8192 bytes


f-paulus:
**Describe the bug**
When the continuation token in the header of the HTTP response from the Cosmos DB is larger than 8192 bytes, the netty that is used by the SDK throws an exception. It seems that this was already fixed for SDK versions <3.0 but the changes were not ported to the SDK 3.x: https://github.com/Azure/azure-cosmosdb-java/issues/24

***Exception or Stack Trace***
```
io.netty.handler.codec.TooLongFrameException: HTTP header is larger than 8192 bytes.
	at io.netty.handler.codec.http.HttpObjectDecoder$HeaderParser.newException(HttpObjectDecoder.java:829)
	at io.netty.handler.codec.http.HttpObjectDecoder$HeaderParser.process(HttpObjectDecoder.java:821)
	at io.netty.buffer.AbstractByteBuf.forEachByteAsc0(AbstractByteBuf.java:1306)
	at io.netty.buffer.AbstractByteBuf.forEachByte(AbstractByteBuf.java:1286)
	at io.netty.handler.codec.http.HttpObjectDecoder$HeaderParser.parse(HttpObjectDecoder.java:793)
	at io.netty.handler.codec.http.HttpObjectDecoder.readHeaders(HttpObjectDecoder.java:592)
	at io.netty.handler.codec.http.HttpObjectDecoder.decode(HttpObjectDecoder.java:218)
	at io.netty.handler.codec.http.HttpClientCodec$Decoder.decode(HttpClientCodec.java:202)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:505)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:444)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:283)
	at io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:253)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:374)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:360)
[...]

PR #6690, opened 2019-12-07
Added response decoder configuration to fix max header size issue

kushagraThapar:
* Added response decoder configuration to fix max header size issue: https://github.com/Azure/azure-sdk-for-java/issues/6069
* Added other response configurations in line with V2

### DIFF
--- a/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/ReactorNettyClient.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/ReactorNettyClient.java
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 package com.azure.data.cosmos.internal.http;
 
+import com.azure.data.cosmos.internal.Configs;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelOption;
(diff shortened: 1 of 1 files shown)


kushagraThapar:
> LGTM. I like small PRs. thank you.
> 
> Just two questions:
> 
> 1. does `. httpResponseDecoder(.)` overwrite the existing http decoder or does it add a new decoder in addition to the existing one?
> 2. I am curious how did you manage to test this?

1. `httpResponseDecoder(.)` adds a new response decoder with provided options. Works as a builder pattern internally. That's why we are doing it once in constructor. 

2. I have tried and haven't been able to test it yet. I am working with folks from backend and query team to reproduce this. The reason to create this PR is to make sure it doesn't have any side effects on http client. Will only merge it after testing it thoroughly.

kushagraThapar:
> > LGTM. I like small PRs. thank you.
> > Just two questions:
> > 
> > 1. does `. httpResponseDecoder(.)` overwrite the existing http decoder or does it add a new decoder in addition to the existing one?
> > 2. I am curious how did you manage to test this?
> 
> 1. `httpResponseDecoder(.)` adds a new response decoder with provided options. Works as a builder pattern internally. That's why we are doing it once in constructor.
> 2. I have tried and haven't been able to test it yet. I am working with folks from backend and query team to reproduce this. The reason to create this PR is to make sure it doesn't have any side effects on http client. Will only merge it after testing it thoroughly.

@desh901 helped with the testing, we are good to go for merge: https://github.com/Azure/azure-sdk-for-java/issues/6069#issuecomment-564933308

Issue #6532, opened 2019-11-25
[Feature] Support SettingSelector literal special character and wild card functions


mssfang:
Support special characters in SettingSetting:
For example, by giving "ab*c"
(1) Literal 
It should treat it as "ab\*c", as the key
(2) Wild Card
All keys that have prefix "ab" and suffix "c" should be valid keys

PR #6693, opened 2019-12-09
Setting selector filter 

mssfang:
Let the users to define their wildcard, we are no longer takes a list of keys, or a list of labels. Instead of, we take a string value that defined by user.

More details: https://github.com/Azure/azure-sdk/issues/842

fixes: #6532

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelector.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelector.java
@@ -4,12 +4,11 @@
 package com.azure.data.appconfiguration.models;
 
 import com.azure.core.annotation.Fluent;
-import com.azure.data.appconfiguration.ConfigurationAsyncClient;
 import com.azure.core.util.CoreUtils;
+import com.azure.data.appconfiguration.ConfigurationAsyncClient;
 
 import java.time.OffsetDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.Objects;
 
 /**
  * A set of options for selecting configuration settings from App Configuration service.


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelector.java
```diff
@@ -50,16 +50,16 @@ public SettingSelector() {
...
      * Examples:
      * <ol>
-     * <li>If keys = "*", settings with any key are returned.</li>
-     * <li>If keys = "abc1234", settings with a key equal to "abc1234" are returned.</li>
-     * <li>If keys = "abc*", settings with a key starting with "abc" are returned.</li>
-     * <li>If keys = "*abc*", settings with a key containing "abc" are returned.</li>
+     * <li>If keyFilter = "*", settings with any key are returned.</li>
+     * <li>If keyFilter = "abc1234", settings with a key equal to "abc1234" are returned.</li>
+     * <li>If keyFilter = "abc*", settings with a key starting with "abc" are returned.</li>
+     * <li>If keyFilter = "*abc*", settings with a key containing "abc" are returned.</li>
```
srnagar: Also, add examples for multiple filters that are comma-separated.

sdk/appconfiguration/azure-data-appconfiguration/src/test/resources/session-records/addConfigurationSettingNullKey.json
```diff
@@ -1,10 +1,10 @@
 {
```
conniey: Don't update *.json files if you aren't modifying the test case. This creates unnecessary diffs and additions to PRs.

mssfang: After chagne to PLAYBACK mode with unchanged json files, it almost failed all tests beside testAzureConfiguration and AadCredentialTest. The existing json files replacement will be needed in this case. So i will leave the PR unchanged.

mssfang: I tried to revert back to unchanged JSON files, but it failed almost all tests besides AadCredentialTest and testAzureConfiguration two tests. So I will still stay what it is in this case.

sdk/appconfiguration/azure-data-appconfiguration/src/test/resources/session-records/setConfigurationSettingNullKey.json
```diff
@@ -1,10 +1,10 @@
 {
   "networkCallRecords" : [ {
     "Method" : "GET",
-    "Uri" : "https://appconfigtestshawnfang.azconfig.io/kv?key=key94864%2a&api-version=1.0",
+    "Uri" : "https://appconfigtestshawnfang.azconfig.io/kv?key=key87027%2a&label=&api-version=1.0",
```
conniey: `&label=` is this the correct behaviour? Having a label with no key?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelector.java
```diff
@@ -93,42 +94,42 @@ public SettingSelector setKeys(String... keys) {
...
+     * <li>If {@code labelFilter = "*abc*"}, settings with a label containing "abc" are returned.</li>
+     * <li>If {@code labelFilter = "abc,def"}, settings with labels "abc" or "def" are returned.</li>
      * </ul>
      *
      * @return labels The labels used to filter GET requests from the service.
      */
-    public String[] getLabels() {
-        return labels == null ? new String[0] : CoreUtils.clone(labels);
+    public String getLabelFilter() {
+        return labelFilter == null ? "" : labelFilter;
```
conniey: An empty string seems different than a collection with no items in it. Is this correct? It may account for why those JSON files are breaking.

mssfang: Thanks. It should be the reason

conniey: Is this the correct behaviour, is what I am wondering

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelector.java
```diff
@@ -49,17 +49,18 @@ public SettingSelector() {
...
+     * <li>If {@code keyFilter = "*abc*"}, settings with a key containing "abc" are returned.</li>
+     * <li>If {@code keyFilter = "abc,def"}, settings with a key equal to "abc" or "def" are returned.</li>
+     * </ul>
      *
      * @return The expressions to filter ConfigurationSetting keys on.
      */
-    public String[] getKeys() {
-        return keys == null ? new String[0] : CoreUtils.clone(keys);
+    public String getKeyFilter() {
+        return keyFilter == null ? "" : keyFilter;
```
conniey: Is this correct then?

mssfang: Let me double check with the feature team. The label is optional so it could be null. But the key should be required.

conniey: In this case, the behaviour is different because you returned an empty set before (nothing in it). But now, you're returning an empty string. Is it possible to have a key with an empty string?

mssfang: You are right. https://github.com/Azure/AppConfiguration/blob/master/docs/REST/kv.md

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelector.java
```diff
@@ -68,18 +70,19 @@ public SettingSelector() {
...
      * </ul>
      *
-     * @param keys The expressions to filter ConfigurationSetting keys on.
+     * @param keyFilter The expressions to filter ConfigurationSetting keys on.
      * @return The updated SettingSelector object
      */
-    public SettingSelector setKeys(String... keys) {
-        this.keys = keys;
+    public SettingSelector setKeyFilter(String keyFilter) {
+        Objects.requireNonNull(keyFilter);
```
conniey: Add a message when using Objects.requireNonNull or it'll throw a NullPointer with no indication of what happened.
Also, update the javadocs with `@throws`.

conniey: Actually, looking at our SDK guidelines, we shouldn't be doing this verification at all and allowing the service to return a bad request.
https://azure.github.io/azure-sdk/general_implementation.html#parameter-validation

mssfang: Ahh. That is correct. Thanks

PR #6735, opened 2019-12-10
AutoRest: update autorest.java cmd

xseeseesee:


eng/mgmt/gulpfile.js
```diff
@@ -39,9 +39,10 @@ gulp.task('default', function() {
     console.log("\tThe version of AutoRest. E.g. 2.0.9, or the location of AutoRest repo, e.g. E:\\repo\\autorest");
 
     console.log("--autorest-java");
-    console.log("\tPath to an autorest.java generator to pass as a --use argument to AutoRest.");
+    console.log("\tOption#1: Path to an autorest.java generator to pass as a --use argument to AutoRest.");
+    console.log("\t\tOption#2: The version of AutoRest.Java. E.g. 2.0.9. You can also pass latest or preview.");
```
ChenTanyi: I think option should use same number of `'\t'`

ChenTanyi: How about use another parameter `--autorest-java-version` rather than use same parameter?

eng/mgmt/gulpfile.js
```diff
@@ -107,9 +112,10 @@ var codegen = function(project, cb) {
 
     console.log('Generating "' + project + '" from spec file ' + specRoot + '/' + mappings[project].source);
 
-    const generatorPath = args['autorest-java']
-        ? `--use=${path.resolve(args['autorest-java'])} `
-        : '';
+    const generatorPath = autoRestJavaVersion == 'preview' || autoRestJavaVersion == 'latest' 
+		|| autoRestJavaVersion.match(/[0-9]+\.[0-9]+\.[0-9]+.*/)
```
ChenTanyi: This could match some path include version, like ~/test-1.2.3/***. May use ^ at first, but I think split into two parameters is better

eng/mgmt/gulpfile.js
```diff
@@ -135,6 +141,7 @@ var codegen = function(project, cb) {
                         ' --azure-arm ' +
                         ' --azure-libraries-for-java-folder=' + sdkRoot + ' ' +
                         ` --license-header=MICROSOFT_MIT_NO_CODEGEN ` +
+                        ` --use=@microsoft.azure/autorest.java@` + autoRestJavaVersion +` `
```
ChenTanyi: This is conflicted with generatorPath

eng/mgmt/gulpfile.js
weidongxu-microsoft: @xccc-msft please check tanyi's suggestion. I think use ^ would be fine.

If possible also print out the resolution of this line, if not already there. Then if anything went wrong caller would have log to see why.

PR #6746, opened 2019-12-10
Add changelog to artifacts if it exists

JimSuplizio:
If the changelog.md file isn't published as part of the artifact then only the commit message from the SHA is used for the release notes. If the changelog.md file is one of the published artifacts then its contents will be used for the release notes instead of the commit message from the release SHA.

With this change, if there's a changelog.md file for a given artifact it'll copy it to the released artifacts. Note: There's other verification that'll be ensuring that client track libraries have changelogs and that is not part of this PR. This only copies them if they exist.

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -1036,6 +1036,39 @@
     </profile>
 
     <profile>
+      <id>changelog</id>
+      <activation>
+        <file>
+          <exists>${basedir}/CHANGELOG.md</exists>
+        </file>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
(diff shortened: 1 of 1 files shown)


pom.client.xml
```diff
@@ -1035,6 +1035,39 @@
       </build>
     </profile>
 
+    <profile>
+      <id>changelog</id>
+      <activation>
+        <file>
+          <exists>${basedir}/CHANGELOG.md</exists>
```
JimSuplizio: ${project.basedir} isn't valid in the <activation> tags, only ${basedir}

PR #6748, opened 2019-12-10
Pretty print the exception error message

sima-zhu:


sdk/core/azure-core/src/main/java/com/azure/core/util/CoreUtils.java
```diff
@@ -183,4 +197,44 @@ public static boolean isNullOrEmpty(CharSequence charSequence) {
...
+
+    /**
+     * Pretty print the json or xml string content. If the content is not in a format of json or xml,
+     * return original string.
+     *
+     * @param content The body content which need to parse.
+     * @param contentType The format of the contents.
+     * @return Pretty json or xml format of the content. If it is not in a format of json or xml, returns original one.
+     */
+    public static String printPrettyFormatJsonOrXml(String content, String contentType) {
```
JonathanGiles: This shouldn't be a public API. It should go into an implementation package utility location.

sdk/core/azure-core-test/src/test/java/com/azure/core/test/implementation/RestProxyWithMockTests.java
```diff
@@ -219,7 +221,7 @@ public void serviceErrorWithResponseContentTypeBadJSON() {
             fail();
         } catch (HttpResponseException ex) {
             assertContains(ex.getMessage(), "Status code 200");
-            assertContains(ex.getMessage(), "\"BAD JSON\"");
+            assertContains(ex.getMessage(), "BAD JSON");
```
JonathanGiles: Where does 'BAD JSON' come from?

sima-zhu: The rest proxy test mock some service response. This is coming from the mock service call. 
```
        ServiceErrorWithCharsetService service = RestProxy.create(
                ServiceErrorWithCharsetService.class,
                new HttpPipelineBuilder().httpClient(new SimpleMockHttpClient() {
                    @Override
                    public Mono<HttpResponse> send(HttpRequest request) {
                        HttpHeaders headers = new HttpHeaders().put("Content-Type", "application/json");

                        HttpResponse response = new MockHttpResponse(request, 200, headers, "BAD JSON".getBytes(StandardCharsets.UTF_8));
                        return Mono.just(response);
                    }
                }).build());
```

sdk/core/azure-core/src/main/java/com/azure/core/util/CoreUtils.java
sima-zhu: Will create an ImplUtil under implementation pkg and put the function there.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/RestProxy.java
```diff
@@ -346,7 +347,8 @@ private static Exception instantiateUnexpectedException(UnexpectedExceptionInfor
         if ("application/octet-stream".equalsIgnoreCase(contentType)) {
             bodyRepresentation = "(" + httpResponse.getHeaderValue("Content-Length") + "-byte body)";
         } else {
-            bodyRepresentation = responseContent.isEmpty() ? "(empty body)" : "\"" + responseContent + "\"";
+            responseContent = ImplUtil.printPrettyFormatJsonOrXml(responseContent, contentType);
```
JonathanGiles: You should only pay the performance penalty of pretty printing the response content in the exception case. Otherwise I would leave it as-is.

sdk/core/azure-core/src/main/java/com/azure/core/http/ContentType.java
```diff
@@ -12,6 +12,11 @@
      */
     public static final String APPLICATION_JSON = "application/json";
 
+    /**
+     * the default JSON Content-Type header.
```
alzimmermsft: This should state XML right? XML isn't JSON :)

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ImplUtil.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.sax.SAXSource;
+import javax.xml.transform.sax.SAXTransformerFactory;
+import javax.xml.transform.stream.StreamResult;
+
+/**
+ * This class contains utility methods useful for internal use.
+ */
+public class ImplUtil {
```
alzimmermsft: I know this is a utility in implementation but I'm not a fan of calling this `ImplUtil`. This feels like it should go into `LoggingUtil` given that this is used to format logging messages, that is also in implementation so it will be hidden from public API.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ImplUtil.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Pretty print the json or xml string content. If the content is not in a format of json or xml,
+     * return original string.
+     *
+     * @param content The body content which need to parse.
+     * @param contentType The format of the contents.
+     * @return Pretty json or xml format of the content. If it is not in a format of json or xml, returns original one.
+     */
+    public static String printPrettyFormatJsonOrXml(String content, String contentType) {
```
alzimmermsft: Should `HttpLoggingPolicy` call into this method instead of having its own implementation of pretty printing JSON?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ImplUtil.java
```diff
@@ -0,0 +1,66 @@
...
+package com.azure.core.implementation;
+
+import com.azure.core.http.ContentType;
+import com.azure.core.util.logging.ClientLogger;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
+import org.xml.sax.InputSource;
```
alzimmermsft: Does this import require any additional Maven dependencies in our library?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/ImplUtil.java
```diff
@@ -0,0 +1,66 @@
...
+            try {
+                Transformer serializer = SAXTransformerFactory.newInstance().newTransformer();
+                serializer.setOutputProperty(OutputKeys.INDENT, "yes");
+                serializer.setOutputProperty("{http://xml.apache.org/xslt}indent-amount", "4");
+                Source xmlSource = new SAXSource(new InputSource(new ByteArrayInputStream(
+                    content.getBytes(StandardCharsets.UTF_8))));
+                StreamResult res =  new StreamResult(new ByteArrayOutputStream());
+                serializer.transform(xmlSource, res);
+                return new String(((ByteArrayOutputStream) res.getOutputStream()).toByteArray(),
+                    StandardCharsets.UTF_8);
```
alzimmermsft: Instead of casting to `ByteArrayOutputStream` why don't we capture a local instance of the `ByteArrayOutputStream` we are creating to construct `StreamResult`, it will allow use to just call `toByteArray` on it.

```java
ByteArrayOutputStream outStream = new ByteArrayOutputStream();
StreamResult res = new StreamResult(outStream);
serializer.transform(xmlSource, res);
return new String(outStream.toByteArray(), StandardCharsets.UTF_8);
```

sdk/core/azure-core/src/main/java/com/azure/core/http/ContentType.java
alzimmermsft: Also lets uppercase `The`, if you could update all instances in this file that would be great!

JonathanGiles:
@sima-zhu Can you please finish this PR off so that we can get it merged?

PR #6753, opened 2019-12-10
Added support for ADLS Error Mapping

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/6358

### DIFF
--- a/sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/specialized/DataLakeLeaseAsyncClient.java
+++ b/sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/specialized/DataLakeLeaseAsyncClient.java
@@ -9,7 +9,6 @@ import com.azure.core.annotation.ServiceMethod;
 import com.azure.core.http.RequestConditions;
 import com.azure.core.http.rest.Response;
 import com.azure.core.util.FluxUtil;
-import com.azure.storage.blob.models.BlobStorageException;
 import com.azure.storage.blob.specialized.BlobLeaseAsyncClient;
 import com.azure.storage.file.datalake.DataLakeDirectoryAsyncClient;
 import com.azure.storage.file.datalake.DataLakeFileAsyncClient;


sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileClient.java
```diff
@@ -276,10 +278,14 @@ public void read(OutputStream stream) {
      */
     public FileReadResponse readWithResponse(OutputStream stream, FileRange range, DownloadRetryOptions options,
         DataLakeRequestConditions requestConditions, boolean getRangeContentMd5, Duration timeout, Context context) {
-        BlobDownloadResponse response = blockBlobClient.downloadWithResponse(stream, Transforms.toBlobRange(range),
-            Transforms.toBlobDownloadRetryOptions(options), Transforms.toBlobRequestConditions(requestConditions),
-            getRangeContentMd5, timeout, context);
-        return Transforms.toFileReadResponse(response);
+        try {
+            BlobDownloadResponse response = blockBlobClient.downloadWithResponse(stream, Transforms.toBlobRange(range),
```
alzimmermsft: Any chance to have the sync `readWithResponse` call into the async `readWithResponse` in a similar fashion that Blobs handles sync download calling into async download? That way the try/catch convert exception won't need to be written everywhere.

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileSystemClient.java
```diff
@@ -195,8 +197,12 @@ public void create() {
...
     public Response<Void> createWithResponse(Map<String, String> metadata, PublicAccessType accessType,
         Duration timeout, Context context) {
-        return blobContainerClient.createWithResponse(metadata, Transforms.toBlobPublicAccessType(accessType), timeout,
-            context);
+        try {
+            return blobContainerClient.createWithResponse(metadata, Transforms.toBlobPublicAccessType(accessType),
+                timeout, context);
+        } catch (BlobStorageException ex) {
+            throw logger.logExceptionAsError(DataLakeImplUtils.transformBlobStorageException(ex));
+        }
```
alzimmermsft: Could `DataLakeImplUtils` add a method such as the following to ensure the same code is called every place a try/catch convert is performed?

```java
public static <T> T returnOrConvertException(Supplier<T> supplier, ClientLogger logger) {
  try {
    return supplier.get();
  } catch (BlobStorageException ex) {
    throw logger.logExceptionAsError(transformBlobStorageException(ex));
  }
}
```

A calling pattern from the clients would look it this.

```java
return DataLakeImplUtils.returnOrConvertException(() -> 
  blobContainerClient.createWithResponse(metadata, Transforms.toBlobPublicAccessType(accessType), timeout, context), logger);
```

Just a thought, current implementation looks correct but changes could be cumbersome.

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakePathAsyncClient.java
```diff
@@ -413,6 +416,7 @@ public DataLakeServiceVersion getServiceVersion() {
     public Mono<Response<PathProperties>> getPropertiesWithResponse(DataLakeRequestConditions requestConditions) {
         try {
             return blockBlobAsyncClient.getPropertiesWithResponse(Transforms.toBlobRequestConditions(requestConditions))
+                .onErrorMap(ex -> DataLakeImplUtils.transformBlobStorageException((BlobStorageException) ex))
```
alzimmermsft: Should this be after the `map`?

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/implementation/util/DataLakeImplUtils.java
```diff
@@ -14,4 +17,24 @@ public static String endpointToDesiredEndpoint(String endpoint, String desiredEn
             return endpoint.replaceFirst(currentStringToMatch, desiredStringToMatch);
         }
     }
+
+    public static RuntimeException transformBlobStorageException(RuntimeException ex) {
```
alzimmermsft: This could be removed given that `RuntimeException` is a subclass of `Throwable` so the method below should handle all cases this method would be used.

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/models/DataLakeExceptionMappingPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+import reactor.core.publisher.Mono;
+
+/**
+ * This is an exception mapping policy in an HttpPipeline to map BlobStorageExceptions to DataLakeStorageExceptions.
+ */
+public class DataLakeExceptionMappingPolicy implements HttpPipelinePolicy {
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        return next.process().onErrorMap(DataLakeExceptionMappingPolicy::transformBlobStorageException);
```
alzimmermsft: Could this call into `DataLakeImplUtils.transformBlobStorageException` instead of creating its own implementation of the method.

gapra-msft: I need to remove this class. I dont really use it anywhere.

gapra-msft: I tried to make a policy but it doesnt work because of the way we handle exceptions

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/implementation/util/DataLakeImplUtils.java
gapra-msft: I needed to add this because otherwise I need to cast the Throwable to a RuntimeException. Let me make that method you suggested and see how it affects the need for the two same methods

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileSystemClient.java
gapra-msft: This looks like it should work better. I'll update to this

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileClient.java
gapra-msft: The reason we cant do this is because we have no way to pass the context (since we dont have the package private method)

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakePathAsyncClient.java
gapra-msft: For some reason putting onErrorMap after gives me a build issue telling me that it expects the method to return a Mono<SimpleResponse<PathProperties>> so any method that maps to a simple response I just do later

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeServiceAsyncClient.java
```diff
@@ -256,6 +257,7 @@ public String getAccountUrl() {
     public PagedFlux<FileSystemItem> listFileSystems(ListFileSystemsOptions options) {
         try {
             return blobServiceAsyncClient.listBlobContainers(Transforms.toListBlobContainersOptions(options))
+//                .onErrorMap(ex -> DataLakeImplUtils.transformBlobStorageException((BlobStorageException) ex))
```
rickle-msft: Can you open an issue for this to track it and reference the issue open against PagedIterable and PagedFlux?

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/specialized/DataLakeLeaseClient.java
```diff
@@ -100,7 +105,12 @@ public String acquireLease(int duration) {
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Response<String> acquireLeaseWithResponse(int duration, RequestConditions modifiedRequestConditions,
         Duration timeout, Context context) {
-        return blobLeaseClient.acquireLeaseWithResponse(duration, modifiedRequestConditions, timeout, context);
+        try {
```
rickle-msft: Can you use the returnOrMapException method in this file? Also, I think when you're constructing the DataLakeStorageException, it might be better to use one of the constructors that accepts a Throwable and pass the original exception in. I think this helps preserve the cause/stack trace

sdk/storage/azure-storage-file-datalake/src/test/java/com/azure/storage/file/datalake/AsyncErrorMappingTest.groovy
```diff
@@ -0,0 +1,158 @@
...
+        fsac = getServiceAsyncClient(primaryCredential).createFileSystem(fileSystemName).block()
+    }
+
+    def "Read file"() {
+        when:
+        def fileName = generatePathName()
+        def fac = fsac.getFileAsyncClient(fileName)
+        def readFileVerifier = StepVerifier.create(fac.readWithResponse(null, null, null, false))
+        then:
+        readFileVerifier.verifyErrorSatisfies {
```
rickle-msft: I think there's an overload of verifyError that just accepts the type of the error. Then you don't have to write these lambdas

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeServiceAsyncClient.java
gapra-msft: Created here https://github.com/Azure/azure-sdk-for-java/issues/6808

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/specialized/DataLakeLeaseClient.java
gapra-msft: Changed to use returnOrMapException

gapra-msft: For some reason that constructor doesn't work

sdk/storage/azure-storage-file-datalake/src/test/java/com/azure/storage/file/datalake/AsyncErrorMappingTest.groovy
gapra-msft: Yep, changed to use that

PR #6766, opened 2019-12-11
Update version tags on tracing readme

samvaity:
- Remove configuring http client from tracing readme.
- Remove update version tags from .md files and follow https://github.com/Azure/azure-sdk-for-java/blob/master/CONTRIBUTING.md#tooling-version-files-and-marker-tags

### DIFF
--- a/sdk/core/azure-core-tracing-opencensus/src/samples/AsyncQueueClientEnqueueMessages.md
+++ b/sdk/core/azure-core-tracing-opencensus/src/samples/AsyncQueueClientEnqueueMessages.md
@@ -18,7 +18,6 @@ Sample uses **[opencensus-impl][opencensus_impl]** as implementation package and
 </dependency>
 ```
 [//]: # ({x-version-update-end})
-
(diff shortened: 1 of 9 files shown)


sdk/core/azure-core-tracing-opencensus/src/samples/AsyncQueueClientEnqueueMessages.md
```diff
@@ -7,29 +7,31 @@ messages on queue client with tracing instrumentation.
...
 <!-- Add Storage Queue dependency -->
 <dependency>
   <groupId>com.azure</groupId>
   <artifactId>azure-storage-queue</artifactId>
-  <version>12.1.0</version> <!-- {x-version-update;com.azure:azure-storage-queue;current} -->
+  <version>12.1.0</version>
 </dependency>
 ```
 [//]: # ({x-version-update-end})
+
```
JimSuplizio: You don't need the extra line in between the two version tags ([//]:) only before/after the entire block in these cases. That's a general comment for the .md files.

PR #6771, opened 2019-12-11
v4: adds an in clause query benchmark

moderakh:
The purpose of this PR is to add a baseline for measure latencies for a in-clause cross partition query for different partition key values.
* adds a in clause query benchmark.
* measures latencies for queries.

sample query:

```
select * from c where c.pk in ("pk1", "pk2", "pk3, ..., "pk2000")
```


sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncQueryBenchmark.java
```diff
@@ -34,7 +71,7 @@ protected void onSuccess() {
 
     @Override
     protected void performWorkload(BaseSubscriber<FeedResponse<Document>> baseSubscriber, long i) throws InterruptedException {
-
+        concurrencyControlSemaphore.acquire();
```
kushagraThapar: Why are we acquiring the lock so early in the `performWorkload`?

moderakh: sure, moved down. DONE

PR #6773, opened 2019-12-12
Add links on batch send operation in eventhubs

### LINKED ISSUE #6758: [FEATURE REQ] EventHub distributed tracing: add links on Send

EventHubs Track2 SDK creates 
- span per each message (when message is added to batch or sent).
- span for sending message/batch.

We assume that users, knowing their code, in most cases can figure out relationships between messages/send operations (they are most likely siblings).

The problem happens when we try to use automated tools and visualizations for the following cases

**Scenario 1**
- message was created at timestamp1
- sent at timestamp2
- processed at timestamp3

**Issue**: How to calculate heuristic like 'Time in queue' for the message (we can't say for sure when it was sent) for arbitrary case (not relying on operation names specific for event hub)

**Scenario 2**
- message was created at timestamp1
- sent at timestamp2 - send failed
- send retried at timestamp3 - succeed
- processed at timestamp4

Same issue.

**Scenario 3**
- message1 was created
- message1 sent 
- message2 was created
- message2 sent 

How do users/tools could find out which message was sent in which operation?

Send span should have links to tracing contexts of all messages being sent.

**Additional context**

e.g. with the following code (sorry for my c#)

```csharp
EventDataBatch batch1 = await eventHubProducerClient1.CreateBatchAsync();
EventDataBatch batch2 = await eventHubProducerClient2.CreateBatchAsync();

batch1.TryAdd(new EventData(Encoding.UTF8.GetBytes("1")));
batch2.TryAdd(new EventData(Encoding.UTF8.GetBytes("2")));

batch1.TryAdd(new EventData(Encoding.UTF8.GetBytes("3")));
batch2.TryAdd(new EventData(Encoding.UTF8.GetBytes("4")));

await eventHubProducerClient1.SendAsync(batch1);
await eventHubProducerClient2.SendAsync(batch2);
```

image

Try to figure out when which message was sent and to which event hub :)

**Information Checklist**
- [x] Description Added

samvaity:
Fixes #6758

### DIFF
--- a/sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
+++ b/sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Release History
 ## Version 1.0.0-beta.2 (2020-01-03)
-- Add eventHub properties to attributes of processing spans.
+- Add `EventHubs.*` properties to attributes of processing spans.
 - Remove `Azure` prefix from convenience layer span names.
-- Add links for batch send operation in EventHubs client library. 
+- Add links for batch send operation in Event Hubs client library. 
 
 This package's
 [documentation](https://github.com/Azure/azure-sdk-for-java/blob/azure-core-tracing-opentelemetry_1.0.0-beta.2/sdk/core/azure-core-tracing-opentelemetry/README.md)
--- a/sdk/core/azure-core-tracing-opentelemetry/src/samples/PublishEventsTracingSample.md
+++ b/sdk/core/azure-core-tracing-opentelemetry/src/samples/PublishEventsTracingSample.md
@@ -60,6 +60,7 @@ public class Sample {
     private static final Logger LOGGER = getLogger("Sample");
     private static  final Tracer TRACER;
     private static final TracerSdkFactory TRACER_SDK_FACTORY;
+    private static final Duration OPERATION_TIMEOUT = Duration.ofSeconds(30);
 
     static {
         TRACER_SDK_FACTORY = configureOpenTelemetryAndLoggingExporter();
@@ -102,7 +103,7 @@ public class Sample {
                 .setMaximumSizeInBytes(256);
 
             final AtomicReference<EventDataBatch> currentBatch = new AtomicReference<>(
-                producer.createBatch(options).block());
+                producer.createBatch(options).block(OPERATION_TIMEOUT));
             
             data.flatMap(event -> {
                 final EventDataBatch batch = currentBatch.get();
@@ -135,8 +136,11 @@ public class Sample {
                     })
                     .subscribe(unused -> System.out.println("Complete"),
                             error -> System.out.println("Error sending events: " + error),
-                            () -> System.out.println("Completed sending events."));
-            span.end();
+                            () -> {
+                                span.end();
+                                producer.close();
+                                System.out.println("Completed sending events.");
+                            });
         }
     }
 }


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -397,17 +398,21 @@ public String getEventHubName() {
             ? new AtomicReference<>(Context.NONE)
             : null;
 
+        // initiate addLink on send span, get shared span builder for all links
+        Context sharedContext = isTracingEnabled ? tracerProvider.startSpan(parentContext.get(), ProcessKind.LINK) :
```
lmolkova: does it really starts a span? we'll start another one somewhere later - why do we need two?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -416,7 +421,7 @@ public String getEventHubName() {
         return getSendLink(batch.getPartitionId())
             .flatMap(link -> {
                 if (isTracingEnabled) {
-                    Context userSpanContext = parentContext.get();
+                    Context userSpanContext = sharedContext;
                     Context entityContext = userSpanContext.addData(ENTITY_PATH_KEY, link.getEntityPath());
                     // start send span and store updated context
                     parentContext.set(tracerProvider.startSpan(
```
lmolkova: will this span reuse span builder we created before?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
samvaity: no this, doesn't it just returns a span builder. I wanted to stay consistent with the exposed methods that are usually used around with tracing, so used an existing one.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
samvaity: yes, the sharedContext on [this](https://github.com/Azure/azure-sdk-for-java/pull/6773/files#diff-afcf12d6b6264dbb96ae63baef89082dR402) line, should get us the span builder returned by the `ProcessKind.LINK`.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
samvaity: no, it will just return a span builder created with "Azure.eventhubs.send" name. Just wanted to keep the naming consistent with the other methods and so used an existing one.

lmolkova: I see. I understand the intention to stay in the existing public API boundaries, but  it seems to be very misleading. startSpan sometimes start span, sometimes just creates a builder in the context: it's completely custom behavior for each kind. It seems very hard to read and maintain. Also I believe it makes almost impossible for someone to implement tracer - they need to understand all the rules hidden in 

https://github.com/Azure/azure-sdk-for-java/blob/9076e2f7bb141d9be4ad9c788c5fa7c4cf12795c/sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java#L67..L88

While it's surely up to you how to proceed, I suggest to split it into a different method. Or at least leave a comment/hints that explain what's going on here.

samvaity: should we rather just add a new method `getSpanBuilder()` instead?

sdk/core/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/OpenCensusTracer.java
```diff
@@ -181,6 +181,12 @@ public Context extractContext(String diagnosticId, Context context) {
         return AmqpPropagationFormatUtil.extractContext(diagnosticId, context);
     }
 
+    @Override
+    public Context getSharedSpanBuilder(String spanName, Context context) {
+        return Context.NONE;
```
conniey: would it be better to throw an UnsupportedOperationException rather than return something that won't work?

sdk/core/azure-core-tracing-opentelemetry/src/test/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracerTest.java
```diff
@@ -117,9 +116,11 @@ public void startSpanTestNoUserParent() {
     public void startSpanProcessKindSend() {
         // Arrange
         final SpanId parentSpanId = parentSpan.getContext().getSpanId();
+        // Start user parent span.
+        final Span.Builder spanBuilder = tracer.spanBuilder(METHOD_NAME);
         // Add additional metadata to spans for SEND
         final Context traceContext = tracingContext.addData(ENTITY_PATH_KEY, ENTITY_PATH_VALUE)
-            .addData(HOST_NAME_KEY, HOSTNAME_VALUE);
+                                         .addData(HOST_NAME_KEY, HOSTNAME_VALUE).addData(SPAN_BUILDER_KEY, spanBuilder);
```
conniey: this is an odd indent level

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -115,8 +116,8 @@
      * load balance the messages amongst available partitions.
      */
     EventHubProducerAsyncClient(String fullyQualifiedNamespace, String eventHubName, EventHubConnection connection,
-        AmqpRetryOptions retryOptions, TracerProvider tracerProvider, MessageSerializer messageSerializer,
-        boolean isSharedConnection) {
+                                AmqpRetryOptions retryOptions, TracerProvider tracerProvider,
```
conniey: revert indentation. if the params are squished against the right margin, it should have an 8 space indentation.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -207,39 +204,40 @@ public String getEventHubName() {
...
-                    final int batchSize = batchMaxSize > 0
-                        ? batchMaxSize
-                        : maximumLinkSize;
-
-                    return Mono.just(new EventDataBatch(batchSize, partitionId, partitionKey, link::getErrorContext,
-                        tracerProvider));
-                }));
+                   .flatMap(link -> link.getLinkSize()
+                                        .flatMap(size -> {
+                                            final int maximumLinkSize = size > 0
```
conniey: this indentation is off

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -394,42 +384,53 @@ public String getEventHubName() {
...
-            ? new AtomicReference<>(Context.NONE)
-            : null;
+                                                           ? new AtomicReference<>(Context.NONE)
+                                                           : null;
+        final AtomicReference<Context> sharedContext = isTracingEnabled
+                                                           ? new AtomicReference<>(Context.NONE)
+                                                           : null;
+        final AtomicReference<Boolean> isFirst = new AtomicReference<>(true);
 
         final List<Message> messages = batch.getEvents().stream().map(event -> {
```
conniey: Consider using a `for` loop rather than the `stream().map()` so you don't need the "isFirst" variable and the parentContext and sharedContext do not have to be atomicreferences.

samvaity: I don't think we can avoid the `isFirst`  because we don't want to send the request to `getSharedSpanBuilder` for every event, we can just do that for one event and get the shared builder for all others. 

Also, `parentContext ` would still have to be an AtomicReference as it is used [here](https://github.com/Azure/azure-sdk-for-java/pull/6773/files#diff-afcf12d6b6264dbb96ae63baef89082dR423) in this lambda.

sdk/core/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/OpenCensusTracer.java
```diff
@@ -181,6 +181,14 @@ public Context extractContext(String diagnosticId, Context context) {
         return AmqpPropagationFormatUtil.extractContext(diagnosticId, context);
     }
 
+    @Override
+    public Context getSharedSpanBuilder(String spanName, Context context) {
+        logger.logExceptionAsError(
```
conniey: You'll have to throw this so you don't need to return context.none

sdk/core/azure-core-tracing-opentelemetry/src/samples/PublishEventsTracingSample.md
```diff
@@ -81,31 +81,20 @@ public class Sample {
...
-            for (EventData event : telemetryEvents) {
-                if (!currentBatch.tryAdd(event)) {
-                    producer.send(currentBatch);
-                    currentBatch = producer.createBatch(options);
-                }
-            }
+            final Context traceContext = new Context(PARENT_SPAN_KEY, span);
+            final Flux<EventData> testData = Flux.range(0, count).flatMap(number -> {
+                final EventData data = new EventData(body, traceContext);
+                return Flux.just(data);
```
conniey: Instead of flatmap, you can use map. There is nothing asynchronous about this.

sdk/core/azure-core-tracing-opentelemetry/src/samples/PublishEventsTracingSample.md
```diff
@@ -81,31 +81,20 @@ public class Sample {
...
+            });
+            producer.send(testData.toIterable(1));
```
conniey: toIterable(1)? There is no exposed overload that allows you to send a single EventData in EventHubProducer.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
conniey: I'm fairly confident this works:

```java
Context parentContext = null;
Context sharedContext = null;
List<Message> messages = new ArrayList<>();

for (int = 0; i < batch.getEvents().size(); i++) {
    final EventData = batch.getEvents().get(i);
    if (isTracingEnabled) {
        parentContext = event.getContext();
        if (i == 0) {
            sharedContext = tracerProvider.getSharedSpanBuilder(parentContext.get());
        }

        Context sharedSpanBuilderContext = sharedContext.get();
        tracerProvider.addSpanLinks(sharedSpanBuilderContext.addData(SPAN_CONTEXT_KEY, event.getContext()));
    }

    final Message message = messageSerializer.serialize(event);
    messages.add(message);
}
```

sdk/core/azure-core-tracing-opencensus/src/main/java/com/azure/core/tracing/opencensus/OpenCensusTracer.java
samvaity: I think we get some error saying you can't directly throw you should use a logger to do that.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
samvaity: We are using sharedContext and parentContext in the lambda expressions below and so they need to be `effectively final`.

sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # Release History
 ## Version 1.0.0-beta.2 (2020-01-03)
-- Add eventhub properties to attributes of processing spans.
+- Add eventHub properties to attributes of processing spans.
```
conniey:
```suggestion
- Add `EventHubs.*` properties to attributes of processing spans.
```

sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
...
 - Remove `Azure` prefix from convenience layer span names.
+- Add links for batch send operation in EventHubs client library.
```
conniey:
```suggestion
- Add links for batch send operation in Event Hubs client library. 
```

sdk/core/azure-core-tracing-opentelemetry/src/samples/PublishEventsTracingSample.md
```diff
@@ -80,35 +80,63 @@ public class Sample {
...
-            EventDataBatch currentBatch = producer.createBatch(options);
-    
-            // For each telemetry event, we try to add it to the current batch.
-            for (EventData event : telemetryEvents) {
-                if (!currentBatch.tryAdd(event)) {
-                    producer.send(currentBatch);
-                    currentBatch = producer.createBatch(options);
+
+            final AtomicReference<EventDataBatch> currentBatch = new AtomicReference<>(
+                producer.createBatch(options).block());
```
conniey: `.block(OPERATION_TIMEOUT);`

sdk/core/azure-core-tracing-opentelemetry/src/samples/PublishEventsTracingSample.md
```diff
@@ -80,35 +80,63 @@ public class Sample {
...
+                        final EventDataBatch batch = currentBatch.getAndSet(null);
+                        if (batch != null) {
+                            producer.send(batch).block(OPERATION_TIMEOUT);
+                        }
+                    })
+                    .subscribe(unused -> System.out.println("Complete"),
+                            error -> System.out.println("Error sending events: " + error),
+                            () -> System.out.println("Completed sending events."));
             span.end();
-            producer.close();
```
conniey: We should close the producer when the send is complete. And same with the span in do finally.

PR #6792, opened 2019-12-12
IoT Central: generate 2018 09 01

ChenTanyi:


ChenTanyi:
@mitchdenny "check-enforcer" is not working in this PR. Could you find someway or just merge it?

mitchdenny:
In this case Check Enforcer is doing the right thing because there are no CI builds that have been run. Should this code change be running some kind of build job to validate that it doesn't break something?

PR #6793, opened 2019-12-12
Policy Insights: generate 2019 10 01

ChenTanyi:


weidongxu-microsoft:
@mitchdenny "check-enforcer" is not working in this PR. Could you find someway or just merge it?

mitchdenny:
Shouldn't this be running some kind of build job? If you just trigger the management build check enforcer will get unblocked. But at the moment it it isn't meeting the minimum required number of checks to allow check enforcer to pass.

weidongxu-microsoft:
@mitchdenny The CI supposed to be ci.mgmt.yml inside the service folders. But I think you have not batch added them yet.

https://github.com/Azure/azure-sdk-for-java/blob/policy_insight/sdk/policyinsights/ci.mgmt.yml

mitchdenny:
OK ... just use the /azp command to trigger the management build then.

weidongxu-microsoft:
@mitchdenny I cannot run it (likely because it exclude /sdk). Should we remove that exclude, before we get mgmt CIs on board?

weidongxu-microsoft:
@mitchdenny Or as a workaround we just touch some file (with empty line e.g) in eng/mgmt to trigger it?

mitchdenny:
I've disabled Check Enforcer as a required check. So you should be unblocked.

ChenTanyi:
Thanks for your help!

PR #6794, opened 2019-12-12
Deployment Manager: generate 2019 11 01

ChenTanyi:
The package would be publish after review.
I wonder it is ok the do such fix.


sdk/deploymentmanager/mgmt-v2019-11-01-preview/pom.xml
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * Code generated by Microsoft (R) AutoRest Code Generator.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.microsoft.azure.deploymentmanager.v2019_11_01_preview</groupId>
+    <parent>
+        <groupId>com.microsoft.azure</groupId>
+        <artifactId>azure-arm-parent</artifactId>
+        <version>1.1.0</version>
```
JimSuplizio: /CC: @weidongxu-microsoft 
com.microsoft.azure:azure-arm-parent (pom.management.xml)'s version is currently 1.2.0, shouldn't this be referencing that version and not the older, 1.1.0, version?

weidongxu-microsoft: Yes, Tanyi please fix it.

This is due to latest autorest.java extension not yet published, else it would be 1.2.0 automatically. So at present we had to do it manually.

I think Tanyi is now trying to double check with service team on that `RolloutRequestInner` in **response** body...

ChenTanyi: Done

sdk/deploymentmanager/mgmt-v2019-11-01-preview/src/main/java/com/microsoft/azure/management/deploymentmanager/v2019_11_01_preview/implementation/RolloutImpl.java
```diff
@@ -31,12 +31,12 @@
         this.createOrUpdateParameter.withLocation(inner().location());
         this.createOrUpdateParameter.withTags(inner().getTags());
         return client.createOrUpdateAsync(this.resourceGroupName(), this.name(), this.createOrUpdateParameter)
-            .map(new Func1<RolloutInner, RolloutInner>() {
-               @Override
-               public RolloutInner call(RolloutInner resource) {
-                   resetCreateUpdateParameters();
-                   return resource;
-               }
+            .to(new Func1<Observable<RolloutRequestInner>, Observable<RolloutInner>>() {
```
weidongxu-microsoft: Check if `flatMap` compiles or not. I still not figure out what `to` method does...

The point here is that the func need to be called **after** first Observable is complete.

ChenTanyi:
@JimSuplizio Could you review it again?

PR #6803, opened 2019-12-12
Add "eng" folder to core path triggers

mikeharder:
- Core is a good pipeline to run on changes under "eng" folder
- Aligns Java with other languages

### DIFF
--- a/sdk/core/ci.yml
+++ b/sdk/core/ci.yml
@@ -21,6 +21,7 @@ trigger:
   paths:
     include:
       - sdk/core/
+      - eng/
 
 pr:
   branches:
@@ -32,6 +33,7 @@ pr:
   paths:
     include:
       - sdk/core/
+      - eng/
 
 stages:
   - template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml


alzimmermsft:
MacOS failure is unrelated to this change. Feel free to merge away.

mikeharder:
> Could this be further scoped to code-quality-reports, spotbugs-aggregate, and the client specific YML templates in the future?

DevOps only supports triggering on folders, not individual files.  So it could be scoped to a more narrow set of folders if it exists.

mikeharder:
> > Could this be further scoped to code-quality-reports, spotbugs-aggregate, and the client specific YML templates in the future?
> 
> DevOps only supports triggering on folders, not individual files. So it could be scoped to a more narrow set of folders if it exists.

Actually it looks like DevOps does support individual files and wildcards, but so far we have kept it simple and just used folders.

https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#paths

mikeharder:
@alzimmermsft: If including all of "eng" causes too many unnecessary builds we can narrow the scope.

weshaggard:
>Could this be further scoped to code-quality-reports, spotbugs-aggregate, and the client specific YML templates in the future?

What are you trying to avoid, under eng?

PR #6810, opened 2019-12-12
Shorten tracing span names

### LINKED ISSUE #6791: [BUG] Distributed tracing: short span names for public api surface
**Describe the bug**
Java SDKs set span names (for public api surface spans) to fully qualified name: 'namespace.class/method'.  Namespace presence makes span names too long and is not very informative. 'Class.method' provides enough granularity and context for users.
 
Python and JS SDKs use class.method (or method only).

N/A

**To Reproduce**

***Code Snippet***

**Expected behavior**
Span names should look like "Appconfig.setKey'. It should affect the convenience layer spans only.

**Screenshots**

This is how it looks now:

image

how it looks with short names (screenshot is taken from the different .NET app with different scenario):

image

 - OS:  N/A
 - IDE : N/A
 - Version of the Library used: -

**Information Checklist**
- [x] Bug Description Added
- [ ] Repro Steps Added
- [ ] Setup information Added

samvaity:
Remove `Azure` prefix from span names.
Fixes #6791

Now,

image

image

### DIFF
--- a/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TracerProvider.java
+++ b/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TracerProvider.java
@@ -115,7 +115,7 @@ public class TracerProvider {
     }
 
     private String getSpanName(ProcessKind processKind) {
-        String spanName = "Eventhubs.";
+        String spanName = "EventHubs.";
         switch (processKind) {
             case SEND:
                 spanName += "send";
--- a/sdk/core/azure-core-amqp/src/test/java/com/azure/core/amqp/implementation/TracerProviderTest.java
+++ b/sdk/core/azure-core-amqp/src/test/java/com/azure/core/amqp/implementation/TracerProviderTest.java
@@ -32,7 +32,7 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 public class TracerProviderTest {
-    private static final String METHOD_NAME = "Eventhubs.send";
+    private static final String METHOD_NAME = "EventHubs.send";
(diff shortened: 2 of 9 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/rest/RestProxy.java
```diff
@@ -196,7 +196,7 @@ private Method determineResumeMethod(Method method, String resumeMethodName) {
      * @return The updated context containing the span context.
      */
     private Context startTracingSpan(Method method, Context context) {
-        String spanName = String.format("Azure.%s/%s", interfaceParser.getServiceName(), method.getName());
+        String spanName = String.format("%s/%s", interfaceParser.getServiceName(), method.getName());
```
lmolkova: for the sake of having the same naming format as other azure SDKs. can we use '.' as a separator everywhere?

E.g. 'Appconfig.getKey' rather than 'Appconfig/getKey'

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TracerProvider.java
```diff
@@ -115,7 +115,7 @@ private void end(String statusMessage, Throwable throwable, Context context) {
     }
 
     private String getSpanName(ProcessKind processKind) {
-        String spanName = "Azure.eventhubs.";
+        String spanName = "Eventhubs.";
```
conniey: `EventHubs.`

conniey: Same with all the other ones.

Issue #6438, opened 2019-11-20
[BUG] Missing Connection String Null or Empty error message.


mssfang:
If user gives a wrong endpoint or missing the endpoint. No exception thrown currently. The methods, listConfigurationSettings() and listRevisions() returns an empty list.

mssfang:
After testing this scenario again. The exception is thrown but there is no customized message telling users, "'Connection string' cannot be null or empty." when the connection string is null or empty

PR #6811, opened 2019-12-12
Add connection string null or empty message error

mssfang:
when Connection string is null or empty, there is not a good appropriate error message to define the error. Added ``
                throw new IllegalArgumentException("'Connection string' cannot be null or empty.");
``

fixes: #6438


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/ConfigurationClientCredentials.java
```diff
@@ -178,7 +178,7 @@ String id() {
 
         CredentialInformation(String connectionString) {
             if (CoreUtils.isNullOrEmpty(connectionString)) {
-                throw new IllegalArgumentException(connectionString);
+                throw new IllegalArgumentException("'Connection string' cannot be null or empty.");
```
conniey: "'connectionString' cannot be null or empty."

PR #6826, opened 2019-12-13
This will snapshot the repo state for analysis if the build fails.

mitchdenny:
Pushing this PR through straight away to try and diagnose a problem with dev packages when builds fail. It basically uploads a copy of the repo state on disk (excluding the .git folder).

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -239,6 +239,11 @@ jobs:
         displayName: 'Publish Report Artifacts'
         artifact: reports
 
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -239,6 +239,11 @@ jobs:
         displayName: 'Publish Report Artifacts'
         artifact: reports
 
+      - publish: $(System.DefaultWorkingDirectory)
```
weshaggard: Should this be `Pipeline.Workspace` instead? I'm not sure if the default working directory can be changed or not.

PR #6833, opened 2019-12-13
Add managed identity live tests in new e2e module

jianghaolu:


JimSuplizio:
I'm curious as to why this is its own entity instead of part of another library? Is this going to get filled in with end to end tests for other areas? This module was added to pom.client.xml but it's not part of any pipeline build, should it be? When/where are these tests supposed to be running.

PR #6834, opened 2019-12-13
Fix dev package production

mitchdenny:
The dev package build/publishing process has not been working correctly since last week. The way the process works is that for each artifact that a particular pipeline produces, it appends a dev.yyyyMMdd.x number to the version and then updates the dependencies for all the packages in the repo that depends on it.

Unfortunately the update logic does a simple append operation to the version number that is already in the ```versions_client.txt``` file. In the case of ```azure-core-http-netty``` library its released version is currently ```1.1.0```, but the repo contains ```1.2.0-beta.1```. When we blindly append ```dev.yyyyMMdd.x ``` to the dependencies we end up with a package that cannot be resolved (```1.1.0-dev.yyyyMMdd.x``` doesn't exist anywhere).

The change in this PR fixes the build by bumping the version of ```azure-core-http-netty``` that they depend on.

HOWEVER, I don't think that this is the best long term solution. We need to come up with a more nuanced ```update_versions.py``` script which will only append the dev package suffix if the version that is already in the dependencies list matches what is defined in ```versions_client.txt``` (for example).

/cc @JimSuplizio I'd be really interested in your opinion on this one!

### DIFF
--- a/eng/versioning/version_client.txt
+++ b/eng/versioning/version_client.txt
@@ -6,7 +6,7 @@ com.azure:azure-sdk-parent;1.6.0;1.6.0
 com.azure:azure-client-sdk-parent;1.7.0;1.7.0
 com.azure:azure-core;1.2.0-beta.1;1.2.0-beta.1
 com.azure:azure-core-amqp;1.0.0-beta.9;1.0.0-beta.9
-com.azure:azure-core-http-netty;1.1.0;1.2.0-beta.1
+com.azure:azure-core-http-netty;1.2.0-beta.1;1.2.0-beta.1
 com.azure:azure-core-http-okhttp;1.1.0-beta.1;1.1.0-beta.1
 com.azure:azure-core-management;1.0.0-beta.8;1.0.0-beta.8
 com.azure:azure-core-test;1.1.0-beta.1;1.1.0-beta.1
--- a/sdk/appconfiguration/azure-data-appconfiguration/README.md
+++ b/sdk/appconfiguration/azure-data-appconfiguration/README.md
@@ -23,7 +23,7 @@ Use the client library for App Configuration to create and manage application co
 <dependency>
     <groupId>com.azure</groupId>
     <artifactId>azure-data-appconfiguration</artifactId>
-    <version>1.0.0-preview.6</version>
+    <version>1.0.0-beta.8</version>
 </dependency>
 ```
 [//]: # ({x-version-update-end})
@@ -37,7 +37,7 @@ AppConfiguration to use Netty HTTP client.
 <dependency>
     <groupId>com.azure</groupId>
     <artifactId>azure-core-http-netty</artifactId>
-    <version>1.0.0</version>
+    <version>1.2.0-beta.1</version>
 </dependency>
 ```
 [//]: # ({x-version-update-end})
@@ -52,7 +52,7 @@ Netty and include OkHTTP client in your pom.xml.
 <dependency>
     <groupId>com.azure</groupId>
     <artifactId>azure-data-appconfiguration</artifactId>
-    <version>1.0.0-preview.6</version>
+    <version>1.0.0-beta.8</version>
     <exclusions>
       <exclusion>
         <groupId>com.azure</groupId>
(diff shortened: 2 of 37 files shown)


alzimmermsft:
Failing validation given the dependencies weren't updated with the update script. Need to bump the dependency version in the versions text file to match `1.2.0-beta.1`.

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -46,7 +46,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-http-netty</artifactId>
-      <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
```
conniey: Did you run the script to update these? The dependency version looks like 1.1.0. @JimSuplizio 
https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/version_client.txt#L9

JimSuplizio: Yeah, it looks like you'd updated the dependency version of azure-core-http-netty without checking in the version_client.txt.

JimSuplizio: > HOWEVER, I don't think that this is the best long term solution. We need to come up with a more nuanced `update_versions.py` script which will only append the dev package suffix if the version that is already in the dependencies list matches what is defined in `versions_client.txt` (for example).

@mitchdenny by "update_versions.py" you mean "set_versions.py" don't you?

I believe that the problem problem is this: You're calling set_versions.py with "--update-type all" and you should really be calling it with "--update-type library". "-all" means update both the dependency and the current. You're creating versions to upload into the dev feed so only the current version needs to be changed and changing the dependency version isn't necessary.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -62,14 +62,12 @@ jobs:
             python3 --version
             python3 eng/versioning/set_versions.py --update-type all --build-type client --build-qualifier dev.$(Build.BuildNumber) --artifact-id ${{artifact.name}}
             python3 eng/versioning/set_versions.py --update-type all --build-type data --build-qualifier dev.$(Build.BuildNumber) --artifact-id ${{artifact.name}}
-          condition: eq(variables['SetDevVersion'],'true')
```
JimSuplizio: I why are you removing the SetDevVersion=true condition in the steps here and below? Without the condition this would end up adding the dev version to everything for every build.

mitchdenny: This is intentional until I merge (I'll add it back in). It just means I can test dev package production in the PR flow (it doesn't publish them).

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
mitchdenny: No I mean update versions. The problem occurs when one of the downstream dependents is relying on the version of core in the repository. When we change that version via update versions (with just - -ut library) it can no longer resolve that version. We either need to flick that down stream dependent to use a dev package or we have to update the whole graph.

mitchdenny: Remember that this update is scoped to just the build, it does t get checked in.

mitchdenny: For now I've just run ```update_versions.py --ut all``` across the repo to snap all the versions to where they need to be - this fixes dev package production. So once I put the condition back in I can merge this PR. I think that there might end up being some nuanced scenarios we need to deal with around versioning when a downstream dependency of core is no longer being worked on and we just leave its dependency version alone.

Thoughts?

mitchdenny:
Looks like Jims update fixed this.

JimSuplizio:
> Looks like Jims update fixed this.

Yeah, when I was in there writing the version verification and increment versioning logic I overhauled this.

PR #6835, opened 2019-12-13
Takes advantage of Reactive Stream Concurrency for Query (Improves latency and fixes earlier issues)

moderakh:
This PR takes advantage of ReactiveStream concurrency for query, and improves query latency.
**Improves the measured latency (for the following cross partition query). Time down from 8s to 0.7 second!**

Using `QueryInClauseParallel` benchmark (added here https://github.com/Azure/azure-sdk-for-java/pull/6771) the latency of the following cross partition query was measured before and after the PR against a collection with 1000,000 RU.

`SELECT * FROM c where c.pk in ("pk1", "pk2", ... "pk2000")`

(where pk is the partition key, and 2000 documents were randomly inserted prior to the test to the collection)

Query e2e latency for retrieving all 2000 documents:

* **before the PR:
            99.9% <= 8,318.88 milliseconds**

* **after the PR:
            99.9% <= 718.13 milliseconds**

**How to repro this result and measure the latency. Use the benchmark tool with the following config (it assumes the database and collection already created)**

on a Ubuntu Linux VM 16 CPU cores against a collection with 1000,000 throughput run

```bash
java -jar azure-cosmos-benchmark/target/azure-cosmos-benchmark-4.0.0-SNAPSHOT-jar-with-dependencies.jar -maxConnectionPoolSize 4000 -serviceEndpoint https://moderakh-perf-account.documents.azure.com:443/ -masterKey XXXX -databaseId testdb -collectionId testcol -consistencyLevel Eventual -concurrency 1 -numberOfOperations 50 -operation QueryInClauseParallel -connectionMode Direct -numberOfPreCreatedDocuments 2000 -documentDataFieldSize 160
```

with improvements in this PR:
Latency
             count = 50
         mean rate = 2.00 calls/second
     1-minute rate = 1.86 calls/second
     5-minute rate = 1.81 calls/second
    15-minute rate = 1.80 calls/second
               min = 416.36 milliseconds
               max = 718.13 milliseconds
              mean = 497.72 milliseconds
            stddev = 82.64 milliseconds
            median = 472.69 milliseconds
              75% <= 493.28 milliseconds
              95% <= 706.60 milliseconds
              98% <= 716.31 milliseconds
              99% <= 718.13 milliseconds
            99.9% <= 718.13 milliseconds

prior to this PR:

Latency
             count = 50
         mean rate = 0.13 calls/second
     1-minute rate = 0.14 calls/second
     5-minute rate = 0.09 calls/second
    15-minute rate = 0.05 calls/second
               min = 7524.18 milliseconds
               max = 8815.75 milliseconds
              mean = 7698.99 milliseconds
            stddev = 101.78 milliseconds
            median = 7683.46 milliseconds
              75% <= 7754.35 milliseconds
              95% <= 7882.09 milliseconds
              98% <= 7897.16 milliseconds
              99% <= 7904.80 milliseconds
            99.9% <= 8318.88 milliseconds

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/Configs.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/Configs.java
@@ -17,8 +17,6 @@ import static com.google.common.base.Strings.emptyToNull;
 
 public class Configs {
     private static final Logger logger = LoggerFactory.getLogger(Configs.class);
-    public static final int CPU_CNT = Runtime.getRuntime().availableProcessors();
-
     private final SslContext sslContext;
 
     // The names we use are consistent with the:
@@ -54,6 +52,7 @@ public class Configs {
 
     private static final int MAX_SHORT_BARRIER_RETRIES_FOR_MULTI_REGION = 4;
     private static final int SHORT_BARRIER_RETRY_INTERVAL_IN_MS_FOR_MULTI_REGION = 10;
+    private static final int CPU_CNT = Runtime.getRuntime().availableProcessors();
     private static final int DEFAULT_DIRECT_HTTPS_POOL_SIZE = CPU_CNT * 500;
 
     //  Reactor Netty Constants
@@ -64,6 +63,10 @@ public class Configs {
 
     public Configs() {
         this.sslContext = sslContextInit();
+    }
+
+    public static int getCPUCnt() {
+        return CPU_CNT;
     }
 
     private SslContext sslContextInit() {
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/ParallelDocumentQueryExecutionContext.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/ParallelDocumentQueryExecutionContext.java
@@ -350,7 +350,7 @@ public class ParallelDocumentQueryExecutionContext<T extends Resource>
     private int fluxSequentialMergeConcurrency(FeedOptions options, int numberOfPartitions) {
         int parallelism = options.getMaxDegreeOfParallelism();
         if (parallelism < 0) {
-            parallelism = Configs.CPU_CNT;
+            parallelism = Configs.getCPUCnt();
         } else if (parallelism == 0) {
             parallelism = 1;
         }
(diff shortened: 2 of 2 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/ParallelDocumentQueryExecutionContext.java
```diff
@@ -337,4 +346,26 @@ public EmptyPagesFilterTransformer(
...
+        }
+
+        return Math.min(numberOfPartitions, parallelism);
+    }
+
+    private int fluxSequentialMergePrefetch(FeedOptions options, int numberOfPartitions, int pageSize, int fluxConcurrency) {
+        int maxBufferedItemCount = options.getMaxBufferedItemCount();
+
+        if (maxBufferedItemCount <= 0) {
+            maxBufferedItemCount = Math.min(Configs.CPU_CNT * numberOfPartitions * pageSize, 100_000);
```
kushagraThapar: Any particular reason to use 100_000 ?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/Configs.java
```diff
@@ -17,6 +17,8 @@
 
 public class Configs {
     private static final Logger logger = LoggerFactory.getLogger(Configs.class);
+    public static final int CPU_CNT = Runtime.getRuntime().availableProcessors();
```
kushagraThapar: To be consistent with other constants in Configs, we define them as `private static final ...` and then provide a getter to get them. 
We should make `CPU_CNT` consistent with other config constants in this class.

kushagraThapar: In fact I just noticed it was `private static final int CPU_CNT = Runtime.getRuntime().availableProcessors();` 
We should change it back to private.

moderakh: added getCPUCnt done.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/ParallelDocumentQueryExecutionContext.java
moderakh: This is kind of heuristic but general idea is that if there are 100_000 documents of size 1KB, that means 100_000*1KB=100MB in memory which will start to put pressure on memory if we go beyond that.

PR #6836, opened 2019-12-13
Remove repo capture logic

mitchdenny:
I added some repo capture logic, its not something that we want to keep on so I've made it conditional.


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -102,7 +102,7 @@ jobs:
         artifact: '$(ArtifactName)'
 
       - publish: $(System.DefaultWorkingDirectory)
-        condition: failed()
+        condition: and(eq(variables['CaptureRepositoryOnFailure'], 'true'), failed())
```
weshaggard: Is this something you plan to include in all the pipelines?

PR #6842, opened 2019-12-13
Add CliCredential

FredGao-new:
Add CliCredential follow https://github.com/Azure/azure-sdk-for-python/pull/8962/files

### DIFF
--- a/sdk/identity/azure-identity/src/test/java/com/azure/identity/AzureCliCredentialTest.java
+++ b/sdk/identity/azure-identity/src/test/java/com/azure/identity/AzureCliCredentialTest.java
@@ -16,8 +16,6 @@ import reactor.core.publisher.Mono;
 import reactor.test.StepVerifier;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
-
-
 import static org.mockito.Mockito.when;
 
 @RunWith(PowerMockRunner.class)


tzhanl:
@jongio

sdk/identity/azure-identity/pom.xml
```diff
@@ -23,7 +23,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
```
jianghaolu: Sorry for the delay in review. This now should be 1.2.0

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -99,6 +105,68 @@
...
 
+    /**
+     * Asynchronously acquire a token from Active Directory with Aure CLI.
+     *
+     * @param request the details of the token request
+     * @return a Publisher that emits an AccessToken
+     */
+    public Mono<AccessToken> authenticateWithAzureCli(TokenRequestContext request) {
+        String azCommand = "az account get-access-token --output json --resource ";
+
```
g2vinay: This function, currently doesn't get tested.
I can add tests for it later, once this PR gets merged.

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredential.java
```diff
@@ -37,6 +38,7 @@
         super(new ArrayDeque<>(Arrays.asList(new EnvironmentCredential(identityClientOptions),
             new ManagedIdentityCredential(null, identityClientOptions),
             new SharedTokenCacheCredential(null, "04b07795-8ddb-461a-bbee-02f9e1bf7b46",
-                identityClientOptions))));
+                identityClientOptions),
+            new AzureCliCredential(identityClientOptions))));
```
g2vinay: Is this correct order for AzureCliCredential @jianghaolu  ?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -99,6 +105,68 @@
...
+            BufferedReader r = new BufferedReader(new InputStreamReader(p.getInputStream()));
+            String line;
+            StringBuilder output = new StringBuilder();
+            while (true) {
+                line = r.readLine();
+                if (line == null) {
+                    break;
+                }
+                if (line.startsWith("'az' is not recognized") || line.matches("(.*)az:(.*)not found")) {
+                    throw new Exception("Azure CLI not installed");
```
g2vinay: this should be 
return Mono.error(logger.logExceptionAsError(new Exception("Azure CL......

Would like to use a better exception class instead of just Exception.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AzureCliCredentialBuilder.java
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT License.
+
+package com.azure.identity;
+
+/**
+ * Fluent credential builder for instantiating a {@link AzureCliCredential}.
+ *
+ * @see AzureCliCredential
+ */
+public class AzureCliCredentialBuilder extends AadCredentialBuilderBase<AzureCliCredentialBuilder> {
```
jianghaolu: What happens if the user sets the `clientId`? I can understand setting the `tenantId`.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -99,6 +105,68 @@
...
+                }
+                output.append(line);
+            }
+            String processOutput = output.toString();
+            if (p.exitValue() != 0) {
+                throw new Exception(processOutput);
```
g2vinay: same here, return Mono.error(logger.logExceptionAsError(new Exception(.....

Would like to use a better exception class instead of just Exception.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -99,6 +105,68 @@
...
+                    SerializerEncoding.JSON);
+            String accessToken = objectMap.get("accessToken");
+            String time = objectMap.get("expiresOn");
+            String timeToSecond = time.substring(0, time.indexOf("."));
+            String timeJoinedWithT = String.join("T", timeToSecond.split(" "));
+            OffsetDateTime expiresOn = LocalDateTime.parse(timeJoinedWithT, DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+                    .atZone(ZoneId.systemDefault()).toOffsetDateTime().withOffsetSameInstant(ZoneOffset.UTC);
+            token = new AccessToken(accessToken, expiresOn);
+        } catch (Exception e) {
+            return Mono.error(e);
```
g2vinay: Okay, so, its being caught here.
better to add here, Mono.error(logger.logExceptionAsError(e)),
logger can be used from azure core

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredential.java
schaabs: Yes this would be the correct ordering.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
FredGao-new: sure, we can follow up after merge.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -99,6 +106,71 @@
...
+            } else {
+                starter = "/bin/sh";
+                switcher = "-c";
+            }
+
+            ProcessBuilder builder = new ProcessBuilder(starter, switcher, command.toString());
+            builder.redirectErrorStream(true);
+            Process p = builder.start();
+
+            BufferedReader r = new BufferedReader(new InputStreamReader(p.getInputStream()));
```
srnagar: This reader must be closed after use. Also, change name to `reader` instead of using a single char.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -99,6 +106,71 @@
...
+                starter = "cmd.exe";
+                switcher = "/c";
+            } else {
+                starter = "/bin/sh";
+                switcher = "-c";
+            }
+
+            ProcessBuilder builder = new ProcessBuilder(starter, switcher, command.toString());
+            builder.redirectErrorStream(true);
+            Process p = builder.start();
```
srnagar: Use `process` instead of `p`.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -99,6 +106,71 @@
...
+
+            BufferedReader r = new BufferedReader(new InputStreamReader(p.getInputStream()));
+            String line;
+            StringBuilder output = new StringBuilder();
+            while (true) {
+                line = r.readLine();
+                if (line == null) {
+                    break;
+                }
+                if(line.startsWith("'az' is not recognized") || line.matches("(.*)az:(.*)not found")) {
```
srnagar: Use string constants here.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -99,6 +106,71 @@
...
+        command.append(azCommand);
+        String scopes = ScopeUtil.scopesToResource(request.getScopes());
+        command.append(scopes);
+    
+        AccessToken token = null;
+        try {
+            String starter;
+            String switcher;
+            if (System.getProperty("os.name").contains("Windows")) {
+                starter = "cmd.exe";
```
srnagar: Use string constants.

sdk/identity/azure-identity/src/test/java/com/azure/identity/AzureCliCredentialTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+import com.azure.core.credential.TokenRequestContext;
+import com.azure.identity.implementation.IdentityClient;
+import com.azure.identity.util.TestUtils;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PowerMockIgnore;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
```
srnagar: Try using just Mockito instead of PowerMockito.

g2vinay: This can be deferred as discussed.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
FredGao-new: https://github.com/Azure/azure-sdk-for-java/blob/41a932e2bc316f58334cd943499991572c283266/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java#L495
In the method `openUrl` of class `IdentityClient`, doing similar things with string, didn't use constants. Is it better to follow that way?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
FredGao-new: https://github.com/Azure/azure-sdk-for-java/blob/41a932e2bc316f58334cd943499991572c283266/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java#L495
In the method `openUrl` of class `IdentityClient`, doing similar things with string, didn't use constants. Is it better to follow that way?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
srnagar: I would suggest creating `private static final String WINDOWS_STARTER = "cmd.exe"` and then use `starter = WINDOWS_STARTER` here.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
srnagar: same as above.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -99,6 +106,72 @@
...
+                line = reader.readLine();
+                if (line == null) {
+                    break;
+                }
+                if(line.startsWith("'az' is not recognized") || line.matches("(.*)az:(.*)not found")) {
+                    throw logger.logExceptionAsError(new ClientAuthenticationException("Azure CLI not installed", null));
+                }
+                output.append(line);
+            }
+            reader.close();
```
srnagar: Close should be done in `finally` block or use try-with-resources.

sdk/identity/azure-identity/src/test/java/com/azure/identity/AzureCliCredentialTest.java
```diff
@@ -0,0 +1,104 @@
...
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+
+
```
srnagar: nit: extra new line.

PR #6855, opened 2019-12-13
File lease

rickle-msft:
I'll revert the stuff I changed in pom.client.xml, add recordings, etc. in a final commit after approval. For some reason that stuff was necessary to build and run tests.

### DIFF
--- a/sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/specialized/LeaseAsyncClientJavaDocCodeSnippets.java
+++ b/sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/specialized/LeaseAsyncClientJavaDocCodeSnippets.java
@@ -3,10 +3,7 @@
 
 package com.azure.storage.file.share.specialized;
 
-import com.azure.core.http.RequestConditions;
 import com.azure.storage.file.share.ShareFileClientBuilder;
-
-import java.time.OffsetDateTime;
 
 
 public class LeaseAsyncClientJavaDocCodeSnippets {
--- a/sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/specialized/LeaseClientJavaDocCodeSnippets.java
+++ b/sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/specialized/LeaseClientJavaDocCodeSnippets.java
@@ -3,12 +3,10 @@
 
 package com.azure.storage.file.share.specialized;
 
-import com.azure.core.http.RequestConditions;
 import com.azure.core.util.Context;
 import com.azure.storage.file.share.ShareFileClientBuilder;
 
 import java.time.Duration;
-import java.time.OffsetDateTime;
 
 public class LeaseClientJavaDocCodeSnippets {
     private ShareLeaseClient client = new ShareLeaseClientBuilder()
@@ -91,10 +89,12 @@ public class LeaseClientJavaDocCodeSnippets {
      * Code snippets for {@link ShareLeaseClient#changeLeaseWithResponse(String, Duration, Context)}
      */
     public void changeLeaseWithResponseCodeSnippets() {
-        // BEGIN: com.azure.storage.file.share.specialized.ShareLeaseClient.changeLeaseWithResponse#String-Duration-Context
+        // BEGIN: com.azure.storage.file.share.specialized.ShareLeaseClient
+        // .changeLeaseWithResponse#String-Duration-Context
         System.out.printf("Changed lease ID is %s%n",
             client.changeLeaseWithResponse("proposedId", timeout, new Context(key, value))
                 .getValue());
-        // END: com.azure.storage.file.share.specialized.ShareLeaseClient.changeLeaseWithResponse#String-Duration-Context
+        // END: com.azure.storage.file.share.specialized.ShareLeaseClient
+        // .changeLeaseWithResponse#String-Duration-Context
     }
 }
--- a/sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/ShareAsyncJavaDocCodeSamples.java
+++ b/sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/ShareAsyncJavaDocCodeSamples.java
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 package com.azure.storage.file.share;
 
-import com.azure.core.util.Context;
 import com.azure.storage.common.StorageSharedKeyCredential;
 import com.azure.storage.file.share.models.ShareAccessPolicy;
 import com.azure.storage.file.share.models.ShareFileHttpHeaders;
--- a/sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/ShareDirectoryAsyncJavaDocCodeSamples.java
+++ b/sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/ShareDirectoryAsyncJavaDocCodeSamples.java
@@ -3,8 +3,8 @@
 package com.azure.storage.file.share;
 
 import com.azure.storage.common.StorageSharedKeyCredential;
-import com.azure.storage.file.share.models.ShareFileHttpHeaders;
 import com.azure.storage.file.share.models.NtfsFileAttributes;
+import com.azure.storage.file.share.models.ShareFileHttpHeaders;
 import com.azure.storage.file.share.models.ShareRequestConditions;
 
 import java.time.LocalDateTime;
@@ -15,7 +15,8 @@ import java.util.EnumSet;
 import java.util.Map;
 
 /**
- * Contains code snippets when generating javadocs through doclets for {@link ShareDirectoryClient} and {@link ShareDirectoryAsyncClient}.
+ * Contains code snippets when generating javadocs through doclets for {@link ShareDirectoryClient} and {@link
+ * ShareDirectoryAsyncClient}.
  */
 public class ShareDirectoryAsyncJavaDocCodeSamples {
 
@@ -35,6 +36,7 @@ public class ShareDirectoryAsyncJavaDocCodeSamples {
 
     /**
      * Generates code sample for creating a {@link ShareDirectoryAsyncClient} with SAS token.
+     *
      * @return An instance of {@link ShareDirectoryAsyncClient}
      */
     public ShareDirectoryAsyncClient createAsyncClientWithSASToken() {
@@ -50,6 +52,7 @@ public class ShareDirectoryAsyncJavaDocCodeSamples {
 
     /**
      * Generates code sample for creating a {@link ShareDirectoryAsyncClient} with SAS token.
+     *
      * @return An instance of {@link ShareDirectoryAsyncClient}
      */
     public ShareDirectoryAsyncClient createAsyncClientWithCredential() {
@@ -65,7 +68,9 @@ public class ShareDirectoryAsyncJavaDocCodeSamples {
     }
 
     /**
-     * Generates code sample for creating a {@link ShareDirectoryAsyncClient} with {@code connectionString} which turns into {@link StorageSharedKeyCredential}
+     * Generates code sample for creating a {@link ShareDirectoryAsyncClient} with {@code connectionString} which turns
+     * into {@link StorageSharedKeyCredential}
+     *
      * @return An instance of {@link ShareDirectoryAsyncClient}
      */
     public ShareDirectoryAsyncClient createAsyncClientWithConnectionString() {
@@ -87,7 +92,8 @@ public class ShareDirectoryAsyncJavaDocCodeSamples {
         ShareDirectoryAsyncClient shareDirectoryAsyncClient = createAsyncClientWithSASToken();
         // BEGIN: com.azure.storage.file.share.ShareDirectoryAsyncClient.create
         shareDirectoryAsyncClient.create().subscribe(
-            response -> { },
+            response -> {
+            },
             error -> System.err.print(error.toString()),
             () -> System.out.println("Completed creating the directory!")
         );
@@ -95,7 +101,8 @@ public class ShareDirectoryAsyncJavaDocCodeSamples {
     }
 
     /**
-     * Generates a code sample for using {@link ShareDirectoryAsyncClient#createWithResponse(FileSmbProperties, String, Map)}
+     * Generates a code sample for using {@link ShareDirectoryAsyncClient#createWithResponse(FileSmbProperties, String,
+     * Map)}
      */
     public void createDirectoryWithResponseAsync() {
         ShareDirectoryAsyncClient shareDirectoryAsyncClient = createAsyncClientWithSASToken();
@@ -123,11 +130,13 @@ public class ShareDirectoryAsyncJavaDocCodeSamples {
     }
 
     /**
-     * Generates a code sample for using {@link ShareDirectoryAsyncClient#createSubdirectoryWithResponse(String, FileSmbProperties, String, Map)}
+     * Generates a code sample for using {@link ShareDirectoryAsyncClient#createSubdirectoryWithResponse(String,
+     * FileSmbProperties, String, Map)}
      */
     public void createSubdirectoryAsyncMaxOverload() {
         ShareDirectoryAsyncClient shareDirectoryAsyncClient = createAsyncClientWithSASToken();
-        // BEGIN: com.azure.storage.file.share.ShareDirectoryAsyncClient.createSubdirectoryWithResponse#String-FileSmbProperties-String-Map
+        // BEGIN: com.azure.storage.file.share.ShareDirectoryAsyncClient
+        // .createSubdirectoryWithResponse#String-FileSmbProperties-String-Map
         FileSmbProperties smbProperties = new FileSmbProperties();
         String filePermission = "filePermission";
         Map<String, String> metadata = Collections.singletonMap("directory", "metadata");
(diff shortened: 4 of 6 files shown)


pom.client.xml
```diff
@@ -777,8 +777,8 @@
                 <arg>-Xlint:try</arg>
                 <arg>-Xlint:unchecked</arg>
                 <arg>-Xlint:varargs</arg>
-                <arg>-Xlint:-module</arg> <!-- FIXME: this is required for now as it introduces a build failure -->
-                <arg>-Xlint:-requires-transitive-automatic</arg> <!-- FIXME: this is required for now as it introduces a build failure -->
+                <!--arg>-Xlint:-module</arg--> <!-- FIXME: this is required for now as it introduces a build failure -->
```
rickle-msft: Revert all this stuff

rickle-msft: Add recordings.

alzimmermsft:
Few random file changes for Application Insights, are those meant to be in this PR?

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareAsyncClient.java
```diff
@@ -780,21 +790,59 @@ public ShareFileAsyncClient getFileClient(String filePath) {
...
+     * If the share or parent directory does not exist.
+     * </li>
+     * <li>
+     * An attempt to create file on a share snapshot will fail with 400 (InvalidQueryParameterValue).
+     * </li>
+     * </ul>
+     */
+    public Mono<Response<ShareFileAsyncClient>> createFileWithResponse(String fileName, long maxSize,
+        ShareFileHttpHeaders httpHeaders, FileSmbProperties smbProperties, String filePermission,
+        Map<String, String> metadata, ShareRequestConditions requestConditions) {
```
alzimmermsft: Is File leasing new as of `2019-07-07`? 

I heard there were discussions about potentially having newer APIs throw if they are supported by older service versions and the client is using said version. Would this need to throw if the client was using service version `2019-02-02`?

rickle-msft: It is new for 2019-07-07. I haven't been clear about the story for specifying the versions. I heard what you mentioned. I also heard something about just clearing the list every time there's a service change that's not supported by old versions? Where can we go to get clarity on that?

rickle-msft:
Changes to app insights aren't supposed to be in there. I'll revert them in the same commit where I undo pom changes and add recordings. I'm not sure where those came from

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/DirectoryAsyncAPITests.groovy
```diff
@@ -541,7 +542,26 @@ class DirectoryAsyncAPITests extends APISpec {
         "fileName"  | -1      | null                                                  | testMetadata                          | ShareErrorCode.OUT_OF_RANGE_INPUT
         "fileName"  | 1024    | new ShareFileHttpHeaders().setContentMd5(new byte[0]) | testMetadata                          | ShareErrorCode.INVALID_HEADER_VALUE
         "fileName"  | 1024    | null                                                  | Collections.singletonMap("", "value") | ShareErrorCode.EMPTY_METADATA_KEY
+    }
+
+    def "Create file lease"() {
+        given:
+        primaryDirectoryAsyncClient.create().block()
+        def leaseId = createLeaseClient(primaryDirectoryAsyncClient.getFileClient("testCreateFile")).acquireLease().block()
```
gapra-msft: Are all the other tests static names too? I thought we're trying to move away from that to prevent tests from stepping on each other (like one test deletes and then the next test fails because its trying to create it)

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareFileAsyncClient.java
```diff
@@ -249,18 +297,45 @@ public ShareServiceVersion getServiceVersion() {
...
+     * @param pollInterval Duration between each poll for the copy status. If none is specified, a default of one second
+     * is used.
+     * @param metadata Optional name-value pairs associated with the file as metadata. Metadata names must adhere to the
+     * naming rules.
+     * @param requestConditions {@link ShareRequestConditions}
+     * @return A {@link PollerFlux} that polls the file copy operation until it has completed or has been cancelled.
+     * @see <a href="https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/">C# identifiers</a>
+     */
+    public PollerFlux<ShareFileCopyInfo, Void> beginCopy(String sourceUrl, Map<String, String> metadata,
+        Duration pollInterval, ShareRequestConditions requestConditions) {
```
gapra-msft: Should this be destinationRequestConditions?

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/DirectoryAsyncAPITests.groovy
rickle-msft: This is standard in file tests right now. We can use #6933 to track this.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareFileProperties.java
```diff
@@ -77,6 +80,55 @@ public ShareFileProperties(final String eTag, final OffsetDateTime lastModified,
         final OffsetDateTime copyCompletionTime, final String copyStatusDescription, final String copyId,
         final String copyProgress, final String copySource, final CopyStatusType copyStatus,
         final Boolean isServerEncrypted, final FileSmbProperties smbProperties) {
+        this(eTag, lastModified, metadata,fileType, contentLength, contentType, contentMd5, contentEncoding,
```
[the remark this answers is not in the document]
rickle-msft: I think based on the discussion around deprecation we had on the cpkn thread, I'll leave this as not deprecated

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareFileAsyncClient.java
rickle-msft: Good call. I have changed it

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareFileProperties.java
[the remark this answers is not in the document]
rickle-msft: @JonathanGiles can you comment on if this would be a case for deprecation?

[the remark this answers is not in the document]
JonathanGiles: It's hard to provide much useful judgement here. It comes down to whether you want users to call the API, if it provides any utility or convenience to the user, if it is behaving in an expected manner, etc. What does it mean to call this method without the lease information, and what would existing users who call this method do (would they all pass in null as you have, or would they provide non-null args which would give them a better UX).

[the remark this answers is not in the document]
rickle-msft: Customers shouldn't really ever be constructing this object themselves. It's only public because it is constructed across package boundaries. Hence, we don't want them to call either the original or the new one, it doesn't provide any utility or convenience. It does behave in the expected manner.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareAsyncClient.java
rickle-msft: We had a conversation about this this morning. I think we are going to come up with a more complete approach to doing service version validation and apply that to all Jul19 features at once, so I will defer the check until that time.

sdk/storage/azure-storage-file-share/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release History
 
+## Version X.X.X (XXXX-XX-XX)
```
weshaggard: The general guidance here is to put the specific version we have in the repo followed by (unreleased) see https://github.com/Azure/azure-sdk/blob/master/docs/policies/releases.md#changelog-guidance

rickle-msft: We haven't run the version update script for this feature branch yet, so the pom still reads 12.0.0. Because we branched a couple weeks ago, we're behind on some of these updates to master. I think our goal is to get the code for the features into the feature branch and then we will rebase and update these sort of infrastructural things for the whole feature branch at once.

weshaggard: I highly suggest you update the version in this branch as well. It is never a good idea to be building the same version as what is published as they will cause conflicts in different modes and you will never know which one you are getting. 

cc @JimSuplizio

gapra-msft: Created a PR here to update versions in our feature branch https://github.com/Azure/azure-sdk-for-java/pull/6981

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareRequestConditions.java
```diff
@@ -0,0 +1,37 @@
...
+ * present. These conditions are entirely optional. The entire object or any of its properties may be set to null when
+ * passed to a method to indicate that those conditions are not desired. Please refer to the type of each field for more
+ * information on those particular access conditions.
+ */
+@Fluent
+public class ShareRequestConditions {
+    private String leaseId;
+
+    /**
+     * Gets the lease ID that blobs and containers must match.
```
alzimmermsft: S
```suggestion
     * Gets the lease ID that shares must match.
```

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareRequestConditions.java
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @return The lease ID that blobs and containers must match.
```
alzimmermsft:
```suggestion
     * @return The lease ID that shares must match.
```

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareRequestConditions.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public String getLeaseId() {
+        return leaseId;
+    }
+
+    /**
+     * Optionally limits requests to blobs and containers that match the lease ID.
```
alzimmermsft:
```suggestion
     * Optionally limits requests to shares that match the lease ID.
```

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareRequestConditions.java
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @param leaseId Lease ID that blobs and containers must match.
```
alzimmermsft:
```suggestion
     * @param leaseId Lease ID that shares must match.
```

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/specialized/ShareLeaseAsyncClient.java
```diff
@@ -0,0 +1,252 @@
...
+ * This client acts as a supplement to that client and only handles leasing operations.
+ *
+ * <p><strong>Instantiating a ShareLeaseAsyncClient</strong></p>
+ *
+ * {@codesnippet com.azure.storage.file.share.specialized.ShareLeaseClientBuilder.asyncInstantiation}
+ *
+ * <p>View {@link ShareLeaseClientBuilder this} for additional ways to construct the client.</p>
+ *
+ * <p>For more information about leasing see the
+ * <a href="https://docs.microsoft.com/en-us/rest/api/storageservices/lease-file">Azure Docs</a>.
```
alzimmermsft: I'm guessing this page will be live once there is general availability?

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/specialized/ShareLeaseAsyncClient.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /**
+     * Breaks the previously acquired lease, if it exists. Leases will break immediately.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.file.share.specialized.ShareLeaseAsyncClient.breakLeaseWithResponse}
+     *
+     * @return A reactive response containing the remaining time in the broken lease in seconds.
```
alzimmermsft: The portion about remaining time can be removed since these leases break immediately and are infinite. Also update `breakLease` to talk about the lease breaking immediately.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/specialized/ShareLeaseClient.java
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Breaks the previously acquired lease, if it exists. Leases will break immediately.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.file.share.specialized.ShareLeaseClient.breakLeaseWithResponse#Duration-Context}
+     *
+     * @param timeout An optional timeout value beyond which a {@link RuntimeException} will be raised.
+     * @param context Additional context that is passed through the Http pipeline during the service call.
+     * @return The remaining time in the broken lease in seconds.
```
alzimmermsft: Same comment as the async client about leases being infinite and breaking immediately.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/specialized/package-info.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Package containing specialized clients for Azure Storage Blobs.
```
alzimmermsft:
```suggestion
 * Package containing specialized clients for Azure Storage File Shares.
```

sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/specialized/LeaseAsyncClientJavaDocCodeSnippets.java
```diff
@@ -0,0 +1,90 @@
...
+        // END: com.azure.storage.file.share.specialized.ShareLeaseAsyncClient.releaseLeaseWithResponse
+    }
+
+    /**
+     * Code snippets for {@link ShareLeaseAsyncClient#breakLeaseWithResponse()}
+     */
+    public void breakLeaseWithResponseCodeSnippets() {
+        // BEGIN: com.azure.storage.file.share.specialized.ShareLeaseAsyncClient.breakLeaseWithResponse
+        client.breakLeaseWithResponse().subscribe(response ->
+            System.out.printf("The broken lease has %d seconds remaining on the lease", response.getValue()));
```
alzimmermsft: Probably should just print `The lease has been successfully broken.` given that they break immediately and are infinite so there is no remaining time on it.

sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/specialized/LeaseAsyncClientJavaDocCodeSnippets.java
```diff
@@ -0,0 +1,90 @@
...
+        // END: com.azure.storage.file.share.specialized.ShareLeaseAsyncClient.releaseLease
+    }
+
+    /**
+     * Code snippets for {@link ShareLeaseAsyncClient#breakLease()}
+     */
+    public void breakLeaseCodeSnippet() {
+        // BEGIN: com.azure.storage.file.share.specialized.ShareLeaseAsyncClient.breakLease
+        client.breakLease().subscribe(response ->
+            System.out.printf("The broken lease has %d seconds remaining on the lease", response));
```
alzimmermsft: Same as the other codesnippet

sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/specialized/LeaseClientJavaDocCodeSnippets.java
```diff
@@ -0,0 +1,100 @@
...
+                .getStatusCode());
+        // END: com.azure.storage.file.share.specialized.ShareLeaseClient.releaseLeaseWithResponse#Duration-Context
+    }
+
+    /**
+     * Code snippets for {@link ShareLeaseClient#breakLeaseWithResponse(Duration, Context)}
+     */
+    public void breakLeaseWithResponseCodeSnippets() {
+        // BEGIN: com.azure.storage.file.share.specialized.ShareLeaseClient.breakLeaseWithResponse#Duration-Context
+        client.breakLeaseWithResponse(timeout, new Context(key, value));
```
alzimmermsft: Probably should have a print statement similar to the `breakLease` codesnippet in this file.

gapra-msft:
Created an issue here https://github.com/Azure/azure-sdk-for-java/issues/7018 regarding all doc issues in this PR

PR #6877, opened 2019-12-16
[After Jan 7] Removing dependency on java.desktop module.

### LINKED ISSUE #5510: azure-identity requires java.desktop to open browser window
It would be much better if we did't need to have a dependency on java.desktop to open a browser window. This won't always necessarily work either - e.g. having this dependency might prevent identity from working on mobile platforms. We should consider having a pluggable system or other approach to not force a java.desktop dependency directly on azure-identity.

JonathanGiles:
Fixes #5510 by removing the java.desktop module dependency and introducing new code to open the users browser. If this fails, a message is logged with the URL they should browse to manually.

### DIFF
--- a/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
+++ b/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
@@ -45,6 +45,7 @@ import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Random;
 import java.util.Scanner;
 import java.util.UUID;
@@ -494,7 +495,7 @@ public class IdentityClient {
     private void openUrl(String url) throws IOException {
         Runtime rt = Runtime.getRuntime();
 
-        String os = System.getProperty("os.name").toLowerCase();
+        String os = System.getProperty("os.name").toLowerCase(Locale.ROOT);
         if (os.contains("win")) {
             rt.exec("rundll32 url.dll,FileProtocolHandler " + url);
         } else if (os.contains("mac")) {


jianghaolu:
Tested on Windows 10 and MacOS

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -488,4 +490,19 @@ private static Proxy proxyOptionsToJavaNetProxy(ProxyOptions options) {
                 return new Proxy(Type.HTTP, options.getAddress());
         }
     }
+
+    private void openUrl(String url) throws IOException {
+        Runtime rt = Runtime.getRuntime();
+
+        String os = System.getProperty("os.name").toLowerCase();
```
jianghaolu: of non-localized String.toUpperCase() or String.toLowerCase() in com.azure.identity.implementation.IdentityClient.openUrl(String) [com.azure.identity.implementation.IdentityClient] At IdentityClient.java:[line 497] DM_CONVERT_CASE

JonathanGiles:
Just suppress this issue so that you can get a clean build - then merge!

JonathanGiles:
@jianghaolu Please can we get this merged sometime in January? Thanks!

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Changed to `toLowerCase(Locale.ROOT)`

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -488,4 +491,19 @@ private static Proxy proxyOptionsToJavaNetProxy(ProxyOptions options) {
...
+    private void openUrl(String url) throws IOException {
+        Runtime rt = Runtime.getRuntime();
+
+        String os = System.getProperty("os.name").toLowerCase(Locale.ROOT);
+        if (os.contains("win")) {
+            rt.exec("rundll32 url.dll,FileProtocolHandler " + url);
+        } else if (os.contains("mac")) {
+            rt.exec("open " + url);
+        } else if (os.contains("nix") || os.contains("nux")) {
+            rt.exec("xdg-open " + url);
```
g2vinay: we need to look into finding a way to write and automate tests for this platform specific commands and browser end result.

alzimmermsft: JUnit offers attributes which can perform per test prerequisites, could look into that.

https://keyholesoftware.com/2018/02/12/disabling-filtering-tests-junit-5/

jianghaolu: Is there a way to test opening a browser?

PR #6878, opened 2019-12-16
Added support for CPK-N

gapra-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/AppendBlobItem.java
```diff
@@ -31,7 +32,9 @@
      * @param encryptionKeySha256 The encryption key used to encrypt the page blob.
      * @param blobAppendOffset The offset at which the block was committed to the block blob.
      * @param blobCommittedBlockCount The number of committed blocks in the block blob.
+     * @deprecated Please use the overload.
      */
+    @Deprecated
```
rickle-msft: We should check on our deprecation policy here. This strikes me as different from what we're doing with sas because over there, we are trying to actively move people away from an inferior workflow, hence deprecation. Here, it's not really a big deal if customers continue to use this overload because it's pretty innocuous and behaves as expected. That said, maybe we want to be proactive about paring down the number of overloads when new functionality is introduced. 

@alzimmermsft  can you weigh in here on what the guidelines are?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/AppendBlobItem.java
```diff
@@ -40,6 +43,32 @@ public AppendBlobItem(final String eTag, final OffsetDateTime lastModified, fina
         this.contentMd5 = CoreUtils.clone(contentMd5);
         this.isServerEncrypted = isServerEncrypted;
```
rickle-msft: I think it is preferable if this overload calls into the next one.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/CPKNTest.groovy
```diff
@@ -0,0 +1,391 @@
+package com.azure.storage.blob
+
+import com.azure.core.http.policy.HttpLogDetailLevel
+import com.azure.core.http.policy.HttpLogOptions
+import com.azure.core.test.TestMode
+import com.azure.core.util.Configuration
+import com.azure.storage.blob.models.BlobContainerEncryptionScope
+import com.azure.storage.blob.models.BlobItem
+import com.azure.storage.blob.models.BlobStorageException
+import com.azure.storage.blob.models.CustomerProvidedKey
```
rickle-msft: I would prefer to have these test methods sit next to the other tests for the same api (i.e. "append blob create with cpkn" next to the other "append blob create" tests). For better or worse, tests have typically been grouped by api rather than by feature, and I think it'll generally better if we keep to one organizational strategy. If you have a compelling reason to put all the cpkn tests in one file, then I'm open to hearing that.

gapra-msft: The reason I chose to do this was because this feature is similar to CPK and CPK has its own test class like this

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/AppendBlobItem.java
JonathanGiles: I generally think of API as something you need to pay rent on. Each public method incurs some small cost - maintenance, testing, documentation, samples, and cognitive load for end users. It's vanishingly small, but not zero cost. 

I don't want us to be in the business of aggressively deprecating anything we no longer like the look of, but I also don't want us keeping around duplicative or redundant API if it is truly unnecessary for any particular use case. Eagerly applying the deprecated annotations in this case should result in less teeth gnashing later, as users will have had potentially many more releases of deprecation warnings before any such breaking changes are applied (when the deprecated API is removed). 

The final point I'll make is that I wouldn't want our libraries to become vast wastelands of deprecation, because that signals to users our libraries are not as stable as they would like, or that we haven't put in enough thought up front to have a library that will stand the tests of time.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerClientBuilder.java
```diff
@@ -99,6 +103,11 @@ public BlobContainerClient buildClient() {
     public BlobContainerAsyncClient buildAsyncClient() {
         BuilderHelper.httpsValidation(customerProvidedKey, "customer provided key", endpoint, logger);
 
+        if (Objects.nonNull(customerProvidedKey) && Objects.nonNull(encryptionScope)) {
+            throw logger.logExceptionAsError(new IllegalArgumentException("Customer provided key and encryption"
+                + "scope cannot both be set"));
```
jaschrep-msft: Spacing again.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClientBuilder.java
```diff
@@ -92,12 +96,18 @@ public BlobServiceAsyncClient buildAsyncClient() {
                 + "anonymously. Please provide a form of authentication"));
         }
 
+        if (Objects.nonNull(customerProvidedKey) && Objects.nonNull(encryptionScope)) {
+            throw logger.logExceptionAsError(new IllegalArgumentException("Customer provided key and encryption"
+                + "scope cannot both be set"));
```
jaschrep-msft: Spacing.

PR #6882, opened 2019-12-16
More Robust BlobBatch Implementation

alzimmermsft:
Makes the `BlobBatch` implementation more robust by adding helper classes `BlobBatchOperation` and `BlobBatchOperationInfo` which maintain state of a batch in a more centralized way. `BlobBatchOperation` allows for the removal of generating `Content-ID`s when a batch operation is submitted, instead it is deferred until the batch is submitted. `BlobBatchOperationInfo` allows for the separation of the batch submission and the batch itself, allowing for the batch to maintain less state about the batch it contains.

### DIFF
--- a/sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchOperationInfo.java
+++ b/sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchOperationInfo.java
@@ -3,6 +3,7 @@
 
 package com.azure.storage.blob.batch;
 
+import com.azure.core.annotation.Immutable;
 import com.azure.core.http.HttpRequest;
 import com.azure.core.util.CoreUtils;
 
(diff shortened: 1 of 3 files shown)


sdk/storage/azure-storage-blob-batch/src/main/java/com/azure/storage/blob/batch/BlobBatchOperationInfo.java
```diff
@@ -0,0 +1,142 @@
...
+        String method = request.getHttpMethod().toString();
+        String urlPath = request.getUrl().getPath();
+        String urlQuery = request.getUrl().getQuery();
+        if (!CoreUtils.isNullOrEmpty(urlQuery)) {
+            urlPath = urlPath + "?" + urlQuery;
+        }
+        appendWithNewline(batchRequestBuilder, String.format(OPERATION_TEMPLATE, method, urlPath, HTTP_VERSION));
+
+        request.getHeaders().stream()
+            .filter(header -> !X_MS_VERSION.equalsIgnoreCase(header.getName()))
```
srnagar: Add a comment as to why this header is filtered.

PR #6883, opened 2019-12-16
Remove Surefire Rerun Failed Tests

alzimmermsft:
Removes the surefire command line configuration to re-run failing tests up to three times before marking them as failing. This is done as this potentially masks transient test failures.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-tests-pre-sdk.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-tests-pre-sdk.yml
@@ -15,7 +15,7 @@ jobs:
         displayName: 'Run Live tests'
         inputs:
           mavenPomFile: ${{ parameters.PomFilePath }}
-          options: '--batch-mode -Dmaven.wagon.http.pool=false -Dsurefire.rerunFailingTestsCount=3'
+          options: '--batch-mode -Dmaven.wagon.http.pool=false'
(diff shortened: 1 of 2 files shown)


alzimmermsft:
@mikeharder and @danieljurek This removes re-running failed tests three times until they are officially considered failed.

sima-zhu:
I am testing under this. Storage has a per-test timeout. It is better to leave this as some of the tests get executed after re-run. Can we try to less time instead of removing this? I think at lease set to 1

sima-zhu:
Another suggestion is to set different variable for different service. Storage is better to have this.

mikeharder:
If retries are needed in the short-term to keep tests green, I'm fine with it.  But the medium-term goal should be to eliminate all test retries, since they can hide real issues in test or product code.  If tests are failing for any reason we need to understand the root cause and the appropriate fix.  I stated earlier the goal of 90% of live test runs passing (all tests on all platforms), and this should include no test retries.

Retries performed within the SDK product code itself are of course fine.

JimSuplizio:
@mikeharder It's unclear whether the retry mechanism helps or hurts based upon the number of logs that we've seen showing the retry mechanism, itself, failing.

alzimmermsft:
> 
> 
> I have tested under this. Storage has a per-test timeout, there are some transient issue kill by timeout have a chance to get executed(Retry did not run if the test kill by timeout). Can we try to use smaller number instead of removing this? At lease set to 1

I don't know if the right approach would be retrying a test failing due to the global timeout, that was put in place to better capture instances where a test just deadlocks/dies and locks the test run until the Azure Pipelines timeout triggers. I'd rather see these fail the test run completely to make them more of the red flag they should be.

PR #6886, opened 2019-12-17
Added canLogAtLevel API in logging.

### LINKED ISSUE #6918: [FEATURE REQ] support performant way of (not) logging in ClientLogger
For fastest way of (not) logging, please add support for
* `clientLogger.isDebugEnabled()`
* `clientLogger.isTraceEnabled()`
* and similar other levels.

(similar to what slf4j provides: http://www.slf4j.org/faq.html#logging_performance)

ask from cosmosdb sdk. discussed offline with Jonathan

sima-zhu:
Fixes #6918


JonathanGiles:
You need to make LogLevel public API.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -173,6 +173,17 @@ private RuntimeException logException(RuntimeException runtimeException, LogLeve
         return runtimeException;
     }
 
+    /*
+     * Determines if the environment and logger support logging at the given log level.
+     *
+     * @param logLevel Logging level for the log message.
+     * @return Flag indicating if logger are configured to support logging at the given log level.
+     */
+    public boolean canLogAtLevel(LogLevel logLevel) {
```
alzimmermsft: Should this be made static and accept a `Class` or `String`? Right now this wouldn't get us too much as almost all use cases of `ClientLogger` are private. From what I understand this is meant to allow someone to determine with their own logic whether or not logging should happen.

sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
```diff
@@ -276,6 +277,18 @@ public void logExceptionAsErrorStackTrace() {
         assertTrue(logValues.contains(runtimeException.getStackTrace()[0].toString()));
     }
 
+    @Test
```
alzimmermsft: Let's just make this a parameterized test that accepts an `int logLevelToConfigure`, `LogLevel logLevelToValidate`, and `boolean expected` to get better coverage on this.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -173,6 +173,17 @@ private RuntimeException logException(RuntimeException runtimeException, LogLeve
         return runtimeException;
     }
 
+    /*
+     * Determines if the environment and logger support logging at the given log level.
+     *
+     * @param logLevel Logging level for the log message.
+     * @return Flag indicating if logger are configured to support logging at the given log level.
```
alzimmermsft:
```suggestion
     * @return Flag indicating if the environment and logger support logging at the given log level.
```

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -173,6 +173,17 @@ private RuntimeException logException(RuntimeException runtimeException, LogLeve
         return runtimeException;
     }
 
+    /*
+     * Determines if the environment and logger support logging at the given log level.
+     *
+     * @param logLevel Logging level for the log message.
```
alzimmermsft:
```suggestion
     * @param logLevel The {@link LogLevel} being validated as supported.
```

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
JonathanGiles: This is to enable the pattern:

```java
if (logger.canLogAtLevel(ERROR)) {
    logger.error("Output string {}", complexOperationToCalculateValue());
}
```

alzimmermsft: Correct, but I am a bit confused on who this change is targeting, is this for us to use internally where we have our `private final ClientLogger`s or externally from customer code?

JonathanGiles: Correct

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,80 @@
...
+        this.numericValue = numericValue;
+        this.allowedLogLevelVariables = new HashSet<>(Arrays.asList(allowedLogLevelVariables));
+    }
+
+    /**
+     * Converts the log level into a numeric representation used for comparisons.
+     *
+     * @return The numeric representation of the log level.
+     */
+    public int toNumeric() {
```
JonathanGiles: I would propose calling this `getLogLevel()`

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,80 @@
...
+        return numericValue;
+    }
+
+    /**
+     * Converts the log level into string representations used for comparisons.
+     *
+     * @return The string representations of the log level.
+     */
+    public Set<String> getAllowedLogLevels() {
```
JonathanGiles: This doesn't need to be public API

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * Convert log level value to LogLevel.
+     *
+     * @param logLevelVal The log level value which needs to convert
+     * @return The LogLevel Enum.
+     */
+    public static LogLevel getLogLevel(String logLevelVal) {
+        return Arrays.stream(LogLevel.values()).filter(logLevel ->
```
alzimmermsft: Any way to convert this into a `Map`? Right now this would require every call to `getLogLevel` (which there will be many) to iterate over the the enum values.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,80 @@
...
+        return allowedLogLevelVariables;
+    }
+
+    /**
+     * Convert log level value to LogLevel.
+     *
+     * @param logLevelVal The log level value which needs to convert
+     * @return The LogLevel Enum.
+     */
+    public static LogLevel getLogLevel(String logLevelVal) {
```
JonathanGiles: This would be better named `fromString` or similar.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,80 @@
...
+     * Converts the log level into string representations used for comparisons.
+     *
+     * @return The string representations of the log level.
+     */
+    public Set<String> getAllowedLogLevels() {
+        return allowedLogLevelVariables;
+    }
+
+    /**
+     * Convert log level value to LogLevel.
```
alzimmermsft:
```suggestion
     * Converts the passed log level string to the corresponding {@link LogLevel}.
```

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
alzimmermsft: Does this need to be publicly accessible?

sima-zhu: No. Will make it private

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Will change to `getLogLevelFromString`

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,80 @@
...
+    public static LogLevel getLogLevel(String logLevelVal) {
+        return Arrays.stream(LogLevel.values()).filter(logLevel ->
+            isValueMatch(logLevel, logLevelVal)).findFirst().orElse(null);
+    }
+
+    /**
+     * Check if the log level value matches the LogLevel enum.
+     */
+    private static boolean isValueMatch(LogLevel logLevel, String logLevelVal) {
+        return String.valueOf(logLevel.toNumeric()).equals(logLevelVal)
```
srnagar: In the string varargs, if you include the numeric value too, you don't need this additional check `String.valueOf(logLevel.toNumeric()).equals(logLevelVal)`.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Do you have any suggestions? We currently have ERROR(1, "error", "err")
I can have a map of `{1: ERROR} ` Numeric value to LogLevel
The string array value needs to iterate anyway. Let me know if you have good idea.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Will turn numeric to string.

sima-zhu: Tried to make the changes. We still need to have numeric, which is using in ClientLogger to check whether log or not.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Will make it private

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: I will change to `getNumericLogLevel`

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
JonathanGiles: This name is even worse. I was proposing `fromString` being the entire name.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
JonathanGiles: Just have a static Map<Object, LogLevel>.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
JonathanGiles: `getLogLevel()` is sufficient.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
JonathanGiles: Srikanta was saying to have the integer and a string representation, so you would do, e.g. VERBOSE(4, "4", "verbose"). Then you don't need to do this extra check.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Turn them into Map.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: I made to VERBOSE("4", "verbose"), and toNumeric API use the first value in string array.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Done

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -173,6 +172,17 @@ private RuntimeException logException(RuntimeException runtimeException, LogLeve
...
 
+    /**
+     * Determines if the environment and logger support logging at the given log level.
+     *
+     * @param logLevel The {@link LogLevel} being validated as supported.
+     * @return Flag indicating if the environment and logger support logging at the given log level.
+     */
+    public boolean canLogAtLevel(LogLevel logLevel) {
+        LogLevel environmentLoggingLevel = getEnvironmentLoggingLevel();
+        return canLogAtLevel(logLevel, environmentLoggingLevel);
```
JonathanGiles: No need to create a local variable here - just call `getEnvironmentLoggingLevel()` directly in the second line.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,77 @@
...
+        this.allowedLogLevelVariables = allowedLogLevelVariables;
+    }
+
+    /**
+     * Converts the log level into a numeric representation used for comparisons.
+     *
+     * @return The numeric representation of the log level.
+     */
+    public int getLogLevel() {
+        return Integer.parseInt(allowedLogLevelVariables[0]);
```
JonathanGiles: This isn't good, as you're now doing this parse every time we need to know the numeric value, which is very frequently. You should do what Srikanta proposed previously and have the int as well as the string values.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,77 @@
...
+     *
+     * @return The string representations of the log level.
+     */
+    private String[] getAllowedLogLevels() {
+        return allowedLogLevelVariables;
+    }
+
+    private static final Map<String, LogLevel> LOG_LEVEL_STRING_MAPPER = Arrays.stream(LogLevel.values())
+        .flatMap(logLevel -> Arrays.stream(logLevel.getAllowedLogLevels()).map(v -> Tuples.of(v, logLevel)))
+        .collect(Collectors.toMap(Tuple2::getT1, Tuple2::getT2));
```
JonathanGiles: This is overly complex - have you tried to just build a single static map in this enum similar to what I proposed?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Done

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,77 @@
...
+    public int getLogLevel() {
+        return Integer.parseInt(allowedLogLevelVariables[0]);
+    }
+
+    /**
+     * Converts the log level into string representations used for comparisons.
+     *
+     * @return The string representations of the log level.
+     */
+    private String[] getAllowedLogLevels() {
```
srnagar: Don't see why this method is required. You can directly use the `allowedLogLevelVariables` field instead.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,77 @@
...
+        .collect(Collectors.toMap(Tuple2::getT1, Tuple2::getT2));
+
+    /**
+     * Converts the passed log level string to the corresponding {@link LogLevel}.
+     *
+     * @param logLevelVal The log level value which needs to convert
+     * @return The LogLevel Enum.
+     */
+    public static LogLevel fromString(String logLevelVal) {
+        if (logLevelVal == null) {
```
srnagar: A `null` arg should instead throw an `IllegalArgumentExcpetion` as it's an unexpected/invalid input.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: What does the object include?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,77 @@
...
+            return null;
+        }
+        return LOG_LEVEL_STRING_MAPPER.get(logLevelVal.toLowerCase(Locale.ROOT));
```
srnagar: If an invalid string is provided, this will return `null`. It would be better to throw an exception here with a detailed error message than to return `null` and causing a NPE elsewhere with no guidance on how it can be fixed.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Removed

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Currently, if no logLevel set, default to NOT_SET level.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: Done by latest commit.

sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
```diff
@@ -105,13 +108,17 @@ public void logAtUnsupportedLevel(int logLevel) {
     @ValueSource(ints = { 1, 2, 3, 4 })
     public void logWhenLoggingDisabled(int logLevel) {
```
alzimmermsft: Should rename this test

sima-zhu: change to `logWhenLoggingInvalidNumeric`

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
JonathanGiles: Srikanta is correct, you should throw.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
JonathanGiles: I'm not seeing this... Have you pushed it?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: logLevelVal is null when env is not set. As we discussed, it is an valid case.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
sima-zhu: It is above. If invalid, then throw.
```
if (!LOG_LEVEL_STRING_MAPPER.containsKey(logLevelVal.toLowerCase(Locale.ROOT))) {
            throw new IllegalArgumentException("We currently do not support the log level you set. LogLevel: "
                + logLevelVal);
        }
```

JonathanGiles: I see it now, it must have been cached and showing the older version. You shouldn't do the to lower case twice, create a local variable.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.util.logging;
```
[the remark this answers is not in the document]
JonathanGiles: LogLevel is now exposed through ClientLogger public API.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,82 @@
...
+        return numericValue;
+    }
+
+    /**
+     * Converts the passed log level string to the corresponding {@link LogLevel}.
+     *
+     * @param logLevelVal The log level value which needs to convert
+     * @return The LogLevel Enum.
+     * @throws IllegalArgumentException if the log level value is invalid.
+     */
```
JonathanGiles: This JavaDoc should be made more clear about valid strings, or pointing to the relevant place where the valid strings are listed. You should also note that NOT_SET is returned if null is passed in.

sima-zhu: Done.

JonathanGiles: I don't see this has been done? I still expect to see more clarity somewhere in the javadoc about the valid string values that can be passed into this method.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,90 @@
...
+
+package com.azure.core.util.logging;
+
+import java.util.HashMap;
+import java.util.Locale;
+
+/**
+ * Enum which represent logging levels used in Azure SDKs. 
+ */
+public enum LogLevel {
```
moderakh: Loggers have distinct log levels for trace vs debug. Any reason we don't have equivalent of trace here?

sima-zhu: We have a discussion here, as there are different LogLevel for different logging framework, there is no need to have 1 on 1 corresponding. We are having minimum set here which is enough for current use. We can add more enums if there is need in future.

PR #6925, opened 2019-12-17
Fix Java Links in the root README

JimSuplizio:
There were 4 broken links in the README.md file which show up on the [root docs page](https://azure.github.io/azure-sdk-for-java/). Basically all 3 links that were linking to contributing.md and the link linking to license.txt

### DIFF
--- a/eng/pipelines/docindex.yml
+++ b/eng/pipelines/docindex.yml
@@ -20,8 +20,8 @@ jobs:
 
       - template: eng/pipelines/templates/scripts/replace-relative-links.yml@azure-sdk-tools
         parameters:
-          TargetFolder: $(Build.SourcesDirectory)
-          RootFolder: $(Build.SourcesDirectory)
+          TargetFolder: .
+          RootFolder: .
           BuildSHA: $(Build.SourceVersion)
           RepoId: 'Azure/azure-sdk-for-java'
 


README.md
```diff
@@ -80,7 +80,7 @@ Format of the release tags are `<package-name>_<package-version>`. For more info
 
 ## Contributing
 
-For details on contributing to this repository, see the [contributing guide](CONTRIBUTING.md).
+For details on contributing to this repository, see the [contributing guide](https://github.com/Azure/azure-sdk-for-java/blob/master/CONTRIBUTING.md).
```
weshaggard: I don't think this is the best way to fix these links. We should try to avoid links to master. Instead we should run the same script we run on the other README's that turn relative links into absolute links as part of the publishing step.

JimSuplizio: Those style of links are all gone with the latest PR.

eng/pipelines/docindex.yml
```diff
@@ -13,6 +13,18 @@ jobs:
     pool:
       vmImage: windows-2019
     steps:
+      - task: UsePythonVersion@0
```
weshaggard: @scbedd orthogonal to Jim's change here but is it worth adding the UsePythonVersion in the template instead of needing to call it explicitly? Or do most pipelines already explicitly setup a python version?

eng/pipelines/docindex.yml
```diff
@@ -13,6 +13,18 @@ jobs:
...
+        displayName: 'Use Python 3.6'
+        inputs:
+          versionSpec: '3.6'
+
+      - template: eng/pipelines/templates/scripts/replace-relative-links.yml@azure-sdk-tools
+        parameters:
+          TargetFolder: .
```
weshaggard: Does this support passing in a single file?

JimSuplizio: Unfortunately, no, the script takes directory arguments.

eng/pipelines/docindex.yml
scbedd: The script is designed to be usable on any of the major python versions we leverage. Originally I was hesitant to add to the script itself because we need to restore the path back to what it was prior to running this script. 

@JimSuplizio  did this fail if you didn't UsePythonVersion?

scbedd: Because if that's the case that is unexpected to me and I'd like to resolve that.

weshaggard: While not exactly the same I tried running the versioning update scripts without setting this and it failed. https://dev.azure.com/azure-sdk/internal/_build/results?buildId=216201&view=logs&j=a8b3ef96-4a4b-559b-f023-10f869b6ec93&t=b95ceac7-6b57-5697-85af-4d84d95bb884. I suspect it is because it ran on an earlier version of Python. I will verify that buy explicitly setting the python version.

PR #6927, opened 2019-12-17
Test against all built

JimSuplizio:
Ensure that the pwsh block doesn't exit with the python script's exit code.
Change the check for SdkType equal client to be not equal data since that variable doesn't have to be set.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -289,7 +289,6 @@ jobs:
 
       - ${{ parameters.PreTestSteps }}
 
-      # JRSSTART
       - task: UsePythonVersion@0
         displayName: 'Use Python 3.6'
         inputs:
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -277,6 +289,36 @@ jobs:
 
       - ${{ parameters.PreTestSteps }}
 
+      # JRSSTART
```
weshaggard: I assume you planned to remove this JRSSTART comment.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -288,9 +330,12 @@ jobs:
           jdkArchitectureOption: 'x64'
           publishJUnitResults: false
           goals: ${{ parameters.TestGoals }}
+        # we want to run this when TestFromSource isn't true (which covers normal running when it isn't set) 
+        # OR when ShouldRunSourceTests is true
+        condition: or(ne(variables['TestFromSource'],'true'), eq(variables['ShouldRunSourceTests'],'true'))
```
weshaggard: You are going to want to and this condition with "succeeded" https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#succeeded so that we don't end up trying to run this step when some other step already failed.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
JimSuplizio: Removed with 5th commit.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
JimSuplizio: Good catch. Updated with latest commit.

PR #6928, opened 2019-12-17
Set core dep version to 1.1.0

JimSuplizio:
There were additive changes made to core which caused the dependency version of core to be updated. These changes are, right now, only needed by a single library and are causing problems with other libraries that need to release since the new version of core has not been released. This is a temporary change to unblock those releases while further discussion around these kinds of changes is reviewed.

### DIFF
--- a/eng/versioning/version_client.txt
+++ b/eng/versioning/version_client.txt
@@ -31,4 +31,4 @@ com.azure:azure-e2e;1.0.0-beta.1;1.0.0-beta.1
 
 # This is a temporary fix as appconfiguration is the only library that needs
 # the updated core dependency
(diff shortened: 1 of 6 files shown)


alzimmermsft:
Should the other Azure Core modules use a current version dependency on Azure Core given that they ship in unison? Also, I think that would resolve all the PR validation failures happening here.

JimSuplizio:
> Should the other Azure Core modules use a current version dependency on Azure Core given that they ship in unison? Also, I think that would resolve all the PR validation failures happening here.

As per our conversation, there was a second additive change to core that wasn't accounted for in the first commit that is for the second.

PR #6934, opened 2019-12-17
Added support for managed disk diff

gapra-msft:
This PR is pending a discussion about how this feature should be exposed to the customer

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/PageBlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/PageBlobAsyncClient.java
@@ -603,13 +603,14 @@ public final class PageBlobAsyncClient extends BlobAsyncClientBase {
     }
 
     /**
-     * Gets the collection of page ranges that differ between a specified snapshot and this page blob. For more
+     * This API only works for managed disk accounts.
+     * <p>Gets the collection of page ranges that differ between a specified snapshot and this page blob. For more
      * information, see the <a href="https://docs.microsoft.com/rest/api/storageservices/get-page-ranges">Azure
-     * Docs</a>.
+     * Docs</a>.</p>
      *
      * <p><strong>Code Samples</strong></p>
      *
-     * {@codesnippet com.azure.storage.blob.specialized.PageBlobAsyncClient.getPageRangesDiffFromUrl#BlobRange-String}
+     * {@codesnippet com.azure.storage.blob.specialized.PageBlobAsyncClient.getManagedDiskRangesDiff#BlobRange-String}
      *
      * @param blobRange {@link BlobRange}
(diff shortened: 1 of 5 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/PageBlobAsyncClient.java
```diff
@@ -596,23 +596,88 @@ private static String pageRangeToString(PageRange pageRange) {
...
+        String prevSnapshotUrl, BlobRequestConditions requestConditions, Context context) {
         blobRange = blobRange == null ? new BlobRange(0) : blobRange;
         requestConditions = requestConditions == null ? new BlobRequestConditions() : requestConditions;
 
-        if (prevSnapshot == null) {
-            throw logger.logExceptionAsError(new IllegalArgumentException("prevSnapshot cannot be null"));
+        URL url = null;
+        if (prevSnapshotUrl == null) {
+            if (prevSnapshot == null) {
+                throw logger.logExceptionAsError(new IllegalArgumentException("prevSnapshot cannot be null"));
```
rickle-msft: I think you mean both can't be null

rickle-msft: They also can't both be specified, right?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/PageBlobAsyncClient.java
```diff
@@ -596,23 +596,88 @@ private static String pageRangeToString(PageRange pageRange) {
...
+     *
+     * {@codesnippet com.azure.storage.blob.specialized.PageBlobAsyncClient.getManagedDiskRangesDiff#BlobRange-String}
+     *
+     * @param blobRange {@link BlobRange}
+     * @param prevSnapshotUrl Specifies the URL of a previous snapshot of the target blob. Specifies that the
+     * response will contain only pages that were changed between target blob and previous snapshot. Changed pages
+     * include both updated and cleared pages. The target blob may be a snapshot, as long as the snapshot specified by
+     * prevsnapshot is the older of the two.
+     *
+     * @return A reactive response emitting all the different page ranges.
```
rickle-msft: This sounds kind of weird. Can you say "all the page ranges that differ between the specified versions of the blob"?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/PageBlobAsyncClient.java
gapra-msft: I think this was from when we planned on adding an overload - it isnt required anymore, I'll remove it in a new PR

PR #6938, opened 2019-12-17
Rev up azure-core-test beta after release

hemanttanwar:
We had to release `1.1.0-beta.1` of azure-core-test since it is dependency for some of the libraries.
And now rev up beta version for this.
Storage depends on unreleased version of azure core test because it needs new feature for Test.LIVE. This will allow us to release azure-core-test and  after that  we will change storage back to released version of azure-core-test.

### DIFF
--- a/sdk/storage/azure-storage-blob-batch/pom.xml
+++ b/sdk/storage/azure-storage-blob-batch/pom.xml
@@ -86,7 +86,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-test</artifactId>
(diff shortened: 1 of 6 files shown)


alzimmermsft:
I don't see Azure Core Test 1.1.0-beta.1 release to Maven central. Is this because it hasn't propagated yet?

hemanttanwar:
> I don't see Azure Core Test 1.1.0-beta.1 in Maven central. Is this because it hasn't propagated yet?

http://repo2.maven.apache.org/maven2/com/azure/azure-core-test/1.1.0-beta.1/

PR #6948, opened 2019-12-18
Update Certificates library current version post preview release

samvaity:


sdk/e2e/pom.xml
```diff
@@ -48,7 +48,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-security-keyvault-certificates</artifactId>
-      <version>4.0.0-beta.7</version> <!-- {x-version-update;com.azure:azure-security-keyvault-certificates;current} -->
+      <version>4.0.0-beta.8</version> <!-- {x-version-update;com.azure:azure-security-keyvault-certificates;current} -->
```
alzimmermsft: Should this be a dependency version? Would this be something we expect a consumer of our SDKs to run to see end-to-end scenarios?

@jianghaolu @joshfree

JimSuplizio: I've asked a similar question on the PR where this was added. I wouldn't have that block this merge though.

Issue #6097, opened 2019-10-30
Add Put Blob Limit to ParallelTransferOptions


alzimmermsft:
To help reduce service calls and potentially optimize throughput offer a `putBlobLimit` property to `ParallelTransferOptions` that will be used in methods which take the parameter to determine where the data should be uploaded using Put Blob or Stage Blocks.

rickle-msft:
Relevant PR was merged as noted above.

PR #6950, opened 2019-12-18
Configure single shot upload

### LINKED ISSUE #5887: Configurable maxSingleShotSize for uploadFileToBlockBlob
This value should be added as a parameter to uploadFile, either in the method signature or in ParallelTransferOptions. It can default to the maxUploadSize of 256MB. The reason for this parameter is allowing customers to configure streaming vs transaction costs. Breaking larger files into chunks can possibily parallelize and therefore speedup transfer, but it can consume more resources and requires more requests, which may be undesirable in certain environments. Therefore, the customer should be able to configure when they want to upload an entire file all at once if the size is permitted by the service.

rickle-msft:
Resolves #5887 and #6097

### DIFF
--- a/sdk/storage/azure-storage-blob/CHANGELOG.md
+++ b/sdk/storage/azure-storage-blob/CHANGELOG.md
@@ -1,11 +1,11 @@
 # Release History
 
 ## 12.2.0-beta.2 (Unreleased)
+- Added a field to ParallelTransferOptions that allows customers to configure the maximum size to upload in a single PUT. Data sizes larger than this value will be chunked and parallelized.
 
 ## 12.2.0-beta.1 (2019-12-17)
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Release History
 
-## 12.2.0-beta.1 (Unreleased)
+## Version XX.X.X (XXXX-XX-XX)
+- Added a field to ParallelTransferOptions that allows customers to configure the maximum size to upload in a single PUT. Data sizes larger than this value will be chunked and parallelized.
+
+## Version 12.2.0-beta1 (2019-12-17)
```
gapra-msft: Add a . between beta and 1

PR #6954, opened 2019-12-18
Implement Digest Authorization Support

### LINKED ISSUE #6109: Support for Digest authenticated proxy
azure-core does not support digest authenticated proxy. This may additionally require changes in azure-identity.

Related customer-reported issue - https://github.com/Azure/azure-sdk-for-java/issues/5372

alzimmermsft:
Fixes #6109 

Digest authorization support is added to Azure Core by `AuthorizationChallengeHandler`. This class handles responding to Basic and Digest authorization requests and is compliant with RFC 2617 and RFC 7616.

### DIFF
--- a/sdk/core/azure-core-http-netty/cgmanifest.json
+++ b/sdk/core/azure-core-http-netty/cgmanifest.json
@@ -1,0 +1,15 @@
+{
+    "Registrations": [
+        {
+            "Component": {
+                "Type": "Maven",
+                "Maven": {
+                    "ArtifactId": "netty-handler-proxy",
+                    "GroupId": "io.netty",
+                    "Version": "4.1.42.Final"
+                }
+            }
+        }
+    ],
+    "Version": 1
+}


sdk/core/azure-core/src/main/java/com/azure/core/http/AuthorizationChallengeHandler.java
```diff
@@ -0,0 +1,402 @@
...
+                // RFC7616 specifies that is the "algorithm" header is null it defaults to MD5.
+                return (algorithmHeader == null) ? MD5 : algorithmHeader.toUpperCase(Locale.ROOT);
+            }));
+    }
+
+    /*
+     * Creates a client nonce.
+     */
+    private static String generateCnonce() {
+        return "0a4f113b";
```
jianghaolu: Does it have to be a new nonce for every new request?

alzimmermsft: Yeah! This was hard coded to verify that a test case worked, I'll update this with something pluggable for testing.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,287 @@
...
+        if (proxyChallengeHolder != null) {
+            // Attempt to apply digest challenges, these are preferred over basic authorization.
+            List<Map<String, String>> digestChallenges = proxyChallengeHolder.getDigestChallenges();
+            if (!CoreUtils.isNullOrEmpty(digestChallenges)) {
+                authorizationHeader = challengeHandler.handleDigest(PROXY_METHOD, PROXY_URI_PATH, digestChallenges,
+                    PROXY_ENTITY_BODY);
+            }
+
+            // If digest challenges exist or all failed attempt to use basic authorization.
+            if (CoreUtils.isNullOrEmpty(authorizationHeader) && proxyChallengeHolder.hasBasicChallenge()) {
```
srnagar: Why do we fallback on basic auth if digest auth failed?

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,287 @@
...
+
+        /*
+         * If the authentication info response contains a cnonce or nc value it MUST match the value sent in the
+         * authorization header. This is the server performing validation to the client that it received the
+         * information.
+         */
+        assert !authenticationInfoPieces.containsKey(CNONCE)
+            || authenticationInfoPieces.get(CNONCE).equals(authorizationPieces.get(CNONCE));
+        assert !authenticationInfoPieces.containsKey(NC)
+            || authenticationInfoPieces.get(NC).equals(authorizationPieces.get(NC));
```
srnagar: Prefer to throw exceptions instead of using `assert`. If these assertions fail, they throw `AssertionError` which is of type [Error](https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html) and is considered a serious non-recoverable problem with the application and is not expected to be handled upstream.

```
An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch. Most such errors are abnormal conditions. The ThreadDeath error, though a "normal" condition, is also a subclass of Error because most applications should not try to catch it.
A method is not required to declare in its throws clause any subclasses of Error that might be thrown during the execution of the method but not caught, since these errors are abnormal conditions that should never occur. That is, Error and its subclasses are regarded as unchecked exceptions for the purposes of compile-time checking of exceptions.
```

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
```diff
@@ -9,68 +9,110 @@
...
 
+    /*
+     * Configures the underlying TcpClient that sends the request.
+     */
+    private TcpClient configureTcpClient(TcpClient tcpClient) {
+        if (eventLoopGroup != null) {
+            tcpClient = tcpClient.runOn(eventLoopGroup);
+        }
+
+        ProxyHandler proxyHandler = proxyHandlerSupplier.get();
```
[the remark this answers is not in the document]
alzimmermsft: Added check for `proxyHandlerSupplier` being `null`.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ChallengeHolder.java
```diff
@@ -0,0 +1,41 @@
...
+
+package com.azure.core.http.netty.implementation;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Model class that contains the authentication challenges returned by a server.
+ */
+public final class ChallengeHolder {
```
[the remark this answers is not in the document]
alzimmermsft: Added annotation

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,287 @@
...
+        this.challengeHandler = challengeHandler;
+        this.proxyChallengeHolderReference = proxyChallengeHolderReference;
+
+        this.codec = new HttpClientCodec();
+        this.authScheme = new AtomicReference<>();
+    }
+
+    @Override
+    public String protocol() {
+        return "http";
```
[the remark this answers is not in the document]
alzimmermsft: Added constant

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    @Override
+    public String authScheme() {
+        String scheme = authScheme.get();
+
+        return (scheme == null) ? "none" : scheme;
```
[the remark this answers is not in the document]
alzimmermsft: Added constant

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,287 @@
...
+import java.util.function.Supplier;
+
+import static com.azure.core.util.AuthorizationChallengeHandler.PROXY_AUTHENTICATE;
+import static com.azure.core.util.AuthorizationChallengeHandler.PROXY_AUTHENTICATION_INFO;
+import static com.azure.core.util.AuthorizationChallengeHandler.PROXY_AUTHORIZATION;
+
+/**
+ * This class handles authorizing requests being sent through a proxy which require authentication.
+ */
+public final class ProxyAuthenticationHandler extends ProxyHandler {
```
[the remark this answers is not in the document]
alzimmermsft: Sadly `HttpProxyHandler` is `final` so it can't be extended.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,287 @@
...
+
+    @Override
+    protected void removeDecoder(ChannelHandlerContext ctx) {
+        this.codec.removeInboundHandler();
+    }
+
+    @Override
+    protected Object newInitialMessage(ChannelHandlerContext ctx) {
+        // This needs to handle no authorization proxying.
+        InetSocketAddress raddr = this.destinationAddress();
```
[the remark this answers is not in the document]
alzimmermsft: Done

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
alzimmermsft: The scenarios where basic is used are the following:

- All digest authentication challenges failed and a basic authentication challenge was returned.
- No digest authentication challenges were returned but a basic challenge was.

The overall idea is that the more secure authentication challenges are attempted first followed by less secure challenges if those fail. If you think we should ignore a basic challenge if digest fails I'd be on-board with that.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
alzimmermsft: Updated to throw an exception instead.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
srnagar: Yeah, I would err on the side of caution before falling back to basic. If users request for this behavior, we can consider relaxing this but if the user started with `digest proxy` in mind, we should fail if that's not possible.

alzimmermsft: With the current design there is no need to indicate Basic or Digest authentication, the `HttpClient` will use whatever challenge (if any) the server returns and it will prefer using Digest before Basic. The only configuration a user will determine is whether `ProxyOptions` has a username and password configured, if the configuration is there an `AuthorizationChallengeHandler` will be added into the `HttpClient` to handle Basic or Digest challenges and if there is no configuration and a challenge is received it will never be able to authorize itself. This is why there is a Basic fallback if all the Digest challenges are malformed or contain expected algorithms the JVM doesn't support.

This could be a place where the design is changed to allow configuring which type of authentication should be used and that could be where a user determines if Basic and Digest or just one of the two is supported for authorization.

srnagar: I am okay with keeping it simple and have a fallback on basic if digest fails. After all, the proxy server is the one that is going to send the challenge and it makes users life easier by not having to split hairs to decide whether to use digest or basic.

sdk/core/azure-core/src/main/java/com/azure/core/util/AuthorizationChallengeHandler.java
```diff
@@ -0,0 +1,521 @@
...
+            }
+
+            Function<byte[], byte[]> digestFunction = getDigestFunction(algorithm);
+
+            // Unable to retrieve a digest for the specified algorithm, skip it.
+            if (digestFunction == null) {
+                continue;
+            }
+
+            Map<String, String> challenge = new ConcurrentHashMap<>(challengesByType.get(algorithm).get(0));
```
srnagar: Is it guaranteed that the first entry in the list will be the latest?

sdk/core/azure-core/src/main/java/com/azure/core/util/AuthorizationChallengeHandler.java
```diff
@@ -0,0 +1,521 @@
...
+     * @param entityBodySupplier Supplies the request entity body, used to compute the hash of the body when using
+     * {@code "qop=auth-int"}.
+     * @return A preemptive authorization header for a potential Digest authentication challenge.
+     */
+    public final String attemptToPipelineAuthorization(String method, String uri, Supplier<byte[]> entityBodySupplier) {
+        String pipeliningType = authorizationPipeliningType.get();
+
+        if (DIGEST.equals(pipeliningType)) {
+            Map<String, String> challenge = lastChallenge.get();
+            String algorithm = challenge.get(ALGORITHM);
```
srnagar: I guess there might be a race condition here - `challenge.set()` is happening in `handleDigest()` and `challenge.get()` is being called here. Multiple threads calling these two methods in random order can lead to unexpected results unless there is some thread-safety built outside of this class that I am missing.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,313 @@
...
+
+    /*
+     * Proxies are always the root path.
+     */
+    private static final String PROXY_URI_PATH = "/";
+
+    /*
+     * Digest authentication to a proxy uses the 'CONNECT' method, these can't have a request body.
+     */
+    private static final Supplier<byte[]> PROXY_ENTITY_BODY = () -> new byte[0];
```
conniey: I'd consider supplying a constant so an allocation doesn't have to be made each time.
```java
private static final byte[] NO_BODY = {};
```

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,313 @@
...
+
+    private final AuthorizationChallengeHandler challengeHandler;
+    private final AtomicReference<ChallengeHolder> proxyChallengeHolderReference;
+    private final HttpClientCodec codec;
+    private final AtomicReference<String> authScheme;
+
+    private HttpResponseStatus status;
+
+    public ProxyAuthenticationHandler(InetSocketAddress proxyAddress, AuthorizationChallengeHandler challengeHandler,
+        AtomicReference<ChallengeHolder> proxyChallengeHolderReference) {
```
conniey: Why is this being passed in as an AtomicReference rather than just a ChallengeHOlder? Are we expecting this to be modified by some other object?

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,313 @@
...
+        super(proxyAddress);
+
+        this.challengeHandler = challengeHandler;
+        this.proxyChallengeHolderReference = proxyChallengeHolderReference;
+
+        this.codec = new HttpClientCodec();
+        this.authScheme = new AtomicReference<>();
```
conniey: When is authScheme updated/set? It isn't clear to me where this happens or why we need an atomic reference for it.

sdk/core/azure-core-http-okhttp/pom.xml
```diff
@@ -101,12 +101,30 @@
...
+      <artifactId>mockito-core</artifactId>
+      <version>3.0.0</version><!-- {x-version-update;org.mockito:mockito-core;external_dependency} -->
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-handler-proxy</artifactId>
+      <version>4.1.42.Final</version>
+      <scope>test</scope>
+    </dependency>
```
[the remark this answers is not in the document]
alzimmermsft: Removed, accidentally added when I was copying tests over from Netty to OkHttp

sdk/core/azure-core/src/main/java/com/azure/core/util/AuthorizationChallengeHandler.java
alzimmermsft: It is guaranteed, this code is a bit odd to cover some potential corner cases where multiple challenges using the same algorithm are returned by a proxy server though it is unlikely.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
alzimmermsft: Updated

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
alzimmermsft: Per design multiple instances of `ProxyAuthenticationHandler` will reference this object, this is being done since Netty will close the channel once the request has been completed and it will need to be passed to each channel opened by the client.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
```diff
@@ -0,0 +1,313 @@
...
+    @Override
+    public String authScheme() {
+        String scheme = authScheme.get();
+
+        return (scheme == null) ? NONE : scheme;
+    }
+
+    @Override
+    protected void addCodec(ChannelHandlerContext ctx) {
+        ctx.pipeline().addBefore(ctx.name(), null, this.codec);
```
anuchandy: just curious why we're add/remove `HttpClientCodec` instance in our proxy-hander? does reactor-netty adds one based on default or user provided `HttpResponseDecoderSpec` ?

anuchandy: never mind, I see Netty's `HttpProxyHandler` doing the same.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/ProxyAuthenticationHandler.java
alzimmermsft: Looks like I forgot to add this functionality, it is used in logging for Netty. It doesn't need to be an `AtomicReference` as it will only exist in the `ProxyAuthenticationHandler` which isn't shared across Netty pipelines, which should mean that it won't be accessed by multiple threads at once.

alzimmermsft: Added functionality to set `authScheme`

sdk/core/azure-core/src/main/java/com/azure/core/util/AuthorizationChallengeHandler.java
alzimmermsft: There isn't thread safety built-in for this, should accesses to `lastChallenge` be synchronized?

alzimmermsft: Synchronized access to `lastChallenge`

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/implementation/HttpProxyHandler.java
```diff
@@ -0,0 +1,331 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+/*
+ * Portions Copyright (c) Microsoft Corporation
+ */
```
srnagar: Would like to get @JonathanGiles opinion on this copyright section.

alzimmermsft: @anuchandy Could you also weigh in on this as you've been investigating this area recently

JonathanGiles: I believe this was the advice that @anuchandy received, so I will defer to him. It seems reasonable to me. I would expect that the THIRD_PARTY_NOTICES file is also updated.

alzimmermsft: Updated the licensing file to include the copyright above.

Issue #6935, opened 2019-12-17
[Event Hubs] Use initial offset position provided by user in initialization handler


srnagar:
Update implementation to use the initial position provided by the user in initialization handler.

srnagar:
Closed in #6978

PR #6978, opened 2019-12-19
Update error handling logic and add tests for EPC

### LINKED ISSUE #6698: [EventHubs] Update error handling implementation

srnagar:
Update the error handling logic as per design and add tests.

Fixes #6698 #6935

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionPumpManager.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionPumpManager.java
@@ -120,9 +120,9 @@ class PartitionPumpManager {
         // the position in set in the InitializationContext (either the earliest event in the partition or
         // the user provided initial position)
         if (checkpoint != null && checkpoint.getOffset() != null) {
-            startFromEventPosition = EventPosition.fromOffset(checkpoint.getOffset(), false);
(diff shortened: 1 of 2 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionPumpManager.java
```diff
@@ -151,42 +144,27 @@ void startPartitionPump(PartitionOwnership claimedOwnership, Checkpoint checkpoi
...
                         partitionEvent.getLastEnqueuedEventProperties()));
                     endProcessTracingSpan(processSpanContext, Signal.complete());
-                } catch (Exception ex) {
-                    /* event processing threw an exception */
-                    handleProcessingError(claimedOwnership, partitionProcessor, ex, partitionContext);
-                    endProcessTracingSpan(processSpanContext, Signal.error(ex));
+                } catch (Throwable throwable) {
+                    /* user code for event processing threw an exception - log and bubble up */
+                    endProcessTracingSpan(processSpanContext, Signal.error(throwable));
+                    throw logger.logExceptionAsError(new RuntimeException(throwable));
```
conniey: Should we add a message to this RuntimeException before bubbling it up? Could be useful because often the first thing  I do is "Ctrl+F" for the message in code. There is another instance.

srnagar: Added message to both exceptions.

PR #6980, opened 2019-12-19
increase and verify version functions

JimSuplizio:
Added functions to increment and verify versions. These should be used by the release pipelines.

### DIFF
--- a/eng/versioning/set_versions.py
+++ b/eng/versioning/set_versions.py
@@ -39,12 +39,10 @@ import time
 from utils import BuildType
 from utils import CodeModule
 from utils import UpdateType
-from utils import version_regex_str_with_anchor
 from utils import version_regex_str_with_names_anchored
 from utils import prerelease_version_regex_with_name
 
 # The regex string we want should be the anchored one since the entire string is what's being matched
-version_regex = re.compile(version_regex_str_with_anchor)
 version_regex_named = re.compile(version_regex_str_with_names_anchored)
 prerelease_regex_named = re.compile(prerelease_version_regex_with_name)
 


eng/versioning/utils.py
```diff
@@ -28,6 +28,10 @@
 # semver format
 # https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
 version_regex_str_with_anchor = r'^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$'
+version_regex_str_with_names_anchored = r'^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$'
```
weshaggard: probably worth just having the one regex with the named anchored instead of both.

eng/versioning/set_versions.py
```diff
@@ -93,23 +109,133 @@ def prep_version_file_for_source_testing(build_type):
...
+       raise ValueError('artifact_id ({}) was not found in version file {}'.format(artifact_id, version_file))
+
 def main():
     parser = argparse.ArgumentParser(description='set version numbers in the appropriate version text file')
     parser.add_argument('--update-type', '--ut', type=UpdateType, choices=list(UpdateType))
     parser.add_argument('--build-type', '--bt', type=BuildType, choices=list(BuildType))
     parser.add_argument('--build-qualifier', '--bq', help='build qualifier to append onto the version string.')
     parser.add_argument('--artifact-id', '--ar', help='artifactId to target.')
     parser.add_argument('--prep-source-testing', '--pst', action='store_true', help='prep the version file for source testing')
+    parser.add_argument('--increase-version', '--iv', action='store_true', help='increase the version for a given artifact')
```
weshaggard: nit: minor but the other scripts call this increment-version.

eng/versioning/set_versions.py
```diff
@@ -93,23 +109,133 @@ def prep_version_file_for_source_testing(build_type):
...
+                if module.artifact_id == artifact_id:
+                    artifact_found = True
+                    vmatch = version_regex_named.match(module.current)
+                    if (vmatch.group('prerelease') is not None):
+                        prever = prerelease_regex_named.match(vmatch.group('prerelease'))
+                        rev = int(prever.group('revision'))
+                        rev += 1
+                        new_version = '{}.{}.{}-beta.{}'.format(vmatch.group('major'), vmatch.group('minor'), vmatch.group('patch'), str(rev))
+                    else:
+                        hotfix = int(vmatch.group('patch'))
```
weshaggard: This should increment the minor version by default. https://github.com/Azure/azure-sdk/blob/master/docs/policies/releases.md#java

JimSuplizio: The next commit will have this behavior as well as the link to the doc.

eng/versioning/utils.py
JimSuplizio: Removed the non-named one. Will be updated with next commit.

eng/versioning/set_versions.py
JimSuplizio: changed to increment, will be updated with next commit

PR #6986, opened 2019-12-20
Updates Event Hubs samples

conniey:
Updates Event Hubs samples to align with scenarios represented in: https://github.com/Azure/azure-sdk-for-js/issues/6552
* Adds sample demonstrating proxy and web sockets.
* Adds sample to show how a streaming producer could be used.
* Renames samples to EventProcessorClient from EventProcessor.
* Adds a more complex sample for EventProcessorClient using demonstrating initalisation, processing, state-management and close context.

TODO:
- [x] Update READMEs to point to these samples.
  - [x] Event Hubs README
  - [x] Samples folder README

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/samples/README.md
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/samples/README.md
@@ -31,8 +31,9 @@ string value can be obtained by:
 - [Publish stream of events][sample_publish_stream_events]
 - [Consume events from an Event Hub partition][sample_consume_event]
 - [Consume events from all partitions using EventProcessorClient][sample_event_processor]
+- [Consume events from all partitions and manage state of processed events][sample_event_processor_state_management]
 - [Consume events from all partitions and manage state of events using
-  EventProcessorClient][sample_event_processor_state_management]
+  EventProcessorClient][sample_event_processor_aggregate_state_management]
 - [Consume events starting from an event sequence number][sample_consume_sequence_number]
 
 ## Troubleshooting
@@ -54,7 +55,8 @@ Guidelines](../../CONTRIBUTING.md) for more information.
 [sample_consume_event]: ./java/com/azure/messaging/eventhubs/ConsumeEvents.java
 [sample_consume_sequence_number]: ./java/com/azure/messaging/eventhubs/ConsumeEventsFromKnownSequenceNumberPosition.java
 [sample_event_processor]: ./java/com/azure/messaging/eventhubs/EventProcessorClientSample.java
-[sample_event_processor_state_management]: ./java/com/azure/messaging/eventhubs/EventProcessorClientWithStateManagement.java
+[sample_event_processor_aggregate_state_management]: ./java/com/azure/messaging/eventhubs/EventProcessorClientAggregateEventsSample.java
+[sample_event_processor_state_management]: ./java/com/azure/messaging/eventhubs/EventProcessorClientStateManagement.java
 [sample_get_event_hubs_metadata]: ./java/com/azure/messaging/eventhubs/GetEventHubMetadata.java
 [sample_publish_custom_metadata]: ./java/com/azure/messaging/eventhubs/PublishEventsWithCustomMetadata.java
 [sample_publish_identity]: ./java/com/azure/messaging/eventhubs/PublishEventsWithAzureIdentity.java


sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/EventProcessorClientWithStateManagement.java
```diff
@@ -0,0 +1,407 @@
...
+ * data can say if a machine is over heating, or if no temperature data for a machine has been collected for a while, it
+ * may be offline.
+ *
+ * The partition key for each produced event is the name of the machine. This ensures that the temperature data for that
+ * machine always gets routed to the same partition. The contents of each event is the temperature of that machine in
+ * Celsius.
+ *
+ * Every 5 seconds, {@link MachineInformation} reports the current average temperature of the machine.
+ */
+public class EventProcessorClientWithStateManagement {
```
srnagar: While this sample is great and closer to real-world scenario, I am wondering if this is too big for users to grasp what is happening. In addition to this sample, having a simpler sample that shows how to maintain state would be a good first step before introducing to this sample.

sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/PublishEventsWithWebSocketsAndProxy.java
```diff
@@ -0,0 +1,80 @@
...
+     * @param args Unused arguments to the program.
+     */
+    public static void main(String[] args) {
+        List<EventData> telemetryEvents = Arrays.asList(
+            new EventData("Roast beef".getBytes(UTF_8)),
+            new EventData("Cheese".getBytes(UTF_8)),
+            new EventData("Tofu".getBytes(UTF_8)),
+            new EventData("Turkey".getBytes(UTF_8)));
+
+        // The connection string value can be obtained by:
```
srnagar: This might be worth putting in README and the samples just provide a link to it.

sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/PublishStreamOfEvents.java
```diff
@@ -0,0 +1,234 @@
...
+                    getOrCreateBatch()
+                        .flatMap(batch -> batch.tryAdd(event) ? Mono.empty() : Mono.just(batch))
+                        .flatMap(fullBatch -> {
+                            sink.next(fullBatch);
+
+                            return createBatch().map(newBatch -> {
+                                if (!newBatch.tryAdd(event)) {
+                                    sink.error(new IllegalArgumentException(String.format(
+                                        "Event is too large for an empty batch. Max size: %s. Event: %s",
+                                        newBatch.getMaxSizeInBytes(), event.getBodyAsString())));
```
srnagar: Instead of printing the event body, it might be better to show how to get the event size. This will show the comparison of the max size allowed and how much bigger the event size is.

sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/EventProcessorClientWithStateManagement.java
conniey: One of the things brought up during our stand up was to have a "real-world" use case of our EPH rather than the contrived printouts.

conniey: I'll create one that just aggregates data from all partitions.

sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/PublishStreamOfEvents.java
conniey: We currently don't persist this information in EventData and it is just calculated. It would be nice to know. I'll create an issue to can add this.

PR #6989, opened 2019-12-20
Changes for unreleased dependencies

JimSuplizio:
The changes here are to better handle the unreleased dependency tags.
- Unreleased_ artifacts should only ever have a dependency version, not a current version. This should remove some ambiguity from the entries. 
- Added verification to the pom file verification script to ensure that pom files never try to use 'current' for an unreleased dependency. Clearly, it works since one of the changes was to fix a pom file that was doing just that.
- The changes to the util classes were necessary to support the fact that unreleased versions will not have a current attribute. 
- Ensure that if we're changing all of the versions for a devops build that we don't accidentally change any of the parent pom versions.
- The update_versions_file_for_nightly_devops function was overhauled a bit to be more correct when processing things for the devops builds. One of the biggest changes was to deal with the unreleased dependencies. When the current versions of an artifact is changed and that artifact has an unreleased dependency, that unreleased dependency version needs to be changed at well.
- The use cases and samples at the top of the set_versions.py file were rushed to get the auto increment/verification functions yesterday and were hilariously incorrect.
- Last but not least, change the error output in the pom verification script to red so it stands out more in the build logs.

### DIFF
--- a/eng/versioning/set_versions.py
+++ b/eng/versioning/set_versions.py
@@ -14,19 +14,19 @@
 # and the current versions.
 #
 #
-#    python eng\versioning\set_versions.py --ut [library|external_dependency|all] --bt [client|data|management] --bq <BuildQualifierString> --ar <artifactId>
+#    python eng/versioning/set_versions.py --ut [library|external_dependency|all] --bt [client|data|management] --bq <BuildQualifierString> --ar <artifactId>
 #
 # Use case: increment the version of a given artifact in the approprate version_[client|data|management].txt file
 #
-#    python eng\versioning\set_versions.py --bt [client|data|management] --increment-version --artifact-id <artifactId>
+#    python eng/versioning/set_versions.py --bt [client|data|management] --increment-version --artifact-id <artifactId>
 # For example: To update increment the version of azure-core
-#    python eng\versioning\set_versions.py --bt client --iv --ar azure-core
+#    python eng/versioning/set_versions.py --bt client --iv --ar azure-core
 #
 # Use case: verify the version of a given artifact in the approprate version_[client|data|management].txt file
 #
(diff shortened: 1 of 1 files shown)


eng/versioning/set_versions.py
```diff
@@ -14,19 +14,19 @@
 # and the current versions.
 #
 #
-#    python utilities/set_versions.py --ut [library|external_dependency|all] --bt [client|data|management] --bq <BuildQualifierString> --ar <artifactId>
+#    python eng\versioning\set_versions.py --ut [library|external_dependency|all] --bt [client|data|management] --bq <BuildQualifierString> --ar <artifactId>
 #
 # Use case: increment the version of a given artifact in the approprate version_[client|data|management].txt file
 #
-#    python eng/versioning/set_versions.py --bt [client|data|management] --increment-version --artifact-id <artifactId>
```
weshaggard: nit: Generally I use forward slashes for these things because they work on both windows and linux but backslashes will only work on windows.

JimSuplizio: Easy to change. Updated with the latest commit,.

sdk/core/azure-core-amqp/pom.xml
```diff
@@ -57,7 +57,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core;current} -->
+      <version>1.2.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core;dependency} -->
```
alzimmermsft: Should Azure Core AMQP just peg to current? Do we foresee situations where we would release Azure Core wihtout releasing Azure Core AMQP?

srnagar: Yes, we might release core without having to release core-amqp.

PR #6990, opened 2019-12-20
Correct documentation of `EventPosition.FromOffset`

ellismg:
In #6986 we changed the semantics of `EventPosition.FromOffset` to
return an exclusive instead of inclusive offset. However, we missed
updating the documentation for the method to match the new semantics.

This corrects that issue.


ellismg:
The wording here is a little clunky, but it matches `EventPosition.FromSequenceNumber`.  Happy to make additional changes if others have better phrasing.

Issue #6949, opened 2019-12-18
Date parsing issue in MSIToken [BUG]


lpghartz:
**Describe the bug**

Azure App Service Environment MSI endpoint is returning a JWT token with an expires_on with "AM" in it. "12/19/2019 5:03:41 AM +00:00"

parseDateToEpochSeconds() in MSIToken is parsing with a fixed pattern of "MM/dd/yyyy HH:mm:ss XXX" which will not parse with the expires_on coming from the MSI endpoint.

***Exception or Stack Trace***
Suppressed: com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `com.azure.identity.implementation.MSIToken`, problem: Unable to parse date time 12/19/2019 5:03:41 AM +00:00
 at [Source: (String)"{"access_token":"..."[truncated 1060 chars]; line: 1, column: 1427]
		at com.fasterxml.jackson.databind.exc.ValueInstantiationException.from(ValueInstantiationException.java:47)
		at com.fasterxml.jackson.databind.DeserializationContext.instantiationException(DeserializationContext.java:1732)
		at com.fasterxml.jackson.databind.deser.std.StdValueInstantiator.wrapAsJsonMappingException(StdValueInstantiator.java:491)
		at com.fasterxml.jackson.databind.deser.std.StdValueInstantiator.rewrapCtorProblem(StdValueInstantiator.java:514)
		at com.fasterxml.jackson.databind.deser.std.StdValueInstantiator.createFromObjectWith(StdValueInstantiator.java:285)
		at com.fasterxml.jackson.databind.deser.ValueInstantiator.createFromObjectWith(ValueInstantiator.java:229)
		at com.fasterxml.jackson.databind.deser.impl.PropertyBasedCreator.build(PropertyBasedCreator.java:198)
		at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeUsingPropertyBased(BeanDeserializer.java:422)
		at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.deserializeFromObjectUsingNonDefault(BeanDeserializerBase.java:1287)
		at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeFromObject(BeanDeserializer.java:326)
		at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:159)
[...]

PR #7046, opened 2019-12-26
change date format in com.azure.identity.implementation.MSIToken.java

Luyunmt:
Fix parse date error in com.azure.identity.implementation.MSIToken.parseDateToEpochSeconds method 
[#6949](https://github.com/Azure/azure-sdk-for-java/issues/6949)

### DIFF
--- a/sdk/identity/azure-identity/src/test/java/com/azure/identity/util/TestUtils.java
+++ b/sdk/identity/azure-identity/src/test/java/com/azure/identity/util/TestUtils.java
@@ -86,7 +86,7 @@ public final class TestUtils {
     }
 
     public static Mono<AccessToken> getMockAccessToken(String accessToken, String expiresOn) {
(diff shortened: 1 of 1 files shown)


JonathanGiles:
My worry with this change would be that other services might have different time format expectations, and in making this change for app service we might be introducing regressions for other services. I will assign this issue to @jianghaolu to investigate further. Thank you for the contribution!

jongio:
@Luyunmt @jianghaolu - I'm curious how the original date/time format string was chosen.  Did that come from a requirement of an existing service?  I share the same concern as Jonathan, we don't want to break other services with this change.

Luyunmt:
@jianghaolu has a commit [commit](https://github.com/Azure/azure-sdk-for-java/commit/0c76069f7865452d92297de286a009f688a96cca) in master fixing the issue

PR #7052, opened 2019-12-26
Add JUnit engine dependency to run unit tests

srnagar:
Text Analytics pom.xml was missing `junit-jupiter-engine` dependency that resulted in tests not running. This PR adds the missing dependency.


sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTest.java
```diff
@@ -29,6 +30,7 @@
     private TextAnalyticsClient client;
 
     @Test
+    @Disabled
```
samvaity: This can just say @Override I think rather than @Test or @Disabled?

Issue #6391, opened 2019-11-18
[FEATURE REQ] Support setting AccessTier when using BlockBlobClient.getBlobOutputStream()


rvervaek:
When I use the BlockBlobClient.getBlobOutputStream() method to upload data it is not possible to set the AccessTier for the blob. 

Provide an overloaded method getBlobOutputStream() that has a parameter AccessTier, that will write the data to the blob using the provided AccessTier.

rickle-msft:
@rvervaek Thank you for pointing out this oversight. We will schedule some work to add an overload that accepts accessTier, metadata, and blob http headers.

rickle-msft:
I am going to close this issue as the fix has been merged.

Issue #6159, opened 2019-11-05
Adds BlockSize Option to getBlobOutputStream in BlockBlob


alzimmermsft:
Add the ability to set the block size for BlockBlobOutputStream. Currently this is hard coded and making this configurable would more closely align to how other size agnostic methods work.

rickle-msft:
This is supported now that BlobOutputStream is based on buffered upload

Issue #6020, opened 2019-10-24
[Storage] BlockBlobClient.getBlobOutputStream() fails with large SequenceInputStream


mikeharder:
`BlockBlobClient.getBlobOutputStream()` fails when used with a large (1 GB + 1 byte) `SequenceInputStream`, but succeeds when used with a large `ByteArrayInputStream`.  It probably repros with other sizes, I just happened to be using "1 GB + 1 byte" when I hit this.

The code to write to the output stream is the same:

```java
OutputStream outputStream = blockBlobClient.getBlobOutputStream();
inputStream.transferTo(outputStream);
outputStream.close();
```

This works fine with `ByteArrayInputStream`:

```java
byte[] byteArray = new byte[(1024 * 1024 * 1024) + 1];
(new Random(0)).nextBytes(byteArray);
InputStream inputStream = new ByteArrayInputStream(byteArray);
```

However, it fails with an equivalent `SequenceInputStream`:

```java
byte[] byteArray = new byte[1024 * 1024];
(new Random(0)).nextBytes(byteArray);

Vector<ByteArrayInputStream> inputStreams = new Vector<>();
for (int i=0; i < 1024; i++) {
    inputStreams.add(new ByteArrayInputStream(byteArray));
}
inputStreams.add(new ByteArrayInputStream(byteArray, 0, 1));

InputStream inputStream = new SequenceInputStream(inputStreams.elements());
```

This `SequenceInputStream` works with `getBlobOutputStream()` in v8 but fails in v12 with this error:

```
        Suppressed: com.azure.storage.blob.models.StorageException: Status code 409, 
"﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>BlockCountExceedsLimit</Code>
 <Message>The uncommitted block count cannot exceed the maximum limit of 100,000 blocks.
RequestId:350bb21a-001e-00bb-58ea-89a084000000
Time:2019-10-23T21:37:10.2060109Z</Message></Error>"
                at com.azure.storage.blob.implementation.PostProcessor.lambda$null$0(PostProcessor.java:30)
                at reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:118)
                ... 80 more
```

mikeharder:
[...]

PR #7067, opened 2019-12-30
BlobOutputStream using BufferedUpload

### LINKED ISSUE #5275: [Blob Storage] Issues with input/output stream uploads, especially large streams.
**Query/Question**
As just figured out in issue #5221 I am able to upload also large amounts of data from files to block blobs via the `BlockBlobClient#uploadFromFile(filePath)` method.

However, I am usually in need of stream operations, with which I have issues. Example: a client uploads data to our backend via octet-stream. The Spring backend already maps this to a Java `InputStream` which I have to channel directly to the cloud storage.

I tried the following:
* `BlockBlobClient#upload(inputStream, length)`: results in a `RequestBodyTooLarge` error when file size is over the API limit of 256MB. How can I solve this? I would have expected the SDK to do the magic like with a file upload which can already be many gigabytes with one SDK upload method call.
* I tried writing to the output stream of a blob directly via `BlockBlobClient#getBlobOutputStream()`, channeling an incoming `InputStream` on the fly via `org.springframework.util.StreamUtils#copy(inputStream, outputStream)`. This operation 
** is very slow compared to file or input stream upload: with the latter I have speeds of 7Mb/s with my connection, the output stream write shows only ~ 140KiB/s, and
** results in completely wrong blob sizes on the storage: when I tried with a random 3Mb input stream, the blob size on the Azure portal is shown as just 4KiB.
** I am also not sure, even if it was faster and would produce a correct file size: would it work for large files?

Generally, I cannot buffer in memory or disk of the host machines due to resource constraints, this wouldn't scale. Also, the data can be potentially large (several gigabytes): if I would buffer this in memory I'd have 'out of memory' exceptions for one client only quickly, and there may be a large number of parallel uploads by thousands of different clients. The same applies for disk space when using temp files to work around this, so it's not an option.

**Code snippets**
```
public void test()
{
    final Path file = Paths.get("~/Downloads/random-file.zip");
    try(final RandomAccessFile raf = new RandomAccessFile("~/Downloads/random-file.zip", "rw"); final InputStream is = Files.newInputStream(file))
    try(final InputStream is = Files.newInputStream(path))
    {
        final long largeFileSize = 1024 * 1024 * 1024 * 10L // 10Gb of size
        final long smallFileSize = 1024 * 1024 * 3L // 3Mb of size
        raf.setLength(largeFileSize);
        //raf.setLength(smallFileSize); // -> use for trying out writing to output stream
        cloudStorageClient.upload("random-file.zip", is, largeFileSize); // -> fails, I would expect to be able to use it with any content length
        
        // writeToOutputStream(is); // -> this is very slow and produces false blob sizes.
    }
    catch (final IOException | CloudStorageException e)
    {
        LOG.error("Test failed", e);
    }
    finally
    {
        try
        {
            Files.deleteIfExists(file);
        }
        catch (final IOException e)
        {
            LOG.error("Test clean-up failed");
        }
    }
}

private void writeToOutputStream(final InputStream inputStream)
{
    try(final OutputStream outputStream = containerClient.getBlockBlobClient("random-file.zip").getBlobOutputStream())
    {
        LOG.debug("Channeling content from input stream to blob storage...");
        org.springframework.util.StreamUtils.copy(inputStream, outputStream);
    }
    catch (final IOException | StorageException e)
    {
        LOG.error("Test failed", e);
    }
}
```

***Why is this not a Bug or a feature Request?***
The input stream upload seems to adhere to the API limit of a 256Mb request body, so it's technically not a bug. But I am not sure if the SDK shouldn't handle this properly in its implementation, which is why I am not filing a feature request for now.

The issue with writing to the output stream *may be a bug*, but I am not sure yet.

**Setup (please complete the following information if applicable):**
 - OS: Mac OS 10.14 (Mojave)
 - IDE : IntelliJ
 - 12.0.0-preview.2

gapra-msft:
Fixes #5275, #6159, #6391, #6020

### DIFF
--- a/sdk/storage/azure-storage-blob/CHANGELOG.md
+++ b/sdk/storage/azure-storage-blob/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 12.2.0-beta.2 (Unreleased)
 - Added a field to ParallelTransferOptions that allows customers to configure the maximum size to upload in a single PUT. Data sizes larger than this value will be chunked and parallelized.
 - Added overloads to downloadToFile to add the option to overwrite existing files. Default behavior is to not overwrite.
-- Changed implementation of BlockBlobOutputStream to take advantage of the buffered upload method, improving performance. 
+- Improved performance of BlockBlobOutputStream.
 - Added overloads to BlockBlobClient.getBlobOutputStream to allow users to provide parallel transfer options, http headers, metadata, access tier, and request conditions.
 
 


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
```diff
@@ -175,21 +178,34 @@ private String getCurrentBlockId() {
...
             Flux<ByteBuffer> fbb = Flux.range(0, 1)
                 .concatMap(pos -> Mono.fromCallable(() -> ByteBuffer.wrap(data, (int) offset, writeLength)));
 
-            return this.writeBlock(fbb.subscribeOn(Schedulers.elastic()), blockID, writeLength);
+            ParallelTransferOptions parallelTransferOptions = ModelHelper.populateAndApplyDefaults(null);
+
+            return this.writeBlock(fbb.subscribeOn(Schedulers.elastic()), parallelTransferOptions);
+        }
+
+        private Mono<Void> writeBlock(Flux<ByteBuffer> data, ParallelTransferOptions parallelTransferOptions) {
```
gapra-msft: ParallelTransferOptions can be instantiated via the getOutputStream method too

gapra-msft: + All the others like metadata, request conditions, etc

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
```diff
@@ -134,62 +139,67 @@ void commit() {
...
-        private Mono<Void> writeBlock(Flux<ByteBuffer> blockData, String blockId, long writeLength) {
-            return client.stageBlockWithResponse(blockId, blockData, writeLength, null,
-                this.requestConditions.getLeaseId())
-                .then()
-                .onErrorResume(BlobStorageException.class, e -> {
-                    this.lastError = new IOException(e);
-                    return Mono.empty();
-                });
+            Flux<ByteBuffer> fbb = Flux.create((FluxSink<ByteBuffer> sink) -> this.sink = sink);
+            fbb.subscribe(); // Subscribe by upload takes too long
```
rickle-msft: Can you expound on this comment--We need to subscribe so that the sink is actually created and because this subscriber doesn't do anything and no data has started flowing, there are no drawbacks to this extra subscribe.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobClient.java
```diff
@@ -115,7 +116,29 @@ public BlobOutputStream getBlobOutputStream(boolean overwrite) {
...
+     * @param parallelTransferOptions {@link ParallelTransferOptions} used to configure buffered uploading.
+     * @param headers {@link BlobHttpHeaders}
+     * @param metadata Metadata to associate with the blob.
+     * @param tier {@link AccessTier} for the destination blob.
+     * @param requestConditions {@link BlobRequestConditions}
+     *
+     * @return A {@link BlobOutputStream} object used to write data to the blob.
+     * @throws BlobStorageException If a storage service error occurred.
+     */
+    public BlobOutputStream getBlobOutputStream(ParallelTransferOptions parallelTransferOptions,
```
rickle-msft: Can you add a changelog entry for this?

PR #7068, opened 2019-12-30
Added logic of checking both slf4j impl and env for logging level

sima-zhu:
The PR is to have more checking on the log level. 
We agreed to check 
* slf4j binding as our log level.
* If slf4j configuration is not set, we will read from env. 
* If both are not present, we will set log level to NOT_SET.
I have implemented src code only for team to review.

TODO: Will add more unit tests for code coverage.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
@@ -51,9 +51,12 @@ public class ClientLogger {
      * Retrieves a logger for the passed class name using the {@link LoggerFactory}.
      *
      * @param className Class name creating the logger.
+     * @throws RuntimeException it is an error.
      */
     public ClientLogger(String className) {
         logger = LoggerFactory.getLogger(className);
+        System.setProperty("java.util.logging.config.file",
+            ClientLogger.class.getClassLoader().getResource("logging.properties").getPath());
         defaultLogger = java.util.logging.Logger.getLogger(className);
     }
 
--- a/sdk/core/azure-core/src/main/resources/logging.properties
+++ b/sdk/core/azure-core/src/main/resources/logging.properties
@@ -1,1 +1,3 @@
-.level = FINE
+handlers= java.util.logging.FileHandler, java.util.logging.ConsoleHandler
+.level = ALL
+java.util.logging.ConsoleHandler.level = FINER
--- a/sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
@@ -78,7 +78,7 @@ public class ClientLoggerTests {
         logMessage(clientLogger, logLevel, logMessage);
 
         sh.flush();
-        String logValues = logCaptureStream.toString(StandardCharsets.UTF_8);
+        String logValues = new String(logCaptureStream.toByteArray(), StandardCharsets.UTF_8);
         assertTrue(logValues.contains(logMessage));
     }
 


sdk/core/azure-core/pom.xml
```diff
@@ -69,7 +69,11 @@
       <artifactId>slf4j-api</artifactId>
       <version>1.7.28</version> <!-- {x-version-update;org.slf4j:slf4j-api;external_dependency} -->
     </dependency>
-
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>1.7.28</version> <!-- {x-version-update;org.slf4j:slf4j-api;external_dependency} -->
+    </dependency>
```
alzimmermsft: @srnagar @JonathanGiles We don't want to force an SLF4J implementation by default do we?

sdk/core/azure-core/pom.xml
```diff
@@ -69,7 +69,11 @@
       <artifactId>slf4j-api</artifactId>
       <version>1.7.28</version> <!-- {x-version-update;org.slf4j:slf4j-api;external_dependency} -->
     </dependency>
-
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>1.7.28</version> <!-- {x-version-update;org.slf4j:slf4j-api;external_dependency} -->
```
alzimmermsft: Wrong version tracking tag

```suggestion
      <version>1.7.28</version> <!-- {x-version-update;org.slf4j:slf4j-simple;external_dependency} -->
```

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/Tracer.java
```diff
@@ -219,7 +219,7 @@
      * @return The updated {@link Context} object containing the span builder.
      * @throws NullPointerException if {@code context} or {@code spanName} is {@code null}.
      */
-    default Context getSharedSpanBuilder(String spanName, Context context) {
+    default Context getSharedSpanBuilder(String spanName, Context context){
```
alzimmermsft: Should revert this, this is also failing checkstyles

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -37,6 +38,7 @@
  */
 public class ClientLogger {
     private final Logger logger;
+    private LogLevel rootLogLevel;
```
alzimmermsft: This introduces potential multi-threading concerns. It'd be better to leave this as a local variable when it is used.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -256,6 +259,33 @@ private boolean canLogAtLevel(LogLevel logLevel, LogLevel environmentLoggingLeve
...
 
+    /**
+     * Checking the system log with the preference order of slf4j and environment variable.
+     *
+     * @return The log level.
+     */
+    private void loadSystemLogLevel() {
+        if (rootLogLevel != null) {
+            return;
+        }
```
alzimmermsft: This breaks a design guideline where the environment logging level must be able to update at any time, once `rootLogLevel` is set in the `ClientLogger` it will never be able to be changed, I think this should be removed.

https://azure.github.io/azure-sdk/general_implementation.html#general-config-behaviour-changes

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -256,6 +259,33 @@ private boolean canLogAtLevel(LogLevel logLevel, LogLevel environmentLoggingLeve
...
+                rootLogLevel = LogLevel.INFORMATIONAL;
+            }
+            if (logger.isWarnEnabled()) {
+                rootLogLevel = LogLevel.WARNING;
+            }
+            if (logger.isErrorEnabled()) {
+                rootLogLevel = LogLevel.ERROR;
+            }
+        }
+        rootLogLevel = getEnvironmentLoggingLevel();
```
alzimmermsft: This needs to be in an `else` block, this code is always called.

sdk/core/azure-core/pom.xml
srnagar:
```suggestion
      <version>1.7.28</version> <!-- {x-version-update;org.slf4j:slf4j-simple;external_dependency} -->
```

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -256,6 +259,33 @@ private boolean canLogAtLevel(LogLevel logLevel, LogLevel environmentLoggingLeve
...
+            if (logger.isInfoEnabled()) {
+                rootLogLevel = LogLevel.INFORMATIONAL;
+            }
+            if (logger.isWarnEnabled()) {
+                rootLogLevel = LogLevel.WARNING;
+            }
+            if (logger.isErrorEnabled()) {
+                rootLogLevel = LogLevel.ERROR;
+            }
+        }
```
alzimmermsft: Would this make more sense to go into `canLogAtLevel`? Have that check if there is a binder and if it contains one use the enabled logging levels on it to determine if it can log, if not set check against the environment variables?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
srnagar: Wouldn't this overwrite the log level from `binder`?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -256,6 +259,33 @@ private boolean canLogAtLevel(LogLevel logLevel, LogLevel environmentLoggingLeve
...
+            }
+            if (logger.isInfoEnabled()) {
+                rootLogLevel = LogLevel.INFORMATIONAL;
+            }
+            if (logger.isWarnEnabled()) {
+                rootLogLevel = LogLevel.WARNING;
+            }
+            if (logger.isErrorEnabled()) {
+                rootLogLevel = LogLevel.ERROR;
+            }
```
alzimmermsft: Should this be an `if/else if` or `switch` block? Right now this will continue updating the `rootLogLevel` for each conditional that passes and will result with whichever passes last as the `rootLogLevel`.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -232,12 +230,17 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
      * Determines if the environment and logger support logging at the given log level.
      *
      * @param logLevel Logging level for the log message.
-     * @param environmentLoggingLevel Logging level the environment is set to support.
      * @return Flag indicating if the environment and logger are configured to support logging at the given log level.
      */
-    private boolean canLogAtLevel(LogLevel logLevel, LogLevel environmentLoggingLevel) {
+    public boolean canLogAtLevel(LogLevel logLevel) {
+        loadSystemLogLevel();
```
srnagar: Is this `load` operation necessary each time this method is called?

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/Tracer.java
srnagar: This change must be reverted. There should be a space after `)`.

sdk/core/azure-core/pom.xml
JonathanGiles: We definitely don't want this. As we discussed previously, we can not include an slf4j implementation.

srnagar: Yeah, we don't want any implementation of SLF4J.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: It is a lazy initialization. It will check rootLogLevel exist or not in loadSystemLogLevel.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Good catch. Will return above

sdk/core/azure-core/pom.xml
sima-zhu: Removed.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Make the call every time instead of storing them.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: This call is to return the logLevel from app or sys. Then both canLogAtLevel and performLogging can use the loglevel.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Done with loading sys env everytime.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -232,28 +231,39 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
      * Determines if the environment and logger support logging at the given log level.
      *
      * @param logLevel Logging level for the log message.
-     * @param environmentLoggingLevel Logging level the environment is set to support.
      * @return Flag indicating if the environment and logger are configured to support logging at the given log level.
      */
-    private boolean canLogAtLevel(LogLevel logLevel, LogLevel environmentLoggingLevel) {
+    public boolean canLogAtLevel(LogLevel logLevel) {
```
alzimmermsft: `LogLevel` is in an implementation package and will need to be moved to public package.

sima-zhu: The PR is for reviewing the code flow, not finished yet.
Will merge the changes in PR:  https://github.com/Azure/azure-sdk-for-java/pull/6886.
Also, will fix the tests after merging.

alzimmermsft:
Need to update some tests or change their setup based on this change. If you want to add more tests around handling when SLF4J has a binder and which levels can be logged at you can look at the [ClientLoggerTests](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java#L38) which setup the logger with a given level of support and here is the conversion of environment log level name to level https://github.com/qos-ch/slf4j/blob/master/slf4j-simple/src/main/java/org/slf4j/simple/SimpleLoggerConfiguration.java#L145.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -232,28 +231,39 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
-                return logger.isErrorEnabled();
-            default:
-                return false;
+    /**
+     * Checking the system log with the preference order of slf4j and environment variable.
+     *
+     * @return The log level.
+     */
+    private LogLevel loadSystemLogLevel() {
+        if (logger.isDebugEnabled()) {
```
alzimmermsft: Should the binder still be checked? That way if an SLF4J binding exists but all logging levels are disabled we won't default into the environment configuration and instead wouldn't log anything as expected.

sima-zhu: Based on my testing, if there is some other logging binding to SLF4J, `logger.isDebugEnabled()` will reflect the log level set by user. So if log4j set log level to info, then the clientLogger `logger.canLogAtLevel(LogLevel.INFO)` returns true. 

Also, Binder check API is inside of `slf4j-simple` pkg which asks to introduce slf4j impl library to src code. This is not ideal as you comment above.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -50,9 +51,13 @@ public ClientLogger(Class<?> clazz) {
      * Retrieves a logger for the passed class name using the {@link LoggerFactory}.
      *
      * @param className Class name creating the logger.
+     * @throws RuntimeException it is an error.
      */
     public ClientLogger(String className) {
         logger = LoggerFactory.getLogger(className);
+        System.setProperty("java.util.logging.config.file",
+            ClientLogger.class.getClassLoader().getResource("logging.properties").getPath());
+        defaultLogger = java.util.logging.Logger.getLogger(className);
```
JonathanGiles: Don't overwrite a system property. Use another means to configure the logger. Also, don't create a default logger unless necessary - you're paying the price here to create two loggers.

sdk/textanalytics/azure-ai-textanalytics/src/samples/java/com/azure/ai/textanalytics/HelloWorld.java
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
mssfang: We have renamed this file. Make sure to exclude this file and HelloWorldAsync.java before merge to master.

alzimmermsft: Should this PR modify this file at all?

mssfang: Noop

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: We can initial it into log method. Try to figure out if we can set the logLevel during runtime.

sima-zhu:
Closed the PR as we agreed to use Sys.out for now.

PR #7075, opened 2020-01-02
KV-Keys code update

g2vinay:
Changes the code to use key id passed from user which is validated, instead of relying on the key to fetch id from, as key gets fetched on demand and is not always available.

### DIFF
--- a/sdk/keyvault/azure-security-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/CryptographyAsyncClient.java
+++ b/sdk/keyvault/azure-security-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/CryptographyAsyncClient.java
@@ -62,7 +62,7 @@ public class CryptographyAsyncClient {
     private LocalKeyCryptographyClient localKeyCryptographyClient;
     private final ClientLogger logger = new ClientLogger(CryptographyAsyncClient.class);
     private String keyCollection;
-    private String keyId;
+    private final String keyId;
 
     /**
      * Creates a CryptographyAsyncClient that uses {@code pipeline} to service requests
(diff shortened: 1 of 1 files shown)


sdk/keyvault/azure-security-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/KeyEncryptionKeyAsyncClient.java
```diff
@@ -27,18 +28,17 @@
      */
     KeyEncryptionKeyAsyncClient(String keyId, HttpPipeline pipeline, CryptographyServiceVersion version) {
         super(keyId, pipeline, version);
+        this.keyId = keyId;
```
srnagar: Since you are passing this to the superclass constructor, why not have a superclass manage the lifecycle of this variable? `CryptographyAsyncClient` can have a protected method `getKeyId()` to get this key id.

g2vinay: This needs to be Public API only in KEK client.
So moved the variable management to super class as package private method.
Kept it public in KEK client.

sdk/keyvault/azure-security-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/CryptographyAsyncClient.java
```diff
@@ -102,6 +104,7 @@
      */
     CryptographyAsyncClient(String keyId, HttpPipeline pipeline, CryptographyServiceVersion version) {
         unpackAndValidateId(keyId);
+        this.keyId = keyId;
```
srnagar: Why doesn't this constructor `initializeCryptoClients()` while the other constructor does (line 95)?

g2vinay: In other constructor, JWK key is passed into the constructor. Key is available to initialize crypto clients.
While in this constructor only key id is available at constructor time. In this scenario the Key gets fetched from the service and crypto clients are initialized on demand when user invokes a crypto operation on the client.

PR #7088, opened 2020-01-03
Net App: generate 2019 10 01

ChenTanyi:


weidongxu-microsoft:
Update pom.mgmt.xml in ./sdk/netapp as well.

PR #7090, opened 2020-01-03
Recreating connections on transient failures

### LINKED ISSUE #4352: [Event Hubs] Reopen connection on transient errors.
* On an AMQP connection failure, create and re-establish existing AMQP sessions and AMQP links.

Related #4372

conniey:
Reopens connection on transient errors.

Fixes #4352


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/EventHubConnectionProcessor.java
```diff
@@ -0,0 +1,268 @@
+/*
+ * Copyright (c) 2011-2017 Pivotal Software Inc, All Rights Reserved.
```
srnagar: This is not the copyright text we use. Is `EventHubConnectionProcessor` code provided by Pivotal Software and they require us to have this copyright text instead?

conniey: Oh no. I have no idea where I had that from.

PR #7097, opened 2020-01-03
Add az namespace span attribute for App config

samvaity:
Ref #7049

In this PR, instrumenting only AppConfig client library for the az namespace span atttribute update.

appConfig

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
@@ -28,6 +28,7 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import java.time.OffsetDateTime;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -58,8 +59,9 @@ public final class ConfigurationAsyncClient {
     private final ClientLogger logger = new ClientLogger(ConfigurationAsyncClient.class);
 
     private static final String ETAG_ANY = "*";
-    private static final Supplier<Map<String, String>> APP_CONFIG_TRACING_PROPERTIES = () ->
-        new HashMap<String, String>() {{ put(AZ_TRACING_NAMESPACE_KEY, "Microsoft.AppConfiguration"); }};
+    private static final Supplier<Map<String, String>> APP_CONFIG_TRACING_PROPERTIES =
+        () -> Collections.unmodifiableMap(
+                new HashMap<String, String>() {{ put(AZ_TRACING_NAMESPACE_KEY, "Microsoft.AppConfiguration"); }});
 
     private final String serviceEndpoint;
     private final ConfigurationService service;


sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/implementation/HttpTraceUtil.java
```diff
@@ -121,5 +125,29 @@ public static Status parseResponseStatus(int statusCode, Throwable error) {
...
+        }
+        if (userSpanName.contains("AppConfig")) {
+            return "Microsoft.AppConfiguration";
+        }
+        if (userSpanName.contains("Storage")) {
+            return "Microsoft.Storage";
+        }
+        if (userSpanName.contains("TextAnalytics")) {
+            return "Microsoft.CognitiveServices";
+        }
```
alzimmermsft: How reliably can we determine whether the `SwaggerInterfaceParser`'s `serviceName` will contain the expected names here? Instead within `RestProxy` could we retrieve the name of the package where the method exists and pass through an additional context value which will be a bit more robust compared to either a codegen name/hand-written name?

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/implementation/HttpTraceUtil.java
```diff
@@ -121,5 +125,28 @@ public static Status parseResponseStatus(int statusCode, Throwable error) {
...
+
+    /**
+     * Parse the resource provider name from the user span name.
+     *
+     * @param userSpanName the span name populated in the RestProxy layer.
+     *
+     * @return the resource provider group name.
+     */
+    public static String parseNamespaceProvider(String userSpanName) {
+        if (userSpanName.contains("KeyVault")) {
```
srnagar: Does this have to be hardcoded? Can this be made configurable instead?

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/implementation/HttpTraceUtil.java
```diff
@@ -25,21 +26,24 @@
...
 
-    private HttpTraceUtil() { }
+    private final ClientLogger logger = new ClientLogger(HttpTraceUtil.class);
+
+    private HttpTraceUtil() {
+    }
 
     /**
-     * Parse OpenTelemetry Status from HTTP response status code.
+     * Parse OpenCensus Status from HTTP response status code.
```
alzimmermsft: This is the OpenTelemetry package, should revert the comment renames.

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/Tracer.java
```diff
@@ -57,6 +57,11 @@
      */
     String SPAN_BUILDER_KEY = "builder";
 
+    /**
+     * Key for {@link Context} which indicates that the context contains the Azure resource provider namespace.
+     */
+    String AZ_NAMESPACE_KEY = "az.namespace";
```
srnagar: Can we include `tracing` in the name of this context key? This will make it clear as to what this key is used for. AZ_NAMESPACE is too generic.

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -48,7 +48,10 @@ public Context start(String spanName, Context context) {
 
         Builder spanBuilder = getSpanBuilder(spanName, context);
         Span span = spanBuilder.startSpan();
-
+        if (span.isRecording()) {
+            span.setAttribute(AZ_NAMESPACE_KEY,
+                AttributeValue.stringAttributeValue(HttpTraceUtil.parseNamespaceProvider(spanName)));
```
alzimmermsft: Should this add the attribute if the parsed name is `""`? It appear in the policy it doesn't.

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/Tracer.java
samvaity: but wouldn't coming from `Tracer` class help to understand that it is coming from tracing concepts.
I think it would repetitive to mention it in the variable name and class name too.

srnagar: No, context is a property bag that is passed around quite a lot in all layers of our library. Having a key named `az.namespace` in context will not necessarily be clear as to what this namespace is and why it's used.

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/implementation/HttpTraceUtil.java
lmolkova: I agree it does not seem like a reliable/easy to maintain way to determine resource provider namespace. Could each library set some common global variable? It should also be possible to fall back to default value (null ?) and don't set az.namespace attribute in this case eliminating libraries that don't support tracing

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/Tracer.java
JonathanGiles: I think there is a slight mismatch in communication here. Srikanta is talking about the string "az.namespace", not the variable name "AZ_NAMESPACE_KEY" (despite the fact he used "AZ_NAMESPACE" in his comment). His argument is valid - imagine have a Map or a Context full of keys, one of which is "az.namespace" - it isn't clear where that came from or how it is relevant. If it had a more specific value then users might better infer that it is tracing related.

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/implementation/HttpTraceUtil.java
JonathanGiles: I agree, this is fraught with danger. Let's find a better solution.

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java
samvaity: I think that is what we are following for other span attributes as well.

samvaity: updated it to not have the span attribute if it does not exist.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/RestProxy.java
```diff
@@ -81,6 +85,27 @@ private RestProxy(HttpPipeline httpPipeline, SerializerAdapter serializer, Swagg
         this.interfaceParser = interfaceParser;
         this.decoder = new HttpResponseDecoder(this.serializer);
         this.responseConstructorsCache = new ResponseConstructorsCache();
+        this.tracerSpanAttributes = null;
```
srnagar: call the other constructor from here instead of duplicating initialization code in both constructors.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/RestProxy.java
```diff
@@ -685,4 +729,25 @@ private static HttpPipeline createDefaultPipeline(HttpPipelinePolicy credentials
...
+     * @param serializer the serializer that will be used to convert POJOs to and from request and
+     *     response bodies
+     * @param <A> the type of the Swagger interface.
+     * @param tracerSpanAttributes the attributes to be set on the tracer spans.
+     * @return a proxy implementation of the provided Swagger interface
+     */
+    @SuppressWarnings("unchecked")
+    public static <A> A create(Class<A> swaggerInterface, HttpPipeline httpPipeline, SerializerAdapter serializer,
+        TracerSpanAttributes tracerSpanAttributes) {
+        final SwaggerInterfaceParser interfaceParser = new SwaggerInterfaceParser(swaggerInterface, serializer);
```
srnagar: same here - call this create method from the other `create` overload method

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/implementation/HttpTraceUtil.java
samvaity: Updated the approach to allow the client libraries to send in their respective `Resource Provider Names` to pass through the context object and skip the hand-written comparisons. Also, updated it to skip adding the attribute if az.namespace value isn't set.

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerSpanAttributes.java
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.util.tracing;
+
+/**
+ * Class to hold the attributes for tracer spans.
+ */
+public class TracerSpanAttributes {
+    private final String resourceProviderName;
```
srnagar: Since this has just one attribute now, I am assuming new attributes will be added later. If that's the intent, then shouldn't we make this a fluent API? A public constructor with just one attribute now will result in breaking changes if we need more attributes here later. 

Alternatively, do we need this attributes class, if all fields in here are going to be `String` types? Can we instead use `Map<String, String>` as properties bag?

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProperties.java
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.util.tracing;
+
+/**
+ * A generic interface for setting tracing properties on outgoing HTTP requests.
+ */
+public interface TracerProperties {
```
srnagar: Can this be marked as a `@FunctionalInterface`? If yes, can we instead use Java's built-in Supplier?

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/Tracer.java
samvaity: Updated both the key and value to include tracing, i.e `AZ_TRACING_NAMESPACE_KEY` and `az.tracing.namespace`. The span attribute will still have the key `az.namespace`

eng/versioning/version_client.txt
```diff
@@ -36,3 +36,4 @@ com.azure:azure-storage-queue;12.2.0-beta.1;12.2.0
 # Format;
 # unreleased_<groupId>:<artifactId>;dependency-version
 unreleased_com.azure:azure-core-amqp;1.0.1
+unreleased_com.azure:azure-core;1.3.0-beta.1
```
srnagar: I am not sure why we are using unreleased for this. There's also this on L8 - `com.azure:azure-core;1.2.0;1.3.0-beta.1`

samvaity: Because I think with recent updates we wanted all libraries to use the released version of core and not the dependency version. And if we wanted to use the current version we will have to include it this way.

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProperties.java
samvaity: updated it to be  a supplier.

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerSpanAttributes.java
samvaity: Span attributes coming from clients would be expected to be strings and hence updating it to use a `Map<String, String>` instead.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationAsyncClient.java
```diff
@@ -54,6 +58,8 @@
     private final ClientLogger logger = new ClientLogger(ConfigurationAsyncClient.class);
 
     private static final String ETAG_ANY = "*";
+    private static final Supplier<Map<String, String>> APP_CONFIG_TRACING_PROPERTIES = () ->
+        new HashMap<String, String>() {{ put(AZ_TRACING_NAMESPACE_KEY, "Microsoft.AppConfiguration"); }};
```
srnagar: This map should be immutable.

PR #7100, opened 2020-01-03
Update Event Hubs and azure-core-amqp current versions and CHANGELOG to GA

conniey:
We are finally GA'ing Event Hubs and azure-core-amqp.
* Updates the current version to 1.0.1 and 5.0.1.
  * Not using 1.0.0 and 5.0.0 because of the version sorting problem in maven.
* Removes the version_client_java_files.txt since the client constants file no longer exists.
* Updates CHANGELOG.

### DIFF
--- a/sdk/core/azure-core-amqp/CHANGELOG.md
+++ b/sdk/core/azure-core-amqp/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 1.0.0 (2020-01-07)
+## 1.0.1 (2020-01-07)
 - Client library name and version are no longer hard coded in connection properties.
 - Update qpid-proton-j-extensions dependency
 - Shorten tracing span names
--- a/sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/CHANGELOG.md
+++ b/sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 1.0.0-beta.5 (Unreleased)
+## 1.0.1 (2020-01-07)
 
 ## 1.0.0-beta.4 (2019-12-02)
 - Artifact name changed from `preview` to `beta`.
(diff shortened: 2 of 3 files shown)


eng/versioning/version_client.txt
```diff
@@ -6,7 +6,7 @@ com.azure:azure-sdk-parent;1.6.0;1.6.0
 com.azure:azure-client-sdk-parent;1.7.0;1.7.0
 com.azure:azure-ai-textanalytics;1.0.0-beta.1;1.0.0-beta.1
 com.azure:azure-core;1.1.0;1.2.0
-com.azure:azure-core-amqp;1.0.0-beta.9;1.0.0-beta.9
+com.azure:azure-core-amqp;1.0.0-beta.9;1.0.1
```
JimSuplizio: There is no beta.9 of azure-core-amqp https://search.maven.org/artifact/com.azure/azure-core-amqp only a beta.8, I'm not sure why both the dependency and current versions were set to the same version here but if you try and set 1.0.0-beta.9 as the dependency version then things should fail as there isn't one in Maven,.

eng/versioning/update_versions.py
```diff
@@ -148,26 +148,6 @@ def update_versions_all(update_type, build_type, target_file, skip_readme):
                 if (file_name.endswith('.md') and not skip_readme) or (file_name.startswith('pom.') and file_name.endswith('.xml')):
                     update_versions(version_map, file_path)
 
-    # This is a temporary stop gap to deal with versions hard coded in java files.
```
JimSuplizio: We can't remove this yet, there are still data files with hard coded versions. eng\versioning\version_data_java_files.txt. My recommendation is to add a file exists check on the file and if the file doesn't exist then skip the replacement.

eng/versioning/version_client.txt
hemanttanwar: Agree This line should look like com.azure:azure-core-amqp;1.0.0-beta.8;1.0.1

sdk/core/azure-core-amqp/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Release History
 
-## Version 1.0.0-beta.9 (2020-01-07)
-- Add links on batch send operation in eventhubs
+## 1.0.0 (2020-01-07)
```
hemanttanwar: Shouldn't this be 1.0.1 ?  Because that is what  you have specified in  version_client.txt

conniey: This is fixed now.

PR #7120, opened 2020-01-04
[After MSAL 1.3] Pluggable HTTP in identity/msal

jianghaolu:


JonathanGiles:
Let's hold this back until MSAL 1.3.0 is released with it's significantly reduced dependency set.

eng/versioning/external_dependencies.txt
```diff
@@ -62,7 +62,7 @@ com.microsoft.azure:azure-mgmt-keyvault;1.11.1
 com.microsoft.azure:azure-mgmt-resources;1.3.0
 com.microsoft.azure:azure-mgmt-storage;1.3.0
 com.microsoft.azure:azure-storage;8.0.0
-com.microsoft.azure:msal4j;0.5.0-preview
+com.microsoft.azure:msal4j;1.1.0
```
JimSuplizio: I like that this is moving past a preview version.

jianghaolu:
Bump version to 1.3.0 on Jan 10 before merging

sdk/identity/azure-identity/pom.xml
```diff
@@ -81,5 +81,11 @@
       <version>3.3.0.RELEASE</version> <!-- {x-version-update;io.projectreactor:reactor-test;external_dependency} -->
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-http-netty</artifactId>
+      <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
```
g2vinay: 1.2.0 is out for this library.
Is that what we intend to add here ?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -137,28 +139,23 @@
...
         String authorityUrl = options.getAuthorityHost().replaceAll("/+$", "") + "/" + tenantId;
-        try {
+        return Mono.fromCallable(() -> {
             ConfidentialClientApplication.Builder applicationBuilder =
-                ConfidentialClientApplication.builder(clientId,
-                    ClientCredentialFactory.create(new FileInputStream(pfxCertificatePath), pfxCertificatePassword))
-                    .authority(authorityUrl);
+                    ConfidentialClientApplication.builder(clientId, ClientCredentialFactory.createFromCertificate(
+                                new FileInputStream(pfxCertificatePath), pfxCertificatePassword))
+                            .httpClient(new HttpPipelineAdapter(options.getHttpPipeline()))
```
g2vinay: so, here we don't check if (options.getHttpPipeline() != null)  before setting the pipeline ?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -137,28 +139,23 @@
...
-        } catch (CertificateException
-            | UnrecoverableKeyException
-            | NoSuchAlgorithmException
-            | KeyStoreException
-            | NoSuchProviderException
-            | IOException e) {
-            return Mono.error(e);
-        }
+            if (options.getHttpPipeline() != null) {
+                applicationBuilder.httpClient(new HttpPipelineAdapter(options.getHttpPipeline()));
```
g2vinay: we're already setting the pipeline on top, even before this check.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -173,12 +170,17 @@
         try {
             byte[] pemCertificateBytes = Files.readAllBytes(Paths.get(pemCertificatePath));
             ConfidentialClientApplication.Builder applicationBuilder =
-                ConfidentialClientApplication.builder(clientId,
-                    ClientCredentialFactory.create(CertificateUtil.privateKeyFromPem(pemCertificateBytes),
-                        CertificateUtil.publicKeyFromPem(pemCertificateBytes))).authority(authorityUrl);
+                ConfidentialClientApplication.builder(clientId, ClientCredentialFactory.createFromCertificate(
+                            CertificateUtil.privateKeyFromPem(pemCertificateBytes),
+                            CertificateUtil.publicKeyFromPem(pemCertificateBytes)))
+                        .httpClient(new HttpPipelineAdapter(options.getHttpPipeline()))
```
g2vinay: same here, setting without checking pipeline being null.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -173,12 +170,17 @@
...
+                        .authority(authorityUrl);
             if (options.getProxyOptions() != null) {
                 applicationBuilder.proxy(proxyOptionsToJavaNetProxy(options.getProxyOptions()));
             }
+            if (options.getHttpPipeline() != null) {
+                applicationBuilder.httpClient(new HttpPipelineAdapter(options.getHttpPipeline()));
```
g2vinay: setting it above already.

sdk/identity/azure-identity/src/test/java/com/azure/identity/implementation/IdentityClientTests.java
```diff
@@ -182,15 +182,16 @@ private void mockForClientCertificate(TokenRequestContext request, String access
...
             String cid = (String) invocation.getArguments()[0];
-            AsymmetricKeyCredential keyCredential = (AsymmetricKeyCredential) invocation.getArguments()[1];
+//            AsymmetricKeyCredential keyCredential = (AsymmetricKeyCredential) invocation.getArguments()[1];
             if (!clientId.equals(cid)) {
                 throw new MsalServiceException("Invalid clientId", "InvalidClientId");
             }
-            if (keyCredential == null || keyCredential.key() == null) {
-                throw new MsalServiceException("Invalid clientCertificate", "InvalidClientCertificate");
-            }
+//            if (keyCredential == null || keyCredential.key() == null) {
```
g2vinay: why are we commenting them out ?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Removing in favor of the other checked one

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: This actually should be used

sdk/identity/azure-identity/src/test/java/com/azure/identity/implementation/IdentityClientTests.java
jianghaolu: The certificate / secret now is not accessible anymore so I cannot verify it here - I was planning to fix the tests later

sdk/identity/azure-identity/pom.xml
jianghaolu: Using 1.2.0 now

sdk/eventhubs/microsoft-azure-eventhubs/pom.xml
```diff
@@ -76,7 +76,7 @@
     <dependency>
       <groupId>com.microsoft.azure</groupId>
       <artifactId>msal4j</artifactId>
-      <version>0.5.0-preview</version> <!-- {x-version-update;com.microsoft.azure:msal4j;external_dependency} -->
+      <version>1.3.0</version> <!-- {x-version-update;com.microsoft.azure:msal4j;external_dependency} -->
```
srnagar: This is track 1 event hubs module. Did you intend to update msal version of track 1 module too? This might need service team's validation.

jianghaolu: Seems to be only used in tests. And I don't know a way to specify a dependency of 2 different versions across our SDKs. 

+ @conniey  @mssfang for their opinion

srnagar: Is this because you are using the script to update the external dependency version and it's updating both track 1 and track 2 dependencies?

sdk/identity/azure-identity/src/test/java/com/azure/identity/implementation/IdentityClientTests.java
jianghaolu: I added it back though we still cannot access the `.key()` value

PR #7126, opened 2020-01-04
Ability to pass connection policy configuration values to http client in transport layer

kushagraThapar:
Currently end users cannot configure http client created in http transport client layer. Connection Policy configuration fields like idleConnectionTimeout, max pool size, and http proxy do not get passed to the http client created in HttpTransportClient. 

* This PR fixes that issue by passing connectionPolicy object to HttpTransportClient so as to be able to configure the http client by these configurations. 
* Walmart is one example, where they want to set a higher value for idle connection timeout.

### DIFF
--- a/sdk/cosmos/CODEOWNERS
+++ b/sdk/cosmos/CODEOWNERS
@@ -7,4 +7,4 @@
 #   https://help.github.com/articles/about-codeowners/  
 
 # Default owner for repo
-*	@moderakh @christopheranderson @kushagraThapar
+*	@moderakh @kushagraThapar @David-Noble-at-work @kirankumarkolli
--- a/sdk/cosmos/changelog/README.md
+++ b/sdk/cosmos/changelog/README.md
@@ -1,5 +1,28 @@
 ## Changelog
+### unreleased
+- Added the ability to select default Direct TCP options
 
+  In priority order we will take default Direct TCP options from:
+  
+  - The string value of system property `"azure.cosmos.directTcp.options"`
+ 
+    Example:
+    ``` 
+    -Dazure.cosmos.directTcp.defaultOptions={\"idleEndpointTimeout\":"PT1M10S",\"maxChannelsPerEndpoint\":10,\"maxRequestsPerChannel\":30}
+    ```
+   - The contents of the file located by the system property "azure.cosmos.directTcp.defaultOptionsFile".
+  
+     Example:
+     ```
+     -Dazure.cosmos.directTcp.defaultOptionsFile=/path/to/default/options/file
+     ```
+  
+   - The contents of the resource file named `"azure.cosmos.directTcp.options.json"`
+  
+   Otherwise, if none of these values are set or an error occurs we create default options based on a set of hard-wired 
+   values defined in the default private parameterless constructor for `RntbdTransportClient.Options` as we did before 
+   this release.
+    
 ### 3.5.0
 - Changefeed bug fixes (includes all the bug fixes released in 3.3.3) 
 - Disaster Recovery related bug fixes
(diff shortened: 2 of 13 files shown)


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpClient.java
```diff
@@ -36,7 +36,7 @@ static HttpClient createFixed(HttpClientConfig httpClientConfig) {
         }
 
         //  Default pool size
-        Integer maxPoolSize = httpClientConfig.getConfigs().getReactorNettyMaxConnectionPoolSize();
+        Integer maxPoolSize = httpClientConfig.getConfigs().getDirectHttpsMaxConnectionLimit();
```
moderakh: isn't `HttpClient.createFixed(.)` also used for GW?
How does `HttpClient.createFixed(.)` distinguish between GW and DirectHttps?

kushagraThapar: Yes, it is used for both Gateway and Direct Https. They are different and I think that is the problem. 
There is no way to set idle connection timeout, max pool size, etc. for Direct Https Client, because it doesn't have `connectionPolicy` object . 
While Gateway Https client reads these values from `connectionPolicy` object which gets passed to it when we create https client. 

To solve this problem, in this PR, I have kept same configuration for Direct Https and Gateway Http Clients. So that end users can configure these http clients through `connectionPolicy` object. 

I believe this problem also persists in V2. 
You can see both  Gateway and Direct Http clients are built using different configuration. 
Gateway: https://github.com/Azure/azure-cosmosdb-java/blob/master/sdk/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/RxDocumentClientImpl.java#L392
Direct Https: https://github.com/Azure/azure-cosmosdb-java/blob/master/direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/HttpTransportClient.java#L81

kushagraThapar:
I looked at the CI failure logs.
ReadMyWrites failed because of Conflict exception (resource already exists). This is the same issue we fixed on v2 (don't retry on writes). .
I have ported the fix to both v3 and v4.
Let's merge those ports first, then we can run the CI on this PR again.

Btw, here is the port for V3: https://github.com/Azure/azure-sdk-for-java/pull/7240

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/HttpClient.java
moderakh: @kushagraThapar that is kind of on purpose.

GW and Direct are two different modes, and even in Direct mode you may need different configuration setting for the connections to BE nodes vs connections to GW (for query).

As an example, RNTBD has its own way of configuration for its connection stack.
I think we might need ConnectionPolicy#idleTimeToGW and ConnectionPolicy#idleTimeForDiredct something along on those lines. similar for connection pool size.

Let's say if you have 4000 replicas your connection pool size for direct at least has to be >= 4000. but as we know even in Direct mode some request go to GW we need a separate connection pool to GW. With the change you made, increasing Direct connection pool size will unnecessarily increase the connection pool size for GW as well which is not the right thing to do.

Ping me to discuss this offline over teams.

kushagraThapar: Agreed, let's talk offline.

joshfree:
@kushagraThapar can this 9-month old PR be closed out?

kushagraThapar:
> @kushagraThapar can this 9-month old PR be closed out?

lol, yeah let me close it :)

kushagraThapar:
This is done using `GatewayConnectionConfig` and `DirectConnectionConfig`. No need for this PR anymore.

PR #7144, opened 2020-01-06
KV version updates Jan 2020

g2vinay:


sdk/keyvault/azure-security-keyvault-keys/README.md
```diff
@@ -16,7 +16,7 @@ Maven dependency for Azure Key Client library. Add it to your project's pom file
 <dependency>
     <groupId>com.azure</groupId>
     <artifactId>azure-security-keyvault-keys</artifactId>
-    <version>4.1.0-beta.1</version>
+    <version>4.0.2</version>
```
srnagar: This may need a minor version bump instead of patch version bump since you are updating core to a version that has minor version bump.

weshaggard: That already happened in 4.0.1 https://search.maven.org/artifact/com.azure/azure-security-keyvault-keys/4.0.1/jar. In either case I don't think that new dependency necessary requires a minor version bump, see https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api which has some info on this case.

srnagar: In most cases, updating dependency without changing public API is fine doesn't require a minor version bump. However, if the underlying dependency upgrades result in a significantly different behavior, then it's worth bumping the minor version of this library too.

In this case, core has several changes that impact how the KV APIs behave. For e.g. iterable stream now lazily fetches pages instead of eagerly fetching all pages.

weshaggard: That is fair if the behavior is detectable by consumers of KV library as a new behavior then you should bump the minor version.

g2vinay: minor version bumped.

PR #7130, opened 2020-01-05
Reactor netty connection pooling

kushagraThapar:
* Removed explicit disposing connections after reading response. 
* Reactor netty automatically puts the connection back to connection pool after response is read: https://gitter.im/reactor/reactor-netty?at=5e11accfa769bf30fd3e8c0f
* Addresses this issue: https://github.com/Azure/azure-sdk-for-java/issues/6973
* Tested this PR with ReadThroughput benchmark and the results show that connections are being re-used
* Here are some snapshots when running readThroughput in Https and gateway mode. 
** Https mode: 
```
consistency-test@consistency-tests-machine:~/azure-sdk-for-java/sdk/cosmos$ netstat -ant | awk '{print $6}' | sort | uniq -c | sort -n
      1 Foreign
      1 established)
      3 ESTABLISHED
      3 LISTEN
consistency-test@consistency-tests-machine:~/azure-sdk-for-java/sdk/cosmos$ netstat -ant | awk '{print $6}' | sort | uniq -c | sort -n
      1 Foreign
      1 established)
      3 LISTEN
   2984 ESTABLISHED
consistency-test@consistency-tests-machine:~/azure-sdk-for-java/sdk/cosmos$ netstat -ant | awk '{print $6}' | sort | uniq -c | sort -n
      1 Foreign
      1 established)
      3 LISTEN
   3094 ESTABLISHED
consistency-test@consistency-tests-machine:~/azure-sdk-for-java/sdk/cosmos$ netstat -ant | awk '{print $6}' | sort | uniq -c | sort -n
      1 Foreign
      1 established)
      3 LISTEN
   3108 ESTABLISHED
consistency-test@consistency-tests-machine:~/azure-sdk-for-java/sdk/cosmos$ netstat -ant | awk '{print $6}' | sort | uniq -c | sort -n
      1 Foreign
      1 established)
      3 LISTEN
   3152 ESTABLISHED
consistency-test@consistency-tests-machine:~/azure-sdk-for-java/sdk/cosmos$ netstat -ant | awk '{print $6}' | sort | uniq -c | sort -n
      1 Foreign
      1 established)
      3 LISTEN
   3156 ESTABLISHED
consistency-test@consistency-tests-machine:~/azure-sdk-for-java/sdk/cosmos$ netstat -ant | awk '{print $6}' | sort | uniq -c | sort -n
      1 Foreign
      1 established)
      3 LISTEN
   3181 ESTABLISHED
[...]

Issue #6973, opened 2019-12-19
Improve Reactor Netty Connection Pooling


kushagraThapar:
**Describe the bug**
Rector Netty Connection Pooling is not properly re-using the connections. 
Either we are disposing the connections too soon, or there is some other configurations which is not setup correctly, but we are seeing a lot of connections getting created in a short period of time frequently. 

**To Reproduce**
This can easily be reproduced by running ReadThroughput benchmarks on v3 / v4 with high concurrency (for example 2000)

* Too many connections seen in TIME_WAIT ("netstat | grep TIME_WAIT  | wc -l" ==> 150k-200k)

***Code Snippet***
We expect ReactorNettyClient.java has the issues. 

**Expected behavior**
These are the goals: 

* Connections creations are expensive and hence re-use aggressively
* In an ideal set-up (with not many NW failures) all connections are re-usable.
* Gateway mode with a fixed concurrency ideally should only create connections linear to concurrency -> connection re-use

kushagraThapar:
This has been fixed on all SDK versions.

PR #7145, opened 2020-01-06
ported fix from v3

moderakh:
ports the fix https://github.com/Azure/azure-sdk-for-java/pull/7130 to v4

Addresses this issue: https://github.com/Azure/azure-sdk-for-java/issues/6973


kushagraThapar:
I looked at the CI failure logs. 
ReadMyWrites failed because of Conflict exception (resource already exists). This is the same issue we fixed on v2 (don't retry on writes). .
I have ported the fix to both v3 and v4. 
Let's merge those ports first, then we can run the CI on this PR again. 

Btw, here is the port for V4: https://github.com/Azure/azure-sdk-for-java/pull/7243

PR #7146, opened 2020-01-06
Cosmosdb release 3.6.0-beta.1

kushagraThapar:
- Added the ability to select default Direct TCP options

  In priority order we will take default Direct TCP options from:
  
  - The string value of system property `"azure.cosmos.directTcp.options"`
 
    Example:
    ``` 
    -Dazure.cosmos.directTcp.defaultOptions={\"idleEndpointTimeout\":"PT1M10S",\"maxChannelsPerEndpoint\":10,\"maxRequestsPerChannel\":30}
    ```
   - The contents of the file located by the system property "azure.cosmos.directTcp.defaultOptionsFile".
  
     Example:
     ```
     -Dazure.cosmos.directTcp.defaultOptionsFile=/path/to/default/options/file
     ```
  
   - The contents of the resource file named `"azure.cosmos.directTcp.options.json"`
  
   Otherwise, if none of these values are set or an error occurs we create default options based on a set of hard-wired 
   values defined in the default private parameterless constructor for `RntbdTransportClient.Options` as we did before 
   this release.
- Fixed Reactor Netty Http Connection Re-use issue.

### DIFF
--- a/eng/versioning/version_data.txt
+++ b/eng/versioning/version_data.txt
@@ -23,8 +23,10 @@ com.microsoft.azure.cognitiveservices:azure-cognitiveservices-contentmoderator;1
 com.microsoft.azure.cognitiveservices:azure-cognitiveservices-customvision-prediction;1.1.0-beta.3;1.1.0-beta.3
 com.microsoft.azure.cognitiveservices:azure-cognitiveservices-customvision-training;1.1.0-beta.3;1.1.0-beta.3
 com.microsoft.azure.cognitiveservices:azure-cognitiveservices-faceapi;1.1.0-beta.1;1.1.0-beta.1
-com.microsoft.azure:azure-cosmos-parent;3.6.0-beta.1;3.6.0-beta.1
-com.microsoft.azure:azure-cosmos;3.6.0-beta.1;3.6.0-beta.1
+# azure-cosmos-parent doesn't have a dependency management section anymore, it doesn't need to change unless parent pom changes 
+com.microsoft.azure:azure-cosmos-parent;3.5.0;3.5.0
+com.microsoft.azure:azure-cosmos;3.5.0;3.6.0-beta.1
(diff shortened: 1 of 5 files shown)


eng/versioning/version_data.txt
```diff
@@ -23,10 +23,12 @@ com.microsoft.azure.cognitiveservices:azure-cognitiveservices-contentmoderator;1
...
 com.microsoft.azure.cognitiveservices:azure-cognitiveservices-customvision-training;1.1.0-beta.3;1.1.0-beta.3
 com.microsoft.azure.cognitiveservices:azure-cognitiveservices-faceapi;1.1.0-beta.1;1.1.0-beta.1
-com.microsoft.azure:azure-cosmos-parent;3.5.0;3.5.1-beta1
-com.microsoft.azure:azure-cosmos;3.5.0;3.5.1-beta1
-com.microsoft.azure:azure-cosmos-benchmark;3.5.1-beta1;3.5.1-beta1
-com.microsoft.azure:azure-cosmos-examples;3.5.1-beta1;3.5.1-beta1
+# azure-cosmos-parent doesn't have a dependency management section anymore, it doesn't need to change unless parent pom changes 
+com.microsoft.azure:azure-cosmos-parent;3.5.0;3.5.0
+com.microsoft.azure:azure-cosmos;3.5.0;3.6.0-beta.1
+# azure-cosmos-benchmark and azure-cosmos-examples don't get released, and should be set to azure-cosmos current-version
```
weshaggard: Why set them to a moving target which is "azure-cosmos" current-version? Wouldn't it be easier if they were set to a fixed version?

kushagraThapar: We have always kept them in sync with azure-cosmos version. That is the only reason to keep incrementing them so as to avoid confusion within the Cosmos SDK team. 

Once we publish v4, we will try to get rid of `azure-cosmos-parent`, `azure-cosmos-benchmark` and `azure-cosmos-examples` 

For now, we would like to keep them updated with azure-cosmos current version.

PR #7151, opened 2020-01-06
Update AzConfig release version to 1.0.1

mssfang:


eng/jacoco-test-coverage/pom.xml
```diff
@@ -80,7 +80,7 @@
     <dependency>
```
conniey: Did you run the update script? I thought this was also referenced in our spotbugs aggregate report pom.

mssfang: Yes. I ran the update script

mssfang: python eng\versioning\update_versions.py --update-type all --build-type client

conniey: Now I'm confused because I can see it here:
https://github.com/Azure/azure-sdk-for-java/blob/master/eng/spotbugs-aggregate-report/pom.xml#L158

mssfang: I asking Jim for the solution. Maybe the script doesn't work properly?

PR #7156, opened 2020-01-06
Adding retry context to diagnostics and moving to JSON format

simplynaveen20:
Adding retry context to diagnostics which will add information -  retry count , time spent in retry and types of error encountered.
Also moving diagnostic format from string builder to json.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/RetryContextOnDiagnosticTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/RetryContextOnDiagnosticTest.java
@@ -155,7 +155,7 @@ public class RetryContextOnDiagnosticTest extends TestSuiteBase {
         validateServiceResponseSuccess(responseFlux);
 
         Mockito.verify(retryPolicy, Mockito.times(2)).getRetryCount();
-        Mockito.verify(retryPolicy, Mockito.times(2)).getStatusAndSubStatusCodes();
+        Mockito.verify(retryPolicy, Mockito.times(3)).getStatusAndSubStatusCodes();
 
         retryPolicy = Mockito.mock(TestRetryPolicy.class);
         Mockito.when(retryPolicy.getRetryCount()).thenReturn(1);
@@ -164,7 +164,7 @@ public class RetryContextOnDiagnosticTest extends TestSuiteBase {
         validateServiceResponseSuccess(responseFlux);
 
         Mockito.verify(retryPolicy, Mockito.times(2)).getRetryCount();
-        Mockito.verify(retryPolicy, Mockito.times(2)).getStatusAndSubStatusCodes();
+        Mockito.verify(retryPolicy, Mockito.times(3)).getStatusAndSubStatusCodes();
 
         retryPolicy = Mockito.mock(TestRetryPolicy.class);
         Mockito.when(retryPolicy.getRetryCount()).thenReturn(1);
@@ -173,7 +173,7 @@ public class RetryContextOnDiagnosticTest extends TestSuiteBase {
         validateServiceResponseSuccess(responseFlux);
 
         Mockito.verify(retryPolicy, Mockito.times(2)).getRetryCount();
-        Mockito.verify(retryPolicy, Mockito.times(2)).getStatusAndSubStatusCodes();
+        Mockito.verify(retryPolicy, Mockito.times(3)).getStatusAndSubStatusCodes();
 
         retryPolicy = Mockito.mock(TestRetryPolicy.class);
         Mockito.when(retryPolicy.getRetryCount()).thenReturn(1);
@@ -182,7 +182,7 @@ public class RetryContextOnDiagnosticTest extends TestSuiteBase {
         validateServiceResponseSuccess(responseFlux);
 
         Mockito.verify(retryPolicy, Mockito.times(2)).getRetryCount();
-        Mockito.verify(retryPolicy, Mockito.times(2)).getStatusAndSubStatusCodes();
+        Mockito.verify(retryPolicy, Mockito.times(3)).getStatusAndSubStatusCodes();
 
         retryPolicy = Mockito.mock(TestRetryPolicy.class);
         Mockito.when(retryPolicy.getRetryCount()).thenReturn(1);
@@ -191,7 +191,7 @@ public class RetryContextOnDiagnosticTest extends TestSuiteBase {
         validateServiceResponseSuccess(responseFlux);
 
         Mockito.verify(retryPolicy, Mockito.times(2)).getRetryCount();
-        Mockito.verify(retryPolicy, Mockito.times(2)).getStatusAndSubStatusCodes();
+        Mockito.verify(retryPolicy, Mockito.times(3)).getStatusAndSubStatusCodes();
 
         cosmosClient.close();
     }


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RetryPolicyWithDiagnostics.java
```diff
@@ -0,0 +1,63 @@
...
+
+package com.azure.cosmos.implementation;
+
+import java.time.Duration;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class RetryPolicyWithDiagnostics implements IRetryPolicy{
```
simplynaveen20: We could remove IRetryPolicy interface altogether and just use this abstract class as RetryPolicy

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RetryContext.java
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.cosmos.implementation;
+
+import java.util.List;
+
+public class RetryContext {
```
simplynaveen20: Did manual testing , also planning to add mock tests to verify all the property of RetryContext in both connection mode and on various operations.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
```diff
@@ -39,20 +44,25 @@
         DateTimeFormatter.ofPattern("dd MMM yyyy HH:mm:ss" + ".SSS").withLocale(Locale.US);
     private final static OperatingSystemMXBean mbean = (com.sun.management.OperatingSystemMXBean)ManagementFactory.getOperatingSystemMXBean();
 
-    private final ZonedDateTime requestStartTime;
-    private ZonedDateTime requestEndTime;
+    private final ObjectMapper objectMapper = new ObjectMapper();
```
moderakh: `ObjectMapper` instantiation is very expensive, and this is instantiated per request.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
```diff
@@ -168,63 +183,25 @@ void recordAddressResolutionEnd(String identifier) {
...
-            for (StoreResponseStatistics storeResponseStatistics : this.responseStatisticsList) {
-                stringBuilder.append(storeResponseStatistics.toString()).append(System.lineSeparator());
-            }
-
-            //  take all responses here - this should be limited in number and each one is important.
-            for (AddressResolutionStatistics value : this.addressResolutionStatistics.values()) {
-                stringBuilder.append(value.toString()).append(System.lineSeparator());
-            }
-
+            requestLatency= getRequestLatency().toMillis();
```
moderakh: toString() should not change the state of the object. This is somewhat an anti-pattern

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
```diff
@@ -168,63 +183,25 @@ void recordAddressResolutionEnd(String identifier) {
...
+                this.supplementalResponseStatisticsList.removeAll(this.supplementalResponseStatisticsList.subList(0, initialIndex));
             }
-            for (int i = initialIndex; i < supplementalResponseStatisticsListCount; i++) {
-                stringBuilder.append(this.supplementalResponseStatisticsList.get(i).toString())
-                    .append(System.lineSeparator());
+            try {
+                printSystemInformation();
+                return objectMapper.writeValueAsString(this);
+            } catch (JsonProcessingException e) {
+                e.printStackTrace();
```
moderakh: failure should be logged using logger

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/IRetryPolicy.java
```diff
@@ -25,6 +26,20 @@
...
+
+    void captureStartTimeIfNotSet();
+
+    void updateEndTime();
+
+    Duration getRetryLatency();
+
+    void addStatusAndSubStatusCode(int statusCode, int subStatusCode);
+
+    List<int[]> getStatusAndSubStatusCodes();
```
moderakh: please note that shouldRetry() is async and some other methods you added are not async. I don't see a specific issue here, but just please be cautious when mixing these sync vs async methods.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/RetryContextOnDiagnosticTest.java
```diff
@@ -0,0 +1,239 @@
...
+        CosmosClientException exception = new CosmosClientException(410, exceptionText);
+        Mockito.when(callbackMethod.call()).thenThrow(exception);
+        ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
+        executor.schedule(() -> {
+            ((TestRetryPolicy) retryPolicy).noRetry = true;
+        }, 10, TimeUnit.SECONDS);
+        Mono<StoreResponse> monoResponse = BackoffRetryUtility.executeRetry(callbackMethod, retryPolicy);
+        validateFailure(monoResponse);
+
+        Assertions.assertThat(retryPolicy.getRetryCount()).isGreaterThanOrEqualTo(5);
```
moderakh: please use static import for `assertThat` so that similar to other tests instead of `Assertions.assertThat` you do `assertThat` here and in other tests.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RetryContext.java
simplynaveen20: Added the mock tests

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
simplynaveen20: Moved to static

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
simplynaveen20: Are we talking about supplementalResponseStatisticsList , where we chopping off the top 10 items? I am aware of it , did it on purpose , no logic issue in this particular case, however if we want to avoid it , then we need to come up with probably another list , which will add overhead. Lets discuss offline for better solution

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/RetryContextOnDiagnosticTest.java
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/IRetryPolicy.java
simplynaveen20: Point noted for future reference.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RetryPolicyWithDiagnostics.java
simplynaveen20: Thoughts ?

kushagraThapar: This provides default implementations for all APIs of IRetryPolicy. 
If we don't plan to use IRetryPolicy altogether, then its fine, but I would suggest keeping it in the code, as it doesn't do any harm, and opens possibilities in future for more development on this if needed.

kushagraThapar: This provides default implementations for all APIs of IRetryPolicy.
If we don't plan to use IRetryPolicy altogether, then its fine, but I would suggest keeping it in the code, as it doesn't do any harm, and opens possibilities in future for more development on this if needed.

simplynaveen20: Yea it doesn't do any harm , will keep it as it is

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ClientSideRequestStatistics.java
simplynaveen20: Moved logic out of toString() of clientSideRequestDiagnostics , and using separate serializer

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosResponseDiagnostics.java
```diff
@@ -31,7 +38,12 @@ CosmosResponseDiagnostics clientSideRequestStatistics(ClientSideRequestStatistic
      */
     @Override
     public String toString() {
-        return this.clientSideRequestStatistics.toString();
+        try {
+            return objectMapper.writeValueAsString(this.clientSideRequestStatistics);
+        }catch (JsonProcessingException e) {
+            logger.error("Error while parsing diagnostics " + e.getOriginalMessage());
```
moderakh: usually errors are logged with the following pattern
```
logger.error("Error while parsing diagnostics", e)
```
any reason we are doing something else?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -658,21 +657,31 @@ private String parentResourceLinkToQueryLink(String parentResouceLink, ResourceT
...
+
         return getStoreProxy(request).processMessage(request);
     }
 
-    private Flux<RxDocumentServiceResponse> read(RxDocumentServiceRequest request) {
+    private Flux<RxDocumentServiceResponse> read(RxDocumentServiceRequest request, DocumentClientRetryPolicy documentClientRetryPolicy) {
         populateHeaders(request, HttpConstants.HttpMethods.GET);
+        if(request.requestContext != null && documentClientRetryPolicy.getRetryCount() > 0) {
+            documentClientRetryPolicy.updateEndTime();
+            request.requestContext.updateRetryContext(documentClientRetryPolicy);
```
moderakh: same question which I had raised offline.

I see that you have different way of capturing retry info in the direct mode, however here on the outer surface of the SDK we are again capturing the retry info, why do we need to capture the retry info in two different places? 

is this only applicable to GW mode if so it should be in GatewayStoreModel. no?

same for all other HttpMethods

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosResponseDiagnostics.java
simplynaveen20: no specific reason , just trying to avoid priting whole error stack . I will move to logger.error("Error while parsing diagnostics", e) in next iteration

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
simplynaveen20: Yes above is applicable to GW mode , and retry in gateway happens via BackoffRetry.executeRetry and at that place we dont have the access to DocumentServiceRequestContext

simplynaveen20: Discussed offline - 
"Lets take the example of ReadDocument()
In Gateway mode ReadDocument() will keep calling readDocumentInternal() based on BackOffRetryUtility.executeRetry() ,and each retry will call the readDocumentInternal() again and update
request.requestContext.updateRetryContext(documentClientRetryPolicy) everytime.
However in Direct mode  ReadDocument() will call readDocumentInternal(), and in case of any  direct exception (Gone) instead of call readDocumentInternal() via  BackOffRetryUtility.executeRetry(), replicate resource client will create it own retry policy and consume failure/success and response will eventually be send via StoreClient to ReadDocument , therefore the code
request.requestContext.updateRetryContext(documentClientRetryPolicy) in read() in RxDocumentClientImpl will never be executed again."

simplynaveen20:
Ran ReadMyWrites on VM , got success , PR is ready to merge
2020-01-28 20:43:13,641       [main] INFO  com.azure.cosmos.benchmark.ReadMyWriteWorkflow - [1000000] operations performed in [2430] seconds.
1/28/20, 8:43:13 PM ============================================================

#Successful Operations
             count = 1000000
         mean rate = 411.36 events/second
     1-minute rate = 424.29 events/second
     5-minute rate = 432.71 events/second
    15-minute rate = 425.19 events/second
#Unsuccessful Operations
             count = 0


2020-01-28 20:43:13,642       [main] INFO  com.azure.cosmos.implementation.RxDocumentClientImpl - Shutting down ...

PR #7175, opened 2020-01-07
Improve logging and update consumer close logic

srnagar:
Added more logging to make debugging easier and added more exception handling.


sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/src/main/java/com/azure/messaging/eventhubs/checkpointstore/blob/Messages.java
```diff
@@ -18,8 +18,10 @@
     private static Properties properties;
```
conniey: Did you update the message.properties file? iirc, these static strings were used in case the property file was not found.

https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/src/main/resources/com/azure/messaging/eventhubs/checkpointstore/blob/messages.properties

srnagar: Good catch - updated!

srnagar:
java - spring - ci build failure can be ignore for now.

PR #7176, opened 2020-01-07
KV changelog updates

g2vinay:
Updates changelog.
Drops commons codec from secrets pom.  (not allowed in track two).

### DIFF
--- a/eng/versioning/external_dependencies.txt
+++ b/eng/versioning/external_dependencies.txt
@@ -22,6 +22,7 @@ com.microsoft.azure:spotbugs-maven-plugin;1.2.1
 com.microsoft.rest:client-runtime;1.6.15
 com.microsoft.rest.v2:client-runtime;2.1.1
 com.squareup.okhttp3:okhttp;4.2.2
+commons-codec:commons-codec;1.13
 io.micrometer:micrometer-core;1.2.0
 io.micrometer:micrometer-registry-azure-monitor;1.2.0
 io.micrometer:micrometer-registry-graphite;1.2.0


sdk/keyvault/azure-security-keyvault-keys/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Release History
 
-## 4.1.0-beta.1 (Unreleased)
+## 4.1.0 (2020-01-07)
+- Fixes the logic of `getKeyId()` method in `KeyEncryptionKeyClient` and `KeyEncryptionKeyAsyncClient` to ensure key id is available in all scenarios.
+- Update azure-core dependency to latest version.
```
srnagar: Instead of saying `latest version`, specify the version number.

sdk/keyvault/azure-security-keyvault-certificates/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Release History
 
-## 4.0.0-beta.8 (Unreleased)
+## 4.0.0 (2020-01-07)
+- Update azure-core dependency to latest version.
```
srnagar: Same here, include version number.

sdk/keyvault/azure-security-keyvault-secrets/pom.xml
```diff
@@ -43,12 +43,6 @@
       <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
     </dependency>
 
-    <dependency>
-      <groupId>commons-codec</groupId>
-      <artifactId>commons-codec</artifactId>
-      <version>1.13</version> <!-- {x-version-update;commons-codec:commons-codec;external_dependency} -->
```
srnagar: Also, update the `version_client.txt` and remove this dependency. I believe this was the only place this dependency was used.

g2vinay: I am assuming external_dependencies.txt

g2vinay: Track 1 KV still references commons-codec.
So, cannot remove from external_dependencies.txt too,

eng/versioning/external_dependencies.txt
```diff
@@ -22,7 +22,6 @@ com.microsoft.azure:spotbugs-maven-plugin;1.2.1
 com.microsoft.rest:client-runtime;1.6.15
 com.microsoft.rest.v2:client-runtime;2.1.1
 com.squareup.okhttp3:okhttp;4.2.2
-commons-codec:commons-codec;1.13
```
JimSuplizio: Your PR is failing because you removed an external_dependency that is used by other libraries. In the future, if you're curious as to whether a dependency is being used you can run one of two quick commands to check:
- From a powershell 6 window, in the root of the enlistment run .\eng\pom_file_version_scanner.ps1 (thought this is only going to tell you after you remove the entry)
- For something quick and dirty on Windows you can just run findstr /spin /c:"commons-codec:commons-codec" pom*.xml

sdk/keyvault/azure-security-keyvault-secrets/pom.xml
JimSuplizio: So nothing should ever be deleted from external_dependencies.txt unless you know for an absolute fact it isn't being used by anywhere else. There are several data track libraries that require this dependency. In my comment above I put in a couple of commands that can be run from the command line to check, in the future.

eng/versioning/external_dependencies.txt
g2vinay: yep, I already reverted this change

PR #7196, opened 2020-01-07
Added basic CI file for spring boot.

mitchdenny:
Quickly merging in the skeleton we need for setting up spring boot on master (to support pipeline generation).


JimSuplizio:
@mitchdenny this pipeline has pretty much been failing since you'd checked in yesterday https://dev.azure.com/azure-sdk/public/_build?definitionId=1326&_a=summary

mitchdenny:
That's deliberate (although the YAML syntax error wasn't so thanks for fixing that). I needed this ci.yml file in to trigger automatic pipeline generation but the spring boot bits will be done on a separate branch.

PR #7228, opened 2020-01-07
Enable build failure when code coverage is below threshold

srnagar:
This PR includes two changes:
- Fails build if the code coverage drops below threshold
- Includes jacoco test coverage reporting steps in `archetype-sdk-client.yml` which were previously configured in `client.yml`

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -104,6 +104,8 @@
     <site.url>https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-java</site.url>
     <issues.url>https://github.com/Azure/azure-sdk-for-java/issues</issues.url>
     <build.context>azure-client-sdk-parent</build.context>
+    <jacoco.min.linecoverage>0.40</jacoco.min.linecoverage>
+    <jacoco.min.branchcoverage>0.30</jacoco.min.branchcoverage>
   </properties>
 
   <build>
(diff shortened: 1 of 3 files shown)


pom.client.xml
```diff
@@ -333,6 +333,31 @@
               <outputDirectory>${project.reporting.outputDirectory}/test-coverage</outputDirectory>
             </configuration>
           </execution>
+          <execution>
```
JimSuplizio: @JonathanGiles is this particular change going to require us to up the version and re-release the parent pom?

srnagar: @JimSuplizio since this change only introduces a check to fail the build when test coverage is low, it's not end user-facing. So, version update and re-release is not necessary.

pom.client.xml
```diff
@@ -333,6 +333,31 @@
...
+                  <limits>
+                    <limit>
+                      <counter>LINE</counter>
+                      <value>COVEREDRATIO</value>
+                      <minimum>0.18</minimum>
+                    </limit>
+                    <limit>
+                      <counter>BRANCH</counter>
+                      <value>COVEREDRATIO</value>
+                      <minimum>0.09</minimum>
```
JonathanGiles: Is it worth extracting these two minimum values out to properties so that child poms may overwrite this (in the future when we start setting a valid minimum and libraries need to opt-out for a valid reason)?

srnagar: Yes, updated this to use properties and also customized the threshold for a couple of storage modules that had really low coverage to:
- demonstrate how to customize the threshold per module
- increase the base threshold to a much higher number than 9%

PR #7243, opened 2020-01-07
Don't retry on network failure on writes

kushagraThapar:
Direct Port of Azure/azure-cosmosdb-java#301

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/ClientRetryPolicy.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/ClientRetryPolicy.java
@@ -158,7 +158,7 @@ public class ClientRetryPolicy implements IDocumentClientRetryPolicy {
         // Therefore just skip the retry here to avoid the delay because retrying won't go through in the end.
 
         Duration retryDelay = Duration.ZERO;
-        if (!this.isReadRequest) {
+        if (!isReadRequest) {
             logger.debug("Failover happening. retryCount {}",  this.failoverRetryCount);
             if (this.failoverRetryCount > 1) {
                 //if retried both endpoints, follow regular retry interval.
@@ -183,7 +183,7 @@ public class ClientRetryPolicy implements IDocumentClientRetryPolicy {
         this.failoverRetryCount++;
 
         // Mark the current read endpoint as unavailable
-        if (this.isReadRequest) {
+        if (isReadRequest) {
             logger.warn("marking the endpoint {} as unavailable for read",this.locationEndpoint);
             this.globalEndpointManager.markEndpointUnavailableForRead(this.locationEndpoint);
         } else {


kushagraThapar:
> Looks good to me. As @simplynaveen20 has done some work on fail over retry. that would be great Naveen if you also take a look.

Yes, I made sure to take a look at DR fixes PR for v2 before porting this change. 
https://github.com/Azure/azure-cosmosdb-java/pull/290

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/ClientRetryPolicy.java
```diff
@@ -173,8 +173,17 @@ private ShouldRetryResult shouldRetryOnSessionNotAvailable() {
             retryDelay = Duration.ofMillis(ClientRetryPolicy.RetryIntervalInMS);
         }
         this.retryContext = new RetryContext(this.failoverRetryCount, false);
-        return this.globalEndpointManager.refreshLocationAsync(null, forceRefresh)
-                .then(Mono.just(ShouldRetryResult.retryAfter(retryDelay)));
+        Mono<Void> completable = this.globalEndpointManager.refreshLocationAsync(null, forceRefresh);
+        if (isReadRequest || WebExceptionUtility.isWebExceptionRetriable(e)) {
```
simplynaveen20: HttpConstants.StatusCodes.FORBIDDEN, HttpConstants.SubStatusCodes.FORBIDDEN_WRITEFORBIDDEN will qualify as isWebExceptionRetriable ? If not then how does that case handled ?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/ClientRetryPolicy.java
```diff
@@ -173,8 +173,17 @@ private ShouldRetryResult shouldRetryOnSessionNotAvailable() {
...
+            // refresh cache and
+            // if it is a read request or if it is a write but we are sure the write
+            // hasn't reached the service retry
+            return completable.then(Mono.just(ShouldRetryResult.retryAfter(retryDelay)));
+        } else {
+            // refresh cache and
+            // no retry for writes which we are not sure if have reached to the service or not
+            return completable.then(Mono.just(ShouldRetryResult.noRetry()));
```
simplynaveen20: Isn't heading of PR is contradicting , we are retrying on write in case of network failure ? Until I am missing anything

kushagraThapar: We are only retrying if it is a readRequest, or WebExceptionRetriable. 
In the else case, we are not retrying.. we are calling noRetry()

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/ClientRetryPolicy.java
kushagraThapar: I don't think it will be qualified as `WebExceptionRetriable`
And in that case, we force refresh but I don't think we will retry. 
Worth discussing over a meeting.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/ClientRetryPolicy.java
simplynaveen20: Exactly , but from heading “dont retry on network failure” Lets discuss offline

simplynaveen20: Ok so it is retry on write in case of connection exception , thinking other way round was confusing.

moderakh:
Thank you Naveen for finding the bug. :-)

PR #7244, opened 2020-01-07
Update identity to 1.0.2 everywhere

jianghaolu:


jianghaolu:
Per @JimSuplizio java-client-reporting should not be here. Ignoring it.

Issue #5445, opened 2019-09-20
Set Container default TTL via CosmosSDK v3


dkroehan:
In the CosmosSDK v2 it was possible to set the default TimeToLive when creating a new DocumentCollection, by using: _DocumentCollection#setDefaultTimeToLive_

I couldn't find a possibility how to do that with the new CosmosSDK v3.

shurd:
Hi @dkroehan thank you for opening this issue. Looking through the V3 code, it does not look like you can currently set the default TTL when creating a container. Based on the implementation of the [.NET v3 SDK](https://docs.microsoft.com/en-us/azure/cosmos-db/how-to-time-to-live#dotnet-enable-withexpiry), the property would likely be in the "CosmosContainerProperties.java" class. @christopheranderson would know more about this issue.

kushagraThapar:
This API has been exposed in all SDK versions now.

PR #7245, opened 2020-01-07
Exposed default time to live API on CosmosContainerProperties

kushagraThapar:
Direct port of this PR: https://github.com/Azure/azure-sdk-for-java/pull/6651
Addresses this: #5445

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainerProperties.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainerProperties.java
@@ -201,7 +201,7 @@ public class CosmosContainerProperties extends Resource {
      *
      * @return the default time-to-live value in seconds.
      */
-    public Integer getDefaultTimeToLive() {
+    public Integer getDefaultTimeToLiveInSeconds() {
         if (super.has(Constants.Properties.DEFAULT_TTL)) {
             return super.getInt(Constants.Properties.DEFAULT_TTL);
         }
@@ -226,7 +226,7 @@ public class CosmosContainerProperties extends Resource {
      *
      * @param timeToLive the default time-to-live value in seconds.
      */
-    public void setDefaultTimeToLive(Integer timeToLive) {
+    public void setDefaultTimeToLiveInSeconds(Integer timeToLive) {
         // a "null" value is represented as a missing element on the wire.
         // setting timeToLive to null should remove the property from the property bag.
         if (timeToLive != null) {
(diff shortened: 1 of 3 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainerProperties.java
```diff
@@ -196,6 +196,46 @@ public CosmosContainerProperties setConflictResolutionPolicy(ConflictResolutionP
         return this;
     }
 
+    /**
+     * Gets the collection's default time-to-live value.
+     *
+     * @return the default time-to-live value in seconds.
+     */
+    public Integer getDefaultTimeToLive() {
```
moderakh: shouldn't we include the unit?

`getDefaultTimeToLiveInSeconds()` ?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainerProperties.java
```diff
@@ -196,6 +196,46 @@ public CosmosContainerProperties setConflictResolutionPolicy(ConflictResolutionP
...
+     * <p>
+     * nonzero positive integer - indicates the default time-to-live value for all documents within the collection. This value can be overridden
+     * by individual documents' time-to-live value.
+     * <p>
+     * -1 - indicates by default all documents within the collection never expire. This value can be overridden by individual documents'
+     * time-to-live value.
+     *
+     * @param timeToLive the default time-to-live value in seconds.
+     */
+    public void setDefaultTimeToLive(Integer timeToLive) {
```
moderakh: ditto

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainerProperties.java
kushagraThapar: We can if we want to, I am fine with it. 
Though the API has been called `getDefaultTimeToLive()` in both v2 and v3. 
Should we change it ?

moderakh: Let's decide what is the right thing to do and we do that for v4 as v4 is not released yet.

kushagraThapar: Agreed, @kirankumarkolli @mbhaskar @David-Noble-at-work  do you guys have any opinion on this ?
Mo and I are in agreement of changing it to `getDefaultTimeToLiveInSeconds()` and `setDefaultTimeToLiveInSeconds()`

kushagraThapar: Done.

PR #7251, opened 2020-01-07
Enable Complete Test Matrix in Storage

alzimmermsft:
Removes the `matrix` configuration in the `tests.yml` file for Storage. This triggers Storage tests to inherit the test matrix from the parent template which specifies the complete testing matrix (Linux, Windows, and Mac X Java 8 and 11 [LTS]).

Additionally, changes the `windowUntil` prefetch from `maxSingleUploadSize` to `Integer.MAX_VALUE` to leverage an unbounded feature in Reactor. Previously, there was an issue that arose when the stream completed before triggering a window to be generated which lead to a deadlock/infinite loop state which hung the application, this was remedied by setting the prefetch limit to an arbitrarily high number. Recently it was discovered that this limit would generate an `Object` array the size of the limit, using `Integer.MAX_VALUE` triggers a specialized unbounded scenario where the prefetch is unlimited but doesn't generate the array. This change continues to fix the initially found issue while reducing the memory overhead introduced by creating the array.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
@@ -430,9 +430,9 @@ public class BlobAsyncClient extends BlobAsyncClientBase {
                 }
             /*
              * Use cutBefore = true as we want to window all data under 4MB into one window.
-             * Set the prefetch to the maxSingleUploadSize in the case that there are numerous tiny buffers,
-             * windowUntil uses a default limit of 256 and once that is hit it will trigger onComplete which causes
-             * downstream issues.
+             * Set the prefetch to 'Integer.MAX_VALUE' to leverage an unbounded fetch limit in case there are numerous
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -434,7 +434,7 @@ as we can guarantee we only need at most two buffers for any call to write (two
              * windowUntil uses a default limit of 256 and once that is hit it will trigger onComplete which causes
              * downstream issues.
              */
-            }, true, parallelTransferOptions.getMaxSingleUploadSize())
+            }, true, Integer.MAX_VALUE)
```
mikeharder: Update code comment to align with this change

alzimmermsft: Done

alzimmermsft: Updated

PR #7253, opened 2020-01-07
Fix for Cosmos dev feed packages

mitchdenny:
Cosmos dev feed packages weren't being published because it uses a different template and the artifact list wasn't being propogated down into the build job template which is necessary for versioning stamping to occur. This PR should fix that.


JimSuplizio:
Thanks for doing this Mitch!

PR #7258, opened 2020-01-08
Use system out to log message when there is no SLF4J binding

sima-zhu:
Logging strategy:
1. When there is slf4j binding, log at the default log level set for slf4j impl.
1. If no slf4j binding, log at env variable log level.

### DIFF
--- a/sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
@@ -16,7 +16,6 @@ import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 


sdk/core/azure-core/pom.xml
```diff
@@ -120,12 +120,12 @@
...
-      <artifactId>slf4j-simple</artifactId>
-      <version>1.7.25</version> <!-- {x-version-update;org.slf4j:slf4j-simple;external_dependency} -->
-      <scope>test</scope>
-    </dependency>
+<!--    <dependency>-->
+<!--      <groupId>org.slf4j</groupId>-->
+<!--      <artifactId>slf4j-simple</artifactId>-->
+<!--      <version>1.7.28</version> &lt;!&ndash; {x-version-update;org.slf4j:slf4j-simple;external_dependency} &ndash;&gt;-->
+<!--      <scope>test</scope>-->
+<!--    </dependency>-->
```
JonathanGiles: Delete this if it is not required.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -50,9 +60,11 @@ public ClientLogger(Class<?> clazz) {
      * Retrieves a logger for the passed class name using the {@link LoggerFactory}.
      *
      * @param className Class name creating the logger.
+     * @throws RuntimeException it is an error.
```
JonathanGiles: Provide better descriptive text than this! What causes the error?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -237,37 +294,44 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
-            case ERROR:
-                return logger.isErrorEnabled();
-            default:
-                return false;
+    /**
+     * Checking the system log with the preference order of slf4j and environment variable.
+     *
+     * @return The log level.
+     */
+    private LogLevel loadSystemLogLevel() {
```
JonathanGiles: This method name is not clear enough. You aren't 'loading', this is more like a 'getConfiguredLogLevel()' or something?

sdk/core/azure-core/src/main/java/module-info.java
```diff
@@ -14,7 +14,7 @@
     requires transitive org.reactivestreams;
 
     requires transitive org.slf4j;
-
+    requires transitive java.logging;
```
alzimmermsft: Why do we require either of these to be transitive? Do we have any public APIs which require an SLF4J of Java logging object?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -36,6 +39,13 @@
  */
 public class ClientLogger {
     private final Logger logger;
+    private final String className;
+
+    private static final String LOG_MESSAGE_FORMAT = "%s: %s %s%n%s";
+    private static final String LOG_LEVEL_VERBOSE = "VERBOSE";
+    private static final String LOG_LEVEL_INFO = "INFO";
+    private static final String LOG_LEVEL_WARN = "WARN";
+    private static final String LOG_LEVEL_ERROR = "ERROR";
```
alzimmermsft: Could this be a property on the `LogLevel` enum?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -207,11 +207,68 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
+                System.out.println(errorMessage);
+                break;
+            default:
+                // Don't do anything, this state shouldn't be possible.
+                break;
+        }
+    }
+
+    private boolean isLogFromEnv() {
+        return logger instanceof NOPLogger;
```
JonathanGiles: We create the logger once and it is final. We should not be wasting cycles on an instanceof check every time we do a log call.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -207,11 +207,68 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
+        Object... args) {
+        if (isFromEnv) {
+            printFromDefaultLogger(throwableMessage, logLevel, format, args);
+        } else {
+            printFromLogger(throwableMessage, logLevel, format, args);
+        }
+    }
+
+    private void printFromDefaultLogger(String throwableMessage, LogLevel logLevel, String format, Object... args) {
+        final DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
```
JonathanGiles: Think more about reducing CPU cycles. This date format does not need to be recreated on every call and could be a static final field in the class.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -207,11 +207,68 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
+        Date date = new Date();
+        String dateTime = dateFormat.format(date);
+        String classPath = className.getClass().getCanonicalName();
```
JonathanGiles: Same here.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -207,11 +207,68 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
+                if (!CoreUtils.isNullOrEmpty(throwableMessage)) {
+                    message += System.lineSeparator() + throwableMessage;
+                }
+                String errorMessage = String.format(LOG_MESSAGE_FORMAT, LOG_LEVEL_ERROR, dateTime, classPath, message);
+                System.out.println(errorMessage);
+                break;
+            default:
+                // Don't do anything, this state shouldn't be possible.
+                break;
+        }
```
JonathanGiles: There is a lot of repetition in here - it looks like you could build the string and do the println outside of the switch statement.

sdk/core/azure-core/src/main/java/module-info.java
JonathanGiles: This is no longer required?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -36,6 +39,13 @@
  */
 public class ClientLogger {
     private final Logger logger;
+    private final String className;
+
+    private static final String LOG_MESSAGE_FORMAT = "%s: %s %s%n%s";
```
alzimmermsft: How did we decide on this log message format? Additionally, should we just have a format for logging stack traces as well instead of manually handling adding a newline and stack trace?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
alzimmermsft: Isn't this just getting the canonical name for `String`?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -207,11 +207,68 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
+            if (rootLogLevel.getLogLevel() > LogLevel.VERBOSE.getLogLevel()) {
                 args = removeThrowable(args);
             }
         }
+        boolean isFromEnv = isLogFromEnv();
+        printInPlace(isFromEnv, throwableMessage, logLevel, format, args);
+
+    }
+
+    private void printInPlace(boolean isFromEnv, String throwableMessage, LogLevel logLevel, String format,
```
alzimmermsft: Could we call this log instead of print?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -207,11 +207,68 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
+        Object... args) {
+        if (isFromEnv) {
+            printFromDefaultLogger(throwableMessage, logLevel, format, args);
+        } else {
+            printFromLogger(throwableMessage, logLevel, format, args);
+        }
+    }
+
+    private void printFromDefaultLogger(String throwableMessage, LogLevel logLevel, String format, Object... args) {
```
alzimmermsft: `logWithDefaultLogger`?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -207,11 +207,68 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
+                String errorMessage = String.format(LOG_MESSAGE_FORMAT, LOG_LEVEL_ERROR, dateTime, classPath, message);
+                System.out.println(errorMessage);
+                break;
+            default:
+                // Don't do anything, this state shouldn't be possible.
+                break;
+        }
+    }
+
+    private boolean isLogFromEnv() {
```
alzimmermsft: Could this be inverted and call `hasSlf4jBinding`? Something more declarative.

sdk/core/azure-core/src/main/java/module-info.java
alzimmermsft: @JonathanGiles

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: It is not needed. Forgot to delete. 
Removed.

sdk/core/azure-core/pom.xml
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Put it in private field and initialize when we have clientlogger instance.

sdk/core/azure-core/src/main/java/module-info.java
JonathanGiles: I commented earlier that this should no longer be necessary given it isn't being used any longer.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Move out some common stuff to printInPlace method.
Exact system out in end of the printFromDefault

sdk/core/azure-core/src/main/java/module-info.java
sima-zhu: Removed.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: className.getClass will return the Class type. 
E.g. 
```
new ClientLogger(A.class);
// What we get from getCanoniacalName is com.azure.classPath.A (String)
```
I have tested in other class. It is working like this.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Since it exists since logger is initialized, I put the check in constructor and have a variable to remember the boolean.

sdk/core/azure-core/src/main/java/module-info.java
sima-zhu: Yes. Removed.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Need to have toString in LogLevel.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Working on stack trace format.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Seems like LogLevel.INFO,name() is working.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
alzimmermsft: That is surprising since `className` is a `String` and from the documentation I've read `"a string".getClass()` should return `String`.

PR #7276, opened 2020-01-08
Storage/feature branch merge with master

gapra-msft:


weshaggard:
@gapra-msft It looks like you did a squash merge into this feature branch which is likely not what you wanted. When moving changes from other branches you generally don't want to merge with github UI but instead do it from the command line. You can us the github UI for PR validation if you like but the actually merge should be pushed from the command line because you generally need to preserve the source history instead of squashing.

PR #7281, opened 2020-01-08
Build system updates to prep for automatic version bump and pre-release verifications

JimSuplizio:
This change is the precursor for getting the automatic version bump and pre-release verification (verify versions are in the accepted format and that we don't try and release anything with "unreleased" dependencies).

The ci.yml file updates are as follows:
- Addition of the SDKType - Right now we're in a state where we need to know if we're building client or data for some of the steps in the client build and, unfortunately, these are only set on the pipelines for those areas that have both client/data building from the same service pom.
- Addition of the groupId to the Artifacts lists - We're manipulating versions of various libraries and, unfortunately, the artifactId isn't actually enough to uniquely identify a library. There's at least one case where there's a similarly named artifactId between client and data which only becomes worse when we look at management track.

The set_versions.py script updates are as follows:
- Do not require an update type - This script was created to manipulate library versions, not external dependencies.
- Plumbed through the new groupId argument
- Reworked the argparse logic to better display which are required and optional. Honestly, the only required argument is the build-type which maps directly to the sdktype. Similarly, added verification to each of the individual commands to verify their individual arguments.
- Added some throws so we'll error our correctly, specifically in the cases where the library being manipulated/verified isn't found.

The archetype-sdk-client.yml updates are as follows:
- Modify the commands that call the set_versions.py script to reflect the updates made above
- Use the SDKType from the parameters in the conditions for Start Jetty, test from source setup

### DIFF
--- a/eng/pipelines/templates/stages/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-client.yml
@@ -8,6 +8,7 @@ stages:
     - template: ../jobs/archetype-sdk-client.yml
       parameters:
         ServiceDirectory: ${{parameters.ServiceDirectory}}
+        SDKType: ${{parameters.SDKType}}
         Artifacts: ${{parameters.Artifacts}}
 
   # The Prerelease and Release stages are conditioned on whether we are building a pull request and the branch.


mitchdenny:
This looks good to me. I think it makes sense to have ```groupId``` on each artifact to avoid complex coalescing object all over the place anywhere we want to use it. Its obvious and in your face.

PR #7292, opened 2020-01-08
Revised All Track 2 README files

mssfang:
- Simplify the `Default HTTP Client` section
- Removed `Alternate HTTP client` and `Configuring HTTP Clients` sections.
- Moved updated `Default HTTP Client` and `Default SSL library` sections to `Troubleshooting` section.

### DIFF
--- a/sdk/storage/azure-storage-file-datalake/README.md
+++ b/sdk/storage/azure-storage-file-datalake/README.md
@@ -120,10 +120,10 @@ In the past, cloud-based analytics had to compromise in areas of performance, ma
 
 Data Lake Storage Gen2 offers two types of resources:
 
-- The _filesystem used via 'DataLakeFileSystemClient'
-- The _path used via 'DataLakeFileClient' or 'DataLakeDirectoryClient'
+- The `_filesystem` used via 'DataLakeFileSystemClient'
+- The `_path` used via 'DataLakeFileClient' or 'DataLakeDirectoryClient'
 
-|ADLS Gen2                     | Blob       |
+|ADLS Gen2                  | Blob       |
 | --------------------------| ---------- |
 |Filesystem                 | Container  | 
 |Path (File or Directory)   | Blob       |


sdk/appconfiguration/azure-data-appconfiguration/README.md
```diff
@@ -28,56 +28,6 @@ Use the client library for App Configuration to create and manage application co
...
-<dependency>
-  <groupId>com.azure</groupId>
-  <artifactId>azure-core-http-okhttp</artifactId>
-  <version>1.1.0</version>
-</dependency>
-```
-[//]: # ({x-version-update-end})
-
-### Configuring HTTP Clients
-When an HTTP client is included on the classpath, as shown above, it is not necessary to specify it in the client library [builders](#create-client), unless you want to customize the HTTP client in some fashion. If this is desired, the `httpClient` builder method is often available to achieve just this, by allowing users to provide a custom (or customized) `com.azure.core.http.HttpClient` instances.
```
srnagar: Also, cleanup `#create-client` link at bottom of the README if it's no longer used. Same for other links that were deleted.

sdk/storage/azure-storage-file-datalake/README.md
```diff
@@ -375,6 +325,17 @@ When interacting with data lake using this Java client library, errors returned
 status codes returned for [REST API][error_codes] requests. For example, if you try to retrieve a file system or path that
 doesn't exist in your Storage Account, a `404` error is returned, indicating `Not Found`.
 
+### Default HTTP Client
```
srnagar: Looks like this has italicized text starting from line 123 `- The _filesystem used via 'DataLakeFileSystemClient'`. There's an underscore before filesystem that's causing this.

sdk/appconfiguration/azure-data-appconfiguration/README.md
mssfang: the link is no longer exists in the first place. Should remove it.

mssfang: We can keep create an azconfig client section for other use.

sdk/storage/azure-storage-file-datalake/README.md
mssfang: updated

PR #7296, opened 2020-01-09
Fix Failing Tests

alzimmermsft:
Fixes tests that were failing when running against the Storage service.

- A few tests were previously relying on a hard-coded default of `4MB` which was recently changed to be a user controlled value which now defaults to `256MB`, these tests are now passing `4MB` for the user controlled value.
- The upload from file code path was altered to eagerly validate `ParallelTransferOptions` which resulted in it failing a check whether HTBB transfer could be applied. Logic was changed to maintain the original state long enough to make this validation to use HTBB when able.
- One test could result in Reactor returning different exceptions, a composite exception can be thrown from a `flatMap` is multiple streams hit an exception instead of returning a single exception. Testing logic was updated to expand the composite exception and do deep validation.


alzimmermsft:
@rickle-msft Mind giving this PR a review? It should resolve a few testing issues and ensures that HTBB upload is being properly applied in the upload from file case.

PR #7298, opened 2020-01-09
Implements a slf4j logger impl as default logger in azure core

sima-zhu:


sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,147 @@
...
+     * DEBUG according to the format outlined above.
+     */
+    public void error(String format, Object... argArray) {
+        formatAndLog(LogLevel.ERROR, format, argArray);
+    }
+
+    /**
+     * For formatted messages, first substitute arguments and then log.
+     *
+     * @param level
```
mssfang: missing description for param `level` and `arguments`

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.implementation.logging;
+
+import com.azure.core.util.logging.LogLevel;
+import java.io.PrintStream;
+import java.text.SimpleDateFormat;
+import java.util.Date;
```
mssfang: At lease do not using 'java.util.Date'.  https://programminghints.com/2017/05/still-using-java-util-date-dont/

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,147 @@
...
+import org.slf4j.helpers.FormattingTuple;
+import org.slf4j.helpers.MessageFormatter;
+
+/**
+ * This class is an internal implementation of slf4j logger.
+ */
+public class DefaultLogger extends DefaultLoggerBase {
```
alzimmermsft: Should this be final?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -35,7 +36,8 @@
  * @see Configuration
  */
 public class ClientLogger {
-    private final Logger logger;
+    private Logger logger;
```
alzimmermsft: Why is this no longer final? Instead load the SLF4J binded instance into a temp variable then set this based on that being a NOPLogger.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Construct DefaultLogger for the given class name.
+     *
+     * @param className Class name creating the logger.
+     * @throws RuntimeException it is an error.
+     */
+    public DefaultLogger(String className) {
+        try {
+            this.classPath = Class.forName(className).getCanonicalName();
```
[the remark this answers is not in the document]
sima-zhu: Switch to
```
        try {
            this.classPath = Class.forName(className).getCanonicalName();
        } catch (ClassNotFoundException e) {
            this.classPath = className;
        }
```

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private String getFormattedDate() {
+        Date now = new Date();
+        return DATE_FORMAT.format(now);
+    }
+
+    void write(StringBuilder buf, Throwable t) {
+        PrintStream targetStream = System.out;
+
+        targetStream.println(buf.toString());
```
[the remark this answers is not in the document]
sima-zhu: https://stackoverflow.com/questions/9459657/is-multi-thread-output-from-system-out-println-interleaved

Make the print process atomic so the log message will not intervene by other message, or print throwable to other places.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
sima-zhu: Made it final

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Revert back to final

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLoggerBase.java
```diff
@@ -0,0 +1,841 @@
...
+public abstract class DefaultLoggerBase implements Logger {
+
+    /**
+     * Return the name of this <code>Logger</code> instance.
+     *
+     * @return name of this logger instance
+     */
+    @Override
+    public String getName() {
+        return null;
```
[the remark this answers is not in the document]
sima-zhu: ClientLogger passes the class name to DefaultLogger. Is it not that useful at DefaultLoggerBase layer, since we need to add private field and pass the class name to base class. I prefer to leave it alone.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLoggerBase.java
```diff
@@ -0,0 +1,841 @@
...
+
+    /**
+     * Is the logger instance enabled for the DEBUG level?
+     *
+     * @return True if this Logger is enabled for the DEBUG level,
+     * false otherwise.
+     */
+    @Override
+    public boolean isDebugEnabled() {
+        return false;
```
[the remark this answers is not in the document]
sima-zhu: * Since default logger base is package private abstract class, we always need to have it in DefaultLogger. However API is not that useful for current structure. 
* It is not always the case to read env. We might need to introduce reading from file and etc in future. So it better to have it in subclass instead of base class.
I think to leave it as it is. And we can easily implement this if there is a need in future.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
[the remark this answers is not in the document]
alzimmermsft: The individual print statement shouldn't interleave but right now we print the message and stack trace in two separate calls which could result in the following scenario.

Two exceptions are being logged, problemA and problemB, logging could have the following outcome

- problemA message
- problemB message
- problemA stack trace
- problemB stack trace

Instead we should build the message + stack trace and log them in one print call.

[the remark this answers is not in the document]
sima-zhu:
```
    void write(StringBuilder buf, Throwable t) {
        if (t != null) {
            StringWriter sw = new StringWriter();
            PrintWriter pw = new PrintWriter(sw);
            t.printStackTrace(pw);
            String sStackTrace = sw.toString();
            buf.append(sStackTrace);
        }
        System.out.println(buf.toString());
    }

```

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,559 @@
...
+
+    /**
+     * Log a message at the INFO level. It is currently not supported.
+     *
+     * @param msg the message string to be logged
+     * @throws UnsupportedOperationException It is currently not supported.
+     */
+    @Override
+    public void info(final String msg) {
+        throw new UnsupportedOperationException();
```
srnagar: These should be implemented too.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,559 @@
...
+
+    private String getFormattedDate() {
+        LocalDateTime now = LocalDateTime.now();
+        return DATE_FORMAT.format(now);
+    }
+
+    void write(StringBuilder buf, Throwable t) {
+        if (t != null) {
+            StringWriter sw = new StringWriter();
+            PrintWriter pw = new PrintWriter(sw);
```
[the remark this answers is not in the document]
sima-zhu: StringWriter does not need to close as it is a no op. 
https://stackoverflow.com/questions/14542535/will-not-closing-a-stringwriter-cause-a-leak

Closed the PrintStream

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLoggerBase.java
```diff
@@ -0,0 +1,886 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.implementation.logging;
+
+import org.slf4j.Logger;
+import org.slf4j.Marker;
+
+abstract class DefaultLoggerBase implements Logger {
```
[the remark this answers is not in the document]
sima-zhu: Removed.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,559 @@
...
+        FormattingTuple tp = MessageFormatter.arrayFormat(format, arguments);
+        log(level, tp.getMessage(), tp.getThrowable());
+    }
+
+    /**
+     * This is our internal implementation for logging regular
+     * (non-parameterized) log messages.
+     *
+     * @param level
+     *            One of the LOG_LEVEL_XXX constants defining the log level
```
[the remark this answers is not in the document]
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,367 @@
...
+     * @param buf Take the log messages.
+     * @param t The exception whose stack trace should be logged
+     */
+    void writeWithThrowable(StringBuilder buf, Throwable t) {
+        if (t != null) {
+            StringWriter sw = new StringWriter();
+            PrintWriter pw = new PrintWriter(sw);
+            t.printStackTrace(pw);
+            buf.append(sw.toString());
+            pw.close();
```
srnagar: Use `try-with-resources` to ensure the resource is always closed.

sima-zhu: Since StringWriter has no-op close(), I only move pw to try with resource. Thanks for pointing this out!

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,559 @@
...
+
+    /**
+     * Return the name of this <code>Logger</code> instance. It is currently not supported.
+     *
+     * @return name of this logger instance
+     * @throws UnsupportedOperationException It is currently not supported.
+     */
+    @Override
+    public String getName() {
+        throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
sima-zhu: Talked offline. Will create an issue and put Up-for-grabs label.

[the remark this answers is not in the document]
sima-zhu: github.com/Azure/azure-sdk-for-java/issues/7341

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
sima-zhu: Same as above

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,559 @@
...
+
+    /**
+     * Log a message at the DEBUG level. It is currently not supported.
+     *
+     * @param msg the message string to be logged
+     * @throws UnsupportedOperationException It is currently not supported.
+     */
+    @Override
+    public void debug(final String msg) {
+        throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
sima-zhu: I enabled the ones used by ClientLogger.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,366 @@
...
+    public boolean isErrorEnabled() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void error(final String msg) {
+        throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
sima-zhu:
```                
logger.error(format, args);
```
This is the one we used.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -235,39 +229,47 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
-                return logger.isErrorEnabled();
-            default:
-                return false;
+    /**
+     * Checking the system log with the preference order of slf4j and environment variable.
+     *
+     * @return The log level.
+     */
+    private LogLevel getConfiguredLogLevel() {
+        if (isFromEnv) {
```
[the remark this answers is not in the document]
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -131,11 +143,7 @@ public void error(String format, Object... args) {
      * @param args Arguments for the message, if an exception is being logged last argument is the throwable.
      */
     private void log(LogLevel logLevel, String format, Object... args) {
-        LogLevel environmentLoggingLevel = getEnvironmentLoggingLevel();
-
-        if (canLogAtLevel(logLevel, environmentLoggingLevel)) {
-            performLogging(logLevel, environmentLoggingLevel, false, format, args);
-        }
+        performLogging(logLevel, false, format, args);
```
alzimmermsft: Could this method be removed and all the higher level methods just call into `performLogging`? This previously existed to abstract a few pieces of logic away from the public API methods which no longer exists based on the changes made in this PR.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -36,6 +35,7 @@
  */
 public class ClientLogger {
     private final Logger logger;
+    private final boolean isFromEnv;
```
alzimmermsft: Is this being used by anything other than the instantiation in the constructor? If not let's remove it.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,376 @@
...
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isDebugEnabled() {
+        String logLevelStr = Configuration.getGlobalConfiguration().get(AZURE_LOG_LEVEL);
+        LogLevel currentLogLevel = LogLevel.fromString(logLevelStr);
```
alzimmermsft: Should this logic be abstracted into `LoggingUtils` or as a private static method on this class?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,376 @@
...
+     *
+     * @param level log level
+     * @param message The message itself
+     * @param t The exception whose stack trace should be logged
+     */
+    private void log(LogLevel level, String message, Throwable t) {
+        String dateTime = getFormattedDate();
+        String threadName = Thread.currentThread().getName();
+        String levelName = level.name();
+        StringBuilder buf = new StringBuilder(32);
```
alzimmermsft: Any reason to instantiate the `StringBuilder` with an internal capacity of 32? Instead could it be instantiated with a capacity large enough to hold the initial string being written to it?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,376 @@
...
+        buf.append(String.format(MESSAGE_TEMPLATE, dateTime, threadName, levelName, classPath, message));
+        writeWithThrowable(buf, t);
```
alzimmermsft: If there is a `Throwable` to write is there a newline after the message template? Looking at the message template there is no newline inserted between the general log message and the stack trace.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,376 @@
...
+        return DATE_FORMAT.format(now);
+    }
+
+    /**
+     * Write the log message with throwable stack trace if any.
+     *
+     * @param buf Take the log messages.
+     * @param t The exception whose stack trace should be logged
+     */
+    void writeWithThrowable(StringBuilder buf, Throwable t) {
```
alzimmermsft: Could this be made private? Additionally, could we change this method to simple add the stack trace to the log message. Adding the `Throwable` stack trace and writing the message to the console don't seem like ideas that should be merged eagerly.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,376 @@
...
+        if (t != null) {
+            StringWriter sw = new StringWriter();
+            try (PrintWriter pw = new PrintWriter(sw)) {
+                t.printStackTrace(pw);
+                buf.append(sw.toString());
+            }
```
alzimmermsft: One of the SDK guidelines it to only print an exception's stack trace when logging is set to verbose. In `ClientLogger` it was set to only log the exception message at a different level.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -235,27 +236,16 @@ private void performLogging(LogLevel logLevel, LogLevel environmentLogLevel, boo
...
+    /**
+     * Determines if the app or environment logger support logging at the given log level.
      *
      * @param logLevel Logging level for the log message.
-     * @param environmentLoggingLevel Logging level the environment is set to support.
      * @return Flag indicating if the environment and logger are configured to support logging at the given log level.
      */
-    private boolean canLogAtLevel(LogLevel logLevel, LogLevel environmentLoggingLevel) {
-        // Do not log if logLevel is null is not set.
-        if (logLevel == null) {
```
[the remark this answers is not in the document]
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,376 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void warn(String format, Object arg) {
+        logFromFormat(LogLevel.WARNING, format, arg);
+    }
+
+    /**
+     * {@inheritDoc}
+     * @param format The formattable message to log.
```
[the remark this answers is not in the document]
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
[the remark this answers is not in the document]
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isDebugEnabled() {
+        String logLevelStr = Configuration.getGlobalConfiguration().get(AZURE_LOG_LEVEL);
+        LogLevel currentLogLevel = LogLevel.fromString(logLevelStr);
+        return LogLevel.VERBOSE.getLogLevel() >= currentLogLevel.getLogLevel();
```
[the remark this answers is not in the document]
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Done.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
sima-zhu: Have a private helper method in class. Env log level is only needed for default logger, so it is ok to leave it here for current use.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
sima-zhu: It is inherited from simple logger. 
The message we know definitely will be printed out is:
2020-01-09 12:35 [main] [Informational]

This is 39 character long. I can put 64 instead of 32 characters to minimize capacity increase.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
sima-zhu: Sure.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
sima-zhu: Talked offline.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
sima-zhu: Added %n in template.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
sima-zhu: Talked offline.

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -184,8 +185,10 @@
   <!-- CodeSnippet Suppression for now, which need code owner's attention -->
   <suppress checks="com.azure.tools.checkstyle.checks.JavadocCodeSnippetCheck" files="com.azure.data.appconfiguration.ConfigurationAsyncClient.java"/>
 
-  <!-- ClientLogger class suppression -->
+  <!-- Logger class suppression -->
   <suppress checks="com.azure.tools.checkstyle.checks.GoodLoggingCheck" files="ClientLogger.java"/>
+  <suppress checks="com.azure.tools.checkstyle.checks.GoodLoggingCheck" files="DefaultLogger.java"/>
+  <suppress checks="com.azure.tools.checkstyle.checks.GoodLoggingCheck" files="DefaultLoggerBase.java"/>
```
srnagar: This line can now be removed as `DefaultLoggerBase` does not exist.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,368 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void info(final String msg, final Throwable t) {
+        throw new UnsupportedOperationException();
+    }
+
+
```
mssfang: nit: extra two lines

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,365 @@
...
+    @Override
+    public void trace(final String format, final Object... arguments) {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void trace(final String msg, final Throwable t) {
```
JonathanGiles: Why overload all of these methods to throw an exception when the base class has an actual useful implementation? Look at https://github.com/qos-ch/slf4j/blob/master/slf4j-api/src/main/java/org/slf4j/helpers/MarkerIgnoringBase.java

sima-zhu: Done

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -0,0 +1,362 @@
...
+     */
+    public DefaultLogger(Class<?> clazz) {
+        this(clazz.getName());
+    }
+
+    /**
+     * Construct DefaultLogger for the given class name.
+     *
+     * @param className Class name creating the logger.
+     * @throws RuntimeException it is an error.
```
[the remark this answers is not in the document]
sima-zhu: Not logger throw an exception. Good catch. Will update the JavaDoc

PR #7306, opened 2020-01-09
Retry transient send errors that occur on Event Hubs Producer

conniey:
When an AMQP link dies, the send operation for a message is retried.


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -44,6 +46,7 @@
 import java.util.function.Supplier;
 import java.util.stream.Collector;
 
+import static com.azure.core.amqp.implementation.RetryUtil.*;
```
alzimmermsft: Wildcard import

PR #7312, opened 2020-01-09
Added version checking for new blob and file APIs

gapra-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobServiceVersionPolicy.java
```diff
@@ -0,0 +1,54 @@
+package com.azure.storage.blob.implementation.util;
```
alzimmermsft: Missing license

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobServiceVersionPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+import com.azure.core.http.HttpPipelineNextPolicy;
+import com.azure.core.http.HttpResponse;
+import com.azure.core.http.policy.HttpPipelinePolicy;
+import com.azure.storage.blob.BlobServiceVersion;
+import com.azure.storage.common.implementation.Constants;
+import reactor.core.publisher.Mono;
+
+import static com.azure.storage.common.implementation.StorageImplUtils.throwIfContainsHeader;
+
+public class BlobServiceVersionPolicy implements HttpPipelinePolicy {
```
alzimmermsft: Mind adding an explanation of the class here

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobServiceVersionPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        HttpHeaders requestHeaders = context.getHttpRequest().getHeaders();
+        BlobServiceVersion serviceVersion = toServiceVersion(
+            requestHeaders.getValue(Constants.HeaderConstants.SERVICE_VERSION));
+        String ANY_BLOB_API = "any blob API";
```
alzimmermsft: Let's make this a `static final` at the class level, no need to instantiate this each time the policy processes.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobServiceVersionPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+
+        if (serviceVersion.ordinal() == BlobServiceVersion.getLatest().ordinal()) {
+            return next.process();
+        }
+
+        // 2019_07_07 check
+        if (serviceVersion.ordinal() < BlobServiceVersion.V2019_07_07.ordinal()) {
+            // Encryption scope
+            String CREATE_CONTAINER = "create container";
```
alzimmermsft: Same comment as `ANY_BLOB_API`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobServiceVersionPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+        return next.process();
+    }
+
+    private static BlobServiceVersion toServiceVersion(String version) {
+        if (version.equals(BlobServiceVersion.V2019_02_02.getVersion())) {
+            return BlobServiceVersion.V2019_02_02;
+        } else if (version.equals(BlobServiceVersion.V2019_07_07.getVersion())) {
+            return BlobServiceVersion.V2019_07_07;
+        } else {
+            throw new NullPointerException(Constants.HeaderConstants.SERVICE_VERSION + " must be set.");
```
alzimmermsft: This should throw an `IllegalStateException` as we don't expect to get here, also if `version` is `null` a NPE would be thrown by the conditional clauses in this method.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobServiceVersionPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+            // Managed disk range diff
+            throwIfContainsHeader(requestHeaders, "x-ms-previous-snapshot-url", ANY_BLOB_API,
+                serviceVersion.getVersion());
+        }
+
+        return next.process();
+    }
+
+    private static BlobServiceVersion toServiceVersion(String version) {
+        if (version.equals(BlobServiceVersion.V2019_02_02.getVersion())) {
```
alzimmermsft: Should check that `BlobServiceVersion.getVersion` equals `version`.

```java
BlobServiceVersion.V2019_02-02.getVersions().equals(version)
```

This is better since the enum has to be set but `version` may not be set (though I think it always will be).

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobServiceVersionPolicyTest.groovy
```diff
@@ -0,0 +1,118 @@
+package com.azure.storage.blob
```
alzimmermsft: Missing license

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/StorageImplUtils.java
```diff
@@ -3,6 +3,8 @@
 
 package com.azure.storage.common.implementation;
 
+import com.azure.core.http.HttpHeaders;
+import com.azure.core.http.HttpRequest;
```
alzimmermsft: Did something pull this dependency in?

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/implementation/util/ShareServiceVersionPolicy.java
```diff
@@ -0,0 +1,75 @@
+package com.azure.storage.file.share.implementation.util;
```
alzimmermsft: Missing license

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/implementation/util/ShareServiceVersionPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+
+public class ShareServiceVersionPolicy implements HttpPipelinePolicy {
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        HttpHeaders requestHeaders = context.getHttpRequest().getHeaders();
+        URL requestUrl = context.getHttpRequest().getUrl();
+        ShareServiceVersion serviceVersion = toServiceVersion(
+            requestHeaders.getValue(Constants.HeaderConstants.SERVICE_VERSION));
+        String ANY_FILE_API = "any file API";
```
alzimmermsft: Similar comments in this file as the Blobs one.

- Make constant strings `static final` at the class level.
- Reverse the version comparison to use the enum as the `.equals` caller since it is known to be set.
- Change the `NullPointerException` to an `IllegalStateException`.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/implementation/util/ShareServiceVersionPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        // 2019_07_07 check
+        if (serviceVersion.ordinal() < ShareServiceVersion.V2019_07_07.ordinal()) {
+            // File lease
+            throwIfContainsHeader(requestHeaders, "x-ms-lease-id", ANY_FILE_API, serviceVersion.getVersion());
+            throwIfContainsHeader(requestHeaders, "x-ms-lease-duration", ANY_FILE_API, serviceVersion.getVersion());
+            throwIfContainsHeader(requestHeaders, "x-ms-proposed-lease-id", ANY_FILE_API, serviceVersion.getVersion());
+
+            if (requestUrl.getQuery() != null && requestUrl.getQuery().contains("comp=lease")) {
```
alzimmermsft: Should a helper method be added into `StorageImplUtils` that validates that the request doesn't contain a given query parameter?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/StorageImplUtils.java
gapra-msft: Just an artefact of me trying something then undoing it

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/implementation/util/ShareServiceVersionPolicy.java
gapra-msft: yeah that's a good idea.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobServiceVersionPolicyTest.groovy
```diff
@@ -0,0 +1,121 @@
...
+            .setDenyEncryptionScopeOverride(true)
+        BlobContainerClient cpkncesContainer = containerBuilder.blobContainerEncryptionScope(ces)
+            .containerName(generateContainerName()).buildClient()
+
+        when:
+        cpkncesContainer.createWithResponse(null, null, null, null)
+
+        then:
+        def e = thrown(IllegalArgumentException)
+        e.getMessage() == "x-ms-deny-encryption-scope-override is not supported for create container in service version 2019-02-02"
```
rickle-msft: I think testing against the literal value of the message is brittle; checking that an IllegalArgumentException is thrown is probably sufficient. 

Also, perhaps IllegalStateException is more appropriate? I think of IlegalArgumentExceptions as resulting from parameter validation on method call time.

gapra-msft: I'll change it to IllegalStateException. That makes sense.

gapra-msft: Though I think it might be best to keep the messages for now - in case we need to change the test later we can. I dont want the wrong message to be called either.

PR #7313, opened 2020-01-09
Storage ARM Template for live tests

hemanttanwar:
Storage will have one common arm template for Queue, File and Blob.
The pipeline will pass on names of these resources as ENV variable into test. For example PRIMARY_STORAGE_ACCOUNT_NAME

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
@@ -118,7 +118,7 @@ jobs:
           PROVISIONER_APPLICATION_ID: $(provisioner-aad-id)
           PROVISIONER_APPLICATION_SECRET: $(provisioner-aad-secret)
         displayName: Remove Test Resources
-        condition: false
+        condition: true
 
 
       - task: PublishTestResults@2


sdk/storage/tests.yml
```diff
@@ -3,30 +3,36 @@ trigger: none
 jobs:
   - template: ../../eng/pipelines/templates/jobs/archetype-sdk-tests.yml
     parameters:
+      Matrix:
+        Windows - Java 8:
+            OSName: 'Windows'
+            OSVmImage: 'windows-2019'
+            JavaVersion: '1.8'
+            DisplayName: 'Windows - Java 8'
```
alzimmermsft: Remove this as it will scope the Storage tests to a smaller matrix

sdk/storage/tests.yml
```diff
@@ -3,30 +3,36 @@ trigger: none
...
+        AZURE_STORAGE_QUEUE_ENDPOINT: $(AZURE_STORAGE_QUEUE_ENDPOINT)
+        AZURE_STORAGE_QUEUE_CONNECTION_STRING: $(AZURE_STORAGE_QUEUE_CONNECTION_STRING)
+        AZURE_STORAGE_BLOB_CONNECTION_STRING: $(AZURE_STORAGE_BLOB_CONNECTION_STRING)
+        STORAGE_DATA_LAKE_ACCOUNT_NAME: $(STORAGE_DATA_LAKE_ACCOUNT_NAME)
+        STORAGE_DATA_LAKE_ACCOUNT_KEY: $(STORAGE_DATA_LAKE_ACCOUNT_KEY)
+        AZURE_STORAGE_FILE_ACCOUNT_NAME: $(AZURE_STORAGE_FILE_ACCOUNT_NAME)
+        AZURE_STORAGE_FILE_ACCOUNT_KEY: $(AZURE_STORAGE_FILE_ACCOUNT_KEY)
+        AZURE_STORAGE_QUEUE_ACCOUNT_NAME: $(AZURE_STORAGE_FILE_ACCOUNT_NAME)
+        AZURE_STORAGE_QUEUE_ACCOUNT_KEY: $(AZURE_STORAGE_QUEUE_ACCOUNT_KEY)
+        TimeoutInMinutes: 120
```
alzimmermsft: Remove this line, its adding an environment variable of `TimeoutInMinutes` which isn't needed

sdk/storage/tests.yml
```diff
@@ -3,30 +3,36 @@ trigger: none
...
   - template: ../../eng/pipelines/templates/jobs/archetype-sdk-tests.yml
     parameters:
+      Matrix:
+        Windows - Java 8:
+            OSName: 'Windows'
+            OSVmImage: 'windows-2019'
+            JavaVersion: '1.8'
+            DisplayName: 'Windows - Java 8'
       ServiceDirectory: storage
+      TimeoutInMinutes: 240
```
alzimmermsft: Let's use a 2 hour timeout since the complete test suite only takes about 60-75 minutes, anything longer and there is likely an issue in our tests.

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -69,6 +102,25 @@ jobs:
...
+          -ProvisionerApplicationId $env:PROVISIONER_APPLICATION_ID
+          -ProvisionerApplicationSecret $env:PROVISIONER_APPLICATION_SECRET
+          -Force
+          -Verbose
+        env:
+          TENANT_ID: $(aad-azure-sdk-test-tenant-id)
+          PROVISIONER_APPLICATION_ID: $(provisioner-aad-id)
+          PROVISIONER_APPLICATION_SECRET: $(provisioner-aad-secret)
+        displayName: Remove Test Resources
+        condition: true
```
mikeharder: @danieljurek: Should this be `condition: always()` instead?  Or are these equivalent?  I assume the intent is to ensure cleanup is always run regardless of earlier failures.

mitchdenny:
Looks like this PR might have broken the Cosmos build

PR #7316, opened 2020-01-09
read point operation for non item resource avoid duplicate json serializat…

moderakh:
similar to https://github.com/Azure/azure-sdk-for-java/pull/7293 avoiding duplicate json serialization and deserialization for read point operation non document resources


kirankumarkolli:
NOTE: Porting it to V3 straight might be a breaking change.

kushagraThapar:
> NOTE: Porting it to V3 straight might be a breaking change.

It shouldn't be a breaking change, as there are no public APIs touched in this PR. All the constructors are package private scoped.

PR #7336, opened 2020-01-10
Continue to use same .receive() method on disconnection

### LINKED ISSUE #7273: On transient connection failures, original `.receive()` call stops.
When a transient failure occurs in the connection, the connection is revived, but the original `.receive(String partitionId, EventPosition position)` stops pushing out events. Consumers need to perform another `.receive()`.

conniey:
* Adds AmqpReceiveLinkProcessor to fetch new links when one dies.
* Fixes logging for ActiveToken provider to give the resource rather than a full scope

Fixes #7273

### DIFF
--- a/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RetryUtil.java
+++ b/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RetryUtil.java
@@ -81,12 +81,12 @@ public class RetryUtil {
                 }
 
                 if (error instanceof TimeoutException) {
-                    LOGGER.info("TimeoutException error occurred. Retrying operation. Attempt: {}. Error: {}",
+                    LOGGER.info("TimeoutException error occurred. Retrying operation. Attempt: {}.",
                         attempt, error);
 
                     return retryPolicy.calculateRetryDelay(error, attempt);
                 } else if (error instanceof AmqpException && (((AmqpException) error).isTransient())) {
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubPartitionAsyncConsumer.java
```diff
@@ -3,127 +3,68 @@
...
         this.consumerGroup = consumerGroup;
         this.partitionId = partitionId;
-        this.emitterProcessor = EmitterProcessor.create(prefetchCount, false);
+        this.emitterProcessor = amqpReceiveLinkProcessor
+            .map(message -> onMessageReceived(message))
+            .doOnNext(event -> {
+                // Keep track of the last position so if the link goes down, we don't start from the original location.
+                final Long offset = event.getData().getOffset();
+                if (offset != null) {
+                    currentEventPosition.set(EventPosition.fromOffset(offset));
```
srnagar: Would it be more efficient if we just kept track of offset (long value) and convert to EventPosition only when the link goes down to restore the link? It prevents creating a lot of EventPosition objects.

conniey: Makes sense. I'll update it.

Issue #5967, opened 2019-10-21
Azure.Core & Azure.Identity needs to be updated to support env vars


joshfree:
The environment variables identified in https://github.com/Azure/azure-sdk/blob/master/_includes/tables/environment_variables.md such as

| Environment Variable          | Purpose                                |
|-------------------------------|----------------------------------------|
| **Proxy Settings**                                                    ||
| HTTP_PROXY                    | Proxy for HTTP connections             |
| HTTPS_PROXY                   | Proxy for HTTPS connections            |
| NO_PROXY                      | Hosts which must not use a proxy       |
| **Identity**                                                          ||
| MSI_ENDPOINT                  | Azure AD MSI Credentials               |
| MSI_SECRET                    | Azure AD MSI Credentials               |
| AZURE_SUBSCRIPTION_ID         | Azure subscription                     |
| AZURE_USERNAME                | Azure username for U/P Auth            |
| AZURE_PASSWORD                | Azure password for U/P Auth            |
| AZURE_CLIENT_CERTIFICATE_PATH | Azure Active Directory                 |
| AZURE_CLIENT_ID               | Azure Active Directory                 |
| AZURE_CLIENT_SECRET           | Azure Active Directory                 |
| AZURE_TENANT_ID               | Azure Active Directory                 |
| AZURE_RESOURCE_GROUP          | Azure Resource Group                   |
| AZURE_CLOUD                   | Name of the sovereign cloud            |
| **Pipeline Configuration**                                            ||
| AZURE_TELEMETRY_DISABLED      | Disables telemetry                     |
| AZURE_LOG_LEVEL               | Enable logging by setting a log level. |
| AZURE_TRACING_DISABLED        | Disables tracing                       |

need to be reviewed, supported, and tested for our Azure.Core and Azure.Identity.  There are at least some of these not currently properly supported, such as AZURE_CLOUD for example.

joshfree:
We will target this work for the December preview

joshfree:
Hi @tzhanl - how is the Identity environment variable support progressing?  Do you need any assistance from @g2vinay or @jianghaolu?  Thanks 😄

tzhanl:
@joshfree we sorted out all the env vars fro all languages. 


Environment   Variable | Purpose | Current Using | Comments
-- | -- | -- | --
Proxy   Settings |   |   |  
[...]

PR #7337, opened 2020-01-10
update EnvironmentCredential and azure.core.util.Configuration

cxznmhdcxz:
Updated EnvironmentCredential to support env vars for UsernamePasswordCredential and ClientCertificateCredential. Added PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH to azure.core.util.Configuration. 
https://github.com/Azure/azure-sdk-for-java/issues/5967

### DIFF
--- a/sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
+++ b/sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
@@ -14,13 +14,27 @@ import com.azure.identity.implementation.IdentityClientOptions;
 import reactor.core.publisher.Mono;
 
 /**
- * A credential provider that provides token credentials based on environment variables.  The environment variables
+ * A credential provider that provides token credentials based on environment variables.  The sets of environment variables
  * expected are:
  * <p>
  * <ul>
  *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_ID AZURE_CLIENT_ID}</li>
  *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_SECRET AZURE_CLIENT_SECRET}</li>
  *     <li>{@link Configuration#PROPERTY_AZURE_TENANT_ID AZURE_TENANT_ID}</li>
+ * </ul>
+ * or:
+ * <p>
+ * <ul>
+ *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_ID AZURE_CLIENT_ID}</li>
+ *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH AZURE_CLIENT_CERTIFICATE_PATH}</li>
+ *     <li>{@link Configuration#PROPERTY_AZURE_TENANT_ID AZURE_TENANT_ID}</li>
+ * </ul>
+ * or:
+ * <p>
+ * <ul>
+ *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_ID AZURE_CLIENT_ID}</li>
+ *     <li>{@link Configuration#PROPERTY_AZURE_USERNAME AZURE_USERNAME}</li>
+ *     <li>{@link Configuration#PROPERTY_AZURE_PASSWORD AZURE_PASSWORD}</li>
  * </ul>
  */
 @Immutable


tzhanl:
@jongio

sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
```diff
@@ -50,6 +50,22 @@
                     configuration.get(Configuration.PROPERTY_AZURE_CLIENT_ID),
                     configuration.get(Configuration.PROPERTY_AZURE_CLIENT_SECRET),
                     identityClientOptions);
+            } else if (configuration.contains(Configuration.PROPERTY_AZURE_CLIENT_ID)
```
jongio: Seems like we should introduce a method something like envContains(params string var) that returns true if the env contains all of the 'var's.

jianghaolu:
Adding @alzimmermsft to review the azure-core change.

sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
```diff
@@ -275,6 +280,21 @@ public boolean contains(String name) {
         return configurations.containsKey(name);
     }
 
+    /**
+     * Determines if multiple configurations all exist.
+     *
+     * @param names Names of the configuration.
+     * @return True if configurations all exist, otherwise false.
+     */
+    public boolean containsAll(String... names){
```
alzimmermsft: I'm not explicitly against this change but I do question how often it will be used outside of the context of `EnvironmentCredential`. Any chance this could be made into a local private method in the interim and if it is needed in additional locations in the future it gets added here.

sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
```diff
@@ -42,14 +42,30 @@
...
+                return new ClientCertificateCredential(configuration.get(Configuration.PROPERTY_AZURE_TENANT_ID),
+                    configuration.get(Configuration.PROPERTY_AZURE_CLIENT_ID),
+                    configuration.get(Configuration.PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH),
+                    null,
+                    identityClientOptions);
+            } else if (configuration.containsAll(Configuration.PROPERTY_AZURE_USERNAME,
+                    Configuration.PROPERTY_AZURE_CLIENT_ID,
+                    Configuration.PROPERTY_AZURE_PASSWORD)) {
+                return new UsernamePasswordCredential(configuration.get(Configuration.PROPERTY_AZURE_CLIENT_ID),
+                    configuration.get(Configuration.PROPERTY_AZURE_TENANT_ID),
```
alzimmermsft: Is this not explicitly needed by `UsernamePasswordCredential`? Otherwise the validation check is missing this.

sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
```diff
@@ -42,14 +42,30 @@
     @Override
     public Mono<AccessToken> getToken(TokenRequestContext request) {
         return Mono.fromSupplier(() -> {
-            if (configuration.contains(Configuration.PROPERTY_AZURE_CLIENT_ID)
-                && configuration.contains(Configuration.PROPERTY_AZURE_CLIENT_SECRET)
-                && configuration.contains(Configuration.PROPERTY_AZURE_TENANT_ID)) {
+            if (configuration.containsAll(Configuration.PROPERTY_AZURE_TENANT_ID,
```
alzimmermsft: Thoughts on changing the `containsAll` method to `getAll`? There are already some race condition chances going on with validating existence then retrieval, if we retrieval all into local instances we could reduce the number of timing issues happening.

sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
```diff
@@ -42,14 +42,30 @@
...
+                    Configuration.PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH)) {
+                return new ClientCertificateCredential(configuration.get(Configuration.PROPERTY_AZURE_TENANT_ID),
+                    configuration.get(Configuration.PROPERTY_AZURE_CLIENT_ID),
+                    configuration.get(Configuration.PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH),
+                    null,
+                    identityClientOptions);
+            } else if (configuration.containsAll(Configuration.PROPERTY_AZURE_USERNAME,
+                    Configuration.PROPERTY_AZURE_CLIENT_ID,
+                    Configuration.PROPERTY_AZURE_PASSWORD)) {
+                return new UsernamePasswordCredential(configuration.get(Configuration.PROPERTY_AZURE_CLIENT_ID),
```
alzimmermsft: Could we retrieve the commonly needed configuration and check those first? Seems like we could make an initial check that `PROPERTY_AZURE_CLIENT_ID` is set and if not just early out.

sdk/identity/azure-identity/src/test/java/com/azure/identity/EnvironmentCredentialTests.java
```diff
@@ -36,4 +36,46 @@ public void testCreateEnvironmentCredential() {
             .expectNextMatches(token -> "token".equals(token.getToken()))
             .verifyComplete();
     }
+
+    @Test
+    public void testCreateEnvironmentClientCertificateCredential() {
+        Configuration configuration = Configuration.getGlobalConfiguration();
```
alzimmermsft: Could remove the local instance since `Configuration` uses the fluent pattern for `put`.

sdk/identity/azure-identity/src/test/java/com/azure/identity/EnvironmentCredentialTests.java
```diff
@@ -36,4 +36,46 @@ public void testCreateEnvironmentCredential() {
...
+                Assert.assertFalse(message != null && message.contains("Cannot create any credentials with the current environment variables"));
+                return Mono.just(new AccessToken("token", OffsetDateTime.MAX));
+            }))
+            .expectNextMatches(token -> "token".equals(token.getToken()))
+            .verifyComplete();
+    }
+
+    @Test
+    public void testCreateEnvironmentUserPasswordCredential() {
+        Configuration configuration = Configuration.getGlobalConfiguration();
```
alzimmermsft: Same as the previous comment

sdk/identity/azure-identity/src/test/java/com/azure/identity/EnvironmentCredentialTests.java
```diff
@@ -36,4 +36,46 @@ public void testCreateEnvironmentCredential() {
...
+    @Test
+    public void testCreateEnvironmentClientCertificateCredential() {
+        Configuration configuration = Configuration.getGlobalConfiguration();
+        configuration.put(Configuration.PROPERTY_AZURE_CLIENT_ID, "foo");
+        configuration.put(Configuration.PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH, "bar");
+        configuration.put(Configuration.PROPERTY_AZURE_TENANT_ID, "baz");
+
+        EnvironmentCredential credential = new EnvironmentCredentialBuilder().build();
+
+        // authentication will fail client-id=foo, but should be able to create ClientSecretCredential
```
ellismg: NIT: I think this comment should say "ClientCertificateCredential" instead of "ClientSecretCredential"

sdk/identity/azure-identity/src/test/java/com/azure/identity/EnvironmentCredentialTests.java
```diff
@@ -36,4 +36,46 @@ public void testCreateEnvironmentCredential() {
...
+    @Test
+    public void testCreateEnvironmentUserPasswordCredential() {
+        Configuration configuration = Configuration.getGlobalConfiguration();
+        configuration.put(Configuration.PROPERTY_AZURE_CLIENT_ID, "foo");
+        configuration.put(Configuration.PROPERTY_AZURE_USERNAME, "bar");
+        configuration.put(Configuration.PROPERTY_AZURE_PASSWORD, "baz");
+
+        EnvironmentCredential credential = new EnvironmentCredentialBuilder().build();
+
+        // authentication will fail client-id=foo, but should be able to create ClientSecretCredential
```
ellismg: NIT: Similar feedback as above, I think this should be UsernamePasswordCredential in the comment.

sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
cxznmhdcxz: No it's not, but supply it anyway if the environment variable is set.

sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
cxznmhdcxz: Removed from `Configuratoin`, now a private method in `EnvironmentCredential`.

PR #7340, opened 2020-01-10
Keyvault ARM template for live test

hemanttanwar:

The pipeline will pass on names of these resources as ENV variable into test. For example AZURE_KEYVAULT_ENDPOINT

### DIFF
--- a/sdk/keyvault/test-resources.json
+++ b/sdk/keyvault/test-resources.json
@@ -1,0 +1,65 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "baseName": {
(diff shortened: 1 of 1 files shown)


hemanttanwar:
Some of the live test failure are known and Vinay is working on them.

sdk/keyvault/test-resources.json
```diff
@@ -0,0 +1,65 @@
...
+            "properties": {
+                "sku": {
+                    "family": "A",
+                    "name": "standard"
+                },
+                "tenantId": "[parameters('tenantId')]",
+                "accessPolicies": [
+                    {
+                        "tenantId": "[parameters('tenantId')]",
+                        "objectId": "[parameters('testApplicationOid')]",
```
g2vinay: IS this application object id, of the azure-sdk-tests (service principal)
@hemanttanwar @danieljurek

Issue #6955, opened 2019-12-19
readmywrites workflow on Direct HTTPS has failures because of reactor netty NotYetConnectedException


mbhaskar:
ERROR com.azure.cosmos.implementation.query.DocumentProducer - Unexpected failure
CosmosClientException{error=null, resourceAddress='null', statusCode=0, message=null, causeInfo=[class: class java.nio.channels.NotYetConnectedException, message: null], responseHeaders={}, requestHeaders={authorization=type%3Dmaster%26ver%3D1.0%26sig%3DZk%2BRtwqDdCvHdWB7tTLpGCyKwstITXzHpHQcq2uJ6mQ%3D, Accept=application/json, x-ms-session-token=0:-1#50102,1:-1#71225,2:-1#71251, x-ms-date=Wed, 18 Dec 2019 22:12:20 GMT, x-ms-documentdb-isquery=true, x-ms-documentdb-partitionkey=["756fdf90-21e3-11ea-8e94-fd4d274e06f7"], x-ms-max-item-count=100, x-ms-continuation=null, x-ms-consistency-level=Session, Content-Type=application/query+json}}
	at com.azure.cosmos.BridgeInternal.createCosmosClientException(BridgeInternal.java:345)
	at com.azure.cosmos.implementation.RxGatewayStoreModel.lambda$toDocumentServiceResponse$2(RxGatewayStoreModel.java:307)
	at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onError(FluxOnErrorResume.java:88)
	at reactor.core.publisher.FluxMapFuseable$MapFuseableSubscriber.onError(FluxMapFuseable.java:134)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.onError(MonoFlatMap.java:165)
	at reactor.core.publisher.MonoSingle$SingleSubscriber.onError(MonoSingle.java:141)
	at reactor.core.publisher.MonoFlatMapMany$FlatMapManyMain.onError(MonoFlatMapMany.java:197)
	at reactor.core.publisher.FluxRetryPredicate$RetryPredicateSubscriber.onError(FluxRetryPredicate.java:101)
	at reactor.core.publisher.MonoCreate$DefaultMonoSink.error(MonoCreate.java:183)
	at reactor.netty.http.client.HttpClientConnect$MonoHttpConnect$TcpClientSubscriber.onError(HttpClientConnect.java:345)
[...]

PR #7345, opened 2020-01-10
Empty response content fix 

kushagraThapar:
Return empty string when receiving empty response from backend. 
Addresses this: https://github.com/Azure/azure-sdk-for-java/issues/6955

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/ErrorUtils.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/ErrorUtils.java
@@ -17,7 +17,7 @@ public class ErrorUtils {
     private static final Logger logger = LoggerFactory.getLogger(ErrorUtils.class);
 
     static Mono<String> getErrorResponseAsync(HttpResponse responseMessage, HttpRequest request) {
-        Mono<String> responseAsString = responseMessage.bodyAsString();
+        Mono<String> responseAsString = responseMessage.bodyAsString().switchIfEmpty(Mono.just(StringUtils.EMPTY));
         if (request.httpMethod() == HttpMethod.DELETE) {
             return Mono.just(StringUtils.EMPTY);
         }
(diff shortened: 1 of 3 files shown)


sdk/cosmos/azure-cosmos-benchmark/src/test/java/com/azure/cosmos/benchmark/ReadMyWritesConsistencyTest.java
```diff
@@ -132,6 +132,7 @@ public void readMyWrites(boolean useNameLink) throws Exception {
         ReadMyWriteWorkflow wf = new ReadMyWriteWorkflow(cfg) {
             @Override
             protected void onError(Throwable throwable) {
+                logger.error("Error occurred in ReadMyWriteWorkflow", throwable);
```
kirankumarkolli: Will error level enabled check come later?

kushagraThapar: Since this is logging error in tests, I don't think if we need error level enabled check for this. 
This is only for debugging CI

moderakh: slf4j internally will check the log level, so checking log level is not needed in this case.

checking log level is needed only when the information needed to be logged requires some pre-computation. For example in the following case checking log level helps:

```java
// checking log level helps
if (logger.isDebugEnabled()) {
 String diagnosticsInfoAsString = diagnosticsInfo.toJson()
 logger.debug("diagnostics info is {}, diagnosticsInfoAsString);
}
```
but here it is not needed
```java
// checking log level is not needed as no pre-computation is done here. slf4 will do internal checking.
 logger.debug("endpoint is {}, endpoint);
}
```

kushagraThapar: Agreed, thanks for the reference @moderakh

PR #7354, opened 2020-01-10
DRAFT: DON'T REVIEW. improve write json serialization/de-serialization path

moderakh:
- when pk is explicitly provided in the request option, there is no need in parsing json
- avoid multiple rounds of json serialization/desrialization when possible in the write path.


moderakh:
withdrawn, still working on it

PR #7363, opened 2020-01-12
Disable feature CI builds by default.

mitchdenny:
This PR disables builds on ```feature/*``` branches by default. Teams that are doing long running builds can enable builds on their feature branch for just their service directories as required, but we found people doing merges from master into feature branches whilst this was the default for all the pipelines was causing build storms.

### DIFF
--- a/sdk/appconfiguration/ci.yml
+++ b/sdk/appconfiguration/ci.yml
@@ -15,7 +15,6 @@ trigger:
   branches:
     include:
       - master
-      - feature/*
       - hotfix/*
       - release/*
   paths:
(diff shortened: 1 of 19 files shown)


alzimmermsft:
What out on this blowing up the agent usage on internal when this is merged.

JimSuplizio:
@mitchdenny we really need to do something with spring and search CI builds, at least turn them off until there's something to build in those branches.

mitchdenny:
> we really need to do something with spring and search CI builds, at least turn them off until there's something to build in those branches.

They were expected to run (and fail) here. Can you point to a PR where you don't think that they should run?

mitchdenny:
I will merge this later today after Redmond rolls of work hours so I don't eat the pool.

PR #7389, opened 2020-01-13
Update documentation and in-memory checkpoint store sample

srnagar:
Updating some documentation and in-memory checkpoint store implementation to support multiple namespaces.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/EventContext.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/EventContext.java
@@ -3,9 +3,6 @@
 
 package com.azure.messaging.eventhubs.models;
 
-import static com.azure.core.util.FluxUtil.monoError;
-
-import com.azure.core.util.logging.ClientLogger;
 import com.azure.messaging.eventhubs.CheckpointStore;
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/EventContext.java
```diff
@@ -103,6 +99,6 @@ public LastEnqueuedEventProperties getLastEnqueuedEventProperties() {
      * successfully processed event in this partition if the update is successful.
      */
     public void updateCheckpoint() {
-        this.updateCheckpointAsync(eventData).block();
+        this.updateCheckpointAsync().block();
```
conniey: for the .block(), should we pass in the "retryOptions.operationTimeout" so it doesn't block forever (or I think default is 5 mins)?

srnagar: Since this is just a delegate method to the checkpoint store's updateCheckpoint(), I think the checkpoint store (blob client or any other store) should be configured with the necessary timeouts and retry options. If the store they use is super-slow, they might want a different timeout setting than the one they use for Event Hubs (in retryOptions).

Maybe we can defer adding a timeout to a later release if that's required.

PR #7401, opened 2020-01-13
fixes some bugs and add tests to improve code coveragwe

mssfang:
- Add client constants class to maintains all cosntants for App configuration client.
- 'id' is type of String, could be null or empty

- Added tests for code coverage for `ConfigurationClientBuilder` and ``SettingSelector.

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
@@ -123,20 +123,17 @@ public class ConfigurationClientBuilderTest extends TestBase {
 
         Objects.requireNonNull(connectionString, "`AZURE_APPCONFIG_CONNECTION_STRING` expected to be set.");
 
-        final ConfigurationClientBuilder clientBuilder;
+        final ConfigurationClientBuilder clientBuilder = new ConfigurationClientBuilder()
+            .connectionString(connectionString)
+            .retryPolicy(new RetryPolicy())
+            .httpLogOptions(new HttpLogOptions().setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS))
+            .serviceVersion(null);
+
         if (interceptorManager.isPlaybackMode()) {
-            clientBuilder = new ConfigurationClientBuilder()
-                .connectionString(connectionString)
-                .httpClient(interceptorManager.getPlaybackClient())
-                .httpLogOptions(new HttpLogOptions().setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS));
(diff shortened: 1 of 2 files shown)


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
```diff
@@ -355,14 +355,14 @@ public ConfigurationClientBuilder configuration(Configuration configuration) {
...
-     * Sets the {@link HttpPipelinePolicy} that is used when each request is sent.
+     * Sets the {@link RetryPolicy} that is used when each request is sent.
      *
      * The default retry policy will be used if not provided {@link ConfigurationClientBuilder#buildAsyncClient()}
      * to build {@link ConfigurationAsyncClient} or {@link ConfigurationClient}.
      * @param retryPolicy user's retry policy applied to each request.
      * @return The updated ConfigurationClientBuilder object.
      */
-    public ConfigurationClientBuilder retryPolicy(HttpPipelinePolicy retryPolicy) {
+    public ConfigurationClientBuilder retryPolicy(RetryPolicy retryPolicy) {
```
alzimmermsft: This breaks public API and won't be allowed. The old method either needs to be deprecated and this put in it's place or remmain as is.

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/ConfigurationClientCredentials.java
```diff
@@ -212,7 +211,7 @@ String id() {
             this.id = id;
             this.secret = secret;
 
-            if (this.baseUri == null || this.id == null || this.secret == null) {
+            if (this.baseUri == null || CoreUtils.isNullOrEmpty(this.id) || this.secret == null) {
```
alzimmermsft: Why is id treated differently than baseUri or secret, won't all have issues if null or empty?

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java
mssfang: ok. I will revert back the change

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/implementation/ConfigurationClientCredentials.java
mssfang: id' is type of String, could be null or empty
but baseUri is URL type, `private final URL baseUri;`.

PR #7403, opened 2020-01-14
Update Download To File Write Logic

alzimmermsft:
`FluxUtil.writeFile` is updated to pass a copied version of the buffer it is writing to file. This change is made since the method cannot be certain of where the data stream is coming from and this copying will prevent the buffer from being changed underneath the method before it is written to the file. This change comes with the cost of duplicating memory but ensures that the buffer will remain consistent during the entirety of the write operation.

### DIFF
--- a/sdk/identity/azure-identity/pom.xml
+++ b/sdk/identity/azure-identity/pom.xml
@@ -12,6 +12,10 @@
   <description>This module contains client library for Microsoft Azure Identity.</description>
   <url>https://github.com/Azure/azure-sdk-for-java</url>
 
+  <properties>
+    <jacoco.min.branchcoverage>0.28</jacoco.min.branchcoverage>
+  </properties>
+
   <parent>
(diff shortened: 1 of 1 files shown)


rickle-msft:
I think we should take this and do a patch release so we can mitigate the data corruption issues, but can we also open another issue to explore some of the optimizations you mentioned in your email to try to ameliorate some of the perf issues this will introduce?

alzimmermsft:
@mitchdenny Mind unblocking the check enforcer step?

mitchdenny:
You just forgot to trigger evaluation :) Normally it won't be required, its just this PR was created whilst I was having some problems ;)

alzimmermsft:
@rickle-msft I created this follow-up [issue](https://github.com/Azure/azure-sdk-for-java/issues/7425) to source when the buffer is being released back into the available pool and potentially finding configuration or new code patterns to reduce the risk of this happening before the operation is complete.

Issue #7271, opened 2020-01-08
Aggregate quality/engineering reports


mitchdenny:
I've created this issue to start a conversation about our strategy around aggregate reporting requirements in the Java repo. My understanding is that the aggregate reports allow the team to get a 30,000ft view of potential quality issues across all track 2 libraries in the repository, and that there is a strong view that we want to retain this capability.

At the same time we've been doing work from the engineering systems perspective to try and make the pipeline for each set of artifacts as independent as possible so that an issue in one part of the code base doesn't break the build and release capability across all pipelines. We've also been wanting to do some structural work to move our parent POMs into the ```sdk/parents``` folder.

As I've been working on this PR I've come up against issues where aggregate reporting, and the need to have modules referenced in parent POMs is at odds with independent pipelines.

https://github.com/Azure/azure-sdk-for-java/pull/6424

It looks like Alan is already doing some work to make it possible to execute spot bugs against the ```pom.service.xml``` aggregators which is complimentary to the PR I am working on above:

https://github.com/Azure/azure-sdk-for-java/pull/6661

What I would like to get out of raising this issue is consensus on:

1. Moving away from ```pom.client.xml``` or any parent POMs having module references, and making them pure parents.
2. Allowing any analysis steps like spotbugs, checkstyle, revapi etc to be run directly against the module (e.g. ```sdk/core/azure-core/pom.xml```) or the aggregator POM (e.g. ```sdk/core/pom.service.xml```).
3. Favor execution of these now more narrowly scoped goals as part of the standard workflow so that developers get feedback on their dev boxes with low effort.

BUT ... the above moves us away from an aggregate reporting capability. And I want to have our cake and eat it too. My proposal is that we have a special pipeline whose purpose is to reference all our modules and execute the aggregate steps that are desired. This would then publish those aggregate reports wherever they need to go, similar to what java - client has done in the past.

The individual pipelines should still produce reports scoped to what they build and ship, but they probably wouldn't publish them anywhere (although they could be run locally and pulled from the pipeline aritfacts).

I feel like we are halfway there now with the work that Alan has been doing and I see that we have some issues in the repo already that are pushing that direction. I just wanted to see if this is something that we can state as an explicit goal.

/cc @JonathanGiles @JimSuplizio @conniey @alzimmermsft @weshaggard

alzimmermsft:
[...]

PR #7406, opened 2020-01-14
Investigation into Dependency Checker performance degredation in in Cosmos

mitchdenny:
Dependency Checker wasn't using our DefaultOptions variable. Unfortunately that meant that it wasn't using our Maven mirror and we are suspecting this might be impacted package download performance. This PR is to validate that.

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -1,6 +1,6 @@
 [
   {
     "projectName": "azure-sdk-for-java",
-    "pomUrls": ["file:./pom.client.xml"]
+    "pomUrls": ["file:./sdk/cosmos/pom.service.xml"]
   }
 ]


mitchdenny:
Yep ... although this doesn't appear to have solved the problem. So I need to continue to investigate.

mitchdenny:
@JonathanGiles and @JimSuplizio I've made some progress here, but I am not 100% sure why. In this PR I modified dependency checker to work against the cosmos POM (just updated dependencies.json in the root of the repo to try it out).

It ran much faster when things were scoped to just the Cosmos aggregator POM. I'm thinking that we should actually do this across all pipelines inline with where I am heading on [this issue](https://github.com/Azure/azure-sdk-for-java/issues/7271).

However, this doesn't entirely explain WHY this is an issue. My initial hypothesis is that the ```Install Module and Run SpotBugs, Checksytle, RevApi, and JavaDoc``` step switched over to running against the service aggregator POM (```sdk/cosmos/pom.service.xml```). As a result the local repository wouldn't have as many packages in it (because it didn't process ```pom.client.xml```). The result would be then that the dependency checker has to download more stuff that isn't available locally which might lead to the timeout. This would also explain why builds on ```sdk/core/pom.service.xml``` don't manifest the problem, because that aggregator pulls in all the track 2 packages.

This is still my best guess, but I'm not sure it fits 100% of the facts just yet.

mitchdenny:
Agreed. Closing this PR in favor of [this PR](https://github.com/Azure/azure-sdk-for-java/pull/7489
) merging in the near future.

Issue #6575, opened 2019-11-27
[QUERY] Anonymous requests to storage blobs


SukruthKS:
**Query/Question**
How do I make anonymous requests to Azure Storage? When I don't provide any credentials to service client builder, it throws:

```
BlobServiceClient client = new BlobServiceClientBuilder()
    .endpoint("https://<account_name>.blob.core.windows.net")
    .buildClient();
```

Error:
```
java.lang.IllegalArgumentException: Authorization credentials must be set.
at com.azure.storage.blob.BlobServiceClientBuilder.lambda$buildAsyncClient$0(BlobServiceClientBuilder.java:98)
at com.azure.storage.blob.implementation.util.BuilderHelper.buildPipeline(BuilderHelper.java:61)
at com.azure.storage.blob.BlobServiceClientBuilder.buildAsyncClient(BlobServiceClientBuilder.java:90)
at com.azure.storage.blob.BlobServiceClientBuilder.buildClient(BlobServiceClientBuilder.java:82)
```

The below code works with version 8 of the SDK:
```
CloudBlobClient client = new CloudBlobClient(<endpoint>, StorageCredentialsAnonymous.ANONYMOUS);
```

***Why is this not a Bug or a feature Request?***
NA

**Setup (please complete the following information if applicable):**
 - OS: Linux
 - IDE : IntelliJ
 - Version of the Library used: 12.0.0
 
 **Information Checklist**
 Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Query Added
- [x] Setup information Added

gapra-msft:
Hi @SukruthKS, 

Only Blobs and Containers are able to allow public access to your so to make an anonymous request to Blob Storage you'll want to use the BlobClientBuilder/ContainerClientBuilder  depending on your needs. 

Feel free to close this issue if this helps unblock you.

SukruthKS:
Thanks @gapra-msft ! We always construct container and blob clients from `BlobServiceClient` and as soon as I build the service client (before making any requests), it fails with the above error. Any possibility of changing the behavior of `BlobServiceClient` to not throw for anonymous access?

rickle-msft:
[...]

PR #7421, opened 2020-01-14
Added ability to create service clients anonymously

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/6575


sdk/storage/azure-storage-blob/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Release History
 
+## 12.3.0-beta.1 (Unreleased)
+- Added ability to create service clients anonymously. Anonymous service clients will throw on attempting to create network requests.
```
rickle-msft: Maybe add "and should only be used to create anonymous container and blob clients." Just in case people wonder "well then what's the point?" lol

alzimmermsft:
Just want to make sure I understand the reasoning behind this change, given that making a call to `checkAccess` in every API and passing along `anonymousAccess` feel like an anti-pattern.

**Scenario**

I have multiple contains which allow public access and I want to generate clients to interact with them.

_Option 1_

I create a `BlobServiceClient` which doesn't have authentication information and begin creating `BlobContainerClient`s with `getBlobContainerClient`.

_Option 2_

I configure a `BlobContainerClientBuilder` which doesn't have authentication information and begin creating `BlobContainerClient`s by setting container name and calling `buildClient`.

This PR is to add support for option 1, correct? Would it be better to push on using option 2 as that falls in-line with the reason why the builder pattern was chosen?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -570,7 +581,18 @@ public String getAccountName() {
...
     public String generateAccountSas(AccountSasSignatureValues accountSasSignatureValues) {
+        checkAccess();
         return new AccountSasImplUtil(accountSasSignatureValues)
             .generateSas(SasImplUtils.extractSharedKeyCredential(getHttpPipeline()));
     }
+
+    /**
+     * Checks if service client was built with credentials.
+     */
+    private void checkAccess() {
```
rickle-msft: I think somehow having the word anonymous in the method name will make it more self descriptive.

gapra-msft: I'll rename it to throwOnAnonymousAccess

gapra-msft:
> Just want to make sure I understand the reasoning behind this change, given that making a call to `checkAccess` in every API and passing along `anonymousAccess` feel like an anti-pattern.
> 
> **Scenario**
> 
> I have multiple contains which allow public access and I want to generate clients to interact with them.
> 
> _Option 1_
> 
> I create a `BlobServiceClient` which doesn't have authentication information and begin creating `BlobContainerClient`s with `getBlobContainerClient`.
> 
> _Option 2_
> 
> I configure a `BlobContainerClientBuilder` which doesn't have authentication information and begin creating `BlobContainerClient`s by setting container name and calling `buildClient`.
> 
> This PR is to add support for option 1, correct? Would it be better to push on using option 2 as that falls in-line with the reason why the builder pattern was chosen?

Option 2 was suggested to the customer but it looks like option 1 is more along the lines of what they wanted as a solution.

Issue #6627, opened 2019-12-04
[BUG] Setting CORS policies on Azure storage account


SukruthKS:
**Describe the bug**
When setting a new CORS rule on the storage account using the SDK, if any of the properties like ExposedHeaders, MaxAgeInSeconds etc are not specified or set to null, the SDK skips those elements in the request body XML and the server sends back an error message saying the XML is not syntactically valid. 

***Exception or Stack Trace***
Request body sent when `ExposedHeaders` property was not specified in the CORS rule:

```
INFO com.azure.storage.blob.implementation.ServicesImpl$ServicesService.setProperties - 913-byte body:%n<?xml version='1.0' encoding='UTF-8'?><StorageServiceProperties><Logging><Version>1.0</Version><Delete>false</Delete><Read>false</Read><Write>false</Write><RetentionPolicy><Enabled>false</Enabled></RetentionPolicy></Logging><HourMetrics><Version>1.0</Version><Enabled>true</Enabled><RetentionPolicy><Enabled>true</Enabled><Days>7</Days></RetentionPolicy><IncludeAPIs>true</IncludeAPIs></HourMetrics><MinuteMetrics><Version>1.0</Version><Enabled>false</Enabled><RetentionPolicy><Enabled>false</Enabled></RetentionPolicy></MinuteMetrics><Cors><CorsRule><AllowedOrigins>snowflakecomputing.com</AllowedOrigins><AllowedMethods>GET,HEAD</AllowedMethods><AllowedHeaders>x-ms-version</AllowedHeaders><MaxAgeInSeconds>60</MaxAgeInSeconds></CorsRule></Cors><DeleteRetentionPolicy><Enabled>true</Enabled><Days>1</Days></DeleteRetentionPolicy><StaticWebsite><Enabled>false</Enabled></StaticWebsite></StorageServiceProperties>
```

Error from the server:

```
<?xml version="1.0" encoding="utf-8"?>
<Error>
  <Code>InvalidXmlDocument</Code>
  <Message>XML specified is not syntactically valid. RequestId:7772028a-401e-005a-6d44-aa80b6000000 Time:2019-12-04T01:44:21.6326037Z</Message>
  <LineNumber>0</LineNumber>
  <LinePosition>0</LinePosition>
  <Reason />
</Error>
```

**To Reproduce**
```
BlobServiceProperties serviceProperties = blobServiceClient.getProperties();

BlobCorsRule rule = new BlobCorsRule();
[...]

PR #7426, opened 2020-01-14
Added client side checks for account set props

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/6627

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
@@ -365,6 +365,8 @@ public final class BlobServiceAsyncClient {
      * <a href="https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-service-properties">Azure Docs</a>.
      * Note that setting the default service version has no effect when using this client because this client explicitly
      * sets the version header on each request, overriding the default.
+     * <p>This method checks to ensure the properties being sent follow the specifications indicated in the Azure Docs.
+     * If CORS policies are set, CORS parameters that are not set default to the empty string.</p>
      *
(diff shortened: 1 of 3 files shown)


sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
```diff
@@ -270,6 +270,22 @@ class ServiceAPITest extends APISpec {
...
+        def serviceProperties = primaryBlobServiceClient.getProperties()
+
+        def rule = new BlobCorsRule()
+        rule.setAllowedOrigins("microsoft.com")
+        rule.setMaxAgeInSeconds(60)
+        rule.setAllowedMethods("GET")
+        rule.setAllowedHeaders("x-ms-version")
+
+        serviceProperties.setCors(Collections.singletonList(rule))
+
```
rickle-msft: Can you leave a comment calling out that some properties are not set and the test validates that we fill them with null? Otherwise it'll be easy to forget what this test is for.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -391,10 +395,89 @@ public String getAccountUrl() {
...
+            // Static Website
+            finalProperties.setStaticWebsite(properties.getStaticWebsite());
+
+        }
+
+        return this.azureBlobStorage.services().setPropertiesWithRestResponseAsync(finalProperties, null, null, context)
             .map(response -> new SimpleResponse<>(response, null));
     }
 
+    private BlobCorsRule validatedCorsRule(BlobCorsRule originalRule) {
```
rickle-msft: Can you leave a comment that mentions that the service requires these fields not be null so we change null to empty string?

PR #7429, opened 2020-01-15
Moving CosmosItem (read, replace, delete) API onto CosmosContainer

mbhaskar:
Moving the following API from CosmosItem to CosmosContainer
- readItem
- replaceItem
- deleteItem

Remove CosmosItem
Add support to typed item API
Modifying tests to reflect the new item api

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseDiagnosticsTest.java
@@ -118,7 +118,7 @@ public class CosmosResponseDiagnosticsTest extends TestSuiteBase {
             CosmosItemRequestOptions cosmosItemRequestOptions = new CosmosItemRequestOptions();
             cosmosItemRequestOptions.setPartitionKey(new PartitionKey("wrongPartitionKey"));
             CosmosItemResponse<CosmosItemProperties> readResponse =
-                this.container.readItem(createResponse.getProperties().getId(), 
+                cosmosContainer.readItem(createResponse.getProperties().getId(), 
                                         new PartitionKey("wrongPartitionKey"),
                                         CosmosItemProperties.class);
             fail("request should fail as partition key is wrong");


sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncReadBenchmark.java
```diff
@@ -53,7 +54,9 @@ protected void performWorkload(BaseSubscriber<CosmosAsyncItemResponse> baseSubsc
         PojoizedJson doc = docsToRead.get(index);
 
         String partitionKeyValue = doc.getId();
-        Mono<CosmosAsyncItemResponse> result = cosmosAsyncContainer.getItem(doc.getId(), partitionKeyValue).read();
+        Mono<CosmosAsyncItemResponse<PojoizedJson>> result = cosmosAsyncContainer.readItem(doc.getId(),
+                                                                                           new PartitionKey(partitionKeyValue),
```
moderakh: can't we assume that the second param is always partition key?

that way the caller doesn't have to explicitly wrap pk value in PartitionKey and implementation will take care of that.

thoughts?

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncWriteBenchmark.java
```diff
@@ -60,13 +61,13 @@ protected void hookOnError(Throwable throwable) {
...
         String partitionKey = uuid + i;
-        Mono<CosmosAsyncItemResponse> obs;
+        Mono<CosmosAsyncItemResponse<PojoizedJson>> obs;
         if (configuration.isDisablePassingPartitionKeyAsOptionOnWrite()) {
             // require parsing partition key from the doc
             obs = cosmosAsyncContainer.createItem(generateDocument(partitionKey, dataFieldValue));
         } else {
             // more optimized for write as partition ke is already passed as config
-            obs = cosmosAsyncContainer.createItem(generateDocument(partitionKey, dataFieldValue), new CosmosItemRequestOptions(partitionKey));
+            obs = cosmosAsyncContainer.createItem(generateDocument(partitionKey, dataFieldValue), new CosmosItemRequestOptions());
```
moderakh: we should pass partition key to CosmosItemRequestOption here.

The intention of this benchmark is to measure perf when pk is passed as request options.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncItemResponse.java
```diff
@@ -6,38 +6,37 @@
...
-     * Gets the itemSettings
+     * Gets the resource .
      *
-     * @return the itemSettings
+     * @return the resource
      */
-    public CosmosItemProperties getProperties() {
-        return super.getProperties();
+    public T getResource(){
+        return super.getProperties().toObject(itemClassType);
```
moderakh: this will result in double serialization (which was fixed prior to this PR)

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncReadBenchmark.java
mbhaskar: We are following the .net model of making partitionkey object manadatory. 
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/8f1375a30799b3acf95d843b2db0c2447cbc3876/Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs#L476

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncItemResponse.java
mbhaskar: Changed this to avoid double serialization. Now we directly use the resource string to convert to user object.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncWriteBenchmark.java
mbhaskar: Added new API accept partition key and changed this call accordingly.

sdk/cosmos/azure-cosmos-examples/src/main/java/com/azure/cosmos/examples/BasicDemo.java
```diff
@@ -2,20 +2,26 @@
...
 import com.azure.cosmos.CosmosContainerProperties;
+import com.azure.cosmos.CosmosItemProperties;
 import com.azure.cosmos.FeedOptions;
 import com.azure.cosmos.FeedResponse;
+import com.azure.cosmos.PartitionKey;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
 
+import static java.lang.System.exit;
```
kushagraThapar: Seems like an unused import, please remove it if it is.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -179,23 +179,34 @@ CosmosAsyncContainer setId(String id) {
...
+     * @return an {@link Mono} containing the single resource response with the created cosmos item or an error.
      */
-    public Mono<CosmosAsyncItemResponse> createItem(Object item, CosmosItemRequestOptions options) {
+    public <T> Mono<CosmosAsyncItemResponse<T>> createItem(T item,
+                                                           PartitionKey partitionKey,
+                                                           CosmosItemRequestOptions options) {
+        if (options == null) {
+            options = new CosmosItemRequestOptions();
+        }
+        options.setPartitionKey(partitionKey);
```
kushagraThapar: What will be the behavior if `partitionKey` is passed as null ?
I see a couple of options here. 
Either default it to `PartitionKey.NONE` or use something like this `checkNoNull(partitionKey, "Partition Key should not be null")` 

Please make sure this behavior is consistent across all the APIs.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -366,14 +375,138 @@ CosmosAsyncContainer setId(String id) {
...
-     * Gets a CosmosAsyncItem object without making a service call
+     * Reads an item.
+     * <p>
+     * After subscription the operation will be performed.
+     * The {@link Mono} upon successful completion will contain a cosmos item response with the read item
+     * In case of failure the {@link Mono} will error.
+     *
+     * @param itemId the item id
+     * @param partitionKey the partition key
+     * @return an {@link Mono} containing the cosmos item response with the read item or an error
```
kushagraThapar: 3rd `@param` is missing. 
Please replace `an` with `a` in return statement before `Mono`. 

Please do this change here and everywhere else.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -366,14 +375,138 @@ CosmosAsyncContainer setId(String id) {
...
+    public <T> Mono<CosmosAsyncItemResponse<T>>  readItem(String itemId, PartitionKey partitionKey, Class<T> itemType) {
+        return readItem(itemId, partitionKey, new CosmosItemRequestOptions(partitionKey), itemType);
+    }
+
+    /**
+     * Reads an item.
+     * <p>
+     * After subscription the operation will be performed.
+     * The {@link Mono} upon successful completion will contain a cosmos item response with the read item
+     * In case of failure the {@link Mono} will error.
```
kushagraThapar: I have got this feedback from various customers, that on `readItem() API`, in v2 SDK if the item is not found, we used to return `empty Observable()`. 
Because it becomes really difficult for the end user to catch 404 and do nothing. 

@moderakh @kirankumarkolli @mbhaskar what should we do with v4 ?

@mbhaskar do you know what .net behavior is ?
As I have also received feedback that we should be consistent across our SDKs.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -366,14 +375,138 @@ CosmosAsyncContainer setId(String id) {
...
+                                                  CosmosItemRequestOptions options, Class<T> itemType) {
+        if (options == null) {
+            options = new CosmosItemRequestOptions();
+        }
+        options.setPartitionKey(partitionKey);
+        RequestOptions requestOptions = options.toRequestOptions();
+        return this.getDatabase().getDocClientWrapper()
+                   .readDocument(getItemLink(itemId), requestOptions)
+                   .map(response -> new CosmosAsyncItemResponse<T>(response, itemType))
+                   .single();
```
kushagraThapar: For V4, I think we can change our `AsyncDocumentClient` APIs to return Mono for single returns so that we won't have to call `.single()` APIs here. 
Right now they return Flux<> for readItem which doesn't make sense. 
I believe @moderakh raised this point with v3. 
We should correct that behavior for v4 for all the APIs if they return a single document / item. 

Calling this extra `.single()` when not needed will affect the performance of these APIs.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncItemResponse.java
```diff
@@ -4,40 +4,41 @@
...
-     * Gets the itemSettings
+     * Gets the resource .
      *
-     * @return the itemSettings
+     * @return the resource
      */
-    public CosmosItemProperties getProperties() {
-        return super.getProperties();
+    public T getResource(){
+        return Utils.parse(responseBodyString, itemClassType);
```
kushagraThapar: Since `responseBodyString` can be empty string, can we please make sure it is handled well when parsing it to an `itemClassType` ?
Do we want to throw error in that case, or return null ?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainer.java
```diff
@@ -175,8 +193,8 @@ public CosmosItemResponse upsertItem(Object item) throws CosmosClientException {
      * @return the cosmos sync item response
      * @throws CosmosClientException the cosmos client exception
      */
-    public CosmosItemResponse upsertItem(Object item, CosmosItemRequestOptions options) throws CosmosClientException {
-        return this.mapItemResponseAndBlock(this.containerWrapper.createItem(item, options));
+    public <T> CosmosItemResponse<T> upsertItem(Object item, CosmosItemRequestOptions options) throws CosmosClientException {
+        return (CosmosItemResponse<T>) this.mapItemResponseAndBlock(this.asyncContainer.upsertItem(item, options));
```
kushagraThapar: I see `mapItemResponseAndBlock()` returns `CosmosItemResponse<T>`. 
Do we need this extra casting here ?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainer.java
```diff
@@ -246,21 +280,61 @@ CosmosItemResponse mapItemResponseAndBlock(Mono<CosmosAsyncItemResponse> itemMon
...
+                                          CosmosItemRequestOptions options, Class<T> itemType) throws CosmosClientException {
+        return this.mapItemResponseAndBlock(asyncContainer.readItem(itemId, partitionKey, options, itemType));
+    }
+
+    /**
+     * Replace cosmos sync item response.
+     *
+     * @param item the item
+     * @param options the options
+     * @return the cosmos sync item response
```
kushagraThapar: Please fill the gaps of missing`@param` here and other API docs.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosResponse.java
```diff
@@ -2,13 +2,14 @@
 // Licensed under the MIT License.
 package com.azure.cosmos;
 
+import com.azure.cosmos.implementation.Constants;
```
kushagraThapar: Seems like unused import.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/Utils.java
```diff
@@ -548,4 +549,12 @@ public ValueHolder(V v) {
         holder.v = dictionary.remove(key);
         return holder.v != null;
     }
+
+    public static <T> T parse(String itemResponseBodyAsString, Class<T> itemClassType) {
```
kushagraThapar: I looked up and there are 4 places where we use `Utils.getsimpleObjectMapper().readValue()` APIs. 
It would be great if we can consolidate them here and have this logic at one place.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -366,14 +375,138 @@ CosmosAsyncContainer setId(String id) {
...
+     * @param itemId the item id
      * @param partitionKey the partition key
-     * @return a cosmos item
+     * @param options the request cosmosItemRequestOptions
+     * @return an {@link Mono} containing the cosmos item response with the read item or an error
      */
-    public CosmosAsyncItem getItem(String id, Object partitionKey) {
-        return new CosmosAsyncItem(id, partitionKey, this);
+    public <T> Mono<CosmosAsyncItemResponse<T>>  readItem(String itemId, PartitionKey partitionKey, 
+                                                  CosmosItemRequestOptions options, Class<T> itemType) {
```
moderakh: passing partition key has ambiguity.

1. partition key can be passed as the second param to readItem
2. CosmosItemRequestOptions has support for `setPartitionKey()`.

if we are always requiring partition key to be passed as a standalone param then the support for setting partition key in `CosmosItemRequestOptions` has to be removed to avoid confusion on public surface area.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosItemRequestOptions.java
```diff
@@ -175,7 +171,7 @@ public CosmosItemRequestOptions setPartitionKey(PartitionKey partitionKey) {
      *
      * @return the partition key
      */
-    public PartitionKey getPartitionKey() {
+    PartitionKey getPartitionKey() {
```
moderakh: the public setter for partition key needs to be removed too because the partition key is a mandatory standalone param on reading/deleting item and doesn't have to be in the cosmosItemRequestOption anymore

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/PartitionKey.java
```diff
@@ -25,7 +25,13 @@
      */
     @SuppressWarnings("serial")
     public PartitionKey(final Object key) {
-        this.internalPartitionKey = PartitionKeyInternal.fromObjectArray(new Object[] {key}, true);
+        // Now that we are making partitionkey as one of the parameters for creating object, just making sure nothing
+        // wrong happens here when user does something like new PartitionKey(PartitionKey.None);
+        if (!(key instanceof PartitionKey)) {
```
moderakh: if the intention is to only capture PartitionKey.None then we should probably do PartitionKey.None == key instead.

Are there any other case other than PartitionKey.None which would get covered here?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncItemResponse.java
```diff
@@ -4,40 +4,41 @@
...
-        String bodyAsString = response.getBodyAsString();
-        if (StringUtils.isEmpty(bodyAsString)) {
+        this.itemClassType = klass;
+        responseBodyString = response.getBodyAsString();
+        if (StringUtils.isEmpty(responseBodyString)){
             super.setProperties(null);
-            itemClient = null;
         } else {
-            CosmosItemProperties props = new CosmosItemProperties(bodyAsString);
+            CosmosItemProperties props = new CosmosItemProperties(responseBodyString);
```
moderakh: If the public API is `T getResource()` the instantiation of `CosmosItemProperties` will do double deserialization. We should remember to remove this altogether and solely rely on `T getResource()`

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseValidator.java
```diff
@@ -210,7 +210,8 @@ public void validate(CosmosAsyncStoredProcedureResponse resourceResponse) {
                 @Override
                 public void validate(T resourceResponse) {
                     assertThat(resourceResponse.getProperties()).isNotNull();
-                    assertThat(resourceResponse.getProperties().getETag()).isNotNull();
+                    //TODO: Uncomment this after checking getETag()
+//                    assertThat(resourceResponse.getETag()).isNotNull();
```
moderakh: what happened to this guy?

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/DocumentCrudTest.java
```diff
@@ -315,17 +335,121 @@ public void upsertDocument_ReplaceDocument(String documentId) throws Throwable {
...
 
-        CosmosResponseValidator<CosmosAsyncItemResponse> validator = new CosmosResponseValidator.Builder<CosmosAsyncItemResponse>()
+        CosmosResponseValidator<CosmosAsyncItemResponse<CosmosItemProperties>> validator = new CosmosResponseValidator.Builder<CosmosAsyncItemResponse<CosmosItemProperties>>()
             .withProperty("newProp", newPropValue).build();
 
         validateSuccess(readObservable, validator);
     }
 
+    @Test(groups = {"simple"}, timeOut = TIMEOUT)
+    public void typedItemsTest() throws Throwable {
```
moderakh: to "Test" suffix or prefix in the test name to look the same as other test methods

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
moderakh: a response has more than just user data, it has activityId, request-charge, etc.

Although returning Mono.empty() may look nice for the end-user who is only interested in user-data, but doesn't provide the the above response header for more advanced user who is interested in response headers.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
moderakh: I don't think this specific case may have any impact on perf as it has to be done somewhere. you are not removing the need for invoking `single()` you are moving it from here to `AsyncDocumentClient`.

and yes ideally it has to be in `AsyncDocumentClient` for cleanness although not directly related to this PR. IMO can be done separately as a good improvement in a different PR.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
mbhaskar: null is a valid partitionkey. Also, please note that null partitionkey and PartitionKey.NONE are totally different

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
mbhaskar: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncItemResponse.java
mbhaskar: Handled the null/empty case. Returning null

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosResponse.java
mbhaskar: Removed

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/Utils.java
mbhaskar: Will check and do the refactoring as required in a different PR as this is not directly related to this.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncWriteBenchmark.java
```diff
@@ -60,13 +62,15 @@ protected void hookOnError(Throwable throwable) {
...
+        Mono<CosmosAsyncItemResponse<PojoizedJson>> obs;
         if (configuration.isDisablePassingPartitionKeyAsOptionOnWrite()) {
             // require parsing partition key from the doc
             obs = cosmosAsyncContainer.createItem(generateDocument(partitionKey, dataFieldValue));
         } else {
             // more optimized for write as partition ke is already passed as config
-            obs = cosmosAsyncContainer.createItem(generateDocument(partitionKey, dataFieldValue), new CosmosItemRequestOptions(partitionKey));
+            obs = cosmosAsyncContainer.createItem(generateDocument(partitionKey, dataFieldValue),
+                                                  new PartitionKey(partitionKey),
+                                                  new CosmosItemRequestOptions());
```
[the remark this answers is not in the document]
mbhaskar: We don't have an overload matching that. Will add if required in next PR.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
mbhaskar: Forgot to remove the setPartitionKey(). Made it package private now.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosItemRequestOptions.java
mbhaskar: Removed

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/PartitionKey.java
mbhaskar: Removed this check altogether. Will add if we see many people using this wrongly.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncItemResponse.java
mbhaskar: Yes, next PR should be for removing CosmosItemProperties as a public type.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/DocumentCrudTest.java
mbhaskar: Sure, always overlook this :-). Removed

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosResponseValidator.java
mbhaskar: Left over from a trial of removing CosmosItemProperties. Got it back.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainer.java
mbhaskar: Can live without it too, but left it for clarity. Will check and remove as required.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainer.java
mbhaskar: Done

sdk/cosmos/azure-cosmos-examples/src/main/java/com/azure/cosmos/examples/BasicDemo.java
mbhaskar: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
mbhaskar: Thanks for the response @moderakh. @kushagraThapar  make a note of this point, and we can discuss further if required and create a workitem for this.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
mbhaskar: +1. We can do it in a separate PR if necessary.

mbhaskar:
The failing test passed locally, so merging.

PR #7430, opened 2020-01-15
Fix cancellation handling

mitchdenny:
This PR extends/adjusts conditions on tasks and jobs to better handle cancellations. We have some tasks in jobs with an ```always()``` condition which effectively stops that job being cancelled. We need to avoid that in most cases because we generally need to cancel jobs in an build storm and ```always()``` conditions slow down recovery time.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -145,7 +145,7 @@ jobs:
       - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
         # ComponentGovernance is currently unable to run on pull requests of public projects. Running on non-PR
         # builds should be sufficient.
-        condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
+        condition: and(succeededOrFailed(), ne(variables['Build.Reason'], 'PullRequest'))
         displayName: 'Component Detection'
 
       # We `install` the code quality reports tooling into our local m2 cache separately from building the Maven project
@@ -311,7 +311,7 @@ jobs:
         condition: and(succeeded(), or(ne(variables['TestFromSource'],'true'), eq(variables['ShouldRunSourceTests'],'true')))
 
       - task: PublishTestResults@2
-        condition: and(succeededOrFailed(), or(ne(variables['TestFromSource'],'true'), eq(variables['ShouldRunSourceTests'],'true')))
+        condition: and(always(), or(ne(variables['TestFromSource'],'true'), eq(variables['ShouldRunSourceTests'],'true')))
         inputs:
           mergeTestResults: true
           testRunTitle: $(RunTitle)
--- a/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
@@ -113,7 +113,7 @@ jobs:
 
 
       - task: PublishTestResults@2
-        condition: succeededOrFailed()
+        condition: always()
         inputs:
           mergeTestResults: true
           testRunTitle: 'Live tests for ${{ parameters.ServiceDirectory }} $(DisplayName)'


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -312,7 +312,7 @@ jobs:
         condition: and(succeeded(), or(ne(variables['TestFromSource'],'true'), eq(variables['ShouldRunSourceTests'],'true')))
 
       - task: PublishTestResults@2
-        condition: and(always(), or(ne(variables['TestFromSource'],'true'), eq(variables['ShouldRunSourceTests'],'true')))
+        condition: and(succeededOrFailed(), or(ne(variables['TestFromSource'],'true'), eq(variables['ShouldRunSourceTests'],'true')))
```
mikeharder: I would prefer to keep `PublishTestResults` as `always()`.  This task should only take a few seconds, and when cancelling a build (say due to a hung test) it's valuable to publish the test results we do have so far.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -146,7 +146,7 @@ jobs:
       - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
         # ComponentGovernance is currently unable to run on pull requests of public projects. Running on non-PR
         # builds should be sufficient.
-        condition: and(succeededOrFailed(), ne(variables['Build.Reason'], 'PullRequest'))
+        condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
```
mikeharder: Why skip ComponentGovernance if an earlier task failed (say Verify Readmes)?  Usually with analyze/test tasks, it's better to run as much as possible even if an earlier step failed.  I agree it should not be `always()` since we want to skip on cancel, but I think `succeededOrFailed()` is better than just `succeeded()`.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -119,7 +119,7 @@ jobs:
         artifact: repository
 
   - job: 'Analyze'
-    condition: ne(variables['Skip.Analyze'], 'true')
+    condition: and(succeeded(), ne(variables['Skip.Analyze'], 'true'))
```
mikeharder: I believe all the jobs are run in parallel, so would there be any difference between `succeeded()` and `succeededAndFailed()` here?  If not then `succeeded()` is fine with me.

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -107,13 +107,13 @@ jobs:
...
             -Verbose
-          condition: ne(variables['AZURE_RESOURCEGROUP_NAME'], '')
+          condition: and(always(), ne(variables['AZURE_RESOURCEGROUP_NAME'], ''))
           continueOnError: true
           displayName: Remove Test Resources
 
 
       - task: PublishTestResults@2
-        condition: always()
+        condition: succeededOrFailed()
```
mikeharder: Same comment as above.  I would prefer to keep `always()` for `PublishTestResults`.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mitchdenny: Using ```always()``` will result in the job ignoring cancellation requests. Perhaps ```succeededOrFailed()``` might be more appropriate?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mitchdenny: I tried to apply a general rule as I was going through which was if a custom condition was defined make sure that a job state condition was also applied. I do actually plan on adding a prep job to all build stages in all pipelines which everything will depend on, so I kinda had that in the back of my head.

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
mitchdenny: ```succeededOrFailed()``` is the same as ```always()``` except that ```always()``` will keep the job running in the event of manual cancellation. When we manually cancel I am assuming we are expressly want the job cancelled now. I don't see the value of partial test results when we explicitly cancel.

mikeharder: The value is this scenario.  Someone notices a test is hung, and would like to manually cancel to stop wasting agent time.  However, they would still like to publish the test results they have so far.  I agree this is a relatively rare scenario, however the cost of delaying cancellation a few seconds (21 seconds for a recent run of `java - storage - tests`) is also low.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mikeharder: I assume you mean the `PublishTestResults` **task** will ignore cancellation, not the whole `Test` job, right?

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
mitchdenny: That's fair enough. Given this is only for the ```- tests``` runs then I guess it won't be a big issue.

mitchdenny: I'll update the PR, unless you see any other issues we should try to get this merged.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mitchdenny: My thinking here was that if we've got a failed job, the validity of the component scan could be invalid anyway, so why bother doing it. But happy to do ```succeededOrFailed()``` in this case since it won't run if we cancel.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mitchdenny: If the job has started then it'll run ```PublishTestResults``` with ```always()```. If the job hasn't started then the task won't run. Its probably not a big deal, if the job hasn't started then it won't run which will help with recovery, and provided we always put a ```succeeded()``` or ```succeededOrFailed()``` on preceding steps where a condition is used then it probably won't be a big issue.

I think you've convinced me ;)

mikeharder: Sounds good.  If this does end up causing bigger problems with cancellation, I agree it's not worth the benefit and should be changed to `succeededOrFailed()`.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -253,7 +253,7 @@ jobs:
 
       - task: Maven@3
         displayName: 'Start Jetty'
-        condition: ne('${{ parameters.SDKType }}', 'client')
+        condition: and(succeeded(), ne(variables['SdkType'], 'client'))
```
JimSuplizio: @mitchdenny why was this changed from parameters back to variables? Isn't variables going to try and pull from the pipeline variable SDKType? Using the parameters should, in theory, guarantee that this only runs for the correct track.

mitchdenny: My bad ... I'll fix it up now. I was getting shown some interesting conflict resolutions when I did my rebase.

mitchdenny:
> I had a question about the Start Jetty task the change from '${{ parameters.SDKType }}' back to variables['SdkType'] in the conditions.

Fixed up in this PR: https://github.com/Azure/azure-sdk-for-java/pull/7512

PR #7476, opened 2020-01-15
Fix Upload From File Tests

alzimmermsft:
This PR is a follow-up to #7403. 

Removed the use of `FluxUtil.writeFile` to write the blob download to disk based on the root cause found in PR 7403.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
@@ -36,11 +36,8 @@ import spock.lang.Requires
 import spock.lang.Unroll
 
 import java.nio.ByteBuffer
-import java.nio.channels.AsynchronousFileChannel
 import java.nio.charset.StandardCharsets
 import java.nio.file.Files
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
```diff
@@ -651,19 +648,20 @@ class BlockBlobAPITest extends APISpec {
...
-            StandardOpenOption.WRITE)
-        StepVerifier.create(FluxUtil.writeFile(blobAsyncClient.download(), outChannel)).verifyComplete()
-        outChannel.close()
+        def outFile = new File(file.getPath().toString() + "result")
+        outFile.createNewFile()
 
-        compareFiles(file, new File(outFile), 0, fileSize)
+        def outStream = new FileOutputStream(outFile)
+        outStream.write(FluxUtil.collectBytesInByteBufferStream(blobAsyncClient.download()).block())
+        outStream.close()
```
srnagar: This can be simplified by using try-with-resources pattern:

```
new FileOutputStream(outFile).with { outStream ->
      outStream.write(FluxUtil.collectBytesInByteBufferStream(blobAsyncClient.download()).block());
}
```

PR #7478, opened 2020-01-15
Post release version update to 3.7.0-beta.1

kushagraThapar:


sdk/cosmos/README.md
```diff
@@ -92,7 +92,7 @@ For example, using maven, you can add the following dependency to your maven pom
 <dependency>
   <groupId>com.microsoft.azure</groupId>
   <artifactId>azure-cosmos</artifactId>
-  <version>3.6.0</version>
+  <version>3.7.0-beta.1</version>
```
JimSuplizio: So, the Java team hasn't been updating the readme versions from the released version until they're about to release that version (it's the same script command with the --sr option which means skip readme). The only reason they do this is because it apparently caused some confusion as some customers go to the repro for the docs. I'm fine with whatever you guys decide.

kushagraThapar: Good point, I think its better to have released version in there, I will update it manually for now, in the future, will use `--sr` option when running the script.

Issue #5715, opened 2019-10-08
[FEATURE REQ] BlobOutputStream Encryption


rickle-msft:
BlobOutputStream does not currently support encryption, though BlobInputStream does. The implementation of BlobOutputStream added additional complexity to supporting encryption, so we deferred this to a later release.

PR #7485, opened 2020-01-15
blob output stream encryption

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/5715

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobClient.java
@@ -139,20 +139,20 @@ public final class BlockBlobClient extends BlobClientBase {
         BlobHttpHeaders headers, Map<String, String> metadata, AccessTier tier,
         BlobRequestConditions requestConditions) {
 
-        BlobAsyncClient blobClient = prepareBuilder(client).buildAsyncClient();
+        BlobAsyncClient blobClient = prepareBuilder().buildAsyncClient();
 
         return BlobOutputStream.blockBlobOutputStream(blobClient, parallelTransferOptions, headers, metadata, tier,
             requestConditions);
     }
 
-    private BlobClientBuilder prepareBuilder(BlobAsyncClientBase client) {
+    private BlobClientBuilder prepareBuilder() {
         BlobClientBuilder builder = new BlobClientBuilder()
-            .pipeline(client.getHttpPipeline())
-            .endpoint(client.getBlobUrl())
-            .snapshot(client.getSnapshotId())
-            .serviceVersion(client.getServiceVersion());
+            .pipeline(getHttpPipeline())
+            .endpoint(getBlobUrl())
+            .snapshot(getSnapshotId())
+            .serviceVersion(getServiceVersion());
 
-        CpkInfo cpk = client.getCustomerProvidedKey();
+        CpkInfo cpk = getCustomerProvidedKey();
         if (cpk != null) {
             builder.customerProvidedKey(new CustomerProvidedKey(cpk.getEncryptionKey()));
         }


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClient.java
```diff
@@ -56,6 +59,59 @@
         this.encryptedBlobAsyncClient = encryptedBlobAsyncClient;
     }
 
+    /**
+     * Creates and opens an output stream to write data to the block blob. If the blob already exists on the service, it
```
gapra-msft: TODO : Change java doc

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
```diff
@@ -40,7 +40,7 @@ static BlobOutputStream appendBlobOutputStream(final AppendBlobAsyncClient clien
         return new AppendBlobOutputStream(client, appendBlobRequestConditions);
     }
 
-    static BlobOutputStream blockBlobOutputStream(final BlockBlobAsyncClient client,
+    public static BlobOutputStream blockBlobOutputStream(final BlobAsyncClient client,
```
gapra-msft: I had to make this public since the crypto package needs access. I dont see a problem doing this.

gapra-msft: should I rename this to blobOutputStream or is that confusing?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClient.java
```diff
@@ -56,6 +58,59 @@
...
+     *
+     * @return A {@link BlobOutputStream} object used to write data to the blob.
+     * @throws BlobStorageException If a storage service error occurred.
+     */
+    public BlobOutputStream getBlobOutputStream() {
+        return getBlobOutputStream(false);
+    }
+
+    /**
+     * Creates and opens an output stream to write data to the block blob. If the blob already exists on the service, it
```
rickle-msft: I don't think the second sentence is relevant here since we accept the overwrite flag on this overload.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClient.java
rickle-msft: Just clarifying that you are changing the Javadoc to say that it does not overwrite by default?

gapra-msft: This is straight copied from block blob. I thought I'd change it to say encrypted blob but the rest of the docs say block blob so I'll probably leave it that way

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClient.java
gapra-msft: I'll change it here and on the block blob doc

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClient.java
```diff
@@ -56,6 +58,57 @@
...
+     * @return A {@link BlobOutputStream} object used to write data to the blob.
+     * @param overwrite Whether or not to overwrite, should data exist on the blob.
+     * @throws BlobStorageException If a storage service error occurred.
+     */
+    public BlobOutputStream getBlobOutputStream(boolean overwrite) {
+        BlobRequestConditions requestConditions = null;
+        if (!overwrite) {
+            if (exists()) {
+                throw logger.logExceptionAsError(new IllegalArgumentException(Constants.BLOB_ALREADY_EXISTS));
+            }
```
alzimmermsft: This case is interesting given that writing to the blob may be delayed a non-determinant amount of time, this differs from the other cases of `overwrite` where their write operation should happen relatively soon compared to the validation check.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobClient.java
```diff
@@ -137,10 +138,28 @@ public BlobOutputStream getBlobOutputStream(BlobRequestConditions requestConditi
...
         BlobRequestConditions requestConditions) {
-        return BlobOutputStream.blockBlobOutputStream(client, parallelTransferOptions, headers, metadata, tier,
+
+        BlobAsyncClient blobClient = prepareBuilder(client).buildAsyncClient();
+
+        return BlobOutputStream.blockBlobOutputStream(blobClient, parallelTransferOptions, headers, metadata, tier,
             requestConditions);
     }
 
+    private BlobClientBuilder prepareBuilder(BlobAsyncClientBase client) {
```
alzimmermsft: Could make this static as it access no instance level properties, on the flip side could remove the parameter and access all the values in the instance.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
alzimmermsft: Leave the name as-is, this stream is meant to solely interact with Block Blobs, don't want to conflate the overarching Blob term to mean Block Blob (though that is the primary use case).

alzimmermsft: I don't have a strong issue with it, we may want to split this and the other implementations into their own class files and make them public as well. I don't see a overwhelmingly strong reason why a customer couldn't pass a client in themselves to create the stream vs needing to get it from the client.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClient.java
gapra-msft: This makes sense to me, the only other way would be to pass it in to the outputstream and it would fail upon closing - so that would be potentially annoying for the customer

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobClient.java
gapra-msft: Just removed the param

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
gapra-msft: I can do this as part of another issue I'm working on with Append and Page Blob Output Stream

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClient.java
alzimmermsft: Agreed, failing fast is the better option here.

PR #7489, opened 2020-01-16
Moving aggregate reports out of the unified pipelines.

mitchdenny:
This PR moves the aggregate reports out of the Analyze stage and drops them into a separate pipeline which is setup to run on a schedule.

### DIFF
--- a/eng/pipelines/aggregate-reports.yml
+++ b/eng/pipelines/aggregate-reports.yml
@@ -27,6 +27,18 @@ jobs:
         goals: 'install'
 
     - task: Maven@3
+      displayName: 'Install Module and Run SpotBugs, Checkstyle, RevApi, and Javadoc'
+      inputs:
+        mavenPomFile: sdk/core/pom.service.xml
+        options: '--no-transfer-progress -DskipTests -Dgpg.skip -Dverify-readme'
+        mavenOptions: '$(MemoryOptions)'
+        javaHomeOption: 'JDKVersion'
+        jdkVersionOption: '1.11'
+        jdkArchitectureOption: 'x64'
+        publishJUnitResults: false
+        goals: 'install'
+
+    - task: Maven@3
       displayName: 'Generate Maven project site, including JavaDocs, SpotBugs, and CheckStyle reports'
       inputs:
         mavenPomFile: pom.client.xml


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -156,17 +156,17 @@ jobs:
...
-        inputs:
-          mavenPomFile: parent/pom.xml
-          options: '$(DefaultOptions) -DskipTests -Dgpg.skip'
-          mavenOptions: '$(MemoryOptions)  $(LoggingOptions)'
-          javaHomeOption: 'JDKVersion'
-          jdkVersionOption: '1.11'
-          jdkArchitectureOption: 'x64'
-          publishJUnitResults: false
-          goals: 'install'
+      # - task: Maven@3
```
alzimmermsft: This doesn't run aggregates or anything, was this removed since parent is less useful given @JimSuplizio's version management changes?

eng/pipelines/aggregate-reports.yml
```diff
@@ -1,6 +1,67 @@
...
+        options: '$(DefaultOptions) -DskipTests -Dgpg.skip'
+        mavenOptions: '$(MemoryOptions)  $(LoggingOptions)'
+        javaHomeOption: 'JDKVersion'
+        jdkVersionOption: '1.11'
+        jdkArchitectureOption: 'x64'
+        publishJUnitResults: false
+        goals: 'install'
+
+    - task: Maven@3
+      displayName: 'Generate Maven project site, including JavaDocs, SpotBugs, and CheckStyle reports'
```
alzimmermsft: @JonathanGiles, should we skip SpotBugs and Checkstyles here given master would be in an invalid state of they failed? Another option would be to continue running them in this step but turn failures off to generate, may want to do that whole sale for Javadocs and RevApi as well here.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mitchdenny: It seemed redundant to me. Do we need to install it? Given we find the parent POMs via links (or via Maven Central) won't they end up getting installed anyway?

eng/pipelines/aggregate-reports.yml
```diff
@@ -1,6 +1,67 @@
...
+      inputs:
+        mavenPomFile: pom.client.xml
+        options: '$(DefaultOptions) -DskipTests -Dgpg.skip'
+        mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
+        javaHomeOption: 'JDKVersion'
+        jdkVersionOption: '1.11'
+        jdkArchitectureOption: 'x64'
+        publishJUnitResults: false
+        goals: 'install site:site site:stage'
```
srnagar: Don't need `install` here. `site:site site:stage` will generate all 3 reports.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
JimSuplizio: @alzimmermsft the parent still contains some plugin definitions. It's unclear if we can simply remove it but maybe we should look at if we can. I'd love to have less 'stuff' in our repo.

alzimmermsft: Ah, didn't realize either of these!

eng/pipelines/aggregate-reports.yml
mitchdenny: Seems we do need it after some experimentation :-)

PR #7495, opened 2020-01-16
Migration: Java perf framework  + Storage perf tests

g2vinay:
Phase 1:Migrates the perf framework and storage tests as it is, and preserves version history.

Phase 2: 
In a follow up PR:  (Captured in this issue): #7496 
1. Wire the framework and tests to master build.
2. Fix checksyle issues
3. Fix spotbugs issues
3. Add javadocs to the framework,
4. Comply with track 2 java guidelines.

Doing this in two phases to make it easier to review things

### DIFF
--- a/sdk/.gitignore
+++ b/sdk/.gitignore
@@ -1,0 +1,64 @@
+*.class
+
+#Javadoc files
+dist/
+.publish/
+
+#External libs
+extlib/
+
+# Auth files
+*.auth
+*.azureauth
+
+# Local checkstyle
+*.checkstyle
+
+# Mobile Tools for Java (J2ME)
+.mtj.tmp/
+
+# Package Files #
+*.jar
+*.war
+*.ear
+
+# Azure Tooling #
+node_modules
+packages
+
+# Eclipse #
+*.pydevproject
+.project
+.metadata
+bin/**
+tmp/**
+tmp/**/*
+*.tmp
+*.bak
+*.swp
+*~.nib
+local.properties
+.classpath
+.settings/
+.loadpath
+
+# Other Tooling #
+.classpath
+.project
+target
+bin
+.idea
+*.iml
+
+# Mac OS #
+.DS_Store
+.DS_Store?
+
+# Windows #
+Thumbs.db
+
+# reduced pom files should not be included
+dependency-reduced-pom.xml
+package-lock.json
+
+.vscode
--- a/sdk/core/azure-perfstress/pom.xml
+++ b/sdk/core/azure-perfstress/pom.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" 
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.azure</groupId>
+  <artifactId>azure-perfstress</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.fasterxml.jackson</groupId>
+        <artifactId>jackson-bom</artifactId>
+        <version>2.10.0</version>
+        <scope>import</scope>
+        <type>pom</type>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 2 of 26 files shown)


g2vinay:
@mikeharder 
This PR migrates the perf code + storage perf tests as is.
It does include the java 8 compatibility changes from this PR: https://github.com/mikeharder/StoragePerf/pull/8#issuecomment-574936694

So, you can review these two PRs at the same time.

sdk/core/azure-perfstress/src/main/java/com/azure/perfstress/SleepTest.java
```diff
@@ -0,0 +1,40 @@
...
+        _secondsPerOperation = Pow(2, instanceCount);
+    }
+
+    private static int Pow(int value, int exponent) {
+        int power = 1;
+        for (int i=0; i < exponent; i++) {
+            power *= value;
+        }
+        return power;
+    }
```
srnagar: Why not use [Math.pow()](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#pow-double-double-) method instead of implementing own power method?

sdk/storage/perf-test/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.azure</groupId>
+  <artifactId>azure-storage-blob-perfstress</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.compiler.source>1.8</maven.compiler.source>
```
srnagar: Are all the perf tests pinned to JDK 8 or can we have two profiles to run these tests for both JDK 8 and 11?

sdk/.gitignore
```diff
@@ -0,0 +1,64 @@
+*.class
```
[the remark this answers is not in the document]
g2vinay: I will be removing this file in the follow up PR of Phase 2.
If its mandatory that this file, should not go in, I can remove it in this PR too.

sdk/core/azure-perfstress/pom.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" 
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.azure</groupId>
+  <artifactId>azure-perfstress</artifactId>
```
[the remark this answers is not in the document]
g2vinay: Yeah, this PR doesn't manipulate the files and codebase.
It will happen as discussed in Phase 2 PR.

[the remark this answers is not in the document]
mikeharder: I chose the name `perfstress` because this is a framework for both perf and stress testing.  But feel free to rename if you want.

sdk/core/azure-perfstress/src/main/java/com/azure/perfstress/SleepTest.java
g2vinay: Yeah, we can use Math.pow(),
unless @mikeharder had any concerns that this might create any performance impact or inconsistencies across languages.

sdk/storage/perf-test/pom.xml
g2vinay: I don't see any issues of having two profiles.
Code is compatible with Java 8 and java 11.
Will add the setup for Java 8 and Java 11 profiles in the Phase 2 PR.

sdk/core/azure-perfstress/src/main/java/com/azure/perfstress/SleepTest.java
mikeharder: I was afraid of the conversion between `double` and `int`.  But if you know this is safe you can replace with `Math.pow()`.

sdk/storage/perf-test/pom.xml
mikeharder: Ideally we should be able to run perf tests on all supported JDK versions.  My Java expertise is limited so if anything looks wrong in the `pom.xml` files it probably is wrong :).

sdk/.gitignore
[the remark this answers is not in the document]
srnagar: If you are planning on removing it later, then it's fine but don't want to have multiple places having this .gitignore configuration.

sdk/core/azure-perfstress/pom.xml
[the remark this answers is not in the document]
srnagar: Since it's not going to be released, I am not too worried about the name as it can be changed later too.

sdk/core/azure-perfstress/src/main/java/com/azure/perfstress/SleepTest.java
srnagar: Math.pow() works for this usecase, we can replace it

sdk/storage/perf-test/pom.xml
srnagar: @mikeharder the pom.xml file looks correct, just wanted to ensure that we can support both JDK 8 and 11. We can add two profiles like @g2vinay said to support both.

PR #7498, opened 2020-01-16
Fix sdk/core dev package builds

mitchdenny:
This PR attempts to resolve an issue that has cropped up with our scheduled nightly builds of ```sdk/core```. As a reminder ```sdk/core/pom.service.xml``` references all track 2 modules and is in effect a monolithic build for our repo (in order to test the impact of core changes on all its dependents).

A problem is occuring because ```sdk/appconfiguration``` depends on an unreleased version of ```com.azure:azure-core``` which it expects to be able to resolve from the repository, but it cannot, because due to our versioning scripts, the version number of ```com.azure:azure-core``` has had the ```.dev.yyyyMMdd.x``` qualifer appended to it.

This is effectively a binary composition/source composition problem that occurs when we dynamically change package versions and don't update the dependency list of downstream dependents.

This change (if it works) aims to push the daily dev feed version numbers for ```com.azure:azure-core``` and friends down into the dependents so that they can build. In theory this should be safe because we shouldn't break backwards compatibility.

@JimSuplizio this was the issue that I pinged you about.

@conniey even if this does fix the issue I don't believe that we publish the event hubs packages from this build into the dev package feed from the ```java - core``` build. You would need to update ```sdk/eventhubs``` to take a dependency of the dev feed version, merge that it and run a dev feed build for the ```java - eventhubs``` pipeline.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -73,14 +73,14 @@ jobs:
       - ${{ each artifact in parameters.Artifacts }}:
         - script: |
             python3 --version
-            python3 eng/versioning/set_versions.py --build-type ${{parameters.SDKType}} --build-qualifier dev.$(Build.BuildNumber) --artifact-id ${{artifact.name}} --group-id ${{artifact.groupId}}
-          condition: eq(variables['SetDevVersion'],'true')
+            python3 eng/versioning/set_versions.py --build-type ${{parameters.SDKType}} --update-type all --build-qualifier dev.$(Build.BuildNumber) --artifact-id ${{artifact.name}} --group-id ${{artifact.groupId}}
+          #condition: eq(variables['SetDevVersion'],'true')
           displayName: Append dev package version suffix for ${{artifact.name}}
 
       - script: |
           python3 --version
-          python3 eng/versioning/update_versions.py --update-type library --build-type ${{parameters.SDKType}}
-        condition: eq(variables['SetDevVersion'],'true')
+          python3 eng/versioning/update_versions.py --update-type all --build-type ${{parameters.SDKType}}
+        #condition: eq(variables['SetDevVersion'],'true')
         displayName: Apply version settings to repository
 
       - script: |


conniey:
@mitchdenny It's fine. If I explicitly choose in my dependencies to use a dev feed of azure-core-amqp, since it's higher in the dependency tree than the one taken by azure-messaging-eventhubs, it'll win.

this is what we use in our stress tests at the moment.
```xml
        <dependency>
            <groupId>com.azure</groupId>
            <artifactId>azure-core-amqp</artifactId>
            <version>[1.0.1-dev.20200107.12,)</version>
        </dependency>

        <dependency>
            <groupId>com.azure</groupId>
            <version>[5.0.1-dev.20200107.7,)</version>
        </dependency>
```

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -73,14 +73,14 @@ jobs:
       - ${{ each artifact in parameters.Artifacts }}:
         - script: |
             python3 --version
-            python3 eng/versioning/set_versions.py --build-type ${{parameters.SDKType}} --build-qualifier dev.$(Build.BuildNumber) --artifact-id ${{artifact.name}} --group-id ${{artifact.groupId}}
-          condition: eq(variables['SetDevVersion'],'true')
+            python3 eng/versioning/set_versions.py --build-type ${{parameters.SDKType}} --update-type all --build-qualifier dev.$(Build.BuildNumber) --artifact-id ${{artifact.name}} --group-id ${{artifact.groupId}}
+          #condition: eq(variables['SetDevVersion'],'true')
```
JimSuplizio: @mitchdenny I'm not sure exactly what you're trying to do but this change isn't correct. The update-type argument is **library**, **external_dependency** or **all** (probably should have been _both_ instead of _all_). set_versions is a script that **only** manipulates library versions which means that the update-type argument was completely unnecessary and was removed from the script. Similarly, the call below to update_versions, was correct with the update_type of library since, the previous set_versions is only ever manipulating library versions.

I'm assuming the commented out conditions were for testing purposes and not part of this PR?

JimSuplizio:
@mitchdenny @conniey I'm going to close this PR, I know what the problem is and I'll submit another PR to fix it.

mitchdenny:
Yep - was just about to do the same.

PR #7501, opened 2020-01-16
Fix set version script for unreleased dependencies for dev feed

JimSuplizio:
When the changes were made to the set_version script to incorporate the groupId the condition for unreleased dependencies wasn't appropriately updated.


mitchdenny:
Thanks for jumping on this @JimSuplizio. Can we test that the build stage works correctly by removing the conditions on the version setting scripts (then put them back prior to merging). I just want to make sure that core, and one of the other pipelines run fine with this change (just to simulate what is going to happen under scheduled build conditions).

JimSuplizio:
@mitchdenny This builds locally with my changes (without them it would fail the same way the pipeline did). I don't think the pipeline is necessary here.

PR #7507, opened 2020-01-16
Updated storage library versions post release

gapra-msft:


sdk/storage/azure-storage-blob-nio/pom.xml
```diff
@@ -55,7 +55,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-storage-blob</artifactId>
-      <version>12.2.0</version> <!-- {x-version-update;com.azure:azure-storage-blob;dependency} -->
+      <version>12.3.0</version> <!-- {x-version-update;com.azure:azure-storage-blob;dependency} -->
```
JimSuplizio: This should be **current**, not **dependency**. Libraries built within the same pipeline should have dependencies on the current version of other libraries within that pipeline. This appears to be the only discrepancy in storage.

gapra-msft: @JimSuplizio Just changed it

PR #7510, opened 2020-01-16
Retry network operations in resources scripts

heaths:
Relates to azure/azure-sdk-tools#326 and other issues.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
@@ -82,7 +82,7 @@ jobs:
             -ProvisionerApplicationId '$(provisioner-aad-id)'
             -ProvisionerApplicationSecret '$(provisioner-aad-secret)'
             -DeleteAfterHours 24
-            -Location westus2
+            -CI
             -Force
             -Verbose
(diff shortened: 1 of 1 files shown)


heaths:
@danieljurek the failure seems to be unrelated, but I'd hate for it to go unchecked:

https://dev.azure.com/azure-sdk/public/_build/results?buildId=236312&view=logs&j=2714c728-d141-5cd1-e4fc-49513190fa65&t=4da905a1-2f38-556a-c21b-c51eb108a69d&l=33061

heaths:
Live tests aren't reporting back status like in other repos for some reason, but the only failure(s) seem to be unrelated and most often not even connecting to Azure, though are numerous: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=236460&view=logs&j=351a4efb-edfe-5bb4-cda1-1ace5fc54b68&t=59f8b7da-4461-55d1-d907-5626116ef0aa&l=2716. Other platform/JRE combinations passed. The "Windows - Java 8" job also appears like it will time out, but that's also not uncommon from past runs, it seems.

heaths:
Had a Java dev look over and agreed none of this would cause the failures we're seeing in the live tests.

PR #6161, opened 2019-11-05
Text analytics Preview 1 

mssfang:
Initial PR for Text Analytics client including samples:
- Added authentication for client builder with AAD and subscription key support.
- Added implementation to Detect Language, Entity, Pii Entity, Linked Entity, Sentiment endpoints.
- Added an initial test framework for text analytics.
- Added samples, tests for Detect Language, Entity, Pii Entity, Linked Entity, Sentiment endpoints.
- Fixed CheckStyle and Spotbugs errors.
- Added README but not fully documented and JavaDoc
- Use the latest swagger

Fixes  #6499, #6498, #6497, #6496, #6495, #6494, #6500, #6501, #6502, #6505, #6507, #6565 , #6564, #6048, #6726, #6728, #6851, #6887,#6870,#6876,#6881,

PR #7514, opened 2020-01-17
Refactor test methods for Text Analytics

### LINKED ISSUE #6966: [Test] Too many helper methods in the tests. Refactor the test framework.
Refactor to reduce the number of helper methods

samvaity:
Closes #6966

Ref https://github.com/Azure/azure-sdk-for-java/pull/6161/files#r359106192

### DIFF
--- a/sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TestUtils.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TestUtils.java
@@ -24,6 +24,8 @@ import com.azure.ai.textanalytics.models.TextSentimentClass;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Contains helper methods for generating inputs for test methods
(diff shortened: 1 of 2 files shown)


sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTestBase.java
```diff
@@ -254,468 +251,179 @@
     abstract void analyseSentimentForListLanguageHint();
 
     // Detect Language runner
-    void detectLanguageShowStatisticsRunner(BiConsumer<List<DetectLanguageInput>,
+    static void detectLanguageShowStatisticsRunner(BiConsumer<List<DetectLanguageInput>,
```
mssfang: What reason to make runner a static method?

samvaity: here, just to keep consistent with other methods in the file.

mssfang: We should make it non-static. I think make the method static will not be safe if we want the tests parallel.

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TestUtils.java
```diff
@@ -0,0 +1,252 @@
...
+import com.azure.ai.textanalytics.models.TextSentimentClass;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Contains helper methods for generating inputs for test methods
+ */
+public final class TestUtils {
```
conniey: Doesn't look like this needs to be public.

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTestBase.java
```diff
@@ -857,156 +561,113 @@ static void validateAnalysedSentiment(TextSentiment expectedSentiment, TextSenti
...
-    static DocumentResultCollection<RecognizeEntitiesResult> getExpectedBatchNamedEntities() {
-        NamedEntity namedEntity1 = new NamedEntity("Seattle", "Location", null, 26, 7, 0.80624294281005859);
-        NamedEntity namedEntity2 = new NamedEntity("last week", "DateTime", "DateRange", 34, 9, 0.8);
-        NamedEntity namedEntity3 = new NamedEntity("Microsoft", "Organization", null, 10, 9, 0.99983596801757812);
+        expected.forEach((key, expectedValue) -> {
+            T actualValue = actual.get(key);
 
-        List<NamedEntity> namedEntityList1 = Arrays.asList(namedEntity1, namedEntity2);
-        List<NamedEntity> namedEntityList2 = Collections.singletonList(namedEntity3);
+            assertTrue(actual.containsKey(key));
```
conniey: This isn't needed. If the key doesn't exist, then the value returned will be null.

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTestBase.java
```diff
@@ -7,6 +7,7 @@
 import com.azure.ai.textanalytics.models.DetectLanguageInput;
```
conniey: Consider setting StepVerifier.defaultTimeout so that it is not infinite. And resetting it when the test class finishes.

samvaity: Updated.

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TestUtils.java
```diff
@@ -0,0 +1,250 @@
...
+        List<DetectLanguageResult> detectLanguageResultList = Arrays.asList(detectLanguageResult1, detectLanguageResult2, detectLanguageResult3);
+
+        return new DocumentResultCollection<>(detectLanguageResultList, DEFAULT_MODEL_VERSION, textDocumentBatchStatistics);
+    }
+
+    /**
+     * Helper method to get the expected Batch Named Entities
+     */
+    static DocumentResultCollection<RecognizeEntitiesResult> getExpectedBatchNamedEntities() {
+        NamedEntity namedEntity1 = new NamedEntity("Seattle", "Location", null, 26, 7, 0.80624294281005859);
```
mssfang: We should remove all the score equal comparison. Currently. the nightly live test failed because of it.
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=242950&view=logs&j=4d5db6ce-0b7f-527e-b115-2367ee6e1fef&t=b7d16dfc-4abf-5ff2-eaa7-e82f3df2ef1b

samvaity: Updated to not check for equality for offset/length/score properties on models.

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TestUtils.java
```diff
@@ -0,0 +1,250 @@
...
+    static List<DetectLanguageInput> getDetectLanguageInputs() {
+        return Arrays.asList(
+            new DetectLanguageInput("0", DETECT_LANGUAGE_INPUTS.get(0), "US"),
+            new DetectLanguageInput("1", DETECT_LANGUAGE_INPUTS.get(1)),
+            new DetectLanguageInput("2", DETECT_LANGUAGE_INPUTS.get(2), "US")
+        );
+    }
+
+    static List<TextDocumentInput> getTextDocumentNamedEntityInputs() {
+        return Arrays.asList(
```
conniey: This will be cumbersome to manage if you add another "named_entity_input" or add to those other lists. I would suggest mapping.

```java
        return IntStream.range(0, contents.size())
            .mapToObj(index -> {
                return new TextDocumentInput(String.valueOf(index), contents.get(index));
            }).collect(Collectors.toUnmodifiableList());
```

samvaity: Updated

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TestUtils.java
mssfang: should the score value need to clean up also?

mssfang: same question apply to other hard-coded numerical values

PR #7534, opened 2020-01-17
Clears closed/errored sessions and links

conniey:
* Fixes issues when a connection/session/link idles and is closed. This cleans up those old sessions and links.
* Updates log messages. There are multiple instances of some classes and it is hard to discern which one is for which.
* Fixes a bug where it would not wait for the reactor to finish scheduling remaining tasks before disposing of instance.


sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClientIntegrationTest.java
```diff
@@ -247,4 +252,41 @@ void worksAfterReconnection() throws InterruptedException {
         TimeUnit.MINUTES.sleep(30);
         System.out.println("Complete.");
     }
+
+    @Disabled("Testing idle/long running operations.")
```
srnagar: Would this be better suited in our stress test client since this is disabled and wouldn't run in any of our live tests?

conniey: Yes, it would for our new stress infrastructure. But I didn't want to manually have to fire up their giant infrastructure when I wanted to test a change. :( I'll create an issue to move these tests.

PR #7538, opened 2020-01-17
Only run Dependency Analyzer for client builds

JimSuplizio:
This is an interim fix as I'm working on spitting up the client/data jobs. Right now the dependency analysis is causing the data track builds to fail and since this task is only generating a report for client, it shouldn't be running for anything other than SDKType client builds.


JimSuplizio:
Closing this pull request. This issue will be fixed as part of https://github.com/Azure/azure-sdk-for-java/pull/7571

PR #7551, opened 2020-01-20
Expose buildUserAgent as protected

### LINKED ISSUE #7549: [FEATURE REQ] Need expose buildUserAgent in UserAgentPolicy
In Fluent Java, we use a rest client for all package. Thus, we need to set different user agent in different request. https://github.com/Azure/azure-libraries-for-java/blob/vnext/azure-mgmt-resources/src/main/java/com/azure/management/UserAgentPolicy.java#L90
In order not to new a policy per request or pass the whole user agent parameter every time, we'd like to ask for expose the [buildUserAgent](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java#L137) function to protected or public.

Solution as above.

**Additional context**

**Information Checklist**
- [x] Description Added

ChenTanyi:
fix #7549

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/UserAgentPolicy.java
@@ -134,7 +134,7 @@ public class UserAgentPolicy implements HttpPipelinePolicy {
     /*
      * Builds the User-Agent header, at minimum this will create a User-Agent header with the DEFAULT_USER_AGENT_FORMAT.
      */
-    private static String buildUserAgent(String applicationId, String sdkName, String sdkVersion,
+    protected static String buildUserAgent(String applicationId, String sdkName, String sdkVersion,
         Configuration configuration) {
         StringBuilder userAgentBuilder = new StringBuilder();
 


alzimmermsft:
Mind offering an example of how this will be used, this will be promoting the private method into public API. Is there a lot of client build-up and tear-down that happens or is the `User-Agent` highly variable between calls?

Based on the outline of the issue have you looked into [AddHeadersFromContextPolicy](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AddHeadersFromContextPolicy.java)? This allows passing a custom set of `HttpHeaders` that will be added into a request, this could be used to pass a `User-Agent` header that is highly variable on a call-by-call basis.

weidongxu-microsoft:
@alzimmermsft 
In Java Fluent SDK, user can choose between using a single lib, with single HttpPipeline, or using a collected lib (for 10+ service), sharing a single HttpPlpeline. We would like different lib having different UserAgent, even when sharing the same HttpPipeline.

Here we wanted to reuse this piece of code, i.e. buildUserAgent, in UserAgentPolicy class if possible (reduce code duplication). Hence we are trying to inherit UserAgentPolicy, instead of using AddHeadersFromContextPolicy.

Let me know if you have any advice. Or if you'd prefer us to duplicate this code.

JonathanGiles:
There is a code smell here with a `protected static` method being introduced. I would rather this change not be made, and either the code extracted into a commonly-accessible utility method, or it simply be copied for the management case, for now. I would suggest further research be done to consider how better to achieve this, but also to consider falling back to copy / paste given the code is relatively simple and minimal.

alzimmermsft:
@weidongxu-microsoft @ChenTanyi 

Following up to get a resolution on this change proposal. Will this logic be copied to where it is needed or should we begin discussing another solution?

ChenTanyi:
@alzimmermsft yeah, we have copied it to where we need it. If you think it is better than use it from base class, I'll close this PR. Is it possible for the function to be public?

PR #7552, opened 2020-01-20
add support for AuthFileCredential

Luyunmt:
Added AuthFileCredential and its tests to azure-identity.
https://github.com/Azure/azure-sdk-for-net/pull/9312

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
@@ -76,6 +76,16 @@ public class Configuration implements Cloneable {
     public static final String PROPERTY_AZURE_TENANT_ID = "AZURE_TENANT_ID";
 
     /**
+     * Path of a PEM certificate file to use when performing service principal authentication with Azure.
+     */
+    public static final String PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH = "AZURE_CLIENT_CERTIFICATE_PATH";
+
+     /**
+     * Path of a authentication file to use when performing service principal authentication with Azure.
+     */
+    public static final String PROPERTY_AZURE_AUTH_LOCATION = "AZURE_AUTH_LOCATION";
+
+    /**
      * Name of the Azure resource group.
      */
     public static final String PROPERTY_AZURE_RESOURCE_GROUP = "AZURE_RESOURCE_GROUP";
@@ -115,6 +125,8 @@ public class Configuration implements Cloneable {
         PROPERTY_AZURE_CLIENT_ID,
         PROPERTY_AZURE_CLIENT_SECRET,
         PROPERTY_AZURE_TENANT_ID,
+        PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH,
+        PROPERTY_AZURE_AUTH_LOCATION,
         PROPERTY_AZURE_RESOURCE_GROUP,
         PROPERTY_AZURE_CLOUD,
         PROPERTY_AZURE_TELEMETRY_DISABLED,
--- a/sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
+++ b/sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
@@ -14,13 +14,32 @@ import com.azure.identity.implementation.IdentityClientOptions;
 import reactor.core.publisher.Mono;
 
 /**
- * A credential provider that provides token credentials based on environment variables.  The environment variables
+ * A credential provider that provides token credentials based on environment variables.  The sets of environment variables
  * expected are:
  * <p>
  * <ul>
  *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_ID AZURE_CLIENT_ID}</li>
  *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_SECRET AZURE_CLIENT_SECRET}</li>
  *     <li>{@link Configuration#PROPERTY_AZURE_TENANT_ID AZURE_TENANT_ID}</li>
+ * </ul>
+ * or:
+ * <p>
+ * <ul>
+ *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_ID AZURE_CLIENT_ID}</li>
+ *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH AZURE_CLIENT_CERTIFICATE_PATH}</li>
+ *     <li>{@link Configuration#PROPERTY_AZURE_TENANT_ID AZURE_TENANT_ID}</li>
+ * </ul>
+ * or:
+ * <p>
+ * <ul>
+ *     <li>{@link Configuration#PROPERTY_AZURE_CLIENT_ID AZURE_CLIENT_ID}</li>
+ *     <li>{@link Configuration#PROPERTY_AZURE_USERNAME AZURE_USERNAME}</li>
+ *     <li>{@link Configuration#PROPERTY_AZURE_PASSWORD AZURE_PASSWORD}</li>
+ * </ul>
+ * or:
+ * <p>
+ * <ul>
+ *     <li>{@link Configuration#PROPERTY_AZURE_AUTH_LOCATION AZURE_AUTH_LOCATION}</li>
  * </ul>
  */
 @Immutable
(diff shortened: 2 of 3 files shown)


sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
```diff
@@ -0,0 +1,120 @@
...
+     * Obtains a token from the Azure Active Directory service, using the specified client detailed specified in the SDK Auth file.
+     * This method is called by Azure SDK clients. It isn't intended for use in application code.  
+     * If the SDK Auth file is missing or invalid, this method throws a <see cref="AuthenticationFailedException"/> exception.
+     * <param name="requestContext">The details of the authentication request</param>
+     * <returns>An <see cref="AccessToken"/> which can be used to authenticate service client calls.</returns>
+     */
+    @Override
+    public Mono<AccessToken> getToken(TokenRequestContext request) {
+        try {
+            ensureCredential();
```
jianghaolu: I appreciate the effort trying to get the credential built regardless of the validity of the auth file. 
In other languages they've been failing fast - throw an error while building the credential, which we are also moving towards in this PR https://github.com/Azure/azure-sdk-for-java/pull/7246/files#diff-56093106beda983b28894478869221c6R128 in case of an invalid certificate file. 
I would recommend moving this to the constructor.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
```diff
@@ -0,0 +1,120 @@
...
+     * <see cref="AuthenticationFailedException"/>.
+     * <returns>A method that will ensure <code>credential</code> has been initialized</returns>
+     */
+    private void ensureCredential() throws Exception {
+        if(credential ==null){
+            try
+            {
+                credential = BuildCredentialForCredentialsFile(ParseCredentialsFile(filepath));
+            } catch (Exception e){  
+                throw new Exception("Error parsing SDK Auth File", e);
```
g2vinay: Good to add logger.logExceptionAsError(new Exception....)
logger can be used from azure core
Ideally, would like to use a better Exception class than just Exception, like InvalidArgumentException.

Plus, IOException will need to be handled separately, and should be bubbled up as it is.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
```diff
@@ -0,0 +1,120 @@
...
+            } 
+        }
+    }
+
+    private static Map<String, String> ParseCredentialsFile(String filePath) throws Exception
+    {
+        File file = new File(filePath);
+        if (!file.exists()) {
+            throw new Exception("Auth File doesn't exist");
```
g2vinay: Good to add logger.logExceptionAsError(new Exception....)
Ideally, would like to use a better Exception class than just Exception, like InvalidArgumentException.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
```diff
@@ -0,0 +1,120 @@
...
+    private TokenCredential BuildCredentialForCredentialsFile(Map<String, String> authData) throws Exception
+    {
+        String clientId = authData.get("clientId");
+        String clientSecret = authData.get("clientSecret");
+        String tenantId = authData.get("tenantId");
+        String activeDirectoryEndpointUrl = authData.get("activeDirectoryEndpointUrl");
+
+        if (clientId == null || clientSecret == null || tenantId == null || activeDirectoryEndpointUrl == null)
+        {
+            throw new Exception("Malformed Azure SDK Auth file. The file should contain 'clientId', 'clientSecret', 'tenentId' and 'activeDirectoryEndpointUrl' values.");
```
g2vinay: same feedback here.

sdk/identity/azure-identity/src/test/java/com/azure/identity/AuthFileCredentialTests.java
```diff
@@ -0,0 +1,36 @@
...
+import reactor.test.StepVerifier;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(fullyQualifiedNames = "com.azure.identity.*")
+@PowerMockIgnore({ "com.sun.org.apache.xerces.*", "javax.xml.*", "org.xml.*" })
+public class AuthFileCredentialTests<T> {
+
+    // @Test
+    // public void SdkAuthFileEnsureCredentialParsesCorrectly() throws Exception {
+        //ClientId,TenantId,ClientSecret in ClientSecretCredential is privated
```
g2vinay: they can be made package private for accessing within the test.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @param filepath              The path to the SDK Auth file.
+     * @param identityClientOptions the options for configuring the identity client
+     */
+    AuthFileCredential(String filepath, IdentityClientOptions identityClientOptions) {
+        Objects.requireNonNull(filepath, "'filepath' cannot be null.");
+        Objects.requireNonNull(identityClientOptions, "'identityClientOptions' cannot be null.");
+        this.filepath = filepath;
+        this.identityClientOptions = identityClientOptions;
+        ensureCredential();
```
ellismg: Can we take the code that is in `ensureCredential` today and pull it into the constructor now?  It's the only place we call it now and it will simplify things.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param filepath              The path to the SDK Auth file.
+     * @param identityClientOptions the options for configuring the identity client
+     */
+    AuthFileCredential(String filepath, IdentityClientOptions identityClientOptions) {
+        Objects.requireNonNull(filepath, "'filepath' cannot be null.");
+        Objects.requireNonNull(identityClientOptions, "'identityClientOptions' cannot be null.");
+        this.identityClientOptions = identityClientOptions;
+        if (credential == null) {
+            try {
+                credential = BuildCredentialForCredentialsFile(ParseCredentialsFile(filepath));
```
schaabs: Since this performs IO I think this should be done in getToken rather than in the constructor.

jianghaolu: Thanks @schaabs for clarifying. I previously asked @Luyunmt  to fail earlier by checking the credentials file earlier. @Luyunmt I think we would still want to throw in `getToken` instead of the constructor.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
```diff
@@ -0,0 +1,132 @@
...
+     * Ensures that credential information is loaded from the SDK Auth file. This
+     * method should be called to initialize <code>_credential</code> before it is
+     * used. If the SDK Auth file is not found or invalid, this method will throw
+     * <see cref="RuntimeException"/>. <returns>A method that will
+     * ensure <code>credential</code> has been initialized</returns>
+     */
+    void ensureCredential() {
+        if (credential == null) {
+            try {
+                credential = BuildCredentialForCredentialsFile(ParseCredentialsFile(filepath));
```
srnagar: Guard this creation of credential in a multi-threaded environment.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
```diff
@@ -0,0 +1,132 @@
...
+    private TokenCredential BuildCredentialForCredentialsFile(Map<String, String> authData) throws Exception
+    {
+        String clientId = authData.get("clientId");
+        String clientSecret = authData.get("clientSecret");
+        String tenantId = authData.get("tenantId");
+        String activeDirectoryEndpointUrl = authData.get("activeDirectoryEndpointUrl");
+
+        if (clientId == null || clientSecret == null || tenantId == null || activeDirectoryEndpointUrl == null)
+        {
+            throw logger.logExceptionAsError(new ClientAuthenticationException("Malformed Azure SDK Auth file. The file should contain 'clientId', 'clientSecret', 'tenentId' and 'activeDirectoryEndpointUrl' values.",null));
```
srnagar: nit: Typo `tenantId`

sdk/identity/azure-identity/src/test/java/com/azure/identity/AuthFileCredentialTests.java
```diff
@@ -0,0 +1,46 @@
...
+package com.azure.identity;
+
+import com.azure.core.credential.TokenRequestContext;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.powermock.core.classloader.annotations.PowerMockIgnore;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
```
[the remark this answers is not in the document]
g2vinay: This can be deferred as discussed.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
Luyunmt: Credential is instance field and method credential.gettoken doesn't change the state of credential,so we don't think it is necessary to take care the  multi-threaded environment

sdk/identity/azure-identity/src/main/java/com/azure/identity/AuthFileCredential.java
```diff
@@ -0,0 +1,132 @@
...
+        } catch (IOException e) {
+            throw logger.logExceptionAsError(new IllegalStateException(e));
+        }
+    }
+
+    private TokenCredential BuildCredentialForCredentialsFile(Map<String, String> authData) throws Exception
+    {
+        String clientId = authData.get("clientId");
+        String clientSecret = authData.get("clientSecret");
+        String tenantId = authData.get("tenantId");
```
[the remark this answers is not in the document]
Luyunmt: In other credentials also using "clientid","tenant" as follow .
 public ClientCertificateCredential build() {
        ValidationUtil.validate(getClass().getSimpleName(), new HashMap<String, Object>() {{
                put("clientId", clientId);
                put("tenantId", tenantId);
                put("clientCertificate", clientCertificate);
            }});
        return new ClientCertificateCredential(tenantId, clientId, clientCertificate, clientCertificatePassword,
            identityClientOptions);
    }
Do we take it into consideration?

jongio:
@g2vinay - Can you help us get this approved for April release?

jongio:
This is on hold for further discussion.

PR #7555, opened 2020-01-21
Increase Identity test coverage

g2vinay:
Increases Identity test coverage from 42 -> 62%
Adds Tests for Identity Client, CertificateUtil, HttpPipelineAdapter.


sdk/identity/azure-identity/src/test/resources/key.pem
```diff
@@ -0,0 +1,16 @@
+-----BEGIN PRIVATE KEY-----
```
g2vinay: This key is generated as a test resource from https://csfieldguide.org.nz/en/interactives/rsa-key-generator/

And it is not the private key of the above certificate.

g2vinay:
> As discussed today it would be great to get rid of PowerMock in these tests.

Filed an issue for this, #7641

PR #7571, opened 2020-01-21
first round archetype-sdk-data.yml

JimSuplizio:
We have a bunch of logic in the [eng\pipelines\templates\jobs\archetype-sdk-client.yml](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/templates/jobs/archetype-sdk-client.yml) that runs exclusively for one branch or another and about half of the logic works correctly. For example the dependency, code coverage and spotbugs reports are being generated every run but are client only.

The purpose of this PR is to create an archetype-sdk-data.yml and clean out the client only jobs. Similarly, remove the data only jobs from the archetype-sdk-client.yml.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-data.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-data.yml
@@ -2,7 +2,7 @@ parameters:
   SDKType: data
   ServiceDirectory: not-specified # Set a default that breaks in obvious ways.
   PreTestSteps: []
-  TestOptions: '$(DefaultOptions)'
+  TestOptions: '$(DefaultOptions) -Dmaven.javadoc.skip=true -Dspotbugs.skip=true -Dcheckstyle.skip=true'
   TestGoals: verify
   TestMatrix:
     Linux - Java 8:
@@ -132,7 +132,7 @@ jobs:
 
       - script: |
           pip install setuptools wheel
-          pip install doc-warden
+          pip install doc-warden==$(DocWardenVersion)
           ward scan -d $(Build.SourcesDirectory) -c $(Build.SourcesDirectory)/eng/.docsettings.yml
         displayName: 'Verify Readmes'
 


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -217,26 +217,13 @@ jobs:
           Options: --batch-mode
           MavenOptions: $(MemoryOptions) $(LoggingOptions)
 
-      - task: Maven@3
```
JimSuplizio: Start Jetty is a data only task and is no longer necessary in client.

eng/pipelines/templates/jobs/archetype-sdk-data.yml
```diff
@@ -0,0 +1,222 @@
...
+      OSName: 'macOS'
+      OSVmImage: 'macOS-10.13'
+      JavaVersion: '1.11'
+      RunTitle: 'macOS on Java 1.11'
+    Windows - Java 11:
+      OSName: 'Windows'
+      OSVmImage: 'windows-2019'
+      JavaVersion: '1.11'
+      RunTitle: 'Windows on Java 1.11'
+
```
JimSuplizio: Data track doesn't have the same interdependencies as client track and the **from source** runs don't really make much sense.

weshaggard: OK we can add it later if we find a gap we need to fill.

eng/pipelines/templates/jobs/archetype-sdk-data.yml
```diff
@@ -0,0 +1,222 @@
...
+      vmImage: 'ubuntu-16.04'
+
+    strategy:
+      matrix:
+        Client Libraries:
+          ArtifactName: 'packages'
+          JavaVersion: '1.11'
+
+    steps:
+      - script: |
```
weshaggard: Orthogonal but potentially more clean-up. I believe this was added to enabling the publishing of dependency reports, see https://github.com/Azure/azure-sdk-for-java/pull/4238, we should see if that is still necessary after all our refactorings and changes to unified pipelines.

sdk/authorization/ci.yml
```diff
@@ -37,7 +37,7 @@ pr:
       - sdk/authorization/mgmt-
 
 stages:
-  - template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
+  - template: ../../eng/pipelines/templates/stages/archetype-sdk-data.yml
     parameters:
       ServiceDirectory: authorization
       SDKType: data
```
weshaggard: Should we set the SDKType in the template now?

eng/pipelines/templates/jobs/archetype-sdk-data.yml
```diff
@@ -0,0 +1,222 @@
...
+    condition: and(succeeded(), ne(variables['Skip.Analyze'], 'true'))
+
+    variables:
+      - template: ../variables/globals.yml
+
+    pool:
+      vmImage: 'ubuntu-16.04'
+
+    steps:
+      - pwsh: |
```
weshaggard: I didn't look super close but it still feels like there might be a fair amount of duplication between the client and data templates. We should consider creating more templates to share as much as possible were it makes sense.

eng/pipelines/templates/jobs/archetype-sdk-data.yml
JimSuplizio: The from source builds were never intended to run for anything other than client. Part of this change removed the [conditions](https://github.com/Azure/azure-sdk-for-java/pull/7571/files#diff-a36606b54a95a0d62cfd379adb42c11eL239) in the archetype-sdk-client that were limiting this run.

eng/pipelines/templates/jobs/archetype-sdk-data.yml
JimSuplizio: @mitchdenny, it looks like the reason this tag step was being done was for the nightly dependency report. It seems that this should be a step within the [nightly aggregate report generation](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/aggregate-reports.yml). Though it's unclear if this step is actually still necessary.

sdk/authorization/ci.yml
JimSuplizio: Good call/catch. I'll remove those from the various ci.yml files.

eng/pipelines/templates/jobs/archetype-sdk-data.yml
mitchdenny: I haven't touched the dependency reports, I think they still run off one of the tagged builds. I think we should file a separate issue for that because we almost certainly will want to move that over to the aggregate reports.

JimSuplizio: OK @mitchdenny, I'll leave this alone in my PR and let you file the issue.

mitchdenny: Logged: https://github.com/Azure/azure-sdk-for-java/issues/7592

PR #7593, opened 2020-01-21
Add Support for Environment Proxy Configurations

alzimmermsft:
Updated the Netty and OkHttp `HttpClient`s to check the environment for proxy configuration if the `ProxyOptions` set in their builder is `null`. Additionally, added the ability to set the `Configuration` object that the builders will use when retrieving proxy information from the environment.

Environment proxy configurations will be loaded in the following order:

1. Azure HTTPS
2. Azure HTTP
3. Java HTTPS
4. Java HTTP

Azure proxy configurations are preferred over the Java proxy configurations as they relate more closely to the SDKs, additionally HTTPS is preferred over HTTP as it is more secure. Java proxies will require `java.net.useSystemProxies` to be `true` before that proxy configuration is applied.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
@@ -88,6 +88,10 @@ public class ProxyOptions {
     /**
      * Sets the hosts which bypass the proxy.
      *
+     * <p>
+     * The expected format of the passed string is a {@code '|'} delimited list of hosts which should bypass the proxy.
+     * Individual host strings may contain regex characters such as {@code '*'}.
+     *
      * @param nonProxyHosts Hosts that bypass the proxy.
      * @return the updated ProxyOptions object
      */


sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -78,50 +85,55 @@ public NettyAsyncHttpClientBuilder(HttpClient nettyHttpClient) {
...
-                        return tcpConfig.proxy(ts -> ts.type(nettyProxy)
-                                .address(proxyOptions.getAddress())
-                                .username(proxyOptions.getUsername())
-                                .password(userName -> proxyOptions.getPassword())
-                                .build());
-                    } else {
-                        return tcpConfig.proxy(ts -> ts.type(nettyProxy).address(proxyOptions.getAddress()));
-                    }
+                ProxyOptions buildProxyOptions = (proxyOptions == null)
+                    ? ProxyOptions.loadFromEnvironment(buildConfiguration)
```
srnagar: nit: `loadFromConfiguration` would be better a name as you are passing the configuration to this method (and configuration can load data from the environment or somewhere else)

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
```diff
@@ -68,6 +115,125 @@ public String getPassword() {
...
+     */
+    public String getNonProxyHosts() {
+        return this.nonProxyHosts;
+    }
+
+    /**
+     * Attempts to load a proxy from the environment.
+     *
+     * <p>
+     * Environment configurations are loaded in this oreder:
```
srnagar:
```suggestion
     * Environment configurations are loaded in this order:
```

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
```diff
@@ -68,6 +115,125 @@ public String getPassword() {
...
+     * @param configuration The {@link Configuration} that is used to load proxy configurations from the environment.
+     * If {@code null} is passed then {@link Configuration#getGlobalConfiguration()} will be used. If
+     * {@link Configuration#NONE} is passed {@link IllegalArgumentException} will be thrown.
+     * @return A {@link ProxyOptions} reflecting a proxy loaded from the environment, if no proxy is found {@code null}
+     * will be returned.
+     * @throws IllegalArgumentException If {@code configuration} is {@link Configuration#NONE}.
+     */
+    public static ProxyOptions loadFromEnvironment(Configuration configuration) {
+        if (configuration == Configuration.NONE) {
+            throw new IllegalArgumentException("'configuration' cannot be 'Configuration.NONE'.");
```
srnagar: This should be wrapped with logger.logAsError().

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
```diff
@@ -68,6 +115,125 @@ public String getPassword() {
...
+                        );
+                    } catch (UnsupportedEncodingException e) {
+                        return null;
+                    }
+                }
+            }
+
+            return proxyOptions;
+        } catch (MalformedURLException ex) {
+            return null;
```
srnagar: Add a log when there's a malformed url. It helps in debugging issues.

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
```diff
@@ -68,6 +115,125 @@ public String getPassword() {
...
+        // No proxy configuration setup.
+        if (CoreUtils.isNullOrEmpty(host)) {
+            return null;
+        }
+
+        int port;
+        try {
+            port = Integer.parseInt(configuration.get(type + "." + JAVA_PROXY_PORT));
+        } catch (NumberFormatException ex) {
+            port = HTTPS.equals(type) ? 443 : 80;
```
srnagar: Consider making these named constants `DEFAULT_HTTPS_PORT` and `DEFAULT_HTTP_PORT`.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
alzimmermsft: Renamed

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
alzimmermsft: Fixed typo

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
alzimmermsft: Made these constant.

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
alzimmermsft: This will require either adding a static logger or create an instance every time this method is called, instead should this scenario return `null`? Throwing here was an arbitrary decision since passing no configuration won't return anything reasonable.

alzimmermsft:
@srnagar I replied to your comments if you want to do another quick pass.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -78,50 +85,55 @@ public NettyAsyncHttpClientBuilder(HttpClient nettyHttpClient) {
...
+        switch (type) {
+            case HTTP:
+                return ProxyProvider.Proxy.HTTP;
+            case SOCKS4:
+                return ProxyProvider.Proxy.SOCKS4;
+            case SOCKS5:
+                return ProxyProvider.Proxy.SOCKS5;
+            default:
+                throw logger.logExceptionAsError(new IllegalStateException(
+                    String.format("Unknown Proxy type '%s' in use. Not configuring Netty proxy.", type)));
```
JonathanGiles: Think of this error message from the users perspective - this will probably require them to do a search for this string and hope they have found a valid result. What additional guidance can you provide in here to help the user solve this without a web search being required?

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -153,59 +171,73 @@ public HttpClient build() {
...
-        if (this.connectionTimeout != null) {
-            httpClientBuilder = httpClientBuilder.connectTimeout(this.connectionTimeout);
-        } else {
-            httpClientBuilder = httpClientBuilder.connectTimeout(DEFAULT_CONNECT_TIMEOUT);
-        }
+
+        // Use the configured read timeout if set, otherwise use the default (120s).
+        httpClientBuilder = (this.readTimeout != null)
+            ? httpClientBuilder.readTimeout(this.readTimeout)
+            : httpClientBuilder.readTimeout(DEFAULT_READ_TIMEOUT);
```
JonathanGiles: You might want to consider moving the ternary operation inside the method call and doing it on one line?

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
```diff
@@ -153,59 +171,73 @@ public HttpClient build() {
...
+    private static Proxy.Type mapProxyType(ProxyOptions.Type type, ClientLogger logger) {
+        switch (type) {
+            case HTTP:
+                return Proxy.Type.HTTP;
+            case SOCKS4:
+            case SOCKS5:
+                return Proxy.Type.SOCKS;
+            default:
+                throw logger.logExceptionAsError(new IllegalStateException(
+                    String.format("Unknown Proxy type '%s' in use. Not configuring OkHttp proxy.", type)));
```
JonathanGiles: Same comment as earlier for here

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
```diff
@@ -40,6 +85,17 @@ public ProxyOptions setCredentials(String username, String password) {
         return this;
     }
 
+    /**
+     * Sets the hosts which bypass the proxy.
+     *
+     * @param nonProxyHosts Hosts that bypass the proxy.
+     * @return the updated ProxyOptions object
+     */
+    public ProxyOptions setNonProxyHosts(String nonProxyHosts) {
```
JonathanGiles: Is 'non-proxy hosts' common terminology? How does the user work with multiple non-proxy hosts? If it is, e.g. comma-separated, would it be better for the API to be varargs or a List<String>?

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
```diff
@@ -68,6 +124,126 @@ public String getPassword() {
...
+     * {@code null} will be returned if no proxy was found in the environment.
+     *
+     * @param configuration The {@link Configuration} that is used to load proxy configurations from the environment.
+     * If {@code null} is passed then {@link Configuration#getGlobalConfiguration()} will be used. If
+     * {@link Configuration#NONE} is passed {@link IllegalArgumentException} will be thrown.
+     * @return A {@link ProxyOptions} reflecting a proxy loaded from the environment, if no proxy is found {@code null}
+     * will be returned.
+     * @throws IllegalArgumentException If {@code configuration} is {@link Configuration#NONE}.
+     */
+    public static ProxyOptions loadFromConfiguration(Configuration configuration) {
```
JonathanGiles: I wonder if we should call this `fromConfiguration`, that feels a little more Java idiomatic

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
alzimmermsft: Updated error message

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
alzimmermsft: Updated

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
alzimmermsft: `nonProxyHosts` is what Java uses for host that should ignore the environment configured proxy, I chose this to stay inline with language specifics.

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
alzimmermsft: Done

sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClientBuilder.java
alzimmermsft: Done

sdk/core/azure-core/src/main/java/com/azure/core/http/ProxyOptions.java
alzimmermsft: This is expected to be a regex based on the documentation for `http.nonProxyHosts` (the Java environment variable for this). Our `AZURE_NO_PROXY` environment variable doesn't state the expected format, only statement made is that is must support CIDR annotation.

PR #7595, opened 2020-01-22
Fixed ShareStats bug

gapra-msft:


sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareAsyncClient.java
```diff
@@ -1086,7 +1087,9 @@ public String generateSas(ShareServiceSasSignatureValues shareServiceSasSignatur
     }
 
     private Response<ShareStatistics> mapGetStatisticsResponse(SharesGetStatisticsResponse response) {
-        ShareStatistics shareStatistics = new ShareStatistics((int) (response.getValue().getShareUsageBytes() / 1024));
+        ShareStatistics shareStatistics =
+            new ShareStatistics((int) (response.getValue().getShareUsageBytes() / (Constants.GB)),
```
rickle-msft: Don't we want Math.ceil() of these values so we round up?

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareStatistics.java
```diff
@@ -16,6 +17,18 @@
      */
     public ShareStatistics(int shareUsageInGB) {
         this.shareUsageInGB = shareUsageInGB;
+        this.shareUsageInBytes = 0;
```
rickle-msft: It might be better to default this to -1 so that customers can tell the difference between not set and empty share

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareAsyncClient.java
rickle-msft: It might also be better for the new constructor to just take a long and we calculate the share usage in bytes in the constructor.

gapra-msft: Done

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/models/ShareStatistics.java
gapra-msft: Done

Issue #7390, opened 2020-01-13
Add ability to rotate subscription keys for Text Analytics service


samvaity:
Please refer to https://github.com/Azure/azure-sdk/issues/947

PR #7596, opened 2020-01-22
[Feature] To be able to rotate subscription key

mssfang:
Add the ability to rotate the subscription key.
In addition, increase test coverage to TextAnalyticsClientBuilder.

Fixes: #7390

### DIFF
--- a/sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTest.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTest.java
@@ -497,7 +497,7 @@ public class TextAnalyticsClientTest extends TextAnalyticsClientTestBase {
     public void nullSubscriptionKey() {
         assertThrows(NullPointerException.class, () -> {
             final TextAnalyticsClientBuilder builder = new TextAnalyticsClientBuilder();
-            builder.endpoint(getEndpoint()).subscriptionKey(null).buildClient();
+            builder.endpoint(getEndpoint()).subscriptionKey(null);
         });
     }
 
@@ -508,7 +508,7 @@ public class TextAnalyticsClientTest extends TextAnalyticsClientTestBase {
     public void nullAADCredential() {
         assertThrows(NullPointerException.class, () -> {
             final TextAnalyticsClientBuilder builder = new TextAnalyticsClientBuilder();
-            builder.endpoint(getEndpoint()).credential(null).buildClient();
+            builder.endpoint(getEndpoint()).credential(null);
         });
     }
 
--- a/sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsAsyncClientTest.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsAsyncClientTest.java
@@ -567,28 +567,6 @@ public class TextAnalyticsAsyncClientTest extends TextAnalyticsClientTestBase {
     }
 
     /**
-     * Test for null subscription key
-     */
-    @Test
-    public void nullSubscriptionKey() {
-        assertThrows(NullPointerException.class, () -> {
-            final TextAnalyticsClientBuilder builder = new TextAnalyticsClientBuilder();
-            builder.endpoint(getEndpoint()).subscriptionKey(null).buildAsyncClient();
-        });
-    }
-
-    /**
-     * Test for null AAD credential
-     */
-    @Test
-    public void nullAADCredential() {
-        assertThrows(NullPointerException.class, () -> {
-            final TextAnalyticsClientBuilder builder = new TextAnalyticsClientBuilder();
-            builder.endpoint(getEndpoint()).credential(null).buildAsyncClient();
-        });
-    }
-
-    /**
      * Test for null service version, which would take take the default service version by default
      */
     @Test


sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
```diff
@@ -0,0 +1,40 @@
...
+        return this.subscriptionKey;
+    }
+
+    /**
+     * Set the subscription key value.
+     *
+     * @param subscriptionKey the subscription key
+     * @return the {@link TextAnalyticsSubscriptionKeyCredential} itself
+     */
+    public TextAnalyticsSubscriptionKeyCredential updateCredential(String subscriptionKey) {
```
srnagar: Don't think returning `TextAnalyticsSubscriptionKeyCredential` is necessary here. Returning `void` should be sufficient.

srnagar: I am not sure if we agreed on the terminology here in previous arch-board review - there were several proposals `refresh`, `update`, `set`, `rotate` etc.

srnagar: Also, when is this method called? Is this method expected to be called by the user when they want to update the key? If so, it would be good to include in samples and code snippets.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.ai.textanalytics.models;
+
+/**
+ * Subscription key credential that shared across cognitive services, or restrict to single service.
```
srnagar: Is this applicable for all cognitive services? If so, is the plan to duplicate this credential in all cognitive services modules or would we create a common module like storage has to put the common stuff there?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Set the subscription key value.
+     *
+     * @param subscriptionKey the subscription key
+     * @return the {@link TextAnalyticsSubscriptionKeyCredential} itself
+     */
+    public TextAnalyticsSubscriptionKeyCredential updateCredential(String subscriptionKey) {
+        this.subscriptionKey = subscriptionKey;
```
srnagar: Can subscriptionKey be null or empty? If not, you may want to check for these conditions here and in the constructor.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
mssfang: I will add the samples and snippet.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
mssfang: We had a discussion about having a common module. We decided not to have a common module now because currently we only have one service implemented. We don't really know the use case.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
mssfang: Agreed. Will add the check

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
```diff
@@ -0,0 +1,40 @@
...
+    public TextAnalyticsSubscriptionKeyCredential(String subscriptionKey) {
+        this.subscriptionKey = subscriptionKey;
+    }
+
+    /**
+     * Get the subscription key value.
+     *
+     * @return the subscription key value
+     */
+    public String getSubscriptionKey() {
```
maririos: Is there a reason why we need this? 
Although the guidelines are not done, during the meeting they agreed that 
>SHOULD NOT define public properties or fields which allow users to access the authentication data directly.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
maririos: Is this thread safe? Is there a way to guarantee that when modified it updates everywhere?

sdk/textanalytics/azure-ai-textanalytics/README.md
```diff
@@ -73,11 +73,11 @@ cognitive services.
...
     
     Use the key as the credential parameter to authenticate the client:
-    <!-- embedme ./src/samples/java/com/azure/ai/textanalytics/ReadmeSamples.java#L45-L48 -->
+    <!-- embedme ./src/samples/java/com/azure/ai/textanalytics/ReadmeSamples.java#L44-L47 -->
     ```java
     TextAnalyticsClient textAnalyticsClient = new TextAnalyticsClientBuilder()
-        .subscriptionKey(SUBSCRIPTION_KEY)
-        .endpoint(ENDPOINT)
+        .subscriptionKey(new TextAnalyticsSubscriptionKeyCredential("{subscription_key}"))
+        .endpoint("https://{servicename}.cognitiveservices.azure.com/")
```
maririos: maybe we should leave it as before `ENDPOINT`? The customer can go to the portal or with Azure CLI and get the endpoint.
Leaving it like this, might confuse people when using a multi-service subscription key as the region goes before the endpoint

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/SubscriptionKeyCredentialPolicy.java
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @param credential the SharedKeyCredential credential used to create the policy.
+     */
+    public SubscriptionKeyCredentialPolicy(TextAnalyticsSubscriptionKeyCredential credential) {
+        this.credential = credential;
+    }
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        context.getHttpRequest().setHeader("Ocp-Apim-Subscription-Key", credential.getSubscriptionKey());
```
hemanttanwar: May be use constant defined in this class instead of string ?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
mssfang: Yes. the property or field "private String subscriptionKey" is a private. User have to invoke the setter `updateCredential` to update the authentication data.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
mssfang: I had a few discussions on the thread-safe concern. some believe it is thread-safe. the reason is it is a single atomic assignment. But depends on how you use it, It could be not threaded safe. This is one I am not sure yet. I will give you an update once finalized.

sdk/textanalytics/azure-ai-textanalytics/README.md
mssfang: Good to know this. I referred to the docs we have. I found nowhere has an example of multi-service key endpoint. Can you share an example of it?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
```diff
@@ -0,0 +1,40 @@
...
+        return this.subscriptionKey;
+    }
+
+    /**
+     * Set the subscription key value.
+     *
+     * @param subscriptionKey the subscription key
```
hemanttanwar: May be adding better description `the subscription key for authentication.`

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientBuilderTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        final TextAnalyticsClient client = clientBuilder.buildClient();
+
+        // Action and Assert
+        validateDetectedLanguages(Arrays.asList(new DetectedLanguage("English", "en", 1.0)),
+            client.detectLanguage("This is a test English Text").getDetectedLanguages());
+
+        // Update to invalid subscription key
+        credential.updateCredential("invalid key");
+
```
hemanttanwar: What happens when client update with valid key and after that all the calls should be successful. Is there a test for that ?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
hemanttanwar: The samples/readme.md : provide an example of update.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
maririos: sure, see here: https://docs.microsoft.com/en-us/azure/cognitive-services/authentication#authenticate-with-a-multi-service-subscription-key

sdk/textanalytics/azure-ai-textanalytics/README.md
maririos: sure, see here: https://docs.microsoft.com/en-us/azure/cognitive-services/authentication#authenticate-with-a-multi-service-subscription-key

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
maririos: thanks for clarifying! this is me not knowing Java :P

sdk/textanalytics/azure-ai-textanalytics/README.md
mssfang: Thank you

g2vinay:
Looking at the 

> From a Java point of view, this all looks fine. I would like @g2vinay and @schaabs to weigh in on how this fits into their vision for auth and identity before this is made public API, though.

Looking at the authentication model for text analytics, a subscription key is passed via header and that's it (it acts as the access token).
TLS encryption will be applied during network calls so subscription key can't be easily intercepted.
Its worth to ensure/add the check in the code that the resource endpoint we're hitting starts with 'https', to ensure TLS protocol comes into play during network calls.

Currently, in Azure Identity there is no support to rotate any of the credentials it supports. The credential needs to be restarted.
In this PR too, the rotation process is manual for the user via the update call on the subscription key credential. So, that looks fine to me.

JonathanGiles:
> Currently, in Azure Identity there is no support to rotate any of the credentials it supports. The credential needs to be restarted.
> In this PR too, the rotation process is manual for the user via the update call on the subscription key credential. So, that looks fine to me.

This was the main point of my request - I do not want this API to be evolving separately than azure-identity intends to evolve, and would therefore like consideration about whether there is any reconciliation that needs to go on here.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/TextAnalyticsClientBuilder.java
```diff
@@ -42,7 +44,7 @@
  *
  * <p>
  * The client needs the service endpoint of the Azure Text Analytics to access the resource service.
- * {@link #subscriptionKey(String) subscriptionKey(String)} or
+ * {@link #subscriptionKey(TextAnalyticsSubscriptionKeyCredential) subscriptionKey(String)} or
```
conniey: The friendly text is incorrect. `subscriptionKey(String)`

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/TextAnalyticsClientBuilder.java
```diff
@@ -87,7 +88,7 @@
     private final String clientVersion;
 
     private String endpoint;
-    private String subscriptionKey;
+    private TextAnalyticsSubscriptionKeyCredential subscriptionKey;
```
conniey: nit: variable name `credential` fits better.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
conniey: In the pipeline policy, it is fetched, so it's possible that the policy picks up a different value than the one that is set. This is not thread safe.

conniey:
> This was the main point of my request - I do not want this API to be evolving separately than azure-identity intends to evolve, and would therefore like consideration about whether there is any reconciliation that needs to go on here.

I think it would be better to have a rotating credential (either manual or on some cadence, etc.) of sorts in azure-identity. We want to support rotating credentials in Event Hubs, too.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsSubscriptionKeyCredential.java
annelo-msft: @mssfang is right - this is still an open question.  We're starting to look at FormRecognizer now, though, so we'll start driving the "where would we put a common Cognitive credential" with the architects.

annelo-msft:
> Looking at the authentication model for text analytics, a subscription key is passed via header and that's it (it acts as the access token).
> TLS encryption will be applied during network calls so subscription key can't be easily intercepted.

Some of the other cognitive services support getting a bearer token from an STS in addition to passing the subscription key.  TextAnalytics doesn't currently support this, which is part of why we're scoping this type to TA-only.

sdk/textanalytics/azure-ai-textanalytics/src/samples/java/com/azure/ai/textanalytics/AnalyzeSentimentAsync.java
```diff
@@ -19,8 +20,8 @@
     public static void main(String[] args) {
         // Instantiate a client that will be used to call the service.
         TextAnalyticsAsyncClient client = new TextAnalyticsClientBuilder()
-            .subscriptionKey("{subscription_key}")
-            .endpoint("https://{servicename}.cognitiveservices.azure.com/")
+            .subscriptionKey(new TextAnalyticsSubscriptionKeyCredential("{subscription_key}"))
+            .endpoint("{endpoint}")
```
conniey: Why the change to `{endpoint}`? It is more clear what the string should look like with the last change.

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTest.java
```diff
@@ -408,4 +414,150 @@ public void analyseSentimentForBatchInputShowStatistics() {
...
+    }
+
+    /**
+     * Test for null AAD credential
+     */
+    @Test
+    public void nullAADCredential() {
+        assertThrows(NullPointerException.class, () -> {
+            final TextAnalyticsClientBuilder builder = new TextAnalyticsClientBuilder();
+            builder.endpoint(getEndpoint()).credential(null).buildClient();
```
conniey: Generally, you try to limit the scope of your assertion. How do you know it threw in credential(null) and not buildclient()?

This goes for all the other tests that do the same.

```java
// arrange
final TextAnalyticsClientBuilder builder = new TextAnalyticsClientBuilder();

// Act & Assert
assertThrows(NullPointerException.class, () -> builder.credential(null));
```

sdk/textanalytics/azure-ai-textanalytics/src/samples/java/com/azure/ai/textanalytics/AnalyzeSentimentAsync.java
mssfang: Based on what Mariana point out here: https://github.com/Azure/azure-sdk-for-java/pull/7596#discussion_r369861889 

The subscription key could apply to both Cognitive Subscription Key or Text Analytics Subscription key.

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTest.java
mssfang: Make sense. I will update it.

sdk/textanalytics/azure-ai-textanalytics/README.md
```diff
@@ -114,14 +114,27 @@ analyze sentiment, recognize entities, detect language, and extract key phrases
...
+    .endpoint("{endpoint}")
     .buildClient();
 ```
 
+#### Rotate existing subscription key
+The Azure Text Analytics client library provide a way to rotate the existing subscription key.
+
+<!-- embedme ./src/samples/java/com/azure/ai/textanalytics/ReadmeSamples.java#L174-L180 -->
+```java
+TextAnalyticsSubscriptionKeyCredential credential = new TextAnalyticsSubscriptionKeyCredential("{expired_subscription_key}");
```
maririos: I wouldn't call it `expired_subscription_key` as it might give the wrong message to people when creating the client.
it could be just `subscription_key`, and leave `new_subscription_key`

JonathanGiles:
I asked for this to be reviewed and approved by @g2vinay or @schaabs from an identity perspective, but I don't see that approval on this PR. Has this been done?

mssfang:
@JonathanGiles  Sorry. No for your question. I will follow up with vinay and scott for further improvement and open a new PR if needed.

PR #7598, opened 2020-01-22
Update test task in CI pipeline to remove unwanted maven steps

### LINKED ISSUE #7591: Look at adding the code coverage verification to a post-test step instead of setting TestGoals: verify on the matrix
When looking at a [run failure in the core pipeline](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=239700&view=results) I noticed that the run tests steps seemed to be running a lot more than tests and coverage checks. In the linked run it looks like the run tests are running javadoc among other things. Doing a bit more digging, when the changes were made to fail PRs due to a code coverage threshold, the [test goal was changed from test to verify](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/templates/jobs/archetype-sdk-client.yml#L6) which ends up running a lot more than just coverage. Eyeballing the test run times for the core pipeline, test runs seem to be taking anywhere from 10-20 minutes longer. We need to be mindful that when we're adding things to a particular step that we don't accidentally pull in extras.

srnagar:
Maven's `verify` goal executes a lot more steps like checkstyle validation, spotbugs validate and javadoc validation which is not required for checking code coverage. So, disabling all these steps for test task in CI pipeline.

Fixes #7591

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -2,7 +2,7 @@ parameters:
   SDKType: not-specified # Set a default that breaks in obvious ways.
   ServiceDirectory: not-specified # Set a default that breaks in obvious ways.
   PreTestSteps: []
-  TestOptions: '$(DefaultOptions) -Dmaven.javadoc.skip -Dspotbugs.skip -Dcheckstyle.skip'
+  TestOptions: '$(DefaultOptions) -Dmaven.javadoc.skip=true -Dspotbugs.skip=true -Dcheckstyle.skip=true'
   TestGoals: verify
   TestMatrix:
     Linux - Java 8:


alzimmermsft:
What about skipping RevApi as well?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -266,7 +266,7 @@ jobs:
         inputs:
           mavenPomFile: sdk/${{parameters.ServiceDirectory}}/pom.service.xml
           options: ${{ parameters.TestOptions }}
-          mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
+          mavenOptions: '$(MemoryOptions) $(LoggingOptions) -Dmaven.javadoc.skip -Dspotbugs.skip -Dcheckstyle.skip'
```
alzimmermsft: What about skipping RevApi as well? Also do these flags need `=true`?

alzimmermsft: Also is this the right file for it to be added to? Doesn't look like the PR validation test runs have these additional arguments.

srnagar: `revapi` does not run for `verify` goal. Also, this is the file I updated when I changed the goal from `test` to `verify`. Not sure about "=true", didn't have to use it on my machine, may be required for CI?

JimSuplizio: I'm pretty sure you're going to want the =true just to be safe. There's an example of skipping these exact items in [eng/pipelines/client.yml](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/client.yml#L217) which sets the options as "-Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true'"

mitchdenny:
Late question on this (sorry). I know that we have generally three jobs in the build stage: Build, Analyze and Test. But one of the things I've been wondering about with our Java builds in particular is the relative benefits of splitting the spotbugs etc runs out into the Analyze jobs.

Idiomatic Maven builds would generally tie these all together so that when we ran ```mvn package``` against the ```pom.service.xml``` file, that verification would be included. Given that we need to download a whole bunch of artifacts in order to do the package step anyway, why don't we just run spotbugs/checkstyle/revapi etc as part of the first Build job anyway?

PR #7599, opened 2020-01-22
Adding PollerFlux::create to support ActivationOp variant that returns PollResponse

anuchandy:


sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -89,17 +90,94 @@ public PollerFlux(Duration defaultPollInterval,
...
+                       Function<PollingContext<T>, Mono<U>> fetchResultOperation,
+                       boolean ignored) {
+        Objects.requireNonNull(defaultPollInterval, "'defaultPollInterval' cannot be null.");
+        if (defaultPollInterval.compareTo(Duration.ZERO) <= 0) {
+            throw logger.logExceptionAsWarning(new IllegalArgumentException(
+                "Negative or zero value for 'defaultPollInterval' is not allowed."));
+        }
+        this.defaultPollInterval = defaultPollInterval;
+        this.activationOperation = null;
+        this.activationOperationEx = Objects.requireNonNull(activationOperation,
```
srnagar: nit: rename this to `activationOperationWithResponse` to make it clear that this is different from `activationOperation` which doesn't have the PollResponse.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -89,17 +90,94 @@ public PollerFlux(Duration defaultPollInterval,
...
+        this.fetchResultOperation = Objects.requireNonNull(fetchResultOperation,
+            "'fetchResultOperation' cannot be null.");
+        this.oneTimeActivationMono = oneTimeActivationMono();
     }
 
     @Override
     public void subscribe(CoreSubscriber<? super AsyncPollResponse<T, U>> actual) {
         this.oneTimeActivationMono
-            .flatMapMany(ignored -> pollingLoop())
+            .flatMapMany(ignored -> {
```
srnagar: Would be good to add a unit test case for this change.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -129,34 +207,43 @@ public void subscribe(CoreSubscriber<? super AsyncPollResponse<T, U>> actual) {
...
-    private Mono<Boolean> oneTimeActivationMono(final Function<PollingContext<T>, Mono<T>> activationOperation) {
+    @SuppressWarnings("unchecked")
+    private Mono<Boolean> oneTimeActivationMono() {
         return Mono.defer(() -> {
             if (this.activated) {
                 return Mono.just(true);
             }
             if (this.guardActivationCall.compareAndSet(this, 0, 1)) {
-                final Mono<T> activationMono;
+                final boolean isEx = this.activationOperationEx != null;
```
srnagar: same here - maybe name this as `isActivationWithResponse`

anuchandy:
@srnagar  - I just restructured the code instead of tweaking existing one by adding if/else and casting etc. I felt it started loosing readability. :(

The updates are:

1. The whole concept of one-time-activation is in its own inner class.
2. No two private vars to store activationOp references.
2. Replaced AtomicFieldUpdater with AtomicBoolean.
3. Removed un-necessary compareAndSet.

I'm adding a test for factory method. 

Please take a look when you get a chance.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
anuchandy: I'm trying to get rid of storing activationOperation in class level, just updated the code to move all activation logic to its own inner class.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
anuchandy: tests added

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -38,10 +38,7 @@
      * Creates DefaultSyncPoller.
      *
      * @param defaultPollInterval the default polling interval
-     * @param activationOperation the activation operation to be invoked at most once across all subscriptions,
-     *                            this parameter is required, if there is no specific activation work to be
-     *                            done then invocation should return Mono.empty(), this operation will be called
-     *                            with a new {@link PollingContext}.
+     * @param syncActivationOperation the activation function upon call synchronously execute the activation operation.
```
JonathanGiles: Can you please improve this wording - I'm not entirely sure what it means.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -87,19 +82,103 @@ public PollerFlux(Duration defaultPollInterval,
...
+            cxt -> new PollResponse<>(LongRunningOperationStatus.NOT_STARTED, activationOperation.apply(cxt).block());
+    }
+
+    /**
+     * Creates PollerFlux.
+     *
+     * PollerFlux obtained from this factory method uses an activationOperation which returns a Mono that
+     * emits {@link PollResponse}, which holds the result.
+     * The PollerFlux created from constructor uses an activationOperation which returns a Mono that directly
+     * emits result.
```
JonathanGiles: If you're meaning to contrast this method against the constructor, be more explicit. "This create method differs from the PollerFlux constructor in that the constructor uses an ...."

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -87,19 +82,103 @@ public PollerFlux(Duration defaultPollInterval,
...
+     * Since the first variant of PollerFlux has access to the response, it can skip the polling loop if the
+     * response indicate that LRO is completed. Whereas the second PollerFlux variant calls pollFunction at
```
JonathanGiles: Avoid acronyms, e.g, "that the long-running operation"

JonathanGiles: Similarly, I'm not entirely clear what you mean by 'first variant' and 'second variant'. Try to be more explicit to avoid any confusion.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -87,19 +82,103 @@ public PollerFlux(Duration defaultPollInterval,
...
+     * least once.
+     *
+     * @param defaultPollInterval the default polling interval
+     * @param activationOperation the activation operation to be invoked at most once across all subscriptions,
+     *                            this parameter is required, if there is no specific activation work to be
+     *                            done then invocation should return Mono.empty(), this operation will be called
+     *                            with a new {@link PollingContext}.
```
JonathanGiles: Don't worry about this kind of indentation - checkstyle should be warning against this too - please check if checkstyle passes.

JonathanGiles: This is a very long, single sentence. Break it down into a few sentences.

anuchandy:
Thanks @JonathanGiles, will cleanup javadoc and get back.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -108,61 +187,12 @@ public void subscribe(CoreSubscriber<? super AsyncPollResponse<T, U>> actual) {
      */
     public SyncPoller<T, U> getSyncPoller() {
         return new DefaultSyncPoller<>(this.defaultPollInterval,
-                this.activationOperation,
+                this.syncActivationOperation,
```
srnagar: Here you can do `ctx -> this.activationOperation.apply(ctx).block()` instead of eagerly creating `syncActivationOperation`

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -87,19 +82,103 @@ public PollerFlux(Duration defaultPollInterval,
...
         this.pollOperation = Objects.requireNonNull(pollOperation, "'pollOperation' cannot be null.");
         this.cancelOperation = Objects.requireNonNull(cancelOperation, "'cancelOperation' cannot be null.");
         this.fetchResultOperation = Objects.requireNonNull(fetchResultOperation,
             "'fetchResultOperation' cannot be null.");
+        this.oneTimeActivationMono = new OneTimeActivation<>(this.rootContext,
+            activationOperation,
+            // mapper
+            activationResult -> new PollResponse<>(LongRunningOperationStatus.NOT_STARTED, activationResult)).getMono();
+        this.syncActivationOperation =
+            cxt -> new PollResponse<>(LongRunningOperationStatus.NOT_STARTED, activationOperation.apply(cxt).block());
```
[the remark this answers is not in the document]
anuchandy: We don't block in the PollerFlux constructor/factory. We are creating  function that get passed to SyncPoller and it's gets blocked when called from SyncPoller.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -87,19 +82,103 @@ public PollerFlux(Duration defaultPollInterval,
...
+        Objects.requireNonNull(activationOperation, "'activationOperation' cannot be null.");
+        this.pollOperation = Objects.requireNonNull(pollOperation, "'pollOperation' cannot be null.");
+        this.cancelOperation = Objects.requireNonNull(cancelOperation, "'cancelOperation' cannot be null.");
+        this.fetchResultOperation = Objects.requireNonNull(fetchResultOperation,
+            "'fetchResultOperation' cannot be null.");
+        this.oneTimeActivationMono = new OneTimeActivation<>(this.rootContext,
+            activationOperation,
+            // mapper
+            Function.identity()).getMono();
+        this.syncActivationOperation = cxt -> activationOperation.apply(cxt).block();
```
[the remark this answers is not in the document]
anuchandy: As mentioned in previous comment, it's function that blocked only when called from SyncPoller 

The reason for wrapping activationOpertion in Func is, we have two types of activationOperation depending on how user want to create poller - factory vs ctr.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
anuchandy: We could do it in previous version but not anymore since we don't have two class members defined to store variations of activationOperations

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -87,19 +82,103 @@ public PollerFlux(Duration defaultPollInterval,
...
+     *                             operation will be called current {@link PollingContext}, if service does not have an
+     *                             api to fetch final result and if final result is same as final poll response value
+     *                             then implementer can choose to simply return value from provided final poll response.
+     *
+     * @param <T> The type of poll response value.
+     * @param <U> The type of the final result of long-running operation.
+     * @return PollerFlux
+     */
+    public static <T, U> PollerFlux<T, U>
+        create(Duration defaultPollInterval,
```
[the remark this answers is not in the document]
anuchandy: Will do

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
anuchandy: @srnagar - or we could re-introduces those two member vars and use it for SyncPoller is created. What do you think?

srnagar: @anuchandy I guess re-introducing two variables will make it less readable. We can leave it as-is for now.

anuchandy: Ok, sounds good.

anuchandy:
@JonathanGiles - updated the JavaDoc, could you re-review?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -22,7 +22,7 @@
  * it is possible to be so, e.g. by using volatile and copying context.
  *
  * @param <T> The type of poll response value
- * @param <U> The type of the final result of long-running operation
+ * @param <U> The type of the final result of long running operation
```
JonathanGiles:
```suggestion
 * @param <U> The type of the final result of the long running operation
```

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -37,42 +37,38 @@
     /**
      * Creates DefaultSyncPoller.
      *
-     * @param defaultPollInterval the default polling interval
-     * @param activationOperation the activation operation to be invoked at most once across all subscriptions,
-     *                            this parameter is required, if there is no specific activation work to be
-     *                            done then invocation should return Mono.empty(), this operation will be called
-     *                            with a new {@link PollingContext}.
+     * @param pollInterval the polling interval
```
JonathanGiles:
```suggestion
     * @param pollInterval the polling interval.
```

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -37,42 +37,38 @@
...
+     * @param syncActivationOperation the operation to synchronously activate (start) the long running operation,
+     *     this operation will be called with a new {@link PollingContext}.
```
JonathanGiles: This still lacks clarity to me. I don't know, as a user of this API, what exactly I should supply here. In other words, you're telling me how this is used, rather than how I use it.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -37,42 +37,38 @@
...
      * @param pollOperation the operation to poll the current state of long running operation, this parameter
-     *                      is required and the operation will be called with current {@link PollingContext}.
-     * @param cancelOperation a {@link Function} that represents the operation to cancel the long-running operation
-     *                        if service supports cancellation, this parameter is required and if service does not
-     *                        support cancellation then the implementer should return Mono.error with an error message
-     *                        indicating absence of cancellation support, the operation will be called with current
-     *                        {@link PollingContext}.
+     *     is required and the operation will be called with current {@link PollingContext}.
+     * @param cancelOperation a {@link Function} that represents the operation to cancel the long running operation
+     *     if service supports cancellation, this parameter is required and if service does not support cancellation
```
JonathanGiles: Break some of these sentences up - in a few places you flow straight into "this parameter is required" when a full stop and a new sentence would improve readability.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/DefaultSyncPoller.java
```diff
@@ -37,42 +37,38 @@
...
-    DefaultSyncPoller(Duration defaultPollInterval,
-                             Function<PollingContext<T>, Mono<T>> activationOperation,
+    DefaultSyncPoller(Duration pollInterval,
+                             Function<PollingContext<T>, PollResponse<T>> syncActivationOperation,
                              Function<PollingContext<T>, Mono<PollResponse<T>>> pollOperation,
                              BiFunction<PollingContext<T>, PollResponse<T>, Mono<T>> cancelOperation,
                              Function<PollingContext<T>, Mono<U>> fetchResultOperation) {
-        Objects.requireNonNull(defaultPollInterval, "'defaultPollInterval' cannot be null.");
-        if (defaultPollInterval.compareTo(Duration.ZERO) <= 0) {
+        Objects.requireNonNull(pollInterval, "'defaultPollInterval' cannot be null.");
```
JonathanGiles:
```suggestion
        Objects.requireNonNull(pollInterval, "'pollInterval' cannot be null.");
```

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -11,13 +11,13 @@
...
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
 /**
- * A Flux that simplifies the task of executing long-running operations against an Azure service.
- * A subscription to {@link PollerFlux} initiate a long-running operation and polls the status
+ * A Flux that simplifies the task of executing long running operations against an Azure service.
+ * A subscription to {@link PollerFlux} initiate a long running operation and polls the status
```
JonathanGiles:
```suggestion
 * A subscription to {@link PollerFlux} initiates a long running operation and polls the status
```

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -38,68 +38,143 @@
...
+        this.oneTimeActivationMono = new OneTimeActivation<>(this.rootContext,
+            activationOperation,
+            // mapper
+            activationResult -> new PollResponse<>(LongRunningOperationStatus.NOT_STARTED, activationResult)).getMono();
+        this.syncActivationOperation =
+            cxt -> new PollResponse<>(LongRunningOperationStatus.NOT_STARTED, activationOperation.apply(cxt).block());
+    }
+
+    /**
+     * Creates PollerFlux.
```
JonathanGiles: Provide a better first sentence, as this is the most visible text in the JavaDoc.

anuchandy:
@JonathanGiles will send follow up Javadoc PR. Merging this to unblock mgmt work.

\\cc @yaohaizh

PR #7625, opened 2020-01-22
Free proton-j Receiver and Reactor resources on close

conniey:
* Frees proton-j resources such as Receiver link and Reactor resources when it is closed.
* Fixes error where we can get "exceeded receiver per partition/consumer group" error.


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/AmqpExceptionHandler.java
```diff
@@ -24,7 +24,7 @@
      * @param exception The exception that caused the connection error.
      */
     void onConnectionError(Throwable exception) {
-        logger.warning("Connection exception encountered: " + exception.toString(), exception);
+        logger.warning("Connection exception encountered: " + exception, exception);
```
alzimmermsft: Should this use a parameterized message?

```java
logger.warning("Connection exception encountered: {}", exception);
```

Right now this will likely log the exception message twice.

PR #7628, opened 2020-01-22
Guava Version Downgrade to 25.0-jre

kushagraThapar:
* Downgraded guava version to 25.0-jre
* Copied `Strings.lenientFormat()` from guava library to azure cosmos internal `Utils.java`

### DIFF
--- a/THIRD-PARTY-NOTICE.txt
+++ b/THIRD-PARTY-NOTICE.txt
@@ -89,3 +89,17 @@ License notice for Slf4j Simple
  OF CONTRACT, TORT OR OTHERWISE,  ARISING FROM, OUT OF OR IN CONNECTION
  WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+License notice for Guava (https://github.com/google/guava)
+------------------------------------------------------------------------------
+
+Copyright (C) 2010 The Guava Authors
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+in compliance with the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License
+is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+or implied. See the License for the specific language governing permissions and limitations under
+the License.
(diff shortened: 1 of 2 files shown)


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/Utils.java
```diff
@@ -552,4 +557,91 @@ public ValueHolder(V v) {
         holder.v = dictionary.remove(key);
         return holder.v != null;
     }
+
```
kirankumarkolli: Lets please get clarification on required claims context from OSS

kushagraThapar: Waiting for PR to be approved, and then will start the OSS process. 
Just want to make sure these changes are good and don't posses any hidden faults before starting the OSS process.

g2vinay:
@JimSuplizio , @danieljurek , @mitchdenny 
The checkout step is failing in Cosmos live test build.
Is it missing some required configuration ?
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=240921&view=results

kushagraThapar:
> @kushagraThapar these changes aren't correct. You know that versions are not changed directly in the POM, they're either changes in the version_*.txt files in the case of built libraries or the external_dependencies.txt file in the case of external dependencies. The file you should have changed is [external_dependencies.txt](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/external_dependencies.txt#L172) and then you should have run python eng/versioning/update_versions.py --ut external_dependency --sr from the root of the enlistment. This is the reason that the Analyze step is failing.
> 
> As for the checkout errors, if you were missing a configuration you would not have had ANY runs passing. @mitchdenny could this be because the pipelines are using [v1 style git tasks](https://github.com/actions/checkout/issues/23) (If you look at the last post on that thread it's someone else seeing a similar issue posted a few days ago).

@JimSuplizio Thank you Jim for correcting me here, I will update the external_dependencies.txt file and run the script and push my changes

JimSuplizio:
> @JimSuplizio , @danieljurek , @mitchdenny
> The checkout step is failing in Cosmos live test build.
> Is it missing some required configuration ?
> https://dev.azure.com/azure-sdk/internal/_build/results?buildId=240921&view=results

Apparently when we're running these runs it's merging your changes on top of master to run. When you see the git exit code of 128 that means that your branch is more than 20 commits behind master. The correct course of action here would be to rebase master and push an updated PR.

kushagraThapar:
Waiting for CELA requirements to be fulfilled. Please don't merge this PR until then.

PR #7629, opened 2020-01-22
improve toString to avoid reconstructing constant string values

moderakh:
improve toString to avoid reconstructing constant string values


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ConflictResolutionMode.java
```diff
@@ -11,37 +11,43 @@
...
-     * Setting the ConflictResolutionMode to "LAST_WRITER_WINS" indicates that conflict resolution should be done by 
+     *
+     * Setting the ConflictResolutionMode to "LAST_WRITER_WINS" indicates that conflict resolution should be done by
      * inspecting a field in the conflicting documents
      * and picking the document which has the higher value in that path. See
      * {@link ConflictResolutionPolicy#getConflictResolutionPath()} for details on how to specify the path
      *
      */
-    LAST_WRITER_WINS,
+    LAST_WRITER_WINS("LastWriterWins"),
```
kirankumarkolli: Is it to cover for POJO renames or its case-sensitive?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -21,12 +21,14 @@
...
     private final CosmosAsyncDatabase database;
-    private String id;
+    private final String id;
+    private final String link;
     private CosmosAsyncScripts scripts;
 
     CosmosAsyncContainer(String id, CosmosAsyncDatabase database) {
         this.id = id;
         this.database = database;
+        this.link = getParentLink() + "/" + getURIPathSegment() + "/" + getId();
```
kirankumarkolli: Nice :smile:

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ConflictResolutionMode.java
moderakh: most of these enum toString are passed as option header in the request.

The most common one is Consistency where for each request we used to do Consistency.toString() and as the java enum naming is different from .Net, prior to this PR toString() used to do extra logic to re-construct the .net friendly version of the enum which is acceptable by the service. My changes in the enum ensures we don't reconstruct the enum serialized to String constant values on each request.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
moderakh: thanks.

PR #7631, opened 2020-01-22
Bump KV versions

g2vinay:
Wasn't aware this process is manual now per SDK after release.
So, bumping versions now.


joshfree:
/cc'ing @alzimmermsft who is going to be the team's build facilitating dev (BFD) / release dev for the next semester; we should consider centralizing the version bump post-shipping to avoid missed version changes.

JimSuplizio:
> Wasn't aware this process is manual now per SDK after release.
> So, bumping versions now.

The PR creation on release is being automated though it'll still require review and submission like any PR.

PR #7635, opened 2020-01-22
Update error and exception model design for Text Analytics

samvaity:
This PR 
- Updates the error model for Text Analytics SDK to map to the [InnerError](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/models/InnerError.java) returned from the service.
- Combine values from [InnerErrorCodeValue](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/models/InnerErrorCodeValue.java) and [ErrorCodeValue](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/models/ErrorCodeValue.java) to a single ErrorCodeValue class.
- Throw exceptions for single input methods when service returns a `200 OK` but with a `TextAnalyticsError` 
- Throw an exception when a success attribute is accessed on a TextAnalyticsError object.

### DIFF
--- a/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/AnalyzeSentimentAsyncClient.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/AnalyzeSentimentAsyncClient.java
@@ -31,7 +31,6 @@ import java.util.stream.Collectors;
 import static com.azure.ai.textanalytics.Transforms.mapByIndex;
 import static com.azure.ai.textanalytics.Transforms.toBatchStatistics;
 import static com.azure.ai.textanalytics.Transforms.toMultiLanguageInput;
-import static com.azure.ai.textanalytics.Transforms.processSingleResponseErrorResult;
 import static com.azure.ai.textanalytics.Transforms.toTextAnalyticsError;
 import static com.azure.ai.textanalytics.Transforms.toTextDocumentStatistics;
 
@@ -57,7 +56,7 @@ class AnalyzeSentimentAsyncClient {
 
         return analyzeBatchSentimentWithResponse(
             Collections.singletonList(new TextDocumentInput("0", text, language)), null, context)
-            .map(response -> processSingleResponseErrorResult(response));
+            .map(Transforms::processSingleResponseErrorResult);
     }
 
     Mono<Response<DocumentResultCollection<AnalyzeSentimentResult>>> analyzeSentimentWithResponse(
--- a/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizeLinkedEntityAsyncClient.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizeLinkedEntityAsyncClient.java
@@ -29,7 +29,6 @@ import java.util.stream.Collectors;
 import static com.azure.ai.textanalytics.Transforms.mapByIndex;
 import static com.azure.ai.textanalytics.Transforms.toBatchStatistics;
 import static com.azure.ai.textanalytics.Transforms.toMultiLanguageInput;
-import static com.azure.ai.textanalytics.Transforms.processSingleResponseErrorResult;
 import static com.azure.ai.textanalytics.Transforms.toTextAnalyticsError;
 import static com.azure.ai.textanalytics.Transforms.toTextDocumentStatistics;
 
@@ -56,7 +55,7 @@ class RecognizeLinkedEntityAsyncClient {
 
         return recognizeBatchLinkedEntitiesWithResponse(
             Collections.singletonList(new TextDocumentInput("0", text, language)), null, context)
-            .map(response -> processSingleResponseErrorResult(response));
+            .map(Transforms::processSingleResponseErrorResult);
     }
 
     Mono<Response<DocumentResultCollection<RecognizeLinkedEntitiesResult>>> recognizeLinkedEntitiesWithResponse(
--- a/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
@@ -32,6 +32,9 @@ import java.util.stream.IntStream;
 final class Transforms {
     private static final ClientLogger LOGGER = new ClientLogger(Transforms.class);
 
+    private Transforms() {
+    }
+
     /**
      * Given a list of inputs will apply the indexing function to it and return the updated list.
      *
@@ -138,4 +141,6 @@ final class Transforms {
         com.azure.ai.textanalytics.models.TextAnalyticsError error) {
         return new TextAnalyticsException(error.getMessage(), error.getCode().toString(), error.getTarget());
     }
+
+
 }
--- a/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/DetectLanguageAsyncClient.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/DetectLanguageAsyncClient.java
@@ -27,7 +27,6 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 
 import static com.azure.ai.textanalytics.Transforms.mapByIndex;
-import static com.azure.ai.textanalytics.Transforms.processSingleResponseErrorResult;
 
 /**
  * Helper class for managing detect language endpoint.
@@ -51,7 +50,7 @@ class DetectLanguageAsyncClient {
         List<DetectLanguageInput> languageInputs = Collections.singletonList(new DetectLanguageInput("0",
             text, countryHint));
         return detectBatchLanguagesWithResponse(languageInputs, null, context)
-            .map(response -> processSingleResponseErrorResult(response));
+            .map(Transforms::processSingleResponseErrorResult);
     }
 
     Mono<Response<DocumentResultCollection<DetectLanguageResult>>> detectLanguagesWithResponse(List<String> textInputs,
--- a/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/ExtractKeyPhraseAsyncClient.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/ExtractKeyPhraseAsyncClient.java
@@ -26,7 +26,6 @@ import java.util.Objects;
 import static com.azure.ai.textanalytics.Transforms.mapByIndex;
 import static com.azure.ai.textanalytics.Transforms.toBatchStatistics;
 import static com.azure.ai.textanalytics.Transforms.toMultiLanguageInput;
-import static com.azure.ai.textanalytics.Transforms.processSingleResponseErrorResult;
 import static com.azure.ai.textanalytics.Transforms.toTextAnalyticsError;
 import static com.azure.ai.textanalytics.Transforms.toTextDocumentStatistics;
 
@@ -53,7 +52,7 @@ class ExtractKeyPhraseAsyncClient {
 
         return extractBatchKeyPhrasesWithResponse(
             Collections.singletonList(new TextDocumentInput("0", text, language)), null, context)
-            .map(response -> processSingleResponseErrorResult(response));
+            .map(Transforms::processSingleResponseErrorResult);
     }
 
     Mono<Response<DocumentResultCollection<ExtractKeyPhraseResult>>> extractKeyPhrasesWithResponse(
--- a/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizeEntityAsyncClient.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizeEntityAsyncClient.java
@@ -28,7 +28,6 @@ import java.util.stream.Collectors;
 
 import static com.azure.ai.textanalytics.Transforms.mapByIndex;
 import static com.azure.ai.textanalytics.Transforms.toBatchStatistics;
-import static com.azure.ai.textanalytics.Transforms.processSingleResponseErrorResult;
 import static com.azure.ai.textanalytics.Transforms.toTextAnalyticsError;
 import static com.azure.ai.textanalytics.Transforms.toTextDocumentStatistics;
 
@@ -55,7 +54,7 @@ class RecognizeEntityAsyncClient {
 
         return recognizeBatchEntitiesWithResponse(
             Collections.singletonList(new TextDocumentInput("0", text, language)), null, context)
-            .map(response -> processSingleResponseErrorResult(response));
+            .map(Transforms::processSingleResponseErrorResult);
     }
 
     Mono<Response<DocumentResultCollection<RecognizeEntitiesResult>>> recognizeEntitiesWithResponse(
--- a/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizePiiEntityAsyncClient.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizePiiEntityAsyncClient.java
@@ -28,7 +28,6 @@ import java.util.stream.Collectors;
 import static com.azure.ai.textanalytics.Transforms.mapByIndex;
 import static com.azure.ai.textanalytics.Transforms.toBatchStatistics;
 import static com.azure.ai.textanalytics.Transforms.toMultiLanguageInput;
-import static com.azure.ai.textanalytics.Transforms.processSingleResponseErrorResult;
 import static com.azure.ai.textanalytics.Transforms.toTextAnalyticsError;
 import static com.azure.ai.textanalytics.Transforms.toTextDocumentStatistics;
 
@@ -55,7 +54,7 @@ class RecognizePiiEntityAsyncClient {
 
         return recognizeBatchPiiEntitiesWithResponse(
             Collections.singletonList(new TextDocumentInput("0", text, language)), null, context)
-            .map(response -> processSingleResponseErrorResult(response));
+            .map(Transforms::processSingleResponseErrorResult);
     }
 
     Mono<Response<DocumentResultCollection<RecognizePiiEntitiesResult>>> recognizePiiEntitiesWithResponse(


sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
```diff
@@ -67,9 +70,9 @@ static TextDocumentBatchStatistics toBatchStatistics(RequestStatistics statistic
     static com.azure.ai.textanalytics.models.TextAnalyticsError toTextAnalyticsError(
         TextAnalyticsError textAnalyticsError) {
         return new com.azure.ai.textanalytics.models.TextAnalyticsError(
-            ErrorCodeValue.fromString(textAnalyticsError.getCode().toString()), textAnalyticsError.getMessage(),
-            textAnalyticsError.getTarget(), textAnalyticsError.getDetails() == null ? null
-            : setErrors(textAnalyticsError.getDetails()));
+            TextAnalyticsErrorCode.fromString(textAnalyticsError.getInnerError().getCode().toString()),
```
kristapratico: I might be missing something, but what happens if no innerError is returned? Will the top level error code/message/target be used?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentResult.java
```diff
@@ -64,4 +66,15 @@ public TextAnalyticsError getError() {
...
     }
+
+    /**
+     * Throw a TextAnalyticsException if result has isError true and when a non-error property was accessed.
+     */
+    void throwExceptionIfError() {
+        if (this.isError()) {
+            throw logger.logExceptionAsError(new TextAnalyticsException(
+                String.format("Error in accessing the property when %s returned with an error.",
+                    this.getClass()), null, null));
```
kristapratico: Would it make sense to return the document id/error message in the exception message here?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
```diff
@@ -0,0 +1,68 @@
...
+ * General exception for Text Analytics related failures.
+ */
+public class TextAnalyticsException extends AzureException {
+    private static final long serialVersionUID = 1L;
+
+    private final String errorCodeValue;
+    private final String target;
+
+    /**
+     * Initializes a new instance of the AmqpException class.
```
kristapratico: Amqp?

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTest.java
```diff
@@ -124,15 +120,12 @@ public void detectLanguageFaultyText() {
...
-     * TODO: update error Model. #6559
+     * Verifies that an TextAnalyticsException is thrown for a text input with invalid country hint.
      */
     @Test
     public void detectLanguageInvalidCountryHint() {
-        TextAnalyticsError expectedError = new TextAnalyticsError(ErrorCodeValue.INVALID_ARGUMENT, "Invalid Country Hint.", null, null);
-        validateErrorDocument(client.detectLanguageWithResponse("Este es un document escrito en Español.", "en", Context.NONE).getValue().getError(),
-            expectedError);
+        Exception exception = assertThrows(TextAnalyticsException.class, () -> client.detectLanguage(""));
+        assertTrue(exception.getMessage().equals(INVALID_COUNTRY_HINT_EXPECTED_EXCEPTION_MESSAGE));
```
kristapratico: Passing in an empty string gives an invalid country hint error?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizeLinkedEntityAsyncClient.java
```diff
@@ -55,7 +57,17 @@
 
         return recognizeBatchLinkedEntitiesWithResponse(
             Collections.singletonList(new TextDocumentInput("0", text, language)), null, context)
-            .map(response -> new SimpleResponse<>(response, response.getValue().iterator().next()));
+            .map(response -> {
+                Iterator<RecognizeLinkedEntitiesResult> linkedEntitiesResultIterator = response.getValue().iterator();
+                RecognizeLinkedEntitiesResult linkedEntitiesResult = null;
+                if (response.getStatusCode() == 200 && linkedEntitiesResultIterator.hasNext()) {
```
srnagar: Use `HttpUrlConnection.HTTP_OK` instead of magic numbers.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public String getMessage() {
+        String baseMessage = super.getMessage();
+
+        if (this.errorCodeValue == null) {
+            return super.getMessage();
+        } else {
+            baseMessage = String.format(Locale.US, "%s %s: {%s}", baseMessage, "ErrorCodeValue",
```
srnagar: Use `Locale.ROOT` instead of `Locale.US`

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
```diff
@@ -87,21 +90,12 @@ static TextDocumentBatchStatistics toBatchStatistics(RequestStatistics statistic
...
-                error.getMessage(),
-                error.getTarget(), error.getDetails() == null ? null : setErrors(error.getDetails())));
-        }
-        return detailsList;
+    static void toTextAnalyticsException(com.azure.ai.textanalytics.models.TextAnalyticsError error,
+        ClientLogger logger) {
+        String baseMessage = String.format(Locale.US, "%s: {%s}, %s",
+            "Status Code", 200, error.getMessage());
+        throw logger.logExceptionAsError(new TextAnalyticsException(baseMessage,
+            error.getCode().toString(),
```
srnagar: Is error guaranteed to have `code` or can `getCode()` result in NPE?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.azure.core.exception.AzureException;
+
+import java.util.Locale;
+
+/**
+ * General exception for Text Analytics related failures.
+ */
+public class TextAnalyticsException extends AzureException {
+    private static final long serialVersionUID = 1L;
```
srnagar: Generate the servialVersionUID instead of setting it to 1L

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
```diff
@@ -87,21 +90,12 @@ static TextDocumentBatchStatistics toBatchStatistics(RequestStatistics statistic
...
-        List<TextAnalyticsError> details) {
-        List<com.azure.ai.textanalytics.models.TextAnalyticsError> detailsList = new ArrayList<>();
-        for (TextAnalyticsError error : details) {
-            detailsList.add(new com.azure.ai.textanalytics.models.TextAnalyticsError(
-                ErrorCodeValue.fromString(error.getCode().toString()),
-                error.getMessage(),
-                error.getTarget(), error.getDetails() == null ? null : setErrors(error.getDetails())));
-        }
-        return detailsList;
+    static void toTextAnalyticsException(com.azure.ai.textanalytics.models.TextAnalyticsError error,
```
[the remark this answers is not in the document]
samvaity: Every TextAnalyticsError returned the service is converted to the sdk equivalent `TextAnalyticsError` i.e uses `toTextAnalyticsError()`.

And in cases of methods taking in a single input if such error is encountered, it is converted to `toTextAnalyticsException()`.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
samvaity: It currently doesn't do that, but I think it should. Will update!

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
samvaity: Error here is guaranteed to have code and message.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/models/ErrorCodeValue.java
```diff
@@ -8,7 +8,7 @@
 import com.fasterxml.jackson.annotation.JsonValue;
 
 /**
- * Defines values for ErrorCodeValue.
+ * Defines values for TextAnalyticsErrorCode.
```
alzimmermsft: Was this a manual change or a regeneration change? If the former this will be overwritten next time the implementation layer is regenerated.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentResult.java
```diff
@@ -64,4 +67,16 @@ public TextAnalyticsError getError() {
     public boolean isError() {
         return isError;
     }
+
+    /**
+     * Throw a {@link TextAnalyticsException} if result has isError true and when a non-error property was accessed.
+     */
+    void throwExceptionIfError() {
```
alzimmermsft: Should this accept a `ClientLogger`? Won't everything calling into this use the `DocumentResult.class` logger?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
```diff
@@ -0,0 +1,68 @@
...
+        return this.target;
+    }
+
+    /**
+     * Gets the String value of TextAnalyticsErrorCode for this exception.
+     *
+     * @return The String value of TextAnalyticsErrorCode for this exception.
+     */
+    public String getErrorCodeValue() {
+        return errorCodeValue;
```
alzimmermsft: Should this be turned back into a `TextAnalyticsErrorCode`? Generally don't want to have to much loose string matching on error codes.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            baseMessage = String.format(Locale.ROOT, "%s %s: {%s}", baseMessage, "ErrorCodeValue",
+                errorCodeValue);
+        }
+
+        if (this.target == null) {
+            return baseMessage;
+        } else {
+            baseMessage = String.format(Locale.ROOT, "%s %s: {%s}", baseMessage, "target", target);
+        }
```
alzimmermsft: Is it possible to have a target without an error code a vice-versa? This seems like it could be cleaned up to check that both aren't 'null' at once and do a full string formatting.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
```diff
@@ -0,0 +1,68 @@
...
+        this.errorCodeValue = errorCodeValue;
+        this.target = target;
+    }
+
+    @Override
+    public String getMessage() {
+        String baseMessage = super.getMessage();
+
+        if (this.errorCodeValue == null) {
+            return super.getMessage();
```
alzimmermsft: `baseMessage` could be returned here instead of retrieving it from the super class again.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            baseMessage = String.format(Locale.ROOT, "%s %s: {%s}", baseMessage, "ErrorCodeValue",
```
alzimmermsft: Let's make 'ErrorCodeValue` a constant

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
```diff
@@ -0,0 +1,68 @@
...
+                errorCodeValue);
+        }
+
+        if (this.target == null) {
+            return baseMessage;
+        } else {
+            baseMessage = String.format(Locale.ROOT, "%s %s: {%s}", baseMessage, "target", target);
```
alzimmermsft: Let's make `target` a constant

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
alzimmermsft: Actually, why isn't this in the format string?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
alzimmermsft: Actually, why isn't this in the format string?

alzimmermsft: Could a `StringBuilder` be used instead of using two `String.format` calls?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentResult.java
```diff
@@ -64,4 +67,16 @@ public TextAnalyticsError getError() {
...
+        if (this.isError()) {
+            throw logger.logExceptionAsError(new TextAnalyticsException(
+                String.format("Error in accessing the property on document id: %s, when %s returned with an error: %s",
+                    this.id, this.getClass().getSimpleName(), this.error.getMessage()),
```
alzimmermsft: Why do the other `String.format` changes use `Locale.ROOT` where this does not?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentResult.java
alzimmermsft: Since this isn't package private it won't be inherited by the subclasses, will this require any additional parameters to be passed?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/models/ErrorCodeValue.java
samvaity: I think the refactor for name change made this change, will revert this.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsException.java
samvaity: It should always have error code but target can be null. Can clean it that way.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
```diff
@@ -88,20 +105,43 @@ static TextDocumentBatchStatistics toBatchStatistics(RequestStatistics statistic
...
-     * Helper method to set error details on {@link TextAnalyticsError}.
+     * Convert the incoming input {@link com.azure.ai.textanalytics.models.TextAnalyticsError}
+     * to a {@link TextAnalyticsException}.
+     *
+     * @param error the {@link com.azure.ai.textanalytics.models.TextAnalyticsError}.
+     * @return the {@link TextAnalyticsException} to be thrown.
+     */
+    static TextAnalyticsException toTextAnalyticsException(com.azure.ai.textanalytics.models.TextAnalyticsError error) {
+        String baseMessage = String.format(Locale.US, "%s: {%s}, %s",
+            "Status Code", HttpURLConnection.HTTP_OK, error.getMessage());
```
annelo-msft: Will this exception always be for 200 only?  A different exception is thrown for 4xx/5xx codes?

It's possible TA service could add another top-level error.  Does this exception type leave room to grow if they do?
Right now they have just these top-level errors:

          "enum": [
            "invalidRequest", // 400
            "invalidArgument", // 200
            "internalServerError", // 500
            "serviceUnavailable"  // 500
          ],

What exception type would we throw if they added another for 200?  Would you inherit from TextAnalyticsException or differentiate it another way?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/models/ErrorCodeValue.java
```diff
@@ -8,7 +8,7 @@
...
  */
 public enum ErrorCodeValue {
```
annelo-msft: We have error code as a string in .NET rather than having an enum for it.  Not sure what the philosophy is on this on the Java side, but in .NET if we did expose this as an enum, we would want a service-specific name for it to avoid type name clashes down the line.  This is a candidate for a common library, but we're still operating under the assumption that we won't get there for TextAnalytics (this could change).

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentResult.java
samvaity: Given that it is an exception to be raised for accessing success attributes on output models, this method would always be accessed in the model classes which exist in the same packages.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
samvaity: > Will this exception always be for 200 only? A different exception is thrown for 4xx/5xx codes?

Yes, we have `HttpResponseException` to deal with the unsuccessful response received with HTTP status code (e.g. 3XX, 4XX, 5XX) from the service request.

>What exception type would we throw if they added another for 200?

For another 200 added from the service side, we can update the `getMessage()` method to handle it accordingly. TextAnalyticsException can still be reused in those cases.

Also, on this, should we consider removing `Status code` from the exception message itself?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/models/ErrorCodeValue.java
samvaity: Correct, this was reverted as it was a code generated file and should not have been changed. 
The enum exposed to the users is named `TextAnalyticsErrorCode`.

annelo-msft: Ah, ok.  I must have read the code incorrectly.

annelo-msft: Do you have the same issue that we have in .NET where it's a breaking change to add values to an enum?

samvaity: No, it wouldn't since in Java we use [ExpandableStringEnum](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/util/ExpandableStringEnum.java) that allows this enum to accept any future values that service team could add.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
```diff
@@ -88,20 +104,40 @@ static TextDocumentBatchStatistics toBatchStatistics(RequestStatistics statistic
...
 
     /**
-     * Helper method to set error details on {@link TextAnalyticsError}.
+     * Convert the incoming input {@link com.azure.ai.textanalytics.models.TextAnalyticsError}
+     * to a {@link TextAnalyticsException}.
+     *
+     * @param error the {@link com.azure.ai.textanalytics.models.TextAnalyticsError}.
+     * @return the {@link TextAnalyticsException} to be thrown.
+     */
+    static TextAnalyticsException toTextAnalyticsException(com.azure.ai.textanalytics.models.TextAnalyticsError error) {
```
srnagar: This can now be made `private`

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentResult.java
```diff
@@ -64,4 +69,19 @@ public TextAnalyticsError getError() {
     public boolean isError() {
         return isError;
     }
+
+    /**
+     * Throw a {@link TextAnalyticsException} if result has isError true and when a non-error property was accessed.
+     *
+     * @param logger the {@link ClientLogger} used to log the exception.
+     */
+    void throwExceptionIfError(ClientLogger logger) {
```
srnagar: Don't use a logger created in another class. For debugging purposes, we should know which class actually has the code that logged something.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/ErrorCodeValue.java
```diff
@@ -1,71 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-// Code generated by Microsoft (R) AutoRest Code Generator.
```
srnagar: Any reason why this generated code was deleted?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/models/ErrorCodeValue.java
maririos: >ExpandableStringEnum that allows this enum to accept any future values that service team could add.

Cool :)

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/TextAnalyticsErrorCode.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Enum value invalidCountryHint.
+     */
+    public static final TextAnalyticsErrorCode INVALID_COUNTRY_HINT = fromString("invalidCountryHint");
+
+    /**
+     * Creates or finds a TextAnalyticsErrorCode from its string representation.
+     *
+     * @param name a name to look for.
```
maririos: NIT: `name a name`

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTest.java
```diff
@@ -110,14 +110,12 @@ public void detectLanguagesNullInput() {
     }
 
     /**
-     * Verifies that the error result is returned when empty text is passed.
+     * Verifies that  an TextAnalyticsException is thrown when empty text is passed.
```
maririos:
```suggestion
     * Verifies that a TextAnalyticsException is thrown when empty text is passed.
```

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/ErrorCodeValue.java
samvaity: It was a duplicate file, the generated file still exists [here](https://github.com/samvaity/azure-sdk-for-java/blob/update-error-model/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/models/ErrorCodeValue.java).

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
```diff
@@ -88,20 +104,40 @@ static TextDocumentBatchStatistics toBatchStatistics(RequestStatistics statistic
...
-    private static List<com.azure.ai.textanalytics.models.TextAnalyticsError> setErrors(
-        List<TextAnalyticsError> details) {
-        List<com.azure.ai.textanalytics.models.TextAnalyticsError> detailsList = new ArrayList<>();
-        for (TextAnalyticsError error : details) {
-            detailsList.add(new com.azure.ai.textanalytics.models.TextAnalyticsError(
-                ErrorCodeValue.fromString(error.getCode().toString()),
-                error.getMessage(),
-                error.getTarget(), error.getDetails() == null ? null : setErrors(error.getDetails())));
+    static <T extends DocumentResult> Response<T> processSingleResponseErrorResult(
+        Response<DocumentResultCollection<T>> response, ClientLogger logger) {
```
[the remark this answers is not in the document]
samvaity: Updated to use the class logger.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/AnalyzeSentimentAsyncClient.java
```diff
@@ -55,7 +57,7 @@
 
         return analyzeBatchSentimentWithResponse(
             Collections.singletonList(new TextDocumentInput("0", text, language)), null, context)
-            .map(response -> new SimpleResponse<>(response, response.getValue().iterator().next()));
+            .map(response -> processSingleResponseErrorResult(response));
```
srnagar: nit: can use method reference instead 
```suggestion
            .map(Transforms::processSingleResponseErrorResult);
```

PR #7667, opened 2020-01-23
temp ci.yml changes to test tools update

JimSuplizio:
These files will not be checked in. They only contain a refs tag so I can test changes to the release pipeline from my private tools branch.


JimSuplizio:
Killing this current PR

PR #7669, opened 2020-01-23
Update Module-Info Jackson Requires

alzimmermsft:
Removes Jackson as a `requires transitive` dependency in Azure Core and updated all modules that were using it to `requires` the portion of Jackson they need.

### DIFF
--- a/sdk/core/azure-core/src/main/java/module-info.java
+++ b/sdk/core/azure-core/src/main/java/module-info.java
@@ -6,10 +6,10 @@ module com.azure.core {
     requires transitive org.reactivestreams;
     requires transitive org.slf4j;
 
+    requires transitive com.fasterxml.jackson.annotation;
     requires transitive com.fasterxml.jackson.core;
     requires transitive com.fasterxml.jackson.databind;
 
-    requires com.fasterxml.jackson.annotation;
     requires com.fasterxml.jackson.dataformat.xml;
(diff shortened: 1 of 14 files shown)


sdk/core/azure-core/src/main/java/module-info.java
```diff
@@ -2,19 +2,17 @@
...
-    requires java.xml;
+    requires transitive reactor.core;
+    requires transitive org.reactivestreams;
+    requires transitive org.slf4j;
 
     requires transitive com.fasterxml.jackson.core;
-    requires transitive com.fasterxml.jackson.annotation;
     requires transitive com.fasterxml.jackson.databind;
+
+    requires com.fasterxml.jackson.annotation;
```
srnagar: Can you add `transitive` for jackson annotation too? Any client library requiring azure-core can also get this without having to explicitly add this in their own modules.

sdk/keyvault/azure-security-keyvault-certificates/src/main/java/module-info.java
```diff
@@ -2,9 +2,10 @@
 // Licensed under the MIT License.
 
 module com.azure.security.keyvault.certificates {
-
     requires transitive com.azure.core;
 
+    requires com.fasterxml.jackson.annotation;
```
srnagar: This can be removed if you add transitive in core.

PR #7670, opened 2020-01-23
Move dependencies.json under the eng folder

JimSuplizio:
As per @JonathanGiles request I'm moving dependencies.json out of the root folder under eng\DependencyChecker. I'm also renaming the file to client_dependencies.json (which gets renamed back to dependencies.json when it's copied into the input/ directory that dependency checker needs to run. The reason for the rename is that, in the future, we may want to run the dependency checker against more than just client.

### DIFF
--- a/eng/pipelines/aggregate-reports.yml
+++ b/eng/pipelines/aggregate-reports.yml
@@ -52,7 +52,7 @@ jobs:
 
     - script: |
         git clone https://github.com/JonathanGiles/DependencyChecker.git
-        mkdir input && cp dependencies.json input/
+        mkdir input && cp eng/DependencyChecker/client_dependencies.json input/dependencies.json
       displayName: 'Download dependency checker'
 
     - task: Maven@3


eng/pipelines/client.yml
```diff
@@ -117,7 +117,7 @@ jobs:
 
       - script: |
           git clone https://github.com/JonathanGiles/DependencyChecker.git
-          mkdir input && cp dependencies.json input/
+          mkdir input && cp eng/DependencyChecker/client_dependencies.json input/dependencies.json
```
weshaggard: Does the relative paths to the pomUrls in the file also need to be updated?

JimSuplizio: The contents of the dependency file do not need to change, I just changed where it's being copied from. It's still going to be copied into the same location that dependency checker needs it to be.

mitchdenny: Now that we have the ```java - aggregate-reports``` pipeline, we should probably make this change there as ```client.yml``` is nearly 100% obsolete. Just need to look at moving a repo wide code coverage collection step across.

JimSuplizio: @mitchdenny, I don't think you saw the second commit (which was in about 30 minutes before your comment). I'd changed it in both places and run the aggregate-reports pipeline to verify.

Issue #7274, opened 2020-01-08
Increase Unit Test Coverage for Blob Cryptography


alzimmermsft:
Azure Storage Blob Cryptography has low [unit test coverage](https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-java/test-coverage/index.html). We are looking to increase unit test coverage of each module to introduce a failing step if a certain threshold isn't met.

Aim to increase unit test coverage to the following:
- [ ] 40%: Baseline
- [ ] 70%: Preferred
- [ ] 80%+: Excellent

PR #7676, opened 2020-01-23
Cryptography Tests Playback

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/7274


sdk/storage/azure-storage-blob-cryptography/src/test/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobDownloadTest.groovy
```diff
@@ -0,0 +1,656 @@
+package com.azure.storage.blob.specialized.cryptography
```
rickle-msft: I'd prefer to not have test files specific to methods. Download tests should be in the file that tests other methods on EncryptedBlob

gapra-msft: I'll move em

PR #7681, opened 2020-01-24
Upgrade Jackson and JaCoCo dependencies to next version

srnagar:
Newer versions of Jackson and Jacoco jars are available. This change is to switch to the latest version.

### DIFF
--- a/eng/jacoco-test-coverage/pom.xml
+++ b/eng/jacoco-test-coverage/pom.xml
@@ -159,7 +159,7 @@
       <plugin>
         <groupId>org.jacoco</groupId>
         <artifactId>jacoco-maven-plugin</artifactId>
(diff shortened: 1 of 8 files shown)


JimSuplizio:
@srnagar, I'm fine being on more recent versions of things but I am curious as to why we're updating. Is there any particular reason that we're updating the dependency versions?

srnagar:
@JimSuplizio MSAL is one of the libraries that we depend on they have upgraded to the newer version of Jackon and we want to be in sync with the latest transitive versions used by any of our direct dependencies.

For JaCoCo, it's just an internal dependency that provides better test coverage reporting.

kushagraThapar:
Failed CIs are known issues, flaky tests. Change looks good to me.

Issue #6838, opened 2019-12-13
[BUG] conditional upsert fails via CosmosSyncContainer


akalfas:
**Describe the bug**

I'm using the sync sdk from v 3.3.3 - and I would like to use some sort of a optimistic locking via the `AccessCondition` construct.

While doing so, I've got back a conflict error and digging in the code I've noticed  - https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/sync/CosmosSyncContainer.java#L189

I can work around the issue via the async sdk, but would like to know if it's a bug or just a intended behaviour. 

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Bug Description Added
- [x] Repro Steps Added
- [x] Setup information Added

kushagraThapar:
@akalfas Can you please provide the stack trace for the conflict ?
I am not sure what you meant by code fragment link, can you please elaborate on that ?

akalfas:
@kushagraThapar  Here is a code snippet that can be used to reproduce the issue that I've described:

```
    ConnectionPolicy connectionPolicy = new ConnectionPolicy()
        .connectionMode(ConnectionMode.GATEWAY);

    CosmosSyncClient client = CosmosClient.builder()
        .endpoint("REDACTED")
        .key("REDACTED")
        .connectionPolicy(connectionPolicy)
        .consistencyLevel(ConsistencyLevel.SESSION)
        .buildSyncClient();

    CosmosSyncDatabaseResponse dbResponse = client.createDatabaseIfNotExists("test-db");

    CosmosSyncDatabase database = dbResponse.database();

    CosmosSyncContainerResponse containerResponse = database
        .createContainerIfNotExists("test-coll", "/pk");

    CosmosSyncContainer container = containerResponse.container();

    String id = UUID.randomUUID().toString();

    ObjectMapper mapper = new ObjectMapper();

    ObjectNode item = mapper.createObjectNode();
    item.put("id", id);
    item.put("pk", id);
    item.put("attr", "before");

    CosmosSyncItemResponse itemResponse = container.upsertItem(item);

    String etag = itemResponse.properties().etag();

    CosmosItemRequestOptions option = new CosmosItemRequestOptions();

    AccessCondition condition = new AccessCondition()
        .type(AccessConditionType.IF_MATCH)
[...]

PR #7692, opened 2020-01-24
Fixed upsertItem API to use upsertItem instead of createItem()

kushagraThapar:
* Fixes this issue: https://github.com/Azure/azure-sdk-for-java/issues/6838


kushagraThapar:
> Thanks for fixing it.
> 
> we should have tests covering v3/v4 public surface area if we already don't (not in the scope of this PR though just something to consider in future)

Good point, we definitely should have it. I will create a github item for this at least for v4.

kushagraThapar:
> > Thanks for fixing it.
> > we should have tests covering v3/v4 public surface area if we already don't (not in the scope of this PR though just something to consider in future)
> 
> Good point, we definitely should have it. I will create a github item for this at least for v4.

We can track it here now: https://github.com/Azure/azure-sdk-for-java/issues/7698

PR #7699, opened 2020-01-24
Reverting secondary key change in testconfig

mbhaskar:


mbhaskar:
> doesn't emulator have a secondary key?

I don't see a secondary key in emulator portal, so I think it doesn't.

PR #7701, opened 2020-01-24
Add missing header field in CBS request.

conniey:
CBS occasionally fails with an invalid signature exception. This fixes the missing header value.

### DIFF
--- a/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/handler/SessionHandler.java
+++ b/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/handler/SessionHandler.java
@@ -84,7 +84,9 @@ public class SessionHandler extends Handler {
 
     @Override
     public void onSessionLocalClose(Event e) {
-        final ErrorCondition condition = e.getSession().getCondition();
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ClaimsBasedSecurityChannel.java
```diff
@@ -46,6 +48,7 @@ public ClaimsBasedSecurityChannel(Mono<RequestResponseChannel> responseChannelMo
                     final Message request = Proton.message();
                     final Map<String, Object> properties = new HashMap<>();
                     properties.put(PUT_TOKEN_OPERATION, PUT_TOKEN_OPERATION_VALUE);
+                    properties.put(PUT_TOKEN_EXPIRY, Date.from(accessToken.getExpiresAt().toInstant()));
```
srnagar: Would be great to have tests that validate all the properties that are in CBS channel.

conniey: I'll add one.

PR #7702, opened 2020-01-24
Improve performance of default logger

srnagar:
This change is to improve the performance of the default logger which will be used in the absence of SLF4J binding. The changes include the following:
- Removed AZURE_LOG_LEVEL environment variable loading for each logging attempt. Instead, the environment variable is loaded once through global configuration and will be used for the duration of the application. Any updates will require application restart which is in line with how SLF4J log bindings work.
- Removed String.format() usage which is not very efficient
https://stackoverflow.com/a/513705/311483
- Removed cyclic dependency on DefaultLogger and Configuration

Performance numbers comparing SLF4J logger with logback binding, ClientLogger with logback binding and ClientLogger using DefaultLogger (when there's no SLF4J binding)

  | SLF4J with   logback binding     (ops/second) | ClientLogger   with logback binding     (ops/second) | SLF4J with   Logback/ClientLogger with Logback | ClientLogger   without binding (using DefaultLogger)     (ops/second) | SLF4J with   Logback/ClientLogger with DefaultLogger
-- | -- | -- | -- | -- | --
logDebug_CheckLevel_DebugIsDisabled | 501866764 | 354051250 | 1.42 | 557664530 | 0.90
logDebug_DebugIsDisabled | 446112655 | 415690208 | 1.07 | 708616204 | 0.63
logInfo_CheckLevel_InfoIsEnabled | 5410 | 5483 | 0.99 | 6382 | 0.85
logInfo_InfoIsEnabled | 6469 | 6796 | 0.95 | 6312 | 1.02
logInfoWithArgs_InfoIsEnabled | 4655 | 4608 | 1.01 | 5232 | 0.89

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
@@ -34,7 +34,11 @@ public final class DefaultLogger extends MarkerIgnoringBase {
     public static final String TRACE = "TRACE";
 
     private final String classPath;
-    private final int configuredLogLevel;
+    private final boolean isTraceEnabled;
+    private final boolean isDebugEnabled;
+    private final boolean isInfoEnabled;
+    private final boolean isWarnEnabled;
+    private final boolean isErrorEnabled;
 
     /**
      * Construct DefaultLogger for the given class.
@@ -59,9 +63,16 @@ public final class DefaultLogger extends MarkerIgnoringBase {
             classPath = className;
         }
         this.classPath = classPath;
-        this.configuredLogLevel =
+        int configuredLogLevel =
             LogLevel.fromString(Configuration.getGlobalConfiguration().get(Configuration.PROPERTY_AZURE_LOG_LEVEL))
                 .getLogLevel();
+
+        isTraceEnabled = LogLevel.VERBOSE.getLogLevel() > configuredLogLevel;
+        isDebugEnabled = LogLevel.VERBOSE.getLogLevel() >= configuredLogLevel;
+        isInfoEnabled = LogLevel.INFORMATIONAL.getLogLevel() >= configuredLogLevel;
+        isWarnEnabled = LogLevel.WARNING.getLogLevel() >= configuredLogLevel;
+        isErrorEnabled = LogLevel.ERROR.getLogLevel() >= configuredLogLevel;
+
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -45,11 +52,16 @@ public DefaultLogger(Class<?> clazz) {
...
-            this.classPath = Class.forName(className).getCanonicalName();
+            classPath = Class.forName(className).getCanonicalName();
         } catch (ClassNotFoundException e) {
-            this.classPath = className;
+            classPath = className;
         }
+        this.classPath = classPath;
+        this.configuredLogLevel =
+            LogLevel.fromString(Configuration.getGlobalConfiguration().get(Configuration.PROPERTY_AZURE_LOG_LEVEL))
+                .getLogLevel();
```
JonathanGiles: `configuredLogLevel` is only ever set here in the constructor. I wonder if it is worth doing a micro-optimisation here to determine which log levels are valid and pre-compute the boolean conditions here, to simplify the `isTraceEnabled`, etc methods?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
```diff
@@ -328,9 +332,22 @@ private void logMessageWithFormat(String levelName, String format, Object... arg
     private void log(String levelName, String message, Throwable t) {
         String dateTime = getFormattedDate();
         String threadName = Thread.currentThread().getName();
-        StringBuilder buf = new StringBuilder(64);
-        buf.append(String.format(MESSAGE_TEMPLATE, dateTime, threadName, levelName, classPath, message));
-        writeWithThrowable(buf, t);
+        StringBuilder stringBuilder = new StringBuilder();
```
moderakh: if the total length is known you can pass that to the constructor of StringBuilder to avoid re-growing cost.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java
srnagar: Yeah, I had initially done it that way to have booleans for each of the log levels but the performance improvements were not noticeable by doing that way vs the current approach. 

Moreover, `is*Enabled` methods in DefaultLogger are much faster than their logback counterparts (557664530 ops/sec vs 501866764 ops/sec). So, didn't want to micro-optimize. 

However, if you think evaluating it once in the constructor will improve overall readability, I am fine with making this change.

Issue #7597, opened 2020-01-22
[BUG] Distributed tracing: EventHubs send span is sent per try (should be one per user call)


lmolkova:
**Describe the bug**
EventHubs  "send" span [is created per each try](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java#L425). 

***Exception or Stack Trace***
N/A

**To Reproduce**
Configure tracing and use EventHub. When Send call is retried, two spans are created.

***Code Snippet***
N/A

**Expected behavior**
However it is a convenience layer span. One span should correspond to one public API call and wrap all retried done under it (we use this approach for e.g. storage). If we'll ever decide to instrument AMQP level, we can start tracing retries with child spans.

Python, .NET, and JS instrument public API call (not each try).

**Screenshots**

**Setup (please complete the following information):**
 - OS: any
 - IDE : any
 - Version of the Library used: azure-messaging-eventhubs 5.0.1

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Bug Description Added
- [ ] Repro Steps Added
- [x] Setup information Added

joshfree:
setting milestone to February

PR #7704, opened 2020-01-24
Fix send span to encapsulate multiple send attempts (retries)

samvaity:
Update in eventhubs to start a send span only once per user send and not for every retry.

Closes: #7597

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
@@ -27,7 +27,6 @@ import org.apache.qpid.proton.amqp.messaging.MessageAnnotations;
 import org.apache.qpid.proton.message.Message;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
-import reactor.core.publisher.Signal;
 
 import java.io.Closeable;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -416,10 +416,11 @@ public String getEventHubName() {
         }
 
         final Context finalSharedContext = sharedContext != null ? sharedContext : Context.NONE;
-
         return withRetry(
             getSendLink(batch.getPartitionId()).flatMap(link -> {
-                if (isTracingEnabled) {
+                // if parent context already has send span context data (in case of retries),
+                // don't start a new send span
+                if (isTracingEnabled && !parentContext.get().getData(HOST_NAME_KEY).isPresent()) {
```
lmolkova: does it mean we'll send one span for the first try only? 
It should be one span per Send call. If there is more than one try - it should wrap all of them:
- duration should be duration of send() call 
- result is the result of all tries (i.e. eventually sent or failed)

samvaity: > it should wrap all of them

does this mean we want to add links between all the send spans but start only a single span?

lmolkova: We still create span per message and link all of these spans to the single 'send' span

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -415,31 +414,24 @@ public String getEventHubName() {
...
-                return messages.size() == 1
+        return withRetry(getSendLink(batch.getPartitionId())
+            .flatMap(link ->
+                messages.size() == 1
                     ? link.send(messages.get(0))
-                    : link.send(messages);
-
-            })
+                    : link.send(messages)), retryOptions.getTryTimeout(), retryPolicy)
             .doOnEach(signal -> {
```
lmolkova: is it called on each try? will it end the span after first retry?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
```diff
@@ -415,31 +414,24 @@ public String getEventHubName() {
...
                 if (isTracingEnabled) {
                     tracerProvider.endSpan(parentContext.get(), signal);
                 }
-            })
-            .doOnError(error -> {
```
lmolkova: does it mean we no longer set error if all retries has failed?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
samvaity: No, so with the updated code, it nows does `doOnEach` after the `withRetry` has completed.
So, once withRetry completes either with `error` or `success` , it will send a signal and the `endSpan` uses the signal type to set error or success.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerAsyncClient.java
samvaity: same as [above](https://github.com/Azure/azure-sdk-for-java/pull/7704/files#r376009331)

PR #7733, opened 2020-01-27
Add az namespace span attribute for Storage

samvaity:
[FEATURE REQ] Distributed tracing: set az.namespace attribute, [here](https://github.com/Azure/azure-sdk-for-java/issues/7049)

Client libraries must provide their respective `Microsoft Resource Provider Namespace` to the tracing plugin package for setting the namespace attribute on the outgoing spans.

Approaches considered:
1. Add an overload of the `RestProxy.create` method that takes in the `Map of the tracing attributes` to be set on the outgoing spans. This update, will require the client builders to use this overload and specify the namespace argument to be passed down.

  - **Drawback**:
    - For AutoRest generated clients, this update will result in changing one of the generated files and the changes will be overwritten every time a new code is generated. 

2. Add a `withContext` method overload into `FluxUtil` that takes in a map to specify tracing attributes.
- **Drawback**:
    - This update will require changing the underlying `withContext` overload on all the methods.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/AppendBlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/AppendBlobAsyncClient.java
@@ -163,6 +163,7 @@ public final class AppendBlobAsyncClient extends BlobAsyncClientBase {
     Mono<Response<AppendBlobItem>> createWithResponse(BlobHttpHeaders headers, Map<String, String> metadata,
         BlobRequestConditions requestConditions, Context context) {
         requestConditions = (requestConditions == null) ? new BlobRequestConditions() : requestConditions;
+        context = context == null ? Context.NONE : context;
 
         return this.azureBlobStorage.appendBlobs().createWithRestResponseAsync(null, null, 0, null, metadata,
             requestConditions.getLeaseId(), requestConditions.getIfModifiedSince(),
@@ -237,6 +238,7 @@ public final class AppendBlobAsyncClient extends BlobAsyncClientBase {
         AppendBlobRequestConditions appendBlobRequestConditions, Context context) {
         appendBlobRequestConditions = appendBlobRequestConditions == null ? new AppendBlobRequestConditions()
             : appendBlobRequestConditions;
+        context = context == null ? Context.NONE : context;
 
         return this.azureBlobStorage.appendBlobs().appendBlockWithRestResponseAsync(
             null, null, data, length, null, contentMd5, null, appendBlobRequestConditions.getLeaseId(),
@@ -322,6 +324,7 @@ public final class AppendBlobAsyncClient extends BlobAsyncClientBase {
         } catch (MalformedURLException ex) {
             throw logger.logExceptionAsError(new IllegalArgumentException("'sourceUrl' is not a valid url."));
         }
+        context = context == null ? Context.NONE : context;
 
         return this.azureBlobStorage.appendBlobs().appendBlockFromUrlWithRestResponseAsync(null, null, url, 0,
             sourceRange.toString(), sourceContentMD5, null, null, null, destRequestConditions.getLeaseId(),
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobLeaseAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobLeaseAsyncClient.java
@@ -133,6 +133,7 @@ public final class BlobLeaseAsyncClient {
         Context context) {
         modifiedRequestConditions = (modifiedRequestConditions == null)
             ? new RequestConditions() : modifiedRequestConditions;
+        context = context == null ? Context.NONE : context;
         if (this.isBlob) {
             return this.client.blobs().acquireLeaseWithRestResponseAsync(null, null, null, duration, this.leaseId,
                 modifiedRequestConditions.getIfModifiedSince(), modifiedRequestConditions.getIfUnmodifiedSince(),
@@ -190,6 +191,7 @@ public final class BlobLeaseAsyncClient {
     Mono<Response<String>> renewLeaseWithResponse(RequestConditions modifiedRequestConditions, Context context) {
         modifiedRequestConditions = (modifiedRequestConditions == null)
             ? new RequestConditions() : modifiedRequestConditions;
+        context = context == null ? Context.NONE : context;
         if (this.isBlob) {
             return this.client.blobs().renewLeaseWithRestResponseAsync(null, null, this.leaseId, null,
                 modifiedRequestConditions.getIfModifiedSince(), modifiedRequestConditions.getIfUnmodifiedSince(),
@@ -199,7 +201,7 @@ public final class BlobLeaseAsyncClient {
         } else {
             return this.client.containers().renewLeaseWithRestResponseAsync(null, this.leaseId, null,
                 modifiedRequestConditions.getIfModifiedSince(), modifiedRequestConditions.getIfUnmodifiedSince(),
-                null, context)
+                null, context.addData(AZ_TRACING_NAMESPACE_KEY, STORAGE_TRACING_NAMESPACE_VALUE))
                 .map(rb -> new SimpleResponse<>(rb, rb.getDeserializedHeaders().getLeaseId()));
         }
     }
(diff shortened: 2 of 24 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobAsyncClient.java
```diff
@@ -186,7 +186,7 @@
     public Mono<Response<BlockBlobItem>> uploadWithResponse(Flux<ByteBuffer> data, long length, BlobHttpHeaders headers,
         Map<String, String> metadata, AccessTier tier, byte[] contentMd5, BlobRequestConditions requestConditions) {
         try {
-            return withContext(context -> uploadWithResponse(data, length, headers, metadata, tier, contentMd5,
+            return FluxUtil.withContext(context -> uploadWithResponse(data, length, headers, metadata, tier, contentMd5,
```
alzimmermsft: Did this need to become qualified?

sdk/storage/azure-storage-blob/pom.xml
```diff
@@ -55,7 +55,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core;dependency} -->
```
alzimmermsft: This dependency could be removed as it will be transitively pulled from Storage Common, and really should be pulled from Storage Common.

sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
```diff
@@ -103,6 +103,34 @@ private static void accept(ByteArrayOutputStream byteOutputStream, ByteBuffer by
...
+     * This method converts the incoming {@code subscriberContext} from {@link reactor.util.context.Context Reactor
+     * Context} to {@link Context Azure Context}, adds the specified tracing attributes and calls the given lambda
+     * function with this context and returns a single entity of type {@code T}
+     * <p>
+     * If the reactor context is empty, {@link Context#NONE} will be used to call the lambda function
+     * </p>
+     *
+     * @param serviceCall serviceCall The lambda function that makes the service call into which azure context
+     * will be passed
+     * @param contextAttributes The map of attributes to be set on the tracer spans.
```
srnagar: The map of attributes doesn't necessarily have to be for tracer spans. Update javadoc here to just say  `The map of attributes to be set on {@link Context}`

sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
```diff
@@ -103,6 +103,34 @@ private static void accept(ByteArrayOutputStream byteOutputStream, ByteBuffer by
...
+     * @param  <T> The type of response returned from the service call
+     * @return The response from service call
+     */
+    public static <T> Mono<T> withContext(Function<Context, Mono<T>> serviceCall,
```
srnagar: There's a bit of duplication of code with these `withContext` methods. The `withContext(Function)` overload should just call into this method with an empty map and you should be able to delete `toAzureContext()` method.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobAsyncClient.java
srnagar: Yeah, this qualifier can be removed.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
```diff
@@ -28,6 +33,10 @@
     private static final ClientLogger LOGGER = new ClientLogger(Utility.class);
     private static final String UTF8_CHARSET = "UTF-8";
     private static final String INVALID_DATE_STRING = "Invalid Date String: %s.";
+    // Please see <a href=https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-services-resource-providers>here</a>
+    // for more information on Azure resource provider namespaces.
+    public static final Map<String, String> STORAGE_TRACING_PROPERTIES = Collections.unmodifiableMap(
```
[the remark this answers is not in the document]
samvaity: updated.

sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
samvaity: Updated to call in the `withContext` overload with an empty map. However, `toAzureContext()` is still needed, for `fluxContext()` method.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobLeaseAsyncClient.java
```diff
@@ -121,7 +122,8 @@ public String getLeaseId() {
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Mono<Response<String>> acquireLeaseWithResponse(int duration, RequestConditions modifiedRequestConditions) {
         try {
-            return withContext(context -> acquireLeaseWithResponse(duration, modifiedRequestConditions, context));
+            return withContext(context -> acquireLeaseWithResponse(duration, modifiedRequestConditions, context),
+                STORAGE_TRACING_PROPERTIES);
```
alzimmermsft: Thoughts on changing this from using `withContext` and a `String` constant in Storage Common to making this a helper method in Storage Common?

sdk/storage/azure-storage-blob-cryptography/pom.xml
```diff
@@ -101,11 +106,11 @@
...
-<!--      <groupId>io.projectreactor</groupId>-->
-<!--      <artifactId>reactor-test</artifactId>-->
-<!--      <version>3.3.0.RELEASE</version>--> <!-- {x-version-update;io.projectreactor:reactor-test;external_dependency} -->
-<!--      <scope>test</scope>-->
-<!--    </dependency>-->
+    <!--      <groupId>io.projectreactor</groupId>-->
+    <!--      <artifactId>reactor-test</artifactId>-->
+    <!--      <version>3.3.0.RELEASE</version>--> <!-- {x-version-update;io.projectreactor:reactor-test;external_dependency} -->
+    <!--      <scope>test</scope>-->
+    <!--    </dependency>-->
```
srnagar: This can be deleted.

sdk/storage/azure-storage-blob/pom.xml
```diff
@@ -52,11 +52,6 @@
   </pluginRepositories>
 
   <dependencies>
-    <dependency>
```
srnagar: This shouldn't be deleted as azure-core-test may bring in an older version of azure-core.

sdk/storage/azure-storage-blob-cryptography/pom.xml
```diff
@@ -43,6 +43,11 @@
       <artifactId>azure-storage-blob</artifactId>
       <version>12.5.0-beta.1</version> <!-- {x-version-update;com.azure:azure-storage-blob;current} -->
     </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
```
rickle-msft: I think cryptography should get its core dependency through blobs, no?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/AppendBlobAsyncClient.java
```diff
@@ -152,7 +153,8 @@
     public Mono<Response<AppendBlobItem>> createWithResponse(BlobHttpHeaders headers, Map<String, String> metadata,
         BlobRequestConditions requestConditions) {
         try {
-            return withContext(context -> createWithResponse(headers, metadata, requestConditions, context));
+            return withContext(context -> createWithResponse(headers, metadata, requestConditions, context),
+                STORAGE_TRACING_PROPERTIES);
```
rickle-msft: Since it seems like pretty much every call to `withContext` needs to pass this constant, I think it'd be helpful if we defined something in storage.common that's like withTracingContext or something to automatically add this. Otherwise, I think us Storage folks are liable to forget and mess up tracing later :)

sdk/storage/azure-storage-queue/pom.xml
```diff
@@ -33,11 +33,6 @@
   </scm>
 
   <dependencies>
-    <dependency>
-      <groupId>com.azure</groupId>
```
rickle-msft: Thanks for cleaning up some of this for us :)

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobLeaseAsyncClient.java
samvaity: Updated with a `withTracingContext` in Storage Common that will update the returned serviceCall context with `STORAGE_TRACING_PROPERTIES`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/AppendBlobAsyncClient.java
samvaity: Updated with a `withTracingContext` in Storage Common that will update the returned serviceCall context with `STORAGE_TRACING_PROPERTIES`.

PR #7734, opened 2020-01-27
Code coverage in aggregate reports

mitchdenny:
This PR adds the code coverage generation in aggregate reports again.

### DIFF
--- a/eng/jacoco-test-coverage/pom.xml
+++ b/eng/jacoco-test-coverage/pom.xml
@@ -134,11 +134,6 @@
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
-      <artifactId>azure-storage-blob-nio</artifactId>
-      <version>12.0.0-beta.1</version> <!-- {x-version-update;com.azure:azure-storage-blob-nio;current} -->
-    </dependency>
-    <dependency>
-      <groupId>com.azure</groupId>
       <artifactId>azure-storage-file-share</artifactId>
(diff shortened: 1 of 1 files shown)


JimSuplizio:
/CC @srnagar 
I'm reasonably certain that just adding a coverage run to the aggregate report is going to cause things to go over the timeline, if not all the time at least intermittently. We're generating coverage numbers on every checkin, is there any way we can start uploading those numbers and having the aggregate report pick those up?

srnagar:
> I'm reasonably certain that just adding a coverage run to the aggregate report is going to cause things to go over the timeline, if not all the time at least intermittently. We're generating coverage numbers on every checkin, is there any way we can start uploading those numbers and having the aggregate report pick those up?

Coverage numbers generated for each checkin is only for that specific service as the PR builds use pom.service.xml. This report must contain coverage for all client libraries.

mitchdenny:
When I was looking at this last night I was starting to think the same thing about somehow merging the results. That said I think we can break this up a bit to keep it under time. And we can increase the agent timeout for this one pipeline.

My main goal is to get us back to a steady state and then we can start looking at how we can make better use of agent resources.

mitchdenny:
@JonathanGiles and @srnagar this should get coverage publishing back up and running. There is one more step that I want to do here which is pull the publishing logic directly into this pipeline rather than relying on a release pipeline. However I want to get this merged first and then work on that modification to the pipeline.

Issue #7497, opened 2020-01-16
[QUERY] Get ETag & Last-Modified while using BlobInputStream


michalmela:
**Query/Question**
In `azure-storage-blob` API, is there a concise way to download a file through a `BlobInputStream` while also getting the blob's ETag & Last-Modified header values?

To elaborate:
As the storage APIs are HTTP at heart, it makes sense to use integration patterns typical for HTTP clients in Java.

On one hand, it is convenient to have a properly buffered input stream to read the data from, like what we can get from `com.azure.storage.blob.specialized.BlobClientBase#openInputStream(com.azure.storage.blob.models.BlobRange, com.azure.storage.blob.models.BlobRequestConditions)`.

On the other, one may want to repeatedly store the blob's ETag and Last-Modified response header values to use them in the subsequent requests (in the `If-Modified-Since` and `If-None-Match` request headers), like what we can get from `com.azure.storage.blob.specialized.BlobClientBase#downloadWithResponse`.

However, I cannot see how a developer can make use of both of those features using a single API. Using a `BlobInputStream`, I can't see how I can access the response headers. Using `downloadWithResponse` I need to manage the stream buffering myself, by either building my own version of `BlobInputStream` or buffering in a huge array in-memory through `ByteArrayOutputStream` / `ByteArrayInputStream`. Additionally, this API throws a `com.azure.storage.blob.models.BlobStorageException` in the case of a `304 Not Modified` response.

***Why is this not a Bug or a feature Request?***
This may well become a feature request (perhaps not only for the blob storage but any other, and perhaps not only for the caching headers) but at this point for me it may just be my lack of knowledge of the API or my lack of understanding of an intentional Azure client design decision.

**Setup (please complete the following information if applicable):**
 - com.azure:azure.storage.blob:12.2.0
 
 **Information Checklist**
 Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Query Added
- [x] Setup information Added

gapra-msft:
Thanks @michalmela for your question. 

You're right in that there is no way to access the response headers using BlobInputStream. We could add that feature by providing an object on the BlobInputStreamthat holds the required headers and is available on close of the BlobInputStream. Would that satisfy your use case?

Also, it's a little unclear to me what you mean by 

> Additionally, this API throws a com.azure.storage.blob.models.BlobStorageException in the case of a 304 Not Modified response.

[...]

PR #7739, opened 2020-01-27
[API Change] Added ability to access BlobProperties from BlobInputStream

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/7497


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobInputStream.java
```diff
@@ -89,4 +102,26 @@ protected synchronized ByteBuffer dispatchRead(final int readLength, final long
...
+            hd.getContentEncoding(), hd.getContentDisposition(), hd.getContentLanguage(), hd.getCacheControl(),
+            hd.getBlobSequenceNumber(), hd.getBlobType(), hd.getLeaseStatus(), hd.getLeaseState(),
+            hd.getLeaseDuration(), hd.getCopyId(), hd.getCopyStatus(), hd.getCopySource(), hd.getCopyProgress(),
+            hd.getCopyCompletionTime(), hd.getCopyStatusDescription(), hd.isServerEncrypted(),
+            null, null, null, null, null, hd.getEncryptionKeySha256(), null, hd.getMetadata(),
+            hd.getBlobCommittedBlockCount());
+    }
+
+    /**
+     *
```
rickle-msft: Extra space.

Looking at this, I wonder if it actually would be good to do a getProperties call if they try to access the properties before reading any data instead of returning null. What do you think?

gapra-msft: Good idea, I'll change it to do that and document it

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlockBlobInputOutputStreamTest.groovy
```diff
@@ -25,6 +26,7 @@ class BlockBlobInputOutputStreamTest extends APISpec {
 
         then:
         def inputStream = bc.openInputStream()
+        !inputStream.getProperties() // before reading properties should be null
```
gapra-msft: need to update this method

Issue #6592, opened 2019-12-01
azure-storage-file-share:12.0.0-preview.5 - ShareDirectoryAsyncClient.getFileClient() should not append "/" for root directory


philiptran:
azure-storage-file-share:12.0.0-preview.5:

`ShareDirectoryAsyncClient.getFileClient(String fileName)` appends "/" to the filePath for root directory , causing Azure to throw error 400 due to malformed URL as observed in the log.

`onStateChange(PUT{uri=/<shareName>/%2f1.jpg, connection=PooledConnection{channel=[id: 0...`

To reproduce:
```
ShareDirectoryClient directoryClient = shareClient.getRootDirectoryClient();
File file = new ClassPathResource("dms/1.jpg").getFile();
try (FileInputStream fis = new FileInputStream(file)) {
  ShareFileHttpHeaders httpHeaders = new ShareFileHttpHeaders().setContentType("image/jpg")
        .setContentDisposition("attachment");
  Response<ShareFileClient> res = directoryClient.createFileWithResponse(file.getName(), file.length(),
      httpHeaders, null, null, null, null, Context.NONE);
  if (res.getStatusCode() != HttpStatus.CREATED.value()) {
    throw new AzureStorageException(res.getStatusCode(), "Could not create Azure file: " + file.getName());
  } else {
    res.getValue().upload(fis, file.length());
  }
} catch (IOException | ShareStorageException ex) {
  throw new AzureStorageException("Could not upload file: " + file.getName(), ex);
}
```

joshfree:
Thanks for reporting this issue @philiptran.  The right storage team contacts will follow up with you shortly.

/cc @rickle-msft

rickle-msft:
@philiptran Thank you for reporting this. We will work on scheduling some time to get a fix in for this.

PR #7748, opened 2020-01-27
Fixed bug where root directory client could not be used to create files

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/6592


sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareDirectoryAsyncClient.java
```diff
@@ -134,6 +134,9 @@ public ShareServiceVersion getServiceVersion() {
      */
     public ShareFileAsyncClient getFileClient(String fileName) {
         String filePath = directoryPath + "/" + fileName;
+        if (directoryPath.isEmpty()) {
+            filePath = fileName;
```
rickle-msft: Can you add a comment that this is to accommodate the root?

gapra-msft: done

PR #7749, opened 2020-01-27
Nio path2

rickle-msft:


sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -58,162 +120,325 @@ public Path getRoot() {
...
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
     public Path getName(int i) {
-        return null;
+        if (i < 0 || i >= this.getNameCount()) {
+            throw new IllegalArgumentException();
```
rickle-msft: Should log these and others

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -38,18 +89,29 @@ public FileSystem getFileSystem() {
...
+     * The root component of this path also identifies the Azure Storage Container in which the file is stored. This
+     * method will not validate that the root component corresponds to an actual file store/container in this
+     * file system. It will simply return the root component of the path if one is present and syntactically valid.
+     *
      * {@inheritDoc}
      */
     @Override
     public Path getRoot() {
+        // Check if the first element of the path is formatted like a root directory.
+        String firstElement = pathString.split(parentFileSystem.getSeparator())[0];
```
rickle-msft: Should use splitToElements

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -58,162 +120,325 @@ public Path getRoot() {
...
      * {@inheritDoc}
      */
     @Override
     public boolean startsWith(Path path) {
-        return false;
+        /*
+        There can only be one instance of a file system with a given id, so comparing object identity is equivalent
+        to checking ids here.
+         */
+        if (path.getFileSystem() != this.parentFileSystem) {
```
rickle-msft: Just use .equals. It's clear and safer and evidently equivalent.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -58,162 +120,325 @@ public Path getRoot() {
...
+            return false;
+        }
+
+        String[] thisPathElements = this.splitToElements();
+        String[] otherPathElements = ((AzurePath) path).splitToElements();
+        if (otherPathElements.length > thisPathElements.length) {
+            return false;
+        }
+        for (int i = 0; i < otherPathElements.length; i++) {
+            if (!otherPathElements[i].equals(thisPathElements[i])) {
```
rickle-msft: What about a path with ".."? In other words, should we normalize first?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -58,162 +120,325 @@ public Path getRoot() {
...
+     * This file system follows the standard practice mentioned in the original docs.
+     *
      * {@inheritDoc}
      */
     @Override
     public Path normalize() {
-        return null;
+        Deque<String> stack = new ArrayDeque<>();
+        String[] pathElements = this.splitToElements();
+        Path root = this.getRoot(); // Refactor so this doesn't access the fs per docs
```
rickle-msft: Can remove the todo as the root validity check is gone.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -226,14 +451,89 @@ public WatchKey register(WatchService watchService, WatchEvent.Kind<?>... kinds)
...
+        return str;
+    }
+
+    private String[] splitToElements() {
+        return this.splitToElements(this.pathString);
+    }
+
+    private String[] splitToElements(String str) {
+        String[] arr = str.split(this.parentFileSystem.getSeparator());
+        if (arr.length == 1 && arr[0].isEmpty()) {
```
rickle-msft: Add a comment here to explain that this is the special case where we split after removing the root of a path that is just the root.

rickle-msft:
Need to add recordings

sdk/storage/ci.yml
```diff
@@ -57,6 +57,9 @@ stages:
           groupId: com.azure
           safeName: azurestorageblobcryptography
           stagingProfileId: 88192f04117501
+        - name: azure-storage-blob-nio
+          safeName: azurestorageblobnio
```
JimSuplizio: This needs the groupId entry which should be
**groupId: com.azure**

sdk/storage/azure-storage-blob-nio/DesignDoc.md
```diff
@@ -0,0 +1,235 @@
+# Azure Storage NIO Design Doc
```
gapra-msft: This doesn't need to be reviewed, just a translation of the original design doc

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystem.java
```diff
@@ -215,6 +215,9 @@ public String getSeparator() {
     }
 
     /**
+     * ':' is only allowed in the root component. Root component will be validated against the list of
+     * file shares?
```
gapra-msft: is there a reason for the ? or am I misreading this sentence.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -3,30 +3,81 @@
...
+import java.util.Deque;
 import java.util.Iterator;
+import java.util.Objects;
 
 /**
+ * The root component, if it is present, is the first element of the path and is denoted by a {@code ':'} as the last
+ * character. Hence, only one instance of {@code ':'} may appear in a path string and it may only be the last character
+ * of the first element in the path. The root component is used to identify which container a path belongs to.
+ * <p>
+ * Constructing a syntactically valid path does not ensure a resource exists at the given path. An error will
```
gapra-msft: Could this be made into more of a bulleted list, kind of like

An error will only be thrown for 
- an invalid root when trying to access a file in that root if it does not exist
- a nonexistent path when trying to access the resource at that location if it does not exist or is inaccessible

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -3,30 +3,81 @@
...
  * {@inheritDoc}
  */
 public final class AzurePath implements Path {
+    private final ClientLogger logger = new ClientLogger(AzurePath.class);
+    private static final String ROOT_DIR_SUFFIX = ":";
+
     private final AzureFileSystem parentFileSystem;
     private final String pathString;
 
     AzurePath(AzureFileSystem parentFileSystem, String s, String... strings) {
```
gapra-msft: can the name of the strings and s variable be more descriptive? are they paths?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -58,162 +120,325 @@ public Path getRoot() {
...
+     * <p>
      * {@inheritDoc}
      */
     @Override
     public Path toRealPath(LinkOption... linkOptions) throws IOException {
-        return null;
+        throw new UnsupportedOperationException();
     }
 
     /**
```
gapra-msft: Add Unsupported in javadoc

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
gapra-msft: I think it makes sense to normalize before comparing

rickle-msft: It looks like the default system (which I'm generally trying to follow) doesn't normalize and will return false for "foo/bar" starts with "foo/.", so I'll leave this as is.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -58,182 +119,422 @@ public Path getRoot() {
      */
     @Override
     public Path getFileName() {
-        return null;
+        if (this.withoutRoot().isEmpty()) {
+            return null;
+        } else {
+            return this.parentFileSystem.getPath(Flux.fromArray(this.splitToElements()).last().block());
```
alzimmermsft: Any reason to use a reactive stream here only to block it? Instead could a `Stream` be used? Another option, based on the logic here, we could have a local instance of the `String[]` and just access and return the last element in the list.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -58,182 +119,422 @@ public Path getRoot() {
...
+        }
+        if (str.startsWith(this.parentFileSystem.getSeparator())) {
+            str = str.substring(1);
+        }
+
+        return str;
+    }
+
+    private String[] splitToElements() {
+        return this.splitToElements(this.pathString);
```
alzimmermsft: Given there is numerous calls to this method should the split pathString be stored as a field on the instance?

rickle-msft: I was wondering that as I wrote it. The consequence is that these objects basically double their memory usage, right? Since we would then store the pathString as one object and then all of its components again separately. I didn't feel like I had enough knowledge of customer scenarios to know if they were cpu or memory bound, so I was going to defer that decision until we hear one way or the other from customers unless you feel strongly about it. Does that make sense?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -3,30 +3,81 @@
...
     private final String pathString;
 
     AzurePath(AzureFileSystem parentFileSystem, String s, String... strings) {
+        if (strings == null) {
+            strings = new String[0]; // Just to make processing easier later. This wont affect the result.
+        }
         this.parentFileSystem = parentFileSystem;
-        this.pathString = String.join(this.parentFileSystem.getSeparator(),
-                Flux.just(s).concatWith(Flux.just(strings)).toIterable());
+        Flux<String> elementFlux =
```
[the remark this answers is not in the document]
rickle-msft: Doesn't have to be a Flux. I can look into switching to a Stream. I'm just more familiar with Flux apis. That's a good point about possible usage in a reactive context.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
rickle-msft: I think I addressed the Stream concern above and the String[] concern below :)

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -3,30 +3,80 @@
...
+            // Strip any trailing, leading, or internal delimiters so there are no duplicates when we join.
+            Flux.fromArray(first.split(this.parentFileSystem.getSeparator()))
+                .concatWith(Flux.fromArray(more)
+                    .flatMap(str -> Flux.fromArray(str.split(this.parentFileSystem.getSeparator()))))
+                .filter(str -> !str.isEmpty());
+
+        this.pathString = String.join(this.parentFileSystem.getSeparator(), elementFlux.toIterable());
+
+        // No element but the first may contain ":"
+        elementFlux.skip(1)
```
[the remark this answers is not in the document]
rickle-msft: I do have tests that cover creating a path with only one element, just a root component, and the root component plus one name element. None of them have any problems, so I'm guessing the Flux just completes if it skips 1 and there's only one element.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -3,30 +3,80 @@
...
+        }
         this.parentFileSystem = parentFileSystem;
-        this.pathString = String.join(this.parentFileSystem.getSeparator(),
-                Flux.just(s).concatWith(Flux.just(strings)).toIterable());
+        Flux<String> elementFlux =
+            // Strip any trailing, leading, or internal delimiters so there are no duplicates when we join.
+            Flux.fromArray(first.split(this.parentFileSystem.getSeparator()))
+                .concatWith(Flux.fromArray(more)
+                    .flatMap(str -> Flux.fromArray(str.split(this.parentFileSystem.getSeparator()))))
+                .filter(str -> !str.isEmpty());
```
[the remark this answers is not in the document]
rickle-msft: I'll switch over the implementation.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -3,30 +3,80 @@
...
+                ? Mono.error(Utility.logError(logger, new InvalidPathException(this.pathString, ROOT_DIR_SUFFIX
+                + " is an invalid character except to identify the root element of this path if there is one.")))
+            : Mono.just(str)).blockLast();
+
+        // There may only be at most one instance of ":" in the root component, and it must be the last character.
+        elementFlux.take(1)
+            .flatMap(str -> str.contains(ROOT_DIR_SUFFIX) && str.indexOf(ROOT_DIR_SUFFIX) < str.length() - 1
+            ? Mono.error(Utility.logError(logger, new InvalidPathException(this.pathString, ROOT_DIR_SUFFIX + " may"
+                + " only be used as the last character in the root component of a path")))
+            : Mono.just(str)).blockLast();
```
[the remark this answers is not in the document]
rickle-msft: Basically I'm mimicking what windows does with its drives/root directories--using <driveName>: to double as the name of the root directory. The ':' helps identify that an element is meant to be the root instead of just any old path element. The offshoot as that ':' can only appear in the first element, otherwise it loses its value in distinguishing the root. If it does appear in the first element, it can only appear as the last character in that element. Does that answer your question? The `FileSystem.getPath()` test cases may also help clarify this. I can clean up the inline docs, too, if you think that would help clarify the logic

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
alzimmermsft: That makes sense to me

PR #7770, opened 2020-01-28
disable tests because react-netty aggregate random order of response …

mssfang:
Disabled listing tests because of unstable react-netty aggregate random order of responses.

Opened a tracking issue for these disabled tests: #7771

### DIFF
--- a/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationAsyncClientTest.java
+++ b/sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationAsyncClientTest.java
@@ -726,7 +726,7 @@ public class ConfigurationAsyncClientTest extends ConfigurationClientTestBase {
                 .verifyComplete();
     }
 
-    @Disabled("reactor netty aggregate method aggregate random order of context")
+    @Disabled("reactor netty aggregate method aggregate random order of context. Tracking issue #7771")
     /**
      * Verifies that, given a ton of revisions, we can list the revisions ConfigurationSettings using pagination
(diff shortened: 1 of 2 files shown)


alzimmermsft:
Not a fan of disabling failing tests due to a legitimate issue in the SDK/underlying infrastructure, this feels like a scenario where we say we'll get around to resolving the underlying issue soon but forget until we get a customer reported issue.

If we end up merging this please submit an issue for the underlying problem and get it triaged appropriately so we don't let the problem fester.

JonathanGiles:
I agree with @alzimmermsft. I do not want to be in the business of disabling tests eagerly. I would like to understand more about what the issue is and why we can't fix the underlying issue, before we disable these tests.

mssfang:
Anu and I had investigated the failure of tests. I am also not a fan of disabling tests. But since it caused by the react-netty aggregate response content in random order, we can't really solve this problem immediately and we want to make sure other tests work property.  Before we file an issue to react-netty, Anu is investigating it and double confirm.

JonathanGiles:
Have these tests ever successfully passed in successive runs, or have they always been flakey?

mssfang:
@JonathanGiles  These tests some times failed some times successfully passed.

JonathanGiles:
@mssfang To avoid any confusion, this PR is now pending you making the changes that Josh has requested.

mssfang:
@JonathanGiles  Yes. Anu wants to have a final verified before file an issue. Will file an issue and merged it after confirmed.

mssfang:
So we are going to merge this PR. Ran a live test pipeline on this PR. All other tests look good. We are keeping investigating it tomorrow. And leave the issue #7771 open.

anuchandy:
Related PR with fix: https://github.com/Azure/azure-sdk-for-java/pull/7837

PR #7793, opened 2020-01-28
Renamed deny encryption scope override to prevent encryption scope ov…

gapra-msft:
…erride


sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
```diff
@@ -124,8 +125,7 @@ class ContainerAPITest extends APISpec {
         response.getValue().getLeaseState() == LeaseStateType.AVAILABLE
         response.getValue().getLeaseStatus() == LeaseStatusType.UNLOCKED
         response.getValue().getMetadata().size() == 0
-        response.getValue().getDefaultEncryptionScope()
-        response.getValue().hasDenyEncryptionScopeOverride() != null
+        response.getValue().getBlobContainerEncryptionScope()
```
rickle-msft: This returns an object, right? I think you need to check the values on the object.

gapra-msft: done

PR #7805, opened 2020-01-29
Generate kusto 2019_11_09 package.

yaohaizh:


weidongxu-microsoft:
According to mitch, check-enforcer need to have at least one CI (but ./sdk is excluded for mgmt CI, and separate CI is not added yet). I think as a workaround, you can modify some file in ./eng/mgmt/ to trigger mgmt CI.

PR #7831, opened 2020-01-29
Making databaseAccount read dynamic

### LINKED ISSUE #7668: Use DatabaseAccount for configurations read dynamically
**Describe the bug**
Currently DatabaseAccount used to read several configurations example maxReplicacount on read calls is the one time snapshot during app initialization.

**Expected behavior**
DatabaseAccount  uses to read configurations  should refresh dynamically with periodic cache refresh in GlobalEndpointManager

**Information Checklist**
- [x] Bug Description Added
- [ ] Repro Steps Added
- [ ] Setup information Added

simplynaveen20:
Currently GatewayServiceConfigurationReader initialize during client start up with one time snapshot of databaseAccount read via GlobalEndPointManager. Its value remain same throughout the lifespan of application.
This PR will read all the databaseAccount configurations via GatewayServiceConfigurationReader dynamically from cache which updates with its scheduling mechanism or any read/write account error.
This will also fix an issue in which application hits DatabaseAccount read call twice during startup.

fixes #7668

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
@@ -26,7 +26,9 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.micrometer.core.instrument.MeterRegistry;
+import org.apache.commons.lang3.tuple.Pair;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 import java.net.URI;
 import java.time.OffsetDateTime;
@@ -489,5 +491,13 @@ public class BridgeInternal {
 
     public static <T> CosmosItemProperties getProperties(CosmosAsyncItemResponse<T> cosmosItemResponse) {
         return cosmosItemResponse.getProperties();
+    }
+
+    public static PartitionKey partitionKeyfromJsonString(String jsonString) {
+        return PartitionKey.fromJsonString(jsonString);
+    }
+
+    public static Object getPartitionKeyObject(PartitionKey right) {
+        return right.getKeyObject();
     }
 }
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainer.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainer.java
@@ -3,7 +3,6 @@
 
 package com.azure.cosmos;
 
-import com.azure.core.util.IterableStream;
 import reactor.core.Exceptions;
 import reactor.core.publisher.Mono;
 
@@ -13,7 +12,7 @@ import reactor.core.publisher.Mono;
  */
 public class CosmosContainer {
 
-    private final CosmosAsyncContainer asyncContainer;
+    final CosmosAsyncContainer asyncContainer;
     private final CosmosDatabase database;
     private final String id;
     private CosmosScripts scripts;
@@ -167,7 +166,7 @@ public class CosmosContainer {
      * @return the cosmos item response
      * @throws CosmosClientException the cosmos client exception
      */
-    
+
     public <T> CosmosItemResponse<T> createItem(T item, CosmosItemRequestOptions options) throws CosmosClientException {
         return this.mapItemResponseAndBlock(this.asyncContainer.createItem(item, options));
     }
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ItemOperations.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ItemOperations.java
@@ -1,0 +1,60 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.cosmos;
+
+import org.apache.commons.lang3.tuple.Pair;
+import reactor.core.publisher.Mono;
+
+import java.util.List;
+
+/**
+ * Note: although this class is public it is not part of our public API and may change in future.
+ */
+public class ItemOperations {
+
+    /**
+     * Note: although this method is public, it is not part of public API and may change in future.
+     * <p>
+     * Reads many documents.
+     *
+     * @param container   the cosmos async container
+     * @param itemKeyList document id and partition key pair that needs to be read
+     * @param klass       class type
+     * @return a feed response of cosmos items
+     */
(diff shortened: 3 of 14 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReader.java
```diff
@@ -41,118 +41,32 @@
...
     }
 
     public ReplicationPolicy getSystemReplicationPolicy() {
-        this.throwIfNotInitialized();
-        return this.systemReplicationPolicy;
-    }
-
-    public boolean enableAuthorization() {
-        return true;
+        DatabaseAccount databaseAccount = this.globalEndpointManager.getDatabaseAccountFromCache(this.serviceEndpoint).block();
```
simplynaveen20: This is blocking call on cache , which should be fine as per me. But if anyone have any strong opinion for these methods to return Mono and used asynchronously please comment. I avoided it as it needed good amount of code touch in different layer of SDK where these properties are getting used.

moderakh: blocking calls inside the SDK are not good. they can cause deadlock, perf hit, etc. We are not doing blocking calls anywhere else.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReader.java
```diff
@@ -41,118 +41,32 @@
...
 
     public ConsistencyLevel getDefaultConsistencyLevel() {
-        this.throwIfNotInitialized();
-        return this.consistencyLevel;
-    }
-
-    public void setDefaultConsistencyLevel(ConsistencyLevel value) {
-        this.throwIfNotInitialized();
-        this.consistencyLevel = value;
+        DatabaseAccount databaseAccount = this.globalEndpointManager.getDatabaseAccountFromCache(this.serviceEndpoint).block();
```
moderakh: blocking calls inside the SDK are not good. they can cause deadlock, perf hit, etc. We are not doing blocking calls anywhere else.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReaderTest.java
```diff
@@ -118,37 +113,21 @@ public static void validateSuccess(Mono<DatabaseAccount> observable) {
...
-        HttpResponse httpResponse = Mockito.mock(HttpResponse.class);
-        Mockito.doReturn(200).when(httpResponse).statusCode();
-        Mockito.doReturn(Flux.just(ByteBufUtil.writeUtf8(ByteBufAllocator.DEFAULT, databaseAccountJson)))
-                .when(httpResponse).body();
-        Mockito.doReturn(Mono.just(databaseAccountJson))
-                .when(httpResponse).bodyAsString();
+    private GatewayServiceConfigurationReader getServiceConfigurationReader(RxDocumentClientImpl rxDocumentClient) throws Exception {
+        Field serviceConfigurationReaderField = RxDocumentClientImpl.class.getDeclaredField("gatewayConfigurationReader");
+        serviceConfigurationReaderField.setAccessible(true);
+        return (GatewayServiceConfigurationReader) serviceConfigurationReaderField.get(rxDocumentClient);
```
moderakh: we have util helper methods for this:
please see this
`com.azure.cosmos.implementation.directconnectivity.get(.)`

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReaderTest.java
```diff
@@ -118,37 +113,21 @@ public static void validateSuccess(Mono<DatabaseAccount> observable) {
...
+    }
 
-        Mockito.doReturn(new HttpHeaders()).when(httpResponse).headers();
-        return httpResponse;
+    private GlobalEndpointManager getGlobalEndpointManager(GatewayServiceConfigurationReader serviceConfigurationReader) throws Exception {
+        Field globalEndpointManagerField = GatewayServiceConfigurationReader.class.getDeclaredField("globalEndpointManager");
+        globalEndpointManagerField.setAccessible(true);
+        return (GlobalEndpointManager) globalEndpointManagerField.get(serviceConfigurationReader);
```
moderakh: ditto.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReader.java
simplynaveen20: We cant really say  at least  for this scenario that this will do deadlock as all are read calls  , even for perf this is cache and not network .But i got your point of avoiding them in code everywhere  and follow async behavior , I will move them to non blocking

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReaderTest.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReaderTest.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReader.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReader.java
simplynaveen20: done

simplynaveen20:
This work done in https://github.com/Azure/azure-sdk-for-java/pull/8466

Issue #6089, opened 2019-10-30
[Storage] Add API BlobClient.upload(InputStream, length)


mikeharder:
Track 1 provides the following API which automatically switches between single-put and stage-block upload based on the length:

```java
CloudBlockBlob.upload(InputStream, long length)
```

For customers migrating from Track 1 to Track 2, I think we should provide a synchronous API as similar as possible.  Currently in Track 2, there is no way to synchronously upload a stream of arbitrary size with good performance for small blobs.  The only option is `BlockBlobClient.getBlobOutputStream()`, which is both more complicated to use and has poor performance for small blobs.

The implementation can simply call into the existing sync APIs:

```java
public void BlobClient.upload(InputStream inputStream, long length) {
    BlockBlobClient blockBlobClient = this.getBlockBlobClient();
    if (length < singlePutMaximum) {
        return blockBlobClient.upload(inputStream, length);
    }
    else {
        BlobOutputStream blobOutputStream = blockBlobClient.getBlobOutputStream();
        inputStream.transferTo(blobOutputStream);
        blobOutputStream.close();
    }
}
```

Related: https://github.com/Azure/azure-sdk-for-java/issues/5275

CC: @rickle-msft

rickle-msft:
I agree that this would make migration easier and more intuitive. I'm ok with this addition.

PR #7835, opened 2020-01-29
[API change] Added convenient upload method to BlobClient

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/6089

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
@@ -17,7 +17,6 @@ import com.azure.storage.blob.specialized.BlobOutputStream;
 import com.azure.storage.blob.specialized.BlockBlobClient;
 import com.azure.storage.blob.specialized.PageBlobClient;
 import com.azure.storage.blob.specialized.SpecializedBlobClientBuilder;
-import com.azure.storage.common.Utility;
 import com.azure.storage.common.implementation.Constants;
 import com.azure.storage.common.implementation.StorageImplUtils;
 import reactor.core.publisher.Mono;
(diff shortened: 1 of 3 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
```diff
@@ -115,6 +120,70 @@ public PageBlobClient getPageBlobClient() {
...
+        final ParallelTransferOptions validatedParallelTransferOptions =
+            ModelHelper.populateAndApplyDefaults(parallelTransferOptions);
+        if (length < validatedParallelTransferOptions.getMaxSingleUploadSize()) {
+            blockBlobClient.uploadWithResponse(data, length, headers, metadata, tier, null, requestConditions,
+                timeout, context);
+        } else {
+            BlobOutputStream blobOutputStream = BlobOutputStream.blockBlobOutputStream(client,
+                validatedParallelTransferOptions, headers, metadata, tier, requestConditions);
+            try {
+                data.transferTo(blobOutputStream);
```
rickle-msft: I thought this method was only java 9?

gapra-msft: yep, I need to change it

gapra-msft: I made the PR before I realized

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/Utility.java
```diff
@@ -241,4 +244,27 @@ public static OffsetDateTime parseDate(String dateString) {
...
     }
+
+    /**
+     * Reads data from an input stream and writes it to an output stream.
+     * @param source {@link InputStream source}
+     * @param writeLength The length of data to write.
+     * @param destination {@link OutputStream destination}
+     * @throws IOException If an I/O error occurs.
+     */
+    public static void copyToOutputStream(InputStream source, long writeLength, OutputStream destination)
```
alzimmermsft: Could this go into `StorageImplUtils` for the time being? Do we expected or would want a customer to use this API in the near-term?

gapra-msft: I have nothing against doing that

gapra-msft: done

PR #7837, opened 2020-01-29
Ensure BufferedResponse type is not caching un-released buffer

anuchandy:


anuchandy:
#### The issue:

The `BufferedHttpResponse` used to cache only references to the source response body buffers. The buffers will be released underneath as first subscription consumes them. For the future subscriptions these released buffers were severed, this result in decoder and other parts depending on  `BufferedHttpResponse` to receive garbage data.

#### consumers of  `BufferedHttpResponse`:

1. test record interceptor
2. body decoder
3. http logging interceptor

#### fixes in pr:

1. updates the `BufferedHttpResponse` to do deep copy caching of source response body buffers.
2. removes `BufferedHttpResponse` use in "body decoder" (that saves extra allocations)
3. updates `StreamResponse::close()` such that it won't attempt to dispose already consumed buffer source.

No update in "test record interceptor" and is fine to buffer as this is not used in production.

#### future

1. we could improve "http logging interceptor" by not doing buffering. This is not included in this PR.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/http/BufferedHttpResponse.java
```diff
@@ -28,7 +28,10 @@
     public BufferedHttpResponse(HttpResponse innerHttpResponse) {
         super(innerHttpResponse.getRequest());
         this.innerHttpResponse = innerHttpResponse;
-        this.cachedBody = innerHttpResponse.getBody().cache();
+        this.cachedBody = FluxUtil.collectBytesInByteBufferStream(innerHttpResponse.getBody())
+            .map(bytes -> ByteBuffer.wrap(bytes))
```
JonathanGiles:
```suggestion
            .map(ByteBuffer::wrap)
```

sdk/core/azure-core/src/main/java/com/azure/core/implementation/http/BufferedHttpResponse.java
```diff
@@ -53,7 +56,7 @@ public HttpHeaders getHeaders() {
 
     @Override
     public Mono<byte[]> getBodyAsByteArray() {
-        return FluxUtil.collectBytesInByteBufferStream(cachedBody.map(ByteBuffer::duplicate));
+        return cachedBody.next().map(byteBuffer -> byteBuffer.array());
```
JonathanGiles:
```suggestion
        return cachedBody.next().map(ByteBuffer::array);
```

anuchandy:
@srnagar thanks for the review. @JonathanGiles, thanks - your suggestions work, validated locally so committing them.

PR #7839, opened 2020-01-30
Cosmos Paged flux initial POC implementation v4

kushagraThapar:
* CosmosPagedFlux initial POC implementation
* To get early feedback and scope the changes to a small subset of files, this PR only covers one API -> `queryItems()`

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainer.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContainer.java
@@ -3,6 +3,7 @@
 
 package com.azure.cosmos;
 
+import com.azure.core.util.IterableStream;
 import reactor.core.Exceptions;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -254,11 +255,10 @@ public class CosmosContainer {
      * @param query the query
      * @param options the options
      * @param klass the class type
-     * @return the iterator
+     * @return the {@link IterableStream}
      */
-    public <T> Iterator<FeedResponse<T>> queryItems(String query, FeedOptions options, Class<T> klass) {
-        //  TODO: Temporary change for testing, remove byPage() once all APIs are migrated and change getFeedIterator() method to accept com.azure.cosmos.CosmosContinuablePagedFlux
-        return getFeedIterator(this.asyncContainer.queryItems(query, options, klass).byPage());
+    public <T> IterableStream<FeedResponse<T>> queryItems(String query, FeedOptions options, Class<T> klass) {
+        return getFeedIterableStream(this.asyncContainer.queryItems(query, options, klass));
     }
 
     /**
@@ -367,6 +367,10 @@ public class CosmosContainer {
 
     private <T> Iterator<FeedResponse<T>> getFeedIterator(Flux<FeedResponse<T>> itemFlux) {
         return itemFlux.toIterable(1).iterator();
+    }
+
+    private <T> IterableStream<FeedResponse<T>> getFeedIterableStream(CosmosContinuablePagedFlux<T> cosmosContinuablePagedFlux) {
+        return IterableStream.of(cosmosContinuablePagedFlux.byPage().toIterable(1));
     }
 
 }
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosItemTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosItemTest.java
@@ -6,6 +6,7 @@
 
 package com.azure.cosmos;
 
+import com.azure.core.util.IterableStream;
 import com.azure.cosmos.rx.TestSuiteBase;
 import com.azure.cosmos.implementation.HttpConstants;
 import org.testng.annotations.AfterClass;
@@ -141,10 +142,10 @@ public class CosmosItemTest extends TestSuiteBase {
         String query = String.format("SELECT * from c where c.id = '%s'", properties.getId());
         FeedOptions feedOptions = new FeedOptions();
 
-        Iterator<FeedResponse<CosmosItemProperties>> feedResponseIterator1 =
+        IterableStream<FeedResponse<CosmosItemProperties>> feedResponseIterator1 =
                 container.queryItems(query, feedOptions, CosmosItemProperties.class);
         // Very basic validation
-        assertThat(feedResponseIterator1.hasNext()).isTrue();
+        assertThat(feedResponseIterator1.iterator().hasNext()).isTrue();
 
         SqlQuerySpec querySpec = new SqlQuerySpec(query);
         Iterator<FeedResponse<CosmosItemProperties>> feedResponseIterator3 =


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -329,22 +329,33 @@ public String getId() {
...
      */
-    public <T> Flux<FeedResponse<T>> queryItems(SqlQuerySpec querySpec, FeedOptions options, Class<T> klass) {
-        return getDatabase().getDocClientWrapper().queryDocuments(getLink(),
-                                                                  querySpec, options)
-                   .map(response -> BridgeInternal.createFeedResponseWithQueryMetrics(
-                       (CosmosItemProperties
-                            .getTypedResultsFromV2Results((List<Document>) (Object) response.getResults(),
-                                                          klass)), response.getResponseHeaders(),
-                       response.queryMetrics()));
+    public <T> CosmosContinuablePagedFlux<T> queryItems(SqlQuerySpec querySpec, FeedOptions options, Class<T> klass) {
```
moderakh: aren't we going to have two options class types? now it is not clear if continuationToken is set in FeedOption and also passed by 'CosmosContinuablePageFlux.byPage(continuationToken)' who overrides who.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContinuablePagedFlux.java
```diff
@@ -0,0 +1,85 @@
...
+ * @param <T> The type of elements in a {@link com.azure.core.util.paging.ContinuablePage}
+ * @see com.azure.core.util.paging.ContinuablePage
+ * @see CosmosPagedFluxOptions
+ * @see FeedResponse
+ */
+public class CosmosContinuablePagedFlux<T> extends ContinuablePagedFlux<String, T, FeedResponse<T>> {
+
+    private final Function<CosmosPagedFluxOptions, Flux<FeedResponse<T>>> optionsFluxFunction;
+
+    public CosmosContinuablePagedFlux(Function<CosmosPagedFluxOptions, Flux<FeedResponse<T>>> optionsFluxFunction) {
```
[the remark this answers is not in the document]
kushagraThapar: Done.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosPagedFluxOptions.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Specifies paging options for Cosmos Paged Flux implementation.
+ * @see CosmosContinuablePagedFlux
+ */
+public class CosmosPagedFluxOptions {
+
+    private String requestContinuation;
+    private Integer maxItemCount;
+
+    public CosmosPagedFluxOptions(FeedOptions feedOptions) {
```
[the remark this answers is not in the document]
kushagraThapar: Its not even used for now :) 
I just added it as it made sense to add something with FeedOptions as parameter for constructor. I will change the public type to package-private.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
kushagraThapar: I agree with the confusion as "who overrides who".
We don't need to necessarily split them. Since `continuationToken` is used internally from `feedOptions` class, which we don't need to change. 
I think a better solution here would be to hide public setter of `continuationToken` and `maxItemCount`. So that end users would have to go via `byPage(continuationToken, pageSize)` route to set them up, and internally we can copy them to feedOptions, and keep things as it is. We will not have to change the business logic for these.

kushagraThapar: > some questions:
> 
> 1. Aren't we going to split the options as discussed earlier?
> 2. can you please run a perf benchmark with and without this PR using this benchmark, and see how different the numbers are?
>    #6771

Regarding options, let's discuss that offline. 
Regarding benchmark, I will run them today and will post the numbers here.

kushagraThapar:
> some questions:
> 
> 1. Aren't we going to split the options as discussed earlier?
> 2. can you please run a perf benchmark with and without this PR using this benchmark, and see how different the numbers are?
>    #6771

As discussed over call, we will split the `FeedOptions` last when all the public surface API changes are finished. 
Regarding perf, I am seeing same numbers with / without Cosmos Paged Flux implementation. 

On feature/cosmos/v4 - WITHOUT Cosmos Paged Flux Implementation:
```
#Successful Operations
             count = 10000
         mean rate = 31.13 events/second
     1-minute rate = 31.11 events/second
     5-minute rate = 29.16 events/second
    15-minute rate = 27.12 events/second
#Unsuccessful Operations
             count = 0

Latency
             count = 10000
         mean rate = 31.13 calls/second
     1-minute rate = 31.11 calls/second
     5-minute rate = 29.16 calls/second
    15-minute rate = 27.12 calls/second
               min = 121.16 milliseconds
               max = 573.10 milliseconds
              mean = 323.81 milliseconds
            stddev = 76.05 milliseconds
            median = 336.66 milliseconds
              75% <= 375.86 milliseconds
              95% <= 429.23 milliseconds
              98% <= 454.12 milliseconds
              99% <= 469.30 milliseconds
            99.9% <= 522.05 milliseconds
```

Cosmos Paged Flux Implementation: 
```
#Successful Operations
             count = 10000
         mean rate = 30.45 events/second
     1-minute rate = 30.43 events/second
     5-minute rate = 28.40 events/second
    15-minute rate = 26.24 events/second
#Unsuccessful Operations
             count = 0

Latency
             count = 10000
         mean rate = 30.45 calls/second
     1-minute rate = 30.43 calls/second
     5-minute rate = 28.40 calls/second
    15-minute rate = 26.24 calls/second
               min = 118.10 milliseconds
               max = 571.99 milliseconds
              mean = 326.78 milliseconds
            stddev = 67.36 milliseconds
            median = 331.55 milliseconds
              75% <= 372.05 milliseconds
              95% <= 425.15 milliseconds
              98% <= 454.38 milliseconds
              99% <= 486.39 milliseconds
            99.9% <= 571.99 milliseconds

```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContinuablePagedFlux.java
```diff
@@ -0,0 +1,85 @@
...
+     * Subscribe to consume all items of type {@code T} in the sequence respectively.
+     * This is recommended for most common scenarios. This will seamlessly fetch next
+     * page when required and provide with a {@link Flux} of items.
+     *
+     * @param coreSubscriber The subscriber for this {@link CosmosContinuablePagedFlux}
+     */
+    @Override
+    public void subscribe(CoreSubscriber<? super T> coreSubscriber) {
+        Flux<FeedResponse<T>> pagedResponse = this.byPage();
+        pagedResponse.flatMap(tFeedResponse -> {
```
moderakh: each of the reactor transformation will have its own internal buffering, we need to make sure after the transformation, we fetch limited number of items/pages, and possibly configure how many pages should be pre-fetched.

We should have some tests similar to `BackPressureCrossPartitionTest` and `BackpressureTest` on the surface query API.

Not necessarily in this PR. but we need to have that.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosPagedFluxOptions.java
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.cosmos;
+
+/**
+ * Specifies paging options for Cosmos Paged Flux implementation.
+ * @see CosmosContinuablePagedFlux
+ */
+public class CosmosPagedFluxOptions {
```
moderakh: I only see this is used in the implementation of CosmosContinuablePagedFlux but not exposed in the public surface area

as it is in `com.azure.cosmos` is it used in the public surface area?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContinuablePagedFlux.java
kushagraThapar: Sure, will add those. Thanks for the suggestion.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosPagedFluxOptions.java
kushagraThapar: Good point, we should have it in implementation folder, I will move it.

kushagraThapar: Done.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContinuablePagedFlux.java
kushagraThapar: @moderakh Added back pressure tests for subscribe call on `CosmosContinuablePagedFlux` object.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosPagedFluxOptions.java
moderakh: This is still in the public package.

kushagraThapar: I see it in implementation package. 

https://github.com/Azure/azure-sdk-for-java/pull/7839/files#diff-2028f0a4901c305ffa200afc82a2cf05

kushagraThapar:
known failures, merging this in now.

PR #7840, opened 2020-01-30
RestProxy: Removing code that reflect on OperationStatus type in unreleased management-core

anuchandy:


sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -375,38 +374,13 @@ private static Object convertToResultType(Object wireResponse, Type resultType,
      */
     private static Type extractEntityTypeFromReturnType(HttpResponseDecodeData decodeData) {
         Type token = decodeData.getReturnType();
-        final ClientLogger logger = new ClientLogger(HttpResponseBodyDecoder.class);
         if (token != null) {
             if (TypeUtil.isTypeOrSubTypeOf(token, Mono.class)) {
                 token = TypeUtil.getTypeArgument(token);
-            } else if (TypeUtil.isTypeOrSubTypeOf(token, Flux.class)) {
```
alzimmermsft: Do we need to handle Flux still?

anuchandy: Not in this path. We handle "Flux<ByteBuffer>" in different code path.

alzimmermsft: Ah, perfect!

PR #7857, opened 2020-01-30
Nio file store2

rickle-msft:


sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureStorageFileAttributeView.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.nio.file.attribute.FileAttributeView;
+
+/**
+ * Provides support for properties specific to Azure Blob Storage such as tier.
+ *
+ * {@inheritDoc}
+ */
+public class AzureStorageFileAttributeView implements FileAttributeView {
+
```
rickle-msft: This type was created in this PR so that the AzureFileStore type can accurately say it is supported, but it will be given actual functionality in a later PR when setting blob/file attributes is added.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureStorageFileAttributeView.java
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Returns {@code "azureStorage"}
```
gapra-msft: why is the string for this azureStorage and not AzureStorage

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileStore.java
```diff
@@ -22,6 +25,11 @@
     private final ClientLogger logger = new ClientLogger(AzureFileStore.class);
     private final AzureFileSystem parentFileSystem;
     private final BlobContainerClient containerClient;
+    private static final Map<Class<? extends FileAttributeView>, String> supportedAttributeViews = Map.of(
```
alzimmermsft: `Map.of` was introduced in JDK 9, this will need to be replaced with a static initializer.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileStore.java
```diff
@@ -54,14 +62,22 @@ public String name() {
...
     /**
+     * Returns the {@code String "AzureBlobContainer"} to indicate that the file store is backed by a remote blob
+     * container in Azure Storage.
+     *
      * {@inheritDoc}
      */
     @Override
     public String type() {
-        return null;
+        return "AzureBlobContainer";
```
alzimmermsft: Let's make this value a constant.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileStore.java
```diff
@@ -70,46 +86,54 @@ public boolean isReadOnly() {
...
 
     /**
+     * Containers do not limit the amount of data stored. This method will always return max long.
+     *
      * {@inheritDoc}
      */
     @Override
     public long getTotalSpace() throws IOException {
-        return 0;
+        return Long.MAX_VALUE;
```
alzimmermsft: Would `-1` or a public constant be a better value for this to have clear indication that this is a unique value and shouldn't be used to determine code paths or in validation.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileStore.java
```diff
@@ -118,10 +142,13 @@ public boolean supportsFileAttributeView(String s) {
...
     /**
+     * This method always throws an {@code UnsupportedOperationException} as no {@link FileStoreAttributeView} is
+     * currently supported.
+     *
      * {@inheritDoc}
      */
     @Override
     public Object getAttribute(String s) throws IOException {
-        return null;
+        throw new UnsupportedOperationException("No FileStoreAttributeViews are currently supported.");
```
alzimmermsft:
```suggestion
        throw new UnsupportedOperationException("FileStoreAttributeViews aren't supported.");
```

Indicating currently isn't needed, let's be more blunt on the exception message.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureStorageFileAttributeView.java
gapra-msft: just asking cause we call containers AzureStorageBlobContainer

rickle-msft: It seems the names of other attribute views begins with a lower case letter

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileStore.java
rickle-msft: I imagine there might be scenarios where customers might try to compare the amount of space in a store with the (possibly predicted) amount of data they have. In that case, a `getTotalSpace() > dataSize` check would always fail. Just speculation, though. Thoughts?

PR #7897, opened 2020-02-01
UX study feedback update part 1

samvaity:
This PR covers the frollowing workitems from the UX study feedback
- [ ] Rename `erroneousDocumentCount` to `invalidDocumentCount`
- [ ] Rename `NamedEntity` -> `CategorizedEntity`
- [ ] Rename `type` and `subtype` properties on `NamedEntity` to `category` and `subCategory`.
- [ ] Create new model `PiiEntitiesResult`  and return it for `getPiiEntities` on `RecognizePiiEntitiesResult`.
- [ ] Delete `s` from all `detectLanguage`  method overloads.

### DIFF
--- a/sdk/textanalytics/azure-ai-textanalytics/README.md
+++ b/sdk/textanalytics/azure-ai-textanalytics/README.md
@@ -232,7 +232,7 @@ for (DetectedLanguage detectedLanguage : textAnalyticsClient.detectLanguage(inpu
 ```java
 String text = "Satya Nadella is the CEO of Microsoft";
 
-for (CategorizedEntity entity : textAnalyticsClient.recognizeEntities(text).getCategorizedEntities()) {
+for (CategorizedEntity entity : textAnalyticsClient.recognizeEntities(text).getEntities()) {
     System.out.printf(
         "Recognized Categorized Entity: %s, Category: %s, SubCategory: %s, Score: %s.%n",
         entity.getText(),
@@ -242,12 +242,12 @@ for (CategorizedEntity entity : textAnalyticsClient.recognizeEntities(text).getC
 }
 ```
 
-### Recognize PII(Personally Identifiable Information) entity
+### Recognize PII (Personally Identifiable Information) entity
 <!-- embedme ./src/samples/java/com/azure/ai/textanalytics/ReadmeSamples.java#L105-L114 -->
 ```java
 String text = "My SSN is 555-55-5555";
 
-for (PiiEntity entity : textAnalyticsClient.recognizePiiEntities(text).getPiiEntities()) {
+for (PiiEntity entity : textAnalyticsClient.recognizePiiEntities(text).getEntities()) {
     System.out.printf(
         "Recognized PII Entity: %s, Category: %s, SubCategory: %s, Score: %s.%n",
         entity.getText(),
@@ -302,7 +302,7 @@ gracefully by catching the exception and display the additional information abou
 ```java
 List<DetectLanguageInput> inputs = Arrays.asList(
     new DetectLanguageInput("1", "This is written in English.", "us"),
-    new DetectLanguageInput("2", "Este es un document escrito en Español.", "es")
+    new DetectLanguageInput("1", "Este es un document escrito en Español.", "es")
 );
 
 try {
@@ -316,7 +316,7 @@ try {
 You can set the `AZURE_LOG_LEVEL` environment variable to view logging statements made in the client library. For
 example, setting `AZURE_LOG_LEVEL=2` would show all informational, warning, and error log messages. The log levels can
 be found here: [log levels][LogLevels].
-
+git add 
 ### Default HTTP Client
 All client libraries by default use the Netty HTTP client. Adding the above dependency will automatically configure 
 the client library to use the Netty HTTP client. Configuring or changing the HTTP client is detailed in the
--- a/sdk/textanalytics/azure-ai-textanalytics/src/samples/java/com/azure/ai/textanalytics/ReadmeSamples.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/samples/java/com/azure/ai/textanalytics/ReadmeSamples.java
@@ -158,7 +158,7 @@ public class ReadmeSamples {
     public void handlingException() {
         List<DetectLanguageInput> inputs = Arrays.asList(
             new DetectLanguageInput("1", "This is written in English.", "us"),
-            new DetectLanguageInput("2", "Este es un document escrito en Español.", "es")
+            new DetectLanguageInput("1", "Este es un document escrito en Español.", "es")
         );
 
         try {


sdk/textanalytics/azure-ai-textanalytics/README.md
```diff
@@ -232,12 +232,12 @@ for (DetectedLanguage detectedLanguage : textAnalyticsClient.detectLanguage(inpu
 ```java
 String text = "Satya Nadella is the CEO of Microsoft";
 
-for (NamedEntity entity : textAnalyticsClient.recognizeEntities(text).getNamedEntities()) {
+for (CategorizedEntity entity : textAnalyticsClient.recognizeEntities(text).getNamedEntities()) {
```
kristapratico: Should this be `getCategorizedEntities()`?

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTestBase.java
```diff
@@ -505,29 +523,48 @@ static void validateKeyPhrases(List<String> expectedKeyPhrases, List<String> act
...
-        List<NamedEntity> actualNamedEntityList) {
-        assertEquals(expectedNamedEntityList.size(), actualNamedEntityList.size());
-        expectedNamedEntityList.sort(Comparator.comparing(NamedEntity::getText));
-        actualNamedEntityList.sort(Comparator.comparing(NamedEntity::getText));
-
-        for (int i = 0; i < expectedNamedEntityList.size(); i++) {
-            NamedEntity expectedNamedEntity = expectedNamedEntityList.get(i);
-            NamedEntity actualNamedEntity = actualNamedEntityList.get(i);
-            validateNamedEntity(expectedNamedEntity, actualNamedEntity);
+    static void validatePiiEntities(List<PiiEntity> expectedPiiEntityList,
```
maririos: I know we talked about this, but could you remind me why we decided on creating a new Pii type?
So far, it looks like a lot of duplication code, only for the name. No new property, functionality, or anything

@kristapratico

kristapratico: My context on this comes from the discussion on Teams: 

> ...that reminds me that users in our study were a little confused that RecognizePiiEntities returned a NamedEntity.  It's worth considering adding a new PiiEntity (or other name) type.  When the healthcare endpoint is up, the shape of the return type will be different for healthcare entities. It's also possible that PiiEntity will evolve differently from NamedEntity

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/TextAnalyticsAsyncClient.java
```diff
@@ -275,7 +275,7 @@ public TextAnalyticsServiceVersion getServiceVersion() {
 
     // Named Entity
```
maririos: Categorized

maririos: or Entity Category... not sure what this comment is though

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/TextAnalyticsClient.java
```diff
@@ -166,7 +166,7 @@ public DetectLanguageResult detectLanguage(String text) {
 
     // Named Entity
```
maririos: same here

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTestBase.java
```diff
@@ -286,20 +287,20 @@ void detectLanguageRunner(Consumer<List<DetectLanguageInput>> testRunner) {
     }
 
     // Named Entity runner
```
maririos: Named

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTestBase.java
```diff
@@ -286,20 +287,20 @@ void detectLanguageRunner(Consumer<List<DetectLanguageInput>> testRunner) {
...
-    void recognizeNamedEntityStringInputRunner(Consumer<List<String>> testRunner) {
+    void recognizeCategorizedEntityStringInputRunner(Consumer<List<String>> testRunner) {
         testRunner.accept(NAMED_ENTITY_INPUTS);
     }
 
     void recognizeNamedEntitiesLanguageHintRunner(BiConsumer<List<String>, String> testRunner) {
```
maririos: Named

sdk/textanalytics/azure-ai-textanalytics/src/test/java/com/azure/ai/textanalytics/TextAnalyticsClientTestBase.java
maririos: Thanks!
It makes me wonder if we should change it now, instead of waiting/asking for more information to the service team.

Also, if we end up going this route, implementation wise, .. should we then just create an `Entity` base class and then `CategoryEntity` and `PiiEntity` inherit from it? and whatever other future Entities? That way we can add new content to it, but the core is the same and for now it will save all the duplicated code

kristapratico: I think we are making the change now (at least it seems Java/Python did at this moment). Entity base class makes sense to me.

samvaity: Although, we would have to come up with a better name as `Entity` would be too generic?
How about calling it `NamedEntity` as that is how even the service derives `CategorizedEntity` and `PiiEntity` from.

maririos: `NamedEntity` sounds good to me :)
@annelo-msft what do you think?

annelo-msft: The concern with NamedEntity is that it's confusing to have this type without a Name property on it.  I'd ask what we gain by having a base class here.  Is it in the public API?  I feel like I remember Krzysztof pushing back on my introducing type hierarchies where there isn't a strong need for customers to learn the base type.  We don't know what other entities for other domains will necessarily look like yet, and is it possible that they wouldn't need Category and Subcategory?  If they don't, should they not derive from Entity?  (Do we need Subcategory for PiiEntity, for example?)  Correct me if I'm wrong here, but would adding a base class be a breaking change here?  I think it's not (could be wrong!), and if not, we could add it later when we've seen more examples that would derive from it.

maririos: I wouldn't see the base class as a public API, more an internal to the API to avoid redundancy.

maririos: to close here, yeah we can always do this later.

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizeEntityAsyncClient.java
```diff
@@ -97,7 +98,7 @@
                 documentEntities.getStatistics() == null ? null
                     : toTextDocumentStatistics(documentEntities.getStatistics()),
                 null, documentEntities.getEntities().stream().map(entity ->
-                new NamedEntity(entity.getText(), entity.getType(), entity.getSubtype(), entity.getOffset(),
+                new CategorizedEntity(entity.getText(), entity.getType(), entity.getSubtype(), entity.getOffset(),
```
kristapratico: Should these be `entity.getCategory()` and `entity.getSubCategory()`?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizePiiEntityAsyncClient.java
```diff
@@ -96,7 +96,7 @@
                 documentEntities.getStatistics() == null ? null
                     : toTextDocumentStatistics(documentEntities.getStatistics()),
                 null, documentEntities.getEntities().stream().map(entity ->
-                new NamedEntity(entity.getText(), entity.getType(), entity.getSubtype(), entity.getOffset(),
+                new PiiEntity(entity.getText(), entity.getType(), entity.getSubtype(), entity.getOffset(),
```
kristapratico: same question here

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/RecognizeEntityAsyncClient.java
samvaity: So this the service `entity` (generated code) model class that we are converting to `CategorizedEntity`, hence the `getType()`

kristapratico: Got it, figured it was fine, but just wanted to make sure

PR #7903, opened 2020-02-02
cosmos sdk re-organized conflicting dependencies.

moderakh:
Fixed a few dependencies.

* rxjava2 is a test dependency. Moved from compile dependency to test dependency
* conflicting versions of netty used to be brought in from reactor-netty and explicit dependencies in the sdk, removed the transitive dependencies from reactor-netty on netty and bring in explicit netty dependencies
* azure-core moved to 1.2.0


moderakh:
> Thanks for fixing these issues. Since we are upgrading azure-core, I would like to also update project reactor-core and reactor-netty dependencies, as they have major improvements and bug fixes, which I believe will boost our performance.
> Here are the versions I would like them to be upgraded to:
> project-reactor-core: https://github.com/reactor/reactor-core/releases/tag/v3.3.2.RELEASE
> reactor-netty: https://github.com/reactor/reactor-netty/releases/tag/v0.9.4.RELEASE

done.

PR #7904, opened 2020-02-03
fixed Mono -> Flux -> Mono back and forth conversion

moderakh:
in different layers we were unnecessarily translating Mono to Flux and back to Mono.

When we know the return type is single we should use Mono and not translate back and forth from Mono to Flux and back to Mono again.

### DIFF
--- a/sdk/cosmos/README.md
+++ b/sdk/cosmos/README.md
@@ -224,6 +224,47 @@ and add the following dependency to your project maven dependencies:
 
 For other platforms (Redhat, Windows, Mac, etc) please refer to these instructions https://netty.io/wiki/forked-tomcat-native.html
 
+### Using system properties to modify default Direct TCP options
+
+We have added the ability to modify the default Direct TCP options utilized by the SDK. In priority order we will take default Direct TCP options from:
+
+1. The JSON value of system property `azure.cosmos.directTcp.defaultOptions`.
+   Example: 
+   ```bash
+   java -Dazure.cosmos.directTcp.defaultOptions={\"idleEndpointTimeout\":\"PT24H\"} -jar target/cosmosdb-sdk-testing-1.0-jar-with-dependencies.jar Direct 10 0 Read
+   ```
+
+2. The contents of the JSON file located by system property `azure.cosmos.directTcp.defaultOptionsFile`.
+   Example: 
+   ```
+   java -Dazure.cosmos.directTcp.defaultOptionsFile=/path/to/default/options/file -jar Direct 10 0 Query
+   ```
+
+3. The contents of the JSON resource file named `azure.cosmos.directTcp.defaultOptions.json`.
+   Specifically, the resource file is read from this stream: 
+   ```java
+   RntbdTransportClient.class.getClassLoader().getResourceAsStream("azure.cosmos.directTcp.defaultOptions.json")
+   ```
+   Example: Contents of resource file `azure.cosmos.directTcp.defaultOptions.json`.
+   ```json
+   {
+     "bufferPageSize": 8192,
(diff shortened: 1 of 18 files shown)


sdk/cosmos/azure-cosmos-examples/src/main/java/com/azure/cosmos/rx/examples/multimaster/Helpers.java
```diff
@@ -38,7 +38,7 @@ static public String createDatabaseUri(String databaseName) {
                                 }
                             }
 
-                            return Flux.error(e);
+                            return Mono.error(e);
                         }
                 ).map(ResourceResponse::getResource).single();
```
kushagraThapar: We don't need this `.single()` anymore, right ?
Here and every other method in this class.

sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/CollectionCRUDAsyncAPITest.java
```diff
@@ -242,7 +243,7 @@ public void createCollection_toBlocking_CollectionAlreadyExists_Fails() {
      */
     @Test(groups = "samples", timeOut = TIMEOUT)
     public void transformObservableToCompletableFuture() throws Exception {
-        Flux<ResourceResponse<DocumentCollection>> createCollectionObservable = client
+        Mono<ResourceResponse<DocumentCollection>> createCollectionObservable = client
                 .createCollection(getDatabaseLink(), collectionDefinition, null);
         CompletableFuture<ResourceResponse<DocumentCollection>> future = createCollectionObservable.single().toFuture();
```
kushagraThapar: Same here, unnecessary conversion to `.single()`

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/caches/RxClientCollectionCache.java
```diff
@@ -94,7 +94,7 @@ public RxClientCollectionCache(ISessionContainer sessionContainer,
             retryPolicyInstance.onBeforeSendRequest(request);
         }
 
-        Flux<RxDocumentServiceResponse> responseObs = this.storeModel.processMessage(request);
+        Mono<RxDocumentServiceResponse> responseObs = this.storeModel.processMessage(request);
         return responseObs.map(response -> BridgeInternal.toResourceResponse(response, DocumentCollection.class)
                 .getResource()).single();
```
kushagraThapar: Please remove unnecessary `.single()`

sdk/cosmos/azure-cosmos-examples/src/main/java/com/azure/cosmos/rx/examples/multimaster/Helpers.java
moderakh: as discussed offline. they are intentionally kept as they will do correctness validation to ensure Mono is not empty.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/caches/RxClientCollectionCache.java
moderakh: as discussed offline. they are intentionally kept as they will do correctness validation to ensure Mono is not empty.

moderakh:
> Thanks for bringing this change in.
> Please remove few more `.single()` instances.

as discussed offline. they are intentionally kept as they will do correctness validation to ensure Mono is not empty.

PR #7908, opened 2020-02-03
minor string handling optimization

moderakh:
* keeping activityId as a UUID in DSR to avoid the cost for translating from String to UUID and parsing back to UUID
* using StringUtils.split instead of split to avoid regex cost
* Cache consistency level enum with their string values to avoid parsing costs. as per my comparison (https://github.com/moderakh/azure-cosmosdb-benchmark) this is 300x faster than `EnumUtils.getEnum`

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/rntbd/RntbdRequestHeaders.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/rntbd/RntbdRequestHeaders.java
@@ -1004,7 +1004,8 @@ final class RntbdRequestHeaders extends RntbdTokenStream<RntbdRequestHeader> {
             // not "apps/appName/partitions/partitionKey/replicas/replicaId/dbs/dbName"
 
             final String address = request.getResourceAddress();
-            final String[] fragments = StringUtils.split(address, UrlTrim);
+            // TODO: this is a reqex. UrlTrim=/+, requires some work to remove the regex portion
+            final String[] fragments = address.split(UrlTrim);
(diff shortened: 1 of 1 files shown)


kirankumarkolli:
>         final String value = request.getResourceId();

@David.Noble@microsoft.com /@mo what's the cost of this extraction?
Is it worth short specializing it?

---
Refers to: sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/rntbd/RntbdRequestHeaders.java:994 in 7ea15c6. [](commit_id = 7ea15c66fedbc8d7c7ef5c7cdb037919c8ed979b, deletion_comment = False)

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/HttpTransportClient.java
```diff
@@ -108,7 +108,7 @@ public void close() {
 
             ResourceOperation resourceOperation = new ResourceOperation(request.getOperationType(), request.getResourceType());
             // uuid correlation manager
-            String activityId = request.getActivityId();
+            String activityId = request.getActivityId().toString();
```
kirankumarkolli: Possible that in future there might be more string serilization consumers. Is it worth request having both byte[] or string part of Request?

moderakh: This is in my list to explore.
Libraries like avro have similar optimization:
https://github.com/rdblue/avro-java/blob/master/avro/src/main/java/org/apache/avro/util/Utf8.java

PR #7914, opened 2020-02-03
Removed AZURE_LOG_LEVEL:3 from test.yml in AzConfig which caused azure core to fail.

mssfang:
- 'AZURE_LOG_LEVEL: 3' which caused nightly live tests to failed on HttpLoggingPolicyTests in azure core.
- In addition, removed unnecessary slf4j log dependency.

### DIFF
--- a/sdk/appconfiguration/tests.yml
+++ b/sdk/appconfiguration/tests.yml
@@ -10,4 +10,3 @@ jobs:
         AZURE_CLIENT_ID: $(aad-azure-sdk-test-client-id)
         AZURE_CLIENT_SECRET: $(aad-azure-sdk-test-client-secret)
         AZURE_TENANT_ID: $(aad-azure-sdk-test-tenant-id)
-        AZURE_LOG_LEVEL: 3


mssfang:
Live tests passed: 

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=250472&view=results

sdk/appconfiguration/tests.yml
```diff
@@ -10,4 +10,3 @@ jobs:
         AZURE_CLIENT_ID: $(aad-azure-sdk-test-client-id)
         AZURE_CLIENT_SECRET: $(aad-azure-sdk-test-client-secret)
         AZURE_TENANT_ID: $(aad-azure-sdk-test-tenant-id)
-        AZURE_LOG_LEVEL: 3
```
srnagar: This shouldn't have been removed as it was pointing to a real issue with `HttpLoggingPolicy`.

conniey: +1 to Srikanta's comment. Logging shouldn't affect whether a test passes or fails.

mssfang: created an issue to azure-core: https://github.com/Azure/azure-sdk-for-java/issues/7919

mssfang: KeyVault and Storage don't use AZURE_LOG_LEVEL in test.yml file. So I think azure core changes should not block Azconfig's nightly live tests.

JonathanGiles:
This PR should not have been merged - you did not have any reviewer approving it in its final state, and it is required that all PRs have review approval before merging. This means that Alan's approval is not based on the final code in the PR, and as shown here, there are questions that are raised based on this final change before you merged.

I suggest you revert this merge and submit a new PR for discussion.

sdk/appconfiguration/tests.yml
conniey: Event Hubs uses AZURE_LOG_LEVEL. It helps a lot in understanding what in the SDK went wrong during a test rather than getting no spew at all. 

The original reason we added AZURE_LOG_LEVEL was because we were trying to understand what was going wrong in the live tests because they were encountering a 429. The error could not be reproduced on our dev machines, but in the build machines. So, setting this log level would hopefully shed some light on the errors happening in the SDK.

sima-zhu: This is some limitation on test setup. In HttpLoggingPolicyTests, we have set the environment variable using `System.setProperty`, but this one will be overwritten if there is environment variable setup in system.

The HttpLoggingPolicyTests is working fine on AZURE_LOG_LEVEL<=2.

sima-zhu: The current workaround is to set AZURE_LOG_LEVEL to 2 if log is needed in live tests

PR #7932, opened 2020-02-04
http client pluggability in Identity

### LINKED ISSUE #4548: Add tracing support for Identity
Add tracing context to Identity package
Add T with Response "overload in sync and async Identity package.

g2vinay:
Adds Http Client pluggability support to Identity.
Enables Distributed tracing.

Fixes #4548

### DIFF
--- a/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
+++ b/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
@@ -115,7 +115,7 @@ public class IdentityClient {
                     publicClientApplicationBuilder.proxy(proxyOptionsToJavaNetProxy(options.getProxyOptions()));
                 } else {
                     //Http Client is null, proxy options are not set, use the default client and build the pipeline.
-                    httpPipelineAdapter = new HttpPipelineAdapter(setupPipeline(httpClient));
+                    httpPipelineAdapter = new HttpPipelineAdapter(setupPipeline(HttpClient.createDefault()));
                     publicClientApplicationBuilder.httpClient(httpPipelineAdapter);
                 }
             }
@@ -163,8 +163,7 @@ public class IdentityClient {
         policies.add(new RetryPolicy());
         HttpPolicyProviders.addAfterRetryPolicies(policies);
         policies.add(new HttpLoggingPolicy(httpLogOptions));
-        HttpClient client = httpClient != null ? httpClient : HttpClient.createDefault();
-        return new HttpPipelineBuilder().httpClient(client)
+        return new HttpPipelineBuilder().httpClient(httpClient)
                    .policies(policies.toArray(new HttpPipelinePolicy[0])).build();
     }
 


sdk/identity/azure-identity/src/main/java/com/azure/identity/CredentialBuilderBase.java
```diff
@@ -68,4 +70,18 @@ public T httpPipeline(HttpPipeline httpPipeline) {
         this.identityClientOptions.setHttpPipeline(httpPipeline);
         return (T) this;
     }
+
+    /**
+     * Sets the HTTP client to use for sending and receiving requests to and from the service.
+     *
+     * @param client The HTTP client to use for requests.
+     * @return itself
```
JonathanGiles: Have something clearer than 'itself', e.g. "An updated instance of this builder with the HttpClient set as specified."

sdk/core/azure-core/src/main/java/com/azure/core/credential/TokenRequestContext.java
```diff
@@ -13,12 +17,16 @@
...
+    private HttpClient httpClient;
+    private HttpLogOptions logOptions;
 
     /**
      * Creates a token request instance.
      */
     public TokenRequestContext() {
         this.scopes = new ArrayList<>();
+        this.httpClient = null;
+        this.logOptions = null;
```
srnagar: This is not necessary.

sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/implementation/KeyVaultCredentialPolicy.java
```diff
@@ -37,8 +41,24 @@
...
+        this.httpClient = null;
+        this.logOptions = null;
     }
 
+    /**
+     * Creates KeyVaultCredentialPolicy.
+     *
+     * @param credential the token credential to authenticate the request
+     */
+    public KeyVaultCredentialPolicy(TokenCredential credential, HttpClient httpCLient, HttpLogOptions logOptions) {
```
srnagar:
```suggestion
    public KeyVaultCredentialPolicy(TokenCredential credential, HttpClient httpClient, HttpLogOptions logOptions) {
```

sdk/core/azure-core/src/main/java/com/azure/core/credential/TokenRequestContext.java
```diff
@@ -50,4 +58,46 @@ public TokenRequestContext addScopes(String... scopes) {
...
+     */
+    public HttpClient getHttpClient() {
+        return httpClient;
+    }
+
+    /**
+     * Get the configured log options.
+     * @return the log options.
+     */
+    public HttpLogOptions getLogOptions() {
```
srnagar: There are other policies that we are not including like retry policy, timeout policy etc. Any reason why only HttpLogOptions is included here? Can this be deferred, so we can think of all other policies that might be needed and add it when required?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -113,21 +123,40 @@
...
             if (options.getProxyOptions() != null) {
                 applicationBuilder.proxy(proxyOptionsToJavaNetProxy(options.getProxyOptions()));
+            } else {
+                HttpPipeline pipeline = setupPipeline(request.getHttpClient(), request.getLogOptions());
+                applicationBuilder.httpClient(new HttpPipelineAdapter(pipeline));
             }
+
             if (options.getHttpPipeline() != null) {
                 applicationBuilder.httpClient(new HttpPipelineAdapter(options.getHttpPipeline()));
             }
```
srnagar: If `options.getHttpPipeline()` is not null, the client provided in `TokenRequestContext` is overwritten. Can you please add comments on why this is done? Also, it would be better to check if `options.getHttpPipeline()` is not null first instead of setting up the pipeline which might get overwritten immediately after.

sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/implementation/KeyVaultCredentialPolicy.java
```diff
@@ -37,8 +41,24 @@
     public KeyVaultCredentialPolicy(TokenCredential credential) {
         Objects.requireNonNull(credential);
         this.cache = new ScopeTokenCache((request) -> credential.getToken(request));
+        this.httpClient = null;
+        this.logOptions = null;
```
srnagar: This ctr should call the other overload.

sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/implementation/KeyVaultCredentialPolicy.java
```diff
@@ -37,8 +41,24 @@
...
+    /**
+     * Creates KeyVaultCredentialPolicy.
+     *
+     * @param credential the token credential to authenticate the request
+     */
+    public KeyVaultCredentialPolicy(TokenCredential credential, HttpClient httpCLient, HttpLogOptions logOptions) {
+        Objects.requireNonNull(credential);
+        this.cache = new ScopeTokenCache((request) -> credential.getToken(request));
+        this.httpClient = httpCLient;
+        this.logOptions = logOptions;
```
srnagar: Do we need logOptions?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
```diff
@@ -21,6 +22,7 @@
     private Function<Duration, Duration> retryTimeout;
     private ProxyOptions proxyOptions;
     private HttpPipeline httpPipeline;
+    private HttpClient httpClient;
```
srnagar: I don't see this being used in this class. Since you are allowing users to set the HttpPipline, `httpClient` can be removed.

sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/implementation/KeyVaultCredentialPolicy.java
```diff
@@ -28,17 +29,29 @@
...
     public KeyVaultCredentialPolicy(TokenCredential credential) {
+        this(credential, null);
+    }
+
+    /**
+     * Creates KeyVaultCredentialPolicy.
+     *
+     * @param credential the token credential to authenticate the request
+     */
+    public KeyVaultCredentialPolicy(TokenCredential credential, HttpClient httpClient) {
```
srnagar: Add some unit tests to cover the changes in this PR.

g2vinay: no more changes needed in KV now with option 1 design.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
g2vinay: With option 1 design, this is needed now, as we are allowing users to set the client too.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -111,23 +128,45 @@
...
             if (options.getHttpPipeline() != null) {
                 applicationBuilder.httpClient(new HttpPipelineAdapter(options.getHttpPipeline()));
+            } else {
+                // If http client is set on the credential, then it should override the proxy options if any configured.
+                if (options.getHttpClient() != null) {
+                    HttpPipeline pipeline = setupPipeline(options.getHttpClient());
+                    applicationBuilder.httpClient(new HttpPipelineAdapter(pipeline));
+                } else if (options.getProxyOptions() != null) {
+                    applicationBuilder.proxy(proxyOptionsToJavaNetProxy(options.getProxyOptions()));
+                }
```
schaabs: should we use the same pipeline that we create for the publicApplicationClient in the constructor so we don't have to create a new pipeline for every call to authenticate through the confidential client?

g2vinay: yeah, pipeline construction is now moved to the constructor.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -111,23 +130,38 @@
...
                 ConfidentialClientApplication.builder(clientId, ClientCredentialFactory.createFromSecret(clientSecret))
                     .authority(authorityUrl);
-            if (options.getProxyOptions() != null) {
-                applicationBuilder.proxy(proxyOptionsToJavaNetProxy(options.getProxyOptions()));
-            }
-            if (options.getHttpPipeline() != null) {
+
+            // If http pipeline is available, then it should override the proxy options if any configured.
+            if (httpPipeline != null) {
                 applicationBuilder.httpClient(new HttpPipelineAdapter(options.getHttpPipeline()));
```
schaabs: I believe this should be `applicationBuilder.httpClient(new HttpPipelineAdapter(httpPipeline));` rather than `applicationBuilder.httpClient(new HttpPipelineAdapter(options.getHttpPipeline()));` because `httpPipeline` could have been set directly from the options or constructed in the ctor, so I think it's possible that `options.getHttpPipeline()` is null when `httpPipeline` is not.
 
Also, the client creates new `HttpPipelineAdapter` instances for each msal client it creates. Does each msal client need it's own adapter? If not I would suggest caching the `HttpPipelineAdapter` rather than just the `HttpPipeline`

g2vinay: added the logic to cache the pipeline adapter.
Thanks for the catch on http pipeline, it was an oversight by me.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -82,17 +93,31 @@
...
+                // If http client is set on the credential, then it should override the proxy options if any configured.
+                HttpClient httpClient = options.getHttpClient();
+                if (httpClient != null) {
+                    httpPipelineAdapter = new HttpPipelineAdapter(setupPipeline(httpClient));
+                    publicClientApplicationBuilder.httpClient(httpPipelineAdapter);
+                } else if (options.getProxyOptions() != null) {
+                    publicClientApplicationBuilder.proxy(proxyOptionsToJavaNetProxy(options.getProxyOptions()));
+                } else {
+                    //Http Client is null, proxy options are not set, use the default client and build the pipeline.
+                    httpPipelineAdapter = new HttpPipelineAdapter(setupPipeline(httpClient));
```
srnagar: Instead of passing `httpClient` here, you can as well use `HttpClient.createDefault()` since you know `httpClient` is null.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -111,23 +136,38 @@
...
     }
 
+    private HttpPipeline setupPipeline(HttpClient httpClient) {
+        List<HttpPipelinePolicy> policies = new ArrayList<>();
+        HttpLogOptions httpLogOptions = new HttpLogOptions();
+        HttpPolicyProviders.addBeforeRetryPolicies(policies);
+        policies.add(new RetryPolicy());
+        HttpPolicyProviders.addAfterRetryPolicies(policies);
+        policies.add(new HttpLoggingPolicy(httpLogOptions));
+        HttpClient client = httpClient != null ? httpClient : HttpClient.createDefault();
```
srnagar: This line can be removed if you pass the default client as suggested above.

PR #7936, opened 2020-02-04
update dependency to current

JimSuplizio:
Dependencies on other libraries built within the same pipeline should be current.

### DIFF
--- a/sdk/core/azure-core-amqp/pom.xml
+++ b/sdk/core/azure-core-amqp/pom.xml
@@ -57,7 +57,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;current} -->
     </dependency>
     <dependency>
       <groupId>com.microsoft.azure</groupId>


alzimmermsft:
@srnagar and @conniey I'll defer to you on this as Azure Core AMQP 1.0.1 hasn't been released.

sdk/core/azure-core-amqp/pom.xml
```diff
@@ -57,7 +57,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;current} -->
```
srnagar: We haven't yet released core-amqp and we need to use the GA version of core for the release. Please hold off on merging this until we release.

JimSuplizio: /CC @conniey 
@srnagar I don't understand using the current version is a problem here. While doing the work for the interdependency graph I noticed that this was the only library built within the core pipeline not using the current version of another core library. Every library building within the same pipeline should be using the **current** version of other libraries within the same pipeline.

srnagar: @JimSuplizio this was because core-amqp had never had a GA version release before yesterday and it was out of sync from the rest of the core modules. Since we were planning on releasing amqp yesterday, we didn't want to have an unreleased version of core when we release amqp. Now, that this is released, it can now switch to using `current`.

JimSuplizio: Yes, I'd posted my reply before knowing that these libraries had shipped.

PR #7938, opened 2020-02-04
Interdependency Graph generation

JimSuplizio:
1. Generate the Interdependency graph data as part of the aggregate report.
2. Copy the eng/common/InterdependencyGraph.html and the data.js into the staging directory (next to the dependency report). Once the report exists within there this should allow a badge to be created, similar to the dependencies report for the interdependency graph.


eng/pipelines/aggregate-reports.yml
```diff
@@ -103,10 +103,21 @@ jobs:
         publishJUnitResults: false
         goals: 'clean package exec:java'
 
+    - task: UsePythonVersion@0
```
weshaggard: Why was this necessary? I don't see any usages of python.

eng/DependencyGraph/gen-dep-graph-data.ps1
```diff
@@ -0,0 +1,203 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# This script requires Powershell 6 which defaults LocalMachine to Restricted on Windows client machines.
+# From a Powershell 6 prompt run 'Get-ExecutionPolicy -List' and if the LocalMachine is Restricted or Undefined then
+# run the following command from an admin Powershell 6 prompt 'Set-ExecutionPolicy -ExecutionPolicy RemoteSigned'. This
+# will enable running scripts locally in Powershell 6.
+
+# Use case: Generate the dependency graph data file (data.js) for track 2 (client) libraries.
```
weshaggard: I assume you wrote this independent from Jonathan's dependency report tools because his didn't have the correct hooks to get the graph file we needed. Do you know if the report and this dependency graph are going have a consistent representation?

eng/pipelines/aggregate-reports.yml
JimSuplizio: Gah! I'm not even sure how I could have copy/paste error'd that. I'll remove it with the next PR.

eng/DependencyGraph/gen-dep-graph-data.ps1
JimSuplizio: I wrote this independent of Jonathan's dependency report because this is only getting first level dependencies which are directly in each of the pom files. Even if I were to use the output from the dependency report it would require some massive transformations/processing to get it into the format that the html page needs. Doing it this way was a bit more straightforward and easy to understand.

JimSuplizio:
> Looks good to me. Although I wonder if we should use singular "dependency" instead of "dependencies" but I leave that up to you.

I kept what was there. Being it's cosmetic and easy to change, if we opt to change it it'll take seconds to so. Let's see what the leads want.

PR #7988, opened 2020-02-05
Query Containers

kushagraThapar:
Updated: 
* queryContainers, 
* readAllContainers, 
* queryConflicts, 
* readConflicts API with CosmosContinuablePagedFlux


kushagraThapar:
> LGTM.
> 
> **Just a thought not in the scope of this PR though:**
> FeedOptions has options which are specific to CosmosItem and not related to CosmosContainer. Any thought for having different option types for different operations?

Good point, but do we have a specific example for this in FeedOptions?
And since FeedOptions is still public, and is not going away, I can't think of a big reason to split them into two different options (one for ComosItem and one for CosmosContainer), and then ultimately merge them to one FeedOptions object internally :)

PR #7989, opened 2020-02-05
Scrub Leading BOM in XML Body Response

alzimmermsft:
Check for a leading BOM in the XML response body and scrub it if it exists.


alzimmermsft:
Upon further investigation the root issue is in Azure Core, closing this PR.

PR #7991, opened 2020-02-05
Remove unnecessary modules.

conniey:
Since we use the dependency version of storage, core, and identity, we shouldn't have to add them to our modules.


JimSuplizio:
@conniey, as per our conversation this probably isn't the correct change to make here. We'll meet up to discuss and figure out the right solution to this problem.

PR #8016, opened 2020-02-07
Add ProxyResource class

weidongxu-microsoft:
Mgmt swagger spec has 3 common base resource types.

SubResource got id (read/write).
ProxyResource (sometime alias as Resource) got id+name+type (read-only).
Resource (usually called TrackedResource) got id+name+type+location+tags.


anuchandy:
@yaohaizh shouldn't this be opened against this PR https://github.com/Azure/azure-sdk-for-java/pull/6303?

weidongxu-microsoft:
@yaohaizh Do you want me to stage this PR on your PR, or do it separately here?

yaohaizh:
@anuchandy @weidongxu-microsoft  it should be part https://github.com/Azure/azure-sdk-for-java/pull/6303  but that PR already have lots of changes. I think could better make several small changes to keep things rolling

PR #8025, opened 2020-02-07
Feature/android release hookup

mitchdenny:
This PR includes some reaction changes to enabling the release of Android packages through the Java release pipeline. At the moment its just adding a ```VerifyVersions``` parameter to the invocation of the release template which turns off the version verification/version update logic which is currently specific to the Java mono repo (its turned on here).

But some work related to artifact segregation may also be required. repo refs will be reverted before merging.


mitchdenny:
This one can be closed now ... it was only necessary for some validation of the new pipeline changes.

PR #8036, opened 2020-02-07
README Documentation Updates

alzimmermsft:


sdk/appconfiguration/azure-data-appconfiguration/README.md
```diff
@@ -18,6 +18,35 @@ Use the client library for App Configuration to create and manage application co
...
+
+<dependencies>
+    <dependency>
+        <groupId>com.azure</groupId>
+        <artifactId>azure-data-appconfiguration</artifactId>
+    </dependency>
+</dependencies>
+```
+
+#### Individual Dependency
```
srnagar: Wonder if this section is necessary. We should encourage users to use the BOM and for individual dependency we can just point them to maven central (have a link here) and ask to copy the latest version if they choose to. This will get us out of managing versions in readme and having an incorrect version for short duration of time from when the readme is updated to when the package is released to maven.

alzimmermsft: So, only have the Managed Dependencies section (probably remove the title) and after it have a link that says `If you wish to use individual references find the latest version here[link to maven]`?

alzimmermsft:
Closing in favor of #8112

PR #8045, opened 2020-02-07
Support setting tenant ID on shared token cache credential

jianghaolu:


jianghaolu:
Added an ignore entry in revapi.json. Otherwise reviewed in #6687

PR #8046, opened 2020-02-07
Deep Clone Buffers Before Returning

alzimmermsft:
This PR adds deep cloning of the `ByteBuffer`s that are returned in a download method. This resolves an issue when the return stream is captured for later use and the backing buffers have been re-consumed.

Reference to related issue: https://github.com/Azure/azure-sdk-for-java/issues/7997

Issue #7997, opened 2020-02-06
[BUG] BlobAsyncClient.download() corrupts the file


Dav1dde:
**Describe the bug**

`BlobAsyncClient.download()` returns a corrupted data stream.

**To Reproduce**
Upload a file to azure storage and download it via the Java Async client.

Demo-Spring application to reproduce the problem (see also screeenshot how to use)
[azure-dl.zip](https://github.com/Azure/azure-sdk-for-java/files/4163856/azure-dl.zip)

In order to launch configure `azure.endpoint` either via commandline, `application.properties` or environment variable `AZURE_ENDPOINT=`, needs to be complete blob service URL including SAS token.

Note: to change the azure container name use `azure.container`, it defaults to `test`.

***Code Snippet***

```java
@RestController
public class WebEndpoint {
    private final WebClient webClient;
    private final BlobContainerAsyncClient azureClient;

    public WebEndpoint(WebClient.Builder webClient,
                       BlobServiceAsyncClient azureClient,
                       @Value("${azure.container:test}") String container) {
        this.webClient = webClient.build();
        this.azureClient = azureClient.getBlobContainerAsyncClient(container);
    }

    @GetMapping("/download")
    public ResponseEntity<Flux<ByteBuffer>> download(
        @RequestParam("file") String filename,
        @RequestParam(value = "wc", defaultValue = "false") boolean useWebClient
    ) {
        return ResponseEntity.ok()
            .body(useWebClient ? webClientDownload(filename) : azureClientDownload(filename));
    }

    private Flux<ByteBuffer> webClientDownload(String filename) {
        return this.webClient.get()
            .uri(this.azureClient.getBlobAsyncClient(filename).getBlobUrl())
            .exchange()
            .flatMapMany(c -> c.body(BodyExtractors.toDataBuffers()))
            .map(DataBuffer::asByteBuffer);
    }

    private Flux<ByteBuffer> azureClientDownload(String filename) {
        return this.azureClient.getBlobAsyncClient(filename).download();
    }
}
```

**Expected behavior**
The file is not corrupt

**Screenshots**
Running the code above:

![image](https://user-images.githubusercontent.com/255721/73917461-c3412000-48bf-11ea-99fb-51327e056be8.png)

![image](https://user-images.githubusercontent.com/255721/73918067-19fb2980-48c1-11ea-8c00-2cacc90dab5a.png)

Part of the corrupted file (in the middle):
[...]

PR #8055, opened 2020-02-09
Deep copy buffer

srnagar:


sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
```diff
@@ -190,7 +195,24 @@ public HttpHeaders getHeaders() {
...
-            }).map(ByteBuf::nioBuffer);
+            }).map(byteBuf -> {
+                if (this.disableBufferCopy) {
+                    return byteBuf.nioBuffer();
+                }
+                return deepCopyBuffer(byteBuf);
+            });
+        }
+
+        private ByteBuffer deepCopyBuffer(ByteBuf byteBuf) {
```
alzimmermsft: Could this be static?

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -203,4 +204,9 @@ private ProxyHandler getProxyHandler(AuthorizationChallengeHandler challengeHand
                     String.format(INVALID_PROXY_MESSAGE, proxyOptions.getType())));
         }
     }
+
+    public NettyAsyncHttpClientBuilder disableBufferCopy() {
```
alzimmermsft: Should the name be inverse and take a boolean to allow for this to be flipped back?

alzimmermsft: Missing Javadoc

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
```diff
@@ -190,7 +195,24 @@ public HttpHeaders getHeaders() {
                 if (!reactorNettyConnection.isDisposed()) {
                     reactorNettyConnection.channel().eventLoop().execute(reactorNettyConnection::dispose);
                 }
-            }).map(ByteBuf::nioBuffer);
+            }).map(byteBuf -> {
```
alzimmermsft: Should this be in the bodyIntern method instead?

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
srnagar: yeah, this can be static too.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
srnagar: The name should be `disable` as, by default, it will be enabled. I can add a param to toggle.

srnagar: will do all the javadoc, checkstyle, spotbugs stuff once we are happy with the shape of the code

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
srnagar: As discussed, changing bodyIntern will result in double copy for other `getBody*` methods that don't return `Flux<ByteBuffer>`

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpPipelineCallContext.java
```diff
@@ -85,4 +85,8 @@ public HttpPipelineCallContext setHttpRequest(HttpRequest request) {
         this.httpRequest = request;
         return this;
     }
+
+    public Context getContext() {
+        return this.getContext();
+    }
```
anuchandy: can this be package private? given HttpPipeline type calls HttpClient::send()?

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
```diff
@@ -235,5 +258,17 @@ private ByteBufFlux bodyIntern() {
         Connection internConnection() {
             return reactorNettyConnection;
         }
+
+        private static ByteBuffer deepCopyBuffer(ByteBuf byteBuf) {
+            ByteBuffer buffer = byteBuf.nioBuffer();
+            int offset = buffer.position();
+            int size = buffer.remaining();
```
anuchandy: minor; maybe not now, but we could have a static empty ByteBuffer in class level and if size == 0, we can just return that, instead of creating a new empty `duplicate` reference every time.

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpPipelineNextPolicy.java
```diff
@@ -41,7 +41,8 @@
 
         this.currentPolicyIndex++;
         if (this.currentPolicyIndex == size) {
-            return this.pipeline.getHttpClient().send(this.context.getHttpRequest());
+            return this.pipeline.getHttpClient().send(this.context.getHttpRequest(), this.context.getContext());
+            // return this.pipeline.getHttpClient().send(this.context);
```
anuchandy: nit: commented code.

sdk/core/azure-core/src/main/java/com/azure/core/util/CoreConstants.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Class to hold commonly used constants.
+ */
+public final class CoreConstants {
+
+    /**
+     * This constant is used as key in {@link Context}.
+     */
+    public static final String DISABLE_BUFFER_COPY = "disable-buffer-copy";
```
anuchandy: nit; good to add bit more context on this const.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
```diff
@@ -549,8 +554,11 @@ public BlobProperties downloadToFile(String filePath, boolean overwrite) {
         ParallelTransferOptions parallelTransferOptions, DownloadRetryOptions downloadRetryOptions,
         BlobRequestConditions requestConditions, boolean rangeGetContentMd5, Set<OpenOption> openOptions,
         Duration timeout, Context context) {
+        Context updatedContext = context == null ? Context.NONE.addData(CoreConstants.DISABLE_BUFFER_COPY, true)
+            : context.addData("disable-buffer-copy", true);
```
anuchandy: nit; Context.NONE.addData(CoreConstants.DISABLE_BUFFER_COPY, true)

alzimmermsft: Do we need to use `Context.NONE` in this situation? Instead could a new `Context` be used?

sdk/storage/azure-storage-common/pom.xml
```diff
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;current} -->
```
alzimmermsft: Shouldn't the version of Netty be updated as well?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
srnagar: Context doesn't have a simple helper method to create one with just 1 key-value pair. I had to create a map first and then create context, instead, I used this alternative which was simpler.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
```diff
@@ -190,7 +208,12 @@ public HttpHeaders getHeaders() {
                 if (!reactorNettyConnection.isDisposed()) {
                     reactorNettyConnection.channel().eventLoop().execute(reactorNettyConnection::dispose);
                 }
-            }).map(ByteBuf::nioBuffer);
+            }).map(byteBuf -> {
+                if (this.disableBufferCopy) {
+                    return byteBuf.nioBuffer();
+                }
+                return deepCopyBuffer(byteBuf);
```
JonathanGiles: Reduce this to a one-liner: `return disableBufferCopy ? byteBuf.nioBuffer() : deepCopyBuffer(byteBuf);`

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -180,6 +182,22 @@ public NettyAsyncHttpClientBuilder configuration(Configuration configuration) {
...
     }
 
+    /**
+     * Disables deep copy of response {@link ByteBuffer} into a heap location that is managed by this client as
+     * opposed to the underlying netty library which may use direct buffer pool.
+     * <br>
+     * <b>
+     * Caution: Disabling this is not recommended as it can lead to data corruption if the downstream consumers
+     * of the response do not handle the byte buffers before netty releases them.
+     * </b>
```
JonathanGiles: It would be good to include code here to demonstrate how to use the buffer correctly if the user chooses to disable this, and to also talk about how this will benefit throughput and memory pressure.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
```diff
@@ -385,6 +386,7 @@ public String copyFromUrl(String copySource) {
      */
     public void download(OutputStream stream) {
         downloadWithResponse(stream, null, null, null, false, null, Context.NONE);
+
```
JonathanGiles: nit: Unnecessary newline

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
```diff
@@ -412,9 +414,12 @@ public void download(OutputStream stream) {
     public BlobDownloadResponse downloadWithResponse(OutputStream stream, BlobRange range,
         DownloadRetryOptions options, BlobRequestConditions requestConditions, boolean getRangeContentMd5,
         Duration timeout, Context context) {
+        Context updatedContext = context == null ? Context.NONE.addData(CoreConstants.DISABLE_BUFFER_COPY, true)
+            : context.addData(CoreConstants.DISABLE_BUFFER_COPY, true);
```
JonathanGiles: I'd like to see this kind of repetitive (and verbose) code moved into a utility method.

sdk/core/azure-core/src/main/java/com/azure/core/util/CoreConstants.java
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.util;
+
+/**
+ * Class to hold commonly used constants.
+ */
+public final class CoreConstants {
```
JonathanGiles: I feel a little uneasy with making this public API right now. I would rather it be impl to start with?

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
srnagar: I will add more docs for this.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
```diff
@@ -550,7 +551,8 @@ public BlobProperties downloadToFile(String filePath, boolean overwrite) {
         BlobRequestConditions requestConditions, boolean rangeGetContentMd5, Set<OpenOption> openOptions,
         Duration timeout, Context context) {
         Mono<Response<BlobProperties>> download = client.downloadToFileWithResponse(filePath, range,
-            parallelTransferOptions, downloadRetryOptions, requestConditions, rangeGetContentMd5, openOptions, context);
+            parallelTransferOptions, downloadRetryOptions, requestConditions, rangeGetContentMd5, openOptions,
+            context);
```
JonathanGiles: This class looks like it can now be reverted

sdk/core/azure-core-http-netty/src/test/java/com/azure/core/http/netty/ReactorNettyClientTests.java
```diff
@@ -281,6 +291,102 @@ public void testConcurrentRequests() {
...
+        response.getBody().doOnNext(delayWriteStream::write).blockLast();
+        assertEquals(LONG_BODY, delayWriteStream.aggregateAsString());
+    }
+
+    /**
+     * Tests that preventing deep copying the buffers returned by Netty won't make the stream returned to the customer
+     * resilient to Netty reclaiming them once the 'onNext' operator chain has completed.
+     */
+    @Test
+    public void ignoreDeepCopyBufferConfiguredInBuilder() {
```
hemanttanwar: Since these two methods `deepCopyBufferConfiguredInBuilder` and `ignoreDeepCopyBufferConfiguredInBuilder`  only differs in boolean values for `disableBufferCopy`. Could we use dataprovider to pass the Boolean and just have one method which test both use cases ?

sdk/core/azure-core-http-netty/src/test/java/com/azure/core/http/netty/ReactorNettyClientTests.java
```diff
@@ -281,6 +291,102 @@ public void testConcurrentRequests() {
...
+        response.getBody().doOnNext(delayWriteStream::write).blockLast();
+        assertEquals(LONG_BODY, delayWriteStream.aggregateAsString());
+    }
+
+    /**
+     * Tests that deep copying of buffers is able to be suppressed via {@link Context}.
+     */
+    @Test
+    public void ignoreDeepCopyBufferConfiguredByContext() {
+        HttpClient client = new ReactorNettyClientProvider().createInstance()
```
hemanttanwar: Same comment , dataprovider for Boolean value ?

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
```diff
@@ -180,6 +182,24 @@ public NettyAsyncHttpClientBuilder configuration(Configuration configuration) {
...
+     *
+     * @param disableBufferCopy If set to {@code true}, the client built from this builder will not deep-copy
+     * response {@link ByteBuffer ByteBuffers}.
+     * @return The updated {@link NettyAsyncHttpClientBuilder} object.
+     */
+    public NettyAsyncHttpClientBuilder disableBufferCopy(boolean disableBufferCopy) {
```
hemanttanwar: Also This would be a good candidate to add sample in `samples\README.md` showing how user can use this.
Also changelog entry for this change.

joshfree:
Related https://github.com/Azure/azure-sdk-for-java/pull/8046 and https://github.com/Azure/azure-sdk-for-java/issues/7997

sdk/core/azure-core/src/main/java/com/azure/core/util/CoreUtils.java
```diff
@@ -183,4 +184,17 @@ public static boolean isNullOrEmpty(CharSequence charSequence) {
         }
         return Collections.emptyMap();
     }
+
+    /**
+     * Convenience method to update the input context to disable copying of HTTP response byte buffers. If the input
+     * context is null, a new instance of context is created with a single entry to disable copying of HTTP response
+     * byte buffers.
```
rickle-msft: Can you add some sort of comment here that mentions this method should be called when building any sync clients because the underlying issue is not a problem in those cases? I think it'll help document why we default to using this in a bunch of our builders.

JonathanGiles: Our plan has just evolved to also do buffer copy on sync clients too. Srikanta is just beginning the process of removing these changes so that sync and async both copy the buffer. This is to remove any possibility of data loss in sync clients when there are intermediate operations during which time the buffer may still be lost.

sdk/core/azure-core/src/main/java/com/azure/core/http/HttpPipelineBuilder.java
```diff
@@ -40,7 +40,6 @@
     private HttpClient httpClient;
     private List<HttpPipelinePolicy> pipelinePolicies;
 
-
```
alzimmermsft: nit: Random newline change

sdk/core/azure-core/src/main/java/com/azure/core/implementation/http/HttpClientProviders.java
```diff
@@ -32,7 +32,6 @@ public static HttpClient createInstance() {
         if (defaultProvider == null) {
             throw new IllegalStateException(CANNOT_FIND_HTTP_CLIENT);
         }
-
```
alzimmermsft: nit: Random newline change

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -36,12 +36,12 @@
...
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.3.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;current} -->
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-http-netty</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core-http-netty;current} -->
```
JimSuplizio: This change and the version of core above are not the correct changes. Dependencies built outside of an area should only ever use the dependency version of a library. If the current version is needed then there should be an unreleased tag and that should exist there until that version is released. This would apply to all of the changes below where dependency was changed to current.

srnagar: `{x-version-update;unreleased_com:azure-core-http-netty;current}` the right tag to use for storage then?

srnagar: Got it to work, I was missing the unreleased tag in version_client.txt

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClientBuilder.java
srnagar: Since this behavior is not something we want users to eagerly use, I don't want to put this in README. I have the javadoc updated with codesnippets. So, if they really want to use, they can find it's usage in javadocs.

sdk/core/azure-core-http-netty/src/test/java/com/azure/core/http/netty/ReactorNettyClientTests.java
```diff
@@ -281,6 +289,56 @@ public void testConcurrentRequests() {
...
+    @Test
+    public void deepCopyBufferConfiguredInBuilder() {
+        HttpClient client = new NettyAsyncHttpClientBuilder().disableBufferCopy(false).build();
+
+        HttpResponse response = client.send(new HttpRequest(HttpMethod.GET, url(server, LONG_BODY_PATH))).block();
+        assertNotNull(response);
+        assertEquals(200, response.getStatusCode());
+
+        DelayWriteStream delayWriteStream = new DelayWriteStream();
+        response.getBody().doOnNext(delayWriteStream::write).blockLast();
```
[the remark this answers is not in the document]
conniey:
```java
StepVerifier.create(response.getBody())
.assertNext(body -> {
     // deserialize and assert
})
.expectComplete()
.verify();
```

PR #8056, opened 2020-02-09
improves logic to avoid deserializing PartitionKeyInternal from json header in AddressResolver

moderakh:
current behaviour:
the user passed PartitionKey (which has PartitionKeyInternal) is serialized to json and put into DSR header, and later AddressResolver deserializes the header to get PartitionKeyInternal to use for routing logic.

this PR:
DSR will contain the PartitionKeyInternal, so AddressResolver can rely on DSR.partitionKeyInternal and avoid deserializing partitionKeyInternal from header.

some optimization in escapeNonAscii to avoid creating StringBuilder if the String is all ASCII

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/AddressResolver.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/AddressResolver.java
@@ -223,7 +223,7 @@ public class AddressResolver implements IAddressResolver {
             PartitionKeyRange range;
             PartitionKeyInternal partitionKeyInternal = request.getPartitionKeyInternal();
 
-            if (partitionKeyInternal != null) {
+            if (partitionKeyInternal != null || request.getHeaders().containsKey(HttpConstants.HttpHeaders.PARTITION_KEY)) {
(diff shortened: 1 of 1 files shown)


sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/directconnectivity/AddressResolverTest.java
```diff
@@ -211,7 +213,9 @@ private void TestCacheRefreshWhileRouteByPartitionKey(
 
         request.forceNameCacheRefresh = forceNameCacheRefresh;
         request.forcePartitionKeyRangeRefresh = forceRoutingMapRefresh;
-        request.getHeaders().put(HttpConstants.HttpHeaders.PARTITION_KEY, new PartitionKey("foo").toString());
+        PartitionKey pk = new PartitionKey("foot");
```
kushagraThapar: What "foot" represents ?

moderakh: that's pk value name. changed back to foo.

kirankumarkolli:
@j82w worth exploring on .NET side also. I can help fill some more.

PR #8064, opened 2020-02-10
disable session capturing by default for non-session modes

moderakh:
SessionContainer session capturing is expensive (impact 5%-10% perf from earlier measurement), however it is enabled in non-session mode (including eventual mode).

The only use case for capturing session tokens in non-session mode is 
* capturing session token from one client and moving it to another client
* setting the captured session token from elsewhere for a one specific request (while being in non-session mode)

These use case is extremely rare. we haven't seen such usage. However, in the presence of session capturing everyone will be paying 5-10% perf degradation.

This PR, disables session capturing for non-session mode by default, and introduces a config flag to enable it for non-session mode, just in case if anyone needs the old behaviour.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
@@ -55,6 +55,8 @@ public class CosmosClientBuilder {
 
     /**
      * Indicates if Session capturing is enabled for non Session modes.
+     * The default is false.
+     *
      * @return the session capturing override
      */
     public boolean isSessionCapturingOverride() {


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -36,10 +36,31 @@
...
     public CosmosClientBuilder() {
     }
 
+    /**
+     * Session capturing is enabled by default for {@link ConsistencyLevel#SESSION}.
+     * For other consistency levels, it is not needed, unless if you need occasionally send requests with Session
+     * Consistency while the client is not configured in session.
+     *
+     * enabling Session capturing for Session mode has no effect.
+     * @param sessionCapturingOverride the session capturing override
```
kushagraThapar: We should mention default value of this parameter here. 
It is `false` by default, and that is the intention, right ?

moderakh: sure, will mention here. yes by default it is false

moderakh: correct. thanks for pointing that out. will update javadoc shortly.

moderakh: done.

PR #8083, opened 2020-02-10
added retry analyzer for failing tests

moderakh:
* retrying methods in CosmosKeyCredentialTest as they perform many master resource operation and they fail often due to 429
* only run CosmosKeyCredentialTest in session mode, it is overkill to run in other modes, which increases the chance of 429.
* added support for special handling for 429 to RetryAnalyzer

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/RetryAnalyzer.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/RetryAnalyzer.java
@@ -55,10 +55,10 @@ public class RetryAnalyzer extends RetryAnalyzerCount {
             return null;
         }
 
-        if (t instanceof CosmosClientException) {
-            return (CosmosClientException) t;
+        while(!(t instanceof CosmosClientException)) {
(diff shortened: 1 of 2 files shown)


sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/RetryAnalyzer.java
```diff
@@ -22,11 +24,41 @@ public RetryAnalyzer() {
...
+    private CosmosClientException extractCosmosClientExceptionIfAny(Throwable t) {
+        if (t == null) {
+            return null;
+        }
+
+        if (t instanceof CosmosClientException) {
+            return (CosmosClientException) t;
+        }
+
+        return extractCosmosClientExceptionIfAny(t.getCause());
```
kushagraThapar: Recursion should be avoided as much as we can because of obvious reasons. 
I believe this can easily be converted to a while loop.

moderakh: why?

PR #8084, opened 2020-02-10
Creating new request object on all retry

simplynaveen20:
This PR will revert changes regarding creation of request once for all retry to handle some border line scenarios of PR https://github.com/Azure/azure-sdk-for-java/pull/7156.
Those border line scenarios will be handle in separate PR of refactoring retry context.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
@@ -389,24 +389,24 @@ public class RxDocumentClientImpl implements AsyncDocumentClient, IAuthorization
 
     @Override
     public Mono<ResourceResponse<Database>> createDatabase(Database database, RequestOptions options) {
-        if (database == null) {
-            throw new IllegalArgumentException("Database");
-        }
-
-        logger.debug("Creating a Database. id: [{}]", database.getId());
-        validateResource(database);
-
-        Map<String, String> requestHeaders = this.getRequestHeaders(options);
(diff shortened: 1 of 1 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -1151,27 +1157,28 @@ void captureSessionToken(RxDocumentServiceRequest request, RxDocumentServiceResp
...
-        Document typedDocument = documentFromObject(document, mapper);
-        logger.debug("Replacing a Document. documentLink: [{}]", documentLink);
-        final String path = Utils.joinPath(documentLink, null);
-        final Map<String, String> requestHeaders = getRequestHeaders(options);
+            if (document == null) {
+                throw new IllegalArgumentException("document");
+            }
 
-        String content = toJsonString(typedDocument, mapper);
+            Document typedDocument = documentFromObject(document, mapper);
```
moderakh: We are not using Document here anymore as this will result in double serialization/desrializaiton.
replace with 
```
String content = toJsonString(typedDocument, mapper);	
```

simplynaveen20: Discussed this offline, it was there before the retryContext pr, so keeping the same for now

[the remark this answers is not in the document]
simplynaveen20: Discussed this offline, it was there before the retryContext pr, so keeping the same for now

PR #8049, opened 2020-02-08
Sync Paged Iterable equivalent for ContinuablePagedFlux

anuchandy:

PR #8086, opened 2020-02-10
Implementation of CosmosContinuablePagedIterable for sync APIs

kushagraThapar:
* CosmosContinuablePagedIterable introduction
* Added ContinuablePagedIterable for until this PR goes in: https://github.com/Azure/azure-sdk-for-java/pull/8049

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosItemTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosItemTest.java
@@ -13,6 +13,8 @@ import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Factory;
 import org.testng.annotations.Test;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.UUID;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContinuablePagedIterable.java
```diff
@@ -0,0 +1,37 @@
...
+ * @see FeedResponse
+ */
+public class CosmosContinuablePagedIterable<T> extends ContinuablePagedIterable<String, T, FeedResponse<T>> {
+
+    /**
+     * Creates instance given {@link ContinuablePagedFlux}.
+     *
+     * @param pagedFlux the paged flux use as iterable
+     */
+    public CosmosContinuablePagedIterable(ContinuablePagedFlux<String, T, FeedResponse<T>> pagedFlux) {
```
moderakh: constructor should not be public.

The end user doesn't need to instantiate this. If you made it public because it has to be instantiated from other packages, please use BridgeInternal technique instead.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContinuablePagedIterable.java
```diff
@@ -0,0 +1,37 @@
...
+        super(pagedFlux);
+    }
+
+    /**
+     * Creates instance given {@link ContinuablePagedFlux}.
+     *
+     * @param pagedFlux the paged flux use as iterable
+     */
+    public CosmosContinuablePagedIterable(ContinuablePagedFlux<String, T, FeedResponse<T>> pagedFlux, int batchSize) {
```
moderakh: ditto

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContinuablePagedIterable.java
kushagraThapar: Done.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosContinuablePagedIterable.java
kushagraThapar: Done.

kushagraThapar:
> LGTM
> minor comment on making CosmosContinuablePagedIterable constructor package private.
> 
> Other than that looks good to me.
> also if we don't have test for resuming from a continuation it is worth to add later.

Added test case for continuation token and page size

Issue #8062, opened 2020-02-10
Update Azure Search to use new azure-core-http-netty version


srnagar:
The latest version of `azure-core-http-netty` ( > 1.2.0) has a critical patch that should be adopted in Azure Search.

sima-zhu:
netty has been updated. Closed the issue

PR #8090, opened 2020-02-10
update http-netty version for search

sima-zhu:
Updated the netty version to the latest version as there is a critical patch.


sima-zhu:
https://github.com/Azure/azure-sdk-for-java/issues/8062

sdk/search/azure-search/pom.xml
```diff
@@ -28,13 +28,14 @@
...
-      <artifactId>azure-core</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <artifactId>azure-core-http-netty</artifactId>
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core-http-netty;current} -->
     </dependency>
     <dependency>
-      <groupId>com.azure</groupId>
-      <artifactId>azure-core-http-netty</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
+      <groupId>com.fasterxml.jackson.core</groupId>
```
alzimmermsft: Why is this adding Jackson as a compile time dependency? Azure Core has this as a transitive dependency which should be pulled in.

sdk/search/azure-search/pom.xml
```diff
@@ -28,13 +28,14 @@
...
+      <artifactId>jackson-databind</artifactId>
+      <version>2.10.1</version> <!-- {x-version-update;com.fasterxml.jackson.core:jackson-databind;external_dependency} -->
+      <scope>compile</scope>
```
srnagar: This shouldn't be added.

sdk/search/azure-search/pom.xml
```diff
@@ -28,13 +28,14 @@
   <dependencies>
     <dependency>
       <groupId>com.azure</groupId>
-      <artifactId>azure-core</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <artifactId>azure-core-http-netty</artifactId>
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core-http-netty;current} -->
```
srnagar: Tags need to be updated to use `{x-version-update;unrleased_com.azure:azure-core-http-netty;dependency}`

sdk/search/azure-search/pom.xml
sima-zhu: The tests failed because of missing the required module. You can check the previous commit.

sdk/search/azure-search/pom.xml
sima-zhu: Shouldn't we use current instead of dependency?

sdk/search/azure-search/pom.xml
sima-zhu: The tests complains about missing module.

sdk/search/azure-search/pom.xml
sima-zhu: Double check the modules dependencies. Remove all unnecessary ones.

sdk/search/azure-search/pom.xml
sima-zhu: Double check the modules dependencies. Remove all unnecessary ones.

sdk/search/azure-search/pom.xml
sima-zhu: Revert the azure-core back as the transitive cannot apply more than multiple layers.

sdk/search/azure-search/pom.xml
sima-zhu: Made the changes.

sdk/search/azure-search/pom.xml
sima-zhu: Removed. 

The jackson core module is transitive from azure-core instead of azure-core-netty. Just added azure-core dependency back to make tests work

sdk/search/azure-search/pom.xml
```diff
@@ -29,12 +29,12 @@
...
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core;dependency} -->
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-http-netty</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core-http-netty;dependency} -->
```
alzimmermsft: This could be changed to dependency which should be 1.3.0 in the latest master.

sima-zhu: Made the change

PR #8094, opened 2020-02-11
avoid serializing Range on each request

moderakh:
current behaviour:
on each request a new Range will be created and upon creating a new Range internally Range will be serialized to the property bag in the constructor.

new behaviour:
on each request as the Range fields are cached, we don't serialize to propertyBag. PropertyBag is used only when absolutely necessary.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/routing/Range.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/routing/Range.java
@@ -19,6 +19,8 @@ public final class Range<T extends Comparable<T>> extends JsonSerializable {
 
     private T minValue;
     private T maxValue;
+    private boolean minInclusive;
+    private boolean maxInclusive;
 
     public Range() {
         super();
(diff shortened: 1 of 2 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/routing/Range.java
```diff
@@ -19,20 +19,32 @@
...
+        this.minInclusive = super.getBoolean(Range.IS_MIN_INCLUSIVE_PROPERTY);
+        this.maxInclusive = super.getBoolean(Range.IS_MAX_INCLUSIVE_PROPERTY);
     }
 
     public Range(T min, T max, boolean isMinInclusive, boolean isMaxInclusive) {
-        this.setMin(min);
-        this.setMax(max);
-        this.setMinInclusive(isMinInclusive);
-        this.setMaxInclusive(isMaxInclusive);
+        this.minValue = min;
```
kushagraThapar: Earlier we were setting these values using setMin(min), which will set the property values in property bag. In this PR, we are not setting the property bag values, but just the local values ?

moderakh: correct. they are only need to be in propertyBag when toJson is invoked() which is what we are doing now

kushagraThapar: sounds good, thanks.

PR #8100, opened 2020-02-11
Deep copy changes for hotfix release

srnagar:
Release hotfix for `azure-core-http-netty` 1.2.0 with changes to deep copy buffers when returning Flux<ByteBuffer> response.

### DIFF
--- a/eng/jacoco-test-coverage/pom.xml
+++ b/eng/jacoco-test-coverage/pom.xml
@@ -44,7 +44,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;current} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;current} -->
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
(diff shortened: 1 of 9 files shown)


eng/jacoco-test-coverage/pom.xml
```diff
@@ -115,42 +115,42 @@
...
       <artifactId>azure-storage-blob-cryptography</artifactId>
-      <version>12.3.0</version> <!-- {x-version-update;com.azure:azure-storage-blob-cryptography;current} -->
+      <version>12.3.1</version> <!-- {x-version-update;com.azure:azure-storage-blob-cryptography;current} -->
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-storage-blob-nio</artifactId>
-      <version>12.0.0-beta.1</version> <!-- {x-version-update;com.azure:azure-storage-blob-nio;current} -->
+      <version>12.0.0-beta.2</version> <!-- {x-version-update;com.azure:azure-storage-blob-nio;current} -->
     </dependency>
```
alzimmermsft: This hasn't been releasd, no need to rev versions.

eng/jacoco-test-coverage/pom.xml
```diff
@@ -44,7 +44,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;current} -->
+      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;current} -->
```
alzimmermsft: Should leave the Azure Core versions as a beta, to verify against `1.2.0` in Maven central.

eng/jacoco-test-coverage/pom.xml
srnagar: I just updated them but in the unified pipeline, we can skip releasing these if they don't have to be and this won't go to master anyway.

PR #8111, opened 2020-02-11
Make changes on sdk build commands in CONTRIBUTING.md

sima-zhu:
* The command in file failed to execute. Update to the working one.
* Making changes to the documents with up-to-date info.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,6 +49,11 @@ Merging Pull Requests (for project contributors with write access)
 - Install [Maven](http://maven.apache.org/download.cgi)
   - add `MAVEN_HOME` to environment variables
 
+>**Note:** If you ran into "long path" issue on `Windows`, enable paths longer than 260 characters by: <br><br>
+1.- Run this as Administrator on a command prompt:<br> 
+`REG ADD HKLM\SYSTEM\CurrentControlSet\Control\FileSystem /v LongPathsEnabled /t REG_DWORD /d 1`<br>*(might need to type `yes` to override key if it already exists)*<br><br>
+2.- Set up `git` by running:<br> `git config --system core.longpaths true`
+
 ### Building and Testing
 
 The easiest way to build is by running the following command from the root folder:


CONTRIBUTING.md
```diff
@@ -49,21 +49,16 @@ Merging Pull Requests (for project contributors with write access)
 - Install [Maven](http://maven.apache.org/download.cgi)
   - add `MAVEN_HOME` to environment variables
```
conniey: The note is still valuable. Unless it is listed elsewhere in this document.

CONTRIBUTING.md
```diff
@@ -49,21 +49,16 @@ Merging Pull Requests (for project contributors with write access)
...
 
-
->**Note:** If you are on `Windows`, enable paths longer than 260 characters by: <br><br>
-1.- Run this as Administrator on a command prompt:<br> 
-`REG ADD HKLM\SYSTEM\CurrentControlSet\Control\FileSystem /v LongPathsEnabled /t REG_DWORD /d 1`<br>*(might need to type `yes` to override key if it already exists)*<br><br>
-2.- Set up `git` by running:<br> `git config --system core.longpaths true`
-
 ### Building and Testing
 
 The easiest way to build is by running the following command from the root folder:
```
conniey: A command should be added that we need to build:
`mvn install -f eng\code-quality-reports\pom.xml` before running this install.

sima-zhu: I added -Dincluding-non-shipping-modules. This will build sdk-build-tool. Is there any special reason to have `mvn install -f eng\code-quality-reports\pom.xml`?

CONTRIBUTING.md
sima-zhu: The long path issue has been solved. I can add it back in case we have the same issue.

CONTRIBUTING.md
JimSuplizio: Can someone outside of Microsoft successfully run the sdk-build-tools install command? It's unclear.

sima-zhu: @JimSuplizio 
We have non shipping module which includes sdk-build-tools build and install. 
I tried in local. It can pass if we simply add `-Dincluding-non-shipping-modules` into the origin command line. So there is no need to run `mvn install -f eng\code-quality-reports\pom.xml` before but simply one command line to build the project. 
`

PR #8117, opened 2020-02-11
removed regex in rntbd, added afterburner to rntbd objectmapper

moderakh:
* Replace regex `String.split()` with non regex `StringUtils.split()` (the rationale is regex is expensive)
* registered jackson after-burner module to rntbd for improving serialization/deserialization (this is based on jackson perf guide and as per my measurement in general improves serialization/deserialization perf by 10%, I noticed some of the rntbd components are using jackson for informational debugging info. So this improves the perf for serializing those debugging constructs rather than the actual data.


moderakh:
azp run java - cosmos - tests

moderakh:
two known tests failures which started failing from earlier PR
 `queryItemsWithContinuationTokenAndPageSize`
 `readDocumentsWithAllPermission`
 which are failing from earlier PR:
https://github.com/Azure/azure-sdk-for-java/pull/8086

other than that CI is green

PR #8118, opened 2020-02-11
live-test, service break change

mssfang:
service no longer supports filters, *a* and *a

remove Javadoc on the examples, but give a link pointer instead, so any future changes on the service side will not require update Javadoc on SDK.


sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelector.java
```diff
@@ -47,15 +47,8 @@ public SettingSelector() {
...
-     * <p>
-     * Examples:
-     * <ul>
-     * <li>If {@code keyFilter = "*"}, settings with any key are returned.</li>
-     * <li>If {@code keyFilter = "abc1234"}, settings with a key equal to "abc1234" are returned.</li>
-     * <li>If {@code keyFilter = "abc*"}, settings with a key starting with "abc" are returned.</li>
-     * <li>If {@code keyFilter = "*abc*"}, settings with a key containing "abc" are returned.</li>
-     * <li>If {@code keyFilter = "abc,def"}, settings with a key equal to "abc" or "def" are returned.</li>
-     * </ul>
+     * <p>See <a href="https://github.com/Azure/AppConfiguration/blob/master/docs/REST/kv.md#filtering">Filtering</a>
```
alzimmermsft: Does AppConfiguration have Microsoft Docs? I'd rather link to that than their GitHub page since those will be more official.

mssfang: their doc hasn't been updated yet. https://docs.microsoft.com/en-us/azure/azure-app-configuration/concept-key-value
The service team will update it when GA.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientTestBase.java
```diff
@@ -286,7 +286,7 @@ void listConfigurationSettingsSelectFieldsRunner(BiFunction<List<ConfigurationSe
         tags.put("tag2", "value2");
 
         final SettingSelector selector = new SettingSelector()
-            .setLabelFilter("*-second*")
+            .setLabelFilter("my-second*")
```
xiangyan99: Maybe we can turn it into a negative test case

sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/models/SettingSelector.java
AlexanderSher: In .net we also use github page.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientTestBase.java
mssfang: if in the future, the service team start to support it again. Then we need to update the negative test case again?

conniey: This would be a service API change. So, I hope the API version would be ramped and we'd expect the old API version to continue throwing the 400.

mssfang: make sense to me. I will add the negative test case

PR #8119, opened 2020-02-12
Update root readme

kaerm:
- Making it more accessible to people consuming this page on azure.github.io to avoid redirecting to github from docs
- Updating the description to point people to the most recent packages

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@
 
 This repository is for active development of the Azure SDK for Java. For consumers of the SDK we recommend visiting our [public developer docs](https://docs.microsoft.com/en-us/java/azure/) or our versioned [developer docs](https://azure.github.io/azure-sdk-for-java).
 
-
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,27 +22,17 @@ Java 8 or later is required to use the January 2020 client libraries, otherwise
 
 Each service might have a number of libraries available from each of the following categories:
 
-- [Client: GA January 2020 Releases](#Client-GA-January-2020-Releases)
+- [Client: New Releases](#Client-New-Releases)
```
srnagar: Update the following text as well in `Important` section since the reference to January 2020 GA libraries is removed.

```
The Azure SDK team is pleased to make available the January 2020 client library General Availability (GA) release.
```

joshfree: The anchor on line 68 will need to be fixed as well.  Please grep the file for "January" There's ~4 spots that will need to be fixed at the same time.

PR #8121, opened 2020-02-12
Nio file system

rickle-msft:


sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -181,11 +195,111 @@ public SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> s
     }
 
     /**
+     * The existence of a directory in the {@code AzureFileSystem} is defined on two levels. Weak existence is defined
```
rickle-msft: Reviewers, please read this Javadoc carefully and ensure we are ok with the content and clarity.

sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/AzureFileSystemProviderSpec.groovy
```diff
@@ -111,4 +114,187 @@ class AzureFileSystemProviderSpec extends APISpec {
...
+    }
+
+    def "FileSystemProvider createDir file already exists"() {
+        setup:
+        def fs = createFS(config)
+        def containerClient =
+            primaryBlobServiceClient.getBlobContainerClient(rootToContainer(fs.getDefaultDirectory().toString()))
+        AppendBlobClient blobClient = containerClient.getBlobClient("foo").getAppendBlobClient()
+
+        when: "A file already exists at the location"
```
rickle-msft: This is a bit of a departure from how we normally do tests. It seemed fitting to me, but if others feel I should break these up into discrete tests, I can do that, too.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystem.java
```diff
@@ -187,18 +209,30 @@ public boolean isReadOnly() {
...
-        return "/";
+        return AzureFileSystem.PATH_SEPARATOR;
     }
 
     /**
+     * The list of root directories corresponds to the list of available file stores and therefore containers specified
+     * upon initialization. A root directory always takes the form {@code "<file-store-name>:"}. This list will not
+     * change during the lifetime of this object. If containers are added to the account after initialization, they
+     * will be ignored. If a container is deleted or otherwise becomes unavailable, its root directory will still be
+     * returned but operations to it will fail.
```
jaschrep-msft: In case we later allow users to specify an account instead of individual containers, let's change `This list will not change during the lifetime of this object. If containers are added to the account after initialization, they will be ignored.` to `Results of this list are scoped based on specifications during FileSystem initialization.`

In this same vein, we can change `If a container is deleted or otherwise becomes unavailable, its root directory will still be returned but operations to it will fail.` to `Containers listed are only guaranteed to have existed at some point in time. Operations on nonexistent containers will fail.`. This gives us some more generic language that allows us to update implementation details later on.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystem.java
```diff
@@ -187,18 +209,30 @@ public boolean isReadOnly() {
...
-        return null;
+        return fileStores.keySet().stream()
+            .map(name -> this.getPath(name + AzurePath.ROOT_DIR_SUFFIX))
+            .collect(Collectors.toList());
     }
 
     /**
+     * This list will not change during the lifetime of this object. If containers are added to the account after
+     * initialization, they will be ignored. If a container is deleted or otherwise becomes unavailable, its
+     * corresponding file store will still be returned but operations to it will fail until it becomes available again.
```
jaschrep-msft: Similar comment as above.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -181,11 +195,111 @@ public SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> s
...
+     * by the presence of a non-zero number of blobs prefixed with the directory's path. This is also known as a virtual
+     * directory and is in order to accommodate containers that were pre-loaded with data by another source but need to
+     * be accessed by this file system. Strong existence is defined as the presence of a zero-length blob whose name is
+     * the directory path with a particular metadata field indicating the blob's status as a directory and is also known
+     * as a concrete directory. Directories created by this file system will strongly exist. Operations targeting
+     * directories themselves as the object (e.g. setting properties) will target marker blobs. Operations using the
+     * directory only as a container (e.g. listing) will operate on the blob-name prefix.
+     *
+     * This method fulfills the contract of: "The check for the existence of the file and the creation of the directory
```
jaschrep-msft: update: `This method fulfills the nio contract`... so we know what you're referencing.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -181,11 +195,111 @@ public SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> s
     }
 
     /**
+     * The existence of a directory in the {@code AzureFileSystem} is defined on two levels. Weak existence is defined
+     * by the presence of a non-zero number of blobs prefixed with the directory's path. This is also known as a virtual
+     * directory and is in order to accommodate containers that were pre-loaded with data by another source but need to
```
gapra-msft: and **exists** in order to accommodate maybe?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
gapra-msft: It might be helpful to highlight some of the key words like strong existence, marker blob, etc in like bold/italics?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -181,11 +195,111 @@ public SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> s
...
+     * be accessed by this file system. Strong existence is defined as the presence of a zero-length blob whose name is
+     * the directory path with a particular metadata field indicating the blob's status as a directory and is also known
+     * as a concrete directory. Directories created by this file system will strongly exist. Operations targeting
+     * directories themselves as the object (e.g. setting properties) will target marker blobs. Operations using the
```
jaschrep-msft: `will target marker blobs and create them if they do not exist.`

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -181,11 +195,111 @@ public SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> s
...
+     * directory only as a container (e.g. listing) will operate on the blob-name prefix.
```
jaschrep-msft: Overloaded use of container. Unsure of a good substitute.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
jaschrep-msft: Additionally, this whole paragraph feels a little off. It reads like you're moving to another topic before I've understood the current one. Most notably, I don't understand how we are obeying the contract from this paragraph.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -181,11 +195,111 @@ public SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> s
...
+     * put an object at the given path. This split is due to limitations in the Storage service API.
+     *
+     * There may be some unintuitive behavior when working with directories in this file system, particularly virtual
+     * directories(usually those not created by this file system). A virtual directory will disappear as soon as all its
+     * children have been deleted. Furthermore, if a directory with the given path weakly exists at the time of calling
+     * this method, this method will still return success and create a concrete directory at the target location.
+     * In other words, it is possible to "double create" a directory if it first weakly exists and then is strongly
+     * created. This is both because it is impossible to atomically check if a virtual directory exists while creating a
+     * concrete directory and because such behavior will have minimal side effects--no files will be overwritten and the
+     * directory will still be available for writing as intended, though it may not be empty.
```
jaschrep-msft: `may not` -> `will not`, as it definitely won't be empty if the directory virtually existed already.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -181,11 +195,111 @@ public SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> s
...
+     * <ul>
+     *     <li>{@code Content-Type}</li>
+     *     <li>{@code Content-Disposition}</li>
+     *     <li>{@code Content-Language}</li>
+     *     <li>{@code Content-Encoding}</li>
+     *     <li>{@code Content-MD5}</li>
+     *     <li>{@code Cache-Control}</li>
+     * </ul>
+     * Note that these properties also have a particular semantic in that if one is specified, all are updated. In other
+     * words, if any of the above is set, all those that are not set will be cleared.
```
jaschrep-msft: Link to rest docs for more details on those specific properties?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/implementation/util/Utility.java
```diff
@@ -13,4 +20,59 @@
...
+    /*
+    Note that this will remove the properties from the list of attributes as it finds them.
+     */
+    public static BlobHttpHeaders extractHttpHeaders(List<FileAttribute<?>> fileAttributes, ClientLogger logger) {
+        BlobHttpHeaders headers = new BlobHttpHeaders();
+        for (Iterator<FileAttribute<?>> it = fileAttributes.iterator(); it.hasNext();) {
+            FileAttribute<?> attr = it.next();
+            boolean propertyFound = true;
+            switch (attr.name()) {
+                case "Content-Type":
```
gapra-msft: are these constants anywhere?

sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/AzureFileSystemProviderSpec.groovy
```diff
@@ -111,4 +114,187 @@ class AzureFileSystemProviderSpec extends APISpec {
...
+        1     | _
+        2     | _
+    }
+
+    def "FileSystemProvider createDir relativePath"() {
+        setup:
+        def fs = createFS(config)
+        def containerClient =
+            primaryBlobServiceClient.getBlobContainerClient(rootToContainer(fs.getDefaultDirectory().toString()))
+        AppendBlobClient blobClient = containerClient.getBlobClient("foo").getAppendBlobClient()
```
gapra-msft: can we use a random name here instead of static?

sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/AzureFileSystemProviderSpec.groovy
gapra-msft: I think using the descriptions of the whens help make it clearer, I have no strong oppositions about this

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
rickle-msft: This isn't actually true. We won't eagerly create marker blobs if we target a directory and notice there's no marker blob.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
rickle-msft: I used "may" because we don't know if we're creating a new directory or double creating a virtual directory. So there are cases of calling this method when the directory will be empty.

sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/AzureFileSystemProviderSpec.groovy
jaschrep-msft: I'd prefer separation, as it's not clear which when/then fails when one does. Additionally, we'll stop when we encounter a single failure and we won't immediately see which specific one failed, which is useful information when initially looking at a failure.

If there is a way to have similar mechanics as table tests with `@Unroll` added, then I think this would be useful, as setup code can be complex sometimes.

PR #8123, opened 2020-02-12
Adding readMany api that reads many documents at once based on id and…

mbhaskar:
Adding readMany api that reads many documents at once based on id and partitionkey

This avoids queryplan round trip, and does query only required partitions for required elements.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
@@ -1414,7 +1414,7 @@ public class RxDocumentClientImpl implements AsyncDocumentClient, IAuthorization
                                     // create the executable query
                                     return createReadManyQuery(collectionLink,
                                                                new SqlQuerySpec(sqlQuery),
-                                                               new FeedOptions(),
+                                                               options,
                                                                Document.class,
                                                                ResourceType.Document,
                                                                collection,
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ItemOperations.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ItemOperations.java
@@ -26,7 +26,7 @@ public class ItemOperations {
     public static <T> FeedResponse<T> readMany(CosmosContainer container,
                                                List<Pair<String, PartitionKey>> itemKeyList,
                                                Class<T> klass) {
-        return readManyInternal(container.asyncContainer, itemKeyList, new FeedOptions(), klass).block();
+        return readManyAsync(container.asyncContainer, itemKeyList, klass).block();
     }
 
     /**
@@ -42,6 +42,10 @@ public class ItemOperations {
     public static <T> Mono<FeedResponse<T>> readManyAsync(CosmosAsyncContainer container,
                                                           List<Pair<String, PartitionKey>> itemKeyList,
                                                           Class<T> klass) {
+
+
+        FeedOptions options = new FeedOptions();
+        options.setMaxDegreeOfParallelism(-1);
         return readManyInternal(container, itemKeyList, new FeedOptions(), klass);
     }
 


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -1328,6 +1332,155 @@ void captureSessionToken(RxDocumentServiceRequest request, RxDocumentServiceResp
...
 
+    @Override
+    public Mono<FeedResponse<Document>> readMany(
+        List<Pair<String, PartitionKey>> itemKeyList,
+        String collectionLink) {
+        
+        DocumentCollection collection = 
+            this.readCollection(collectionLink, new RequestOptions()).block().getResource();
+
+        return this.readCollection(collectionLink, new RequestOptions())
```
moderakh: does this actually read the collection? or uses collection cache.

Reading collection is expensive and we should be using CollectionCache.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/AsyncDocumentClient.java
```diff
@@ -1322,6 +1325,14 @@ public CosmosKeyCredential getCosmosKeyCredential() {
      */
     Mono<DatabaseAccount> getDatabaseAccount();
 
+    /**
+     * Reads many documents at once
+     * @param itemKeyList document id and partition key pair that needs to be read
+     * @param collectionLink link for the documentcollection/container to be queried
+     * @return
+     */
+    Mono<FeedResponse<Document>> readMany(List<Pair<String, PartitionKey>> itemKeyList,
```
moderakh: we should have a similar sync API here as well.

```java
Mono<FeedResponse<Document>> readManyAsync()
FeedResponse<Document>> readManySync(.)
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -1328,6 +1332,155 @@ void captureSessionToken(RxDocumentServiceRequest request, RxDocumentServiceResp
...
+            .flatMap(documentCollectionResourceResponse -> {
+                         Map<PartitionKeyRange, List<Pair<String, PartitionKey>>> partitionRangeItemKeyMap =
+                             new HashMap<>();
```
moderakh: shouldn't this be concurrentHashMap as it is modified on a flatMap?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/ParallelDocumentQueryExecutionContext.java
```diff
@@ -96,6 +100,36 @@ private ParallelDocumentQueryExecutionContext(
...
     }
 
+    public static <T extends Resource> Flux<IDocumentQueryExecutionComponent<T>> createReadManyQueryAsync(
+        IDocumentQueryClient queryClient,
+        String collectionResourceId, SqlQuerySpec sqlQuery,
+        Map<PartitionKeyRange, String> rangeQueryMap,
+        FeedOptions feedOptions, String collectionRid, String collectionLink, UUID activityId, Class<T> klass,
+        ResourceType resourceTypeEnum) {
+        
+        List<PartitionKeyRange> ranges = new ArrayList<>();
```
moderakh: same about this one too. shouldn't this be thread-safe?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
```diff
@@ -489,4 +491,12 @@ public static void setFeedOptionsMaxItemCount(FeedOptions feedOptions, Integer m
     public static <T> CosmosContinuablePagedFlux<T> createCosmosContinuablePagedFlux(Function<CosmosPagedFluxOptions, Flux<FeedResponse<T>>> pagedFluxOptionsFluxFunction) {
         return new CosmosContinuablePagedFlux<>(pagedFluxOptionsFluxFunction);
     }
+
+    public static Mono<FeedResponse<Document>> readMany(CosmosAsyncContainer container, 
+                                                     List<Pair<String, PartitionKey>> itemKeyList){
```
kirankumarkolli: >String, PartitionKey [](start = 63, length = 20)

Parameter order consistency w.r.t. other API's

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
```diff
@@ -489,4 +491,12 @@ public static void setFeedOptionsMaxItemCount(FeedOptions feedOptions, Integer m
     public static <T> CosmosContinuablePagedFlux<T> createCosmosContinuablePagedFlux(Function<CosmosPagedFluxOptions, Flux<FeedResponse<T>>> pagedFluxOptionsFluxFunction) {
         return new CosmosContinuablePagedFlux<>(pagedFluxOptionsFluxFunction);
     }
+
+    public static Mono<FeedResponse<Document>> readMany(CosmosAsyncContainer container,
```
kirankumarkolli: Is a sync version of this contract also needed?

kirankumarkolli: Is BridgeInernal right place? May be a new targetted contract
BridgeInternal has lot of other contracts which might get attention.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -1328,6 +1332,155 @@ void captureSessionToken(RxDocumentServiceRequest request, RxDocumentServiceResp
...
+        String partitionKeyPath = (new StringBuilder(partitionKeyPathString)).deleteCharAt(0).toString();
+        
+        Map<PartitionKeyRange, String> rangeQueryMap = new HashMap<>();
+        
+        for (PartitionKeyRange range : partitionRangeItemKeyMap.keySet()) {
+            StringBuilder queryStringBuilder = new StringBuilder();
+            queryStringBuilder.append("select * from c where ");
+            List<Pair<String, PartitionKey>> pairsList = partitionRangeItemKeyMap.get(range);
+            for (Pair p : pairsList) {
+                queryStringBuilder.append("(c.id = \"");
```
moderakh: we can create the template of query using StringBuilder, but for the values of id and pk, can we use SqlParameter?

that way you don't have to worry about serializing the query and also it is safe against sql query injection attack?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -1328,6 +1332,155 @@ void captureSessionToken(RxDocumentServiceRequest request, RxDocumentServiceResp
         return queryDocuments(collectionLink, "SELECT * FROM r", options);
     }
 
+    @Override
+    public Mono<FeedResponse<Document>> readMany(
+        List<Pair<String, PartitionKey>> itemKeyList,
+        String collectionLink) {
+        
+        DocumentCollection collection = 
+            this.readCollection(collectionLink, new RequestOptions()).block().getResource();
```
moderakh: is this some leftover code? is this needed?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
moderakh: how about this class names?
1. `ReadManyItems`
2. `Experimental`
3. `SpecialQuery`

any thought? any suggestion on the class name?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -1328,6 +1332,155 @@ void captureSessionToken(RxDocumentServiceRequest request, RxDocumentServiceResp
...
+                                        .map(feedList -> {
+                                            List<Document> finalList = new ArrayList<>();
+                                            HashMap<String, String> headers = new HashMap<>();
+                                            double requestCharge = 0;
+                                            for (FeedResponse<Document> page : feedList) {
+                                                requestCharge += page.getRequestCharge();
+                                                finalList.addAll(page.getResults());
+                                            }
+                                            headers.put(HttpConstants.HttpHeaders.REQUEST_CHARGE, Double.toString(requestCharge));
+                                            FeedResponse<Document> frp = BridgeInternal.createFeedResponse(finalList, headers);
```
kirankumarkolli: Split handling: Do query layer take care?

moderakh: `DocumentProducer` internally/recursively will handle split, if DocumentProducers correctly wired up no additional work for split proofing should be needed.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
mbhaskar: Moved to using collectioncache

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
mbhaskar: Did minor refactoring to ensure that these calls are only made once in the flatmap and no other thread is going to write on to this. Also made the rangeQueryMap unmodifiable

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/ParallelDocumentQueryExecutionContext.java
mbhaskar: It will be modified only inside this flatmap and is called only once.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
mbhaskar: At the moment, I added async version, sync wrapper can be added quickly if required.

mbhaskar: Leaving it in the BridgeInternal until a decision is taken on where it should actually go

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
mbhaskar: Thank you for the suggestion, moved to use  SqlQuerySpec and SqlParameters, looks much cleaner and also helps handle some additional types without heave conversions

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
mbhaskar: Yup. Deleted

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -1328,6 +1334,182 @@ void captureSessionToken(RxDocumentServiceRequest request, RxDocumentServiceResp
...
+            for (Pair p : pairsList) {
+                paramCnt++;
+                queryStringBuilder.append("(c.id = ");
+                queryStringBuilder.append(getCurentParamName(paramCnt));
+                parameters.add(new SqlParameter(getCurentParamName(paramCnt), p.getLeft()));
+                paramCnt++;
+                queryStringBuilder.append(" and ");
+                queryStringBuilder.append(" c.");
+                // partition key def
+                queryStringBuilder.append(partitionKeyPath);
```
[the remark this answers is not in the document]
moderakh: I addressed this one myself.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
mbhaskar: This API is kind of different from the other we have. Still we are sort of consistent

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
mbhaskar: We moved it to ItemOperations

moderakh: we added sync version too.

mbhaskar: Thanks Mo for adding sync API. We have both sync and async API now.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
```diff
@@ -26,7 +26,9 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.micrometer.core.instrument.MeterRegistry;
+import org.apache.commons.lang3.tuple.Pair;
```
kushagraThapar: Nit: No need for this import.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ParallelDocumentQueryTest.java
```diff
@@ -539,4 +547,44 @@ private void queryWithContinuationTokensAndPageSizes(String query, int[] pageSiz
 
         return receivedDocuments;
     }
+
+    //TODO: Fix the test for GW mode
```
kushagraThapar: Do we still need to fix this for gateway ?

moderakh: later this needs to be fixed. for some reason doesn't work now.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
moderakh: will fix. ack.

Issue #8127, opened 2020-02-12
[BUG] EventContext.updateCheckpoint throws PartitionProcessorException when checkpoint store is BlobCheckpointStore


cykl:
**Describe the bug**

Invoking EventContext.updateCheckpoint from an EventProcessorClient event processor throws a PartitionProcessorException / IllegalStateException when underlying checkpoint store is a BlobCheckpointStore.
 
***Exception or Stack Trace***

```
2020-02-11 18:39:54,136 ERROR com.azure.messaging.eventhubs.PartitionPumpManager [] - Error in event processing callback
2020-02-11 18:39:54,136 ERROR reactor.core.publisher.Operators [] - Operator called default onErrorDropped
com.azure.messaging.eventhubs.implementation.PartitionProcessorException: Error in event processing callback
	at com.azure.messaging.eventhubs.PartitionPumpManager.processEvent(PartitionPumpManager.java:184) ~[azure-messaging-eventhubs-5.0.1.jar:?]
	at com.azure.messaging.eventhubs.PartitionPumpManager.lambda$startPartitionPump$1(PartitionPumpManager.java:148) ~[azure-messaging-eventhubs-5.0.1.jar:?]
	at reactor.core.publisher.LambdaSubscriber.onNext(LambdaSubscriber.java:160) ~[reactor-core-3.3.0.RELEASE.jar:3.3.0.RELEASE]
	at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:192) ~[reactor-core-3.3.0.RELEASE.jar:3.3.0.RELEASE]
	at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:192) ~[reactor-core-3.3.0.RELEASE.jar:3.3.0.RELEASE]
	at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:192) ~[reactor-core-3.3.0.RELEASE.jar:3.3.0.RELEASE]
	at reactor.core.publisher.EmitterProcessor.drain(EmitterProcessor.java:426) ~[reactor-core-3.3.0.RELEASE.jar:3.3.0.RELEASE]
	at reactor.core.publisher.EmitterProcessor.onNext(EmitterProcessor.java:268) ~[reactor-core-3.3.0.RELEASE.jar:3.3.0.RELEASE]
[...]

PR #8128, opened 2020-02-12
Ensure PublishEventWithAzureIdentity publishes all events

cykl:
PublishEventWithAzureIdentity demonstrates how to send events using EventDataBatch to an Event Hub. Unfortunately, current sample does not send anything because last EventDataBatch is not sent.


cykl:
A co-worker executed this sample to reproduce https://github.com/Azure/azure-sdk-for-java/issues/8127 and was puzzled to not observe new events :smile:

PR #8132, opened 2020-02-12
Rev Azure Core Versions After Release

alzimmermsft:


eng/versioning/version_client.txt
```diff
@@ -5,10 +5,10 @@ com.azure:azure-sdk-all;1.0.0;1.0.0
 com.azure:azure-sdk-parent;1.6.0;1.6.0
 com.azure:azure-client-sdk-parent;1.7.0;1.7.0
 com.azure:azure-ai-textanalytics;1.0.0-beta.1;1.0.0-beta.2
-com.azure:azure-core;1.2.0;1.3.0-beta.1
+com.azure:azure-core;1.2.0;1.3.0-beta.2
```
JimSuplizio: I'm not sure if these version changes are OK. [In theory, the dependency-version should be set to the most recent released version and the current-version is set to the next version to be released](https://github.com/Azure/azure-sdk-for-java/blob/master/CONTRIBUTING.md#versions-and-versioning). Further, there's still the entry for unreleased_com.azure:azure-core;1.3.0-beta.1. At a minimum, I'm going to insist that the unreleased core version be set to the current version of core, which is how these entries are supposed to be setup.

alzimmermsft: Updated the unreleased version of Azure Core to `1.3.0-beta.2`. The reason I want to keep the dependency version of Azure Core to `1.2.0` is it will allow the repository to be in a ready to ship state in case we need to do an off-cycle release we won't need to figure out what, if any, changes depended on a version of Azure Core that is in beta and can't be used as a GA dependency.

alzimmermsft:
@g2vinay I added an update for the Identity version as well, please take a look.

PR #8137, opened 2020-02-12
Releasing TA. no longer depends on core 1.3.0.beta-1 but 1.2.0

mssfang:
No longer depends on azure core 1.3.0.beta.1 but back to 1.2.0
which compatible with azure-core-http-netty 1.3.0


conniey:
@mssfang : Can you update your commit message next time before doing the squash/merge? "init" isn't helpful.

image

Issue #6945, opened 2019-12-18
[BUG] ListBlobs failing with Azurite


rfang-srpg:
**Describe the bug**
Cannot use Azurite for local development as listBlobs() (in BlobContainerClient) fails.

***Exception or Stack Trace***
local:
```
[WARNING]
com.azure.storage.blob.models.BlobStorageException: Status code 400, (empty body)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance0 (Native Method)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance (NativeConstructorAccessorImpl.java:62)
    at sun.reflect.DelegatingConstructorAccessorImpl.newInstance (DelegatingConstructorAccessorImpl.java:45)
    at java.lang.reflect.Constructor.newInstance (Constructor.java:423)
    at com.azure.core.http.rest.RestProxy.instantiateUnexpectedException (RestProxy.java:357)
    at com.azure.core.http.rest.RestProxy.lambda$ensureExpectedStatus$6 (RestProxy.java:422)
    at reactor.core.publisher.MonoDefer.subscribe (MonoDefer.java:44)
    at reactor.core.publisher.Mono.subscribe (Mono.java:4087)
    at reactor.core.publisher.FluxSwitchIfEmpty$SwitchIfEmptySubscriber.onComplete (FluxSwitchIfEmpty.java:75)
    at reactor.core.publisher.MonoFlatMap$FlatMapMain.onComplete (MonoFlatMap.java:174)
    at reactor.core.publisher.FluxMap$MapSubscriber.onComplete (FluxMap.java:136)
    at reactor.core.publisher.MonoProcessor$NextInner.onComplete (MonoProcessor.java:518)
    at reactor.core.publisher.MonoProcessor.onNext (MonoProcessor.java:308)
    at reactor.core.publisher.MonoProcessor.onComplete (MonoProcessor.java:265)
    at reactor.core.publisher.FluxDoFinally$DoFinallySubscriber.onComplete (FluxDoFinally.java:138)
    at reactor.core.publisher.FluxHandle$HandleSubscriber.onComplete (FluxHandle.java:206)
    at reactor.core.publisher.FluxMap$MapConditionalSubscriber.onComplete (FluxMap.java:262)
    at reactor.core.publisher.FluxDoFinally$DoFinallySubscriber.onComplete (FluxDoFinally.java:138)
[...]

PR #8153, opened 2020-02-12
Fix list blob items request for empty list of 'include' items

janjaali:
Relates to #6945.

This PR aims to fix an issue in the creation of requests for listing blob items from Azure Storage if no 
datasets-include options are provided. 

The `BlobContainerAsyncClient` creates requests to retrieve the list of blob items by applying a provided list of dataset-include options as comma-separated query-parameter to the request. The created request corresponds to [List Blobs (REST API)](https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs). 

Prior to this change the behavior `BlobContainerAsyncClient` was to add a not-filled 'include' query-parameter to requests without any provided include option:

```http
GET /devstoreaccount1/mycontainer?include=&restype=container&comp=list
```

This behavior is unexpected and does not correspond to the REST API. (Besides of that, Azurite rejects such requests making local development against a "dockerized" solution cumbersome).

The introduced change will modify the behavior of the `BlobContainerAsyncClient` in such a way that empty lists of dataset-include options will be interpreted as `null` causing that no 'include' query-parameter will be attached to the created request.

### DIFF
--- a/sdk/storage/azure-storage-blob/CHANGELOG.md
+++ b/sdk/storage/azure-storage-blob/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Release History
 
 ## 12.5.0-beta.1 (Unreleased)
-
+- Fixed a bug that was adding an invalid 'include' query-parameter for list blob item requests if no dataset-include options were specified.
 
 ## 12.4.0 (2020-02-12)
 - Added ability to access BlobProperties from BlobInputStream.


rickle-msft:
@janjaali Thank you for contributing this fix! Can you please add an entry to the changelog that notes this bugfix?

rickle-msft:
Resolves #6945

rickle-msft:
It looks like there are a couple tests that still need their recordings updated, specifically the listing tests in the CPK-N test suite. Would you mind updating those to see if that'll help pass CI?

janjaali:
I added an entry to the change and also updated the records. The tests are succeeding now, but unfortunately the pom version verifier seems to complain about a not released [azure core version](https://dev.azure.com/azure-sdk/public/_build/results?buildId=260896&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=e75da802-0139-5977-bd1e-91175bec9059&l=60) and the "Test Windows From Source - Java 8" job is running now for 40 min :/

Issue #8044, opened 2020-02-07
[SB Track2] Design Discussion between Unified or Separate Client for Queue/Topic/Subscriber


hemanttanwar:
**Query/Question**
There is wide spectrum of options on how many clients this SDK might have. Other languages are opting for either one client for Queue,Topic,Subscriber or more clients.

This thread discusses pro and con of these options for java.

**Track 1 Service Bus API**
 
Reference : https://apiview.dev/Assemblies/Review/845defea57e84a2885d3b26384f9fe7d
Current API have four separate Client 1. `QueueReceiverClient` 2. `QueueSenderClient`  3.`SubscriptionClient` 4. `TopicClient` 

Following are options for Track 2 API

**Track 2 API Option 1**
 
**APIVIEW:** 
https://apiview.dev/Assemblies/Review/5fe1667bd1cd4b4cb744629fe0ccef70

**Git Prototype and Example:**
**Queue Sender** : https://github.com/Azure/azure-sdk-for-java/compare/master...hemanttanwar:feature/servicebus-track2-queuec-client-api-7697?expand=1#diff-207736c4d6b42231118367d56d2e1698R49
**Queue Receiver** :  https://github.com/Azure/azure-sdk-for-java/compare/master...hemanttanwar:feature/servicebus-track2-queuec-client-api-7697?expand=1#diff-9ea8a00279aa393cd81ce495c498bf9cR120
**Queue Receiver Session** : https://github.com/Azure/azure-sdk-for-java/compare/master...hemanttanwar:feature/servicebus-track2-queuec-client-api-7697?expand=1#diff-1c69d62457074295b52c06558685667eR142

•	Three  separate Client for Queue, Topic and Subscriber
•	Each client have its own builder
•	No Separate Client for basic V/s advance features users
**Pro:** 
    1. Clear delineation of API bases on Client and Client are separated out from Builder (i.e QueueSender, QueueReceiver, Topic and Subscriber Client)
    2. Similar Track 1 API 
**Con:**
   1. Duplication of API, more public API footprint. Example log of function on QueueSender and TopicClient are same.

 
**Track 2 API Option 2**
https://apiview.dev/Assemblies/Review/6af850ed7eb04843a6897afbe13cdd28
**Git:**
[...]

PR #8155, opened 2020-02-12
Prototype for Service Bus Track 2 Queue Receiver/Sender

hemanttanwar:
Draft/ Prototype for Service Bus Track 2

• Three separate Client for Queue, Topic and Subscriber
• Each client have its own builder
• No Separate Client for basic V/s advance features users
**Pro:**
1. Clear delineation of API bases on Client and Client are separated out from Builder (i.e QueueSender, QueueReceiver, Topic and Subscriber Client)
2. Similar Track 1 API
**Con:**
Duplication of API, more public API footprint. 

**Features Considered** 
1.  Read Message/batch from Queue
2. Send message/batch to Queue
3. Implement Session id 

All the options for API design are discussed here.

https://github.com/Azure/azure-sdk-for-java/issues/8044

PS: I will work on check style and spotbugs suppressions now.

### DIFF
--- a/sdk/servicebus/ci.data.yml
+++ b/sdk/servicebus/ci.data.yml
@@ -38,7 +38,7 @@ pr:
       - sdk/servicebus/mgmt-
 
 stages:
-  - template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
+  - template: ../../eng/pipelines/templates/stages/archetype-sdk-data.yml
     parameters:
       ServiceDirectory: servicebus
       Artifacts:
--- a/sdk/servicebus/ci.yml
+++ b/sdk/servicebus/ci.yml
@@ -21,7 +21,7 @@ trigger:
     include:
       - sdk/servicebus/
     exclude:
-      - sdk/servicebus/mgmt-
+      - sdk/servicebus/microsoft-azure
       - sdk/servicebus/mgmt-
 
 pr:
@@ -39,7 +39,7 @@ pr:
       - sdk/servicebus/mgmt-
 
 stages:
-  - template: ../../eng/pipelines/templates/stages/archetype-sdk-data.yml
+  - template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
     parameters:
       ServiceDirectory: servicebus
       Artifacts:


sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/MessageBatch.java
```diff
@@ -0,0 +1,254 @@
...
+ * A class for aggregating {@link Message} into a single, size-limited, batch. It is treated as a single message when
+ * sent to the Azure Service Bus service.
+ *
+
+ *     producer.
+ */
+ final class MessageBatch {
+    private final ClientLogger logger = new ClientLogger(MessageBatch.class);
+    private final Object lock = new Object();
+    private final int maxMessageSize;
```
hemanttanwar: This is a feature which is requested by customer where SDK calculate max size allowed for a batch message and warn client if their messages are larger than max allowed size.

conniey: Do we have a concept of message batch in service bus? If this isn't part of a feature we have committed to, we shouldn't add it in.

hemanttanwar: Yes https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.servicebus.queueclient.sendbatch?view=azure-java-stable

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -16,7 +16,41 @@
 
   <!-- Add Javadoc suppression for any test files (ie. ones that live under src/test/java). -->
   <suppress checks="Javadoc" files="src[/\\]test[/\\]java[/\\].*.java"/>
-
+ <!-- <suppress checks="Javadoc" files="EventPosition.java"/>
```
conniey: Why are we adding all these suppressions?

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ClientConstants.java
```diff
@@ -3,10 +3,13 @@
 
 package com.azure.core.amqp.implementation;
 
+import java.time.Duration;
```
conniey: Why is this static field here?

sdk/core/azure-core-amqp/src/main/java/module-info.java
```diff
@@ -3,18 +3,23 @@
 
 module com.azure.core.amqp {
     requires transitive com.azure.core;
-    requires transitive reactor.core;
 
-    requires com.microsoft.azure.qpid.protonj.extensions;
-    requires proton.j;
-    requires org.reactivestreams;
+    requires transitive com.microsoft.azure.qpid.protonj.extensions;
```
conniey: is there a reason we're adding transitive here?

sdk/core/azure-core-amqp/src/main/java/module-info.java
```diff
@@ -3,18 +3,23 @@
...
     exports com.azure.core.amqp.implementation to
+        com.azure.messaging.servicebus.implementation,
+        com.azure.messaging.servicebus,
         com.azure.messaging.eventhubs;
+
     exports com.azure.core.amqp.implementation.handler to
+        com.azure.messaging.servicebus,
         com.azure.messaging.eventhubs;
+
+    opens com.azure.core.amqp.implementation;
```
conniey: why are we adding this opens here?

sdk/core/azure-core-amqp/src/main/java/module-info.java
```diff
@@ -3,18 +3,23 @@
...
+    requires transitive com.microsoft.azure.qpid.protonj.extensions;
+    requires transitive proton.j;
+    requires transitive org.reactivestreams;
 
     exports com.azure.core.amqp;
     exports com.azure.core.amqp.exception;
 
     // FIXME this should not be a long-term solution
     exports com.azure.core.amqp.implementation to
+        com.azure.messaging.servicebus.implementation,
```
conniey: nit: alphabetical order

sdk/servicebus/azure-messaging-servicebus/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+
+  <scm>
+    <url>scm:git:https://github.com/Azure/azure-sdk-for-java</url>
+    <connection>scm:git:git@github.com:Azure/azure-sdk-for-java.git</connection>
+    <tag>HEAD</tag>
+  </scm>
+
+  <properties>
+    <project.automatic.module.name>com.azure.messaging.servicebus</project.automatic.module.name>
+    <checkstyle.skip>true</checkstyle.skip>
```
conniey: these shouldn't be here

sdk/servicebus/azure-messaging-servicebus/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+    <spotbugs.version>4.0.0-beta3</spotbugs.version>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-amqp</artifactId>
+      <version>1.1.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core-amqp;dependency} -->
```
conniey: This should be the dependency version. We haven't updated anything in azure-core-amqp.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
```diff
@@ -0,0 +1,265 @@
...
+ * <li>{@link #getBody()} - if AMQPMessage.Body has Data section</li>
+ * </ol>
+ *
+ * <p>
+ * Serializing a received {@link Message} with AMQP sections other than ApplicationProperties (with primitive Java
+ * types) and Data section is not supported.
+ * </p>
+ *
+ * @see MessageBatch
+
```
conniey: nit: empty line

eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
```diff
@@ -408,10 +408,25 @@
   </Match>
 
   <!-- The HttpRequest fields are serializable. Flux<T> and HttpHeaders are able to serialize and deserialize. -->
+  <Match>
```
conniey: Shouldn't have these.

eng/versioning/version_client.txt
```diff
@@ -30,6 +30,7 @@ com.azure:azure-storage-common;12.3.1;12.4.0-beta.1
 com.azure:azure-storage-file-share;12.1.1;12.2.0-beta.1
```
conniey: alphabetical order

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    public UUID getLockToken() {
+        return lockToken;
+    }
+
+    public void setLockToken(UUID lockToken) {
+        this.lockToken = lockToken;
+    }
+
```
conniey: nit: empty line.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
```diff
@@ -0,0 +1,265 @@
...
+        this.context = context.addData(key, value);
+
+        return this;
+    }
+
+    public UUID getLockToken() {
+        return lockToken;
+    }
+
+    public void setLockToken(UUID lockToken) {
```
conniey: Should we allow a user to set the lock token? this seems dangerous.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/MessageBatch.java
```diff
@@ -0,0 +1,251 @@
...
+        return this.sizeInBytes;
+    }
+
+    /**
+     * Tries to add an {@link Message event} to the batch.
+     *
+     * @param message The {@link Message} to add to the batch.
+     * @return {@code true} if the event could be added to the batch; {@code false} if the event was too large to fit in
+     *     the batch.
+     * @throws IllegalArgumentException if {@code eventData} is {@code null}.
```
conniey: The javadocs don't match up.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/MessageBatch.java
```diff
@@ -0,0 +1,251 @@
...
+                    final MessageAnnotations messageAnnotations = (message.getMessageAnnotations() == null)
+                        ? new MessageAnnotations(new HashMap<>())
+                        : message.getMessageAnnotations();
+                    messageAnnotations.getValue().put(Symbol.getSymbol(key), value);
+                    message.setMessageAnnotations(messageAnnotations);
+                }
+            });
+        }
+
+
```
conniey: nit: extra line

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
+
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+import java.util.Map;
+import java.util.Objects;
+import java.util.ServiceLoader;
+
+@ServiceClientBuilder(serviceClients = {})
```
conniey: Should have a service client here

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+public final class QueueClientBuilder {
+
+    private final ClientLogger logger = new ClientLogger(QueueClientBuilder.class);
```
conniey: move final variable below static fields.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+    private TokenCredential credentials;
+    private Configuration configuration;
+    private AmqpRetryOptions retryOptions;
+    private Scheduler scheduler;
+    private AmqpTransportType transport = AmqpTransportType.AMQP;
+    private String fullyQualifiedNamespace;
+    private String queueName;
+
+
+    private final String connectionId;
```
conniey: group final and other writable variables together.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+    private ServiceBusConnectionProcessor servicerBusConnectionProcessor;
+    private boolean isSharedConnection;
+
+
+    private SendOptions defaultSenderOptions;
+
```
conniey: nit: extra lines

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+    private String fullyQualifiedNamespace;
+    private String queueName;
+
+
+    private final String connectionId;
+    private ServiceBusConnectionProcessor servicerBusConnectionProcessor;
+    private boolean isSharedConnection;
+
+
+    private SendOptions defaultSenderOptions;
```
conniey: What are send options?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+    private AmqpRetryOptions retryOptions;
+    private Scheduler scheduler;
+    private AmqpTransportType transport = AmqpTransportType.AMQP;
+    private String fullyQualifiedNamespace;
+    private String queueName;
+
+
+    private final String connectionId;
+    private ServiceBusConnectionProcessor servicerBusConnectionProcessor;
+    private boolean isSharedConnection;
```
conniey: Does the concept of a shared connection exist in SB? I'd prefer to remove it until there is a feature request. That's why we had this in Event Hubs.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+    }
+    private ConnectionOptions getConnectionOptions() {
+        configuration = configuration == null ? Configuration.getGlobalConfiguration().clone() : configuration;
+
+        if (credentials == null) {
+            final String connectionString = configuration.get(AZURE_SERVICE_BUS_CONNECTION_STRING);
+
+            if (CoreUtils.isNullOrEmpty(connectionString)) {
+                throw logger.logExceptionAsError(new IllegalArgumentException("Credentials have not been set. "
+                    + "They can be set using: connectionString(String), connectionString(String, String), "
```
conniey: This message is incorrect.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueReceiverAsyncClient.java
```diff
@@ -0,0 +1,269 @@
...
+    private final String queueName;
+    private final ServiceBusConnectionProcessor connectionProcessor;
+    private final MessageSerializer messageSerializer;
+    private final int prefetchCount;
+    private final boolean isSharedConnection;
+    private final TracerProvider tracerProvider;
+    private final ReceiveMode defaultReceiveMode = ReceiveMode.PEEKLOCK;
+
+
+    AmqpRetryOptions defaultRetryOptions =  new AmqpRetryOptions();
```
conniey: Visibility on this should be private

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueReceiverAsyncClient.java
```diff
@@ -0,0 +1,269 @@
...
+
+    /**
+     * Keeps track of the open consumers keyed by linkName. The link name is generated as: {@code
+     * "partitionId_GUID"}. For receiving from all partitions, links are prefixed with {@code "all-GUID-partitionId"}.
+     */
+    private final ConcurrentHashMap<String, ServiceBusAsyncConsumer> openConsumers =
```
conniey: The concept of partitions doesn't exist here. This class was introduced so that one consumer could listen to multiple partitions in Event Hubs. What you're copying is the incorrect class.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ReceiveMode.java
```diff
@@ -0,0 +1,20 @@
...
+     * <ul>
+     * <li>If the receiver processes the message successfully, it calls <code>complete</code> and the message will be deleted.</li>
+     * <li>If the receiver decides that it can't process the message successfully, it calls <code>abandon</code> and the message will be unlocked and made available to other receivers.</li>
+     * <li>If the receiver wants to defer the processing of the message to a later point in time, it calls <code>defer</code> and the message will be deferred. A deferred can only be received by its sequence number.</li>
+     * <li>If the receiver wants to dead-letter the message, it calls <code>deadLetter</code> and the message will be moved to a special sub-queue called deadletter queue.</li>
+     * <li>If the receiver calls neither of these methods within a configurable period of time (by default, 60 seconds), the service assumes the receiver has failed. In this case, it behaves as if the receiver had called <code>abandon</code>, making the message available to other receivers</li>
+     * </ul>
+     */
+    PEEKLOCK,
```
conniey: PEEK_LOCK, and RECEIVE_AND_DELETE

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Rule.java
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.messaging.servicebus;
+
+public class Rule {
```
conniey: Are rules something we have in the first preview? if not, we should remove it for now. Else, this is in the public API.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
```diff
@@ -0,0 +1,388 @@
...
+        return message;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T> T deserialize(org.apache.qpid.proton.message.Message message, Class<T> clazz) {
+        Objects.requireNonNull(message, "'message' cannot be null.");
+        Objects.requireNonNull(clazz, "'clazz' cannot be null.");
+
+        if (clazz == ServiceBusProperties.class) {
```
conniey: There isn't ServiceBusProperties in SB API

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
```diff
@@ -0,0 +1,388 @@
...
+        } else if (clazz == Message.class) {
+            return (T) deserializeEventData(message);
+        } else {
+            throw logger.logExceptionAsError(new IllegalArgumentException(
+                "Deserialization only supports ServiceBusProperties."));
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T> T deserializeManagementResponse(org.apache.qpid.proton.message.Message message, Class<T> deserializedType) {
```
conniey: We don't have a management node ($management) in SB afaik. Do we need this?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
```diff
@@ -0,0 +1,388 @@
...
+        if (deserializedType == ServiceBusProperties.class) {
+            return (T) toServiceBusProperties(amqpBody);
+        } else {
+            throw logger.logExceptionAsError(new IllegalArgumentException(String.format(
+                Messages.CLASS_NOT_A_SUPPORTED_TYPE, deserializedType)));
+        }
+    }
+
+
+    private Message deserializeEventData(org.apache.qpid.proton.message.Message message) {
```
conniey: deserializeEventData -> deserializeMessage. Does this match the track 1 deserialization logic?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
```diff
@@ -0,0 +1,388 @@
...
+                    ? new MessageAnnotations(new HashMap<>())
+                    : message.getMessageAnnotations();
+                messageAnnotations.getValue().put(Symbol.getSymbol(key), value);
+                message.setMessageAnnotations(messageAnnotations);
+            }
+        });
+    }
+
+    private static int getPayloadSize(org.apache.qpid.proton.message.Message msg) {
+
```
conniey: Should make sure this aligns with the track 1 serialization code for a service bus message. iirc, it's different.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/TransactionContext.java
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.messaging.servicebus;
+
+import java.nio.ByteBuffer;
+
+public class TransactionContext {
+
+    public ByteBuffer getTransactionId() {
```
conniey: Why is tranactionId a ByteBuffer? Did we have an API review on this?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/TransactionContext.java
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.messaging.servicebus;
+
+import java.nio.ByteBuffer;
+
+public class TransactionContext {
```
conniey: I would hold off on adding types we haven't discussed until they are fleshed out.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/TransactionContext.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public ByteBuffer getTransactionId() {
+        return null;
+    }
+
+    public void commit()  {}
+
+    public void commitAsync() {}
```
conniey: commitAsync is very .NET. `public Mono<Void> commit();`

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/CreateBatchOptions.java
```diff
@@ -0,0 +1,40 @@
...
+import com.azure.core.annotation.Fluent;
+
+/**
+ * The set of options that can be specified when creating an batch of messages. This wrapper will help to limit
+ * the messages with maximum allowed size.
+ *
+
+ */
+@Fluent
+public class CreateBatchOptions {
```
conniey: The "MessageBatch" is not publicly exposed. We haven't discussed having this made public.

conniey: The sendBatch APIs don't expose any of these options.

https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.servicebus.imessagesender.sendbatch?view=azure-java-stable#com_microsoft_azure_servicebus_IMessageSender_sendBatch_Collection___extends_IMessage__

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/Messages.java
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
conniey: We have this here because there are error messages and a corresponding ".properties" resource file to map to. There should be that file made available.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/SendOptions.java
```diff
@@ -0,0 +1,77 @@
...
+
+
+import com.azure.core.annotation.Fluent;
+
+/**
+ * The set of options that can be specified when sending a set of events to influence the way in which events are sent
+ * to the Event Hubs service.
+ */
+@Fluent
+public class SendOptions {
```
conniey: The message class itself has "PartitionKey". Partition id is not a concept in SB. Do we need this class?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusProperties.java
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
conniey: This class should be removed.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/AsyncReceiverTest.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.Map;
+import java.util.UUID;
+import java.util.function.Supplier;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class AsyncReceiverTest {
```
conniey: Should be named QueueReceiverAsyncClientTest

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/AsyncSenderTest.java
```diff
@@ -0,0 +1,162 @@
...
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class AsyncSenderTest {
```
conniey: Should be named QueueSenderAsyncClientTest

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
hemanttanwar: All the suppression is going to be removed before PR is merged as I work on them .

sdk/core/azure-core-amqp/src/main/java/module-info.java
hemanttanwar: emoved

sdk/core/azure-core-amqp/src/main/java/module-info.java
hemanttanwar: `transitive` is  for implicit readability,  so in service I do not need to depend on it and it will be available implecitly.

sdk/servicebus/azure-messaging-servicebus/pom.xml
hemanttanwar: We changed module-info.java

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueReceiverAsyncClient.java
hemanttanwar: We have multiple session , similar concept and  we could use it for that purpose.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
hemanttanwar: I will check for track 1 deserialization .

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/MessageBatch.java
```diff
@@ -0,0 +1,248 @@
...
+        this.eventBytes = new byte[maxMessageSize];
+        this.tracerProvider = tracerProvider;
+    }
+
+    /**
+     * Gets the number of {@link Message events} in the batch.
+     *
+     * @return The number of {@link Message events} in the batch.
+     */
+    public int getCount() {
```
conniey: This class is package-private. These don't need to be public.

sdk/servicebus/pom.service.xml
```diff
@@ -10,5 +10,6 @@
   <version>1.0.0</version><!-- Need not change for every release-->
   <modules>
     <module>microsoft-azure-servicebus</module>
```
conniey: This should be updated to have the profiles.
https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/eventhubs/pom.service.xml#L10-L45

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueSenderAsyncClientTest.java
```diff
@@ -0,0 +1,137 @@
...
+            .buildAsyncSenderClient();
+
+        // Create an event to send.
+        Message message = new Message("Hello world!".getBytes(UTF_8));
+
+        // Send that event. This call returns a Mono<Void>, which we subscribe to. It completes successfully when the
+        // event has been delivered to the Event Hub. It completes with an error if an exception occurred while sending
+        // the event.
+
+        asyncSender.send(message)
```
conniey: This should be using StepVerifier.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueReceiverAsyncClientTest.java
```diff
@@ -0,0 +1,204 @@
...
+        String connectionString = baseConnectionString + ";EntityPath=hemant-test1";
+
+        // Instantiate a client that will be used to call the service.
+
+        consumer = new QueueClientBuilder()
+            .connectionString(connectionString)
+            .createAsyncReceiverClient(numberOfEvents);
+        Thread.sleep(3000);
+
+        consumer.receive()
```
conniey: This should be using StepVerifier.

sdk/servicebus/azure-messaging-servicebus/src/main/java/module-info.java
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module com.azure.messaging.servicebus {
+    requires transitive com.azure.core.amqp;
```
conniey: nit: keep your requires, opens and uses grouped together.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusSharedKeyCredential.java
```diff
@@ -0,0 +1,158 @@
...
+import java.time.Duration;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Base64;
+import java.util.Locale;
+import java.util.Objects;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
```
conniey: Do you have an issue open to update all the documentation?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusAsyncConsumer.java
```diff
@@ -0,0 +1,74 @@
...
+import com.azure.messaging.servicebus.implementation.ServiceBusReceiveLinkProcessor;
+import reactor.core.publisher.EmitterProcessor;
+import reactor.core.publisher.Flux;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * A package-private consumer responsible for reading {@link Message} from a specific Service Bus.
+ */
+class ServiceBusAsyncConsumer implements AutoCloseable {
```
conniey: This should be in the implementation package.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueSenderAsyncClient.java
```diff
@@ -0,0 +1,374 @@
...
+    public String getQueueName() {
+        return this.queueName;
+    }
+
+    /**
+     *
+     * @param batch of messages which allows client to send maximum allowed size for a batch of messages.
+     * @return The {@link Mono} the finishes this operation on service bus resource.
+     */
+    public Mono<Void> send(MessageBatch batch) {
```
conniey: This should be package-private. MessageBatch is not accessible outside of this package.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueReceiverAsyncClient.java
conniey: It wouldn't be structured in this way and be returned from the .receive() function. A QueueReceiver would be tied to a single `ServiceBusAsyncConsumer`. You'd probably have another class SessionReceiver or something internally.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
```diff
@@ -0,0 +1,262 @@
...
+ * @see MessageBatch
+ */
+public class Message {
+    /*
+     * These are properties owned by the service and set when a message is received.
+     */
+    static final Set<String> RESERVED_SYSTEM_PROPERTIES;
+
+    private final Map<String, Object> properties;
+    private final byte[] body;
```
yvgopal: >private final byte[] body; [](start = 4, length = 26)

Accepting only bytes as body will break interoperability with other clients sending messages to Service Bus. If an application uses any standard AMQP client or another Service Bus client to send a non-byte message, that message cannot be represented by this class and bytes will not mean anything to a receiving application. We started the V1 Java SDK with bytes only message and then added support for other message types. If the V2 SDK is supporting only bytes means starting from square one again.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusSharedKeyCredential.java
hemanttanwar: https://github.com/Azure/azure-sdk-for-java/issues/8225

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ClientConstants.java
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License.
+
+package com.azure.messaging.servicebus.implementation;
+
+import java.time.Duration;
+
+/**
+ *  Constants defined for service bus.
+ */
+public class ClientConstants {
```
JonathanGiles: This should have a more descriptive name

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
hemanttanwar: I am adding String , contentType, byte[] constructor for now and will compare with what else is needed. I will add more add needed.

JonathanGiles:
It would be best to request another review when all feedback is taken into consideration, to save from reviewing the large PR too frequently.

sdk/servicebus/azure-messaging-servicebus/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+
+  <scm>
+    <url>scm:git:https://github.com/Azure/azure-sdk-for-java</url>
+    <connection>scm:git:git@github.com:Azure/azure-sdk-for-java.git</connection>
+    <tag>HEAD</tag>
+  </scm>
+
+  <properties>
+    <project.automatic.module.name>com.azure.messaging.servicebus</project.automatic.module.name>
+    <spotbugs.version>4.0.0-beta3</spotbugs.version>
```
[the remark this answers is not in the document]
hemanttanwar: removing it.

sdk/servicebus/azure-messaging-servicebus/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+  </distributionManagement>
+
+  <scm>
+    <url>scm:git:https://github.com/Azure/azure-sdk-for-java</url>
+    <connection>scm:git:git@github.com:Azure/azure-sdk-for-java.git</connection>
+    <tag>HEAD</tag>
+  </scm>
+
+  <properties>
+    <project.automatic.module.name>com.azure.messaging.servicebus</project.automatic.module.name>
```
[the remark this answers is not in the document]
hemanttanwar: removing it.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueReceiverAsyncClient.java
```diff
@@ -0,0 +1,282 @@
...
+        return null;
+    }
+
+    /**
+     *
+     * @param lockToken to be used.
+     * @param context The {@link TransactionContext} to be used.
+     * @return The {@link Mono} the finishes this operation on service bus resource.
+     */
+    public Mono<Void> complete(UUID lockToken, TransactionContext context) {
```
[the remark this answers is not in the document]
hemanttanwar: Since we are not implementing in first releasing. I will remove `TransactionContext ` for now. And we will add it in later beta releases.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ClientConstants.java
hemanttanwar: I have updated documentation `Declares at which time the message should appear on the Service Bus Queue.`

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueSenderAsyncClient.java
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+
+    /**
+     *
+     * @param message to be scheduled
+     * @param scheduledEnqueueTimeUt Time of the enqueue.
+     * @return The {@link Mono} the finishes this operation on service bus resource.
+     */
+    public Mono<Long> schedule(Message message, Instant scheduledEnqueueTimeUt) {
```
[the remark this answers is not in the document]
hemanttanwar: I have updated documentation Declares at which time the message should appear on the Service Bus Queue.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ClientConstants.java
hemanttanwar: `ServiceBusClientConstants`    Does this make sense ?

pom.client.xml
```diff
@@ -1231,6 +1231,7 @@
     <module>sdk/core/azure-core-test</module>
     <module>sdk/core/azure-core-tracing-opentelemetry</module>
     <module>sdk/e2e</module>
```
conniey: nit: alphabetise.

pom.client.xml
```diff
@@ -1231,6 +1231,7 @@
...
+    <module>sdk/servicebus/azure-messaging-servicebus</module>
```
conniey: Did you also add this to the jacoco test coverage? I'm not sure if spotbugs aggregate report also requires an entry.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
conniey: It might be worth creating an issue for it so we don't lose track of this.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
```diff
@@ -0,0 +1,335 @@
...
+    public String getBodyAsString() {
+        return new String(body, UTF_8);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
```
conniey: Is this equality check correct? I'd assume there are other properties to compare for a Message.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,338 @@
...
+ * The builder to create {@link QueueReceiverAsyncClient} and {@link QueueSenderAsyncClient}.
+ */
+@ServiceClientBuilder(serviceClients = { QueueReceiverAsyncClient.class, QueueSenderAsyncClient.class})
+public final class QueueClientBuilder {
+
+    private static final String AZURE_SERVICE_BUS_CONNECTION_STRING = "AZURE_SERVICE_BUS_CONNECTION_STRING";
+    private static final AmqpRetryOptions DEFAULT_RETRY =
+        new AmqpRetryOptions().setTryTimeout(ServiceBusClientConstants.OPERATION_TIMEOUT);
+
+    private static final String SERVICEBUS_PROPERTIES_FILE = "azure-messaging-servicebus.properties";
```
conniey: This doesn't match the actual file name.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,338 @@
...
+     */
+    public QueueClientBuilder connectionString(String connectionString) {
+        final ConnectionStringProperties properties = new ConnectionStringProperties(connectionString);
+        final TokenCredential tokenCredential;
+        try {
+            tokenCredential = new ServiceBusSharedKeyCredential(properties.getSharedAccessKeyName(),
+                properties.getSharedAccessKey(), ServiceBusClientConstants.TOKEN_VALIDITY);
+        } catch (Exception e) {
+            throw logger.logExceptionAsError(
+                new AzureException("Could not create the ServiceBusSharedKeyCredential.", e));
```
conniey: I'd expect an IllegalArgumentException.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,338 @@
...
+     * @param connectionString to connect to Queue.
+     * @return The updated {@link QueueClientBuilder} object.
+     */
+    public QueueClientBuilder connectionString(String connectionString) {
+        final ConnectionStringProperties properties = new ConnectionStringProperties(connectionString);
+        final TokenCredential tokenCredential;
+        try {
+            tokenCredential = new ServiceBusSharedKeyCredential(properties.getSharedAccessKeyName(),
+                properties.getSharedAccessKey(), ServiceBusClientConstants.TOKEN_VALIDITY);
+        } catch (Exception e) {
```
conniey: Catching exception? Shouldn't we catch a specific one?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,338 @@
...
+     * @param fullyQualifiedNamespace for the Service Bus.
+     * @param queueName The name of the queue.
+     * @param credential {@link TokenCredential} to be used for authentication.
+     * @return The updated {@link QueueClientBuilder} object.
+     */
+    public QueueClientBuilder credential(String fullyQualifiedNamespace, String queueName, TokenCredential credential) {
+        this.fullyQualifiedNamespace = Objects.requireNonNull(fullyQualifiedNamespace,
+            "'fullyQualifiedNamespace' cannot be null.");
+        this.credentials = Objects.requireNonNull(credential, "'credential' cannot be null.");
+        this.queueName = Objects.requireNonNull(queueName, "'entityPath' cannot be null.");
```
conniey: ''queueName" not "entityPath"

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,338 @@
...
+        final ServiceBusConnectionProcessor connectionProcessor = createConnectionProcessor(messageSerializer);
+
+        final TracerProvider tracerProvider = new TracerProvider(ServiceLoader.load(Tracer.class));
+
+        return new QueueReceiverAsyncClient(connectionProcessor.getFullyQualifiedNamespace(), queueName,
+            connectionProcessor, tracerProvider, messageSerializer, prefetchCount);
+    }
+
+    QueueReceiverAsyncClient createAsyncReceiverClient(ReceiveMode receiveMode, int prefetchCount) {
+        return createAsyncReceiverClient(prefetchCount);
```
conniey: Why does this method exist?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueReceiverAsyncClient.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /**
+     *
+     * @param lockToken to be used.
+     * @return The {@link Mono} the finishes this operation on service bus resource.
+     */
+    public Mono<Void> abandon(UUID lockToken) {
+        //TODO(feature-to-implement)
+        return null;
```
conniey: Imho, returning Mono.empty() is more friendly than returning null for writing API tests before implementation.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueSenderAsyncClient.java
```diff
@@ -0,0 +1,352 @@
...
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiConsumer;
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collector;
+
+import static com.azure.core.util.tracing.Tracer.ENTITY_PATH_KEY;
```
conniey: Do your import statements need to be reordered?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueSenderAsyncClient.java
```diff
@@ -0,0 +1,352 @@
...
+     */
+    public static final int MAX_MESSAGE_LENGTH_BYTES = 256 * 1024;
+
+    /**
+     * Creates a new instance of this {@link QueueSenderAsyncClient} that sends messages to
+     */
+    QueueSenderAsyncClient(String queueName, ServiceBusConnectionProcessor connectionProcessor,
+                           AmqpRetryOptions retryOptions, TracerProvider tracerProvider,
+                           MessageSerializer messageSerializer) {
+        // Caching the created link so we don't invoke another link creation.
```
conniey: This comment doesn't apply.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueSenderAsyncClient.java
```diff
@@ -0,0 +1,352 @@
...
+        this.messageSerializer = Objects.requireNonNull(messageSerializer,
+            "'messageSerializer' cannot be null.");
+        this.retryOptions = Objects.requireNonNull(retryOptions, "'retryOptions' cannot be null.");
+        this.queueName = Objects.requireNonNull(queueName, "'entityPath' cannot be null.");
```
conniey: This message is incorrect.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueSenderAsyncClient.java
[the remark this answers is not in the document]
conniey: I think he wants to see scheduledEnqueueTimeUtc

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ReceiveMode.java
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.messaging.servicebus;
+
+/**
+ *  Defines the Receive modes.
+ */
+public enum ReceiveMode {
```
conniey: Should this be in a model package class?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
```diff
@@ -0,0 +1,349 @@
...
+     *
+     * @throws IllegalArgumentException if {@code object} is not an instance of {@link Message}.
+     */
+    @Override
+    public <T> org.apache.qpid.proton.message.Message serialize(T object) {
+        Objects.requireNonNull(object, "'object' to serialize cannot be null.");
+
+        if (!(object instanceof Message)) {
+            throw logger.logExceptionAsError(new IllegalArgumentException(
+                "Cannot serialize object that is not EventData. Clazz: " + object.getClass()));
```
conniey: This message is incorrect.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
```diff
@@ -0,0 +1,349 @@
...
+    @Override
+    public <T> T deserialize(org.apache.qpid.proton.message.Message message, Class<T> clazz) {
+        Objects.requireNonNull(message, "'message' cannot be null.");
+        Objects.requireNonNull(clazz, "'clazz' cannot be null.");
+
+        if (clazz == Message.class) {
+            return (T) deserializeMessage(message);
+        } else {
+            throw logger.logExceptionAsError(new IllegalArgumentException(
+                "Deserialization only supports ServiceBusProperties."));
```
conniey: What? ServiceBusproperties doesn't exist.

conniey: It's easier to reason about by flipping this if statement:

```java
if (clazz != Message.class) {
    throw logger...
}

return (T)...
````

File: sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java [hunk 69]
```diff
@@ -0,0 +1,349 @@
...
+            Data bodyData = (Data) bodySection;
+            body = bodyData.getValue().getArray();
+        } else {
+            logger.warning(String.format(Messages.MESSAGE_NOT_OF_TYPE,
+                bodySection != null ? bodySection.getType() : "null"));
+
+            body = new byte[0];
+        }
+
+        final Message eventData = new Message(body, receiveProperties, Context.NONE);
```
conniey: I like how it's called eventData here.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/CreateBatchOptions.java
conniey: Why does this exist??

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
conniey: Do you have a test for this?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/Messages.java
```diff
@@ -0,0 +1,50 @@
...
+import com.azure.core.util.logging.ClientLogger;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * Define various messages for different error conditions.
+ */
+public class Messages {
```
conniey: Messages is an overloaded term because SB exposes a "Message" class. I would rename this to something else like ResourceStrings or something.

sdk/servicebus/pom.service.xml
```diff
@@ -8,7 +8,34 @@
...
+        <property>
+          <name>env.SDKTYPE</name>
+          <value>!data</value>
+        </property>
+      </activation>
+      <modules>
+        <module>../core/azure-core</module>
+        <module>../core/azure-core-test</module>
+        <module>../identity/azure-identity</module>
+        <module>../core/azure-core-amqp</module>
```
conniey: alphabetise.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueSenderAsyncClientTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+        // Create an event to send.
+        Message message = new Message("Hello world!".getBytes(UTF_8));
+
+        // Send that event. This call returns a Mono<Void>, which we subscribe to. It completes successfully when the
+        // event has been delivered to the Service Bus. It completes with an error if an exception
+        // occurred while sending the event.
+        StepVerifier.create(asyncSender.send(message))
+            .verifyComplete();
+        Thread.sleep(1000 * 5);
```
conniey: This isn't needed.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueSenderAsyncClientTest.java
```diff
@@ -0,0 +1,79 @@
...
+        // The connection string value can be obtained by:
+        // 1. Going to your Event Hubs namespace in Azure Portal.
+        // 2. Creating an Event Hub instance.
+        // 3. Creating a "Shared access policy" for your Event Hub instance.
+        // 4. Copying the connection string from the policy's properties.
+
+        String connectionString = baseConnectionString + ";EntityPath=queue-test1";
+
+        // Instantiate a client that will be used to call the service.
+        QueueSenderAsyncClient asyncSender = new QueueClientBuilder()
```
conniey: Make sure to dispose of your client in a try/finally.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueReceiverAsyncClientTest.java
```diff
@@ -0,0 +1,192 @@
...
+    private static final byte[] PAYLOAD_BYTES = PAYLOAD.getBytes(UTF_8);
+    private static final int PREFETCH = 1;
+
+    private final ClientLogger logger = new ClientLogger(QueueReceiverAsyncClientTest.class);
+    private final String messageTrackingUUID = UUID.randomUUID().toString();
+    private final DirectProcessor<org.apache.qpid.proton.message.Message> messageProcessor = DirectProcessor.create();
+    private final DirectProcessor<Throwable> errorProcessor = DirectProcessor.create();
+    private final DirectProcessor<AmqpEndpointState> endpointProcessor = DirectProcessor.create();
+    private final DirectProcessor<AmqpShutdownSignal> shutdownProcessor = DirectProcessor.create();
+    private final String baseConnectionString = System.getenv("AZURE_SERVICEBUS_CONNECTION_STRING");
```
conniey: This doesn't match the one you fetch from configuration in QueueClientBuilder.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueReceiverAsyncClientTest.java
```diff
@@ -0,0 +1,192 @@
...
+            applicationProperties.put(MESSAGE_TRACKING_ID, messageTrackingValue);
+        }
+
+        message.setApplicationProperties(new ApplicationProperties(applicationProperties));
+        message.setBody(new Data(new Binary(contents)));
+
+        return message;
+    }
+}
+
```
conniey: nit: uneeded extra line.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueSenderAsyncClientTest.java
```diff
@@ -0,0 +1,79 @@
...
+        // Send that event. This call returns a Mono<Void>, which we subscribe to. It completes successfully when the
+        // event has been delivered to the Event Hub. It completes with an error if an exception occurred while sending
+        // the event.
+        List<Message> list = new ArrayList<Message>();
+
+        list.add(message1);
+        list.add(message2);
+        StepVerifier.create(asyncSender.send(list))
+            .verifyComplete();
+        Thread.sleep(5000);
```
conniey: You don't need this.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueSenderAsyncClientTest.java
```diff
@@ -0,0 +1,79 @@
...
+    public void testPublishMultipleMessage() throws Exception {
+        // The connection string value can be obtained by:
+        // 1. Going to your Service Bus namespace in Azure Portal.
+        // 2. Creating an Service Bus instance.
+        // 3. Creating a "Shared access policy" for your Service Bus instance.
+        // 4. Copying the connection string from the policy's properties.
+
+        String connectionString = baseConnectionString + ";EntityPath=queue-test1";
+        // Instantiate a client that will be used to call the service.
+        QueueSenderAsyncClient asyncSender = new QueueClientBuilder()
```
conniey: Dispose of client.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueSenderAsyncClientTest.java
```diff
@@ -0,0 +1,79 @@
...
+            .verifyComplete();
+        Thread.sleep(1000 * 5);
+    }
+
+    /**
+     * Main method to invoke this demo on how to send a message to an Azure Service Bus.
+     */
+    @Test
+    public void testPublishMultipleMessage() throws Exception {
+        // The connection string value can be obtained by:
```
conniey: This isn't a sample. You can remove these comments.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/QueueSenderAsyncClientTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+public class QueueSenderAsyncClientTest {
+
+    private final String baseConnectionString = System.getenv("AZURE_SERVICEBUS_CONNECTION_STRING");
+    /**
+     * Main method to invoke this demo on how to send a message to an Azure Event Hub.
+     */
+    @Test
+    public void testPublishSingleMessage() throws Exception {
+        // The connection string value can be obtained by:
```
conniey: This isn't a sample, you can remove these comments.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
```diff
@@ -0,0 +1,343 @@
...
+
+        RESERVED_SYSTEM_PROPERTIES = Collections.unmodifiableSet(properties);
+    }
+
+    /**
+     * Creates an event containing the {@code body}.
+     *
+     * @param content The data to set for this event.
+     * @throws NullPointerException if {@code body} is {@code null}.
+     */
```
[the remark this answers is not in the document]
hemanttanwar: changed everything to `body`  as AMQP protocol uses this term.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,338 @@
...
+
+    private ProxyOptions proxyOptions;
+    private TokenCredential credentials;
+    private Configuration configuration;
+    private AmqpRetryOptions retryOptions;
+    private Scheduler scheduler;
+    private AmqpTransportType transport = AmqpTransportType.AMQP;
+    private String fullyQualifiedNamespace;
+    private String queueName;
+    private ServiceBusConnectionProcessor servicerBusConnectionProcessor;
```
richardpark-msft:
```suggestion
    private ServiceBusConnectionProcessor serviceBusConnectionProcessor;
```
Or go full bore and add more 'r's.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /**
+     *
+     * @param fullyQualifiedNamespace for the Service Bus.
+     * @param queueName The name of the queue.
+     * @param credential {@link TokenCredential} to be used for authentication.
+     * @return The updated {@link QueueClientBuilder} object.
+     */
+    public QueueClientBuilder credential(String fullyQualifiedNamespace, String queueName, TokenCredential credential) {
```
richardpark-msft: Probably a good point for us to sync on this - I think we're calling the `fullyQualifiedNamespace` just `host`.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueReceiverAsyncClient.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /**
+     * Completes a {@link Message} using its lock token. This will delete the message from the service.
+     *
+     * @param lockToken to be used.
+     * @return The {@link Mono} the finishes this operation on service bus resource.
+     */
+    public Mono<Void> complete(UUID lockToken) {
+        //TODO(feature-to-implement)
```
richardpark-msft: In the JS stuff I just did I just took the entire message as the parameter. What do you think about that? I figure it requires the user to know less - just give me the whole object.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueSenderAsyncClient.java
```diff
@@ -0,0 +1,353 @@
...
+     * The default maximum allowable size, in bytes, for a batch to be sent.
+     */
+    public static final int MAX_MESSAGE_LENGTH_BYTES = 256 * 1024;
+
+    /**
+     * Creates a new instance of this {@link QueueSenderAsyncClient} that sends messages to
+     */
+    QueueSenderAsyncClient(String queueName, ServiceBusConnectionProcessor connectionProcessor,
+                           AmqpRetryOptions retryOptions, TracerProvider tracerProvider,
+                           MessageSerializer messageSerializer) {
```
richardpark-msft: What are some examples of message serializers? I'm guessing we're all going to have JSON, strings. Do you have any Java specific serializers?

pom.client.xml
JimSuplizio: @conniey, good catch. Looking at the spotbugs aggregate report pom file, it only seems to include data libraries. I think this just need an entry in jacoco.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
hemanttanwar: Created https://github.com/Azure/azure-sdk-for-java/issues/8384

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueReceiverAsyncClient.java
hemanttanwar: I like the idea of "user know less". This is one place where all the languages can be same. Lets get it finalized during API review. One point to consider , does user have to cache whole object instead of id if they are dealing with multiple messages.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
hemanttanwar: I got impression that EH had issues where useres were confused with `host` . I am open to idea but we should get review by EH team.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueSenderAsyncClient.java
hemanttanwar: Currently, this serializer will support converting AMQP message in Java SDK representation `Message` . Making sure the property from AMQP message are translated properly. As of now Binary and json will be supported. We have to bring parity with track 1.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/QueueClientBuilder.java
hemanttanwar: Added test for this and others.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/Message.java
hemanttanwar: Right now, It is comparing object and `Message` body. As I understand more about properties, we can add those checks.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ClientConstants.java
hemanttanwar: Removed this class for now.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RequestResponseChannel.java
```diff
@@ -91,18 +91,25 @@
         this.operationTimeout = retryOptions.getTryTimeout();
         this.retryPolicy = RetryUtil.getRetryPolicy(retryOptions);
 
+        // Create sender
         this.replyTo = entityPath.replace("$", "") + "-client-reply-to";
         this.messageSerializer = messageSerializer;
+        session.setOutgoingWindow(Integer.MAX_VALUE);
```
conniey: This is used by Event Hubs too. Why are we setting You should be creating PRs so we can review these changes.

session.open() has to be invoked on the reactor dispatcher thread.

PR #8186, opened 2020-02-13
Added test.yml file for live test pipeline

sima-zhu:
Connie is working on decoupling search live tests from storage, so I did not add storage credential in test.yml. It will fail for a while until #8190 is checked in.


eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -92,7 +94,7 @@ jobs:
       - task: Maven@3
         displayName: $(DisplayName)
         inputs:
-          mavenPomFile: sdk/${{parameters.ServiceDirectory}}/pom.service.xml
+          mavenPomFile: ${{ parameters.MavenPom}} #${{ parameters.MavenPom}} 'sdk/${{parameters.ServiceDirectory}}/azure-search/pom.xml' 'sdk/${{parameters.ServiceDirectory}}/pom.service.xml'
```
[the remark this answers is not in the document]
mitchdenny: Yep. What we should be doing here is using the -pl and -am switches in our pipelines to scope the test runs more appropriately. And Connie is cooking that up right now.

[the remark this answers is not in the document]
conniey: https://github.com/Azure/azure-sdk-for-java/pull/8190 is the PR that should fix this. :)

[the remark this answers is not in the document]
sima-zhu: Revert the changes

sima-zhu:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=263613&view=logs&j=4d5db6ce-0b7f-527e-b115-2367ee6e1fef

sdk/search/azure-search/src/test/java/com/azure/search/SearchIndexClientTestBase.java
```diff
@@ -92,22 +92,20 @@
...
-                .indexName(indexName)
-                .httpClient(interceptorManager.getPlaybackClient());
+        SearchIndexClientBuilder builder = new SearchIndexClientBuilder().endpoint(endpoint)
+            .indexName(indexName);
+        if (interceptorManager.isPlaybackMode()) {
+            return builder.httpClient(interceptorManager.getPlaybackClient());
+        }
+        builder.httpClient(new NettyAsyncHttpClientBuilder().wiretap(true).build())
+            .credential(searchApiKeyCredential)
+            .addPolicy(new RetryPolicy())
```
alzimmermsft: _Not this PR and for documentation purposes._

A `RetryPolicy` should be implicitly added in the builder, right now this will cause the create pipeline to have two retry policies.

sdk/search/azure-search/src/test/java/com/azure/search/SearchIndexClientTestBase.java
```diff
@@ -92,22 +92,20 @@
...
+            .addPolicy(new HttpLoggingPolicy(
+                new HttpLogOptions().setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS)));
```
alzimmermsft: _Not this PR and for documentation purposes._

This change would ideally be done by setting the `HttpLogOptions` within the builder. Right now this will cause the generated pipeline to have two logging policies which is less than ideal when running live tests with logging on.

sdk/search/test.yml
```diff
@@ -0,0 +1,14 @@
+trigger: none
+
+jobs:
+    - template: ../../eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+      parameters:
+          ServiceDirectory: search
+          DisableAzureResourceCreation: true
+          Timeout: 120
```
alzimmermsft: Given that we've done a live test run about how long did it take to complete? We don't want to have timeouts that are too long in cases where tests deadlock and hold a DevOps VMs for longer than needed. Is it possible to use the default timeout of 60 minutes for the Search live tests?

sdk/search/azure-search/src/test/java/com/azure/search/SearchServiceTestBase.java
```diff
@@ -656,4 +660,22 @@ ServiceStatistics getExpectedServiceStatistics() {
             .setCounters(serviceCounters)
             .setLimits(serviceLimits);
     }
+
+    static boolean liveMode() {
+        return setupTestMode() == TestMode.LIVE;
+    }
```
alzimmermsft: Should this be replaced with the `isLiveMode` method in `InterceptorManager`?

sdk/search/ci.yml
```diff
@@ -3,9 +3,9 @@
 
 resources:
   repositories:
-    # - repository: azure-sdk-build-tools
-    #   type: git
-    #   name: internal/azure-sdk-build-tools
+     - repository: azure-sdk-build-tools
+       type: git
+       name: internal/azure-sdk-build-tools
```
alzimmermsft: `yml` files are space sensitive, this is using the wrong character offset. This will either lead to a parsing exception or the configuration being ignored.

sdk/search/test.yml
```diff
@@ -0,0 +1,14 @@
...
+          EnvVars:
+              AZURE_TEST_MODE: LIVE
+              AZURE_TENANT_ID: $(aad-azure-sdk-test-tenant-id)
+              AZURE_CLIENT_ID: $(aad-azure-sdk-test-client-id)
+              AZURE_CLIENT_SECRET: $(aad-azure-sdk-test-client-secret)
+              AZURE_SUBSCRIPTION_ID: $(aad-azure-sdk-test-subscription-id)
```
alzimmermsft: `yml` files are space sensitive, this is using the wrong character offset. This will either lead to a parsing exception or the configuration being ignored.

sdk/search/azure-search/src/test/java/com/azure/search/SearchServiceTestBase.java
sima-zhu: Yes. Try to replace with the one.

sima-zhu: My concern is next release will depend on azure-core-test. So I put a helper method currently in search. Will move to interceptorManager.isLiveMode once azure-core-test is released.

sdk/search/test.yml
sima-zhu: Search is depending storage currently, so 60 minute usually cannot finish search tests. Besides of that, search also takes long time to run, which might need to improve in future.

sdk/search/test.yml
```diff
@@ -0,0 +1,14 @@
+trigger: none
+
+jobs:
+    - template: ../../eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+      parameters:
+          ServiceDirectory: search
```
alzimmermsft: Could we add a `maxParallel` configuration and set it to 2, maybe 3.

https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#matrix

FYI @tg-msft, this should help reduce the number of concurrent testing jobs happening at the same time.

sima-zhu: Set to 3 for initial check in

PR #8187, opened 2020-02-13
Update library versions

conniey:
* Updates Event Hubs and Checkpoint Store versions. So we can release a minor version fix.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/InMemoryCheckpointStore.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/samples/java/com/azure/messaging/eventhubs/InMemoryCheckpointStore.java
@@ -117,7 +117,7 @@ public class InMemoryCheckpointStore implements CheckpointStore {
         }
 
         String prefix = prefixBuilder(checkpoint.getFullyQualifiedNamespace(), checkpoint.getEventHubName(),
-            checkpoint.getEventHubName(), CHECKPOINT);
+            checkpoint.getConsumerGroup(), CHECKPOINT);
         checkpointsMap.put(prefix + SEPARATOR + checkpoint.getPartitionId(), checkpoint);
         logger.info("Updated checkpoint for partition {} with sequence number {}", checkpoint.getPartitionId(),
             checkpoint.getSequenceNumber());


sdk/eventhubs/azure-messaging-eventhubs/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Release History
 
-## 5.1.0-beta.1 (Unreleased)
-
+## 5.0.2 (2020-02-13)
+- Fix bug where consumers would not be able to block in their consuming code.
```
srnagar: Is this only for the consumer or producer too?

sdk/search/azure-search/pom.xml
```diff
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.3.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core;dependency} -->
+      <version>1.3.0-beta.2</version> <!-- {x-version-update;unreleased_com.azure:azure-core;dependency} -->
```
srnagar: This is azure-search. Don't think this should be updated.

sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/pom.xml
```diff
@@ -40,7 +40,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-messaging-eventhubs</artifactId>
-      <version>5.1.0-beta.1</version> <!-- {x-version-update;com.azure:azure-messaging-eventhubs;current} -->
+      <version>5.0.2</version> <!-- {x-version-update;com.azure:azure-messaging-eventhubs;current} -->
```
srnagar: Interesting, we have already update storage blob version too.

sdk/search/azure-search/pom.xml
conniey: The build was failing before this change because it was complaining that a version wasn't updated.

JimSuplizio: This was fixed yesterday with https://github.com/Azure/azure-sdk-for-java/pull/8188. This mismatch was the result of two independency PRs being auto-merged. You'll have to rebase and push another commit without this file.

PR #8189, opened 2020-02-13
Use Unreleased Dependency of Azure Core Netty in Search

alzimmermsft:
Currently Search is depending on version `1.3.0-beta.2` of Azure Core but is using version `1.3.0` of Netty which depends on `1.2.0` Azure Core. This PR updates Search to use the unreleased version of Netty, `1.4.0-beta.2`, which depends on Azure Core `1.3.0-beta.2` to align the dependencies of Azure Core Search has. Additionally, updates the CODEOWNERS for Search to list @sima-zhu and myself.

### DIFF
--- a/eng/versioning/version_client.txt
+++ b/eng/versioning/version_client.txt
@@ -38,4 +38,4 @@ com.azure:azure-storage-queue;12.3.0;12.4.0-beta.1
 # note: the unrelease dependencies do not require the empty line as their versions
 # will not be manipulated with the automatic PR creation code.
 unreleased_com.azure:azure-core;1.3.0-beta.2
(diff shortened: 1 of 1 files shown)


sdk/search/azure-search/pom.xml
```diff
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.3.0-beta.2</version> <!-- {x-version-update;unreleased_com.azure:azure-core;dependency} -->
+      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
```
sima-zhu: Search asks to use new page API which depends on azure-core 1.3.0.

PR #8190, opened 2020-02-13
Only invoking tests in a module.

conniey:
* Add PS script to select modules to run
* Update Event Hubs live test not to create storage resources.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
@@ -103,7 +103,7 @@ jobs:
         displayName: Install modules
         inputs:
           mavenPomFile: sdk/${{parameters.ServiceDirectory}}/pom.service.xml
-          options: --no-transfer-progress -Dmaven.test.skip -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true
+          options: --no-transfer-progress -DskipTests -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true
           mavenOptions: '$(MemoryOptions)'
           javaHomeOption: 'JDKVersion'
           jdkVersionOption: '1.11'


eng/Get-ServiceModules.ps1
```diff
@@ -0,0 +1,64 @@
...
+    [ValidateSet("data", "client")]
+    [string]$SdkType
+)
+
+# By default stop for any error.
+if (!$PSBoundParameters.ContainsKey("ErrorAction")) {
+    $ErrorActionPreference = "Stop"
+}
+
+$root = Join-Path "$PSScriptRoot/../sdk" $ServiceDirectory | Resolve-Path
```
mitchdenny: To make this script more reusable I would recommend that the ServiceDirectory just be changed to Directory and that it scans for all POMs under that directory regardless. It doesn't need to know anything about the directory structure in the script since once it has the POMs it is just producing the argument set based on the groupId and artifactId.

mitchdenny:
Hrm ... interesting side effect of doing this is that its picking up ```azure-core-management```. I wonder whether we should opt for a different approach here and only build the things that we intend to ship. in that case we would process the ```ci.yml``` file to get the groupId and artifactId (name) to build the string.

eng/Get-ServiceModules.ps1
```diff
@@ -0,0 +1,64 @@
...
+
+switch($SdkType) {
```
JimSuplizio: @conniey I'm going to recommend that you not use the directory names to filter but instead look at the pom.xml files under that area subdirectory and use the parent entry to filter out which is which.
Take a look at what I'm doing for the dependency graph generation which only looks at client but you can see the logic. Here's the [loop](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/DependencyGraph/gen-dep-graph-data.ps1#L103). Here is where I'm [checking the parent](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/DependencyGraph/gen-dep-graph-data.ps1#L119). What you'd do set the ValidParent to azure-client-sdk-parent or azure-data-sdk-parent and key off of that. I've found that to be more accurate and it has the added benefit of not needing to be changed if we ever have to manipulate 'too long' directory names.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -228,11 +228,34 @@ jobs:
         condition: and(succeeded(), eq(variables['ShouldRunSourceTests'],'true'))
         displayName: 'Update versions for source build'
 
+      - pwsh: |
```
sima-zhu: In CI, we build all modules originally.  How do you use TestFromSource to indicate whether it is from maven or from local module?

conniey: There is a set-version in python that updates all the pom.xmls to use current-version.

sima-zhu:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=261975&view=logs&j=27eb5d14-7214-5d9c-6bac-653552bd18b3&t=8c9b85f5-3240-5af8-b929-e53e815dfe57

sima-zhu:
My thinking here.
* If you are using dependency from local, then the live tests still need to run tests from the module. e.g. If you are depending unreleased storage (or core) in EH, then storage tests will get executed. It still have a chance to couple with other module.
* If say, EH rarely depend on unreleased version. Then there is no need to put it in pom.service.xml.

My suggestion is to install pom.service.xml, but only test on modules you fetch from script.

conniey:
> My suggestion is to install pom.service.xml, but only test on modules you fetch from script.

We do install pom.service.xml. The arguments are only added to the "Run tests" step.

eng/Get-ServiceModules.ps1
```diff
@@ -0,0 +1,99 @@
...
+        $parentArtifactId = "azure-data-sdk-parent"
+        break
+    }
+    default {
+        throw "'$SdkType' is not supported."
+    }
+}
+
+# If it is not a module that is built by pom.client.xml, we should skip it.
+$clientModules = New-Object System.Collections.Generic.HashSet[string]
```
mitchdenny: This is going to be fragile. In the near future we are going to be removing the ```<module />``` elements from the parent POM which means you won't be able to rely on this.

eng/Get-ServiceModules.ps1
```diff
@@ -0,0 +1,99 @@
...
+[xml]$clientPom = Get-Content -Path (Join-Path "$PSScriptRoot/../" "pom.client.xml")
```
mitchdenny: Hard coded relative path.

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -89,6 +90,41 @@ jobs:
...
+          options: --no-transfer-progress -DskipTests -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true
+          mavenOptions: '$(MemoryOptions)'
+          javaHomeOption: 'JDKVersion'
+          jdkVersionOption: '1.11'
+          jdkArchitectureOption: 'x64'
+          publishJUnitResults: false
+          goals: 'install'
+
+      - pwsh: |
+          $modules = $(eng/Get-ServiceModules.ps1 ${{parameters.ServiceDirectory}} ${{parameters.SDKType}})
```
mitchdenny: This code is duplicated. Can it be placed into a wrapper script around your script so that it isn't duplicated? Another option that we've been thinking of is adding a Prepare job into the Build stage which Build, Test and Analyze depend on which could run this script in one location. It is possible to export variables between jobs for this purpose.

eng/Get-ServiceModules.ps1
conniey: That's why it was trying to build azure-core-management, because it couldn't be found in pom.client.xml and azure-core's pom.service.xml is a reference to pom.client.xml. Any other suggestions?

eng/Get-ServiceModules.ps1
conniey: I saw this in other build scripts. Suggestions?

mitchdenny: If we must take a dependency on ```pom.client.xml``` then I'd prefer to pass the path in since I know its location is going to change. Preferably I would hope that we don't need to worry what the parent POM is. I mean in effect don't we just want to build all the modules that are referenced by ```pom.service.xml``` that sit under ```sdk/[servicedir]```? Why do we need to consult the parent at all?

conniey: The reason for this is this. We were getting the error that it didn't know about module `azure-core-management`.
https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/pom.service.xml#L11-L13

eng/Get-ServiceModules.ps1
mitchdenny: This is going to be messy :(

We kinda lean on the ```java - core -ci```/```java - core``` pipelines to build everything as a bit of a broad integration test, which is why we reference ```pom.client.xml``` as a proxy for building those things that depend on core in any way. It's a bit of a hack.

The more correct thing to do is basically have another script which scans everything under SDK looking for modules that take a dependency on the modules under ```sdk/[servicedir]``` and also build them. So for example storage would be unable to check anything in that broke event hubs (its not just about core).

The problem is that without restructuring our parent POMs (or at least using pure aggregators) we aren't going to be able to do the right thing here. Which means that this code will need to be reworked pretty soon.

The right thing (but big effort thing) to do here is:

1. Place a true aggregate POM in ```./pom.xml``` which then references each service POM.
2. Change build scripts to build against that root POM using the arguments that you are generating here.
3. Change your script to do two scans. The first scan scans ```sdk/[service]/*``` for the complete list of POMs under a service directory. Then scan everything under ```sdk/*``` and build a list of targets that depend directly on the modules in the target ```sdk/[service]``` folder. Then taking that list, and the target directory modules create the ```-pl``` switch.

This would net us a few benefits. First, you wouldn't inadvertently build storage for event hubs. However, storage would build event hubs, because it depends on storage. For live testing we probably wouldn't include downstream dependents in the test.

Sadly the correct thing to do here seems like a lot of work.

JimSuplizio: > This is going to be fragile. In the near future we are going to be removing the `<module />` elements from the parent POM which means you won't be able to rely on this.

@mitchdenny There is a lot of tooling which relies on inspecting the parent pom entry to figure out what track pom belongs to which track.

weshaggard:
Can I get a little more context on this change? Before we go and start parsing the pom files directly have we investigated a way to do this in maven itself?

conniey:
@weshaggard The problem is that many of our pipelines ultimately call `sdk/<client-library>/pom.service.xml`. The problem is that maven runs the goal on _all_ the modules in pom.service.xml. You can't tell it to skip particular modules unless you use the --projects parameter.

Since we have module references to azure-core, netty, these things end up being compiled/tested/analyzed, which takes up unnecessary computation time and test failures unrelated to the `<client-library>`. This is why event hubs live tests are always failing because it also runs the storage live tests.

weshaggard:
Is there not a way to use profiles for this? We could include the source dependencies for the build goal and not include them for the test goal?

conniey:
@weshaggard Good idea! We had thought of using profiles but not for that specific goal. I'll take a look

sdk/eventhubs/test-resources.json
```diff
@@ -49,17 +45,11 @@
...
-        "contributorRoleId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c')]",
         "eventHubsApiVersion": "2017-04-01",
         "eventHubsDataOwnerRoleId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/f526a384-b230-433a-b45c-95f59c4a2dec')]",
         "eventHubsNamespaceKeyName": "RootManageSharedAccessKey",
         "eventHubsNamespaceName": "[concat('eventhub', parameters('baseName'))]",
-        "location": "[resourceGroup().location]",
-        "premiumAccountName": "[concat('prem', parameters('baseName'))]",
-        "primaryAccountName": "[concat('prim', parameters('baseName'))]",
-        "secondaryAccountName": "[concat('sec', parameters('baseName'))]",
-        "storageApiVersion": "2019-04-01"
```
alzimmermsft: Does Checkpoint Store have tests that would require these?

conniey:
Closed in favour of #8242

PR #8223, opened 2020-02-14
PR to fix publisher issue

conniey:


sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/pom.xml
```diff
@@ -40,7 +40,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-messaging-eventhubs</artifactId>
-      <version>5.0.1</version> <!-- {x-version-update;com.azure:azure-messaging-eventhubs;current} -->
+      <version>5.0.2</version> <!-- {x-version-update;com.azure:azure-messaging-eventhubs;current} -->
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
```
srnagar: Should we take this opportunity to update blob storage to 12.3.1 as well?

PR #8242, opened 2020-02-18
POM Restructuring Phase #1

mitchdenny:
This PR is the first step in a number of changes designed to restructure the relationship between our builds and parent POMs. Before this PR our parent POMs functioned as aggregators for developers who wanted to build the whole tree, we then had seperate ```pom.service.xml``` files that the pipelines used.

This change repurposes ```pom.xml``` in the root of the repo and makes it a true top level aggregate for the entire repo (aside from management plane, but we'll get to that). All of our track 1 and track 2 modules are referenced in the aggregate and you use the ```-pl``` and ```-am``` switches.

Our pipelines have been reconfigured to use this root aggregate and we generate the ```-pl``` switch based on the artifact list in the ```ci.yml``` file.

One of the benefits we picked up was that we no longer need to do the seperate install of ```sdk-build-tools``` because it is in the graph, so anything that depends on it will end up triggering it to build.


mitchdenny:
@JimSuplizio what are you thoughts on the side effects of this change. I know that some of the tooling that exists today depends on ```pom.client.xml``` having a list of modules. Do we know all the places that will be impacted?

mitchdenny:
@weidongxu-microsoft what would be the implications if we removed the ```<modules />``` section from ```pom.management.xml``` like I have done in this PR?

weidongxu-microsoft:
@mitchdenny I think it is fine. I don't think mgmt SDK depends on these 2 hybrid builds.

sdk/core/pom.service.xml
```diff
@@ -9,6 +9,13 @@
   <packaging>pom</packaging>
   <version>1.0.0</version><!-- Need not change for every release-->
   <modules>
-    <module>../../pom.client.xml</module>
+    <module>azure-core</module>
```
JimSuplizio: /CC @weshaggard 
The reason this was changed to pom.client.xml was to ensure that when core built everything else built. Swapping this back to just the list of core libraries is going to revert that and we're probably going to want to have whatever the "build everything for client" equivalent is going to be.

weshaggard: Correct. We needed a way to have the core pipeline build and test everything that depended on core. If you can come up with another clever way to do that we can switch otherwise I'd keep this the way it is.

mitchdenny: Yeah so I created this PR knowing this problem wasn't solved with the hope that it would spur some conversation about the issues created by the way we have this setup and other approaches that we could possibly take.

mitchdenny: Coming back to this. The way to make this build process the downstream dependents is to use the ```-amd``` switch.

mitchdenny:
@JonathanGiles @conniey @alzimmermsft @weshaggard @JimSuplizio this PR is pretty much ready to roll. Everything that needs to be building seems to be building.

mitchdenny:
> Basically it would let us build any client/data library directly from project root, by applying correct `-pl -am` option.

Correct, although I'm planning on bringing in the management plane libraries into this model as well. You'll soon be able to add all of the management plane libraries to the ```sdk/[service]/pom.xml``` file and because the ```ci.*.yml``` files explicitly reference the artifacts they produce the different pipelines will just produce the right ```-pl``` switch to build them.

To make this a reality there will need to be a few small tweaks to the management plane built template, but it should be pretty minimal.

It also means that should you wish to, we could push management plane packages through our unified pipelines.

weidongxu-microsoft:
> > Basically it would let us build any client/data library directly from project root, by applying correct `-pl -am` option.
> 
> Correct, although I'm planning on bringing in the management plane libraries into this model as well. You'll soon be able to add all of the management plane libraries to the `sdk/[service]/pom.xml` file and because the `ci.*.yml` files explicitly reference the artifacts they produce the different pipelines will just produce the right `-pl` switch to build them.
> 
> To make this a reality there will need to be a few small tweaks to the management plane built template, but it should be pretty minimal.
> 
> It also means that should you wish to, we could push management plane packages through our unified pipelines.

Em, mgmt plane could wait a bit actually. We are planning to move new libraries into azure-sdk-for-java on Peter's request. These will be the ones that integrated with Azure Core. So we might first integrate these latest libraries into `pom.xml` first. We could start with a proposal in a few weeks (conform to this pom structure).

mitchdenny:
Sounds good. I'm not in a hurry to push management plane into this model until you are ready.

eng/pipelines/templates/steps/generate-project-list.yml
```diff
@@ -0,0 +1,18 @@
...
+
+steps:
+- pwsh: |
+    Write-Host "##vso[task.setvariable variable=ProjectList;]"
+  displayName: Initialize project list variable
+
+- ${{each artifact in parameters.Artifacts }}:
+  - pwsh: |
+      Write-Host "##vso[task.setvariable variable=ProjectList;]$(ProjectList),${{artifact.groupId}}:${{artifact.name}}"
+    displayName: Append ${{artifact.groupId}}:${{artifact.name}} to project list
```
conniey: nit: add new empty line

pom.data.xml
```diff
@@ -602,6 +602,7 @@
     </profile>
   </profiles>
 
+  <!--
```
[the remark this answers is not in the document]
mitchdenny: > There's an open question about the need for the additional test module entries but that only really a concern, for me, until it's understood why they're necessary. If the situation arises where someone is adding a new library and/or area and a similar issue pops up, needing a dependency where seemingly we shouldn't, I'd like to be able to explain why it's necessary.

This is a side effect of the ```-amd``` approach I was trying to use. I'll go through the PR and remove the ones that are unnecessary now that I don't need to force the dependency graph into a particular shape.

I'll also remove the commented out secitons.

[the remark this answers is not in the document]
mitchdenny: Wait ... I need to live those modules in there for some tooling that still depends on them being there. I'll do that instead. Glad you mentioned it.

sdk/search/pom.xml
```diff
@@ -5,10 +5,10 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.azure</groupId>
-  <artifactId>azure-core-service</artifactId>
+  <artifactId>azure-search-service-parent</artifactId>
```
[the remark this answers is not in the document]
mitchdenny: Fixed.

pom.xml
```diff
@@ -4,281 +4,25 @@
...
-          <name>include-template</name>
-        </property>
-      </activation>
-      <modules>
-        <module>./sdk/template/azure-sdk-template</module>
-      </modules>
-    </profile>
-  </profiles>
-
+  <version>1.0.0</version><!-- Need not change for every release-->
```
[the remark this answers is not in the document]
mitchdenny: I changed it back to what it was, the name wasn't important to me when I started doing this I basically did a ```CTRL-A```, ```DEL``` on the file and started again and that is just the name I came back with.

sdk/storage/azure-storage-common/pom.xml
```diff
@@ -70,6 +70,12 @@
       <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core-test;dependency} -->
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>perf-test-core</artifactId>
+      <version>1.0.0-beta.1</version> <!-- {x-version-update;com.azure:perf-test-core;dependency} -->
```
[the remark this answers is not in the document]
mitchdenny: This was to work around the ```-amd``` switch issue. Will remove in an update to the PR.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -102,12 +102,16 @@ jobs:
...
+        parameters:
+          Artifacts: ${{parameters.Artifacts}}
+
       - task: Maven@3
         displayName: 'Build and Package'
         inputs:
-          mavenPomFile: sdk/${{parameters.ServiceDirectory}}/pom.service.xml
+          mavenPomFile: pom.xml
           goals: 'package'
-          options: '$(DefaultOptions) "-DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -DskipTests -Dgenerate-overview' # We include template-module so we ensure it always builds in CI
```
weshaggard: nit: I don't think the comment at the end of the line is needed any longer.

eng/pipelines/templates/jobs/archetype-sdk-data.yml
```diff
@@ -90,12 +90,16 @@ jobs:
...
+        parameters:
+          Artifacts: ${{parameters.Artifacts}}
+  
       - task: Maven@3
         displayName: 'Build and Package'
         inputs:
-          mavenPomFile: sdk/${{parameters.ServiceDirectory}}/pom.service.xml
+          mavenPomFile: pom.xml
           goals: 'package'
-          options: '$(DefaultOptions) "-DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -DskipTests -Dgenerate-overview' # We include template-module so we ensure it always builds in CI
```
weshaggard: nit: end of line comment can be deleted.

sdk/eventhubs/azure-messaging-eventhubs/pom.xml
```diff
@@ -58,6 +58,12 @@
       <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core-test;dependency} -->
       <scope>test</scope>
     </dependency>
+    <dependency> <!-- HACK: Adding this here to work around a limitation with the -amd switch in Maven. -->
```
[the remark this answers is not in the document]
mitchdenny: It was to work around a problem with the way that the ```-amd``` switch works on Maven. Given a project (e.g. ```com.azure:azure-storage-blob```) it resolves that it needs to build its dependants (e.g. ```com.azure:azure-messaging-eventhubs```) but doesn't then resolve their dependencies. So you end up with projects in the graph not being added to the build list.

Graphically:
https://miro.com/app/board/o9J_kulp0yo=/

eng/pipelines/templates/steps/generate-project-list.yml
```diff
@@ -0,0 +1,18 @@
+# This job template uses the artifact list in the ci.yml file
+# to generate a variable at build time that can be passed into
+# the Maven -pl command-line switch to constrain what subset of
+# projects are targetted based on the artifacts this pipeline
```
joshfree: British spelling of targeted? :)

mitchdenny: I am a subject of the crown guvna

pom.data.xml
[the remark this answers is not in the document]
JimSuplizio: >> This is a side effect of the -amd approach
I recognize that but is this something we can predict? What's the pattern that causes us to need the extra entry and is it something we can effectively document/communicate with the Java team so everyone knows the deal and is on the same page?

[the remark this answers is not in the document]
mitchdenny: The most likely culprits are going to be things like identity, storage, perf-test-core. Hard to predict where it'll crop up because its a function of the dependencies that individual libraries take. I think it might happen more as we introduce new services that take dependencies on fundamental libraries like identity and storage.

PR #8273, opened 2020-02-18
Make facet object strongly typed

sima-zhu:
* Make `RangeFacetResult` and `ValueFacetResult` generic type.

### DIFF
--- a/sdk/search/azure-search/src/main/java/com/azure/search/models/RangeFacetResult.java
+++ b/sdk/search/azure-search/src/main/java/com/azure/search/models/RangeFacetResult.java
@@ -4,11 +4,6 @@
 package com.azure.search.models;
 
 import com.azure.core.annotation.Immutable;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
 
 /**
  * A single bucket of a range facet query result that reports the number of documents


sdk/search/azure-search/src/main/java/com/azure/search/models/RangeFacetResult.java
```diff
@@ -10,20 +10,22 @@
...
+     * @param to Value indicates the upper bound of facet's range
      */
-    public RangeFacetResult(FacetResult facetResult) {
-        count = facetResult.getCount();
-        from = facetResult.getDocument().get("from");
-        to = facetResult.getDocument().get("to");
+    public RangeFacetResult(Long count, T from, T to) {
+        this.count = count;
+        this.from = from;
+        this.to = to;
```
brjohnstmsft: The accessor methods `getFrom` and `getTo` should return type `T` as well.

sdk/search/azure-search/src/main/java/com/azure/search/models/ValueFacetResult.java
```diff
@@ -11,27 +11,17 @@
...
-     * Constructor
-     *
-     * @param count count
-     * @param value value
-     */
-    public ValueFacetResult(Long count, Object value) {
+    public ValueFacetResult(Long count, T value) {
         this.count = count;
         this.value = value;
     }
```
brjohnstmsft: The accessor method `getValue` should return type `T` as well.

sdk/search/azure-search/src/test/java/com/azure/search/SearchTestBase.java
```diff
@@ -121,17 +124,21 @@ void assertRangeFacets(List<RangeFacetResult> baseRateFacets, List<RangeFacetRes
         Assert.assertEquals(2, lastRenovationDateFacets.get(1).getCount().intValue());
     }
 
-    List<RangeFacetResult> getRangeFacetsForField(
+    @SuppressWarnings("unchecked")
+    <T> List<RangeFacetResult<T>> getRangeFacetsForField(
         Map<String, List<FacetResult>> facets, String expectedField, int expectedCount) {
         List<FacetResult> facetCollection = getFacetsForField(facets, expectedField, expectedCount);
-        return facetCollection.stream().map(RangeFacetResult::new)
+        return facetCollection.stream().map(facetResult -> new RangeFacetResult<T>(facetResult.getCount(),
```
brjohnstmsft: The logic to create `RangeFacetResult` and `ValueFacetResult` from a `FacetResult` should be part of the SDK, not just the tests. Every customer that uses facets will need to do this, so we might as well make it easy.

sima-zhu: Do we want to add a public API in FacetResult to get RangeFacetResult/ValueFacetResult?

sima-zhu: FacetResult is auto-gen code. For the effort of getting RangeFacetResult/ValueFacetResult from FacetResult, we have to change the swagger definition.

brjohnstmsft: By "change the swagger definition" I assume you mean use an AutoRest transform directive? Because there's no way to express local language-specific methods on a Swagger type. I'm assuming you will handle that, and our team will own the actual Swagger definition of the REST API.

sima-zhu: I might misunderstood your statement before.

I currently added back the constructor of taking FacetResult inside of RangeFacetResult/ValueFacetResult. 

Let me know if furthermore functionalities needed here.

brjohnstmsft: It's definitely a step in the right direction. I'm not sure what more we can do to make this more discoverable. @tg-msft had similar concerns about this pattern for the new .NET SDK. Maybe we can find a common solution for Preview 2.

tg-msft: Tracking with https://github.com/Azure/azure-sdk-for-net/issues/10613

Issue #8149, opened 2020-02-12
[InProgress]Address API Feedback


AlexGhiondea:
**TODO:**
**Done:**
- ~~FacetResult.java property is "documents" when it can reasonably be left as "additionalProperties".~~
- ~~Rename IndexBatchImpl -> IndexBatchBase~~
- ~~Try making SuggestRequest and AutocompleteRequest package-scope (will have to make a lot of generated classes package-scope too probably)~~
- ~~Move to azure-core 1.3.0 and use ContinuablePagedFluxCore instead of PagedFluxBase~~
- ~~GeoPoint shouldn't expose coordinates directly~~
- ~~Review naming of DataSources static methods for API guidelines compliance~~
- ~~getDocumentCount() should return long, not Long~~
- ~~SearchDocumentsResult - getNextPageParameters and getNextLink should be package-scope. So should SearchRequest (if possible).~~
- ~~The document property of SearchResult and SuggestResult should not be mutable. Probably the result of an over-broad regex rule.~~
- ~~Singular names for IndexBatch methods -- apparently intentional, need to check with API review board if they should be plural.~~

sima-zhu:
* Autocomplete -> AutoComplete (Auto-gen)

sima-zhu:
Auto-gen code issue:
1. `Mono<SimpleResponse<Void>>` -> ` Mono<Response<Void>> delete`
1. `final String apiVersion = "2019-05-06";` should removed this.client.getApiVersion()
1.` VisualFeature` there are several addition.
1. `ImageAnalysisSkill` addition
1.` ImageAnalysisSkillLanguage` addition
1. `SearchMode` enum is not the same.
1. `TextSplitMode` and `TextTranslationSkillLanguage` are being deleted.

brjohnstmsft:
@sima-zhu Please do not change names unless there is a compelling reason to do so. `Skillset` is the correct casing as it is all one word (see [here](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/search/Microsoft.Azure.Search.Service/src/Generated/Models/Skillset.cs)), as is [Autocomplete](https://github.com/Azure/azure-sdk-for-net/blob/2d4097a23917c4a35ca75bc2dda1c16761b4170b/sdk/search/Microsoft.Azure.Search.Data/src/Customizations/Documents/DocumentsOperationsExtensions.cs#L67).

[...]

PR #8274, opened 2020-02-18
Add Missing Tests and Remove Exists APIs

alzimmermsft:
This PR adds tests that were missing from the Track 1 port and removes the `exists` APIs from clients. Additionally, general changes were made to documentation to follow Javadoc expectations, JUnit 4 code was removed from tests, and replaced a Storage Base64 encoding call with java.util.Base64.

The `exists` APIs were removed per request of Bruce as they didn't fit into the Search API well. The create APIs allow for create if not exists functionality which is one of the general uses of exists and is safe from race conditions, in the future if exists is wanted it is easier to add the APIs than remove them if we don't want them.


sdk/search/azure-search/src/test/java/com/azure/search/DocumentConverterTests.java
```diff
@@ -242,15 +243,55 @@ public void canReadComplexCollection() {
...
+
+        Document actualDoc = deserialize(json);
+        Assert.assertEquals(expectedDoc, actualDoc);
+    }
+
+    @Test
+    public void arraysWithOnlyNullsReadAsStringArrays() {
+        String json = "{ \"field\": [null, null] }";
+
+        // With only null elements, we can't tell what type of collection it is. For backward compatibility, we assume type string.
```
brjohnstmsft: I'm assuming that `Arrays.asList(null, null)` is returning an `ArrayList` of `Object`, not `String`. This is fine (since this SDK has never shipped, there are no backcompat concerns), but the comment and name of the test no longer reflect reality.

sdk/search/azure-search/src/test/java/com/azure/search/LookupAsyncTests.java
```diff
@@ -281,9 +267,90 @@ public void emptyDynamicObjectsInCollectionExpandedOnGetWhenCollectionFieldSelec
...
+        indexedDoc.put("Rooms", rooms);
+
+        Document expectedDoc = new Document();
+        expectedDoc.put("HotelId", "1");
+        expectedDoc.put("HotelName", OffsetDateTime.of(2015, 2, 11, 12, 58, 0, 9, ZoneOffset.UTC));
+        expectedDoc.put("Location", GeoPoint.create(40.760586, -73.975403));
+        expectedDoc.put("Rooms", Collections.singleton(new Document(Collections.singletonMap("BaseRate", "NaN"))));
+
+        IndexBatch<Document> batch = new IndexBatch<>();
+        batch.addUploadAction(expectedDoc);
```
brjohnstmsft: Looks like the bug from the Track 1 tests got ported over. This should be uploading `indexedDoc`, not `expectedDoc`.

sdk/search/azure-search/src/test/java/com/azure/search/LookupSyncTests.java
```diff
@@ -266,8 +273,67 @@ public void emptyDynamicObjectsInCollectionExpandedOnGetWhenCollectionFieldSelec
...
+        indexedDoc.put("Location", GeoPoint.create(40.760586, -73.975403)); // Test that we don't confuse Geo-JSON & complex types.
+        indexedDoc.put("Rooms", rooms);
+
+        Document expectedDoc = new Document();
+        expectedDoc.put("HotelId", "1");
+        expectedDoc.put("HotelName", OffsetDateTime.of(2015, 2, 11, 12, 58, 0, 9, ZoneOffset.UTC));
+        expectedDoc.put("Location", GeoPoint.create(40.760586, -73.975403));
+        expectedDoc.put("Rooms", Collections.singleton(new Document(Collections.singletonMap("BaseRate", "NaN"))));
+
+        client.index(new IndexBatch<>().addUploadAction(expectedDoc));
```
brjohnstmsft: Should be uploading `indexedDoc`. I'm getting paranoid now that we've made this mistake in many places and copy-pasted. 😬

sdk/search/azure-search/src/test/java/com/azure/search/SearchTestBase.java
```diff
@@ -32,6 +31,10 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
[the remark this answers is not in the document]
alzimmermsft: I'll delete all the async tests and the base test classes in the next commit 😄

sdk/search/azure-search/src/test/java/com/azure/search/LookupAsyncTests.java
```diff
@@ -281,9 +267,90 @@ public void emptyDynamicObjectsInCollectionExpandedOnGetWhenCollectionFieldSelec
...
+            .assertNext(response -> assertEquals(expectedDoc, response.getValue()))
+            .verifyComplete();
+    }
+
+    @Override
+    public void canGetDocumentWithBase64EncodedKey() {
+        createHotelIndex();
+        client = getSearchIndexClientBuilder(INDEX_NAME).buildAsyncClient();
+
+        String complexKey = Base64.getEncoder().encodeToString(new byte[]{1, 2, 3, 4, 5});
```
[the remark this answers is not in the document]
alzimmermsft: Appears that I need to use `Base64.getUrlEncoder`, that will be URL safe 😄 

https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html#getUrlEncoder--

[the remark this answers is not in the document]
alzimmermsft: If the URL encoded characters of `+` and `/` are expected on the service side this could be wrapped in a called to URLEncode instead.

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
```diff
@@ -140,256 +140,196 @@ HttpPipeline getHttpPipeline() {
...
+     * Deletes a collection of documents from the target index. See https://docs.microsoft.com/rest/api/searchservice/addupdate-or-delete-documents
      *
-     * Exception IndexBatchException is thrown when some of the indexing actions failed,
-     * but other actions succeeded and modified the state of the index.
-     * This can happen when the Search Service is under heavy indexing load.
-     * It is important to explicitly catch this exception and check the return value of
-     * its getIndexingResult method. The indexing result reports the status of each indexing
-     * action in the batch, making it possible to determine the state of the index after a
-     * partial failure.
+     * Exception IndexBatchException is thrown when some of the indexing actions failed, but other actions succeeded and
```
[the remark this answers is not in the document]
alzimmermsft: Was planning to fix those, my IntelliJ started acting up and those dropped off my mind. Will be adding in a commit soon

sdk/search/azure-search/src/test/java/com/azure/search/test/environment/models/LoudHotel.java
```diff
@@ -12,166 +12,166 @@
 
 public class LoudHotel {
     @JsonProperty(value = "HotelId")
-    private String HOTELID;
+    private String hotelId;
```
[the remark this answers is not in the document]
alzimmermsft: In tests using `LoudHotel` the response JSON would be something like the following?

```json
{
  "HOTELID": "id",
}
```

I was changing this to follow Java code style guidelines where camel casing is expected and we have linting rules that fail build if they aren't followed. I will revert these changes and add a suppression on this file.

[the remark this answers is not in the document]
brjohnstmsft: No, the JSON is expected to have camelCase property names (at least, [it is in .NET](https://github.com/Azure/azure-sdk-for-net/blob/1d2ee8b78d9c6ebdd5b83c6ae5fae6a8e80138fc/sdk/search/Microsoft.Azure.Search/tests/Tests/Models/LoudHotel.cs#L116) -- looks like the Java tests were setup for PascalCase instead).

sdk/search/azure-search/src/test/java/com/azure/search/LookupAsyncTests.java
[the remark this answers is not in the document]
brjohnstmsft: The service doesn't interpret or try to decode the document key (at least not in this API). It's just that the client needs to construct a valid URL that the service will understand.

sdk/search/azure-search/src/test/java/com/azure/search/LookupAsyncTests.java
alzimmermsft: Removed this class

sdk/search/azure-search/src/test/java/com/azure/search/LookupSyncTests.java
alzimmermsft: Using `indexDoc` now

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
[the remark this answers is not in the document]
alzimmermsft: Updated other JavaDocs

sdk/search/azure-search/src/test/java/com/azure/search/SearchTestBase.java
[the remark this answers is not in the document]
alzimmermsft: Removed most async and base class combinations, I'll hold off on removing the rest for another PR since the few that are left contain more technical details I don't want to handle in this PR.

sdk/search/azure-search/src/test/java/com/azure/search/test/environment/models/LoudHotel.java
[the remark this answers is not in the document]
alzimmermsft: Reverting the changes made to this class and will add a linting suppression to this class

sdk/search/azure-search/src/test/java/com/azure/search/DocumentConverterTests.java
alzimmermsft: I'll update this to make sure it uses `List<String>`.

sdk/search/azure-search/src/main/java/com/azure/search/models/GeoPoint.java
```diff
@@ -116,27 +116,19 @@ public String toString() {
         return "";
     }
 
-    /**
```
sima-zhu: Why do we remove the getCoordinates?

sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyPipelinePolicy.java
```diff
@@ -25,14 +26,13 @@
      * @throws IllegalArgumentException when the api key is an empty string
      */
     public SearchApiKeyPipelinePolicy(SearchApiKeyCredential apiKey) {
-        Objects.requireNonNull(apiKey);
+        Objects.requireNonNull(apiKey, "'apiKey' cannot be null.");
```
sima-zhu: Not necessary to have the message.
Consistent with other class
this.apiKey = Objects.requireNonNull(apiKey);

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
```diff
@@ -140,265 +140,207 @@ HttpPipeline getHttpPipeline() {
...
     }
 
     /**
-     * Gets Client Api Version.
+     * Gets the version of the Search service the client is using.
      *
-     * @return the apiVersion value.
+     * @return The version of the Search service the client is using.
      */
     public SearchServiceVersion getApiVersion() {
```
sima-zhu: I changed the API and document in my PR

sdk/search/azure-search/src/main/java/com/azure/search/models/GeoPoint.java
alzimmermsft: It was an ask to remove them, https://github.com/azure/azure-sdk-for-java/issues/8149

sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyPipelinePolicy.java
alzimmermsft: It's been a general guideline to include a message on `Objects.requireNonNull` calls.

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
alzimmermsft: Mind linking the change?

sdk/search/azure-search/src/main/java/com/azure/search/SearchServiceAsyncClient.java
```diff
@@ -1300,265 +1009,187 @@ public String getEndpoint() {
...
-            .map(Function.identity());
-    }
-
-    /**
-     * Determines whether or not the given skillset exists.
-     *
-     * @param skillsetName the name of the skillset
-     * @return true if the skillset exists; false otherwise.
-     */
-    public Mono<Boolean> skillsetExists(String skillsetName) {
```
sima-zhu: Why do we remove this API?

alzimmermsft: Per request from Bruce, it will be easier to add this back in the future than removing it from our APIs.

PR #8289, opened 2020-02-19
transport client sharing support

moderakh:
Adds support for sharing TransportClient across multiple instances of CosmosClients using a new API in the `CosmosClientBuilder#setConnectionSharingAcrossClientsEnabled(true)`

example:
```java
CosmosClient client1 = new CosmosClientBuilder()
    .setKey(xxx1)
    .setEndpoint(yyy1)
    .setConnectionSharingAcrossClientsEnabled(true)
    .build();

CosmosClient client2 = new CosmosClientBuilder()
    .setKey(xxx2)
    .setEndpoint(yyy2)
    .setConnectionSharingAcrossClientsEnabled(true)
    .build()
```
In scenarios where from the same JVM the end user needs to connect to multiple cosmos db account endpoints and so they have multiple CosmosClient instances per endpoint in the JVM, currently we don't do connection sharing.

This PR provides the foundation for sharing TransportClient. more work will be coming after this PR:
* I will investigate the perf implication in such scenarios
* is it worth to also share connections to GW in these scenarios? (probably yes, if so it will be in upcoming PR)

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
@@ -37,7 +37,7 @@ public class CosmosClientBuilder {
     private TokenResolver tokenResolver;
     private CosmosKeyCredential cosmosKeyCredential;
     private boolean sessionCapturingOverrideEnabled;
-    private boolean connectionSharingAcrossClientsEnabled;
+    private boolean connectionReuseAcrossClientsEnabled;
 
     public CosmosClientBuilder() {
     }
@@ -100,11 +100,11 @@ public class CosmosClientBuilder {
      * Please note, when setting this option, the connection configuration (e.g., socket timeout config, idle timeout config)
      * of the first instantiated client will be used for all other client instances.
      *
-     * @param connectionSharingAcrossClientsEnabled connection sharing
+     * @param connectionReuseAcrossClientsEnabled connection sharing
      * @return current cosmosClientBuilder
      */
-    public CosmosClientBuilder setConnectionSharingAcrossClientsEnabled(boolean connectionSharingAcrossClientsEnabled) {
-        this.connectionSharingAcrossClientsEnabled = true;
+    public CosmosClientBuilder setConnectionReuseAcrossClientsEnabled(boolean connectionReuseAcrossClientsEnabled) {
+        this.connectionReuseAcrossClientsEnabled = true;
         return this;
     }
 
@@ -116,8 +116,8 @@ public class CosmosClientBuilder {
      *
      * @return the connection sharing across multiple clients
      */
-    public boolean isConnectionSharingAcrossClientsEnabled() {
-        return this.connectionSharingAcrossClientsEnabled;
+    public boolean isConnectionReuseAcrossClientsEnabled() {
+        return this.connectionReuseAcrossClientsEnabled;
     }
 
     /**
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
@@ -50,7 +50,7 @@ public class CosmosAsyncClient implements AutoCloseable {
         this.tokenResolver = builder.getTokenResolver();
         this.cosmosKeyCredential = builder.getCosmosKeyCredential();
         this.sessionCapturingOverride = builder.isSessionCapturingOverrideEnabled();
-        this.enableTransportClientSharing = builder.isConnectionSharingAcrossClientsEnabled();
+        this.enableTransportClientSharing = builder.isConnectionReuseAcrossClientsEnabled();
         this.asyncDocumentClient = new AsyncDocumentClient.Builder()
                                        .withServiceEndpoint(this.serviceEndpoint)
                                        .withMasterKeyOrResourceToken(this.keyOrResourceToken)
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/MultipleCosmosClientsWithTransportClientSharingTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/MultipleCosmosClientsWithTransportClientSharingTest.java
@@ -45,8 +45,8 @@ public class MultipleCosmosClientsWithTransportClientSharingTest extends TestSui
         CosmosAsyncContainer asyncContainer = getSharedMultiPartitionCosmosContainer(this.client.asyncClient());
         container1 = client.getDatabase(asyncContainer.getDatabase().getId()).getContainer(asyncContainer.getId());
 
-        client1 = copyCosmosClientBuilder(clientBuilder()).setConnectionSharingAcrossClientsEnabled(true).buildClient();
-        client2 = copyCosmosClientBuilder(clientBuilder()).setConnectionSharingAcrossClientsEnabled(true).buildClient();
+        client1 = copyCosmosClientBuilder(clientBuilder()).setConnectionReuseAcrossClientsEnabled(true).buildClient();
+        client2 = copyCosmosClientBuilder(clientBuilder()).setConnectionReuseAcrossClientsEnabled(true).buildClient();
 
         container1 = client1.getDatabase(asyncContainer.getDatabase().getId()).getContainer(asyncContainer.getId());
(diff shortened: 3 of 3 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -59,8 +62,33 @@ public void enableSessionCapturing(boolean sessionCapturingOverride) {
...
+
+    /**
+     * Enables transport client sharing allows. The default is false.
+     *
+     * When you have multiple instances of Cosmos Client in the same JVM interacting to multiple Cosmos accounts,
+     * enabling this allows connection sharing in Direct mode if possible between instances of Cosmos Client.
+     * @param transportClientSharingEnabled transport client sharing
+     * @return current cosmosClientBuilder
+     */
+    public CosmosClientBuilder setTransportClientSharingEnabled(boolean transportClientSharingEnabled) {
```
kirankumarkolli: >setTransportClientSharingEnabled [](start = 31, length = 32)

TransportClient is an implementation detail. How about name reflecting of connection sharing across clients?

kirankumarkolli: Connection sharing assumed with same configuration. Other alternatives of modeling sharing?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/StoreClientFactory.java
```diff
@@ -13,28 +13,36 @@
...
         this.protocol = configs.getProtocol();
         this.requestTimeoutInSeconds = requestTimeoutInSeconds;
         this.maxConcurrentConnectionOpenRequests = maxConcurrentConnectionOpenRequests;
+        this. enableTransportClientSharing = enableTransportClientSharing;
 
         if (protocol == Protocol.HTTPS) {
             this.transportClient = new HttpTransportClient(configs, requestTimeoutInSeconds, userAgent);
         } else if (protocol == Protocol.TCP){
-            this.transportClient = new RntbdTransportClient(configs, requestTimeoutInSeconds, userAgent);
+            if (enableTransportClientSharing) {
```
kirankumarkolli: Can the sharing be transport agnostic?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/SharedRntbdTransportClient.java
```diff
@@ -0,0 +1,67 @@
...
+        return transportClient.invokeStoreAsync(physicalAddress, request);
+    }
+
+    public int getReferenceCounter() {
+        return counter.get();
+    }
+
+    @Override
+    public void close() throws Exception {
+        synchronized (TransportClient.class) {
```
simplynaveen20: What is the reason of synchronizing on TransportClient and not on SharedRntbdTransportClient, isn't the threads entering this block on the instance of SharedRntbdTransportClient ?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
moderakh: If connection sharing is enabled, the first config will win. assuming if anyone enables this feature doesn't want different config across multiple clients.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/StoreClientFactory.java
moderakh: I thought about that earlier. However port-reuse option (coming later) is only applicable to TCP. 

However it is worth to add it for Direct Https regardless of port-reuse. I will make the change to make it agnostic. sure

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/SharedRntbdTransportClient.java
moderakh: good question, here we are not locking on an instance of class, this lock is a global JVM lock on the class object rather than an instance. So doing it on either on `SharedRntbdTransportClient.class` or `TransportClient.class` both are JVM global lock.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
moderakh: agree. I am changing to `setConnectionSharingAcrossClientsEnabled()`

moderakh: addressed.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/StoreClientFactory.java
moderakh: made it transport agnostic. addressed.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/SharedRntbdTransportClient.java
moderakh: change that to the `SharedTransportClient.class` to be cleaner in the code.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -59,8 +62,58 @@ public void enableSessionCapturing(boolean sessionCapturingOverride) {
...
+     * // when configured this way client1 and client2 will share connections when possible.
+     * }
+     * </pre>
+     *
+     * When you have multiple instances of Cosmos Client in the same JVM interacting to multiple Cosmos accounts,
+     * enabling this allows connection sharing in Direct mode if possible between instances of Cosmos Client.
+     * @param connectionSharingAcrossClientsEnabled connection sharing
+     * @return current cosmosClientBuilder
+     */
+    public CosmosClientBuilder setConnectionSharingAcrossClientsEnabled(boolean connectionSharingAcrossClientsEnabled) {
```
simplynaveen20: In SetConnectionSharingAcrossClientsEnabled  "acrossClients" sounds little verbose , how about setClientsConnectionSharingEnabled()/ isClientsConnectionSharingEnabled().

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
kirankumarkolli: Please document the behaviors very explicitly. 
May be easier to say to use it with homogeneous configuration/client-options (first one times might be also ambiguous to them, with concurrent creations).

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/SharedTransportClient.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public int getReferenceCounter() {
+        return counter.get();
+    }
+
+    @Override
+    public void close() throws Exception {
+        synchronized (SharedTransportClient.class) {
+            final int numberOfActiveTransportClients = counter.decrementAndGet();
+            logger.info("closing one reference to the shared RntbdTransportClient, the number of remaining references is {}", numberOfActiveTransportClients);
```
kirankumarkolli: Is there UT covering ref-counting and re-creation assertion?

moderakh: of course. on the first commit it had UT. please see SharedTransportClientTest

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
simplynaveen20: resolving comment here , offline conversation going on

PR #8290, opened 2020-02-19
adding support for loading test config (account info) from a properties file in the home folder

moderakh:
in addition to old ways of setting test CI config, this PR adds support for setting test config using a properties file in the home folder.

This makes local development and debugging in intelliji easier as you don't have to change environment variable for the test.

you need to have `cosmos-v4.properties` in the project folder or `~/cosmos-v4.properties` file with the following content:

```properties
DESIRED_CONSISTENCIES=["Session"]
PROTOCOLS=["Tcp"]
ACCOUNT_HOST=https://xxx.documents.azure.com:443/
ACCOUNT_KEY=YYY
ACCOUNT_CONSISTENCY=Session
```

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TestConfigurations.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TestConfigurations.java
@@ -8,6 +8,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.InputStream;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TestConfigurations.java
```diff
@@ -14,57 +22,75 @@
...
 
     public static String PROTOCOLS =
-            System.getProperty("PROTOCOLS",
+        properties.getProperty("PROTOCOLS",
                                StringUtils.defaultString(Strings.emptyToNull(
                                        System.getenv().get("PROTOCOLS")),
                                                          null));
+
+    private static Properties loadProperties() {
+        Path path = Paths.get(System.getProperty("user.home"), "cosmos-v4.properties");
```
simplynaveen20: How about having default file cosmos-v4.properties in project directory itself , so dev just need to change that instead of creating new in local

moderakh: good suggestion Naveen.
I added the support for that as well.

moderakh: done

moderakh:
cherry picked in a new PR to master:

PR #8303, opened 2020-02-19
Use Latest PagedFlux Changes from Azure Core

alzimmermsft:
Updates the use of `PagedFlux` and `PagedIterable` to use the new features from Azure Core.

### DIFF
--- a/sdk/search/azure-search/src/main/java/com/azure/search/util/AutocompletePagedResponse.java
+++ b/sdk/search/azure-search/src/main/java/com/azure/search/util/AutocompletePagedResponse.java
@@ -23,7 +23,7 @@ public final class AutocompletePagedResponse extends PagedResponseBase<Void, Aut
      *
      * @param autocompleteResponse Autocomplete response returned from the service.
      */
-    AutocompletePagedResponse(SimpleResponse<AutocompleteResult> autocompleteResponse) {
+    public AutocompletePagedResponse(SimpleResponse<AutocompleteResult> autocompleteResponse) {
         super(autocompleteResponse.getRequest(),
             autocompleteResponse.getStatusCode(),
             autocompleteResponse.getHeaders(),
--- a/sdk/search/azure-search/src/main/java/com/azure/search/util/SearchPagedResponse.java
+++ b/sdk/search/azure-search/src/main/java/com/azure/search/util/SearchPagedResponse.java
@@ -44,7 +44,7 @@ public final class SearchPagedResponse implements ContinuablePage<SearchRequest,
      *
      * @param documentSearchResponse an http response with the results
      */
-    SearchPagedResponse(SimpleResponse<SearchDocumentsResult> documentSearchResponse) {
+    public SearchPagedResponse(SimpleResponse<SearchDocumentsResult> documentSearchResponse) {
         this.statusCode = documentSearchResponse.getStatusCode();
         this.headers = documentSearchResponse.getHeaders();
         this.request = documentSearchResponse.getRequest();
--- a/sdk/search/azure-search/src/main/java/com/azure/search/util/SuggestPagedResponse.java
+++ b/sdk/search/azure-search/src/main/java/com/azure/search/util/SuggestPagedResponse.java
@@ -37,7 +37,7 @@ public final class SuggestPagedResponse extends PagedResponseBase<Void, SuggestR
      *
      * @param documentSearchResponse an http response with the results
      */
-    SuggestPagedResponse(SimpleResponse<SuggestDocumentsResult> documentSearchResponse) {
+    public SuggestPagedResponse(SimpleResponse<SuggestDocumentsResult> documentSearchResponse) {
         super(documentSearchResponse.getRequest(),
             documentSearchResponse.getStatusCode(),
             documentSearchResponse.getHeaders(),


alzimmermsft:
> 
> 
> So this change doesn't impact the usage of these APIs in any way? I recall the base class was used before because it was the only way to access the `SearchPagedResponse`, `SuggestPagedResponse`, etc.

This won't change how these APIs are used, `PagedFlux<T>` extends `PagedFluxBase<T, PagedResponse<T>>` and the `SearchPagedResponse`, etc classes implement `PagedResponse<T>`. This will work as it did before, is a little cleaner on the API level, and fixes the fact that in the latest versions of Azure Core `PagedFluxBase` is now deprecated.

brjohnstmsft:
@alzimmermsft Ok. My main concern is that it's still possible to access top-level properties of the response like facets, coverage, etc.

sdk/search/azure-search/src/main/java/com/azure/search/util/ContinuablePagedIterable.java
```diff
@@ -0,0 +1,151 @@
...
+ * {@link Iterable} interfaces.
+ *
+ * @param <C> the type of the continuation token
+ * @param <T> The type of elements in a {@link ContinuablePage}
+ * @param <P> The {@link ContinuablePage} holding items of type {@code T}.
+ *
+ * @see IterableStream
+ * @see ContinuablePagedFlux
+ */
+public abstract class ContinuablePagedIterable<C, T, P extends ContinuablePage<C, T>> extends IterableStream<T> {
```
alzimmermsft: This class will be removed once this is merged into Azure Core, this was added into Search directly to unblock this change.

sdk/search/azure-search/src/main/java/com/azure/search/util/SearchPagedFlux.java
```diff
@@ -0,0 +1,29 @@
...
+ * element type is {@link SearchResult}, and the page type is {@link SearchPagedResponse}.
+ */
+public final class SearchPagedFlux extends ContinuablePagedFluxCore<SearchRequest, SearchResult, SearchPagedResponse> {
+    /**
+     * Creates an instance of {@link SearchPagedFlux}.
+     *
+     * @param pageRetrieverProvider Supplied that handles retrieving {@link SearchPagedResponse SearchPagedResponses}.
+     */
+    public SearchPagedFlux(
+        Supplier<PageRetriever<SearchRequest, SearchPagedResponse>> pageRetrieverProvider) {
```
sima-zhu: The indentation is not right.

sdk/search/azure-search/src/main/java/com/azure/search/util/AutocompletePagedFlux.java
```diff
@@ -0,0 +1,26 @@
...
+import com.azure.search.models.AutocompleteItem;
+import reactor.core.publisher.Mono;
+
+import java.util.function.Supplier;
+
+/**
+ * Implementation of {@link PagedFluxBase} where the element type is {@link AutocompleteItem} and the page type is
+ * {@link AutocompletePagedResponse}.
+ */
+public final class AutocompletePagedFlux extends PagedFluxBase<AutocompleteItem, AutocompletePagedResponse> {
```
sima-zhu: I did not understand why we still use  PagedFluxBase here. Do we want to deprecate it?

sdk/search/azure-search/src/main/java/com/azure/search/util/ContinuablePagedIterable.java
sima-zhu: Do we have it for preview 1? Can u put it under impl?

alzimmermsft: The changes for this in Azure Core should be merged before preview 1 ships. This can't be under implementation as it is a super class to a public API and we don't want to have any classes in the hierarchy for a public API be non-public API.

sdk/search/azure-search/src/main/java/com/azure/search/util/SearchPagedFlux.java
alzimmermsft: Not sure how the indentation is wrong, I thought it was 4 spaces per level of indentation?

sdk/search/azure-search/src/main/java/com/azure/search/util/AutocompletePagedFlux.java
alzimmermsft: The deprecation of PagedFluxBase is still under discussion, I believe we are leaning towards not deprecating it as one of the most common scenarios for specialized `PagedFlux`s is what is covered by `PagedFluxBase`.

@JonathanGiles @anuchandy @srnagar

sdk/search/azure-search/src/main/java/com/azure/search/AutocompletePagedResponse.java
```diff
@@ -3,29 +3,44 @@
...
-    public AutocompletePagedResponse(SimpleResponse<AutocompleteResult> autocompleteResponse) {
+    AutocompletePagedResponse(SimpleResponse<AutocompleteResult> autocompleteResponse) {
         super(autocompleteResponse.getRequest(),
             autocompleteResponse.getStatusCode(),
             autocompleteResponse.getHeaders(),
             autocompleteResponse.getValue().getResults(),
             null,
             null);
+
+        this.coverage = autocompleteResponse.getValue().getCoverage();
```
srnagar: Can getValue() be null?

sdk/search/azure-search/src/main/java/com/azure/search/SearchPagedResponse.java
```diff
@@ -24,8 +26,14 @@
  * information is: count - number of total documents returned. Will be returned only if isIncludeTotalResultCount is set
  * to true coverage - coverage value.
  */
-public final class SearchPagedResponse extends PagedResponseBase<Void, SearchResult> {
+public final class SearchPagedResponse implements ContinuablePage<SearchRequest, SearchResult>,
```
srnagar: Add `@Immutable` annotation

sdk/search/azure-search/src/main/java/com/azure/search/util/AutocompletePagedFlux.java
srnagar: That's right, since PagedFluxBase is still a common type for some services where continuation token is string and the response is a type of `PagedResponse`, we don't want to eagerly deprecate something that might be useful.

sdk/search/azure-search/src/main/java/com/azure/search/util/SearchPagedIterable.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.azure.search.SearchPagedResponse;
+import com.azure.search.models.SearchRequest;
+import com.azure.search.models.SearchResult;
+
+/**
+ * Implementation of {@link ContinuablePagedIterable} where the continuation token type is {@link SearchRequest}, the
+ * element type is {@link SearchResult}, and the page type is {@link SearchPagedResponse}.
+ */
+public class SearchPagedIterable extends ContinuablePagedIterable<SearchRequest, SearchResult, SearchPagedResponse> {
```
srnagar: `final`?

sdk/core/azure-core/src/main/java/module-info.java
```diff
@@ -36,6 +36,7 @@
     opens com.azure.core.implementation.logging to com.fasterxml.jackson.databind;
     opens com.azure.core.implementation.serializer to com.fasterxml.jackson.databind;
     opens com.azure.core.implementation.serializer.jsonwrapper to com.fasterxml.jackson.databind;
+    exports com.azure.core.util.paging;
```
srnagar: Keep all `exports` together (move this to line 29)

sdk/search/azure-search/pom.xml
```diff
@@ -37,6 +37,16 @@
...
 
+    <!-- Added this dependency to include necessary annotations used by reactor core.
+    Without this dependency, javadoc throws a warning as it cannot find enum When.MAYBE
+    which is used in @Nullable annotation in reactor core classes -->
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>3.0.2</version> <!-- {x-version-update;com.google.code.findbugs:jsr305;external_dependency} -->
+      <scope>provided</scope>
+    </dependency>
```
srnagar: This is already added in azure-core. Do we need to add this here as well?

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
```diff
@@ -479,11 +482,12 @@ public String getEndpoint() {
...
      * @param suggesterName The name of the suggester as specified in the suggesters collection that's part of the index
      * definition
-     * @return A {@link PagedFluxBase} that iterates over {@link SuggestResult} objects and provides access to the
+     * @return A {@link SuggestPagedFlux} that iterates over {@link SuggestResult} objects and provides access to the
      * {@link SuggestPagedResponse} object for each page containing HTTP response and coverage information.
      * @see <a href="https://docs.microsoft.com/rest/api/searchservice/Suggestions">Suggestions</a>
      */
-    public PagedFluxBase<SuggestResult, SuggestPagedResponse> suggest(String searchText, String suggesterName) {
+    public SuggestPagedFlux suggest(String searchText,
+        String suggesterName) {
```
srnagar: nit: this can be in a single line.

brjohnstmsft:
@alzimmermsft Won't have time to review this in detail; Feel free to not block on me.

sdk/search/azure-search/pom.xml
alzimmermsft: `provided` dependencies don't appear to flow transitively to downstream modules, this is needed for `azure-search`.

sdk/search/azure-search/src/main/java/com/azure/search/AutocompletePagedResponse.java
alzimmermsft: Shouldn't, `null` value is likely an error scenario.

srnagar: What does a HTTP 204 resolve to? Would this have a non-null value?

sdk/core/azure-core/src/main/java/module-info.java
alzimmermsft: Changed ordering

alzimmermsft:
> 
> 
> @alzimmermsft Won't have time to review this in detail; Feel free to not block on me.

Thanks for the heads up Bruce! 😄

sdk/search/azure-search/src/main/java/com/azure/search/AutocompletePagedResponse.java
alzimmermsft: The autocomplete API only expects `200` as a valid response code.

PR #8247, opened 2020-02-18
Update dependencies for azure-eventgrid

sanderkoenders:
To resolve some of the CVE issues in this library I updated a couple of dependencies:
- azure-client-runtime 1.5.0 -> 1.7.1
- azure-client-authentication 1.5.0 -> 1.7.1

The new vulnerability report looks like:
```
"Project","ScanDate","DependencyName","DependencyPath","Description","License","Md5","Sha1","Identifiers","CPE","CVE","CWE","Vulnerability","Source","CVSSv2_Severity","CVSSv2_Score","CVSSv2","CVSSv3_BaseSeverity","CVSSv3_BaseScore","CVSSv3","CPE Confidence","Evidence Count"
Microsoft Azure SDK for eventgrid,"Tue, 18 Feb 2020 09:35:09 +0100",guava-20.0.jar,/home/archcry/.m2/repository/com/google/guava/guava/20.0/guava-20.0.jar,"Guava is a suite of core and expanded libraries that include     utility classes, google's collections, io classes, and much     much more.      Guava has only one code dependency - javax.annotation,     per the JSR-305 spec.",http://www.apache.org/licenses/LICENSE-2.0.txt,f32a8a2524620dbecc9f6bf6a20c293f,89507701249388e1ed5ddcf8c41f4ce1be7831ef,pkg:maven/com.google.guava/guava@20.0,cpe:2.3:a:google:guava:20.0:*:*:*:*:*:*:*,CVE-2018-10237,CWE-502 Deserialization of Untrusted Data,"Unbounded memory allocation in Google Guava 11.0 through 24.x before 24.1.1 allows remote attackers to conduct denial of service attacks against servers that depend on this library and deserialize attacker-provided data, because the AtomicDoubleArray class (when serialized with Java serialization) and the CompoundOrdering class (when serialized with GWT serialization) perform eager allocation without appropriate checks on what a client has sent and whether the data size is reasonable.",NVD,MEDIUM,4.3,/AV:N/AC:M/Au:N/C:N/I:N/A:P,MEDIUM,5.9,/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:H,HIGH,24
[...]

PR #8309, opened 2020-02-20
Adding script to generate cgmanifest.json for transitive dependencies

conniey:
Our current tooling for searching for dependencies only examines immediate dependencies. It's possible for dependency `Foo` to rely on a version of `Bar` that requires an update. Our tooling doesn't pick up dependency `Bar`.

* This script generates a `cgmanifest.json`, or appends to an existing one. 
* Sample: `pwsh .\eng\Set-ComponentRegistrations.ps1 D:\git\conniey\azure-sdk-for-java-2\sdk\eventhubs\ -ExcludeRegex "mgmt-*"`
  * Will search under _eventhubs_ for pom.xml and will exclude any directories that start with "mgmt-*".

/cc @JimSuplizio @joshfree

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -191,7 +191,7 @@ jobs:
         displayName: 'Run SpotBugs, Checkstyle, RevApi, and Javadoc'
         inputs:
           mavenPomFile: sdk/${{parameters.ServiceDirectory}}/pom.service.xml
-          options: '--no-transfer-progress -DskipTests -Dgpg.skip -Dverify-readme'
+          options: '--no-transfer-progress -DskipTests -Dgpg.skip -Dverify-readme-windows'
           mavenOptions: '$(MemoryOptions)'
           javaHomeOption: 'JDKVersion'
           jdkVersionOption: '1.11'


sdk/servicebus/microsoft-azure-servicebus/cgmanifest.json
```diff
@@ -0,0 +1,475 @@
+{
```
JimSuplizio: @conniey If we're generating a manifest as part of the build step should this cgmanifest file be checked in?

conniey: Yeah, this should be removed. I'll delete it.

mitchdenny:
Feels like we are doing a lot of heavy lifting here for something that component governance should be doing. Did you speak to anyone on the component governance team about this already?

conniey:
@mitchdenny Hey. Yes, I did. They said it's not a scenario they support and to create a cgmanifest.json for these transitive dependencies.

mitchdenny:
> @mitchdenny Hey. Yes, I did. They said it's not a scenario they support and to create a cgmanifest.json for these transitive dependencies.

Interesting. If it doesn't matter to them, why does it matter to us?

conniey:
I tried to explain it in the summary, but concrete examples are here, where the dependency itself does not contain any issues, but the dependencies it pulls in does. Tools like [DependencyCheck](https://owasp.org/www-project-dependency-check/) do check for these transitive dependencies.
* https://github.com/Azure/azure-sdk-for-java/issues/8440
* https://github.com/Azure/azure-sdk-for-java/pull/8247

PR #8328, opened 2020-02-20
Identity bug fixes

g2vinay:


sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
```diff
@@ -80,7 +82,7 @@
                     .authority(authorityUrl)
                     .setTokenCacheAccessAspect(accessAspect)
                     .build();
-            } catch (MalformedURLException e) {
+            } catch (IOException | PlatformNotSupportedException e) {
```
jianghaolu: Can you catch `Exception` here? We don't want any exception to be directly thrown

g2vinay: done

Issue #6385, opened 2019-11-18
[FEATURE REQ] group common CI job from mgmt sdk to shared template


weidongxu-microsoft:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Description Added
- [x] Expected solution specified

mitchdenny commented 
> I noticed that your ci.mgmt.yml files are all duplicates, this might make maintenance a bit challenging for you in the future, so you might like to consider refactoring into a template, but in terms of the substance of the changes I don't think that this will negatively impact the client SDK builds.

weidongxu-microsoft:
script for automation
https://github.com/weidongxu-microsoft/misc-automation/blob/master/validate_java_mgmt_sdk.py

PR #8358, opened 2020-02-20
Fix up trigger filters for search mgmt pipeline.

mitchdenny:
@weidongxu-microsoft @alzimmermsft just found this bug in the management pipeline path filters for Search. Could you go and double check the rest that were created? 

Also I noticed that all the ci.yml files pretty much just have the same code. I think it would be a good idea to start looking at using a template to reduce all the duplication here. You could setup a template in ```eng/pipelines/templates/stages/archetype-sdk-management.yml``` which contains these steps and use parameters to pass in the values. Similar to how we do for our pipelines. Would probably make maintenance easier.

### DIFF
--- a/sdk/search/ci.mgmt.yml
+++ b/sdk/search/ci.mgmt.yml
@@ -15,7 +15,7 @@ trigger:
     include:
       - sdk/search/
     exclude:
-      - sdk/search/microsoft-azure
+      - sdk/search/azure-
 
 pr:
   branches:
@@ -28,7 +28,7 @@ pr:
     include:
       - sdk/search/
     exclude:
-      - sdk/search/microsoft-azure
+      - sdk/search/azure-
 
 
 variables:


weidongxu-microsoft:
The exclude section is automatically generated (depends on the folder structure of that time, also existed yml is automatically modified to exclude mgmt-) when I batch add all the ci.mgmt.yml (maybe last year?). `microsoft-azure` is the default one which is usually used by client or data.

The `azure-search` folder in search service is added about half a month ago. Without premonition I won't be able to fix that...

I still have the task https://github.com/Azure/azure-sdk-for-java/issues/6385
I will try to find a time to do this. Also I will try to have a script to enable me to periodically verify the correctness of the ci.mgmt.yml and pom.mgmt.xml.

mitchdenny:
> Without premonition I won't be able to fix that...

That's fair enough. Given it was automated you can ignore my request to double check them all ;) We can pick them up by exception as libraries transition from track 1 to track 2.

PR #8370, opened 2020-02-21
use common ci mgmt definition

weidongxu-microsoft:
Fix https://github.com/Azure/azure-sdk-for-java/issues/6385

### DIFF
--- a/sdk/vmwarecloudsimple/ci.mgmt.yml
+++ b/sdk/vmwarecloudsimple/ci.mgmt.yml
@@ -13,7 +13,11 @@ trigger:
       - release/*
   paths:
     include:
-      - sdk/vmwarecloudsimple/mgmt-
+      - sdk/vmwarecloudsimple/
+    exclude:
+      - sdk/vmwarecloudsimple/azure
+      - sdk/vmwarecloudsimple/microsoft-azure
+      - sdk/vmwarecloudsimple/ms-azure
 
 pr:
   branches:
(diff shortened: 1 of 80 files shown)


weidongxu-microsoft:
@mitchdenny 

Let me know whether this is OK for you.

CI fails on client/data probably not related to this PR. I only modifies mgmt CI.

weidongxu-microsoft:
BTW, mgmt CI works pretty good. You can delete the Azure.azure-sdk-for-java-mgmt CI if you wishes.

sdk/vmwarecloudsimple/ci.mgmt.yml
```diff
@@ -13,9 +13,7 @@ trigger:
       - release/*
   paths:
     include:
-      - sdk/vmwarecloudsimple/
-    exclude:
-      - sdk/vmwarecloudsimple/microsoft-azure
+      - sdk/vmwarecloudsimple/mgmt-
```
weshaggard: Are you sure this path prefix works? I don't believe DevOps allows such syntax so I'd suggest you validate it.

mitchdenny: The matching logic for pipelines is effectively a starts-with check. So if your folkders start with ```sdk/vmwareclouresimple/mgmt-``` it should match (and in this case exclude) them.

mitchdenny: ... you just can't/don't need the wildcard symbols like ```*``` etc.

mitchdenny:
This looks good. I've disabled the old management pipeline and moved it into the ```\java\legacy``` folder in Azure DevOps.

sdk/vmwarecloudsimple/ci.mgmt.yml
weidongxu-microsoft: @weshaggard @mitchdenny 

I've tried part of path in my own repo. It indeed does not work.

I set the "include" to "src/ma", not work, "src/main", works.

I will revert it back to "include" + "exclude" setting. Even if "exclude" part not working correctly, it still be better compared to "include" part not working.

PR #8371, opened 2020-02-21
Spliting SentimenLabel class

mssfang:
Spliting `SentimenLabel` to `DocumentSentimentLabel` and `SentenceSentimentLabel` two classes.

### DIFF
--- a/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentSentiment.java
+++ b/sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentSentiment.java
@@ -13,21 +13,21 @@ import java.util.List;
  */
 @Immutable
 public final class DocumentSentiment {
-    private final DocumentSentimentLabel sentimentLabel;
+    private final DocumentSentimentLabel sentiment;
     private final SentimentConfidenceScorePerLabel confidenceScores;
     private final List<SentenceSentiment> sentences;
 
     /**
      * Creates a {@link DocumentSentiment} model that describes the sentiment of the document.
      *
-     * @param sentimentLabel the sentiment label of the document.
+     * @param sentiment the sentiment label of the document.
      * @param confidenceScores the sentiment confidence score (Softmax score) between 0 and 1, for each sentiment label.
      *   Higher values signify higher confidence.
      * @param sentences a list of sentence sentiments.
      */
-    public DocumentSentiment(DocumentSentimentLabel sentimentLabel, SentimentConfidenceScorePerLabel confidenceScores,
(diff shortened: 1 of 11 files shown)


sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentSentiment.java
```diff
@@ -8,43 +8,44 @@
...
 
 /**
- * The DocumentSentiment model
+ * The {@link DocumentSentiment} model that contains sentiment label of a document, confidence score of the sentiment
+ * label, and a list of {@link SentenceSentiment}.
  */
 @Immutable
 public final class DocumentSentiment {
-    private final SentimentLabel sentiment;
+    private final DocumentSentimentLabel sentimentLabel;
```
maririos: Is there a reason why we change the property/attribute name from `sentiment` to `sentimentLabel`?

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/SentenceSentiment.java
```diff
@@ -6,27 +6,28 @@
...
+ * The {@link SentenceSentiment} model that contains a sentiment label of a sentence, confidence score of the sentiment
+ * label, length of the sentence and offset of the sentence within a document.
  */
 @Immutable
 public final class SentenceSentiment {
     private final int length;
     private final int offset;
     private final SentimentConfidenceScorePerLabel confidenceScores;
-    private final SentimentLabel sentiment;
+    private final SentenceSentimentLabel sentimentLabel;
```
maririos: same comment here

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/SentenceSentiment.java
```diff
@@ -51,19 +52,19 @@ public int getOffset() {
     }
 
     /**
-     * Get the text sentiment labels: POSITIVE, NEGATIVE, NEUTRAL.
+     * Get the text sentiment label: POSITIVE, NEGATIVE, or NEUTRAL. Sentence sentiment label could not be MIXED.
```
maririos: NIT: not sure we want to specify that mixed is not an option. It could trigger the question, but why not?
instead of just not mentioning it

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/models/DocumentSentiment.java
mssfang: Not a strong reason why should change to `sentimentLabel`. Will change back to sentiment.

PR #8385, opened 2020-02-21
Update azure e2e build config

g2vinay:


jianghaolu:
We will need to update the live test scripts with this change. But why?

PR #8442, opened 2020-02-24
Issue 8079: Non thread safe call to write blob operation.

carniwar:
Enabling api rest calls using the TracerProxy to run in multi thread a environment. This doesn't consider changes in classloader (like adding a new jar containing a new implementation of Tracer), but at least works for threads running in parallel.

Needed to converted to a singleton because lock mechanisms in lazy load for static fields is not the best practise. The lazy load could be removed, but depending on the order/priority for loading classes, we could introduce another bug.

To avoid any problem, a singleton with lazy load on the list of Tracer providers was introduced.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProxy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProxy.java
@@ -17,10 +17,15 @@ import java.util.ServiceLoader;
  */
 public final class TracerProxy {
 
-    private static final TracerProxy INSTANCE = new TracerProxy();
-
-    private List<Tracer> tracers = null;
-    private final Object lock = new Object();
+    private static final List<Tracer> TRACERS;
+    static {
+        ServiceLoader<Tracer> serviceLoader = ServiceLoader.load(Tracer.class);
+        List<Tracer> tracers = new ArrayList<>();
+        for (Tracer tracer : serviceLoader) {
+            tracers.add(tracer);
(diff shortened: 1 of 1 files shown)


srnagar:
@carniwar Thank you for reporting this issue and submitting the PR. `TracerProxy` indeed will have issues in a multi-threaded environment and needs to be fixed. 

However, we have a well-defined pattern for service loaders as seen here - [HttpClientProviders.java](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/implementation/http/HttpClientProviders.java#L18)

It would be great if you could update your PR to use this pattern instead. 
```java
private static final List<Tracer> TRACERS = new ArrayList<>();
static {
    ServiceLoader<? extends Tracer> serviceLoader = ServiceLoader.load(Tracer.class);
    Iterator<Tracer> iter = serviceLoader.iterator();
    while(iter.hasNext()) {
        TRACERS.add(iter.next());
    }
}
```
Thank you once again and we truly appreciate your contribution!

carniwar:
Hi @srnagar ,
That was my first option, but I was trying to do the most similar to a lazy load as possible.
No problem, I'll update the PR soon.
Thank you for the feedback.

carniwar:
Hi @srnagar ,
I pushed the changes according to the example provided. I tested in our scenario and it's working fine. The only change is that I'm returning an immutable list, since TRACERS should be final.
Let me know if you have further questions.
=)

PR #8466, opened 2020-02-24
Making databaseAccount read dynamic after V4 master merge

### LINKED ISSUE #7668: Use DatabaseAccount for configurations read dynamically

simplynaveen20:
This is same as PR on V4 branch - https://github.com/Azure/azure-sdk-for-java/pull/7831
Created the new PR altogether as it was easier after V4 merge.

Below is the context
Currently GatewayServiceConfigurationReader initialize during client start up with one time snapshot of databaseAccount read via GlobalEndPointManager. Its value remain same throughout the lifespan of application.
This PR will read all the databaseAccount configurations via GatewayServiceConfigurationReader dynamically from cache which updates with its scheduling mechanism or any read/write account error.
This will also fix an issue in which application hits DatabaseAccount read call twice during startup.

fixes #7668


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/GlobalEndpointManager.java
```diff
@@ -159,6 +166,20 @@ public void close() {
         });
     }
 
+    public Mono<DatabaseAccount> getDatabaseAccountFromCache(URI defaultEndpoint) {
+        return this.databaseAccountAsyncCache.getAsync(StringUtils.EMPTY, null, () -> this.owner.getDatabaseAccountFromEndpoint(defaultEndpoint).single().doOnSuccess(databaseAccount -> {
+            if(databaseAccount != null) {
+                this.latestDatabaseAccount = databaseAccount;
+            }
+
+            this.refreshLocationAsync(databaseAccount, false);
```
moderakh: there is a bug here in `refreshLocationAsync(.)` invocation

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -250,29 +250,12 @@ private RxDocumentClientImpl(URI serviceEndpoint,
...
-        this.gatewayConfigurationReader = new GatewayServiceConfigurationReader(this.serviceEndpoint,
-                this.hasAuthKeyResourceToken,
-                resourceToken,
-                this.connectionPolicy,
-                this.authorizationTokenProvider,
-                this.reactorHttpClient);
-
-        DatabaseAccount databaseAccount = this.gatewayConfigurationReader.initializeReaderAsync().block();
+        this.gatewayConfigurationReader = new GatewayServiceConfigurationReader(this.serviceEndpoint, this.globalEndpointManager);
+        DatabaseAccount databaseAccount = this.globalEndpointManager.getLatestDatabaseAccount();
```
moderakh: where are we doing the initialization blocking call now?

If the global endpoint manager is not fully initialized with first databaseAccount fetch could this return null?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/GlobalEndpointManager.java
```diff
@@ -159,6 +166,20 @@ public void close() {
...
+        }));
+    }
+
+    public DatabaseAccount getLatestDatabaseAccount() {
+        return this.latestDatabaseAccount;
```
moderakh: there is a time window where GlobalEndpointManager is instantiated but hasn't fetched DatabaseAccount, within that period how do we ensure this never returns null?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
simplynaveen20: GatewayServiceConfigurationReader constructor is doing blocking call on async cache , so for first time if cache in null it will hit the BE

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/GlobalEndpointManager.java
simplynaveen20: We are calling init on GlobalEndpointManager  from RxDocumentClientImpl which will update this value , any future call if return null due to any exception or error we will not update latestDatabaseAccount and async cahce , they will hold the previous not null value

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/GlobalEndpointManager.java
simplynaveen20: What bug , please explain . this is same as the code initially doing in RxDocumentClientImpl, via this.globalEndpointManager.refreshLocationAsync(databaseAccount, false).block();

simplynaveen20: I think i see the bug will fix in next iteration

simplynaveen20: Made refreshLocationAsync in async pipeline , added test cases to make sure this method is getting called

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/GlobalEndpointManager.java
moderakh: Please document the new API saying what you said above.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
moderakh: could you please add an `assert(databaseAccount != null)` here. for sanity check and readibility.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReader.java
```diff
@@ -42,118 +24,28 @@
...
-        httpHeaders.set(HttpConstants.HttpHeaders.VERSION, HttpConstants.Versions.CURRENT_VERSION);
-
-        UserAgentContainer userAgentContainer = new UserAgentContainer();
-        String userAgentSuffix = this.connectionPolicy.getUserAgentSuffix();
-        if (userAgentSuffix != null && userAgentSuffix.length() > 0) {
-            userAgentContainer.setSuffix(userAgentSuffix);
-        }
-
-        httpHeaders.set(HttpConstants.HttpHeaders.USER_AGENT, userAgentContainer.getUserAgent());
-        httpHeaders.set(HttpConstants.HttpHeaders.API_TYPE, Constants.Properties.SQL_API_TYPE);
```
moderakh: as this `httpHeaders.set(HttpConstants.HttpHeaders.API_TYPE, Constants.Properties.SQL_API_TYPE);` is removed, we are not passing this header to the service in the initialization time. 

* what does .Net do?
* Is that ok not to set SQL_API_TYPE? what happens on non sql type account?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/GlobalEndpointManager.java
```diff
@@ -253,8 +274,27 @@ private void startRefreshLocationTimerAsync() {
...
+        };
+
+        Mono<DatabaseAccount> obsoleteValueMono = databaseAccountAsyncCache.getAsync(StringUtils.EMPTY, null, fetchDatabaseAccount);
+        return obsoleteValueMono.flatMap(obsoleteValue -> {
+            if (firstTimeDatabaseAccountInitialization.compareAndSet(true, false)) {
+                return Mono.just(obsoleteValue);
+            }
+            return databaseAccountAsyncCache.getAsync(StringUtils.EMPTY, obsoleteValue, fetchDatabaseAccount).doOnError(t -> {
+                //Putting back the old value in cache, this will avoid cache corruption
+                databaseAccountAsyncCache.set(StringUtils.EMPTY, obsoleteValue);
```
moderakh: does this mean there is a window of time where there is an invalid result in the cache?

simplynaveen20: Yes

simplynaveen20: Removed cache as per offline discussion

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReader.java
simplynaveen20: Checked with Jake , for .NET code is there to send apiType on all gateway calls , but currently it is taking ApiType.None, which is is filtered out when call goes to gateway.
Now for java we have one place for all gateway calls ,we should include that in common place . We can track this item separately

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/GlobalEndpointManager.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/GatewayServiceConfigurationReader.java
moderakh: thanks

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
simplynaveen20: done

Issue #8183, opened 2020-02-13
[BUG] OkHttpAsyncHttpClient throws IOException downloading blob from Azure container


TimOMalley:
**Describe the bug**
OkHttpAsyncHttpClient throws an IOException: closed when it's reading the InputStream during the BlobClient downloadToFile.
***Exception or Stack Trace***
`2020-02-11 15:54:00,928 | INFO  | -pubblobsdk_Worker-2 | o.q.c.JobRunShell                | 715 - org.quartz-scheduler.quartz - 2.3.0 | Job myGroup.us_gov_dod_af_cce_mm_pubblobsdk_route2 threw a JobExecutionException: 
org.quartz.JobExecutionException: reactor.core.Exceptions$ReactiveException: java.io.IOException: closed
	at org.apache.camel.component.quartz2.CamelJob.execute(CamelJob.java:61) ~[!/:2.21.0.fuse-750033-redhat-00001]
	at org.quartz.core.JobRunShell.run(JobRunShell.java:202) [!/:?]
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573) [!/:?]
Caused by: reactor.core.Exceptions$ReactiveException: java.io.IOException: closed
	at reactor.core.Exceptions.propagate(Exceptions.java:336) ~[!/:3.3.0.RELEASE]
	at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:91) ~[!/:3.3.0.RELEASE]
	at reactor.core.publisher.Mono.block(Mono.java:1663) ~[!/:3.3.0.RELEASE]
	at com.azure.storage.common.implementation.StorageImplUtils.blockWithOptionalTimeout(StorageImplUtils.java:94) ~[!/:?]
	at com.azure.storage.blob.specialized.BlobClientBase.downloadToFileWithResponse(BlobClientBase.java:481) ~[!/:?]
	at com.azure.storage.blob.specialized.BlobClientBase.downloadToFile(BlobClientBase.java:442) ~[!/:?]
	at mil.af.cce2.mm.templates.azure.lib.AzureLibBean.downloadBlobFromContainer(AzureLibBean.java:480) ~[!/:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_212]
[...]

PR #8476, opened 2020-02-25
Fixing OkHttp Async Client early stream closing

anuchandy:
Addressing https://github.com/Azure/azure-sdk-for-java/issues/8183

The stream get closed by `Mono.using()` as soon as `ResponseBody` is emitted.
 
The `ResponseBody` instance hold the content stream, we want to wait for the entire emission of content stream to complete to close the stream.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
+++ b/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
@@ -1902,9 +1902,9 @@
 
   <!-- Empty byte array is returned as null to reactor. Reactor transform null from Callable to downstream onComplete  -->
   <Match>
-    <Class name="com.azure.core.http.okhttp.OkHttpAsyncHttpClient$OkHttpResponse"/>
-    <Method name="getBodyAsByteArray"/>
+    <Class name="~com\.azure\.core\.http\.okhttp\.OkHttpAsyncHttpClient\$OkHttpResponse(.*)"/>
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
```diff
@@ -178,29 +178,21 @@ public void onResponse(okhttp3.Call call, okhttp3.Response response) {
...
-                    // square.github.io/okhttp/4.x/okhttp/okhttp3/-response-body/#the-response-body-must-be-closed
-                    ResponseBody::close, /* Change in behavior since reactor-core 3.3.0.RELEASE */ false);
-            }
+            // innerResponse.body() getter will not return null for server returned responses.
+            // It can be null:
+            // [a]. if response is built manually with null body (e.g for mocking)
+            // [b]. for the cases described here
+            // [ref](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response/body/).
+            //
+            this.responseBody = innerResponse.body();
```
srnagar: Do we still need the null check for the above mocking (a) and cacheResponse, networkResponse etc (b)?

anuchandy: we do that in each of the places `this.responseBody` is used.

PR #8485, opened 2020-02-26
Remove JavadocInlineTagCheck Checkstyle Rule

alzimmermsft:
This PR removes the `JavadocInlineTagCheck` Checkstyle rule as it was causing more noise than catching. This rule was originally written while we were transitioning our Javadoc code snippets from using `<pre>`, `<code>`, `{@code}` to using the codesnippet tool we now use. This check generally causes conflicts with auto formatting which will split `{@code}` tags into multiple lines leading to scenarios where Checkstyle's will fail.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml
+++ b/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml
@@ -362,7 +362,10 @@ page at http://checkstyle.sourceforge.net/config.html -->
     1) Use {@codesnippet ...} instead of '<code>', '<pre>', or '{@code ...}' if these tags span multiple lines.
        Inline code sample are fine as-is.
     2) No check on class-level Javadoc. -->
-    <module name="com.azure.tools.checkstyle.checks.JavadocInlineTagCheck"/>
+    <!-- This check was removed as it causes more noise than fixes. The purpose of this check was to move away from -->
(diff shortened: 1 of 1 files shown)


eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml
```diff
@@ -362,7 +362,10 @@ page at http://checkstyle.sourceforge.net/config.html -->
     1) Use {@codesnippet ...} instead of '<code>', '<pre>', or '{@code ...}' if these tags span multiple lines.
        Inline code sample are fine as-is.
     2) No check on class-level Javadoc. -->
-    <module name="com.azure.tools.checkstyle.checks.JavadocInlineTagCheck"/>
+    <!-- This check was removed as it causes more noise than fixes. The purpose of this check was to move away from -->
+    <!-- using <code>, <pre>, and {@code} tags while we were on-boarding codesnippets. These tags have since been   -->
+    <!-- removed and codesnippets are expected.                                                                     -->
+    <!-- <module name="com.azure.tools.checkstyle.checks.JavadocInlineTagCheck"/> -->
```
srnagar: Can be deleted instead of commenting this out.

alzimmermsft: I felt leaving it with an explanation would be better if we ever wanted to re-purpose it in the future.

PR #8488, opened 2020-02-26
Cosmos DB: generate 2019 12 12

ChenTanyi:


ChenTanyi:
The track 1 cosmos management sdk named `com.microsoft.azure.management.cosmosdb` from beginning, we follow the naming and not change yet. (I 'm not certain if it would change in track 2)

There is no change for data-plane, you can check it by yourself.

PR #8492, opened 2020-02-26
Enabled Bounded Staleness and Consistent Prefix Fast tests

kushagraThapar:
* Enabled Bounded Staleness and Consistent Prefix Fast tests on Tcp and Https connection mode


kushagraThapar:
> thank you. We should add more tests later. Good to start.
> 
> For bounded-staleness the account consistency can be strong (as strong >= bounded-staleness), however in the long run we need tests where account default consistency is bounded staleness.
> 
> LGTM

Thanks Mo, 
Regarding Strong >= Bounded-staleness, this was the quickest way I could test this. 
For sure, we will add more tests in upcoming days

PR #8441, opened 2020-02-24
Upgrade `azure-client-authentication` package to 1.7.2

raghuvaran:
Closes #8440

Issue #8270, opened 2020-02-18
Standardize external dependency versions


JimSuplizio:
The [eng/versioning/external_dependencies.txt](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/external_dependencies.txt) file currently has a number of [external dependency exceptions](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/external_dependencies.txt#L152). The vast majority of these are for Data track libraries but that doesn't change the fact that there some of these old libraries and dependencies have CVE issues that need to get fixed. For example: https://github.com/Azure/azure-sdk-for-java/pull/8247

PR #8497, opened 2020-02-26
Consolidates various versions of autorest runtime for java in data libraries

conniey:
* Updates libraries to 1.7.2 release of autorest-runtime-for-java
* Removes and fixes compile errors when removing different versions of the dependency.
  * The test-jar dependencies was because it used:
    * InterceptorManager (okhttp3)
    * TestBase
    * TestNamer*
    * I copied these classes rather than creating another common library. (Not sure if it is worth the investment if they are being maintained and not in active dev.)
* Removes cosmos library from pom.data.xml
   * It's no longer a data library. Only the client one exists.
* Moves applicationinsights record files into proper location.

Related to #8270 
Supercedes #8441

### DIFF
--- a/eng/versioning/external_dependencies.txt
+++ b/eng/versioning/external_dependencies.txt
@@ -205,8 +205,5 @@ media_io.jsonwebtoken:jjwt;0.5.1
 media_org.mockito:mockito-all;1.9.0
 media_com.microsoft.azure:adal4j;1.2.0
 
-# sdk\servicebus\microsoft-azure-servicebus\pom.xml
-servicebus_com.microsoft.azure:azure-client-authentication;1.7.0
-
 # sdk\storage\azure-storage-blob-cryptography\pom.xml
 storage_com.microsoft.azure:azure-storage;8.4.0
--- a/sdk/servicebus/microsoft-azure-servicebus/pom.xml
+++ b/sdk/servicebus/microsoft-azure-servicebus/pom.xml
@@ -78,7 +78,7 @@
     <dependency>
       <groupId>com.microsoft.azure</groupId>
       <artifactId>azure-client-authentication</artifactId>
-      <version>1.7.0</version> <!-- {x-version-update;servicebus_com.microsoft.azure:azure-client-authentication;external_dependency} -->
+      <version>1.7.2</version> <!-- {x-version-update;com.microsoft.azure:azure-client-authentication;external_dependency} -->
       <scope>compile</scope>
     </dependency>
     <dependency>


sdk/eventhubs/microsoft-azure-eventhubs/pom.xml
```diff
@@ -52,7 +52,7 @@
     <dependency>
       <groupId>com.microsoft.azure</groupId>
       <artifactId>azure-client-authentication</artifactId>
-      <version>1.6.15</version> <!-- {x-version-update;com.microsoft.azure:azure-client-authentication;external_dependency} -->
+      <version>1.7.2</version> <!-- {x-version-update;com.microsoft.azure:azure-client-authentication;external_dependency} -->
```
alzimmermsft: Should this have the relocation section added since there is a Track 2 client for EventHubs?

sdk/keyvault/microsoft-azure-keyvault/pom.xml
```diff
@@ -56,7 +56,7 @@ the MIT License. See License.txt in the project root for license information. --
     <dependency>
       <groupId>com.microsoft.azure</groupId>
       <artifactId>azure-client-runtime</artifactId>
-      <version>1.6.15</version> <!-- {x-version-update;com.microsoft.azure:azure-client-runtime;external_dependency} -->
+      <version>1.7.2</version> <!-- {x-version-update;com.microsoft.azure:azure-client-runtime;external_dependency} -->
```
alzimmermsft: Should this have the relocation section added since there is a Track 2 client for LeyVault?

conniey: I didn't want to mix two different issue fixes. I can create another PR that addresses all the relocations.

conniey: Does LeyVault fall along some ley lines?! 😆

g2vinay: I am working on a PR for relocation section in Track1 KV POMs. Should be out today.

sdk/eventhubs/microsoft-azure-eventhubs/pom.xml
JonathanGiles: That is a good point Alan - ideally yes, we should start relocating as soon as we believe we have a replacement for the existing track one library.

sdk/keyvault/microsoft-azure-keyvault/pom.xml
JonathanGiles: I prefer a separate PR for the relocations too, thanks!

PR #8500, opened 2020-02-26
Nio copy

rickle-msft:


sdk/storage/azure-storage-common/pom.xml
```diff
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
```
gapra-msft: revert

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -325,13 +369,184 @@ public void delete(Path path) throws IOException {
...
+        There is an optimization here where we try to do the copy first and only check for a virtual directory if
+        there's a 404. In the cases of files and concrete directories, this only requires one request. For virtual
+        directories, however, this requires three requests: failed copy, check status, create directory. Depending on
+        customer scenarios and how many virtual directories they copy, it could be better to check the directory status
+        first and then do a copy or createDir, which would always be two requests for all resource types.
+         */
+        try {
+            SyncPoller<BlobCopyInfo, Void> pollResponse =
+                destinationBlob.beginCopy(sourceBlob.getBlobUrl(), null, null, null, null, requestConditions, null);
+            pollResponse.waitForCompletion(Duration.ofSeconds(30));
```
gapra-msft: Is there a reason for the duration being 30 here?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -325,13 +369,184 @@ public void delete(Path path) throws IOException {
...
+        }
+        if (!optionsList.isEmpty()) {
+            throw Utility.logError(logger, new UnsupportedOperationException("Unsupported copy option found. Only "
+                + "StandardCopyOption.COPY_ATTRIBUTES and StandareCopyOption.REPLACE_EXISTING are supported."));
+        }
 
+        // Validate paths.
+        // Copying a root directory or attempting to create/overwrite a root directory is illegal.
+        if (source.equals(source.getRoot()) || destination.equals(destination.getRoot())) {
+            throw Utility.logError(logger, new IllegalArgumentException(String.format("Neither source nor destination "
```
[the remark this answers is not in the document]
rickle-msft: I actually just refactored this to do exactly that :). It'll be in my next PR.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
rickle-msft: Nope. Just a magic number plus being a little lazy about adding a constant/option. I figure we can add an option for it if/when people are dissatisfied with this. Or I can add an option now. :P

sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/AzureFileSystemProviderSpec.groovy
```diff
@@ -303,15 +306,509 @@ class AzureFileSystemProviderSpec extends APISpec {
...
+        setup:
+        def fs = createFS(config)
+
+        // Generate resource names.
+        // Don't use default directory to ensure we honor the root.
+        def rootName = fs.getRootDirectories().last().toString()
+        def sourceName = generateBlobName()
+        def sourcePath = fs.getPath(rootName, sourceName)
+        def destName = generateBlobName()
+        def destPath = fs.getPath(rootName, destName)
```
gapra-msft: why dont we add this repeated to the setup and create global variables?

gapra-msft: I guess its not really repeated in every test

rickle-msft: I didn't quite follow this suggestion because I think the tests in this spec will vary too much to make them global, but I did a lot of refactoring of tests and cut out a lot of lines. That'll show up in my next PR.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
gapra-msft: might at least be helpful to make it a constant so we dont have to dig around for them later. ya know?

sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/AzureFileSystemProviderSpec.groovy
```diff
@@ -303,15 +306,509 @@ class AzureFileSystemProviderSpec extends APISpec {
...
+
+        // Generate clients to resources.
+        def containerClient = primaryBlobServiceClient
+            .getBlobContainerClient(containerName)
+        def sourceClient = containerClient.getBlobClient(sourceName)
+        def destinationClient = containerClient.getBlobClient(destName)
+        def sourceChildClient = null
+        def destChildClient = null
```
jaschrep-msft: Can you throw this in the file's setup method? It looks mostly the same throughout.

jaschrep-msft: Or if they differ enough for this to not be an option, it looks like you always use these variables. Have them defined on the class and have a "default populate" method that you can ignore for the times where you really do need different values. That will also help identify when a test is doing something different where that setup matters.

rickle-msft: This will come in my PR for delete, but I did add this.

PR #8393, opened 2020-02-21
Make changes on public APIs direct under 'search'

sima-zhu:
First round of API changes.
Only make changes on APIs direct under com.azure.search.

PR #8503, opened 2020-02-26
Standardize Search Client Builders

alzimmermsft:
This PR standardizes the Search client builders and adds additional APIs commonly found in other builders. Additionally, this PR cleans up the Javadocs for the builders, making them more comprehensive and adding insight into what certain APIs entail.

### DIFF
--- a/sdk/search/azure-search/src/test/java/com/azure/search/SearchIndexClientBuilderTests.java
+++ b/sdk/search/azure-search/src/test/java/com/azure/search/SearchIndexClientBuilderTests.java
@@ -7,8 +7,7 @@ import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 public class SearchIndexClientBuilderTests {
     private final SearchApiKeyCredential searchApiKeyCredential = new SearchApiKeyCredential("0123");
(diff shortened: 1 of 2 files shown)


sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
```diff
@@ -95,17 +95,6 @@
         HttpPipeline httpPipeline) {
 
         SearchServiceUrlParts parts = SearchServiceUrlParser.parseServiceUrlParts(endpoint);
-
```
sima-zhu: Are these checked anywhere?

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexClientBuilder.java
```diff
@@ -34,81 +33,131 @@
...
-    private String clientVersion;
+    private final ClientLogger logger = new ClientLogger(SearchIndexClientBuilder.class);
+    private final List<HttpPipelinePolicy> policies = new ArrayList<>();
+    private final HttpHeaders headers = new HttpHeaders().put(ECHO_REQUEST_ID_HEADER, "true");
 
+    private final String clientName;
+    private final String clientVersion;
+
+    private SearchApiKeyCredential searchApiKeyCredential;
+    private SearchServiceVersion apiVersion;
```
sima-zhu: I have renamed this field. I am going to check in https://github.com/Azure/azure-sdk-for-java/pull/8393. Please merge my PR.

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexClientBuilder.java
```diff
@@ -34,81 +33,131 @@
...
     public SearchIndexClientBuilder() {
-        apiVersion = SearchServiceVersion.getLatest();
-        policies = new ArrayList<>();
-        httpClient = HttpClient.createDefault();
-        httpLogOptions = new HttpLogOptions();
-
         Map<String, String> properties = CoreUtils.getProperties(SEARCH_PROPERTIES);
         clientName = properties.getOrDefault(NAME, "UnknownName");
         clientVersion = properties.getOrDefault(VERSION, "UnknownVersion");
-        headers = new HttpHeaders()
```
sima-zhu: Why removed the default value?

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexClientBuilder.java
```diff
@@ -121,140 +170,139 @@ public SearchIndexClientBuilder endpoint(String endpoint) {
     }
 
     /**
-     * Sets the index name
+     * Sets the {@link SearchApiKeyCredential} used to authenticate HTTP requests.
      *
-     * @param indexName name of the index
-     * @return the updated SearchIndexClientBuilder object
+     * @param searchApiKeyCredential SearchApiKeyCredential used to authenticate HTTP requests.
```
sima-zhu: Add @link for SearchApiKeyCredential . The SearchApiKeyCredential `is` used to authenticate HTTP requests.

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
sima-zhu: Never mind. Found in API

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexClientBuilder.java
alzimmermsft: This was inline instantiated at the property, no need to include this in the constructor if it will be created for every client.

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexClientBuilder.java
alzimmermsft: Will do

sdk/search/azure-search/src/test/java/com/azure/search/SearchIndexClientBuilderTests.java
```diff
@@ -131,123 +130,41 @@ public void whenBuildClientAndVerifyPropertiesThenSuccess() {
...
 
     @Test
-    public void verifyEmptyEndpointIsInvalidAsyncTest() {
-        expectThrowsWithMessage("'endpoint' must be a valid URL", () -> new SearchIndexClientBuilder()
-            .endpoint("")
-            .credential(searchApiKeyCredential)
-            .searchServiceVersion(apiVersion)
-            .indexName(indexName)
-            .buildAsyncClient());
+    public void emptyEndpointThrowsIllegalArgumentException() {
```
sima-zhu: To my point of view, error message change is also a breaking change. To verify the message right here, make sure developer pay attention when they make a change to the error message.
I prefer to assert the message as well.

alzimmermsft: Exception messages are treated on a per case basis on whether they are considered a breaking change, with the general decision being would this message be programmatically consumed. Given that this is a runtime exception that isn't networking related I believe this would be unlikely so there isn't a strong conviction on testing the message of the exception.

PR #8579, opened 2020-02-28
fix: some location LRO return 201 rather than 200

ChenTanyi:


ChenTanyi:
@yaohaizh CI has passed.

yaohaizh:
@anuchandy please help review this one

PR #8604, opened 2020-02-28
moving away from intermediate String in more places and using byte[] instead

moderakh:
moving away from intermediate String in more places and using byte[] instead.

this is a follow up PR in the https://github.com/Azure/azure-sdk-for-java/pull/8486 series


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/Utils.java
```diff
@@ -65,6 +65,18 @@
         Utils.simpleObjectMapper.registerModule(new AfterburnerModule());
     }
 
+    public static byte[] getUTF8BytesOrNull(String str) {
+        if (str == null) {
+            return null;
+        }
+
+        try {
+            return str.getBytes("UTF-8");
```
kushagraThapar: Instead of using String "UTF-8", we should use the import -> StandardCharsets.UTF_8
don't need to change on this PR as CI have passed, but please change it in next iteration in another PR.

moderakh: you are right. will fix. missed this one.

moderakh: will fix in the next PR. thank you

PR #8605, opened 2020-02-28
Avoid extra polling.

yaohaizh:
@anuchandy 

Currently, the LRO will always do extra polling then exam the last poll state, which take extra internal time wasted. Given this,

This PR makes some changes: 
1. Send an initialize response in any case. 
2. Do polling and exam the result. 

Please provide your oponion here. Or do we need extra LRO status?


anuchandy:
@yaohaizh we have a case where we need to do a final get to receive the fully populated resource. This is mainly for provisioning-State based polling.

yaohaizh:
I'm not sure exactly what the  provisioning-State based polling meaning? @anuchandy any suggestion here to fix the extra polling issue? The extra polling makes every LRO quite slow.

yaohaizh:
@yungezz

yungezz:
hello and thanks. :)

PR #8608, opened 2020-02-28
Allow unreleased dependencies if the scope is test

JimSuplizio:
The scan_for_unreleased_dependencies.ps1 script was a bit too restrictive and would fail for any unreleased dependency even if that dependency was scoped to test. The code change is pretty straightforward, when an unreleased dependency is discovered and its scope is test, don't fail the check.


weshaggard:
What cases do we want to allow the test scope to release with an unreleased dependency? If we do that will we ever block anyone from using our package at runtime? For example if someone pulls our project in to there test project does that cause our test scope dependencies to need to resolve?

PR #8631, opened 2020-03-02
A policy that timeout the response content that is blocked and retries

anuchandy:


anuchandy:
This was a prototype while investigating a customer issue, not required anymore given we have support for a more granular level timeout in netty level.

PR #8632, opened 2020-03-02
HttpClient module for azure-core-http based on the JDK 11+ HttpClient

JonathanGiles:
This PR introduces a new http client module for use by users of the Azure SDK for Java libraries. They may opt into this in a fashion [similar to what they can do today to use OkHttp](https://github.com/Azure/azure-sdk-for-java/wiki/HTTP-clients#replacing-the-default-http-client).

Whilst the HttpClient API was first introduced into JDK 9 as a preview feature, and made GA in JDK 11, there is a limiting feature in JDK 11 which means that this module will only be usable in JDK 12 and later.

We are actively benchmarking this module, and will soon also integrate it into our parameterised unit tests (as a follow-up PR after this PR is merged).

This PR is not yet complete - review is welcome but there are still tasks to be done before merging will be considered, and a comment will be left inviting PR review at that point prior to merging.

### DIFF
--- a/sdk/core/azure-core-http-jdk-httpclient/src/samples/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilderJavaDocCodeSnippets.java
+++ b/sdk/core/azure-core-http-jdk-httpclient/src/samples/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilderJavaDocCodeSnippets.java
@@ -57,11 +57,11 @@ public class JdkAsyncHttpClientBuilderJavaDocCodeSnippets {
 
     public void connectionTimeoutSample() {
 
-        // BEGIN: com.azure.core.http.jdk.httpclient.JdkAsyncHttpClientBuilder#connectionTimeout
+        // BEGIN: com.azure.core.http.jdk.httpclient.JdkAsyncHttpClientBuilder.connectionTimeout#Duration
         HttpClient client = new JdkAsyncHttpClientBuilder()
                 .connectionTimeout(Duration.ofSeconds(250)) // connection timeout of 250 seconds
                 .build();
-        // END: com.azure.core.http.jdk.httpclient.JdkAsyncHttpClientBuilder#connectionTimeout
+        // END: com.azure.core.http.jdk.httpclient.JdkAsyncHttpClientBuilder.connectionTimeout#Duration
 
     }
 }


anuchandy:
@JonathanGiles I've updated the PR to switch to `fromPublisher` for response body.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk11/httpclient/Jdk11AsyncHttpClient.java
```diff
@@ -6,57 +6,59 @@
...
 class Jdk11AsyncHttpClient implements HttpClient {
     private final java.net.http.HttpClient jdk11HttpClient;
 
-    private static final Mono<BodyPublisher> EMPTY_BODY_PUBLISHER_MONO = Mono.just(noBody());
+    private static final Set<String> JDK11_RESTRICTED_HEADERS;
+    static {
+        TreeSet<String> treeSet = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        treeSet.addAll(Set.of("connection",
+            "content-length",
+            "date", // BUG: Java11: https://bugs.openjdk.java.net/browse/JDK-8213189 (Fixed in JDK12)
```
anuchandy: @JonathanGiles I think this bug in JDK11 HttpClient may be a blocker for this HttpClient adoption. "Date" is a necessary header for many of the track2 libs, something we need to look into.

JonathanGiles: In that case we would only support this on JDK 12 and later, which is unfortunate.

JonathanGiles:
Converting to a real PR to start PR validation steps.

sdk/core/azure-core-http-jdk11-httpclient/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.azure</groupId>
+    <artifactId>azure-client-sdk-parent</artifactId>
+    <version>1.7.0</version> <!-- {x-version-update;com.azure:azure-client-sdk-parent;current} -->
+    <relativePath>../../../pom.client.xml</relativePath>
+  </parent>
+
+  <groupId>com.azure</groupId>
+  <artifactId>azure-core-http-jdk-httpclient</artifactId>
```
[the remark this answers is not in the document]
JonathanGiles: I initially named it jdk11 but moved away from that - I now want it all to just be jdk

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import java.net.http.HttpRequest.BodyPublisher;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.concurrent.Flow;
+
+import static java.net.http.HttpResponse.BodyHandlers.*;
+import static java.net.http.HttpRequest.BodyPublishers.*;
```
alzimmermsft: `*` imports.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
```diff
@@ -0,0 +1,197 @@
...
+ */
+class JdkAsyncHttpClient implements HttpClient {
+    private final java.net.http.HttpClient jdk11HttpClient;
+
+    private static final Set<String> JDK11_RESTRICTED_HEADERS;
+    static {
+        TreeSet<String> treeSet = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        treeSet.addAll(Set.of("connection",
+            "content-length",
+            "date", // BUG: Java11: https://bugs.openjdk.java.net/browse/JDK-8213189 (Fixed in JDK12)
```
alzimmermsft: How will this work for libraries which require fine tuning for these headers? In the case of Storage the upload could fail on a mismatch `Content-Length` header and shared key credential authentication will fail if the `Date` header is mismatch compared to what was used to generate the key.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+        if (buildProxyOptions != null) {
+            httpClientBuilder = httpClientBuilder.proxy(new JdkHttpClientProxySelector(
+                buildProxyOptions.getType().toProxyType(),
+                buildProxyOptions.getAddress(),
+                buildProxyOptions.getNonProxyHosts()));
+
+            if (buildProxyOptions.getUsername() != null) {
+                httpClientBuilder.authenticator(new Authenticator() {
+                    @Override
+                    protected PasswordAuthentication getPasswordAuthentication() {
```
alzimmermsft: Would this handle digest authentication?

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Creates JdkAsyncHttpClientBuilder.
+     */
+    public JdkAsyncHttpClientBuilder() {
+    }
+
+    /**
+     * Creates JdkAsyncHttpClientBuilder from the builder of an existing {@link java.net.http.HttpClient.Builder}.
+     *
+     * @param httpClientBuilder the HttpClient builder to use
```
alzimmermsft: Add a `@throws`

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * Sets the executor to be used for asynchronous and dependent tasks. This cannot be null.
+     *
+     * <p> If this method is not invoked prior to {@linkplain #build() building}, a default executor is created for each
+     * newly built {@code HttpClient}.
+     *
+     * @param executor the executor to be used for asynchronous and dependent tasks
+     * @return the updated Jdk11AsyncHttpClientBuilder object
```
alzimmermsft: Add a `@throws`

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+     */
+    public JdkAsyncHttpClientBuilder executor(Executor executor) {
+        this.executor = Objects.requireNonNull(executor, "executor can not be null");
+        return this;
+    }
+
+    /**
+     * Sets the connection timeout.
+     *
```
alzimmermsft: `<p>`

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/implementation/JdkHttpClientProvider.java
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.http.jdk.httpclient.implementation;
+
+import com.azure.core.http.HttpClient;
+import com.azure.core.http.HttpClientProvider;
+import com.azure.core.http.jdk.httpclient.JdkAsyncHttpClientBuilder;
+
+public class JdkHttpClientProvider implements HttpClientProvider {
```
alzimmermsft: Missing class level Javadoc

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/implementation/JdkHttpClientProxySelector.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * This class handles selecting the proxy during a request.
+ */
+public final class JdkHttpClientProxySelector extends ProxySelector {
+    private final Proxy.Type proxyType;
+    private final SocketAddress proxyAddress;
+    private final Pattern nonProxyHostsPattern;
+
+    public JdkHttpClientProxySelector(Proxy.Type proxyType, SocketAddress proxyAddress, String nonProxyHosts) {
```
alzimmermsft: Constructor Javadoc?

sdk/core/azure-core-http-jdk11-httpclient/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+      <id>microsoft</id>
+      <name>Microsoft</name>
+    </developer>
+  </developers>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core</artifactId>
+      <version>1.3.0-beta.2</version> <!-- {x-version-update;com.azure:azure-core;current} -->
```
alzimmermsft: Needs a version change

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * HttpClient implementation for the JDK 11 HttpClient.
+ */
+class JdkAsyncHttpClient implements HttpClient {
+    private final java.net.http.HttpClient jdk11HttpClient;
+
+    private static final Set<String> JDK11_RESTRICTED_HEADERS;
+    static {
+        TreeSet<String> treeSet = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
```
alzimmermsft: This should be more flexible in its configuration, if someone is using Java 12+ and set the system property to allow restricted headers we should allow these headers.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilder.java
anuchandy: good question :) it won't. Is our current digest auth code just pluggable? or it's tailored per HTTP client? I haven't looked into that yet.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
anuchandy: I added this initially to avoid JDK11 HttpClient native builder from throwing an exception. Thinking what would be the best approach here:

Maybe we can copy over the same jdk11+ logic around the header whitelisting here or if the exception thrown by the JDK11+ HttpClient native builder has a special error code for disallowed headers then we can catch, log and continue.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
anuchandy: If JDK httpClient supports completely disabling its header checks programmatically then that would be the easiest path. So far we didn't see that Netty/OkHttp blocking any headers.

If full disabling is not possible then we need to see if the native HTTP client has a programmatic way to white list subset of headers, if yes we can identify the headers that JDK blocks by default but required by our SDKs to function properly and white list them.

Yet to explore.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
```diff
@@ -0,0 +1,253 @@
...
+            }
+        }
+    }
+
+    /**
+     * Get the java runtime major version.
+     *
+     * @return the java major version
+     */
+    private int getJavaVersion() {
```
alzimmermsft, line 146: This could be made static right?

anuchandy: could be, it uses instance-level logger, so static means we need to pass it.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
anuchandy: Relaxed validations and added necessary log messages as agreed offline.

sdk/core/azure-core-http-jdk11-httpclient/pom.xml
anuchandy: updated to the lastest azure-core.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/implementation/JdkHttpClientProxySelector.java
anuchandy: It's an implementation package type but added minimal Javadoc for this.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilder.java
anuchandy: digest-auth is on hold for now, like we discussed this will be picked up once we hear from native HTTP client owner.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
anuchandy: Relaxed validations and added necessary log messages as agreed offline. 

Regarding the headers we care about:

`Date` header is allowed from JDK12+
`Content-Length` header will be set by JDK native HttpClient before sending out the requests.

sdk/core/azure-core-http-jdk11-httpclient/README.md
```diff
@@ -0,0 +1,68 @@
...
+- Java Development Kit (JDK) with version 11 or above
+
+### Adding the package to your product
+
+[//]: # ({x-version-update-start;com.azure:azure-core-http-jdk-httpclient;current})
+```xml
+<dependency>
+    <groupId>com.azure</groupId>
+    <artifactId>azure-core-http-jdk-httpclient</artifactId>
+    <version>1.1.0</version>
```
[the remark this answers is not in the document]
anuchandy: updated to `1.0.0-beta.1`

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
```diff
@@ -0,0 +1,197 @@
...
+            try {
+                builder.uri(request.getUrl().toURI());
+            } catch (URISyntaxException e) {
+                throw Exceptions.propagate(e);
+            }
+            final HttpHeaders headers = request.getHeaders();
+            if (headers != null) {
+                for (HttpHeader header : headers) {
+                    final String headerName = header.getName();
+                    if (!JDK11_RESTRICTED_HEADERS.contains(headerName)) {
```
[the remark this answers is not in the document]
anuchandy: yes, updated to log these as errors.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
```diff
@@ -0,0 +1,197 @@
...
+            return getBodyAsByteArray()
+                .map(bytes -> new String(bytes, charset));
+        }
+
+        @Override
+        public void close() {
+            if (!this.disposed) {
+                this.contentFlux
+                    .subscribe()
+                    .dispose();
```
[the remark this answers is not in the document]
anuchandy: fixed.

sdk/core/azure-core-http-jdk11-httpclient/pom.xml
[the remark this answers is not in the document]
anuchandy: removed the `11` from all places (dir, files, types, etc..).

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
anuchandy: going with a private instance-level method, Can be static if we ever move this method to utils.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk11/httpclient/Jdk11AsyncHttpClient.java
anuchandy: added necessary validation to log error in case this client is ever used with java_version <= 11.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilder.java
anuchandy: good catch, added code snippet.

sdk/core/azure-core-http-jdk11-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
```diff
@@ -0,0 +1,197 @@
...
+            }
+        }
+
+        private static HttpHeaders fromJdk11HttpHeaders(java.net.http.HttpHeaders headers) {
+            final HttpHeaders httpHeaders = new HttpHeaders();
+            for (final String key : headers.map().keySet()) {
+                for (final String value : headers.allValues(key)) {
+                    httpHeaders.put(key, value);
```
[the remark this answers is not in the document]
anuchandy: cool, updated to join multi-values using comma separator.

anuchandy:
@JimSuplizio thank you for explaining offline to me how ci.yml will be used. I've updated ci.yml to include this new module in the artifact list.

PR #8638, opened 2020-03-02
Distributed tracing for EventHubs: add enqueueTime to Process span links 

samvaity:
This PR adds enqueuedTime attribute on each link when processing messages: unix epoch time with milliseconds precision representing when the message was enqueued (x-opt-enqueued-time system property).

Ref : #7726

### DIFF
--- a/sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java
+++ b/sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java
@@ -35,7 +35,6 @@ public class OpenTelemetryTracer implements com.azure.core.util.tracing.Tracer {
     static final String AZ_NAMESPACE_KEY = "az.namespace";
     static final String MESSAGE_BUS_DESTINATION = "message_bus.destination";
     static final String PEER_ENDPOINT = "peer.address";
-    static final String MESSAGE_ENQUEUED_TIME = "x-opt-enqueued-time";
 
     private final ClientLogger logger = new ClientLogger(OpenTelemetryTracer.class);
 


sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventProcessorClientTest.java
```diff
@@ -200,9 +202,10 @@ public void testProcessSpans() throws Exception {
             .createConsumer(anyString(), anyInt()))
             .thenReturn(consumer1);
         when(eventData1.getSequenceNumber()).thenReturn(1L);
-        when(eventData2.getSequenceNumber()).thenReturn(2L);
+        when(eventData1.getSequenceNumber()).thenReturn(2L);
```
conniey: Did you mean to change it from eventData2 -> eventData1? The same event data returns two different sequence values?

samvaity: Was meaning to remove eventData2 altogether.  Removed now.

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/Tracer.java
```diff
@@ -62,6 +62,11 @@
      */
     String SPAN_BUILDER_KEY = "builder";
 
+    /**
+     * Key for {@link Context} which indicates the the time of the last enqueued message in the partition's stream.
+     */
+    String MESSAGE_ENQUEUED_TIME = "x-opt-enqueued-time";
```
srnagar: Same constant is declared in OpenTelemetryTracer too. Can this defined in just one place?

sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -256,28 +260,32 @@ private static Context setContextData(Span span) {
...
-            AttributeValue.stringAttributeValue(getOrDefault(context, AZ_TRACING_NAMESPACE_KEY, "", String.class)));
+        String entityPath = getOrDefault(context, ENTITY_PATH_KEY, null, String.class);
+        if (entityPath != null) {
+            span.setAttribute(MESSAGE_BUS_DESTINATION, AttributeValue.stringAttributeValue(entityPath));
+        }
+        String hostName = getOrDefault(context, HOST_NAME_KEY, null, String.class);
+        if (hostName != null) {
+            span.setAttribute(PEER_ENDPOINT, AttributeValue.stringAttributeValue(hostName));
+        }
+        Long messageEnqueuedTime = getOrDefault(context, MESSAGE_ENQUEUED_TIME, null, Long.class);
```
srnagar: Should we convert time to `Instant` type? On the span, would a string formatted datetime make it easier than a long?

samvaity: There is a direct specification to have this as a long value. It could be directly be used as-is if kept long, maybe?
>Please add enqueuedTime attribute on each link when processing messages: unix epoch time with milliseconds precision representing when message was enqueued (x-opt-enqueued-time system property). Attribute value should have long type.

PR #8645, opened 2020-03-03
ContentRetryableResponseUtil to create response with retryable content

anuchandy:


anuchandy:
don't review this code yet, it's a solution I came up with while looking into customer case where the read operation gets hanged. There is an existing type in storage named `ReliableDownload`, I'm also looking into it to see whether it can be extended to handle this case, if so then that is preferred.

anuchandy:
Closing this PR: A similar concept is now incorporated in the storage ReliableDownload type.

PR #8648, opened 2020-03-03
Set return-client-request-id

alzimmermsft:
This PR updates the builder in azure-search to set `return-client-request-id` instead of `x-ms-return-client-request-id`.

### DIFF
--- a/sdk/search/azure-search/src/test/java/com/azure/search/SearchServiceSyncTests.java
+++ b/sdk/search/azure-search/src/test/java/com/azure/search/SearchServiceSyncTests.java
@@ -6,11 +6,11 @@ import com.azure.core.http.rest.Response;
 import com.azure.core.util.Context;
 import com.azure.search.models.RequestOptions;
 import com.azure.search.models.ServiceStatistics;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import java.util.UUID;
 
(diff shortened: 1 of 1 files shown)


alzimmermsft:
> 
> 
> Is there test coverage for this?

Will add it soon, wanted to verify this was the correct direction for the change.

sdk/search/azure-search/src/test/java/com/azure/search/SearchServiceSyncTests.java
```diff
@@ -26,4 +31,16 @@ public void getServiceStatsReturnsCorrectDefinitionWithResponse() {
...
+
+    @Test
+    public void getServiceStatsReturnsRequestId() {
+        SearchServiceClient serviceClient = getSearchServiceClientBuilder().buildClient();
+
+        UUID expectedUuid = UUID.randomUUID();
+        RequestOptions requestOptions = new RequestOptions().setClientRequestId(expectedUuid);
+        Response<ServiceStatistics> response = serviceClient.getServiceStatisticsWithResponse(requestOptions, Context.NONE);
+
+        assertEquals(expectedUuid.toString(), response.getHeaders().getValue("client-request-id"));
```
brjohnstmsft: @alzimmermsft In the next month or so, we'll be deploying a service update that will set both `client-request-id` and `x-ms-client-request-id` on the response if either `return-client-request-id` or `x-ms-return-client-request-id` is specified.

PR #8679, opened 2020-03-03
Added Okhttp and ServiceVesion into client test framework

sima-zhu:
The changes are made because we want to extend the ability of our test framework to test on Okhttp and serviceVersion and etc (may have more in future). 

The change is currently made to AppConfig and KeyVault. Will run live tests for a while to learn the behavior.

### DIFF
--- a/sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
+++ b/sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
@@ -96,7 +96,7 @@ public abstract class TestBase implements BeforeEachCallback {
      *
      * @return The TestMode that has been initialized.
      */
-    public static TestMode getTestMode() {
+    public TestMode getTestMode() {
         return testMode;
     }
 


pom.xml
```diff
@@ -6,7 +6,15 @@
...
-  <version>1.0.0</version> <!-- {x-version-update;com.azure:azure-sdk-all;current} -->
+  <version>1.0.0</version>
+    <dependencies>
+        <dependency>
+            <groupId>com.azure</groupId>
+            <artifactId>azure-core</artifactId>
+            <version>1.4.0-beta.2</version>
+            <scope>compile</scope>
+        </dependency>
+    </dependencies> <!-- {x-version-update;com.azure:azure-sdk-all;current} -->
```
alzimmermsft: This will add Azure Core as a dependency to all modules depending pom.xml, this dependency should be removed.

JimSuplizio: @sima-zhu I don't believe this is the correct place to have this dependency (which, btw, is declared incorrectly anyways, the version tag belongs on the version line). This is azure-sdk-all which includes client, data and management in its module list. This change needs to be more limited, possibly just in pom.client.xml

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationAsyncClientTest.java
```diff
@@ -91,17 +97,21 @@ public void addConfigurationSetting() {
     /**
      * Tests that we cannot add a configuration setting when the key is an empty string.
      */
-    @Test
-    public void addConfigurationSettingEmptyKey() {
+    @ParameterizedTest(name = DISPLAY_NAME_WITH_ARGUMENTS)
+    @MethodSource("getTestParameters")
+    public void addConfigurationSettingEmptyKey(HttpClient httpClient, ConfigurationServiceVersion serviceVersion) {
```
alzimmermsft: Not a huge fan that every test will need to become parameterized to test each client, are there any ways to abstract this into Azure Core Test or does JUnit have the ability to parameterize an entire test run?

sdk/keyvault/azure-security-keyvault-certificates/pom.xml
```diff
@@ -97,7 +97,14 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-test</artifactId>
-      <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core-test;dependency} -->
+      <version>1.2.0-beta.1</version> <!-- {x-version-update;unrelease_com.azure:azure-core-test;current} -->
```
JimSuplizio: This tag is wrong, unreleased dependencies should only ever be dependency. The tag should be x-version-update;unreleased_com.azure:azure-core-test;dependency. As it is right now it'll definitely fail the analyze step's version scanner.

sdk/keyvault/azure-security-keyvault-keys/pom.xml
```diff
@@ -95,7 +95,14 @@
       <dependency>
         <groupId>com.azure</groupId>
         <artifactId>azure-core-test</artifactId>
-        <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core-test;dependency} -->
+        <version>1.2.0-beta.1</version> <!-- {x-version-update;unrelease_com.azure:azure-core-test;current} -->
```
JimSuplizio: Same comment as above, the tag is incorrect and should be x-version-update;unreleased_com.azure:azure-core-test;dependency

sdk/core/azure-core/src/main/java/com/azure/core/implementation/http/HttpClientProviders.java
```diff
@@ -34,4 +36,19 @@ public static HttpClient createInstance() {
         }
         return defaultProvider.createInstance();
     }
+
+    /**
+     * Returns a list of all {@link HttpClient HttpClients} that are discovered in the classpath.
+     *
+     * @return A list of all {@link HttpClient HttpClients} discovered in the classpath.
+     */
+    public static List<HttpClient> getAllHttpClients() {
```
alzimmermsft: I don't think we would want this to be public API

pom.xml
sima-zhu: Wrong place to add. Will remvoe.

sima-zhu: Not intended to put here.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyAsyncClientTest.java
```diff
@@ -30,24 +36,31 @@
...
+    }
+
+
+    private void getKeyAsyncClient(HttpClient httpClient,
+        KeyServiceVersion serviceVersion) {
+
+        client = clientSetup(pipeline -> new KeyClientBuilder()
+            .vaultUrl(getEndpoint())
+            .pipeline(pipeline)
+            .httpClient(httpClient == null ? interceptorManager.getPlaybackClient() : httpClient)
```
srnagar: Same as above. Pipeline and http client both should not be set.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyClientTest.java
```diff
@@ -17,54 +18,65 @@
...
 
     private KeyClient client;
 
     @Override
     protected void beforeTest() {
         beforeTestSetup();
+        client = clientSetup(pipeline -> new KeyClientBuilder()
+            .vaultUrl(getEndpoint())
+            .pipeline(pipeline)
+            .buildClient());
```
srnagar: This can be deleted.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyClientTest.java
```diff
@@ -17,54 +18,65 @@
...
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class KeyClientTest extends KeyClientTestBase {
+    static final String DISPLAY_NAME_WITH_ARGUMENTS = "{displayName} with [{arguments}]";
```
srnagar: Move this to `KeyClientTestBase`

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyClientTest.java
```diff
@@ -17,54 +18,65 @@
...
-                .pipeline(pipeline)
-                .buildClient());
-        }
+    private void getKeyClient(HttpClient httpClient,
+        KeyServiceVersion serviceVersion) {
+
+        client = clientSetup(pipeline -> new KeyClientBuilder()
+            .vaultUrl(getEndpoint())
+            .pipeline(pipeline)
+            .httpClient(httpClient == null ? interceptorManager.getPlaybackClient() : httpClient)
```
srnagar: Set the httpclient in pipeline instead.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/http/HttpClientProviders.java
sima-zhu: This is decided by Srikanta and Jonathan. @srnagar introduce the logic here. We want to expose the clients to test framework, so we are able to get all clients from test without manual adding clients in test every time. It is under implementation, the only pain is to expose the API to azure-core-test. I think we agree to approach in this way.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
```diff
@@ -113,8 +123,9 @@ public void timeoutPolicy() {
         assertThrows(RuntimeException.class, () -> client.setConfigurationSetting(key, null, value));
     }
 
-    @Test
-    public void nullServiceVersion() {
+    @ParameterizedTest(name = DISPLAY_NAME_WITH_ARGUMENTS)
+    @MethodSource("getTestParameters")
+    public void nullServiceVersion(HttpClient httpClient) {
```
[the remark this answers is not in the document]
sima-zhu: This is testing null Service Version. 
I test it. It is working for not adding version. The display name makes more sense

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyAsyncClientTest.java
sima-zhu: Will rework here. Thanks!

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationAsyncClientTest.java
JonathanGiles: JUnit 4 did support this, but as far as I know, JUnit 5 does not - you now parameterise each test separately.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
```diff
@@ -184,4 +193,23 @@ private static URI getURI(String endpointFormat, String namespace, String domain
...
+        // cartesian product of arguments - https://github.com/junit-team/junit5/issues/1427
+        List<Arguments> argumentsList = new ArrayList<>();
+        getHttpClients()
+            .forEach(httpClient -> {
+                for (ConfigurationServiceVersion serviceVersion : ConfigurationServiceVersion.values()) {
+                    argumentsList.add(Arguments.of(httpClient, serviceVersion));
+                }
+            });
+        return argumentsList.stream();
+    }
```
JonathanGiles: It's really critical that for a client library, you only have this method once. All test classes should reference a single version of this method.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationAsyncClientTest.java
sima-zhu: > First, I'd like to see a description so I can understand why these changes are necessary. Second, I don't believe that pom.xml isn't the correct place to add a core dependency.

Thanks for suggestion! Added the description.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
sima-zhu: Will add TestHelper for having this.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyClientTest.java
```diff
@@ -30,41 +34,43 @@
...
-                .vaultUrl(getEndpoint())
-                .pipeline(pipeline)
-                .buildClient());
-        }
+    private void getKeyClient(HttpClient httpClient, KeyServiceVersion serviceVersion) {
+        HttpPipeline httpPipeline = getHttpPipeline(httpClient, serviceVersion);
+        client = new KeyClientBuilder()
+            .vaultUrl(getEndpoint())
+            .pipeline(httpPipeline)
+            .buildClient();
```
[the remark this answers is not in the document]
sima-zhu: Ah, right. Making changes in UserAgent one. Forgot to update here. 
Thanks!

sdk/keyvault/azure-security-keyvault-secrets/src/test/java/com/azure/security/keyvault/secrets/SecretClientTestBase.java
```diff
@@ -437,4 +433,23 @@ public void sleep(long millis) {
...
+     *
+     * @return A stream of HttpClient and service version combinations to test.
+     */
+    static Stream<Arguments> getTestParameters() {
+        // when this issues is closed, the newer version of junit will have better support for
+        // cartesian product of arguments - https://github.com/junit-team/junit5/issues/1427
+        List<Arguments> argumentsList = new ArrayList<>();
+        getHttpClients()
+            .forEach(httpClient -> {
+                for (SecretServiceVersion serviceVersion : SecretServiceVersion.values()) {
```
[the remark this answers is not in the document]
sima-zhu: Do we want to introduce new Environment variable to control this?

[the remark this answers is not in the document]
JonathanGiles: Yes. My proposal to @srnagar yesterday was the following:

* AZURE_TEST_SERVICE_VERSIONS = all | latest
* AZURE_TEST_HTTP_CLIENTS = all | netty

If no environment variable is set, use `latest` and `netty`.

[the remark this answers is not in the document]
sima-zhu: Talked with Srikanta. We only introduce the AZURE_TEST_SERVICE_VERSIONS variables here for the following two reasons.
1. The client library has no ability to tell the httpClient is instance of Netty or Okhttp in current structure.
1. HttpClient has no impact on PR validation, as we only run playbackClient for PR.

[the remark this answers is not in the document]
srnagar: I later spoke to @sima-zhu and provided sample for enabling HttpClient env variable as well in TestBase's `shouldClientBeTested()` method.

So, both http client and service version should be supported now.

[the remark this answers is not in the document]
sima-zhu: Yes. Tested over the method. We can support both variables.

pom.client.xml
```diff
@@ -958,7 +957,11 @@
                 --add-opens com.azure.core/com.azure.core.http=ALL-UNNAMED
                 --add-opens com.azure.core/com.azure.core.http.policy=ALL-UNNAMED
                 --add-opens com.azure.core/com.azure.core.util=ALL-UNNAMED
+                --add-opens com.azure.core/com.azure.core.implementation.http=ALL-UNNAMED
                 --add-opens com.azure.ai.textanalytics/com.azure.ai.textanalytics=ALL-UNNAMED
+                --add-opens com.azure.data.appconfiguration/com.azure.data.appconfiguration=ALL-UNNAMED
+                --add-opens com.azure.security.keyvault.certificates/com.azure.security.keyvault.certificates=ALL-UNNAMED
+                --add-opens com.azure.security.keyvault.secrets/com.azure.security.keyvault.secrets=ALL-UNNAMED
```
srnagar: Don't you need KV Keys too?

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/AadCredentialTest.java
```diff
@@ -3,29 +3,35 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static com.azure.data.appconfiguration.TestHelper.DISPLAY_NAME_WITH_ARGUMENTS;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
 
 /**
  * Unit test for construct a configuration client by using AAD token credential.
  */
+@TestInstance(PER_CLASS)
```
srnagar: Now that the MethodSource is a `static` method, is this still required?

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
```diff
@@ -22,9 +22,15 @@
...
 import java.util.Objects;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 
+import static com.azure.data.appconfiguration.TestHelper.DISPLAY_NAME_WITH_ARGUMENTS;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
 
+@TestInstance(PER_CLASS)
```
srnagar: Is this still required?

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
```diff
@@ -127,22 +134,21 @@ public void nullServiceVersion() {
...
         ConfigurationSetting addedSetting = clientBuilder.buildClient().setConfigurationSetting(key, null, value);
         Assertions.assertEquals(addedSetting.getKey(), key);
         Assertions.assertEquals(addedSetting.getValue(), value);
     }
 
-    @Test
-    public void defaultPipeline() {
+    @ParameterizedTest(name = DISPLAY_NAME_WITH_ARGUMENTS)
+    @MethodSource("com.azure.data.appconfiguration.TestHelper#getTestParameters")
+    public void defaultPipeline(HttpClient httpClient, ConfigurationServiceVersion serviceVersion) {
```
srnagar: Since this test is to use the default pipeline (i.e. whichever httpclient is selected by default along with all other policies), this test should only have service version as a parameter.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientTest.java
```diff
@@ -5,31 +5,37 @@
...
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 
+import static com.azure.data.appconfiguration.TestHelper.DISPLAY_NAME_WITH_ARGUMENTS;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
 
+@TestInstance(PER_CLASS)
```
srnagar: Check if this is still required.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/TestHelper.java
```diff
@@ -0,0 +1,41 @@
...
+class TestHelper extends TestBase {
+    static final String DISPLAY_NAME_WITH_ARGUMENTS = "{displayName} with [{arguments}]";
+
+    /**
+     * Returns a stream of arguments that includes all combinations of eligible {@link HttpClient HttpClients} and
+     * service versions that should be tested.
+     *
+     * @return A stream of HttpClient and service version combinations to test.
+     */
+    protected static Stream<Arguments> getTestParameters() {
```
srnagar: package-private instead of protected.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/TestHelper.java
```diff
@@ -0,0 +1,41 @@
...
+        List<Arguments> argumentsList = new ArrayList<>();
+        getHttpClients()
+            .forEach(httpClient -> {
+                Arrays.stream(ConfigurationServiceVersion.values()).filter(TestHelper::shouldServiceVersionBeTested)
+                    .forEach(serviceVersion -> argumentsList.add(Arguments.of(httpClient, serviceVersion)));
+            });
+        return argumentsList.stream();
+    }
+
+    protected static boolean shouldServiceVersionBeTested(ConfigurationServiceVersion serviceVersion) {
```
srnagar: private instead of protected

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/TestHelper.java
```diff
@@ -0,0 +1,41 @@
...
+        if (Configuration.getGlobalConfiguration().get(AZURE_TEST_SERVICE_VERSIONS) == null) {
+            return ConfigurationServiceVersion.getLatest().equals(serviceVersion);
+        }
+        return true;
```
srnagar: This should check what was in the environment variable before returning true.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
```diff
@@ -121,6 +131,37 @@ protected void beforeTest() {
...
+    }
+
+    /**
+     * Filter out the http clients needs to run in test framework
+     * @param client Http client needs to check
+     * @return Boolean indicates whether filters out the client or not.
+     */
+    protected static boolean shouldClientBeTested(HttpClient client) {
+        // This is for when we decide to filter some http clients based on some criteria
+        // to reduce the time take to run the tests.
```
srnagar: Replace this comment with what this method is doing. We are using Netty http client as default if env variable is not set. If it's set to ALL, all HttpClients in the classpath will be tested else the name of the HttpClient class should match env variable.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
```diff
@@ -121,6 +131,37 @@ protected void beforeTest() {
...
+        String configuredHttpClientToTest = Configuration.getGlobalConfiguration().get(AZURE_TEST_HTTP_CLIENTS);
```
srnagar: This should support a comma-separated list of HttpClient names.

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTest.java
```diff
@@ -6,44 +6,65 @@
...
+    }
 
-    @Test
-    public void createCertificate() {
+    @ParameterizedTest(name = DISPLAY_NAME_WITH_ARGUMENTS)
+    @MethodSource("getTestParameters")
+    public void createCertificate(HttpClient httpClient, CertificateServiceVersion serviceVersion) {
+        getCertificateClient(httpClient, serviceVersion);
         createCertificateRunner((policy) -> {
+            getCertificateClient(httpClient, serviceVersion);
```
srnagar: Why's getCertificateClient() called twice?

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTest.java
```diff
@@ -6,44 +6,65 @@
...
-                .buildClient());
-        } else {
-            client = clientSetup(pipeline -> new CertificateClientBuilder()
-                .vaultUrl(getEndpoint())
-                .pipeline(pipeline)
-                .buildClient());
-        }
     }
 
+    private void getCertificateClient(HttpClient httpClient,
```
srnagar: Since this is not returning anything, consider renaming this to createCertificateClient() or setCertificateClient()

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
```diff
@@ -64,7 +65,8 @@ protected String getTestName() {
     void beforeTestSetup() {
     }
 
-    <T> T clientSetup(Function<HttpPipeline, T> clientBuilder) {
+    HttpPipeline getHttpPipeline(HttpClient httpClient, CertificateServiceVersion serviceVersion) {
+   // <T> T clientSetup(Function<HttpPipeline, T> clientBuilder) {
```
srnagar: nit: delete this line

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
```diff
@@ -605,5 +592,31 @@ public void sleep(long millis) {
...
+                    .forEach(serviceVersion -> argumentsList.add(Arguments.of(httpClient, serviceVersion)));
+            });
+        return argumentsList.stream();
+    }
+
+    static boolean shouldServiceVersionBeTested(CertificateServiceVersion serviceVersion) {
+        if (Configuration.getGlobalConfiguration().get(AZURE_TEST_SERVICE_VERSIONS) == null) {
+            return CertificateServiceVersion.getLatest().equals(serviceVersion);
+        }
+        return true;
```
srnagar: This should check the environment variable before returning true.

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
```diff
@@ -402,11 +396,11 @@ void importCertificateRunner(Consumer<ImportCertificateOptions> testRunner) {
     }
 
     @Test
-    public abstract  void importPemCertificate() throws IOException;
+    public abstract  void importPemCertificate(HttpClient httpClient, CertificateServiceVersion serviceVersion) throws IOException;
 
     void importPemCertificateRunner(Consumer<ImportCertificateOptions> testRunner) throws IOException {
 
-        byte[] certificateContent = readCertificate("pemCert.pem");
+        byte[] certificateContent = readCertificate("certificate.pem");
```
srnagar: The file name was changed. Was this intended?

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
```diff
@@ -419,24 +413,17 @@ void importPemCertificateRunner(Consumer<ImportCertificateOptions> testRunner) t
...
-        BufferedReader br = new BufferedReader(new FileReader(pemPath));
-        try {
-            String line;
-            while ((line = br.readLine()) != null) {
-                pemCert += line + "\n";
-            }
-        } finally {
-            br.close();
+        if (pemPath.contains(":")) {
+            pemPath = pemPath.substring(1);
```
srnagar: Why was this change required?

sdk/keyvault/azure-security-keyvault-keys/pom.xml
```diff
@@ -41,6 +41,11 @@
         <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-surefire-plugin;external_dependency} -->
         <configuration>
           <rerunFailingTestsCount>3</rerunFailingTestsCount>
+          <argLine>
+            --add-exports com.azure.core/com.azure.core.implementation.http=ALL-UNNAMED
+            --add-opens com.azure.security.keyvault.keys/com.azure.security.keyvault.keys.cryptography=ALL-UNNAMED
+            --add-opens com.azure.security.keyvault.keys/com.azure.security.keyvault.keys=ALL-UNNAMED
+          </argLine>
```
srnagar: Since all other `--add-opens` are in `pom.client.xml`, it's better to move this there to keep them all in one place.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyAsyncClientTest.java
```diff
@@ -29,25 +33,25 @@
...
-                .pipeline(pipeline)
-                .buildAsyncClient());
-        } else {
-            client = clientSetup(pipeline -> new KeyClientBuilder()
-                .pipeline(pipeline)
-                .vaultUrl(getEndpoint())
-                .buildAsyncClient());
-        }
+
+    private void getKeyAsyncClient(HttpClient httpClient, KeyServiceVersion serviceVersion) {
```
srnagar: Update name of the method.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyClientTest.java
```diff
@@ -426,4 +475,29 @@ private DeletedKey pollOnKeyPurge(String keyName) {
         return null;
     }
 
+    /**
+     * Returns a stream of arguments that includes all combinations of eligible {@link HttpClient HttpClients} and
+     * service versions that should be tested.
+     *
+     * @return A stream of HttpClient and service version combinations to test.
+     */
+    static Stream<Arguments> getTestParameters() {
```
srnagar: This method is a duplicate of `getTestParameters()` method in `KeyClientTestBase`.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTest.java
```diff
@@ -32,35 +36,34 @@
...
+    private void initializeKeyClient(HttpClient httpClient) {
+        pipeline = getHttpPipeline(httpClient, KeyServiceVersion.getLatest());
+        client = new KeyClientBuilder()
+            .pipeline(pipeline)
+            .vaultUrl(getEndpoint())
+            .buildClient();
+    }
+
+    private CryptographyClient initializeCryptographyClient(String keyId, HttpClient httpClient, CryptographyServiceVersion serviceVersion) {
+        pipeline = getHttpPipeline(httpClient, serviceVersion);
```
srnagar: Shouldn't the same pipeline be used for both KeyClient and CryptographyClient?

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTest.java
```diff
@@ -32,35 +36,34 @@
...
-                    .vaultUrl(getEndpoint())
-                    .buildClient();
-            });
-        }
     }
 
-    @Test
-    public void encryptDecryptRsa() throws Exception {
+    private void initializeKeyClient(HttpClient httpClient) {
+        pipeline = getHttpPipeline(httpClient, KeyServiceVersion.getLatest());
```
srnagar: KeyServiceVersion should also be parameterized

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTestBase.java
```diff
@@ -209,4 +202,23 @@ public void sleep(long millis) {
...
+     *
+     * @return A stream of HttpClient and service version combinations to test.
+     */
+    protected Stream<Arguments> getTestParameters() {
+        // when this issues is closed, the newer version of junit will have better support for
+        // cartesian product of arguments - https://github.com/junit-team/junit5/issues/1427
+        List<Arguments> argumentsList = new ArrayList<>();
+        getHttpClients()
+            .forEach(httpClient -> {
+                for (KeyServiceVersion serviceVersion : KeyServiceVersion.values()) {
```
srnagar: Since this is cryptography client test, should this be CryptographServiceVersion?

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/KeyEncryptionKeyClientTestBase.java
```diff
@@ -176,4 +159,21 @@ public void sleep(long millis) {
...
+     *
+     * @return A stream of HttpClient and service version combinations to test.
+     */
+    protected Stream<Arguments> getTestParameters() {
+        // when this issues is closed, the newer version of junit will have better support for
+        // cartesian product of arguments - https://github.com/junit-team/junit5/issues/1427
+        List<Arguments> argumentsList = new ArrayList<>();
+        getHttpClients()
+            .forEach(httpClient -> {
+                argumentsList.add(Arguments.of(httpClient));
```
srnagar: Should this also include service versions?

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTest.java
sima-zhu: Named to createCertificateClient

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTest.java
sima-zhu: Paste error. Removed.

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
sima-zhu: This is merge issue. Thanks for catching this.

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
sima-zhu: Not merge the most recent. 
Changed to master one.

sdk/keyvault/azure-security-keyvault-keys/pom.xml
sima-zhu: OK. Will move there.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTest.java
sima-zhu: Not exactly, `initializeKeyClient` is using the latest KeyServiceVersion, as the main purpose of the test is to test over CryptographyServiceClient, which is using CryptographyServiceVersion. There are some other tests for KeyClient

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTest.java
sima-zhu: Explained above. 

The test class is to test over CryptographyServiceClient instead of KeyClient.
There are some other test classes which tests KeyClient, and we have parameterized test on KeyServiceVersion over there.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTestBase.java
sima-zhu: This is left over. I forgot to remove. 
All tests here are using the helper method under TestHelper.

Removed

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/KeyEncryptionKeyClientTestBase.java
sima-zhu: Left over. It is supposed to remove.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientTest.java
sima-zhu: Removed

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/ConfigurationClientBuilderTest.java
sima-zhu: Removed

pom.client.xml
sima-zhu: I put it over keys pom.
Moved back to keep them in one place as you comment below.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyClientTest.java
sima-zhu: Removed.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
sima-zhu: Made changes to support comma.

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -36,19 +36,26 @@
...
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.3.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.4.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core;dependency} -->
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-http-netty</artifactId>
       <version>1.4.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
+      <scope>test</scope>
```
srnagar: Netty dependency should not be test scoped.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/AadCredentialTest.java
```diff
@@ -24,8 +27,8 @@
     static String connectionString;
     static TokenCredential tokenCredential;
 
-    @BeforeEach
-    public void setup() throws InvalidKeyException, NoSuchAlgorithmException {
+    public void setup(HttpClient httpClient, ConfigurationServiceVersion serviceVersion)
```
srnagar: This method can now be private.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/TestHelper.java
```diff
@@ -0,0 +1,43 @@
...
+        getHttpClients()
+            .forEach(httpClient -> {
+                Arrays.stream(ConfigurationServiceVersion.values()).filter(TestHelper::shouldServiceVersionBeTested)
+                    .forEach(serviceVersion -> argumentsList.add(Arguments.of(httpClient, serviceVersion)));
+            });
+        return argumentsList.stream();
+    }
+
+    private static boolean shouldServiceVersionBeTested(ConfigurationServiceVersion serviceVersion) {
+        String serviceVersionFromEnv = Configuration.getGlobalConfiguration().get(AZURE_TEST_SERVICE_VERSIONS);
```
srnagar: This should be comma-separated list of versions just like http client env variable and should support "ALL".

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
```diff
@@ -23,6 +28,12 @@
 public abstract class TestBase implements BeforeEachCallback {
     // Environment variable name used to determine the TestMode.
     private static final String AZURE_TEST_MODE = "AZURE_TEST_MODE";
+    private static final String AZURE_TEST_HTTP_CLIENTS = "AZURE_TEST_HTTP_CLIENTS";
+    protected static final String AZURE_TEST_HTTP_CLIENTS_VALUE_ALL = "ALL";
+    protected static final String AZURE_TEST_HTTP_CLIENTS_VALUE_NETTY = "NettyAsyncHttpClient";
+    protected static final String AZURE_TEST_HTTP_CLIENTS_VALUE_OKHTTP = "OkHttpAsyncHttpClient";
```
srnagar: This is not required

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
```diff
@@ -23,6 +28,12 @@
 public abstract class TestBase implements BeforeEachCallback {
     // Environment variable name used to determine the TestMode.
     private static final String AZURE_TEST_MODE = "AZURE_TEST_MODE";
+    private static final String AZURE_TEST_HTTP_CLIENTS = "AZURE_TEST_HTTP_CLIENTS";
+    protected static final String AZURE_TEST_HTTP_CLIENTS_VALUE_ALL = "ALL";
```
srnagar: protected and static are contradictory. Change this to public instead.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
```diff
@@ -121,6 +132,50 @@ protected void beforeTest() {
     protected void afterTest() {
     }
 
+    /**
+     * Returns a list of {@link HttpClient HttpClients} that should be tested.
+     * @return A list of {@link HttpClient HttpClients} to be tested.
+     */
+    protected static Stream<HttpClient> getHttpClients() {
```
srnagar: public static instead of protected static.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
srnagar: Update the comment here.

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
srnagar: This should be updated too.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyClientTestBase.java
```diff
@@ -413,4 +404,30 @@ public void sleep(long millis) {
...
+                    .forEach(serviceVersion -> argumentsList.add(Arguments.of(httpClient, serviceVersion)));
+            });
+        return argumentsList.stream();
+    }
+
+    private static boolean shouldServiceVersionBeTested(KeyServiceVersion serviceVersion) {
+        if (Configuration.getGlobalConfiguration().get(AZURE_TEST_SERVICE_VERSIONS) == null) {
+            return KeyServiceVersion.getLatest().equals(serviceVersion);
+        }
+        return true;
```
srnagar: Check the env variable before returning true.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTest.java
srnagar: Before this change, the same pipeline was used for both.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/TestHelper.java
```diff
@@ -0,0 +1,43 @@
...
+                    .forEach(serviceVersion -> argumentsList.add(Arguments.of(httpClient, serviceVersion)));
+            });
+        return argumentsList.stream();
+    }
+
+    private static boolean shouldServiceVersionBeTested(CryptographyServiceVersion serviceVersion) {
+        if (Configuration.getGlobalConfiguration().get(AZURE_TEST_SERVICE_VERSIONS) == null) {
+            return CryptographyServiceVersion.getLatest().equals(serviceVersion);
+        }
+        return true;
```
srnagar: Check env variable before returning true.

sdk/keyvault/azure-security-keyvault-secrets/src/test/java/com/azure/security/keyvault/secrets/SecretClientTestBase.java
```diff
@@ -437,4 +430,30 @@ public void sleep(long millis) {
...
+                    .forEach(serviceVersion -> argumentsList.add(Arguments.of(httpClient, serviceVersion)));
+            });
+        return argumentsList.stream();
+    }
+
+    private static boolean shouldServiceVersionBeTested(SecretServiceVersion serviceVersion) {
+        if (Configuration.getGlobalConfiguration().get(AZURE_TEST_SERVICE_VERSIONS) == null) {
+            return SecretServiceVersion.getLatest().equals(serviceVersion);
+        }
+        return true;
```
srnagar: Same here - check env variable before returning true.

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
sima-zhu: Done

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/AadCredentialTest.java
sima-zhu: Done.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
sima-zhu: Done

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
sima-zhu: Removed

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/KeyClientTestBase.java
sima-zhu: Done

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTest.java
```diff
@@ -32,27 +35,20 @@
...
     }
 
-    @Test
-    public void encryptDecryptRsa() throws Exception {
+    private void initializeClient(HttpClient httpClient, CryptographyServiceVersion serviceVersion) {
+        pipeline = getHttpPipeline(httpClient, serviceVersion);
+        client = new KeyClientBuilder()
+            .pipeline(pipeline)
+            .vaultUrl(getEndpoint())
+            .buildClient();
```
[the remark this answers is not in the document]
sima-zhu: `initializeKeyClient` is not the major client we want to test in this class but `initializeCryptographyClient`. There are another test classes testing KeyClient. What we use for keyClientBuilder are default ones (netty, latest)

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/TestHelper.java
conniey: yes, iirc, the guidelines say to use service specific names for env variables.
https://azure.github.io/azure-sdk/java_implementation.html#service-specific-environment-variables

sima-zhu: Will make the change

srnagar: Also, don't read from env var every single time `shouldServiceVersionBeTested` is called. Read it once before and pass the value to this method.

sima-zhu: Put it to static final field, so we can fetch the value after initialization.

sdk/keyvault/azure-security-keyvault-keys/src/test/java/com/azure/security/keyvault/keys/cryptography/CryptographyClientTest.java
sima-zhu: Talked offline. This does not affect the logic for tests. Besides of that, These two do not use exactly same pipelines. UserAgentPolicy are not the same.

eng/versioning/version_client.txt
```diff
@@ -44,6 +44,8 @@ com.azure:perf-test-core;1.0.0-beta.1;1.0.0-beta.1
 # note: The unreleased dependencies will not be manipulated with the automatic PR creation code.
 unreleased_com.azure:azure-core;1.4.0-beta.1
 unreleased_com.azure:azure-core-amqp;1.1.0-beta.1
+unreleased_com.azure:azure-core-test;1.2.0-beta.1
+unreleased_com.azure:azure-core;1.4.0-beta.1
```
JimSuplizio: @sima-zhu this is the reason you're seeing the analyze step fail. There are two unreleased_com.azure:azure-core;1.4.0-beta.1 entries, one on line 48 and the other on line 43. When a duplicate entry is encountered my script was doing a write-error which apparently caused the wrapper script to exit without flushing the buffer first so it was never seen. I'll be fixing my script so that doesn't happen again.

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/TestHelper.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.provider.Arguments;
+
+import static com.azure.core.test.TestBase.getHttpClients;
+
+class TestHelper {
+    static final String DISPLAY_NAME_WITH_ARGUMENTS = "{displayName} with [{arguments}]";
+    private static final String AZURE_TEST_SERVICE_VERSIONS = "AZURE_APPCONFIG_SERVICE_VERSIONS";
```
srnagar: Name the variable and value as `AZURE_APPCONFIG_TEST_SERVICE_VERSIONS` as we don't want it to be mistaken for some non-test related env variable. 
```suggestion
    private static final String AZURE_APPCONFIG_TEST_SERVICE_VERSIONS = "AZURE_APPCONFIG_TEST_SERVICE_VERSIONS";
```

sdk/appconfiguration/azure-data-appconfiguration/src/test/java/com/azure/data/appconfiguration/TestHelper.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String serviceVersionFromEnv = Configuration.getGlobalConfiguration().get(AZURE_TEST_SERVICE_VERSIONS);
+    private static final String AZURE_TEST_SERVICE_VERSIONS_VALUE_ALL = "ALL";
```
srnagar: This can be declared once in `TestBase.java` in core-test module as ALL can be common for all libraries.

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
```diff
@@ -605,5 +601,53 @@ public void sleep(long millis) {
...
+     *
+     * Environment values currently supported are: "ALL", "${version}".
+     * Use comma to separate http clients want to test.
+     * e.g. {@code set AZURE_TEST_SERVICE_VERSIONS = V1_0, V2_0}
+     *
+     * @param serviceVersion ServiceVersion needs to check
+     * @return Boolean indicates whether filters out the service version or not.
+     */
+    private static boolean shouldServiceVersionBeTested(CertificateServiceVersion serviceVersion) {
+        String serviceVersionFromEnv = Configuration.getGlobalConfiguration().get(AZURE_TEST_SERVICE_VERSIONS);
```
srnagar: This should be fetched just once instead of reading from env variables for all client-service version combinations. Same for all other clients.

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
```diff
@@ -34,27 +34,31 @@
...
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.fail;
 
 public abstract class CertificateClientTestBase extends TestBase {
+    static final String DISPLAY_NAME_WITH_ARGUMENTS = "{displayName} with [{arguments}]";
     private static final String SDK_NAME = "client_name";
     private static final String SDK_VERSION = "client_version";
+    private static final String AZURE_TEST_SERVICE_VERSIONS = "AZURE_KEYVAULT_CERTIFICATE_SERVICE_VERSIONS";
```
srnagar: Add `TEST` to the name of env variable so that it's not mistaken to be non-test related config. Same for all other service version env variables used in tests.

sdk/keyvault/azure-security-keyvault-certificates/src/test/java/com/azure/security/keyvault/certificates/CertificateClientTestBase.java
sima-zhu: That's a miss. Make the changes.

PR #8683, opened 2020-03-03
Change searchServiceVersion to serviceVersion

sima-zhu:


sdk/search/azure-search/src/test/java/com/azure/search/AutocompleteSyncTests.java
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reservgited.
```
alzimmermsft: Typo here, revert this change

PR #8687, opened 2020-03-03
Add Additional UTF8 BOM Check

alzimmermsft:
This PR adds an additional UTF8 BOM check by checking for the String representation of the UTF8 BOM.

### DIFF
--- a/eng/spotbugs-aggregate-report/pom.xml
+++ b/eng/spotbugs-aggregate-report/pom.xml
@@ -122,7 +122,7 @@
         <dependency>
           <groupId>com.microsoft.azure</groupId>
           <artifactId>azure-eventhubs-eph</artifactId>
-          <version>3.1.0</version> <!-- {x-version-update;com.microsoft.azure:azure-eventhubs-eph;current} -->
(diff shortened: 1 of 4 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/JacksonAdapter.java
```diff
@@ -158,12 +161,17 @@ public String serializeList(List<?> list, CollectionFormat format) {
...
-        if (value.startsWith(BOM)) {
-            value = value.replaceFirst(BOM, "");
+
+        // Remove any leading BOM from the XML.
+        if (encoding == SerializerEncoding.XML) {
+            if (value.startsWith(BOM)) {
+                value = value.replaceFirst(BOM, "");
+            } else if (value.startsWith(BOM_STRING)) {
+                value = value.replaceFirst(BOM_STRING, "");
+            }
```
srnagar: As discussed, please file an issue for further investigation on this. This is good for now to release this as this issue was already noticed in smoke tests and would be better to get this fix out before customers are impacted.

alzimmermsft: https://github.com/Azure/azure-sdk-for-java/issues/8710

PR #8688, opened 2020-03-04
Added code for initial Blob Quick Query Input Stream implementation

gapra-msft:


sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryAsyncClient.java
```diff
@@ -4,10 +4,202 @@
...
+     * Queries the entire blob. NOTE: Returns raw avro.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.quickquery.BlobQuickQueryAsyncClient.query#String}
+     *
+     * @param expression The query expression.
+     * @return A reactive response containing the queried data.
+     */
+    public Flux<ByteBuffer> query(String expression) {
```
gapra-msft: for preview I can make these package private - eventually we want these methods to work though

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryClient.java
```diff
@@ -4,10 +4,138 @@
...
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.quickquery.BlobQuickQueryClient.query#OutputStream-String}
+     *
+     * @param stream A non-null {@link OutputStream} instance where the downloaded data will be written.
+     * @param expression The query expression.
+     * @throws UncheckedIOException If an I/O error occurs.
+     * @throws NullPointerException if {@code stream} is null.
+     */
+    public void query(OutputStream stream, String expression) {
```
gapra-msft: These can be package private for preview purposes - not meant to be exposed now

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/DelimitedTextConfiguration.java
```diff
@@ -129,7 +129,7 @@ public DelimitedTextConfiguration setEscapeChar(String escapeChar) {
      *
      * @return the hasHeaders value.
      */
-    public String getHasHeaders() {
+    public boolean isHasHeaders() {
```
rickle-msft: isHas

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ErrorReceiver.java
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.storage.common;
+
+/**
+ * An {@code ErrorReceiver} is an object that can be used to report errors on transfers. When specified on transfer
+ * operations, the {@code reportError} method will be called whenever errors are encountered. The user may configure
```
rickle-msft: 1. Is this still just for non-fatal errors?
2. Why is it generic? Are there multiple kinds of errors?

gapra-msft: I made it generic in case we want to reuse this for something else some day, but I might be being over eager.

gapra-msft: For now at least, this is for fatal and non fatal errors after our discussion with the service team.

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryClient.java
```diff
@@ -4,10 +4,142 @@
...
+    /**
+     * Package-private constructor for use by {@link BlobQuickQueryClientBuilder}.
+     * @param client the async blob quick query client
+     */
+    BlobQuickQueryClient(BlobQuickQueryAsyncClient client) {
+        this.client = client;
+    }
+
+    /**
+     * Opens a blob input stream to query the blob.
```
rickle-msft: May want to note that this returns parsed avro

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryClient.java
```diff
@@ -4,10 +4,142 @@
...
+        ErrorReceiver<BlobQuickQueryError> nonFatalErrorReceiver, ProgressReceiver progressReceiver) {
+
+        // Data to subscribe to and read from.
+        Flux<ByteBuffer> data = client.queryWithResponse(expression, input, output, requestConditions)
+            .flatMapMany(ResponseBase::getValue);
+
+        // Create input stream from the data.
+        FluxInputStream fluxInputStream = new FluxInputStream(data, logger);
+
+        return new BlobQuickQueryInputStream(fluxInputStream, nonFatalErrorReceiver, progressReceiver);
```
rickle-msft: Why a progress receiver? Typically we only take that when we're doing parallel operations

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
```diff
@@ -0,0 +1,192 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+public class BlobQuickQueryInputStream extends InputStream {
+
+    private DataFileStream<GenericRecord> parsedStream;
+    private ByteArrayInputStream userBuffer;
```
rickle-msft: A description of these fields would be helpful.

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
```diff
@@ -0,0 +1,192 @@
...
+import org.apache.avro.file.DataFileStream;
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.generic.GenericRecord;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+public class BlobQuickQueryInputStream extends InputStream {
```
rickle-msft: docs

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
```diff
@@ -0,0 +1,192 @@
...
+
+        /* If len is 0, then no bytes are read and 0 is returned. */
+        if (len == 0) {
+            return 0;
+        }
+        /* Attempt to read at least one byte. If no byte is available because the stream is at end of file,
+           the value -1 is returned; otherwise, at least one byte is read and stored into b. */
+
+        /* First read? */
+        if (firstRead) {
```
rickle-msft: I think we can generally skip a `firstRead` variable by just checking if `parsedStream` is null

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryClient.java
gapra-msft: The service returns "progress" records of how many bytes of the original blob have been parsed so far, so we wanted to expose this to the user as theyre reading from the stream

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
```diff
@@ -0,0 +1,192 @@
...
+            firstRead = false;
+            /* Reads from the network to determine schema. */
+            parsedStream = new DataFileStream<>(networkStream, new GenericDatumReader<>());
+            this.userBuffer = new ByteArrayInputStream(new byte[0]);
+        }
+
+        /* Now we are guaranteed that buffer is SOMETHING. */
+        /* No data is available in the buffer.  */
+        if (this.userBuffer.available() == 0) {
+            /* Try to get more data */
```
rickle-msft: I think adding a comment that we need the while loop because not every record gives us data, and we need to keep parsing until we find a data record would be helpful

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
```diff
@@ -0,0 +1,192 @@
...
+            while (this.userBuffer.available() == 0 && this.parsedStream.hasNext()) {
+                parseRecord(this.parsedStream.next());
+            }
+        }
+        /* Data is now available in the buffer. */
+        if (this.userBuffer.available() > 0) {
```
rickle-msft: If we are sure that data is available in the buffer, why do we need the if statement?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
```diff
@@ -0,0 +1,192 @@
...
+            return this.userBuffer.read(b, off, len);
+        }
+
+        /* End record was seen, there is no more data available to be read from the stream. Return -1. */
+        if (this.endRecordSeen) {
+            return -1;
+        }
+
+        return 0;
```
rickle-msft: Do we ever actually reach this case?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
```diff
@@ -0,0 +1,220 @@
...
+    private Subscription subscription; // Subscription to request more data from as needed
+
+    private ByteArrayInputStream buffer;
+
+    private boolean subscribed;
+    private boolean fluxComplete;
+    private boolean waitingForData;
+
+    private final Lock lock;
+    private final Condition dataAvailable;
```
rickle-msft: Also commending on some of these fields would be helpful

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
gapra-msft: good point

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
```diff
@@ -0,0 +1,220 @@
...
+        /* At this point, buffer should not be null. If it is, that indicates either an error or completion event
+           was emitted by the Flux. */
+        if (this.buffer == null) { // Only executed on first subscription.
+            if (this.lastError != null) {
+                throw lastError;
+            }
+            if (this.fluxComplete) {
+                return -1;
+            }
+            this.buffer = new ByteArrayInputStream(new byte[0]);
```
rickle-msft: What's the purpose of this line?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
```diff
@@ -0,0 +1,220 @@
...
+
+        /* Now we are guaranteed that buffer is SOMETHING. */
+        /* No data is available in the buffer.  */
+        if (this.buffer.available() == 0) {
+            /* If the flux completed, there is no more data available to be read from the stream. Return -1. */
+            if (this.fluxComplete) {
+                return -1;
+            }
+            /* If we are not waiting for another request to come in, request more data. */
+            if (!waitingForData) {
```
rickle-msft: How could we be here if we are waiting for data?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
```diff
@@ -0,0 +1,220 @@
...
+        if (this.buffer.available() > 0) {
+            return this.buffer.read(b, off, len);
+        }
+
+        /* If the flux completed, there is no more data available to be read from the stream. Return -1. */
+        if (this.fluxComplete) {
+            return -1;
+        }
+
+        return 0;
```
rickle-msft: Under what conditions could we end up here?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Override
+    public void close() throws IOException {
+        super.close();
```
rickle-msft: We may also want to cancel our subscription.

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
gapra-msft: I added a clarifying comment - at this point either data is available in buffer or we hit the end of the stream

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryClient.java
gapra-msft: I'm not sure this is something we want to expose to the user. They dont particularly care that it's avro that's been parsed right?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
gapra-msft: We shouldnt ever reach this case. I'll add a throws IOException since this indicates the end record wasnt seen

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
gapra-msft: That whole if block is really a "this is no longer the first read check." InputStreams are supposed to throw if an error happens on the first read.

gapra-msft: So I make buffer non null once the first read is done

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
gapra-msft: Removed that - I think it was a remnant of previous logic

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
gapra-msft: I think its an error - I'll throw an ioException but I'm not sure what the message should be - like this indicates that the flux just abruptly stopped? but I'm not sure if that's possible

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryClient.java
rickle-msft: I mostly mention this because the other apis mention that they return raw avro

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
rickle-msft: A possible alternative is IllegalStateException

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
rickle-msft: Again, I think IllegalStateException might be more descriptive. It doesn't have the same expectations as IOException for readable messages, so putting something vague and unhelpful like "Reached end of method. Should never happen," is more acceptable.

sdk/storage/azure-storage-quickquery/src/test/java/com/azure/storage/quickquery/APISpec.groovy
```diff
@@ -0,0 +1,398 @@
...
+            responseLeaseId = createLeaseClient(bc).acquireLease(-1)
+        }
+        if (leaseID == receivedLeaseID) {
+            return responseLeaseId
+        } else {
+            return leaseID
+        }
+    }
+
+    def setupBlobLeaseCondition(BlobAsyncClientBase bac, String leaseID) {
```
rickle-msft: If you can, cleaning out some of these helpers that we don't need would be helpful.

sdk/storage/azure-storage-quickquery/src/test/java/com/azure/storage/quickquery/BlobQQInputStreamTest.groovy
```diff
@@ -0,0 +1,448 @@
...
+import com.azure.storage.quickquery.models.BlobQuickQueryError
+
+import com.azure.storage.quickquery.models.BlobQuickQueryJsonSerialization
+import spock.lang.Unroll
+
+class BlobQQInputStreamTest extends APISpec {
+
+    // Generates and uploads a small 1KB CSV file
+    def uploadCsv(BlobQuickQueryDelimitedSerialization s, int numCopies) {
+        byte[] headers = ("rn1" + s.getColumnSeparator() + "rn2" + s.getColumnSeparator() + "rn3" + s.getColumnSeparator() + "rn4" + s.getRecordSeparator()).getBytes()
```
rickle-msft: `Strings.join(s.getColumnSeparator(), "rn1", "rn2"...) would be more concise

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryClient.java
gapra-msft: Theyre marked non public for now so it should be fine

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
```diff
@@ -0,0 +1,232 @@
...
+
+    private IOException lastError;
+
+    /**
+     * Creates a new FluxInputStream
+     *
+     * @param data The data to subscribe to and read from.
+     * @param logger {@link ClientLogger}
+     */
+    public FluxInputStream(Flux<ByteBuffer> data, ClientLogger logger) {
```
rickle-msft: I think FluxInputStream should have its own client logger rather than being passed one

sdk/storage/azure-storage-quickquery/src/test/java/com/azure/storage/quickquery/FluxInputStreamAPITest.groovy
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        is.close()
+
+        then:
+        for (int i = 0; i < num; i++) {
+            assert bytes[i] == i.byteValue()
+        }
+
```
rickle-msft: Can you add a couple more checks to assert that everything is also the same size? Like is.available == 0 and totalRead == num?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
```diff
@@ -0,0 +1,232 @@
...
+        this.data = data;
+        this.logger = logger;
+        this.lock = new ReentrantLock();
+        this.dataAvailable = lock.newCondition();
+    }
+
+    @Override
+    public int read() throws IOException {
+        return read(new byte[1], 0, 1);
+    }
```
alzimmermsft: `read()` is supposed to return the next byte read from the stream, right now this will be returning the number of bytes read from the stream which is what is the expected return for `read(byte[], int, int)`.

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
```diff
@@ -0,0 +1,232 @@
...
+    private void validateParameters(byte[] b, int off, int len) {
+        if (b == null) {
+            throw new NullPointerException("'b' cannot be null");
+        }
+        if (off < -1) {
+            throw new IndexOutOfBoundsException("'off' cannot be less than -1");
+        }
+        if (len < -1) {
+            throw new IndexOutOfBoundsException("'len' cannot be less than -1");
+        }
```
alzimmermsft: Shouldn't these be checking `< 0` and the exception stating that these cannot be `less than 0`?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryAsyncClient.java
```diff
@@ -4,10 +4,202 @@
...
- * TEMP
+ * This class provides a client that contains all operations that apply to Azure Storage Blob quick query.
+ *
+ * <p>This client offers the ability to query contents of a structured blob. </p>
+ *
+ * @see BlobQuickQueryClientBuilder
  */
 @ServiceClient(builder = BlobQuickQueryClientBuilder.class, isAsync = true)
-public class BlobQuickQueryAsyncClient {
+class BlobQuickQueryAsyncClient {
```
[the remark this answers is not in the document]
gapra-msft: It seems like the plan going forward is to add these to the blob package, and once I implement avro parsing in async we can add the async functions back.

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
gapra-msft: you're right this should be less than or equal to. My bad

gapra-msft: I'll be eventually putting out another PR to put this in the blobs package. I'll fix it then.

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryAsyncClient.java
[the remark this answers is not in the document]
gapra-msft: This was just done for a really quick effort that uses the Apache Avro parser.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ErrorReceiver.java
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.storage.common;
+
+/**
+ * An {@code ErrorReceiver} is an object that can be used to report errors on transfers. When specified on transfer
```
kasobol-msft, line 7: nit: we should refer to this as "class" not "object"

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryClient.java
```diff
@@ -4,10 +4,143 @@
...
+import java.nio.ByteBuffer;
+import java.time.Duration;
+
+import static com.azure.storage.common.implementation.StorageImplUtils.blockWithOptionalTimeout;
 
 /**
- * TEMP
+ * This class provides a client that contains all operations that apply to Azure Storage Blob quick query.
+ *
+ * <p>This client offers the ability to query contents of a structured blob. </p>
```
kasobol-msft, line 36: Do we want to mention what kind of structured blobs does this support?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
```diff
@@ -0,0 +1,207 @@
...
+                }
+                this.endRecordSeen = true;
+                break;
+            } case "progress": {
+                if (this.progressReceiver != null) {
+                    Object scanned = record.get("bytesScanned");
+                    Object total = record.get("totalBytes");
+
+                    if (checkParametersNotNull("progress", scanned, total)) {
+                        this.progressReceiver.reportProgress((Long) scanned);
```
kasobol-msft, line 156: Why do we care about "total" if we don't emit that?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
```diff
@@ -0,0 +1,207 @@
...
+                }
+                break;
+            } case "error": {
+                Object fatal = record.get("fatal");
+                Object name = record.get("name");
+                Object description = record.get("description");
+                Object position = record.get("position");
+
+                if (checkParametersNotNull("error", fatal, name, description, position)) {
+                    BlobQuickQueryError error = new BlobQuickQueryError((Boolean) fatal, name.toString(),
```
kasobol-msft, line 167: This is quite confusing. The parameter client takes is named `nonFatalErrorReceiver` but we're passing both fatals and non-fatals there.

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
```diff
@@ -0,0 +1,232 @@
...
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * An InputStream interface that subscribes to a Flux allows data to be read.
+ */
+public class FluxInputStream extends InputStream {
```
kasobol-msft, line 24: This thing doesn't seem to be qq/avro specific. Should this go to common?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/models/BlobQuickQueryDelimitedSerialization.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Defines the input or output delimited (CSV) serialization for a blob quick query request.
+ */
+public class BlobQuickQueryDelimitedSerialization extends
+    BlobQuickQuerySerialization {
+
+    private char columnSeparator;
+    private char fieldQuote;
+    private char escapeChar;
```
kasobol-msft, line 14: Is it possible that for example separator can be multi-character? If yes we might reconsider data type here to avoid future migration/break.

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/models/BlobQuickQueryError.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Creates a new BlobQuickQueryError object.
+     * @param fatal Whether or not the error is fatal.
+     * @param name The name of the error.
+     * @param description A description of the error.
+     * @param position The blob offset at which the error occurred.
+     */
+    public BlobQuickQueryError(boolean fatal, String name, String description, long position) {
+        this.fatal = fatal;
```
kasobol-msft, line 24: I wonder what's example of "fatal" that we don't want to just throw exception. If it's fatal then don't we want to stop the operation?

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/models/BlobQuickQuerySerialization.java
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Gets the record separator.
+     * @return the record separator.
+     */
+    public char getRecordSeparator() {
+        return recordSeparator;
+    }
+
+    abstract BlobQuickQuerySerialization setRecordSeparator(char recordSeparator);
+
```
kasobol-msft, line 23: By looking on derivatives of BlobQuickQuerySerialization I wonder why setRecordSeparator is abstract - both overrides are exactly the same.
Also by looking on apache avro implementation it seems that record separator is format independent property. What's the scenario where implementation of this would be different?

kasobol-msft: Maybe record separator could be independent input to QQ APIs ? It doesn't seem to be coupled to format (all avro-parsable content types must have some kind of separator I think).

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/BlobQuickQueryInputStream.java
```diff
@@ -0,0 +1,207 @@
...
+    private final ErrorReceiver<BlobQuickQueryError> errorReceiver;
+    private final ProgressReceiver progressReceiver;
+
+    /**
+     * Creates a new {@code BlobQuickQueryInputStream}
+     * @param networkStream {@link InputStream network stream}
+     * @param errorReceiver {@link ErrorReceiver<BlobQuickQueryError>}
+     * @param progressReceiver {@link ProgressReceiver}
+     */
+    BlobQuickQueryInputStream(InputStream networkStream, ErrorReceiver<BlobQuickQueryError> errorReceiver,
```
kasobol-msft, line 45: nit: naming. This class doesn't seem to assume that stream it's wrapping is coming from network. That's piece of knowledge caller of this constructor would have.

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
gapra-msft: moving it to common

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/implementation/util/FluxInputStream.java
gapra-msft: fixed in new implementation

sdk/storage/azure-storage-quickquery/src/main/java/com/azure/storage/quickquery/models/BlobQuickQueryDelimitedSerialization.java
gapra-msft: Added a comment, we can discuss better options

PR #8698, opened 2020-03-04
move dependency on azure core to a released version 1.3.0-beta.1

moderakh:
move dependency on azure core to a released version 1.3.0-beta.1 getting rid of azure dev feed. preparing for a release to maven central repo.

### DIFF
--- a/sdk/cosmos/azure-cosmos-benchmark/pom.xml
+++ b/sdk/cosmos/azure-cosmos-benchmark/pom.xml
@@ -36,19 +36,6 @@ Licensed under the MIT License.
     <jacoco.min.branchcoverage>0.01</jacoco.min.branchcoverage>
   </properties>
 
-  <repositories>
-    <repository>
-      <id>azure-sdk-for-java</id>
-      <url>https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-java/maven/v1</url>
-      <releases>
-        <enabled>true</enabled>
-      </releases>
-      <snapshots>
-        <enabled>true</enabled>
-      </snapshots>
-    </repository>
-  </repositories>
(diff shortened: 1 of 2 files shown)


JimSuplizio:
> As far as I know, this is not the correct way to make changes to pom.xml
> 
> @JimSuplizio - Since my team is not familiar with the new process, can you please suggest correct python script that needs to be run in order to change this ?

/CC @kurtzeborn @joshfree @alzimmermsft 

Couple of things here:
1. The contributing document contains how [versions and versioning](https://github.com/Azure/azure-sdk-for-java/blob/master/CONTRIBUTING.md#versions-and-versioning) should be done and has a sub-sections for [tools, versions and marking tags](https://github.com/Azure/azure-sdk-for-java/blob/master/CONTRIBUTING.md#tooling-version-files-and-marker-tags). 
2. This particular change is not going to work. Just removing the tag will trip the analyze step, every dependency and every plugin is verified to have a groupId, artifactId and versionId and each versionId has its version tag and version verified. The choices in the version tag are current/dependency and unreleased dependency (which means that you're relying on the 'current' version of a library outside of that library's build pipeline). 

At this exact moment, I am sorry to say, what you are trying to do cannot be done. I'm going to look at changing this in the near future but these changes are non-trivial and, with down with a cold, won't be in this week.

alzimmermsft:
@moderakh Is this PR still required as 1.3.0 has shipped and Cosmos can bump to/has been bumped to that version?

kushagraThapar:
> @moderakh Is this PR still required as 1.3.0 has shipped and Cosmos can bump to/has been bumped to that version?

We don't need this anymore. @moderakh is out for a couple of days, I will close this on his behalf

PR #8700, opened 2020-03-04
Fixed bug where EncryptedBlobClient would not encrypt data before upl…

gapra-msft:
…oad on new methods


sdk/storage/azure-storage-blob-cryptography/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Release History
 
+## 12.4.1 (2020-03-04)
+- Fixed a bug where EncryptedBlobClient.upload methods would not encrypt users data. Now the method throws an UnsupportedOperationException.
```
jaschrep-msft: Can you change this to call out the specific method that wasn't encrypting? It's a bad look in our changelog to say "the encryption client never encrypted" and that's not quite true.

gapra-msft: done

PR #8712, opened 2020-03-05
Prepare Azure Search for Release

alzimmermsft:
This PR prepares Azure Search for release through the following:

- Updates the version number.
- Adds a `CHANGELOG.md` explaining what is being released.
- Modifies the `README.md`.
- Changes class names based on arch board review.
- Moves package private helper classes into `implementation`.

### DIFF
--- a/eng/versioning/version_client.txt
+++ b/eng/versioning/version_client.txt
@@ -22,7 +22,7 @@ com.azure:azure-identity;1.0.3;1.1.0-beta.2
 com.azure:azure-messaging-eventhubs;5.0.2;5.1.0-beta.1
 com.azure:azure-messaging-eventhubs-checkpointstore-blob;1.0.2;1.1.0-beta.1
 com.azure:azure-messaging-servicebus;1.0.0-beta.1;1.0.0-beta.1
-com.azure:azure-search;1.0.0-beta.2;1.0.0-beta.2
+com.azure:azure-search;11.0.0-beta.1;11.0.0-beta.1
 com.azure:azure-security-keyvault-certificates;4.0.0;4.1.0-beta.1
 com.azure:azure-security-keyvault-keys;4.2.0-beta.1;4.2.0-beta.2
 com.azure:azure-security-keyvault-secrets;4.1.0;4.2.0-beta.1


alzimmermsft:
> 
> 
> @alzimmermsft, I'm good with the changes but why the version change? We released [1.0.0-beta.1](https://search.maven.org/search?q=g:com.azure%20AND%20a:azure-search) back in December and I'm curious as to why we're jumping to 11.0.0-beta.1 now.

@JimSuplizio It was decided to use 11 as the major version to make it clear that this SDK comes after the last release from track 1 which was using major version 10.

sdk/search/azure-search/README.md
```diff
@@ -132,62 +142,37 @@ Samples are explained in detail [here][samples_readme].
 
 ### General
 
-When you interact with Azure Cognitive Search using this Java client library, errors returned by the service correspond to the same HTTP status codes returned for [REST API][rest_api] requests. For example, if you try to retrieve a document that doesn't exist in your index, a `404` error is returned, indicating `Not Found`.
-
-<!-- embedme ./src/samples/java/com/azure/search/ReadmeSamples.java#L75-L83 -->
-```java
```
sima-zhu: Why do we remove this?

sdk/search/azure-search/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Release History
+
+## 11.0.0-beta.1 (Unreleased)
```
weshaggard: If you are preparing for this to be the last PR before release you may want to stamp in the date here instead of Unreleased.

sdk/search/azure-search/src/main/java/com/azure/search/implementation/util/DocumentResponseConversions.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Utility class for Document Response conversions.
+ */
+public final class DocumentResponseConversions {
+
+    /*
+     * Represents the OData context field.
+     */
+    private static final String ODATA_CONTEXT = "@odata.context";
```
brjohnstmsft: Did you make the change to emit `application/json;odata.metadata=none` for the `Accept` header of each request from `SearchIndexClient`? If so, you shouldn't need this anymore.

sima-zhu: Removed

sima-zhu: I took over Alan's PR as he is OOF, I found it is not possible to checkin his repo recently. As this is internal use class. We can leave the change for next release.

Issue #6182, opened 2019-11-06
[QUERY] AccessToken refresh - should cache expiry be made configurable?


kagkarlsson:
**Query/Question**
An `AccessToken` for a `ManagedIdentityCredential` expires 2 minutes prior to the actual expiry of the underlying token. As these jwt-tokens seem to have a lifetime of 6-8 hours, would it not make sense to have the refresh happen a bit sooner? Or make it configurable?  (to prevent disruptions if the token-refresh fails for more than 2min due to service disruptions)

***Why is this not a Bug or a feature Request?***
It may be that the answer is "Refresh of tokens via metadata service instance will never fail for more than 2 minutes"..

maggiepint:
@jianghaolu or @conniey can one of you help here?

jianghaolu:
The answer is that the HTTP request containing the token should take < 2 minutes to get to the service. When we inspect a cached access token and make sure there's more than 2 minutes life left for it, it's safe to authenticate a request with this token.

Manged identity credentials don't have refresh tokens. Only user based credentials have refresh tokens. If a refresh token expires - there's nothing we can do thus it's unnecessary to have this 2 minute check. But sharing one token implementation with access tokens is just easier :), and 2 minutes is neglectable compared to refresh tokens' usual 90-day life.

We can surely make this configurable - but I'm curious to know your use case for it.

kagkarlsson:
My point was, if we start trying to get a new token say 30 minutes prior to the expiry of the previous one, then we would be able to handle a 30-minute service disruption where getting a new token fails (for some reason)

jianghaolu:
@kagkarlsson This is a great scenario I didn't think of. I plan to get this in our next month's release. You are welcome to contribute too!

kagkarlsson:
Great! Will the PR be public?

jianghaolu:
Sorry for the delay folks. A PR is open here https://github.com/Azure/azure-sdk-for-java/pull/8715. Feel free to comment there. 
We are shipping a preview with this fix next week so further feedback can be addressed afterwards.

jianghaolu:
@kagkarlsson and anyone interested

[...]

PR #8715, opened 2020-03-05
Add configuration for refresh token before expiry

jianghaolu:
Add tests later but just opening a PR for design discussion.

Fixes https://github.com/Azure/azure-sdk-for-java/issues/6182

Sample usage:
```java
ManagedIdentityCredential credential = new ManagedIdentityCredentialBuilder()
    .clientId("user managed identity client id")
    .tokenRefreshOffset(Duration.ofMinutes(10))
    .build();
```

### DIFF
--- a/sdk/identity/azure-identity/src/main/java/com/azure/identity/CredentialBuilderBase.java
+++ b/sdk/identity/azure-identity/src/main/java/com/azure/identity/CredentialBuilderBase.java
@@ -94,6 +94,7 @@ public abstract class CredentialBuilderBase<T extends CredentialBuilderBase<T>> 
      * token takes longer than normal to get to the server.
      *
      * @param tokenRefreshOffset the duration before the actual expiry of a token to refresh it
+     * @return An updated instance of this builder with the token refresh offset set as specified.
      */
     @SuppressWarnings("unchecked")
     public T tokenRefreshOffset(Duration tokenRefreshOffset) {
--- a/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
+++ b/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
@@ -127,7 +127,6 @@ public final class IdentityClientOptions {
     }
 
     /**
-<<<<<<< HEAD
      * @return how long before the actual token expiry to refresh the token.
      */
     public Duration getTokenRefreshOffset() {
@@ -145,7 +144,9 @@ public final class IdentityClientOptions {
      * @param tokenRefreshOffset the duration before the actual expiry of a token to refresh it
      */
     public IdentityClientOptions setTokenRefreshOffset(Duration tokenRefreshOffset) {
-        this.tokenRefreshOffset = tokenRefreshOffset;
+        if (tokenRefreshOffset != null) {
+            this.tokenRefreshOffset = tokenRefreshOffset;
+        }
         return this;
     }
 
--- a/sdk/identity/azure-identity/src/test/java/com/azure/identity/ClientSecretCredentialTest.java
+++ b/sdk/identity/azure-identity/src/test/java/com/azure/identity/ClientSecretCredentialTest.java
@@ -17,6 +17,7 @@ import org.powermock.modules.junit4.PowerMockRunner;
 import reactor.core.publisher.Mono;
 import reactor.test.StepVerifier;
 
+import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.util.UUID;
@@ -59,6 +60,40 @@ public class ClientSecretCredentialTest {
             .expectNextMatches(accessToken -> token2.equals(accessToken.getToken())
                 && expiresAt.getSecond() == accessToken.getExpiresAt().getSecond())
             .verifyComplete();
+    }
+
+    @Test
+    public void testValidSecretsWithTokenRefreshOffset() throws Exception {
+        // setup
+        String secret = "secret";
+        String token1 = "token1";
+        String token2 = "token2";
+        TokenRequestContext request1 = new TokenRequestContext().addScopes("https://management.azure.com");
+        TokenRequestContext request2 = new TokenRequestContext().addScopes("https://vault.azure.net");
+        OffsetDateTime expiresAt = OffsetDateTime.now(ZoneOffset.UTC).plusHours(1);
+        Duration offset = Duration.ofMinutes(10);
+
+        // mock
+        IdentityClient identityClient = PowerMockito.mock(IdentityClient.class);
+        when(identityClient.authenticateWithClientSecret(secret, request1)).thenReturn(TestUtils.getMockAccessToken(token1, expiresAt, offset));
+        when(identityClient.authenticateWithClientSecret(secret, request2)).thenReturn(TestUtils.getMockAccessToken(token2, expiresAt, offset));
+        PowerMockito.whenNew(IdentityClient.class).withAnyArguments().thenReturn(identityClient);
+
+        // test
+        ClientSecretCredential credential = new ClientSecretCredentialBuilder()
+                .tenantId(tenantId)
+                .clientId(clientId)
+                .clientSecret(secret)
+                .tokenRefreshOffset(offset)
+                .build();
+        StepVerifier.create(credential.getToken(request1))
+                .expectNextMatches(accessToken -> token1.equals(accessToken.getToken())
+                        && expiresAt.getSecond() == accessToken.getExpiresAt().getSecond())
+                .verifyComplete();
+        StepVerifier.create(credential.getToken(request2))
+                .expectNextMatches(accessToken -> token2.equals(accessToken.getToken())
+                        && expiresAt.getSecond() == accessToken.getExpiresAt().getSecond())
+                .verifyComplete();
     }
 
     @Test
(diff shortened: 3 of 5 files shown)


sdk/identity/azure-identity/src/main/java/com/azure/identity/CredentialBuilderBase.java
```diff
@@ -68,4 +68,17 @@ public T httpPipeline(HttpPipeline httpPipeline) {
...
+
+    /**
+     * Sets the duration before the actual expiry of a token to refresh it.
+     * This is useful when network is congested and a request containing the
+     * token takes longer than normal to get to the server.
+     *
+     * @param refreshBeforeExpiry the duration before the actual expiry of a token to refresh it
+     */
+    @SuppressWarnings("unchecked")
+    public T setRefreshBeforeExpiry(Duration refreshBeforeExpiry) {
```
jianghaolu: A better name suggestion is welcome

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
```diff
@@ -21,6 +21,7 @@
     private Function<Duration, Duration> retryTimeout;
     private ProxyOptions proxyOptions;
     private HttpPipeline httpPipeline;
+    private Duration refreshBeforeExpiry = Duration.ofMinutes(2);
```
jianghaolu: Default matches the 2 minutes in AccessToken: https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/AccessToken.cs

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/MsalToken.java
```diff
@@ -22,9 +22,10 @@
      *
      * @param msalResult the raw authentication result returned by MSAL
      */
-    public MsalToken(IAuthenticationResult msalResult) {
-        super(msalResult.accessToken(), OffsetDateTime.ofInstant(msalResult.expiresOnDate().toInstant(),
-            ZoneOffset.UTC));
+    public MsalToken(IAuthenticationResult msalResult, IdentityClientOptions options) {
+        super(msalResult.accessToken(), OffsetDateTime.ofInstant(
+                msalResult.expiresOnDate().toInstant().minus(options.getRefreshBeforeExpiry()), ZoneOffset.UTC)
+            .plusMinutes(2));
```
jianghaolu: Reverts the minus 2 minutes in AccessToken: https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/credential/AccessToken.java#L22

sdk/identity/azure-identity/src/main/java/com/azure/identity/CredentialBuilderBase.java
JonathanGiles: How about `tokenRefreshOffset' or similar?

sdk/identity/azure-identity/src/main/java/com/azure/identity/CredentialBuilderBase.java
```diff
@@ -68,4 +68,17 @@ public T httpPipeline(HttpPipeline httpPipeline) {
         this.identityClientOptions.setHttpPipeline(httpPipeline);
         return (T) this;
     }
+
+    /**
+     * Sets the duration before the actual expiry of a token to refresh it.
+     * This is useful when network is congested and a request containing the
+     * token takes longer than normal to get to the server.
```
JonathanGiles: You should document what the default value is if it is not set.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -403,7 +401,9 @@
                             .useDelimiter("\\A");
                     String result = s.hasNext() ? s.next() : "";
 
-                    return SERIALIZER_ADAPTER.<MSIToken>deserialize(result, MSIToken.class, SerializerEncoding.JSON);
+                    MSIToken msiToken = SERIALIZER_ADAPTER.<MSIToken>deserialize(result, MSIToken.class, SerializerEncoding.JSON);
+                    return new AccessToken(msiToken.getToken(),
+                            msiToken.getExpiresAt().plusMinutes(2).minus(options.getRefreshBeforeExpiry()));
```
JonathanGiles: There is duplicate code here and above for creating the `AccessToken`, which is a possible source of bugs in the future if they are not kept in sync. Consider creating a method to centralise it.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/MsalToken.java
JonathanGiles: I feel like we are spreading the two minute concept out throughout the code base. Can you have it as a private static final value that is referred to everywhere?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -350,7 +346,9 @@
             Scanner s = new Scanner(connection.getInputStream(), StandardCharsets.UTF_8.name()).useDelimiter("\\A");
             String result = s.hasNext() ? s.next() : "";
 
-            return Mono.just(SERIALIZER_ADAPTER.deserialize(result, MSIToken.class, SerializerEncoding.JSON));
+            MSIToken msiToken = SERIALIZER_ADAPTER.deserialize(result, MSIToken.class, SerializerEncoding.JSON);
+            return Mono.just(new AccessToken(msiToken.getToken(),
+                    msiToken.getExpiresAt().plusMinutes(2).minus(options.getRefreshBeforeExpiry())));
```
anuchandy: This adjustment of expires time looks good to me, but I have a question on when should we acquire the token from MSI endpoint. This method seems to retrieve the token as soon as it is called. Let's say it is called at 1 pm so token acquired at 1.00 pm, now let's caller subscribed to the returned Mono at 1.10 pm, now caller is served with 10 min old token. Should we delay the token acquisition until the caller subscribes to the returned mono? (I mean putting the entire acquisition logic in Mono.defer).

jianghaolu: yes I can move it inside a Mono.defer.

sdk/identity/azure-identity/src/main/java/com/azure/identity/CredentialBuilderBase.java
jianghaolu: Using tokenRefreshOffset

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Moved to a common base class IdentityToken

sdk/identity/azure-identity/src/main/java/com/azure/identity/CredentialBuilderBase.java
jianghaolu: Documented.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
```diff
@@ -125,6 +126,29 @@ public IdentityClientOptions setHttpPipeline(HttpPipeline httpPipeline) {
...
+     * token will be considered expired at and after the time of (actual
+     * expiry - token refresh offset). The default offset is 2 minutes.
+     *
+     * This is useful when network is congested and a request containing the
+     * token takes longer than normal to get to the server.
+     *
+     * @param tokenRefreshOffset the duration before the actual expiry of a token to refresh it
+     */
+    public IdentityClientOptions setTokenRefreshOffset(Duration tokenRefreshOffset) {
+        this.tokenRefreshOffset = tokenRefreshOffset;
```
anuchandy: nit; `null` check.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
```diff
@@ -125,6 +126,29 @@ public IdentityClientOptions setHttpPipeline(HttpPipeline httpPipeline) {
         return this;
     }
 
+    /**
+<<<<<<< HEAD
```
anuchandy: nit; `<<<<<<< HEAD` from conflict resoln.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityToken.java
```diff
@@ -0,0 +1,24 @@
...
+public class IdentityToken extends AccessToken {
+    /**
+     * Creates an identity token instance.
+     *
+     * @param token the token string.
+     * @param expiresAt the expiration time.
+     * @param options the identity client options.
+     */
+    public IdentityToken(String token, OffsetDateTime expiresAt, IdentityClientOptions options) {
+        super(token, expiresAt.plusMinutes(2).minus(options.getTokenRefreshOffset()));
```
anuchandy: nit; `Object.requireNonNull` checks.

anuchandy: discussed with Jinaghao offline, these are impl classes and these values are derived from service returned response. Hence we don't have to check here given we shouldn't test service behavior.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
jianghaolu: Removed.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
jianghaolu: Added a nullcheck

PR #8760, opened 2020-03-05
reduced response header copying

moderakh:
Prior to this PR:
1. Response Headers were constructed in the TransportClient
2. StoreReponse copied the response headers from 1.
3. DocumentServiceResponse created another copy of response headers from 2.

after this PR.
2 and 3 will share the same reference to the response headers, avoiding copying response headers multiple times.


moderakh:
as per my testing ironically this change lowered the throughput somewhat ~5%. so not moving forward with this.

PR #8762, opened 2020-03-05
Add Azure Cli credentials to Identity

g2vinay:


sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -122,6 +140,97 @@
...
+     * @param request the details of the token request
+     * @return a Publisher that emits an AccessToken
+     */
+    public Mono<AccessToken> authenticateWithAzureCli(TokenRequestContext request) {
+        String azCommand = "az account get-access-token --output json --resource ";
+
+        StringBuilder command = new StringBuilder();
+        command.append(azCommand);
+        String scopes = ScopeUtil.scopesToResource(request.getScopes());
+        command.append(scopes);
```
schaabs: Here we need to validate the scope passed in doesn't contain any characters which could cause us to execute something unintended. See https://github.com/Azure/azure-sdk-for-net/blob/23cc9455cf501d6a65ce8faaedb27242b27c3b51/sdk/identity/Azure.Identity/src/ScopeUtilities.cs#L46-L54

PR #8772, opened 2020-03-06
Regen from new swagger and address some API changes

sima-zhu:
We are adopting new swagger changes. This is initial change from auto-gen. Will get sort of the everything in the following commits.

### DIFF
--- a/sdk/search/azure-search/src/samples/java/com/azure/search/AutoCompleteExample.java
+++ b/sdk/search/azure-search/src/samples/java/com/azure/search/AutoCompleteExample.java
@@ -22,7 +22,7 @@ public class AutoCompleteExample {
 
     /**
      * From the Azure portal, get your Azure Cognitive Search service URL and API key,
-     * and set aathe values of these environment variables:
+     * and set the values of these environment variables:
      */
     private static final String ENDPOINT = Configuration.getGlobalConfiguration().get("AZURE_COGNITIVE_SEARCH_ENDPOINT");
     private static final String API_KEY = Configuration.getGlobalConfiguration().get("AZURE_COGNITIVE_SEARCH_API_KEY");


sdk/search/azure-search/src/main/java/com/azure/search/implementation/SynonymMapsImpl.java
```diff
@@ -60,28 +63,33 @@ public SynonymMapsImpl(SearchServiceRestClientImpl client) {
...
 
         @Delete("synonymmaps('{synonymMapName}')")
         @ExpectedResponses({204, 404})
-        Mono<SimpleResponse<Void>> delete(@PathParam("synonymMapName") String synonymMapName, @HostParam("searchServiceName") String searchServiceName, @HostParam("searchDnsSuffix") String searchDnsSuffix, @QueryParam("api-version") String apiVersion, @HeaderParam("client-request-id") UUID clientRequestId, @HeaderParam("If-Match") String ifMatch, @HeaderParam("If-None-Match") String ifNoneMatch, Context context);
+        @UnexpectedResponseExceptionType(SearchErrorException.class)
+        Mono<Response<Void>> delete(@PathParam("synonymMapName") String synonymMapName, @HostParam("endpoint") String endpoint, @QueryParam("api-version") String apiVersion, @HeaderParam("x-ms-client-request-id") UUID xMsClientRequestId, @HeaderParam("If-Match") String ifMatch, @HeaderParam("If-None-Match") String ifNoneMatch, Context context);
```
alzimmermsft: Any reason that just delete changed from `SimpleResponse<Void>` to `Response<Void>`?

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchResult.java
```diff
@@ -46,7 +46,7 @@
      * @return the additionalProperties value.
      */
     @JsonIgnore
-    public Document getDocument() {
+public Document getDocument() {
```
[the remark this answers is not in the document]
brjohnstmsft: Probably the result of a regex transform.

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchDocumentsResult.java
```diff
@@ -111,7 +111,7 @@ public Double getCoverage() {
      *
      * @return the nextPageParameters value.
      */
-    public SearchRequest getNextPageParameters() {
+    SearchRequest getNextPageParameters() {
```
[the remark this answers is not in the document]
brjohnstmsft: Do customers need access to this? I thought the paging mechanism was supposed to encapsulate this detail.

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchResult.java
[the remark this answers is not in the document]
sima-zhu: Yes, Haven't got a chance to work on that. Will work on a fix after release.

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchErrorException.java
```diff
@@ -0,0 +1,41 @@
...
+
+package com.azure.search.models;
+
+import com.azure.core.exception.HttpResponseException;
+import com.azure.core.http.HttpResponse;
+
+/**
+ * Exception thrown for an invalid response with SearchError information.
+ */
+public final class SearchErrorException extends HttpResponseException {
```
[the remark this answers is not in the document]
sima-zhu: It is updated. Thanks!

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchDocumentsResult.java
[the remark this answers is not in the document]
sima-zhu: This is internal used method. However, the SearchPageResponse is not in the same folder as SearchDocumentsResult.java. I reverted the change as it cannot be called if put it package-scope. Need to find a way to solve this.

sdk/search/azure-search/src/main/java/com/azure/search/implementation/SynonymMapsImpl.java
sima-zhu: It is newly introduced feature in azure core. My understanding here we do not need to check Value in simpleResponse. So Response<Void> is enough and simple for use.

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
```diff
@@ -468,7 +465,7 @@ SearchPagedFlux search(String searchText, SearchOptions searchOptions, RequestOp
             return restClient.documents()
                 .getWithRestResponseAsync(key, selectedFields, requestOptions, context)
                 .map(res -> {
-                    Document doc = res.getValue();
+                    Document doc = new Document(res.getValue());
```
brjohnstmsft: Isn't this going to copy the entire map?

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchErrorException.java
```diff
@@ -0,0 +1,41 @@
...
+     * @param message the exception message or the response content if a message is not available.
+     * @param response the HTTP response.
+     * @param value the deserialized response value.
+     */
+    public SearchErrorException(String message, HttpResponse response, SearchError value) {
+        super(message, response, value);
+    }
+
+    @Override
+    public SearchError getValue() {
```
brjohnstmsft: I had to add `SearchError` to the Swagger for the sake of JS/TS, but it's my understanding that this is a subset of a well-defined [error contract for Microsoft REST APIs](https://github.com/Microsoft/api-guidelines/blob/master/Guidelines.md#7102-error-condition-responses). Do we actually need to expose this, or is there some Azure Core type for error payloads that overlaps `SearchError`?

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchResult.java
```diff
@@ -45,8 +44,7 @@
      *
      * @return the additionalProperties value.
      */
-    @JsonIgnore
-    public Document getDocument() {
+    public Map<String, Object> getDocument() {
```
brjohnstmsft: This is a public contract -- Shouldn't the return type be `SearchDocument` (renamed from `Document`)?

sdk/search/azure-search/src/main/java/com/azure/search/models/SuggestResult.java
```diff
@@ -33,7 +34,7 @@
      *
      * @return the additionalProperties value.
      */
-    public Document getDocument() {
+    public Map<String, Object> getDocument() {
```
brjohnstmsft: Same question -- Shouldn't this be `SearchDocument` instead of just `Map`?

sdk/search/azure-search/src/main/java/com/azure/search/models/VisualFeature.java
```diff
@@ -13,15 +13,25 @@
  * Defines values for VisualFeature.
  */
 public enum VisualFeature {
```
brjohnstmsft: I think this is one Swagger version behind, because we recently removed two of these properties.

sdk/search/azure-search/src/samples/java/com/azure/search/SearchAsyncWithFullyTypedDocumentsExample.java
```diff
@@ -49,7 +49,7 @@ public static void main(String[] args) {
         SearchPagedFlux results = searchClient.search("searchText");
         results
             .subscribe(item -> {
-                Document document = item.getDocument();
+                Document document = new Document(item.getDocument());
```
brjohnstmsft: Customers shouldn't have to create `Document` instances explicitly.

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
sima-zhu: Yes, it will copied the map.

Here is the constructor JavaDoc
```
    /**
     * Constructs a new {@code HashMap} with the same mappings as the
     * specified {@code Map}.  The {@code HashMap} is created with
     * default load factor (0.75) and an initial capacity sufficient to
     * hold the mappings in the specified {@code Map}.
     *
     * @param   m the map whose mappings are to be placed in this map
     * @throws  NullPointerException if the specified map is null
     */
    public HashMap(Map<? extends K, ? extends V> m) {
```

brjohnstmsft: That doesn't seem great for performance.

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchErrorException.java
sima-zhu: It's actually useful to have getCode, getMessage etc APIs in SearchError. Otherwise we have to parse the error message.
Autorest will generate exception depending on the name default error (`SearchError` -> `SearchErrorException`) from Swagger.
```
"default": {
            "description": "Error response.",
            "schema": {
              "$ref": "#/definitions/SearchError"
            }
          }
```

In storage, we use the same structure. 
That is expected. What error type do you expect Java use, HttpResponseException?

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchResult.java
sima-zhu: Yes. This is manually craft code. So Alan's PR will address this.

sdk/search/azure-search/src/main/java/com/azure/search/models/SuggestResult.java
sima-zhu: Search result is auto-gen code. As @alzimmermsft brought up yesterday. Our auto-gen code is better not to depend on any manually craft APIs. That's why I have changed the return type here. Do we still want the Document here?

sdk/search/azure-search/src/main/java/com/azure/search/models/VisualFeature.java
sima-zhu: Will generate again. Thanks!

sdk/search/azure-search/src/samples/java/com/azure/search/SearchAsyncWithFullyTypedDocumentsExample.java
sima-zhu: Depend on whether we want to have auto-gen code depends on manual codes.

sdk/search/azure-search/src/main/java/com/azure/search/models/SearchResult.java
sima-zhu: Made the changes.

sdk/search/azure-search/src/main/java/com/azure/search/models/SuggestResult.java
sima-zhu: Revert it back

sdk/search/azure-search/src/samples/java/com/azure/search/AutoCompleteExample.java
```diff
@@ -22,7 +22,7 @@
 
     /**
      * From the Azure portal, get your Azure Cognitive Search service URL and API key,
-     * and set the values of these environment variables:
+     * and set aathe values of these environment variables:
```
brjohnstmsft: Typo

sdk/search/azure-search/src/samples/java/com/azure/search/SearchAsyncWithFullyTypedDocumentsExample.java
brjohnstmsft: It should depend on what's best for the user experience.

sdk/search/azure-search/src/samples/java/com/azure/search/AutoCompleteExample.java
sima-zhu: Done.

sdk/search/azure-search/src/main/java/com/azure/search/SearchIndexAsyncClient.java
JonathanGiles: Do we need Document to extend from HashMap, or is it sufficient that it have a map-like API?

brjohnstmsft: @JonathanGiles Strictly speaking no. I think that was done for implementation convenience. It also means that `SearchDocument` (formerly `Document`) implements `Map`. The only scenario I can imagine where that's necessary is interop with other JVM languages.

PR #8774, opened 2020-03-06
Reliable download fix

rickle-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/ReliableDownload.java
```diff
@@ -67,15 +70,16 @@ BlobDownloadHeaders getDeserializedHeaders() {
...
             : applyReliableDownload(rawResponse.getValue(), -1, options);
 
         return value.switchIfEmpty(Flux.just(ByteBuffer.wrap(new byte[0])));
     }
 
     private Flux<ByteBuffer> tryContinueFlux(Throwable t, int retryCount, DownloadRetryOptions options) {
-        // If all the errors are exhausted, return this error to the user.
-        if (retryCount > options.getMaxRetryRequests() || !(t instanceof IOException)) {
+        // If all the errors are exhausted or the error is not retryable, return this error to the user.
+        if (retryCount >= options.getMaxRetryRequests()
```
anuchandy: So the `retryCount` is global for the `Download` not per `read()` invocation?

rickle-msft: It is as I've coded it here. I picked this a little bit arbitrarily. Since this was easier/faster and both of us seemed a little unsure that one option was strictly better than the other, I figured sticking with the existing plan would make it more likely to be ready to ship next week. We also see these stale streams pretty rarely, and I think even one retry per download operation is enough in the vast vast majority of cases. If you feel strongly that resetting the count on each successful item is better or if we see some Timeouts still trickling through, I can update it in a later release.

anuchandy: sounds good.

PR #8794, opened 2020-03-06
cache range in PartitionKeyRange to avoid Range parsing json on every query

moderakh:
cache Range in PartitionKeyRange to avoid json parsing range on each query

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/PartitionKeyRange.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/PartitionKeyRange.java
@@ -93,7 +93,7 @@ public class PartitionKeyRange extends Resource {
     public Range<String> toRange() {
         if (this.range == null) {
             synchronized (this) {
-                if (range != null) {
+                if (range == null) {
                     this.range = new Range<String>(this.getMinInclusive(), this.getMaxExclusive(), true, false);
                 }
             }


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/PartitionKeyRange.java
```diff
@@ -76,20 +78,28 @@ public String getMinInclusive() {
...
-    public void setMaxExclusive(String maxExclusive) {
+    private void setMaxExclusive(String maxExclusive) {
         BridgeInternal.setProperty(this, "maxExclusive", maxExclusive);
     }
 
     public Range<String> toRange() {
-        return new Range<String>(this.getMinInclusive(), this.getMaxExclusive(), true, false);
+        if (this.range == null) {
+            synchronized (this) {
+                if (range != null) {
```
simplynaveen20: Should'n it be range  == null

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/PartitionKeyRange.java
```diff
@@ -76,20 +78,28 @@ public String getMinInclusive() {
...
 
-    public void setMaxExclusive(String maxExclusive) {
+    private void setMaxExclusive(String maxExclusive) {
         BridgeInternal.setProperty(this, "maxExclusive", maxExclusive);
     }
 
     public Range<String> toRange() {
-        return new Range<String>(this.getMinInclusive(), this.getMaxExclusive(), true, false);
+        if (this.range == null) {
+            synchronized (this) {
```
simplynaveen20: Loud thinking  - If toRange () method return same value for a PartitionKeyRange object , I am seeing a trade off between locking (making sure only one thread create object ) vs time saving if we dont do synchronized and letting two thread create object if they simultaneously comes in this block

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/PartitionKeyRange.java
moderakh: this was on earlier commit. already fixed.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/PartitionKeyRange.java
moderakh: as per Happen-Before relationship in java. To allow one thread sees the effect of the other, either this field has to be volatile or we should do synchronized lock.

as synchrnozied is only used on set assignment it doesn't have any cost on the get, however there is always a cost associated with get in volatile. synchronized doesn't add any cost on the get which is the intention.

simplynaveen20: As i said in this particular scenario as both thread end up creating same value we can deviate from Happens-Before , how ever as you mentioned synchronized  doesn't add any cost then this is good . All set

Issue #6558, opened 2019-11-26
Client builders should accept another client for initialization


rickle-msft:
Coming from #6536, builders should be able to accept another client and extract the pipeline and client for initialization. This helps conveniently construct clients that don't share package internals or inheritance hierarchies.

SukruthKS:
When I opened #6536, it was in the context of `EncryptedBlobClient`.  It'd be great if builders for other specialized clients like `BlobLeaseClient` can also get this capability. Looks like this feature request is already generalized to apply to all builders, but just wanted to point out in case it was going to be added only to `EncryptedBlobClientBuilder` :)

SukruthKS:
Never mind. `BlobLeaseClientBuilder` already has this.

gapra-msft:
Hi @SukruthKS, 

Just to help clarify things, are you only concerned about caching credentials when needing to use the EncryptedBlobClientBuilder or are you also concerned about the other optional pipeline settings that are configurable from the builder? 

Would it be sufficient if we could provide you with a method on the builder like `encryptedBuilder.credential(existingClient.getHttpPipeline()) `

This would be a method that sets the credential found in the httpPipeline provided. Then you would have to go in and configure optional policies if need be.

SukruthKS:
We are not setting any pipeline policies currently so getting just the creds from an existing pipeline should be sufficient. However, I could see us adding pipeline policies in the future for any custom requirements and having the ability to generate EncryptedBlobClient from that pipeline would be useful.

gapra-msft:
Hi @SukruthKS I recently merged a PR that should address this issue. It should go out in our next release.

SukruthKS:
@gapra-msft when I create `EncryptedBlobClient` from an existing `BlobClient` instance as shown below, it doesn't seem to copy over `CustomerProvidedKey` set in the blob client and I don't see an option to set it explicitly on `EncryptedBlobClientBuilder` either. Is there a way to do it?

```
EncryptedBlobClient encryptedBlobClient = new EncryptedBlobClientBuilder()
    .blobClient(existingBlobClient)
    .buildEncryptedBlobClient;
```

gapra-msft:
Hi @SukruthKS 

Sorry about that. Originally I didn't copy over the CustomerProvidedKey since I didn't realize people would want to double encrypt their data. We will work on adding that back in.

Thanks

rickle-msft:
[...]

PR #8825, opened 2020-03-06
EncryptedBlobClientBuilder enhancements

gapra-msft:
Resolves https://github.com/Azure/azure-sdk-for-java/issues/6558
Added CPK support
Fixed bug where decryption policy would not be added on the pipeline method

### DIFF
--- a/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
+++ b/sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
@@ -472,9 +472,10 @@ public final class EncryptedBlobClientBuilder {
     }
 
     /**
-     * Sets the {@link HttpPipeline} to use for the service client, and adds a decryption policy.
+     * Sets the {@link HttpPipeline} to use for the service client, and adds a decryption policy if one is not present.
      *
-     * If {@code pipeline} is set, all other settings are ignored, aside from {@link #endpoint(String) endpoint}.
+     * If {@code pipeline} is set, all other settings are ignored, aside from {@link #endpoint(String) endpoint}
+     * and {@link #customerProvidedKey(CustomerProvidedKey) customer provided key}.
      *
      * <p>Use this method after setting the key in {@link #key(AsyncKeyEncryptionKey, String) key} and keyResolver in
      * {@link #keyResolver(AsyncKeyEncryptionKeyResolver)}.</p>
@@ -491,11 +492,16 @@ public final class EncryptedBlobClientBuilder {
         HttpPipeline pipeline = null;
         if (httpPipeline != null) {
             List<HttpPipelinePolicy> policies = new ArrayList<>();
-            policies.add(new BlobDecryptionPolicy(keyWrapper, keyResolver));
+            // Check that BlobDecryptionPolicy not already present while copying them over to a new policy list.
+            boolean decryptionPolicyPresent = false;
             for (int i = 0; i < httpPipeline.getPolicyCount(); i++) {
                 HttpPipelinePolicy currPolicy = httpPipeline.getPolicy(i);
+                decryptionPolicyPresent |= currPolicy instanceof BlobDecryptionPolicy;
                 policies.add(currPolicy);
-
+            }
+            // If a decryption policy is not already present, add it to the front.
+            if (!decryptionPolicyPresent) {
+                policies.add(0, new BlobDecryptionPolicy(keyWrapper, keyResolver));
             }
             pipeline = new HttpPipelineBuilder()
                 .httpClient(httpPipeline.getHttpClient())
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
```diff
@@ -93,13 +94,17 @@
...
      */
     EncryptedBlobAsyncClient(HttpPipeline pipeline, String url, BlobServiceVersion serviceVersion, String accountName,
-        String containerName, String blobName, String snapshot, AsyncKeyEncryptionKey key, String keyWrapAlgorithm) {
-        super(pipeline, url, serviceVersion, accountName, containerName, blobName, snapshot, null /*cpk*/,
-            null /*encryption scope*/);
+        String containerName, String blobName, String snapshot, CpkInfo customerProvidedKey,
+        AsyncKeyEncryptionKey key, String keyWrapAlgorithm) {
+        super(pipeline, url, serviceVersion, accountName, containerName, blobName, snapshot, customerProvidedKey,
+            null);
+        // Note encryption scope is always null when passed in from the client builder.
```
gapra-msft: remove this comment

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -469,19 +472,37 @@ public EncryptedBlobClientBuilder retryOptions(RequestRetryOptions retryOptions)
     }
 
     /**
-     * Sets the {@link HttpPipeline} to use for the service client.
+     * Sets the {@link HttpPipeline} to use for the service client, and adds a decryption policy.
```
rickle-msft: If one is not present

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -469,19 +472,37 @@ public EncryptedBlobClientBuilder retryOptions(RequestRetryOptions retryOptions)
...
      *
      * If {@code pipeline} is set, all other settings are ignored, aside from {@link #endpoint(String) endpoint}.
```
rickle-msft: cpk isn't ignored, right? Maybe for a bit of future-proofing, we should specify that any options used to configure a pipeline are ignored?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -469,19 +472,37 @@ public EncryptedBlobClientBuilder retryOptions(RequestRetryOptions retryOptions)
...
         if (this.httpPipeline != null && httpPipeline == null) {
             logger.info("HttpPipeline is being set to 'null' when it was previously configured.");
         }
+        checkValidEncryptionParameters();
+
+        HttpPipeline pipeline = null;
+        if (httpPipeline != null) {
+            List<HttpPipelinePolicy> policies = new ArrayList<>();
+            policies.add(new BlobDecryptionPolicy(keyWrapper, keyResolver));
+            for (int i = 0; i < httpPipeline.getPolicyCount(); i++) {
```
rickle-msft: Can we check to ensure one isn't already present?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -512,6 +552,9 @@ public EncryptedBlobClientBuilder serviceVersion(BlobServiceVersion version) {
      * <p>If {@code pipeline} is set, all other settings are ignored, aside from {@link #endpoint(String) endpoint} and
      * {@link #serviceVersion(BlobServiceVersion) serviceVersion}.</p>
      *
+     * <p>Note that this method does not copy over the {@link CustomerProvidedKey} and encryption scope properties
```
rickle-msft: Can you add "for security reasons" to hopefully preempt some customers who may ask for this?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
rickle-msft: It's also not copied since it isn't a pipeline option. I guess it feels a little weird to stay stuff outside the pipeline won't be copied, but the pipeline is a weird concept for people, so maybe it's best to be explicit.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
gapra-msft: will do

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
gapra-msft: done

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
gapra-msft: done

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
gapra-msft: I think they might get confused cause they dont know what options are pipeline and what arent, I'll just say except CPK

Issue #6536, opened 2019-11-25
[QUERY] Client side encryption using `azure-storage-blob-cryptography` library


SukruthKS:
**Query/Question**
We are currently using client side encryption functionality provided in v8 SDK. I'm upgrading our application to use v12 and I have a few questions/issues:

- Passing a key to `EncryptedBlobClient` requires implementing `AsyncKeyEncryptionKey` interface. Are there any existing implementations I can use? With v8, we were using `SymmetricKey` class provided in `com.microsoft.azure.keyvault.cryptography`.

- `EncryptedBlobClient` does not support uploading a blob from a stream like in `BlockBlobClient`. It only supports uploading from a file. Is there a way to do it using a stream? I'd prefer not to create files on disk just to do this.

- `EncryptedBlobClient` cannot be created from `BlobContainerClient` like other clients (e.g., `BlockBlobClient`). In our application, we'd initialize `BlobServiceClient` once and derive other clients like container or blob depending on the access needs. Since `EncryptedBlobClient` requires full standalone construction, I either need to cache credentials (all supported types as any of them could be used) or always create `EncryptedBlobClient` when creating `BlobServiceClient` even if client side encryption is not going to be used. 

- Client side encryption feature was seamlessly integrated in v8 compared to v12. Depending on whether a blob needs to be encrypted, I'd have to use two different clients in v12 whereas in v8 I just have to set `BlobEncryptionPolicy` in upload options.

***Why is this not a Bug or a feature Request?***
NA

**Setup (please complete the following information if applicable):**
 - OS: Linux
 - IDE : IntelliJ
 - Version of the Library used: 12.0.0
 
 **Information Checklist**
 Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Query Added
- [x] Setup information Added

rickle-msft:
Hi, @SukruthKS. Responding to your points in order:

- The keyvault.keys.cryptography package has a KeyEncryptionKeyBuilder that you can use to generate implementations of the type to pass to the encrypted blob builder. The builder requests an async key encryption key, but only because we use one internally. I don't think you should really have to deal with any async objects because of this interface, but please let me know if I have overlooked something. 
- I am planning on doing some refactor work to BlobOutputStream that I think should enable this scenario. We are aware of the need and are tracking it, but it is not currently supported. You can track this at #5715
[...]

PR #8826, opened 2020-03-06
Changed EncryptedBlobClient to properly override upload method

rickle-msft:
Addresses the most recent discussion on #6536


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClient.java
```diff
@@ -109,6 +116,32 @@ public BlobOutputStream getBlobOutputStream(ParallelTransferOptions parallelTran
...
+     */
+    @Override
+    public void uploadWithResponse(InputStream data, long length, ParallelTransferOptions parallelTransferOptions,
+        BlobHttpHeaders headers, Map<String, String> metadata, AccessTier tier, BlobRequestConditions requestConditions,
+        Duration timeout, Context context) {
+        final ParallelTransferOptions validatedParallelTransferOptions =
+            ModelHelper.populateAndApplyDefaults(parallelTransferOptions);
+
+        // BlobOutputStream passes to buffered upload, which has logic for switching based on stream size
+        BlobOutputStream blobOutputStream = BlobOutputStream.blockBlobOutputStream(encryptedBlobAsyncClient,
```
gapra-msft: don't feel two strongly about this but could we make this logic protected in regular Blob so its not duplicated?

rickle-msft: Good call!

rickle-msft:
@JonathanGiles are you ok with this refactor? It adds a protected method to avoid some code duplication.

JonathanGiles:
I would prefer to avoid the protected API if possible. What is the impact of duplication?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
```diff
@@ -167,24 +168,41 @@ public void upload(InputStream data, long length, boolean overwrite) {
...
+     *
+     * @param parallelTransferOptions The options passed to upload. Pre-verified.
+     * @param headers The headers passed to upload.
+     * @param metadata The metadata passed to upload.
+     * @param tier The tier passed to upload.
+     * @param requestConditions The request conditions passed to upload.
+     *
+     * @return The {@link BlobOutputStream} initialized with the appropriate client.
+     */
+    protected BlobOutputStream getBlobOutputStreamInternal(ParallelTransferOptions parallelTransferOptions,
```
srnagar: Can this be private instead? You can declare another private method in `EncryptedBlobClient` instead of overriding.

PR #8833, opened 2020-03-07
Nio delete

rickle-msft:
There are two parts to this PR.
1. Adding the delete api and relevant tests
2. A reasonably large refactor of tests that I promised I would fix when I merged the last PR.

### DIFF
--- a/sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
+++ b/sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
@@ -213,6 +213,8 @@ public final class AzureFileSystemProvider extends FileSystemProvider {
     }
 
     /**
+     * Creates a new directory at the specified path.
+     * <p>
      * The existence of a directory in the {@code AzureFileSystem} is defined on two levels. <i>Weak existence</i> is
      * defined by the presence of a non-zero number of blobs prefixed with the directory's path. This concept is also
      * known as a  <i>virtual directory</i> and enables the file system to work with containers that were pre-loaded
@@ -358,7 +360,9 @@ public final class AzureFileSystemProvider extends FileSystemProvider {
     }
 
     /**
-     * As noted by the NIO docs, this method is not atomic. It is possible to delete a file in use by another process,
+     * Deletes the specified resource.
+     * <p>
+     * This method is not atomic. It is possible to delete a file in use by another process,
      * and doing so will not immediately invalidate any channels open to that file--they will simply start to fail.
      * Root directories cannot be deleted even when empty.
      *
@@ -386,12 +390,17 @@ public final class AzureFileSystemProvider extends FileSystemProvider {
         try {
             blobClient.delete();
         } catch (BlobStorageException e) {
+            if (e.getErrorCode().equals(BlobErrorCode.BLOB_NOT_FOUND)) {
+                throw Utility.logError(logger, new NoSuchFileException(path.toString()));
+            }
             throw Utility.logError(logger, new IOException(e));
         }
     }
 
     /**
-     * As stated in the nio docs, this method is not atomic. More specifically, the checks necessary to validate the
+     * Copies the resource at the source location to the destination.
+     * <p>
+     * This method is not atomic. More specifically, the checks necessary to validate the
      * inputs and state of the file system are not atomic with the actual copying of data. If the copy is triggered,
      * the copy itself is atomic and only a complete copy will ever be left at the destination.
      *
--- a/sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/APISpec.groovy
+++ b/sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/APISpec.groovy
@@ -523,8 +523,9 @@ class APISpec extends Specification {
     }
 
     def checkBlobIsDir(BlobClient blobClient) {
-        return blobClient.getPropertiesWithResponse(null, null, null).getValue().getMetadata()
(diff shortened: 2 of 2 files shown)


sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzurePath.java
```diff
@@ -531,10 +555,24 @@ BlobClient toBlobClient() throws IOException {
 
         String blobName = this.withoutRoot();
         if (blobName.isEmpty()) {
+<<<<<<< HEAD
```
rickle-msft: Oops!

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -363,11 +358,36 @@ BlobContainerClient getContainerClient(BlobClient client) {
     }
 
     /**
+     * As noted by the NIO docs, this method is not atomic. It is possible to delete a file in use by another process,
+     * and doing so will not immediately invalidate any channels open to that file--they will simply start to fail.
+     * Root directories cannot be deleted even when empty.
```
alzimmermsft: I don't think we need to call out what the NIO documents state explicitly, let's just include it as a statement in our docs. Would also like to change the first line to call out what exactly this method does, such as `This API deletes the specified path/file`, then adding the ancillary info like this isn't atomic, what can't be deleted, etc.

```suggestion
     * Deletes the specified path.
     * <p>
     * This method is not atomic, it is possible to delete a file in use by another process. Open channels to the file won't immediately invalidate, they will simply start to fail.
     * <p>
     * Root directories cannot be deleted even when empty.
```

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -363,11 +358,36 @@ BlobContainerClient getContainerClient(BlobClient client) {
...
+        }
+        if (dirStatus.equals(DirectoryStatus.NOT_EMPTY)) {
+            throw Utility.logError(logger, new DirectoryNotEmptyException(path.toString()));
+        }
 
+        // After all validation has completed, delete the resource.
+        try {
+            blobClient.delete();
+        } catch (BlobStorageException e) {
+            throw Utility.logError(logger, new IOException(e));
```
alzimmermsft: Do we want to log and throw an exception if the exception is that the blob doesn't exist? I'm good with either, just a theoretical question on what delete on something that doesn't exist means.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
gapra-msft: If we really want to call out the NIO doc, we could also just link it

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
gapra-msft: Won't checkDirStatus do the blob doesn't exist validation for us? It gets the container client and lists paths with path = blob name and returns diff codes depending on the number of elements underneath it

gapra-msft: I suppose something could also happen between that check and the delete

sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/APISpec.groovy
```diff
@@ -481,10 +488,45 @@ class APISpec extends Specification {
...
+        return pathStr
+    }
+
+    def putDirectoryBlob(BlockBlobClient blobClient) {
+        blobClient.commitBlockListWithResponse(Collections.emptyList(), null,
+            [(AzureFileSystemProvider.DIR_METADATA_MARKER): "true"], null, null, null, null)
+    }
+
+    def checkBlobIsDir(BlobClient blobClient) {
+        return blobClient.getPropertiesWithResponse(null, null, null).getValue().getMetadata()
```
gapra-msft: You could also make sure the value here is true just to prevent accidental fake news. One day the service could decide they want to set this to false for non-dir like blobs

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
rickle-msft: Below the javadocs I've written, I always inherit the javadocs from the parent, so all the extra info should be there anyway. I'll update the docs to remove the NIO reference and change the first sentence.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
rickle-msft: It's true we do check for no such blob based on the dir status earlier and also true that something else could happen in between. I added a specific case for a BlobNotFound error code and throw a NoSuchFileException in that case as well. 

I would prefer to throw in the case of no such file because the docs say we can, so it's more informative and because there's a deleteIfExists api that customers can call if they don't want the exception behavior.

PR #8847, opened 2020-03-09
Public api duration change

kushagraThapar:
* Updated public APIs with Duration
* Updated documentation with word "duration"

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/model/ConsistencyPolicy.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/model/ConsistencyPolicy.java
@@ -127,7 +127,7 @@ public final class ConsistencyPolicy extends JsonSerializable {
         if (maxStalenessInterval == null) {
             throw new IllegalArgumentException("maxStalenessInterval should not be null");
         }
-        super.set(Constants.Properties.MAX_STALENESS_INTERVAL_IN_SECONDS, maxStalenessInterval.toSeconds());
+        super.set(Constants.Properties.MAX_STALENESS_INTERVAL_IN_SECONDS, maxStalenessInterval.getSeconds());
         return this;
     }
 }


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientException.java
```diff
@@ -193,13 +194,13 @@ void setError(CosmosError cosmosError) {
...
-     * Gets the recommended time interval after which the client can retry failed
+     * Gets the recommended time duration after which the client can retry failed
      * requests
      *
-     * @return the recommended time interval after which the client can retry failed
+     * @return the recommended time duration after which the client can retry failed
      * requests.
      */
-    public long getRetryAfterInMilliseconds() {
+    public Duration getRetryAfterDuration() {
```
moderakh: Duration is not used as a suffix in ConnectionPolicy#setRequestTimeout() and getter.

any reason we are using it here?

what are your thought on the method naming?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/model/ConsistencyPolicy.java
```diff
@@ -105,21 +107,21 @@ public ConsistencyPolicy setMaxStalenessPrefix(int maxStalenessPrefix) {
      *
      * @return the max staleness prefix.
      */
-    public int getMaxStalenessIntervalInSeconds() {
+    public Duration getMaxStalenessIntervalInSeconds() {
```
moderakh: hmm, do we want "InSeconds()" as suffix in the method name?

the time resolution for this method is seconds. If we are using Duration here we should update the documentation mentioning that the resolution is in seconds.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/model/ConsistencyPolicy.java
```diff
@@ -105,21 +107,21 @@ public ConsistencyPolicy setMaxStalenessPrefix(int maxStalenessPrefix) {
...
 
     /**
      * Sets the in bounded staleness consistency, the maximum allowed staleness in terms time interval.
      *
-     * @param maxStalenessIntervalInSeconds the max staleness interval in seconds.
+     * @param maxStalenessIntervalInSeconds the max staleness duration.
      * @return the ConsistenctPolicy.
      */
-    public ConsistencyPolicy setMaxStalenessIntervalInSeconds(int maxStalenessIntervalInSeconds) {
+    public ConsistencyPolicy setMaxStalenessIntervalInSeconds(Duration maxStalenessIntervalInSeconds) {
```
moderakh: the same.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientException.java
kushagraThapar: The reason I added Duration here is because `getRetryAfter` seems very general. It was not clear to me that it returns duration to retry after. So I added that. I can change it to `getRetryAfter` if you think that is much better.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/model/ConsistencyPolicy.java
kushagraThapar: We don't want inSeconds() as the suffix in the method name. I will remove it.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/model/ConsistencyPolicy.java
kushagraThapar: agreed.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientException.java
moderakh: ok makes sense, `getRetryAfterDuration()` looks good

PR #8866, opened 2020-03-09
Azure sdk/increment package version core 282785

JimSuplizio:
I took the automated [version bump PR](https://github.com/Azure/azure-sdk-for-java/pull/8834) created by the release pipeline and added a commit it to remove the unreleased dependency entries for azure-core and azure-core-http-netty which were just released. The second commit I added was to fix azure-data-appconfiguration's pom.xml file which was incorrectly updated to point at the current version core.


JimSuplizio:
> Search looks good to me. For other library depends on azure-core, and azure-core-netty previous version. Do we want to bump to new release version?

With the exception when core is released as a beta and we don't move the dependency version forward, we should be moving the dependency version forward after we release a GA version. The only differences between the versions should be additive.

PR #8879, opened 2020-03-09
Update AMQP's core dependency

JimSuplizio:
Core released on 3/6 with the exception of azure-core-amqp which has a fix in progress. While libraries built within the same pipeline should always be using the current version of those libraries, an exception needs to be made here for amqp in order for it to be released.  A separate issue will be filed to flip this back to current after it has been released.


conniey:
Thanks Jim!

PR #8881, opened 2020-03-09
fixes live tests for text analytics

mssfang:
fixes offset position and correct ARM template `AZURE_TEXT_ANALYTICS_SUBSCRIPTION_KEY` to `AZURE_TEXT_ANALYTICS_API_KEY`.


samvaity:
Recommend running the live test pipeline against this PR to confirm if it actually fixes the failing tests for API key.

PR #8901, opened 2020-03-10
Adding distinct support

mbhaskar:
- Adding support for distinct queries
- Doesn't support continuation token at the moment

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/OrderedDistinctMap.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/OrderedDistinctMap.java
@@ -3,9 +3,9 @@
 package com.azure.cosmos.implementation.query;
 
 import com.azure.cosmos.implementation.Utils;
+import com.azure.cosmos.implementation.apachecommons.lang.StringUtils;
 import com.azure.cosmos.models.Resource;
 import com.fasterxml.jackson.core.JsonProcessingException;
-import org.apache.commons.lang3.StringUtils;
 
 import java.security.NoSuchAlgorithmException;
 


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctQueryType.java
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+package com.azure.cosmos.implementation.query;
+
+public enum DistinctQueryType {
+    // This means that the query does not have DISTINCT.
+    None,
```
moderakh: Java code style for enum. If the service sends the string value of these in the query plan you can utilize the pattern I added in the ConsistencyLevel

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/OrderedDistinctMap.java
```diff
@@ -0,0 +1,40 @@
...
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
+
+public class OrderedDistinctMap extends DistinctMap {
+    private final ObjectMapper mapper;
+    private String lastHash;
+
+    public OrderedDistinctMap(String lastHash) {
+        mapper = new ObjectMapper().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);
```
moderakh: ObjectMapper instantiation is expensive.
you should have a static ObjectMapper instead:
`OrderedDistinctMap.OBJECT_MAPPER`

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/UnorderedDistinctMap.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Base64;
+import java.util.HashSet;
+
+public class UnorderedDistinctMap extends DistinctMap {
+    private final ObjectMapper mapper;
+    private final HashSet<String> resultSet;
+
+    public UnorderedDistinctMap() {
+        resultSet = new HashSet<>();
+        mapper = new ObjectMapper().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);
```
moderakh: ditto, ObjectMapper instantiation is expensive. we should use a static one. You can create here `UnorderedDistinctMap.OBJECT_MAMMER`

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/DistinctQueryTests.java
```diff
@@ -0,0 +1,351 @@
...
+    public void generateTestData() {
+
+        Random rand = new Random();
+        ObjectMapper mapper = new ObjectMapper();
+        for (int i = 0; i < 40; i++) {
+            Person person = getRandomPerson(rand);
+            try {
+                docs.add(new CosmosItemProperties(mapper.writeValueAsString(person)));
+            } catch (JsonProcessingException e) {
+                e.printStackTrace();
```
moderakh: use logger please

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/DistinctQueryTests.java
```diff
@@ -0,0 +1,351 @@
...
+
+        bulkInsert();
+
+        waitIfNeededForReplicasToCatchUp(clientBuilder());
+    }
+
+    public enum City {
+        NewYork,
+        LosAngeles,
+        Seattle
```
moderakh: java enum code style. take a look at ConsistencyLevel if string value of this matters.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctDocumentQueryExecutionContext.java
```diff
@@ -0,0 +1,115 @@
...
+        this.distinctMap = DistinctMap.create(distinctQueryType, previousHash);
+    }
+
+    public static <T extends Resource> Flux<IDocumentQueryExecutionComponent<T>> createAsync(
+        Function<String, Flux<IDocumentQueryExecutionComponent<T>>> createSourceComponentFunction,
+        DistinctQueryType distinctQueryType,
+        String continuationToken) {
+
+        if (distinctQueryType == DistinctQueryType.Ordered) {
+            CosmosClientException dce = new BadRequestException("Ordered distinct queries are not supported yet");
```
bchong95: why are these blocked?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctDocumentQueryExecutionContext.java
```diff
@@ -0,0 +1,115 @@
...
+            return Flux.error(dce);
+        }
+
+        Utils.ValueHolder<DistinctContinuationToken> outDistinctcontinuationtoken = new Utils.ValueHolder<>();
+        DistinctContinuationToken distinctContinuationToken = new DistinctContinuationToken(null /*lasthash*/,
```
bchong95: This should be set in the else case of `if(continuationToken != null)`.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
```diff
@@ -0,0 +1,109 @@
...
+        assertThat(add).as("Value should be added first time").isTrue();
+        boolean add2 = distinctMap.add(resource, outHash);
+        assertThat(add2).as("same value should not be added again").isFalse();
+        resource = String.format("{ " + "\"id\": \"6\" + \"}");
+        boolean add3 = distinctMap.add(resource, outHash);
+        assertThat(add3).as("different value should be added again").isTrue();
+    }
+
+    @Test(groups = "unit", dataProvider = "distinctMapArgProvider")
+    public void objectValue(DistinctQueryType queryType) {
```
bchong95: We need a test for two objects where the property values are just shuffeled:

```
{
     "name" : "John",
     "age": 24
}
```

vs 

```
{
     "age": 24,
     "name" : "John"
}
```

And they should hash to the same, since ordered of JSON properties does not matter.

bchong95: I see that you have one.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
```diff
@@ -0,0 +1,109 @@
...
+                                             + "\"id\": \"12345\""
+                                             + "} ");
+
+        DistinctMap distinctMap = DistinctMap.create(queryType, null);
+
+        Utils.ValueHolder<String> outHash = new Utils.ValueHolder<>();
+        boolean add = distinctMap.add(resource1, outHash);
+        assertThat(add).as("Value should be added first time").isTrue();
+        boolean add2 = distinctMap.add(resource2, outHash);
+        assertThat(add2).as("Order of objects in map should be treated same").isTrue();
```
bchong95: This should be `false` right? since the ordering does not change the hash.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @DataProvider(name = "distinctMapArgProvider")
+    public Object[][] distinctMapArgProvider() {
+        return new Object[][] {
+            {DistinctQueryType.Ordered},
+            {DistinctQueryType.Unordered},
+        };
+    }
+
+    @Test(groups = "unit", dataProvider = "distinctMapArgProvider")
+    public void integerValue(DistinctQueryType queryType) {
```
bchong95: We need a test for `1.0` vs `1`

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
```diff
@@ -0,0 +1,109 @@
...
+        assertThat(add).as("Value should be added first time").isTrue();
+        boolean add2 = distinctMap.add(resource, outHash);
+        assertThat(add2).as("same value should not be added again").isFalse();
+        resource = String.format("{ " + "\"id\": %d + \"}", 3);
+        boolean add3 = distinctMap.add(resource, outHash);
+        assertThat(add3).as("different value should be added again").isTrue();
+    }
+
+    @Test(groups = "unit", dataProvider = "distinctMapArgProvider")
+    public void stringValue(DistinctQueryType queryType) {
```
bchong95: We a test for strings vs their escaped counterpart

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @Test(groups = "unit", dataProvider = "distinctMapArgProvider")
+    public void nullValue(DistinctQueryType queryType) {
+        DistinctMap distinctMap = DistinctMap.create(queryType, null);
+        Utils.ValueHolder<String> outHash = new Utils.ValueHolder<>();
+        boolean add = distinctMap.add(null, outHash);
+        assertThat(add).as("Value should be added first time").isTrue();
+        boolean add2 = distinctMap.add(null, outHash);
+        assertThat(add2).as("same value should not be added again").isFalse();
+    }
+
```
bchong95: We need a test for arrays.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctQueryType.java
mbhaskar: Thankyou. Changed to follow java code style

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/OrderedDistinctMap.java
mbhaskar: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/UnorderedDistinctMap.java
mbhaskar: Done

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/DistinctQueryTests.java
mbhaskar: Done

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/DistinctQueryTests.java
mbhaskar: changed the test

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctDocumentQueryExecutionContext.java
mbhaskar: Left over as part of testing! Removed this

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
mbhaskar: Yeah added one for this

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
mbhaskar: Yeah it should be, corrected

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
mbhaskar: added an integration test for this

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
mbhaskar: added test for arrays

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/DistinctMapTest.java
mbhaskar: Already creating the object using escaped strings and they are eventually read as Strings by Jackson, so this should be covered

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctDocumentQueryExecutionContext.java
mbhaskar: Initialized it early as I am using outDistinctcontinuationtoken in if condition. And distinctContinuationToken is being used even if it doesnt go into else. So initialized these here

mbhaskar:
> ALSO, what is the implication on the normal query follow (without distinct)?
> 
> can we do some perf testing? please use `QueryInClauseParallel`

This doesnt impact the existing flow. Only addition is an extra if condition in pipeline to check if it is a distinct query.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctDocumentQueryExecutionContext.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+public class DistinctDocumentQueryExecutionContext<T extends Resource> implements IDocumentQueryExecutionComponent<T> {
+    private final IDocumentQueryExecutionComponent<T> component;
+    private final DistinctMap distinctMap;
+    private final DistinctQueryType distinctQueryType;
+    private String lastHash;
```
moderakh: is `lastHash` field value needed?
if so shouldn't this be volatile to ensure threads have consistent view on the value?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/OrderedDistinctMap.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import java.nio.charset.Charset;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
+
+public class OrderedDistinctMap extends DistinctMap {
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);;
+    private String lastHash;
```
moderakh: is `lastHash` field value needed?
if so shouldn't this be volatile to ensure threads have consistent view on the value?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/UnorderedDistinctMap.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.nio.charset.Charset;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
+import java.util.HashSet;
+
+public class UnorderedDistinctMap extends DistinctMap {
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);;
+    private final HashSet<String> resultSet;
```
moderakh: if `resultSet` can be accessed concurrently you should consider making this ConcurrentHashSet

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/OrderedDistinctMap.java
mbhaskar: yup, made this volatile

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/UnorderedDistinctMap.java
mbhaskar: Yes, made it a ConcurrentHashSet

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctDocumentQueryExecutionContext.java
mbhaskar: Using AtomicReference now.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctMap.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public abstract boolean add(Resource object, Utils.ValueHolder<String> outHash);
+
+    String getHash(Resource resource) throws JsonProcessingException,
+                                                 NoSuchAlgorithmException {
+        final Object obj = OBJECT_MAPPER.treeToValue(ModelBridgeInternal.getPropertyBagFromJsonSerializable(resource)
+            , Object.class);
+        final String sortedJson =
+            OBJECT_MAPPER.writeValueAsString(obj);
+        MessageDigest md = MessageDigest.getInstance("SHA-1");
```
moderakh: @bchong95 sha1 is not secure. as this is only used for hashing that shouldn't be a concern. But I wonder why didn't we use other hashes?

@mbhaskar how does this do in terms of perf?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctMap.java
```diff
@@ -0,0 +1,47 @@
...
+        byte[] digest = md.digest(sortedJson.getBytes(Charset.defaultCharset()));
```
moderakh: shouldn't we be using utf-8 as encoding? default charset may change from one machine to another machine

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctMap.java
mbhaskar: @moderakh .NET doesnt use SHA-1, they use their CosmosElement framework and then use Murmurhash3 on individual types. JS and Java use SHA-1

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctMap.java
mbhaskar: Unless we run partial queries in different machines, this might not be a concern! but in general it might not be a good practise to use defaultCharset, will move to UTF-8 in next PR

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/query/DistinctMap.java
mbhaskar: Also, we can continue to work on this hash as a util function and make this stronger and robust.

mbhaskar: this is also used in group by, so will try to test it further and improve

PR #8934, opened 2020-03-11
Remove build steps not required for report generation

srnagar:


srnagar:
Updated yml to skip revapi in `Run Tests` task

eng/pipelines/aggregate-reports.yml
```diff
@@ -28,10 +28,10 @@ jobs:
         goals: 'install'
 
     - task: Maven@3
-      displayName: 'Install Module and Run SpotBugs, Checkstyle, RevApi, and Javadoc'
+      displayName: 'Install Module'
       inputs:
         mavenPomFile: sdk/core/pom.service.xml
-        options: '--no-transfer-progress -DskipTests -Dgpg.skip -Dverify-readme'
+        options: '--no-transfer-progress -DskipTests -Dgpg.skip -Dcheckstyle.skip=true -Dspotbugs.skip=true -Dmaven.javadoc.skip=true -Drevapi.skip=true'
```
alzimmermsft: Should `jacoco.skip=true` be added here as well?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -2,7 +2,7 @@ parameters:
   SDKType: client
   ServiceDirectory: not-specified # Set a default that breaks in obvious ways.
   PreTestSteps: []
-  TestOptions: '$(DefaultOptions) -Dmaven.javadoc.skip=true -Dspotbugs.skip=true -Dcheckstyle.skip=true'
+  TestOptions: '$(DefaultOptions) -Dmaven.javadoc.skip=true -Dspotbugs.skip=true -Dcheckstyle.skip=true -Drevapi.skip=true'
```
alzimmermsft: Should jacoco be skipped here as well?

srnagar: We need the jacoco report generated here because we want to fail the build if coverage drops below the threshold.

PR #8936, opened 2020-03-11
Remove keyvault track 1

mitchdenny:
This PR removes traces of KeyVault track 1 from the repository. I'm proposing this change because we are making changes to the engineering system and removing track 1 of keyvault at this time would make this change easier. Should we wish to make a servicing release to the track 1 version of the library we would make a branch from before this commit and make the changes and release the library.

Before merging this change we would mothball the pipeline definition for track 1 by moving it into the legacy folder and disabling it.

Summary of changes:

1. Remove track 1 libraries from sdk/keyvault
2. Remove references in pom.data.xml
3. Remove ci.data.yml file.


weshaggard:
closing based on conversation of keeping this working for servicing.

PR #8975, opened 2020-03-11
Replace HttpBin Usage with WireMock

### LINKED ISSUE #3255: Check and update the jetty-maven-plugin and exec-maven-plugin
The jetty-maven-plugin and exec-maven-plugin may outdated with WireWork. Need to check and update.

### LINKED ISSUE #8096: Azure Core HTTP Netty Flaky test
The following test fails occasionally as part of live tests and CI on PRs.

[ERROR] Tests run: 128, Failures: 0, Errors: 1, Skipped: 3, Time elapsed: 3.009 s <<< FAILURE! - in com.azure.core.http.netty.implementation.RestProxyWithNettyTests
[ERROR] service19PutWithNoContentTypeAndByteArrayBodyWithNullBody  Time elapsed: 0.031 s  <<< ERROR!

Live test failure: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=256676&view=logs&j=5f969c9e-a71b-579d-7ec1-8fe84403e65c&s=b15d9194-8f26-5328-b47f-5968c76b37e7&t=b69a79bc-1781-5a49-93cb-bc6f011737ea&l=414

alzimmermsft:
Fixes #8096

This PR ports `RestProxyTests` from using the `HttpBin` service to using mocked responses via `WireMock`. This change reduces the likelihood that a test is flaky due to HttpBin service issues or flaky networking as this isn't the purpose of the tests.

Also fixes #3255 by removing `RestProxyStressTests` and `MockServer` which allows for the Jetty dependencies to be removed from Azure Core's testing scope.

### DIFF
--- a/sdk/core/azure-core-test/src/test/java/com/azure/core/test/RestProxyTestsWireMockServer.java
+++ b/sdk/core/azure-core-test/src/test/java/com/azure/core/test/RestProxyTestsWireMockServer.java
@@ -102,9 +102,12 @@ public final class RestProxyTestsWireMockServer {
             rawHeaders.put("Content-Type", "application/octet-stream");
             rawHeaders.put("Content-Length", String.valueOf(bodySize));
 
+            byte[] body = new byte[bodySize];
+            new SecureRandom().nextBytes(body);
+
             return new ResponseDefinitionBuilder().withStatus(200)
                 .withHeaders(toWireMockHeaders(rawHeaders))
-                .withBody(new SecureRandom().generateSeed(bodySize))
+                .withBody(body)
                 .build();
         }
 


sdk/core/azure-core-http-netty/pom.xml
```diff
@@ -116,7 +116,14 @@
...
       <artifactId>azure-core-test</artifactId>
-      <version>1.2.0-beta.1</version>  <!-- {x-version-update;com.azure:azure-core-test;current} -->
+      <version>1.2.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core-test;current} -->
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-test</artifactId>
+      <version>1.2.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core-test;current} -->
+      <type>test-jar</type>
```
srnagar: If you have the test-jar, do you need the one without the test-jar too?

sdk/core/azure-core-http-netty/src/test/java/com/azure/core/http/netty/implementation/RestProxyWithNettyTests.java
```diff
@@ -4,12 +4,36 @@
...
+    public static void shutdownWireMockServer() {
+        if (server != null) {
+            server.shutdown();
+        }
+    }
+
+    @Override
+    protected int getWireMockPort() {
+        return server.port();
+    }
```
srnagar: This is the same code in `RestProxyWithHttpProxyNettyTests`. Can you check if this can be de-duplicated?

sdk/core/azure-core-http-netty/pom.xml
alzimmermsft: I'll try this out locally, if it continues to work the same I'll remove the other.

alzimmermsft: Build failed

sdk/core/azure-core-http-netty/src/test/java/com/azure/core/http/netty/implementation/RestProxyWithNettyTests.java
alzimmermsft: For now I will be keep these separate as I believe WireMock is capable of allowing proxying which we may want to take advantage of to final begin testing proxying.

PR #8981, opened 2020-03-11
Make changes on Analyzer property in Field, and CustomerAnalyzer

sima-zhu:


sdk/search/azure-search/src/main/java/com/azure/search/models/Field.java
```diff
@@ -447,12 +493,35 @@ public String getAnalyzer() {
...
+     * 'th.microsoft', 'th.lucene', 'tr.microsoft', 'tr.lucene',
+     * 'uk.microsoft', 'ur.microsoft', 'vi.microsoft', 'standard.lucene',
+     * 'standardasciifolding.lucene', 'keyword', 'pattern', 'simple', 'stop',
+     * 'whitespace'.
      *
      * @param analyzer the analyzer value to set.
      * @return the Field object itself.
      */
-    public Field setAnalyzer(String analyzer) {
+    public Field setAnalyzer(AnalyzerName analyzer) {
```
brjohnstmsft: `indexAnalyzer` and `searchAnalyzer` should also be of type `AnalyzerName`. The Swagger is defined this way; I'm not sure why the generated code isn't.

sima-zhu: Swagger made the right return type for the three class.
In transformer, we covert the AnalyzerName Enum type to String. I only made the changes on the one on issue. Thanks for pointing this out. 
I will change them all.

Issue #8741, opened 2020-03-05
[SB Track2] Implement `peek()` and `peekBatch` in ServiceBus*Receiver


conniey:
## Tasks
- [ ] `ServiceBusAsyncReceiver.peek()` and `peekBatch()` (including necessary overloads)
  - [x] Add unit tests
  - [x] Add integration tests
- [ ] `ServiceBusReceiver.peek()` and `peekBatch()` (including necessary overloads)
  - [ ] Add unit tests
  - [ ] Add integration tests

hemanttanwar:
Not doing  peek for Batch, I will prioritize it low,  first I will push other feature and come back to Batch. Also core team is discussing Batching design options.

PR #8983, opened 2020-03-11
ServiceBus Track 2 Peek operation on Management channel  ( Include $cbs channel for auth)

hemanttanwar:
This include following feature

1. $cbs channel for sending auth token.
2. peek operation 
3. peek from a sequence number

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClient.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClient.java
@@ -30,7 +30,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * An <b>asynchronous</b> receiver responsible for receiving {@link ServiceBusReceivedMessage} from a specific queue
- * or topic on Azure Service Bus.
+ * or topic on Azure Service Bus. 
  */
 @ServiceClient(builder = ServiceBusClientBuilder.class, isAsync = true)
 public final class ServiceBusReceiverAsyncClient implements Closeable {


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/AmqpConstants.java
```diff
@@ -19,4 +19,8 @@
 
     static final int AMQP_BATCH_MESSAGE_FORMAT = 0x80013700; // 2147563264L;
 
+    public static final String REQUEST_RESPONSE_LEGACY_STATUS_CODE = "status-code";
```
conniey: This already exists in RequestResponseChannel.java. Is there a reason it is here?

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/MessageSerializerResult.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/***
+ * This will hole additional data which can not be stored in AMQP Message.
+ * Example last  serial number.
+ */
+public class MessageSerializerResult<T> {
+    List<T> messages = null;
```
conniey: Access modifier and using final instead. Consider setting Collections.emptyList instead.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/MessageSerializerResult.java
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
conniey: This class isn't used anywhere?

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ReactorConnection.java
```diff
@@ -77,10 +81,13 @@
...
      * @param product The name of the product this connection is created for.
      * @param clientVersion The version of the client library creating the connection.
+     * @param senderSettleMode to set as {@link SenderSettleMode} on sender.
+     * @param receiverSettleMode to set as {@link ReceiverSettleMode} on receiver.
      */
     public ReactorConnection(String connectionId, ConnectionOptions connectionOptions, ReactorProvider reactorProvider,
         ReactorHandlerProvider handlerProvider, TokenManagerProvider tokenManagerProvider,
-        MessageSerializer messageSerializer, String product, String clientVersion) {
+        MessageSerializer messageSerializer, String product, String clientVersion,
+        SenderSettleMode senderSettleMode, ReceiverSettleMode receiverSettleMode) {
```
conniey: Why does a connection know about these modes? I'd expect them to be parameters when creating a receiver or sender or when creating the management node itself.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RequestResponseUtils.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.azure.core.amqp.implementation;
+
+import org.apache.qpid.proton.message.Message;
+
+/**
+ * This consists of various utilities needed to manage Request/Response  channel.
+ */
+public class RequestResponseUtils {
+    public static final int REQUEST_RESPONSE_UNDEFINED_STATUS_CODE = -1;
```
conniey: Can this live as a private static method in RequestResponseChannel.

hemanttanwar:
fixes #8741   ( No batch API in this PR)

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
```diff
@@ -48,6 +53,7 @@
     private static final String VIA_PARTITION_KEY_NAME = "x-opt-via-partition-key";
     private static final String DEAD_LETTER_SOURCE_NAME = "x-opt-deadletter-source";
 
+
```
conniey: nit:  empty line

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RequestResponseUtils.java
conniey: Since this is related to reading the status code from AmqpMessages, a better place would be to put it in MessageUtils. There should be a class already

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
```diff
@@ -164,20 +170,37 @@ public int getSize(org.apache.qpid.proton.message.Message amqpMessage) {
...
         }
     }
 
+    private List<ServiceBusReceivedMessage> deserializeListOfMessages(Message amqpMessage) {
+        //maintain the order of elements because last sequence number needs to be maintain.
+        List<Message> listAmqpMessages = convertAMQPValueMessageToBrokeredMessage(amqpMessage);
+
+        List<ServiceBusReceivedMessage> receivedMessageList = new ArrayList<>();
+        for (Message oneAmqpMessage:listAmqpMessages
+             ) {
```
conniey: nit: new line

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -59,9 +62,17 @@
...
+    private final TokenManager cbsBasedTokenManager;
+
+    // Maintain last peek sequence number
+    private AtomicReference<Long>  lastPeekedSequenceNumber = new AtomicReference<>(0L);
+
+
     ManagementChannel(Mono<RequestResponseChannel> responseChannelMono, String topicOrQueueName,
                       TokenCredential credential, TokenManagerProvider tokenManagerProvider,
-                      MessageSerializer messageSerializer, Scheduler scheduler) {
+                      MessageSerializer messageSerializer, Scheduler scheduler, TokenManager cbsBasedTokenManager
```
conniey: Why do we pass in the TokenManager here. iirc, before we create the ManagementChannel, we already create a token manager that authorizes with it. We shouldn't need to manually authorize ourselves.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementConstants.java
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the MIT License.
+
+package com.azure.messaging.servicebus.implementation;
+
+import com.azure.core.amqp.implementation.AmqpConstants;
+
+/**
+ * Constants which is used for management calls to support operations for example renewlock, schedule, defer etc.
+ */
+public class ManagementConstants {
```
conniey: This doesn't need to be public

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusReactorAmqpConnection.java
```diff
@@ -101,10 +117,14 @@ public ServiceBusReactorAmqpConnection(String connectionId, ConnectionOptions co
                     logger.info("Creating management node. entityPath: [{}]. address: [{}]. linkName: [{}]",
                         entityPath, address, linkName);
 
+                    TokenManager cbsBasedTokenManager =  new AzureTokenManagerProvider(
```
conniey: Every time a peek operation is called, it'll create another management channel node.  This is a bit wasteful if we keep getting the management node and have to recreate it.

sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientPeek.java
```diff
@@ -143,7 +143,7 @@ public void peekOneMessage() {
                 });
 
         try {
-            Thread.sleep(5000);
+            Thread.sleep(90000);
```
conniey: Move these to tests and use stepverifier.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
```diff
@@ -54,4 +54,28 @@ void receiveMessageAutoComplete() {
...
+        final String messageId = UUID.randomUUID().toString();
+        final String contents = "Some-contents";
+        final ServiceBusMessage message = TestUtils.getServiceBusMessage(contents, messageId, 0);
+        final ReceiveMessageOptions options = new ReceiveMessageOptions().setAutoComplete(true);
+        receiver = createBuilder()
+            .receiveMessageOptions(options)
+            .buildAsyncReceiverClient();
+
+        // Assert & Act
+        StepVerifier.create(sender.send(message).thenMany(receiver.peek()))
```
conniey: sender.send(message).then(receiver.peek()) is probably what you want. thenMany suggests it returns a flux.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientTest.java
```diff
@@ -122,6 +131,41 @@ void teardown() {
...
+     * Verifies that this peek one messages from a sequence Number.
+     */
+    @Test
+    void peekWithSequenceOneMessage() {
+        // Arrange
+        final int numberOfEvents = 1;
+        final int fromSequenceNumber = 10;
+
+        when(managementNode.peek(fromSequenceNumber))
+            .thenReturn(Mono.just(mock(ServiceBusReceivedMessage.class)));
```
conniey: When you return inline mocks like this, you need to add the Mockito.clearInlinMocks() or else it won't be garbage collected.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/MessageSerializerResult.java
hemanttanwar: I have deleted it.  I thought of using it but survived without it.  Will Make sure to remove it

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
hemanttanwar: Good catch...

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RequestResponseUtils.java
hemanttanwar: RequestResponseUtils.getResponseStatusCode( is used by )  RequestResponseChannel  and also by  ServiceBusMessageSerializer and shared. Thuis I put it in amqp-core and  there is no MessageUtils in amqp-core  .

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/MessageSerializerResult.java
hemanttanwar: removed this class.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusReactorAmqpConnection.java
hemanttanwar: I have tested with 4 peek() calls from one client , since we are storing `ServiceBusManagementNode` in `ConcurrentHashMap`  and return cached managementNode after once created, 

https://github.com/Azure/azure-sdk-for-java/pull/8983/files#diff-6460b86560d643d02d2e58af8b6213cfL96

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
hemanttanwar: The `TokenManager` instance is created before `ManagementChannel` is created but does not call `TokenManager.authorize()`. So we pass this instance in `ManagementChannel`  because  it knows when user makes first call to `peek`. First call to `authorize()` should happen when user does peek operation, So  we call it `once only` just before peek is called.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusReactorAmqpConnection.java
```diff
@@ -36,6 +41,11 @@
     private static final String MANAGEMENT_LINK_NAME = "mgmt";
     private static final String MANAGEMENT_ADDRESS = "$management";
 
+    /** This is used in setting up management chhannel and it is always fixed.
+     * This is not used by sdk user trying to receive/send messages.*/
+    public static final SenderSettleMode MANAGEMENT_SEND_SETTLE_MODE = SenderSettleMode.SETTLED;
```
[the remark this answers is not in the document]
conniey: It doesn't need to be a variable because enums are basically static. You can pass the values in as-is in the constructor.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -46,79 +47,107 @@
...
+    public Mono<ServiceBusReceivedMessage> peek() {
 
-            if (!Objects.isNull(sessionId)) {
-                requestBodyMap.put(ServiceBusConstants.REQUEST_RESPONSE_SESSION_ID, sessionId);
-            }
+        return peek(this.lastPeekedSequenceNumber.get() + 1);
+    }
 
-            request.setBody(new AmqpValue(requestBodyMap));
+    private <T> Flux<T> peek(Map<String, Object> appProperties, Class<T> responseType,
```
conniey: AppProperties as a parameter isn't necessary, you always pass the same parameter.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
```diff
@@ -54,4 +54,53 @@ void receiveMessageAutoComplete() {
...
+     * Verifies that we can send and peek a message.
+     */
+    @Test
+    void peekMessage() {
+        // Arrange
+        final String messageId = UUID.randomUUID().toString();
+        final String contents = "Some-contents";
+        final ServiceBusMessage message = TestUtils.getServiceBusMessage(contents, messageId, 0);
+        final ReceiveMessageOptions options = new ReceiveMessageOptions().setAutoComplete(true);
+        receiver = createBuilder()
```
conniey: You should dispose of the existing one (created in BeforeEach) if you are going to recreate it. It'll still consume resources.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientTest.java
```diff
@@ -122,6 +130,43 @@ void teardown() {
...
+
+        // Act & Assert
+        StepVerifier.create(consumer.peek(fromSequenceNumber))
+            .expectNextCount(numberOfEvents)
+            .verifyComplete();
+
+        Mockito.framework().clearInlineMocks();
```
conniey: Should put this in AfterEach, if the test fails at line 163, it'll never be run.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientTest.java
```diff
@@ -122,6 +130,43 @@ void teardown() {
...
+     */
+    @Test
+    void peekOneMessage() {
+        // Arrange
+        final int numberOfEvents = 1;
+        when(managementNode.peek())
+            .thenReturn(Mono.just(mock(ServiceBusReceivedMessage.class)));
+
+        // Act & Assert
+        StepVerifier.create(consumer.peek())
```
conniey: Do you have a test that peeks the next sequence number?

conniey:
Apply the small fixes. I am working with RequestResponse and Management Node, I'll do the refactors.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientTest.java
hemanttanwar: It is already there in `@AfterEach`

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -46,79 +47,107 @@
...
+                             int maxMessages, long fromSequenceNumber, UUID sessionId) {
 
-            return channelMono.flatMap(requestResponseChannel -> requestResponseChannel.sendWithAck(request))
-                .map(message -> messageSerializer.deserialize(message, responseType));
-        });
+        return
+            Mono.defer(() -> {
+                if (!cbsBasedTokenManagerCalled.get()) {
```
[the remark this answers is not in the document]
hemanttanwar: We can look into it in more detail and make it better.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientTest.java
hemanttanwar: added this test

PR #8995, opened 2020-03-11
Remove the appinsights data plane library.

mitchdenny:
This PR removes ```sdk/applicationinsights/microsoft-azure-applicationinsights-query```. This library never shipped and the application insights SDK is being maintained out of a separate repo so removing this to reduce maintenance burden around the ES and to avoid any confusion.

/cc @jianghaolu


mitchdenny:
Pipelines also disabled and moved to \java\legacy (for now)

PR #8998, opened 2020-03-12
update Identity versions in Track 2 SDKs - march 

g2vinay:


alzimmermsft:
Shouldn't the Identity README also use 1.0.4?

g2vinay:
> Shouldn't the Identity README also use 1.0.4?

yeah, it should, 
But will do it in a separate PR as pushing again will trigger all pipelines again,

PR #9016, opened 2020-03-12
Remove loganalytics data plane module.'

mitchdenny:
This PR removes the data plane loganalytics module. It doesn't look like it has ever had a GA release.

@jianghaolu I think you probably know about this one too?


mitchdenny:
Ping @jianghaolu

PR #9030, opened 2020-03-12
Integrate Dependency Checker Plugin into pom.client.xml

alzimmermsft:
This PR integrates the dependency checker's Maven plugin into `pom.client.xml`. The analyses pipeline will no longer need to checkout the source code from GitHub and install the checker locally. A new profile was added into `pom.client.xml` so this check won't be triggered by default and the plugin will only run for `pom.client.xml` and none of the inheriting projects.

Additionally, this PR also fixes the improperly configured dependency checker plugin used in the `azure-sdk-bom` project and revs the version being used.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/index.html
+++ b/eng/code-quality-reports/src/main/resources/index.html
@@ -16,19 +16,19 @@
             font-family: sans-serif;
             -ms-text-size-adjust: 100%
         }
-        
+
         a {
             background-color: transparent
         }
-        
+
         h1 {
             margin: .67em 0
         }
-        
+
         hr {
             height: 0
         }
-        
+
         *,
         :after,
         :before,
@@ -38,41 +38,41 @@
         input[type=search] {
             box-sizing: border-box
         }
-        
+
         html {
             font-size: 10px;
             -webkit-tap-highlight-color: transparent
         }
-        
+
         body {
             font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
             font-size: 14px;
             line-height: 1.428571429;
             color: #333
         }
-        
+
         a {
             color: #67808c;
             text-decoration: none
         }
-        
+
         a:focus,
         a:hover {
             color: #465760;
             text-decoration: underline
         }
-        
+
         a:focus {
             outline: -webkit-focus-ring-color auto 5px;
             outline-offset: -2px
         }
-        
+
         hr {
             margin-top: 20px;
             margin-bottom: 20px;
             border-top: 1px solid #eee
         }
-        
+
         .h1,
         .h2,
         .h3,
@@ -90,7 +90,7 @@
             line-height: 1.1;
             color: inherit
         }
-        
+
         .h1 .small,
         .h1 small,
         .h2 .small,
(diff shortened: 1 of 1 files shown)


eng/pipelines/aggregate-reports.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       displayName: 'Generate aggregate code coverage report'
```
JonathanGiles: Update title to say code coverage and dependencies

pom.client.xml
```diff
@@ -113,7 +113,8 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.8.1</version> <!-- {x-version-update;org.apache.maven.plugins:maven-compiler-plugin;external_dependency} -->
+        <version>3.8.1
+        </version> <!-- {x-version-update;org.apache.maven.plugins:maven-compiler-plugin;external_dependency} -->
```
JonathanGiles: What's up with the line wrapping?

alzimmermsft: IntelliJ auto formatting breaking lines after 120 characters, will update my configuration to ignore XML files (or set it to something large like 200 characters).

eng/pipelines/aggregate-reports.yml
alzimmermsft: Updating to `Generate aggregate code coverage and dependency report`

mitchdenny:
How will this change socialize if we remove the list of modules from ```pom.client.xml``` which is one of the changes landing with the POM restructuring. Will dependency checker continue to work in that secnario.

Note ... that for the core pipeline we'll be running tests with the -am and -amd switches which means it should build the full graph of track 2 libraries. Take a look at this PR to see what the incoming changes look like:

https://github.com/Azure/azure-sdk-for-java/pull/8242

alzimmermsft:
@mitchdenny That is a good question, I'll need to review what is being changed in that PR. If there is still a parent that lists all modules this change is still possible and can be ported to the new pom, if not this change will basically be a moot point.

eng/pipelines/aggregate-reports.yml
```diff
@@ -86,31 +86,14 @@ jobs:
...
-
     - pwsh: |
         $(Build.SourcesDirectory)/eng/DependencyGraph/gen-dep-graph-data.ps1
       displayName: Generate Interdependency Report Data
 
     - pwsh: |
         copy -r target/staging $(Build.ArtifactStagingDirectory)
         copy eng/code-quality-reports/src/main/resources/index.html $(Build.ArtifactStagingDirectory)
-        copy output/dependencies.html (Join-Path $(Build.ArtifactStagingDirectory) "staging")
+        copy output/pom.client.html (Join-Path $(Build.ArtifactStagingDirectory) "staging")
```
JonathanGiles: You will need to update the link in /eng/code-quality-reports/src/main/resources/index.html to change based on this file rename.

alzimmermsft: Should I remove the link out to the SpotBugs aggregate report as well since that doesn't get generated anymore?

PR #9031, opened 2020-03-12
Leveraged common datalake buffered upload classes in blob

gapra-msft:


gapra-msft:
@JonathanGiles is it ok to remove the protected method in BlobAsyncClient since it was marked Internal Use Only in Doc? If not, I can add it back. It is currently not used in the class since I refactored some of the buffered upload common logic.

JonathanGiles:
Don't remove it, because it is a breaking change. Add it back in and deprecate it appropriately.

PR #9035, opened 2020-03-12
Fix version scanner for a couple of scenarios

JimSuplizio:
Fix the pom file version scanner for a couple of scenarios
1. $versionNode.InnerText should be trimmed before it's passed off to the validation function. There's a case where intellij breaks up lines that are over a certain number of characters which ends up splitting the line because it's too long. For example:
`      <version>1.1.0`
`      </version> <!-- {x-version-update;com.azure:azure-core-test;dependency} -->`

2. The dependencyManagement check should be looking for whether or not $xmlPomFile.project.dependencyManagement exists instead of scanning for an element named dependencyManagement. There's a scenario where one of the plugins had a dependencyManagement configuration option which was tripping the dependencyManagement check and failing incorrectly.

### DIFF
--- a/eng/versioning/pom_file_version_scanner.ps1
+++ b/eng/versioning/pom_file_version_scanner.ps1
@@ -257,7 +257,7 @@ Get-ChildItem -Path $Path -Filter pom*.xml -Recurse -File | ForEach-Object {
             return
         }
     }
-    
+
     Write-Host "processing pomFile=$($pomFile)"
     if ($xmlPomFile.project.dependencyManagement)
     {


mitchdenny:
nit: this file is snake case where convention for PowerShell is Verb-Noun.ps1 ... even for scripts :)

JimSuplizio:
At this point I'm going to ignore the runs failing to sync azure-sdk-build-tools. Since this is a powershell script that only runs during the analyze step and contains changes necessary to get the dependency checker changes that @alzimmermsft is making into the system, I'm going to submit it on the merit of the successful core run.

PR #9036, opened 2020-03-12
Updated service versions

gapra-msft:
Also regenerated off swagger


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/AzureBlobStorageBuilder.java
```diff
@@ -97,7 +97,7 @@ public AzureBlobStorageImpl build() {
         if (this.version != null) {
             client.setVersion(this.version);
         } else {
-            client.setVersion("2019-07-07");
+            client.setVersion("2019-12-12");
```
alzimmermsft: Should this be `2019-10-10` based on the comment in `BlobServiceVersion`?

gapra-msft: Eventually we want it to be 2019-12-12 anyway and the way the builders are set up they always provide the latest service version in the enum  (i.e. this.version will never be null) so it will send 10-10 out on the wire.

PR #9057, opened 2020-03-13
Add event processor sample for Azure Stack usecase

srnagar:
Azure Stack has an older version of Blob Storage and Track 2 Storage SDK only support newer versions. In order to use the blob checkpoint store with Event Processor, older version of storage has to be added as a header policy when creating the container client. This PR has a sample to demonstrate how this can be done.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/src/samples/java/com/azure/messaging/eventhubs/checkpointstore/blob/EventProcessorWithOlderStorageVersion.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/src/samples/java/com/azure/messaging/eventhubs/checkpointstore/blob/EventProcessorWithOlderStorageVersion.java
@@ -17,7 +17,7 @@ import java.util.concurrent.CountDownLatch;
  */
 public class EventProcessorWithOlderStorageVersion {
 
-    private static final String STORAGE_CONNECTION_STRING ="";
+    private static final String STORAGE_CONNECTION_STRING = "";
     private static final String SAS_TOKEN = "";
     private static final String CONTAINER_NAME = "";
 


sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/src/samples/java/com/azure/messaging/eventhubs/checkpointstore/blob/EventProcessorWithOlderStorageVersion.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+public class EventProcessorWithOlderStorageVersion {
+
+    private static final String STORAGE_CONNECTION_STRING ="";
+    private static final String SAS_TOKEN = "";
+    private static final String CONTAINER_NAME = "";
+
+    private static final String EH_CONNECTION_STRING = "";
+    private static final String CONSUMER_GROUP = "";
+    public static final String EVENT_HUB_NAME = "";
```
conniey: Nit: private to be consistent

sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/src/samples/java/com/azure/messaging/eventhubs/checkpointstore/blob/EventProcessorWithOlderStorageVersion.java
```diff
@@ -0,0 +1,86 @@
...
+import com.azure.messaging.eventhubs.EventProcessorClient;
+import com.azure.messaging.eventhubs.EventProcessorClientBuilder;
+import com.azure.storage.blob.BlobContainerAsyncClient;
+import com.azure.storage.blob.BlobContainerClientBuilder;
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * This sample shows how to setup an Event Processor that uses an older version of blob storage service as
+ * checkpoint store.
+ */
```
jfggdl: Consider adding more context to the situation on which a user will need to use this sample. Something like: The following sample can be used if the environment you are targeting supports a different version of Storage Blob SDK than those typically available on Azure. For example, if you are running Event Hubs on an Azure Stack Hub version 2002, the highest available version for the Storage service is version 2017-11-09. In this case, you will need to use the following code to change the Storage service API version to 2017-11-09. For more information on the Azure Storage service versions supported on Azure Stack Hub, please refer to docs.microsoft.com/en-us/azure-stack/user/azure-stack-acs-differences.

PR #9072, opened 2020-03-13
Make changes to hide SearchDocumentsResult and SuggestDocumentsResult 

sima-zhu:
The PR is to address the API changes around `SearchDocumentsResult` and `SuggestDocumentsResult`.

This two classes internally transfer to `SearchPagedResponse/SuggestPagedResponse` field, which are not used for end-user anymore. 
Hide these two classes.

Remove ODATA_CONTEX cleanup as we added Accept Header in service API.

### DIFF
--- a/sdk/search/azure-search/src/test/resources/session-records/canGetDynamicDocumentWithNullOrEmptyValues.json
+++ b/sdk/search/azure-search/src/test/resources/session-records/canGetDynamicDocumentWithNullOrEmptyValues.json
@@ -1,48 +1,52 @@
 {
-  "networkCallRecords" : [ {
-    "Method" : "POST",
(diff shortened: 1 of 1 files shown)


sdk/search/azure-search/src/main/java/com/azure/search/implementation/models/SearchError.java
```diff
@@ -4,7 +4,7 @@
 // Changes may cause incorrect behavior and will be lost if the code is
 // regenerated.
 
-package com.azure.search.models;
+package com.azure.search.implementation.models;
```
alzimmermsft: This is part of the public API for `SearchErrorException` and will need to remain under models, not implementation.models.

sima-zhu: Right, will move it back

PR #9074, opened 2020-03-13
added a blob changefeed package

gapra-msft:


sdk/storage/azure-storage-blob-changefeed/pom.xml
```diff
@@ -0,0 +1,264 @@
...
+      <artifactId>azure-core-http-netty</artifactId>
+      <version>1.4.0</version> <!-- {x-version-update;com.azure:azure-core-http-netty;dependency} -->
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>1.7.25</version> <!-- {x-version-update;org.slf4j:slf4j-simple;external_dependency} -->
+      <scope>test</scope>
+    </dependency>
```
alzimmermsft: Mind removing this test dependency? We're looking to remove all `slf4j-simple` dependencies in test as we now have a default fallback logger which will be used.

sdk/storage/azure-storage-blob-changefeed/src/main/java/module-info.java
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module azure.storage.blob.changefeed {
+    exports com.azure.storage.blob.changefeed;
```
alzimmermsft: Need to `requires transitive com.azure.core` here.

gapra-msft:
@JimSuplizio could you take a look at this PR please?

Issue #9080, opened 2020-03-13
[BUG] buffer reallocate exception 


VIJAY-SUBRAMANIAM:
BufferOverflowException has thrown when netty trying to reallocate buffer.   The issue with azure v12.5.0 and netty 4.1.44.Final

It seems issue related to https://github.com/netty/netty/issues/9911
 
2020-03-13 20:37:55,967 [a8c597cc8a78] [reactor-http-epoll-21] WARN  reactor.netty.http.client.HttpClientConnect - [id: 0x09bbb32b, L:/172.17.0.5:52526 - R:lntsonymusicdsrv.blob.core.windows.net/52.239.220.100:443] The connection observed an error
java.nio.BufferOverflowException: null
	at java.nio.DirectByteBuffer.put(DirectByteBuffer.java:373) ~[?:?]
	at io.netty.buffer.PoolArena$DirectArena.memoryCopy(PoolArena.java:795) ~[netty-buffer-4.1.44.Final.jar:?]
	at io.netty.buffer.PoolArena$DirectArena.memoryCopy(PoolArena.java:704) ~[netty-buffer-4.1.44.Final.jar:?]
	at io.netty.buffer.PoolArena.reallocate(PoolArena.java:405) ~[netty-buffer-4.1.44.Final.jar:?]
	at io.netty.buffer.PooledByteBuf.capacity(PooledByteBuf.java:118) ~[netty-buffer-4.1.44.Final.jar:?]
	at io.netty.buffer.AbstractByteBuf.ensureWritable(AbstractByteBuf.java:334) ~[netty-buffer-4.1.44.Final.jar:?]
	at io.netty.handler.ssl.SslHandler.attemptCopyToCumulation(SslHandler.java:2210) ~[netty-handler-4.1.44.Final.jar:?]
	at io.netty.handler.ssl.SslHandler.access$2800(SslHandler.java:166) ~[netty-handler-4.1.44.Final.jar:?]
	at io.netty.handler.ssl.SslHandler$SslHandlerCoalescingBufferQueue.compose(SslHandler.java:2171) ~[netty-handler-4.1.44.Final.jar:?]
	at io.netty.channel.AbstractCoalescingBufferQueue.remove(AbstractCoalescingBufferQueue.java:175) [netty-transport-4.1.44.Final.jar:?]
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:814) [netty-handler-4.1.44.Final.jar:?]
[...]

PR #9081, opened 2020-03-13
Update Reactor and Netty Dependencies

alzimmermsft:
This PR updates the Netty dependencies to `4.1.45` and the Reactor dependency to `3.3.3` to match the version used by Reactor Netty.

This resolves issue #9080 by updating the version of Netty being used to a version that fixes the mentioned issues.

### DIFF
--- a/eng/versioning/external_dependencies.txt
+++ b/eng/versioning/external_dependencies.txt
@@ -180,7 +180,7 @@ cosmos_io.netty:netty-handler;4.1.45.Final
 cosmos_io.netty:netty-handler-proxy;4.1.45.Final
 cosmos_io.netty:netty-tcnative;2.0.27.Final
 cosmos_io.netty:netty-transport-native-epoll;4.1.45.Final
-cosmos_io.projectreactor:reactor-core;3.3.2.RELEASE
+cosmos_io.projectreactor:reactor-core;3.3.3.RELEASE
 cosmos_org.mockito:mockito-core;1.10.19
 
 # sdk\eventgrid\microsoft-azure-eventgrid\pom.xml
--- a/sdk/cosmos/azure-cosmos/pom.xml
+++ b/sdk/cosmos/azure-cosmos/pom.xml
@@ -92,7 +92,7 @@ Licensed under the MIT License.
     <dependency>
       <groupId>io.projectreactor</groupId>
       <artifactId>reactor-core</artifactId>
-      <version>3.3.2.RELEASE</version> <!-- {x-version-update;cosmos_io.projectreactor:reactor-core;external_dependency} -->
+      <version>3.3.3.RELEASE</version> <!-- {x-version-update;cosmos_io.projectreactor:reactor-core;external_dependency} -->
       <exclusions>
         <exclusion>
           <groupId>io.netty</groupId>


sdk/cosmos/azure-cosmos-benchmark/pom.xml
```diff
@@ -90,7 +90,7 @@ Licensed under the MIT License.
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative</artifactId>
-      <version>2.0.26.Final</version> {x-version-update;io.netty:netty-tcnative;external_dependency}
+      <version>2.0.27.Final</version> {x-version-update;io.netty:netty-tcnative;external_dependency}
```
alzimmermsft: Should this use the Cosmos specific tag instead of the general one? Should we use this PR to get the general and Cosmos versions inline?

sdk/cosmos/azure-cosmos/pom.xml
```diff
@@ -244,7 +244,7 @@ Licensed under the MIT License.
     <dependency>
       <groupId>io.projectreactor.netty</groupId>
       <artifactId>reactor-netty</artifactId>
-      <version>0.9.4.RELEASE</version> <!-- {x-version-update;io.projectreactor.netty:reactor-netty;external_dependency} -->
+      <version>0.9.5.RELEASE</version> <!-- {x-version-update;io.projectreactor.netty:reactor-netty;external_dependency} -->
```
alzimmermsft: Same question.

sdk/cosmos/azure-cosmos-benchmark/pom.xml
kushagraThapar: @alzimmermsft please update cosmos versions to reactor-core - 3.3.3
in external_dependencies.txt file. 
For now, keep using cosmos versions.

sdk/cosmos/azure-cosmos/pom.xml
kushagraThapar: Same response.

sdk/cosmos/azure-cosmos-benchmark/pom.xml
srnagar: Looks like this dependency is commented out but in general, since cosmos is preparing for release, it's better to check with the Cosmos team if they are willing to update the dependency version close to the release. But after release, we should consolidate and use the same versions in Cosmos and other track 2 libraries.

sdk/cosmos/azure-cosmos/pom.xml
srnagar: Same answer :)

sdk/cosmos/azure-cosmos-benchmark/pom.xml
kushagraThapar: I agree with @srnagar  Yeah, let's use the same versions, but have diff sources - cosmos will take from cosmos dependency section.. 
We have an issue with netty-tcnative.. We are investigating it ..

alzimmermsft: I'll update the version 😄

eng/versioning/external_dependencies.txt
```diff
@@ -180,7 +180,7 @@ cosmos_io.netty:netty-handler;4.1.45.Final
 cosmos_io.netty:netty-handler-proxy;4.1.45.Final
 cosmos_io.netty:netty-tcnative;2.0.27.Final
 cosmos_io.netty:netty-transport-native-epoll;4.1.45.Final
-cosmos_io.projectreactor:reactor-core;3.3.2.RELEASE
+cosmos_io.projectreactor:reactor-core;3.3.3.RELEASE
```
JimSuplizio: @alzimmermsft @kushagraThapar 
Right now, with these changes, every cosmos_io.* variable is now duplicated since the the common versions of these external dependencies has been brought forward. The cosmos* variables, that are duplicates, should no longer be here.

alzimmermsft: Cosmos is currently investigating an issue where they would like to continue using their own dependency set, maybe this could be unified once Cosmos releases their GA'd library.

@kushagraThapar

JimSuplizio: That's fine for the meantime and I'll approve the PR but please file an issue for Cosmos to remove the cosmos_* dependencies that are the same version as the common dependency versions before they GA.

PR #9114, opened 2020-03-16
Nio azure resource

rickle-msft:
This PR is purely refactor and should change no functionality. It became clear that most of the helper methods written up to this point were generally operations on some sort of remote resource, so this PR introduces the abstraction of AzureResource to be a central, logical location for all those methods.

### DIFF
--- a/sdk/storage/azure-storage-blob-nio/pom.xml
+++ b/sdk/storage/azure-storage-blob-nio/pom.xml
@@ -244,12 +244,12 @@
                 --add-exports com.azure.core/com.azure.core.implementation.serializer.jackson=ALL-UNNAMED
                 --add-exports com.azure.core/com.azure.core.implementation.util=ALL-UNNAMED
                 --add-opens com.azure.storage.common/com.azure.storage.common.implementation=ALL-UNNAMED
-                --add-opens com.azure.storage.common/com.azure.storage.common.sas=ALL-UNNAMED
-                --add-opens com.azure.storage.blob/com.azure.storage.blob.nio=ALL-UNNAMED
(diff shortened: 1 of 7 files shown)


sdk/storage/azure-storage-blob-nio/src/test/resources/session-records/AzureFileSystemProviderSpecfilesystemprovidercopynorootdir.json
```diff
@@ -1,11 +1,19 @@
 {
   "networkCallRecords" : [ {
     "Method" : "GET",
+<<<<<<< HEAD
```
alzimmermsft: Merge conflict here

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureResource.java
```diff
@@ -0,0 +1,249 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.storage.blob.nio;
```
alzimmermsft: If this class is for internal use only would it make sense to move it into `implementation`?

rickle-msft: I'm not sure. It's pretty integral to some of the functionality and seems like it's of the same kind as the other types in the package. I think I'll leave it here for now, and if it seems it doesn't make sense in the package any more, I can move it later.

PR #9123, opened 2020-03-16
Generalized azurite endpoints

### LINKED ISSUE #7909: [BUG] Support non-IP address based development URLs
**Describe the bug**

Since migrating to version 12 of the SDK it will no longer connect to Azurite in a docker container.

```
dm-store_1                        | 2020-02-03 13:51:19.946 ERROR 1 --- [nio-8080-exec-2] u.g.h.d.errorhandler.ApiErrorAttributes  : Status code 501, "Path or Http-Method does not match any emulated command.
dm-store_1                        | Possible causes include
dm-store_1                        |   - missing account name path parameter
dm-store_1                        |   - Unsupported Http-Method on resource
dm-store_1                        |   - Unsupported / Not implemented "comp" query parameter on resource"
```

**To Reproduce**

Create a docker compose file with Azurite and try to access it using it's container name.  

For example: https://github.com/hmcts/rpa-em-ccd-orchestrator/blob/master/docker-compose-dependencies.yml#L152-L202

***Code Snippet***

This is the change that causes the problem:

https://github.com/Azure/azure-sdk-for-java/pull/6156/files#diff-c498bc9cb41a9cfda139edbe2bd692c9R105-R110

**Expected behavior**

It should be possible to use Azurite with a non-IP address connection string.

**Screenshots**

N/A

 - OS: Ubuntu 18.04
 - IDE : IntelliJ
 - Version of the Library used: 12.3.0

**Additional context**

Relates to #6140

**Information Checklist**
- [x] Bug Description Added
- [x] Repro Steps Added
- [x] Setup information Added

rickle-msft:
Resolves #7909

I ended up doing this is the way I explicitly said I didn't want to because I think the other way I proposed in the issue would leave accountName as null on the clients, which isn't good behavior. If there are thoughts on a better way to detect this, I am open to them.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
@@ -476,6 +476,4 @@ public final class BlobUrlParts {
 
         return retVals;
     }
-
-
 }
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
@@ -18,8 +18,7 @@ import java.net.URL;
 public class ModelHelper {
 
     /**
-     * Determines whether or not the passed authority is IP style, that is it is of the format
-     * {@code <host>:<port>}.
+     * Determines whether or not the passed authority is IP style, that is, it is of the format {@code <host>:<port>}.
      *
      * @param authority The authority of a URL.
      * @throws MalformedURLException If the authority is malformed.


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
```diff
@@ -15,8 +18,18 @@
...
+    /**
+     * Determines whether or not the passed authority is IP style, that is it is of the format
+     * "&lt;host&gt;:&lt;port&gt;".
+     *
+     * @param authority The authority of a URL.
+     * @throws MalformedURLException If the authority is malformed.
+     * @return Whether the authority is IP style.
+     */
+    public static boolean determineAuthorityIsIpStyle(String authority) throws MalformedURLException {
+        return new URL("http://" +  authority + "/path").getPort() != -1;
```
alzimmermsft: Any reason `/path` is required?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
```diff
@@ -15,8 +18,18 @@
  * RESERVED FOR INTERNAL USE.
  */
 public class ModelHelper {
-    public static final Pattern IP_V4_URL_PATTERN = Pattern
-        .compile("(?:\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3})|(?:localhost)");
+
+    /**
+     * Determines whether or not the passed authority is IP style, that is it is of the format
+     * "&lt;host&gt;:&lt;port&gt;".
```
alzimmermsft:
```suggestion
     * {@code "<host><port>"}.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
```diff
@@ -102,7 +103,12 @@ public String getHost() {
      */
     public BlobUrlParts setHost(String host) {
         this.host = host;
-        this.isIpUrl = ModelHelper.IP_V4_URL_PATTERN.matcher(host).find();
+        try {
+            this.isIpUrl = ModelHelper.determineAuthorityIsIpStyle(host);
+        } catch (MalformedURLException e) {
+            throw logger.logExceptionAsError(new IllegalStateException("Authority is malformed. Host: "
+                + host));
+        }
```
alzimmermsft: Does this issue affect Queues as well? I know it also supports IP-style URLs in Azurite.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
rickle-msft: It was a required parameter for an overload of URL constructor that I was initially trying to use, but I can try this without it and see if it still works

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
rickle-msft: Yea I should probably bring it over there as well. Thanks for thinking of that.

rickle-msft:
@alzimmermsft You wanna take another look at queues?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
```diff
@@ -338,10 +343,15 @@ public static BlobUrlParts parse(String url) {
...
-        } else {
-            parseNonIpUrl(url, parts);
+        try {
+            if (ModelHelper.determineAuthorityIsIpStyle(url.getAuthority())) {
+                parseIpUrl(url, parts);
+            } else {
+                parseNonIpUrl(url, parts);
+            }
+        } catch (MalformedURLException e) {
+            throw parts.logger.logExceptionAsError(new IllegalStateException("Authority is malformed. Host: "
```
gapra-msft: Should this be illegal state or illegal argument? isnt the url technically something the user provided to us?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
```diff
@@ -466,4 +476,6 @@ private static void parseNonIpUrl(URL url, BlobUrlParts parts) {
 
         return retVals;
     }
+
+
```
gapra-msft: nit. extra new lines

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
```diff
@@ -7,16 +7,27 @@
...
  * This class provides helper methods for common model patterns.
  *
  * RESERVED FOR INTERNAL USE.
  */
 public class ModelHelper {
-    public static final Pattern IP_V4_URL_PATTERN = Pattern
-        .compile("(?:\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3})|(?:localhost)");
+
+    /**
+     * Determines whether or not the passed authority is IP style, that is it is of the format
```
gapra-msft: Doesnt matter too much but the wording here is a little weird "that is it is", should there be a comma somewhere or something?

gapra-msft:
> @alzimmermsft You wanna take another look at queues?

Datalake will probably also have this problem. Though not all of the APIs in datalake work with Azurite, a few APIs that call into blobs do.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobUrlParts.java
rickle-msft: I think it should be IllegalState. The parameter type is URL, so that means we're starting with valid URL. The only way it's possibly malformed is if we constructed the dummy URL wrong, which is our fault. And given that we're using syntax that is known to be valid, if it somehow becomes malformed, I think it's fair to call the state illegal.

PR #9127, opened 2020-03-16
Disable Consistently Failing Live Tests

alzimmermsft:
This PR disables live tests that haven't passed once in the last 14 days. Corresponding issue #9128 was filed to re-enable these tests once the failure reason is determined and fixed.

### DIFF
--- a/sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileServiceAPITests.groovy
+++ b/sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileServiceAPITests.groovy
@@ -4,16 +4,15 @@
 package com.azure.storage.file.share
 
 import com.azure.storage.common.StorageSharedKeyCredential
-import com.azure.storage.file.share.models.ShareMetrics
 import com.azure.storage.file.share.models.ListSharesOptions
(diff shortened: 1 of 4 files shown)


alzimmermsft:
@gapra-msft or @rickle-msft mind giving this a quick review?

rickle-msft:
It looks like the CPK-N tests are failing because the accounts used don't support CPK-N. @joshfree where is the discussion on how that's being resolved?

I suspect the NPEs on the file share tests are coming from not having configured credentials for a premium file account, but that's just speculation. Are we sure the environment variables were set? Or do we have a stack trace for those errors? @alzimmermsft

alzimmermsft:
Appears that the environment variable output from the ARM template doesn't match what is expected in the test for premium file shares. The template generates using the prefix `PREMIUM_STORAGE_FILE_` where the test looks for the prefix `PREMIUM_FILE_STORAGE_`, aligning these will fix the four tests being disabled in File Shares.

alzimmermsft:
@rickle-msft I've re-enabled the premium file share tests based on the issue outlined in an earlier comment, I don't know where the conversations are at for supporting CPK-N. Any chance we could merge this PR to allow further tests in the pipeline to hopefully run?

PR #9134, opened 2020-03-16
Couple of minor fixes

JimSuplizio:
1. pom_file_version_scanner.ps1 was incorrectly using Write-Error instead of the Write-Error-With-Color function (which uses Write-Host in red). The only places using this were the functions that build the version hashtables from the files. As it turns out, the Write-Error, when running in a devops environment, causes the generated wrapper script to exit without flushing the buffer so it errors but it never ends up in the log. With this change it'll write to the host in red the script will exit with a non-zero exit code (which it already did).
2. update_versions.py - This wasn't checking if the hashtable had an entry before adding one. It does now and will throw if one is detected.


JimSuplizio:
This script only affects the build analyze step. There are a metric ton of failures in the PR from the azure-sdk-build-tools permissions issue which have nothing to do with what I've changed. Once the java - core -ci run completes successfully I'm going to submit.

PR #9138, opened 2020-03-16
Replace relative file paths with URLs

maggiepint:
Replaces relative file paths in the readme with URLs to allow for automation to copy the reference into the docs.microsoft.com ref


scbedd:
@maggiepint We do resolve relative paths prior to deployment. In my understanding we weren't going to check in removal of all the relative paths.

joshfree:
Closing per offline conversation

PR #9143, opened 2020-03-17
remove relative links

maggiepint:
Remove relative links from Readme to create automation to copy to docs.microsoft.com ref

### DIFF
--- a/sdk/storage/azure-storage-queue/README.md
+++ b/sdk/storage/azure-storage-queue/README.md
@@ -432,8 +432,8 @@ If you would like to become an active contributor to this project please follow 
 5. Create new Pull Request
 
 <!-- LINKS -->
-[source_code]: src
-[samples_readme]: src/samples/README.md
+[source_code]: https://github.com/Azure/azure-sdk-for-java/edit/master/sdk/storage/azure-storage-queue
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-queue/README.md
```diff
@@ -432,8 +432,8 @@ If you would like to become an active contributor to this project please follow
 5. Create new Pull Request
 
 <!-- LINKS -->
-[source_code]: src
-[samples_readme]: src/samples/README.md
+[source_code]: https://github.com/Azure/azure-sdk-for-java/edit/master/sdk/storage/azure-storage-queue
```
weshaggard: Why are we switching these? This goes against our general guidance at https://github.com/Azure/azure-sdk/blob/master/docs/policies/README-TEMPLATE.md.

maggiepint: Soooo :-). Right now, until we have an automation, I have to copy/paste these into the docs repo to deploy them out to docs.microsoft.com. @scbedd  and I are working our way there on the automation, but in the interim fixing every link every time is super tedious.

weshaggard: Yeah @scbedd talked to me about the samples stuff I didn't know it affected readmes as well. That is very unfortunate as those links might be broken in the future without much of a path to update them.

joshfree:
Closing PR based on offline discussion

PR #9144, opened 2020-03-17
Add Release Infrastructure for BOMs

alzimmermsft:
This PR adds release infrastructure for BOMs.

### DIFF
--- a/sdk/boms/pom.service.xml
+++ b/sdk/boms/pom.service.xml
@@ -9,8 +9,6 @@
   <packaging>pom</packaging>
   <version>1.0.0</version><!-- Need not change for every release-->
   <modules>
-    <!-- NOTE: Commenting out because we should now be binary composed.
-    -->
     <module>azure-sdk-bom</module>
   </modules>
 </project>


eng/pipelines/templates/jobs/archetype-sdk-bom.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - task: UsePythonVersion@0
+        displayName: 'Use Python 3.6'
+        inputs:
+          versionSpec: '3.6'
+
+      - template: tools/daily-dev-build-variable/daily-dev-build-variable.yml@azure-sdk-tools
+
+      - pwsh: |
+        $(Build.SourcesDirectory)/eng/versioning/pom_file_version_scanner.ps1
```
JimSuplizio: Is this step actually necessary? The BOM file doesn't have tags, the right parentage to ensure that it's been scanned and even if it did, the script would have to be modified for this specific case to allow a dependencyManagement sections (which is effectively what a BOM file is).

eng/pipelines/templates/stages/archetype-sdk-bom.yml
```diff
@@ -0,0 +1,23 @@
...
+  - stage: Build
+    jobs:
+      - template: ../jobs/archetype-sdk-bom.yml
+        parameters:
+          ServiceDirectory: ${{parameters.ServiceDirectory}}
+          SDKType: ${{parameters.SDKType}}
+          Artifacts: ${{parameters.Artifacts}}
+
+  # The Prerelease and Release stages are conditioned on whether we are building a pull request and the branch.
+  - ${{if and(ne(variables['Build.Reason'], 'PullRequest'), eq(variables['System.TeamProject'], 'internal'))}}:
```
weshaggard: As we chatted I suspect this will not work as is. I'll defer to @mitchdenny  to help deep dive into this and figure out what needs to happen.

eng/pipelines/templates/jobs/archetype-sdk-bom.yml
alzimmermsft: This can be removed, really the only checks I would want to do eagerly with the BOM is verifying that all dependencies references are in Maven central and that they align on common dependencies.

sdk/boms/pom.service.xml
```diff
@@ -0,0 +1,16 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.azure</groupId>
+  <artifactId>azure-boms</artifactId>
+  <packaging>pom</packaging>
+  <version>1.0.0</version><!-- Need not change for every release-->
+  <modules>
+    <!-- NOTE: Commenting out because we should now be binary composed.
+    -->
```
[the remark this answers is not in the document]
alzimmermsft: Removed

eng/pipelines/templates/jobs/archetype-sdk-bom.yml
alzimmermsft: Removed

eng/pipelines/templates/stages/archetype-sdk-bom.yml
mitchdenny: The POM publishing should work, and I think that signing will be a no-op. There could be some things we need to react to in the release legs of the pipeline such as version bump logic.

mitchdenny: My proposal is that given this won't break any current release pipelines is that we get it in and start tweaking things.

PR #9149, opened 2020-03-17
Resources: generate package-subscriptions-2019-11

xseeseesee:


ChenTanyi:
@weidongxu-microsoft It seems all the mgmt ci have problem.

weidongxu-microsoft:
@mitchdenny 

Any idea why this error suddenly happens in CI?
> The repository azure-sdk-build-tools in project internal could not be retrieved. Verify the name and credentials being used.

I saw this error in non mgmt CI as well. https://dev.azure.com/azure-sdk/internal/_build/results?buildId=304795&view=results

BTW, I actually do not know whether mgmt-plane CI depends on the azure-sdk-build-tools repository. If not I can remove them.

mitchdenny:
There is a service issue with pipelines right now. We have an IcM ticket tracking it.

PR #9159, opened 2020-03-17
ServiceBus Track2 - Message lock renew  by user

hemanttanwar:
Implement following feature
Read message lock value while receiving
User calls renewMessageLock API
Added test

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
@@ -97,7 +97,7 @@ public class ManagementChannel implements ServiceBusManagementNode {
     @Override
     public Mono<Instant> renewMessageLock(UUID lockToken) {
         return renewMessageLock(new UUID[]{lockToken})
-            .last()
+            .next()
             .publishOn(scheduler);
     }
 
@@ -228,7 +228,8 @@ public class ManagementChannel implements ServiceBusManagementNode {
         }).flatMapMany(responseMessage -> {
             int statusCode = RequestResponseUtils.getResponseStatusCode(responseMessage);
             if (statusCode !=  AmqpResponseCode.OK.getValue()) {
-                return Mono.error(ExceptionUtil.amqpResponseCodeToException(statusCode, "", getErrorContext()));
+                return Mono.error(ExceptionUtil.amqpResponseCodeToException(statusCode, "Could not renew the lock.",
+                    getErrorContext()));
             }
 
             return Flux.fromIterable(messageSerializer.deserializeList(responseMessage, Instant.class));
--- a/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
@@ -44,7 +44,7 @@ class ServiceBusReceiverAsyncClientIntegrationTest extends IntegrationTestBase {
 
     @Override
     protected void afterTest() {
-        dispose(receiver, sender);
+        dispose(receiver, receiverManual, sender);
     }
 
     /**
@@ -161,7 +161,7 @@ class ServiceBusReceiverAsyncClientIntegrationTest extends IntegrationTestBase {
             .map(receivedMessage -> {
                 //  keep renewing the lock whole you process the message.
                 Disposable disposable = receiverManual.renewMessageLock(receivedMessage.getLockToken())
-                    .repeat(() -> true)
+                    .repeat()
                     .delayElements(renewAfterSeconds)
                     .doOnNext(instant -> {
                         // This will ensure that we are getting valid refresh time
@@ -182,7 +182,9 @@ class ServiceBusReceiverAsyncClientIntegrationTest extends IntegrationTestBase {
                 disposable.dispose();
                 return receivedMessage;
             })))
-            .assertNext(Assertions::assertNotNull)
+            .assertNext(receivedMessage -> {
+                Assertions.assertNotNull(receivedMessage.getLockedUntil());
+            })
             .verifyComplete();
         // ensure that renew lock is called atleast once.
         Assertions.assertTrue(renewMessageLockCounter.get() > 0);
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java
@@ -197,8 +197,7 @@ class ServiceBusMessageSerializer implements MessageSerializer {
     }
 
     private List<Instant> deserializeListOfInstant(Message amqpMessage) {
-        List<Instant> listInstant = new ArrayList<>();
-
+       
         if (amqpMessage.getBody() instanceof AmqpValue) {
             AmqpValue amqpValue = ((AmqpValue) amqpMessage.getBody());
             if (amqpValue.getValue() instanceof  Map) {
@@ -206,14 +205,14 @@ class ServiceBusMessageSerializer implements MessageSerializer {
                 Map<String, Object> responseBody = (Map<String, Object>) amqpValue.getValue();
                 Object expirationListObj = responseBody.get(REQUEST_RESPONSE_EXPIRATIONS);
 
-                if (expirationListObj instanceof  Date[]) {
-                    listInstant = Arrays.stream((Date[]) expirationListObj)
+                if (expirationListObj instanceof Date[]) {
+                    return Arrays.stream((Date[]) expirationListObj)
                         .map(Date::toInstant)
                         .collect(Collectors.toList());
                 }
             }
         }
-        return listInstant;
+        return Collections.emptyList();
     }
     private List<ServiceBusReceivedMessage> deserializeListOfMessages(Message amqpMessage) {
         final List<ServiceBusReceivedMessage> messageList = new ArrayList<>();
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusManagementNode.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusManagementNode.java
@@ -59,7 +59,7 @@ public interface ServiceBusManagementNode extends AutoCloseable {
      * Queue/Subscription creation (LockDuration). If processing of the message requires longer than this duration,
      * the lock needs to be renewed. For each renewal, the lock is reset to the entity's LockDuration value.
      *
-     * @param messageLock The {@link UUID} of the mesage {@link ServiceBusReceivedMessage} to be renewed.
+     * @param messageLock The {@link UUID} of the message {@link ServiceBusReceivedMessage} to be renewed.
      * @return {@link Instant} representing the pending renew.
      */
     Mono<Instant> renewMessageLock(UUID messageLock);


sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -87,6 +96,27 @@
         }));
     }
 
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Mono<Instant> renewMessageLock(ServiceBusReceivedMessage messageForLockRenew) {
```
conniey: You don't need this override, calling the management node method with a lock token is sufficient.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -87,6 +96,27 @@
...
+            .publishOn(scheduler);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Mono<Instant> renewMessageLock(UUID lockToken) {
+        return renewMessageLock(new UUID[]{lockToken})
+            .last()
```
conniey: I don't know if this is the correct operator, what if there are multiple instants? `.single()` would throw an exception if there were more than one element.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -202,6 +232,39 @@ private Message createDispositionMessage(UUID[] lockTokens, DispositionStatus di
         return message;
     }
 
+    private Flux<Instant> renewMessageLock(UUID[] renewLockList) {
+
+        return  isAuthorized(PEEK_OPERATION_VALUE).thenMany(createRequestResponse.flatMap(channel -> {
+            UUID[] lockTokens = Arrays.stream(renewLockList)
+                //.map(ServiceBusReceivedMessage::getLockToken)
```
conniey: Is there a reason you're creating another array with the same content?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -202,6 +232,39 @@ private Message createDispositionMessage(UUID[] lockTokens, DispositionStatus di
...
+                .toArray(UUID[]::new);
+
+            Message requestMessage = createManagementMessage(REQUEST_RESPONSE_RENEWLOCK_OPERATION,
```
conniey: REQUEST_RESPONSE_RENEWLOCK_OPERATION -> REQUEST_RESPONSE_RENEW_LOCK_OPERATION

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -38,6 +43,10 @@
 import static com.azure.messaging.servicebus.implementation.ManagementConstants.REQUEST_RESPONSE_MESSAGE_COUNT;
 import static com.azure.messaging.servicebus.implementation.ManagementConstants.SERVER_TIMEOUT;
 import static com.azure.messaging.servicebus.implementation.ManagementConstants.UPDATE_DISPOSITION_OPERATION;
+import static com.azure.messaging.servicebus.implementation.ManagementConstants.REQUEST_RESPONSE_LOCKTOKENS;
```
conniey: Prefixing these constants with request_response isn't necessary. It was because all their message constants were in the same file. Ones relevant to Management are in the ManagementConstants file.

conniey: LOCKTOKENS -> LOCK_TOKENS

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -202,6 +232,39 @@ private Message createDispositionMessage(UUID[] lockTokens, DispositionStatus di
...
+
+            Message requestMessage = createManagementMessage(REQUEST_RESPONSE_RENEWLOCK_OPERATION,
+                channel.getReceiveLinkName());
+
+            requestMessage.setBody(new AmqpValue(Collections.singletonMap(REQUEST_RESPONSE_LOCKTOKENS, lockTokens)));
+            return channel.sendWithAck(requestMessage);
+        }).flatMapMany(responseMessage -> {
+            int statusCode = RequestResponseUtils.getResponseStatusCode(responseMessage);
+            List<Instant> expirationsForLocks = new ArrayList<>();
+            if (statusCode ==  REQUEST_RESPONSE_OK_STATUS_CODE) {
```
conniey: What if the status isn't OK? We should return an error.

conniey: You need to run Alt+Shift+L to reformat the code. There is incorrect spacing on a few lines.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -202,6 +232,39 @@ private Message createDispositionMessage(UUID[] lockTokens, DispositionStatus di
...
+                //.map(ServiceBusReceivedMessage::getLockToken)
+                .toArray(UUID[]::new);
+
+            Message requestMessage = createManagementMessage(REQUEST_RESPONSE_RENEWLOCK_OPERATION,
+                channel.getReceiveLinkName());
+
+            requestMessage.setBody(new AmqpValue(Collections.singletonMap(REQUEST_RESPONSE_LOCKTOKENS, lockTokens)));
+            return channel.sendWithAck(requestMessage);
+        }).flatMapMany(responseMessage -> {
+            int statusCode = RequestResponseUtils.getResponseStatusCode(responseMessage);
```
conniey: All this logic can be moved to MessageSerializer.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -202,6 +232,39 @@ private Message createDispositionMessage(UUID[] lockTokens, DispositionStatus di
...
+            List<Instant> expirationsForLocks = new ArrayList<>();
+            if (statusCode ==  REQUEST_RESPONSE_OK_STATUS_CODE) {
+                @SuppressWarnings("unchecked")
+                Map<String, Object> responseBody = (Map<String, Object>) ((AmqpValue) responseMessage
```
conniey: This unchecked cast makes me nervous. I would be more defensive about casting (ie. checking instanceof).

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -202,6 +232,39 @@ private Message createDispositionMessage(UUID[] lockTokens, DispositionStatus di
...
+                    .getBody()).getValue();
+                Object expirationListObj = responseBody.get(REQUEST_RESPONSE_EXPIRATIONS);
+
+                if (expirationListObj instanceof  Date[]) {
+                    Date[] expirations = (Date[]) expirationListObj;
```
conniey: This variable creation isn't necessary if you're going to pass it immediately to the next line.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -202,6 +232,39 @@ private Message createDispositionMessage(UUID[] lockTokens, DispositionStatus di
...
+                    expirationsForLocks =  Arrays.stream(expirations)
+                        .map(Date::toInstant)
+                        .collect(Collectors.toList());
+                }
+            }
+            return Flux.fromIterable(expirationsForLocks);
```
conniey: Instead of returning Flux.fromIterable, you could have kept it as a stream instead of collecting it twice and using fromStream

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementConstants.java
```diff
@@ -40,4 +40,8 @@
     static final String REQUEST_RESPONSE_VIA_PARTITION_KEY = "via-partition-key";
     static final String REQUEST_RESPONSE_SCHEDULE_MESSAGE_OPERATION = AmqpConstants.VENDOR + ":schedule-message";
     static final String REQUEST_RESPONSE_ASSOCIATED_LINK_NAME = "associated-link-name";
+    static final String REQUEST_RESPONSE_LOCKTOKENS = "lock-tokens";
+    static final String REQUEST_RESPONSE_RENEWLOCK_OPERATION = AmqpConstants.VENDOR + ":renew-lock";
+    static final int REQUEST_RESPONSE_OK_STATUS_CODE = 200;
```
conniey: This already exists in AmqpResponseCode.

sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ReceiveMessageAndExtendLockSample.java
```diff
@@ -0,0 +1,130 @@
...
+    private ServiceBusReceiverAsyncClient consumer;
+
+
+    @BeforeEach
+    public void setup() {
+        MockitoAnnotations.initMocks(this);
+        receiveLinkMono = Mono.fromCallable(() -> amqpReceiveLink);
+
+        when(amqpReceiveLink.receive()).thenReturn(messageProcessor);
+        //when(amqpReceiveLink.getErrors()).thenReturn(errorProcessor);
```
conniey: Remove these commented lines

sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ReceiveMessageAndExtendLockSample.java
```diff
@@ -0,0 +1,130 @@
...
+        //when(amqpReceiveLink.getConnectionStates()).thenReturn(endpointProcessor);
+        //when(amqpReceiveLink.getShutdownSignals()).thenReturn(shutdownProcessor);
+
+        String connectionString = System.getenv("AZURE_SERVICEBUS_CONNECTION_STRING")
```
conniey: Why not use a fake connection string rather than reading from the environment.

sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ReceiveMessageAndExtendLockSample.java
```diff
@@ -0,0 +1,130 @@
...
+
+        // Arrange
+        final int numberOfEvents = 1;
+
+        // Act & Assert
+        StepVerifier.create(
+            consumer.receive()
+                .take(numberOfEvents))
+            .assertNext(receivedMessage -> {
+                AtomicReference<Instant> timeToRefresh = new AtomicReference<>(receivedMessage.getLockedUntil());
```
conniey: Asserts are for checking if the message you received is correct. You are doing a lot more logic here.
Consider using `.then(Runnable)`

sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ReceiveMessageAndExtendLockSample.java
```diff
@@ -0,0 +1,130 @@
...
+                        timeToRefresh.set(instant);
+                    });
+
+                // processing the messaging
+                int count = 0;
+                while (count < 15) {
+                    ++count;
+                    log(count + ". processing message ");
+                    try {
+                        Thread.sleep(1000);
```
conniey: We shouldn't be using Thread.sleep() to test or wait. StepVerifier was created so it could wrap the async operations and wait correctly.

sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ReceiveMessageAndExtendLockSample.java
```diff
@@ -0,0 +1,130 @@
...
+
+                    }
+                }
+                log("processing done");
+                renewDisposable.dispose();
+            }).verifyComplete();
+    }
+
+    private void log(String message) {
+        System.out.println(message);
```
conniey: Use clientlogger

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
hemanttanwar: removing it.

sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ReceiveMessageAndExtendLockSample.java
```diff
@@ -0,0 +1,140 @@
...
+        // 4. Copying the connection string from the policy's properties.
+        String connectionString = "<< CONNECTION STRING FOR THE SERVICE BUS QUEUE or TOPIC >>";
+        connectionString = "Endpoint=sb://sbtrack2-hemanttest-prototype.servicebus.windows.net/;SharedAccessKeyName=manage;SharedAccessKey=T3wSc5Zp91BC1kw2bnLlNJYiBogrKRe+eBO0ST9ejCY=;EntityPath=hemant-test1";
+        ServiceBusReceiverAsyncClient consumer = new ServiceBusClientBuilder()
+            .connectionString(connectionString)
+            .scheduler(Schedulers.elastic())
+            .buildAsyncReceiverClient();
+
+        // Act & Assert
+        StepVerifier.create(
```
conniey: This is a test. it should be in the test folder.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusAsyncConsumer.java
```diff
@@ -20,13 +21,15 @@
     private final ServiceBusReceiveLinkProcessor amqpReceiveLinkProcessor;
     private final MessageSerializer messageSerializer;
     private final ServiceBusMessageProcessor processor;
+    private final Scheduler scheduler;
 
     public ServiceBusAsyncConsumer(ServiceBusReceiveLinkProcessor amqpReceiveLinkProcessor,
-        MessageSerializer messageSerializer, boolean isAutoComplete,
-        Function<ServiceBusReceivedMessage, Mono<Void>> onComplete,
-        Function<ServiceBusReceivedMessage, Mono<Void>> onAbandon) {
+                                   MessageSerializer messageSerializer, boolean isAutoComplete,
```
conniey: Undo this reformat. The parameters are all mushed against the right margin. I have it set in intellij not to align it.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
hemanttanwar: Normally service would return one because we are sending one lock token . But  in track1 also , they are using same logic since AMQP message comes as array in case of one lock token.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClient.java
```diff
@@ -194,6 +194,46 @@ public String getServiceBusResourceName() {
...
+     *
+     * @return The {@link Mono} the finishes this operation on service bus resource.
+     */
+    public Mono<Instant> renewMessageLock(ServiceBusReceivedMessage receivedMessage) {
+        return connectionProcessor
+            .flatMap(connection -> connection.getManagementNode(entityPath, entityType))
+            .flatMap(serviceBusManagementNode -> serviceBusManagementNode
+                .renewMessageLock(receivedMessage.getLockToken())
+                .map(instant -> {
+                    receivedMessage.setLockedUntil(instant);
```
conniey: is this what they do in track 1? update the actual parameter object itself? Since we want to make ServicebusReceivedMessage immutable, id expect a clone to be created.

hemanttanwar: Track 1: They update same object in this way.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
hemanttanwar: moved to serializer.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
```diff
@@ -131,4 +141,50 @@ void peekBatchMessagesFromSequence() {
...
+        final ServiceBusMessage message = TestUtils.getServiceBusMessage(contents, "id-1", 0);
+
+        AtomicReference<Integer> renewMessageLockCounter = new AtomicReference<>(0);
+        // Assert & Act
+        StepVerifier.create(sender.send(message).thenMany(receiverManual.receive()
+            .take(1)
+            .delayElements(renewAfterSeconds)
+            .map(receivedMessage -> {
+                //  keep renewing the lock whole you process the message.
+                Disposable disposable = receiverManual.renewMessageLock(receivedMessage.getLockToken())
```
conniey: I know you added comments and it doesn't make it reactive nor does it use StepVerifier correctly.
* You've only hooked up the `doOnNext`, which is used for side effects like logging, but you use it for validation. the first overload of `subscribe` is a better fit.
* What if the renewal fails? There is no handling of any error cases to fail the test
* What if the operation never emits a value because it was hung? You don't handle that case either.

Since we are pressed for time, can you create an issue to write tests for this?

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
```diff
@@ -131,4 +141,50 @@ void peekBatchMessagesFromSequence() {
...
+                    .repeat(() -> true)
```
conniey: You could have used. repeat() here.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
hemanttanwar: that should be fine I guess.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
hemanttanwar: https://github.com/Azure/azure-sdk-for-java/issues/9193

PR #9168, opened 2020-03-17
Decouple commons dependency 

anuchandy:
Removed azure-cosmos source dependency on external commons packages (commons-io, commons-collections4, commons-text). 

Created a local package 
com.azure.cosmos.implementation.apachecommons that contains only necessary commons types. The azure-cosmos source types are updated to use types from this package.

The non-shipping packages, sample, benchmark, and test will continue to take a dependency on external commons packages.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/ConsistencyTestsBase.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/implementation/ConsistencyTestsBase.java
@@ -4,6 +4,7 @@
 
 package com.azure.cosmos.implementation;
 
+import com.azure.cosmos.implementation.apachecommons.collections.map.UnmodifiableMap;
 import com.azure.cosmos.models.AccessCondition;
 import com.azure.cosmos.models.AccessConditionType;
 import com.azure.cosmos.BridgeInternal;
@@ -18,7 +19,6 @@ import com.azure.cosmos.models.Resource;
 import com.azure.cosmos.implementation.directconnectivity.WFConstants;
 import com.azure.cosmos.implementation.routing.PartitionKeyInternalHelper;
 import com.azure.cosmos.implementation.routing.Range;
-import org.apache.commons.collections4.map.UnmodifiableMap;
 import org.apache.commons.lang3.StringUtils;
 import org.assertj.core.api.Assertions;
 import org.testng.SkipException;


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/apachecommons/collections/AbstractCollectionDecorator.java
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Portions Copyright (c) Microsoft Corporation
```
moderakh: we should track the source library version, for tracking future fixes, etc.

Where are we tracking the source version?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/apachecommons/lang/Range.java
```diff
@@ -0,0 +1,292 @@
...
+/*
+ * Portions Copyright (c) Microsoft Corporation
+ */
+
+package com.azure.cosmos.implementation.apachecommons.lang;
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+public final class Range<T> implements Serializable {
```
moderakh: bug: this is a cosmos class why is this coming here?
@kushagraThapar

kushagraThapar: Yeah, we have our own Range.java class, which doesn't use any of the external libraries other than jackson. 

@anuchandy  can you please confirm if we didn't replace this class by mistake ?

```
package com.azure.cosmos.implementation.routing;

import com.azure.cosmos.BridgeInternal;
import com.azure.cosmos.models.JsonSerializable;
import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.databind.node.ObjectNode;

import java.io.Serializable;
import java.util.Comparator;

@JsonIgnoreProperties({ "empty", "singleValue", "hashMap" })
public final class Range<T extends Comparable<T>> extends JsonSerializable {
```

kushagraThapar: @moderakh  - The Range.java class that Anu has added should be coming from `org.apache.commons.lang3`
We use commons-lang3 Range.java in `ParallelAsync.java` and `ConsistencyTests2.java`

moderakh: ok correct. looks good thanks.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/apachecommons/collections/AbstractCollectionDecorator.java
anuchandy: I'll add source version in all common files, good catch!

anuchandy: Discussed with Kushagra. [Updated](https://github.com/Azure/azure-sdk-for-java/pull/9168/commits/69e6342ce5ad07b65f3a4adb4d510182795f29ab) the cgmanifest with the version information, this file can be used as the central place to track version, instead of adding same info in all source files.

PR #9171, opened 2020-03-17
Enable all http clients, service versions for appconfig and keyvault

sima-zhu:
Our current live tests are only testing on netty client and latest service version.

The change here is to enable KeyVault and Appconfig to test over all combination of httpclients and service versions.

If say app conifg is having service versions of V1_0, V2_0
Then we run test suites on:
1. NettyClient, V1_0
1. NettyClient, V2_0
1. OkhttpClient, V1_0
1. OkhttpClient, V2_0

There is commit to support the change. Please check [here](https://github.com/Azure/azure-sdk-for-java/commit/c07367d2b652690585cc8c28c986a6ce25991fbe)

### DIFF
--- a/sdk/keyvault/tests.yml
+++ b/sdk/keyvault/tests.yml
@@ -3,7 +3,7 @@ trigger: none
 jobs:
   - template: ../../eng/pipelines/templates/jobs/archetype-sdk-tests.yml
     parameters:
-      TimeoutInMinutes: 120
+      TimeoutInMinutes: 200
       MaxParallel: 3
       ServiceDirectory: keyvault
       EnvVars:


sdk/appconfiguration/tests.yml
```diff
@@ -6,6 +6,8 @@ jobs:
       ServiceDirectory: appconfiguration
       EnvVars:
         AZURE_TEST_MODE: RECORD
+        AZURE_TEST_HTTP_CLIENTS: ALL
+        AZURE_APPCONFIG_TEST_SERVICE_VERSIONS: ALL
```
alzimmermsft: Mind providing some background or explanation in the PR comments on what this and how it works? Based on the PR alone I'm assuming there is some internal pluming that handles running these permutations but it is not abundantly clear on what based on the code changes.

srnagar: @alzimmermsft the plumbing to support this happened in this PR - https://github.com/Azure/azure-sdk-for-java/pull/8679

sima-zhu: Updated the description. Thanks for comments.

sima-zhu:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=309364&view=results

v.s. 

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=305139&view=ms.vss-test-web.build-test-results-tab

We can see we run more tests on new live test builds.

sima-zhu:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=309511&view=results

v.s.

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=305153&view=ms.vss-test-web.build-test-results-tab

PR #9172, opened 2020-03-17
Remove Unused/Unnecessary Code from SwaggerMethodParser

### LINKED ISSUE #7577: Deprecate SwaggerMethodParser's expectsResponseBody
`SwaggerMethodParser`'s [https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/SwaggerMethodParser.java#L503](expectsResponseBody) method doesn't get used and is part of the public API, this should be investigated if it could be deprecated.

alzimmermsft:
Fixes #7577 

Removed unused methods and unnecessary checks from `SwaggerMethodParser`.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/SwaggerMethodParser.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/SwaggerMethodParser.java
@@ -3,10 +3,6 @@
 
 package com.azure.core.http.rest;
 
-import com.azure.core.implementation.UnixTime;
-import com.azure.core.implementation.http.UnexpectedExceptionInformation;
-import com.azure.core.util.Base64Url;
-import com.azure.core.util.DateTimeRfc1123;
 import com.azure.core.annotation.BodyParam;
 import com.azure.core.annotation.Delete;
 import com.azure.core.annotation.ExpectedResponses;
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/rest/SwaggerMethodParser.java
```diff
@@ -310,7 +294,7 @@ public String setPath(Object[] methodArguments) {
                 final Object methodArgument = swaggerMethodArguments[substitution.getMethodParameterIndex()];
                 String parameterValue = serialize(methodArgument);
                 if (parameterValue != null) {
-                    if (substitution.shouldEncode() && escaper != null) {
```
alzimmermsft: `escaper` was set to a constant a few lines above, will never be null.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/SwaggerMethodParser.java
```diff
@@ -322,38 +306,38 @@ public String setPath(Object[] methodArguments) {
...
     /**
      * Get the headers that have been added to this value based on the provided method arguments.
+     *
      * @param swaggerMethodArguments The arguments that will be used to create the headers' values.
      * @return An Iterable with the headers.
      */
     public Iterable<HttpHeader> setHeaders(Object[] swaggerMethodArguments) {
         final HttpHeaders result = new HttpHeaders(headers);
 
-        if (headerSubstitutions != null) {
```
alzimmermsft: `headerSubstituations` is instantiated to an empty `List` and will only be added to, will never be null.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/SwaggerMethodParser.java
```diff
@@ -550,7 +503,7 @@ private String serializeFormData(String key, Object value) {
     }
 
     private String applySubstitutions(String originalValue, Iterable<Substitution> substitutions,
-                                      Object[] methodArguments, PercentEscaper escaper) {
```
alzimmermsft: `escaper` is always passed as the same constant and this method is private.

sdk/core/azure-core/src/main/java/com/azure/core/http/rest/SwaggerMethodParser.java
```diff
@@ -560,10 +513,8 @@ private String applySubstitutions(String originalValue, Iterable<Substitution> s
                     final Object methodArgument = methodArguments[substitutionParameterIndex];
 
                     String substitutionValue = serialize(methodArgument);
-                    if (substitutionValue != null
-                        && !substitutionValue.isEmpty()
-                        && substitution.shouldEncode() && escaper != null) {
```
alzimmermsft: `escaper` is always a constant.

PR #9176, opened 2020-03-17
Added update APIs to Credential class and add https check on policy.

sima-zhu:
The change is asking from API review tool. 
https://apiview.dev/Assemblies/Review/175a6f2bd4224031b1291f364c5691f0#com.azure.search.SearchApiKeyCredential

@brjohnstmsft Want to confirm with you. We have to use https protocol to access the search client, right?

### DIFF
--- a/sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyPipelinePolicy.java
+++ b/sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyPipelinePolicy.java
@@ -7,7 +7,6 @@ import com.azure.core.http.HttpPipelineCallContext;
 import com.azure.core.http.HttpPipelineNextPolicy;
 import com.azure.core.http.HttpResponse;
 import com.azure.core.http.policy.HttpPipelinePolicy;
-import com.azure.search.SearchApiKeyCredential;
 import reactor.core.publisher.Mono;
 
 import java.util.Objects;


sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyCredential.java
```diff
@@ -11,7 +11,7 @@
  * @see SearchServiceClientBuilder
  */
 public final class SearchApiKeyCredential {
-    private final String apiKey;
+    private String apiKey;
```
alzimmermsft: Should this be made atomic to prevent race conditions if this credential is being shared? `AtomicReference<String>`?

sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyCredential.java
```diff
@@ -30,4 +30,15 @@ public SearchApiKeyCredential(String apiKey) {
     public String getApiKey() {
         return apiKey;
     }
+
+    /**
+     * Rotates the credential on the same Search Client.
```
alzimmermsft: Credentials may be shared across clients, so this will affect all clients using this credential.

```suggestion
     * Rotates the Search credential.
```

sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyPipelinePolicy.java
```diff
@@ -32,6 +32,9 @@ public SearchApiKeyPipelinePolicy(SearchApiKeyCredential apiKey) {
 
     @Override
     public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+        if ("http".equals(context.getHttpRequest().getUrl().getProtocol())) {
+            return Mono.error(new RuntimeException("token credentials require a URL using the HTTPS protocol scheme"));
```
alzimmermsft: Is this using the same exception message as credentials in AAD use? It would be nice to turn it into a proper sentence here and file an issue for the other locations.

brjohnstmsft:
@sima-zhu Yes, we only use HTTPS to call the service in the cloud. We use HTTP for calling local instances on our dev machines, but that's not a concern here.

sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyPipelinePolicy.java
sima-zhu: Thanks for catching this.

sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyCredential.java
sima-zhu: @schaabs @g2vinay Any convention on this?

sdk/search/azure-search/src/main/java/com/azure/search/implementation/SearchApiKeyPipelinePolicy.java
```diff
@@ -1,12 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-package com.azure.search;
+package com.azure.search.implementation;
```
alzimmermsft: We shouldn't move this into implementation as a user may want to use this to create their own custom pipeline. Also, this is going to trigger a CheckStyle rule to fail that validates all `HttpPipelinePolicy` implementations are in the public API.

sima-zhu: We made this class final. How do customer make use of the policy?

alzimmermsft: They may want to create their own pipeline and code such as this:

```java
HttpPipeline myCustomSearchPipeline = new HttpPipelineBuilder()
    .addPolicy(new SearchApiKeyPipelinePolicy(myCredentials)
    .addPolicy(myOtherPolicies)
    .build()
```

This is completely allowed and in some cases expected for a customer to roll their own custom pipeline to handle specific cases they want to cover, remove policies our builders add that they don't want, etc.

sdk/search/azure-search/src/main/java/com/azure/search/SearchApiKeyCredential.java
sima-zhu: Talked offline. As the api-key is at instance level. Rotate the credential for clients is thread safe.

Will leave as it is.

PR #9188, opened 2020-03-18
Persistence token cache for Mac & Linux with new MSAL ext

jianghaolu:
Here lists the set of design decisions we need to make, to go from the MSAL shared token cache accessors to a user-friendly SharedTokenCacheCredential design that works seamlessly and effortlessly across platforms and environments. 

 

### scope 

Q: What credentials are we supporting shared token cache? 

A: All user account credentials and service principal credentials. 

 

### user account credentials (public client applications) 

Q: Enable or disable reading from the shared token cache by default? 

A: Disable. 

Q: Enable or disable writing to the shared token cache by default? 

A: Disable. 

Q: What environment variable to specify which account if there are multiple? 

A: AZURE_USERNAME. 

 

### service principal credentials (confidential client applications) 

Q: Do we want to support token cache for this? 

A: Yes. 

Q: Enable or disable reading from the shared token cache by default? 

A: Disable. 

Q: Enable or disable writing from the shared token cache by default? 

A: Disable. 

Q: What environment variable to specify which service principal if there are multiple? 

A: N/A – will be specified on the ConfidentialClientApplication 

Q: If both AZURE_USERNAME and AZURE_CLIENT_ID are present, what does it do? 

A: N/A 

 

### configurations 

Q: Do we have one method to enable both read & write, or separate methods? 

A: One method. 

Q: Do we start writing to cache in SharedTokenCacheCredential now? 

A: Yes 

Q: Do we allow configuring where the token cache is stored?  

A: No 

 

### operating system specifics 

Q: Do we automatically fallback to unprotected file on Linux if Gnome keyring item is unavailable? 

A: No. Allow through a property “allowUnencryptedCache”. 

 

### storage locations for user accounts 

Q: Where to store the DPAPI encrypted file on Windows? 

A: {user.home}\AppData\Local\.IdentityService\msal.cache 

Q: Where to store the secret item in Keychain on Mac? 

A: Service = Microsoft.Developer.IdentityService, Account = MSALCache 

Q: Where to store the secret item in Gnome Keyring? 

A: Under default keyring, with item label “MSALCache”, and schema “msal.cache”. 

Q: Where to store the secret item as an unprotected file on Linux? 

A: $HOME/.IdentityService/msal.cache 

 

(Document all these to be implementations and shouldn’t be taken as a hard dependency) 

 

### storage locations for service principals 

Q: Do we merge the token cache data for service principals with the user accounts? 

A: Hopefully no, to avoid service principals peeking into the user tokens 

Q: If the answer is no for the above question, specify the locations for all questions in Article 2.1. 

A: MSALConfidentialCache, msal.confidential.cache 

 

### exceptions 

Q: What exceptions do we throw if DPAPI encrypted file cannot be found on Windows, or Keychain item is unavailable on Mac, or Gnome Keyring item is unavailable on Linux, or unprotected file is unavailable? 

A: CredentialUnavailableException for SharedTokenCacheCredential, ClientAuthenticationException for others if enabled 

Q: What exceptions do we throw if the file, or the Keychain/Keyring item is available, but cannot be parsed or error out for any other reasons? 

A: Wrap them in ClientAuthenticationException / AuthenticationFailedException. 

 

### logging 

Q: What information do we log in this process?  

A: 1) What platform we are on and 2) if the token cache access succeeded or failed. No information of where the token cache is located will be logged. 

2) for sure, 1) potentially in the future 

### transient errors 

Q: Is there a plan to detect and retry transient errors, like a race condition that MSAL fails to prevent? 

A: for now no.

### DIFF
--- a/sdk/keyvault/azure-security-keyvault-secrets/pom.xml
+++ b/sdk/keyvault/azure-security-keyvault-secrets/pom.xml
@@ -105,7 +105,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-identity</artifactId>
-      <version>1.1.0-beta.4</version> <!-- {x-version-update;com.azure:azure-identity;dependency} -->
+      <version>1.0.5</version> <!-- {x-version-update;com.azure:azure-identity;dependency} -->
       <scope>test</scope>
     </dependency>
 
--- a/sdk/keyvault/azure-security-keyvault-secrets/src/samples/java/com/azure/security/keyvault/secrets/PersistentTokenCacheDemo.java
+++ b/sdk/keyvault/azure-security-keyvault-secrets/src/samples/java/com/azure/security/keyvault/secrets/PersistentTokenCacheDemo.java
@@ -24,7 +24,7 @@ public class PersistentTokenCacheDemo {
             .build();
 
         SecretClient client = new SecretClientBuilder()
-            .vaultUrl("https://jianghaovaultb.vault.azure.net")
+            .vaultUrl("https://persistentcachedemo.vault.azure.net")
             .credential(defaultCredential)
             .buildClient();
 


sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredential.java
```diff
@@ -23,6 +27,18 @@
  */
 @Immutable
 public final class DefaultAzureCredential extends ChainedTokenCredential {
+    private static final String DEFAULT_CACHE_FILE_NAME = "msal.cache";
+    private static final Path DEFAULT_CACHE_DIRECTORY = Platform.isWindows() ?
+            Paths.get(System.getProperty("user.home"), "AppData", "Local", ".IdentityService", "msal.cache") :
+            Paths.get(System.getProperty("user.home"),".IdentityService", "msal.cache");
+    private static final String DEFAULT_KEYCHAIN_SERVICE = "Microsoft.Developer.IdentityService";
```
jianghaolu: Keychain service & account names are consistent with Python ext at https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/71b5301cb8c4768a9f439bde20750d448062944c/msal_extensions/token_cache.py#L141 and this test in .NET ext: https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/85c1e8ec283b940b1cc34eae44ac37fa645738ff/tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MsalCacheStorageTests.cs

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredential.java
```diff
@@ -23,6 +27,18 @@
  */
 @Immutable
 public final class DefaultAzureCredential extends ChainedTokenCredential {
+    private static final String DEFAULT_CACHE_FILE_NAME = "msal.cache";
+    private static final Path DEFAULT_CACHE_DIRECTORY = Platform.isWindows() ?
+            Paths.get(System.getProperty("user.home"), "AppData", "Local", ".IdentityService", "msal.cache") :
```
jianghaolu: This is consistent with Visual Studio: https://github.com/microsoft/Git-Credential-Manager-Core/blob/master/src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs#L228

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredential.java
```diff
@@ -23,6 +27,18 @@
...
+            Paths.get(System.getProperty("user.home"),".IdentityService", "msal.cache");
```
jianghaolu: This is invented by me - open for discussion

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredential.java
```diff
@@ -23,6 +27,18 @@
...
+    private static final String DEFAULT_KEYCHAIN_SERVICE = "Microsoft.Developer.IdentityService";
+    private static final String DEFAULT_KEYCHAIN_ACCOUNT = "MSALCache";
+    private static final String DEFAULT_KEYRING_NAME = "default";
+    private static final KeyRingItemSchema DEFAULT_KEYRING_SCHEMA = KeyRingItemSchema.GENERIC_SECRET;
+    private static final String DEFAULT_KEYRING_ITEM_NAME = DEFAULT_KEYCHAIN_ACCOUNT;
+    private static final String DEFAULT_KEYRING_ATTR_NAME = "MsalClientID";
```
jianghaolu: Line 39 and 40 are also from the test https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/85c1e8ec283b940b1cc34eae44ac37fa645738ff/tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MsalCacheStorageTests.cs

jianghaolu:
Linux tests are failing on CI but passing on Ubuntu Desktop 16.04. Possibly the CI doesn't have libsecret installed. Looking into possible mitigations while the PR is ready for review.

schaabs:
@jianghaolu CI machines are not going to have libsecret installed. We have run into the same issue with the `VisualStudioCodeCredential` live tests. In this case we added mocking and disabled the live tests but left them so they could be run manually for now.

eng/versioning/version_client.txt
```diff
@@ -19,8 +19,9 @@ com.azure:azure-cosmos-examples;4.0.1-beta.1;4.0.1-beta.1
 com.azure:azure-cosmos-benchmark;4.0.1-beta.1;4.0.1-beta.1
 com.azure:azure-data-appconfiguration;1.1.1;1.2.0-beta.1
 com.azure:azure-e2e;1.0.0-beta.1;1.0.0-beta.1
-com.azure:azure-identity;1.0.4;1.1.0-beta.3
+com.azure:azure-identity;1.0.5;1.1.0-beta.4
 com.azure:azure-identity-perf;1.0.0-beta.1;1.0.0-beta.1
+>>>>>>> 3a255eb005748305722895bc50d3dd4ff5789bdb
```
weshaggard: Looks like a bad merge here.

jianghaolu: Thanks! But you seems to be reviewing an outdated diff.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AadCredentialBuilderBase.java
```diff
@@ -70,4 +70,31 @@ public T executorService(ExecutorService executorService) {
...
         return (T) this;
     }
+
+    /**
+     * Sets whether to use an unprotected file specified by <code>cacheFileLocation()</code> instead of
+     * Gnome keyring on Linux. This is false by default.
+     *
+     * @param useUnprotectedFileOnLinux whether to use an unprotected file for cache storage.
+     *
+     * @return The updated T object.
```
srnagar, line 80: nit: to be consistent with the rest of the javadoc change this to 
`@return An updated instance of this builder with useUnprotectedFileOnLinux set as specified.`

sdk/identity/azure-identity/src/main/java/com/azure/identity/AadCredentialBuilderBase.java
```diff
@@ -70,4 +70,31 @@ public T executorService(ExecutorService executorService) {
...
+        this.identityClientOptions.setUseUnprotectedTokenCacheFileOnLinux(useUnprotectedFileOnLinux);
+        return (T) this;
+    }
+
+    /**
+     * Disable using the shared token cache.
+     *
+     * @param disabled whether to disable using the shared token cache.
+     *
+     * @return The updated identity client options.
```
srnagar, line 93: same as above

sdk/identity/azure-identity/src/main/java/com/azure/identity/KeyringItemSchema.java
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.identity;
+
+/**
+ * An expandable enum for types of item schema in a Keyring.
+ */
+public final class KeyringItemSchema {
```
srnagar, line 9: Since the javadoc mentions that this is an expandable enum, should this class extend from `ExpandableStringEnum`?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -490,6 +529,86 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+                                .switchIfEmpty(Mono.defer(() -> Mono.fromFuture(() -> {
+                                        try {
+                                            return getPublicClientApplication(false).acquireTokenSilently(forceParams);
+                                        } catch (MalformedURLException e) {
+                                            throw logger.logExceptionAsWarning(new RuntimeException(e));
+                                        }
+                                    }
+                                ).map(result -> new MsalToken(result, options))));
+                    } catch (MalformedURLException e) {
+                        e.printStackTrace();
```
srnagar, line 606: Use log instead of `e.printStackTrace()`

sdk/identity/pom.xml
```diff
@@ -12,5 +12,6 @@
     <!-- NOTE: Commenting out becuase we are now binary composed.
     -->
     <module>azure-identity</module>
+    <module>perf-test</module>
```
srnagar, line 15: Use the artifact id here: `azure-identity-perf`

sdk/identity/azure-identity/src/main/java/com/azure/identity/AadCredentialBuilderBase.java
jianghaolu: Updated

sdk/identity/azure-identity/src/main/java/com/azure/identity/AadCredentialBuilderBase.java
jianghaolu: Updated

sdk/identity/azure-identity/src/main/java/com/azure/identity/KeyringItemSchema.java
jianghaolu: Changed to extend from `ExpandableStringEnum`

sdk/identity/pom.xml
jianghaolu: Updated to azure-identity-perf

sdk/identity/azure-identity/src/main/java/com/azure/identity/AadCredentialBuilderBase.java
```diff
@@ -70,4 +70,31 @@ public T executorService(ExecutorService executorService) {
...
+
+    /**
+     * Disable using the shared token cache.
+     *
+     * @param disabled whether to disable using the shared token cache.
+     *
+     * @return An updated instance of this builder with if the shared token cache disabled specified.
+     */
+    @SuppressWarnings("unchecked")
+    public T disableSharedTokenCache(boolean disabled) {
```
schaabs, line 96: I think that persistence of the token cache should be opt-in rather than opt out. Also I'm not sure we can put this on the base builder for all credentials, for instance what does it mean to create a `SharedTokenCachceCredential` with the shared token cache disabled.

sdk/identity/azure-identity/src/main/java/com/azure/identity/KeyringItemSchema.java
```diff
@@ -0,0 +1,20 @@
...
+
+package com.azure.identity;
+
+import com.azure.core.util.ExpandableStringEnum;
+
+/**
+ * An expandable enum for types of item schema in a Keyring.
+ */
+
+public final class KeyringItemSchema extends ExpandableStringEnum<KeyringItemSchema> {
```
schaabs, line 12: Is this meant to be part of the public API? I'd prefer to keep these details internal.

jianghaolu: Good catch - It was made public to allow configurations - but since for now we are not allowing users to configure it I'll hide it.

sdk/identity/azure-identity/src/main/java/com/azure/identity/AadCredentialBuilderBase.java
jianghaolu: For SharedTokenCacheCredential, yes you can disable shared token cache, but assuming no one will do this does save us a ton of duplicate code.

jianghaolu:
Closing this PR for now as too many design decisions are still pending. It's been going back and forth multiple times in the past 6 weeks with multiple variables changing. Conclusion: this is not ready.

sdk/identity/azure-identity-perf/src/main/java/com/azure/identity/perf/core/ServiceTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.azure.core.credential.TokenRequestContext;
+import com.azure.identity.InteractiveBrowserCredential;
+import com.azure.identity.InteractiveBrowserCredentialBuilder;
+import com.azure.perf.test.core.PerfStressOptions;
+import com.azure.perf.test.core.PerfStressTest;
+import reactor.core.publisher.Mono;
+
+public abstract class ServiceTest<TOptions extends PerfStressOptions> extends PerfStressTest<TOptions> {
+    protected static final String CLI_CLIENT_ID = "04b07795-8ddb-461a-bbee-02f9e1bf7b46";
```
g2vinay, line 14: do we need them to be static ?

sdk/identity/azure-identity/src/main/java/com/azure/identity/AadCredentialBuilderBase.java
g2vinay: ManagedIdentityCredential and DefaultAzureCredential inherit from CredentialBuilderBase, so they will not inherit this setter currently.

sdk/identity/azure-identity-perf/src/main/java/com/azure/identity/perf/core/ServiceTest.java
jianghaolu: I'm simply defining it as a constant: https://stackoverflow.com/questions/9639007/defining-constant-string-in-java

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -144,7 +153,22 @@
...
+                try {
+                    publicClientApplicationBuilder.setTokenCacheAccessAspect(
+                            new PersistenceTokenCacheAccessAspect(options.getPersistenceSettings()));
+                } catch (Throwable t) {
+                    if (sharedTokenCacheCredential) {
+                        throw logger.logExceptionAsError(new CredentialUnavailableException(
+                            "Shared token cache is unavailable in this environment.", t));
+                    } else {
+                        throw logger.logExceptionAsError(new ClientAuthenticationException(
+                            "Shared token cache is unavailable in this environment.", null, t));
```
alzimmermsft, line 166: Is this exception message meant to be the same as the one above?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -380,29 +404,44 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
      */
     public Mono<MsalToken> authenticateWithUsernamePassword(TokenRequestContext request,
                                                             String username, String password) {
-        return Mono.fromFuture(publicClientApplication.acquireToken(
+        return Mono.fromFuture(() -> getPublicClientApplication(false).acquireToken(
             UserNamePasswordParameters.builder(new HashSet<>(request.getScopes()), username, password.toCharArray())
                 .build()))
-            .map(ar -> new MsalToken(ar, options));
+            .onErrorResume(t -> Mono.error(new ClientAuthenticationException("Failed to acquire token with username "
```
alzimmermsft, line 410: Should this be `onErrorMap`? This logic looks to be more of an error mapping than an alternate resume from an exception given this will propagate an error.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -380,29 +404,44 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
      */
     public Mono<MsalToken> authenticateWithUsernamePassword(TokenRequestContext request,
                                                             String username, String password) {
-        return Mono.fromFuture(publicClientApplication.acquireToken(
+        return Mono.fromFuture(() -> getPublicClientApplication(false).acquireToken(
```
alzimmermsft, line 407: Do we ever want this future to be cancellable? If yes, we will need to add a `doFinally` operator that will check for cancelled to cancel the future.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -380,29 +404,44 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+            forceParameters = SilentParameters.builder(new HashSet<>(request.getScopes())).forceRefresh(true).build();
         }
         return Mono.defer(() -> {
             try {
-                return Mono.fromFuture(publicClientApplication.acquireTokenSilently(parameters))
-                        .map(ar -> new MsalToken(ar, options));
+                return Mono.fromFuture(getPublicClientApplication(false).acquireTokenSilently(parameters))
+                        .map(ar -> new MsalToken(ar, options))
+                        .filter(t -> !t.isExpired())
+                        .switchIfEmpty(Mono.defer(() -> Mono.fromFuture(() -> {
```
alzimmermsft, line 437: The defer isn't needed here.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -380,29 +404,44 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+                    .forceRefresh(true).build();
         } else {
             parameters = SilentParameters.builder(new HashSet<>(request.getScopes())).build();
+            forceParameters = SilentParameters.builder(new HashSet<>(request.getScopes())).forceRefresh(true).build();
         }
         return Mono.defer(() -> {
             try {
-                return Mono.fromFuture(publicClientApplication.acquireTokenSilently(parameters))
-                        .map(ar -> new MsalToken(ar, options));
+                return Mono.fromFuture(getPublicClientApplication(false).acquireTokenSilently(parameters))
```
alzimmermsft, line 434: Should this use a future supplier as below?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -380,29 +404,44 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
      * @return a Publisher that emits an AccessToken
      */
-    public Mono<MsalToken> authenticateWithUserRefreshToken(TokenRequestContext request, MsalToken msalToken) {
+    public Mono<MsalToken> authenticateWithMsalAccount(TokenRequestContext request, IAccount account) {
         SilentParameters parameters;
-        if (msalToken.getAccount() != null) {
-            parameters = SilentParameters.builder(new HashSet<>(request.getScopes()), msalToken.getAccount()).build();
+        SilentParameters forceParameters;
+        if (account != null) {
+            parameters = SilentParameters.builder(new HashSet<>(request.getScopes()), account).build();
```
alzimmermsft, line 425: Does account here have to be non-null? If it can be null, could the if/else block be merged to always pass `account`?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -380,29 +404,44 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
-                return Mono.fromFuture(publicClientApplication.acquireTokenSilently(parameters))
-                        .map(ar -> new MsalToken(ar, options));
+                return Mono.fromFuture(getPublicClientApplication(false).acquireTokenSilently(parameters))
+                        .map(ar -> new MsalToken(ar, options))
+                        .filter(t -> !t.isExpired())
+                        .switchIfEmpty(Mono.defer(() -> Mono.fromFuture(() -> {
+                                try {
+                                    return getPublicClientApplication(false).acquireTokenSilently(forceParameters);
+                                } catch (MalformedURLException e) {
+                                    throw logger.logExceptionAsWarning(new RuntimeException(e));
```
alzimmermsft, line 441: Need to use `Exceptions.propagate` here.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -425,8 +464,9 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
             DeviceCodeFlowParameters parameters = DeviceCodeFlowParameters.builder(new HashSet<>(request.getScopes()),
                 dc -> deviceCodeConsumer.accept(new DeviceCodeInfo(dc.userCode(), dc.deviceCode(),
                     dc.verificationUri(), OffsetDateTime.now().plusSeconds(dc.expiresIn()), dc.message()))).build();
-            return publicClientApplication.acquireToken(parameters);
-        }).map(ar -> new MsalToken(ar, options));
+            return getPublicClientApplication(false).acquireToken(parameters);
+        }).onErrorResume(t -> Mono.error(new ClientAuthenticationException("Failed to acquire token with device code",
```
alzimmermsft, line 468: `onErrorMap`?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -439,11 +479,12 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
      */
     public Mono<MsalToken> authenticateWithAuthorizationCode(TokenRequestContext request, String authorizationCode,
                                                              URI redirectUrl) {
-        return Mono.fromFuture(() -> publicClientApplication.acquireToken(
+        return Mono.fromFuture(() -> getPublicClientApplication(false).acquireToken(
             AuthorizationCodeParameters.builder(authorizationCode, redirectUrl)
                 .scopes(new HashSet<>(request.getScopes()))
                 .build()))
-            .map(ar -> new MsalToken(ar, options));
+            .onErrorResume(t -> Mono.error(new ClientAuthenticationException("Failed to acquire token with "
```
alzimmermsft, line 486: `onErrorMap`?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -490,6 +531,56 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
             });
     }
 
+    /**
+     * Gets token from shared token cache
+     * */
+    public Mono<AccessToken> authenticateWithSharedTokenCache(TokenRequestContext request, String username) {
+        // find if the Public Client app with the requested username exists
+        return Mono.fromFuture(() -> getPublicClientApplication(true).getAccounts())
+                .onErrorResume(t -> Mono.error(new CredentialUnavailableException(
```
alzimmermsft, line 540: `onErrorMap`?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Yes - I'm not aware of this operator, changing them now

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Removed.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Yeah we should

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Unfortunately in the first builder overload the account is required to be non-null.

jianghaolu: But it does seem like I can add account afterwards

jianghaolu:
Thanks for the review Alan! I've addressed the comments.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Reorganized a little bit to avoid creating 2 parameters upfront

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Synced offline - due to the small cost of the getToken() call we will skip the hassle to cancel the future here

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: They are indeed the same - just wrapping in different exceptions to be handled differently downstream.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -490,6 +530,56 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+
+                    for (IAccount cached : set) {
+                        if (username == null || username.equals(cached.username())) {
+                            if (!accounts.containsKey(cached.homeAccountId())) { // only put the first one
+                                accounts.put(cached.homeAccountId(), cached);
+                            }
+                        }
+                    }
+
+                    if (accounts.size() == 0) {
```
g2vinay, line 553: This logic here was updated before. Please get the one from master.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -490,6 +530,56 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+                        if (username == null) {
+                            return Mono.error(new CredentialUnavailableException("No accounts were discovered in the "
+                                    + "shared token cache. To fix, authenticate through tooling supporting azure "
+                                    + "developer sign on."));
+                        } else {
+                            return Mono.error(new CredentialUnavailableException(String.format("User account '%s' was "
+                                    + "not found in the shared token cache. Discovered Accounts: [ '%s' ]", username,
```
g2vinay, line 560: Error messages and checks were changed here and below. 
This one is PII exposure.

jianghaolu: Thanks! I lost the update during the merge.

sdk/identity/azure-identity/src/main/java/com/azure/identity/DeviceCodeCredential.java
```diff
@@ -51,7 +51,7 @@
     public Mono<AccessToken> getToken(TokenRequestContext request) {
         return Mono.defer(() -> {
             if (cachedToken.get() != null) {
-                return identityClient.authenticateWithUserRefreshToken(request, cachedToken.get())
+                return identityClient.authenticateWithMsalAccount(request, cachedToken.get().getAccount())
                     .onErrorResume(t -> Mono.empty());
```
conniey, line 55: Would you want this to resume on an error and return an empty Mono? I'd expect the error to be propagated downstream and allow downstream subscribers to do with that error what they will. In  `globalSetupAsync` this is call is chained with `.then()`, so it would hide the error.

There are similar instances in other credentials below.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -380,33 +403,47 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+        .switchIfEmpty(Mono.fromFuture(() -> {
+            SilentParameters.SilentParametersBuilder forceParametersBuilder = SilentParameters.builder(
+                new HashSet<>(request.getScopes())).forceRefresh(true);
+            if (account != null) {
+                forceParametersBuilder = forceParametersBuilder.account(account);
+            }
+            try {
+                return getPublicClientApplication(false).acquireTokenSilently(forceParametersBuilder.build());
+            } catch (MalformedURLException e) {
+                throw logger.logExceptionAsError(Exceptions.propagate(e));
```
conniey, line 444: Since you're using futures, it may be easier to follow using:
```java
var completableFuture = new CompletableFuture<MsalToken>();
completableFuture.completeExceptionally(logger.logExceptionAsError(Exceptions.propagate(e));
return completableFuture;
```

sdk/identity/azure-identity/src/main/java/com/azure/identity/DeviceCodeCredential.java
jianghaolu: Yeah it should be probably be handled downstream - it's out of the scope of this PR but I can see if it can be done with minimal footprint

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -490,6 +529,56 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+
+                    for (IAccount cached : set) {
+                        if (username == null || username.equals(cached.username())) {
+                            if (!accounts.containsKey(cached.homeAccountId())) { // only put the first one
+                                accounts.put(cached.homeAccountId(), cached);
+                            }
+                        }
+                    }
+
+                    if (set.size() == 0) {
```
conniey, line 552: nit: set.isEmpty()

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -490,6 +529,56 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+                        return Mono.error(new CredentialUnavailableException("SharedTokenCacheCredential "
```
conniey, line 553: Should we use `return FluxUtils.monoError(logger, new ....)`? Or is it preferable not to log it

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -490,6 +529,56 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+                    }
+
+                    if (CoreUtils.isNullOrEmpty(username)) {
+                        return Mono.error(new CredentialUnavailableException("SharedTokenCacheCredential "
+                                + "authentication unavailable. Multiple accounts were found in the cache. Use "
+                                + "username and tenant id to disambiguate."));
+                    }
+
+                    if (accounts.size() != 1) {
+                        if (accounts.size() == 0) {
```
conniey, line 564: nit: Same with .isEmpty() here. But I won't lose sleep. :)

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -109,8 +113,14 @@
...
-        if (clientId == null) {
-            this.publicClientApplication = null;
+    }
+
+    private PublicClientApplication getPublicClientApplication(boolean sharedTokenCacheCredential) {
+        if (publicClientApplication != null) {
+            return publicClientApplication;
+        } else if (clientId == null) {
+            throw logger.logExceptionAsError(new IllegalArgumentException(
+                "A non-null value for client ID must be providedfor user authentication."));
```
conniey, line 123: providedfor -> provided for

sdk/identity/azure-identity/src/test/java/com/azure/identity/implementation/IdentityClientIntegrationTests.java
```diff
@@ -51,7 +51,7 @@ public void deviceCodeCanGetToken() {
         Assert.assertNotNull(token.getToken());
         Assert.assertNotNull(token.getExpiresAt());
         Assert.assertFalse(token.isExpired());
-        token = client.authenticateWithUserRefreshToken(new TokenRequestContext().addScopes("https://vault.azure.net/.default"), token).block();
+        token = client.authenticateWithMsalAccount(new TokenRequestContext().addScopes("https://vault.azure.net/.default"), token.getAccount()).block();
```
conniey, line 54: These tests could be using StepVerifier rather than block.
```java
// Arrange
var context = new TokenRequestContext().addScopes("https://vault.azure.net/.default");

// Act & Assert
StepVerifier.create(client.authenticateWithMsalAccount(context, token.getAccount()))
    .assertNext(token -> {
        assertNotNull(token.getToken());
        assertNotNull(token.getExpiresAt());
    })
    .verifyComplete();
```

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Thanks Connie for the tip but in this code the exception is thrown from creating the future, not from executing it.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: We talked to Jonathan and since this IdentityClient isn't used directly by customers and we have downstream credentials to do the logging - we shouldn't log here.

sdk/identity/azure-identity/src/test/java/com/azure/identity/implementation/IdentityClientIntegrationTests.java
jianghaolu: This is out of scope - if this PR stays open next week I can fix these. But for now I'm planning to get some sleep and enjoy my day off 😸

jianghaolu:
Connie's 2 feedback are out of scope of this PR. Shouldn't block this PR so I plan to fix in a separate PR.

sdk/identity/azure-identity/src/main/java/com/azure/identity/DeviceCodeCredentialBuilder.java
```diff
@@ -29,6 +29,31 @@ public DeviceCodeCredentialBuilder challengeConsumer(
...
+    }
+
+    /**
+     * Sets whether to enable using the shared token cache. This is disabled by default.
+     *
+     * @param enabled whether to enabled using the shared token cache.
+     *
+     * @return An updated instance of this builder with if the shared token cache enabled specified.
+     */
+    public DeviceCodeCredentialBuilder enableSharedTokenCache(boolean enabled) {
```
schaabs, line 52: In .NET and python we named this option EnablePersistentCache.

sdk/identity/azure-identity/src/main/java/com/azure/identity/InteractiveBrowserCredentialBuilder.java
```diff
@@ -27,6 +27,31 @@ public InteractiveBrowserCredentialBuilder port(int port) {
...
+    }
+
+    /**
+     * Sets whether to enable using the shared token cache. This is disabled by default.
+     *
+     * @param enabled whether to enabled using the shared token cache.
+     *
+     * @return An updated instance of this builder with if the shared token cache enabled specified.
+     */
+    public InteractiveBrowserCredentialBuilder enableSharedTokenCache(boolean enabled) {
```
schaabs, line 50: Same comment about EnablePersistentCache

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -109,8 +113,14 @@
         this.tenantId = tenantId;
         this.clientId = clientId;
         this.options = options;
-        if (clientId == null) {
-            this.publicClientApplication = null;
+    }
+
+    private PublicClientApplication getPublicClientApplication(boolean sharedTokenCacheCredential) {
+        if (publicClientApplication != null) {
+            return publicClientApplication;
```
schaabs, line 120: This method is called from numerous methods on the IdentityClient some which pass `sharedTokenCacheCredential` as `true` and some as `false`. Whichever calls this method first will cache the publicClientApplication and then all subsequent calls will effectively ignore the `sharedTokenCacheCredential` parameter. I'm guessing this doesn't matter in practice because certain credentials only call certain methods, and each credential has it's own instance of `IdentityClient`, and `IdentityClient` is internal. However, this makes the functionality of this class and the classes that use it interdependent in a non-obvious way which could easily be overlooked if we try to make optimizations like sharing an instance of `IdentityClient`. I think we should consider refactoring this.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
schaabs: `acquireToken` here will most likely make an HttpRequest, would this normally be cancellable in other service clients? If so we should probably do the same here.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -144,7 +154,21 @@
...
+                } catch (Throwable t) {
+                    String message = "Shared token cache is unavailable in this environment.";
+                    if (sharedTokenCacheCredential) {
+                        throw logger.logExceptionAsError(new CredentialUnavailableException(message, t));
+                    } else {
+                        throw logger.logExceptionAsError(new ClientAuthenticationException(message, null, t));
+                    }
+                }
+            }
             this.publicClientApplication = publicClientApplicationBuilder.build();
```
schaabs, line 170: This is not thread safe. If two clients using the same credential call get token at the same time it's possible that multiple threads attempt to initialize `publicClientApplication` at the same time. In this case they might each get different instances and any tokens cached on any instances other than the final one will be lost.

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
```diff
@@ -63,92 +51,18 @@
...
-                    return Mono.fromFuture(() -> future).map(result ->
-                        new AccessToken(result.accessToken(),
-                            result.expiresOnDate().toInstant().atOffset(ZoneOffset.UTC)));
-
-                } catch (MalformedURLException e) {
-                    e.printStackTrace();
-                    return Mono.error(new RuntimeException("Token was not found"));
-                }
-            });
+        return identityClient.authenticateWithSharedTokenCache(request, username);
```
schaabs, line 66: `authenticateWithSharedTokenCache` searches the shared token cache for an account each time. Shouldn't the account be saved off after the first time this method is called so that `authenticateWithMsalAccount` can be called instead?

sdk/identity/azure-identity/src/main/java/com/azure/identity/UsernamePasswordCredentialBuilder.java
```diff
@@ -36,6 +36,31 @@ public UsernamePasswordCredentialBuilder password(String password) {
...
+    }
+
+    /**
+     * Sets whether to enable using the shared token cache. This is disabled by default.
+     *
+     * @param enabled whether to enabled using the shared token cache.
+     *
+     * @return An updated instance of this builder with if the shared token cache enabled specified.
+     */
+    public UsernamePasswordCredentialBuilder enableSharedTokenCache(boolean enabled) {
```
schaabs, line 59: We hadn't added the ability to use the shared token cache to the `UsernamePasswordCredential` in .NET and Python yet. I'm not totally against it, but I think it adds limited value since `UsernamePasswordCredential` has no user interaction.

sdk/identity/azure-identity/src/main/java/com/azure/identity/DeviceCodeCredentialBuilder.java
g2vinay: updated

sdk/identity/azure-identity/src/main/java/com/azure/identity/InteractiveBrowserCredentialBuilder.java
g2vinay: updated

sdk/identity/azure-identity/src/main/java/com/azure/identity/SharedTokenCacheCredential.java
g2vinay: added the logic.

sdk/identity/azure-identity/src/main/java/com/azure/identity/UsernamePasswordCredentialBuilder.java
g2vinay: we'll be keeping it for preview, as discussed.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: Opened an issue #10675

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: Opened an issue #10675

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: As discussed, there's nothing much to gain to chain doOnCancel signal to CompletableFuture, as it gets triggered right away and not when subscribed to.

jianghaolu:
Thanks for the comments. They are extremely helpful. However, some comments may be better suited as general issues for the Identity library for Java and should be considered out of scope for this PR.

I'd like to narrow down the comments to only applicable for the changes in this PR. For the others, thanks @g2vinay for opening separate issues.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: Resolving as separate issues opened

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
jianghaolu: This method is only called once per IdentityClient and there's exactly one IdentityClient instance per credential. 
Yes it can have nicer designs, as something we can do for #10675

Issue #9190, opened 2020-03-18
ClientLogger's logExceptionAsError Checks Wrong Log Level


alzimmermsft:
The `logExceptionAsError` method in `ClientLogger` checks if the logger is able to log at `VERBOSE` instead of `ERROR`.

alzimmermsft:
Fixed in #9194

PR #9194, opened 2020-03-18
Add APIs to Log Checked Exceptions as Warning and Error

### LINKED ISSUE #7633: Add Support for Non-RuntimeExceptions in ClientLogger logException
`ClientLogger`'s `logException` methods only allow for `RuntimeException`s to be passed. This results in a conflict with a Checkstyle rule where thrown exceptions should be thrown as `logException` calls. Either add another overload which takes a `Throwable` in the cases where a checked exception is being thrown or update the Checkstyle rule to validate that the exception being thrown isn't checked.

alzimmermsft:
Fixes #7633 and #9190 

This PR adds two new APIs to `ClientLogger` to support logging checked exceptions as either a warning or error, `logCheckedExceptionAsWarning` and `logCheckedExceptionAsError`. These APIs fill a gap in functionality where it is expected that exceptions are logged but there was no APIs to support logging `Throwable` based exceptions, there was only support for `RuntimeException` subclasses.

This PR has a few additional minor changes that were added as well.

- Fixed an issue where `logExceptionAsError` was using `VERBOSE` as the log level instead of the expected `ERROR` level.
- Removed `slf4j-simple` test dependencies from all Azure Core modules. This fixes part of a long term goal to completely remove this from all modules.
- Adds `5` as a supported environment variable for `AZURE_LOG_LEVEL` to configure the `NOT_SET` log level.

### DIFF
--- a/sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
@@ -18,6 +18,7 @@ import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
 import java.util.stream.Stream;
 
+import static com.azure.core.util.Configuration.PROPERTY_AZURE_LOG_LEVEL;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
```diff
@@ -57,233 +58,167 @@ public void revertLoggingConfiguration() throws Exception {
...
-        RuntimeException runtimeException = createRuntimeException(exceptionMessage);
+        IOException ioException = createIOException(exceptionMessage);
 
-        String originalLogLevel = setupLogLevel(1);
-        new ClientLogger(ClientLoggerTests.class).logExceptionAsWarning(runtimeException);
-        setPropertyToOriginalOrClear(Configuration.PROPERTY_AZURE_LOG_LEVEL, originalLogLevel);
+        String originalLogLevel = setupLogLevel(logLevelToConfigure.getLogLevel());
+        try {
+            throw new ClientLogger(ClientLoggerTests.class).logCheckedExceptionAsError(ioException);
+        } catch (Throwable throwable) {
```
srnagar: Can this be changed to catch IOException instead?

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
```diff
@@ -134,37 +135,76 @@ public void error(String format, Object... args) {
...
+     * <p>
+     * This API covers the cases where a checked exception type needs to be thrown and logged. If a {@link
+     * RuntimeException} is being logged use {@link #logExceptionAsError(RuntimeException)} instead.
+     *
+     * @param checkedException Throwable to be logged and returned.
+     * @param <T> Type of the Throwable being logged.
+     * @return The passed {@link Throwable}.
+     * @throws NullPointerException If {@code checkedException} is {@code null}.
+     */
+    public <T extends Throwable> T logCheckedExceptionAsError(T checkedException) {
```
srnagar: Since this applies to all throwables, we can name this `logThrowableAsError`

sdk/core/azure-core/src/test/java/com/azure/core/util/logging/ClientLoggerTests.java
alzimmermsft: I was looking to catch the generic type to ensure that `instanceof` it would be what is expected.

sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java
alzimmermsft: Renamed

PR #9202, opened 2020-03-18
Refactoring perf tests

srnagar:


sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadBlobTest.java
```diff
@@ -21,6 +27,6 @@ public void run() {
 
     @Override
     public Mono<Void> runAsync() {
-        return blobAsyncClient.upload(RandomFlux.create(options.getSize()), null, true).then();
+        return blobAsyncClient.upload(randomByteBufferFlux, null, true).then();
```
mikeharder: Should be better for perf since we don't create new flux every iteration.

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadFromFileTest.java
```diff
@@ -3,33 +3,31 @@
...
-
-import com.azure.perf.test.core.RandomStream;
-import com.azure.perf.test.core.SizeOptions;
-import com.azure.storage.blob.perf.core.BlobTestBase;
 import reactor.core.publisher.Mono;
 
-public class UploadFromFileTest extends BlobTestBase<SizeOptions> {
+public class UploadFromFileTest extends BlobTestBase<PerfStressOptions> {
+
+    private final Path tempFile;
```
mikeharder: Why move from static to instance?  All instances of this test should be able to upload from the same file on disk, right?

mikeharder: Either way would work, but my preference is to share the same file unless each test needs its own file.  It may also be slightly faster since the OS disk cache only needs to cache one file, not N files.

mikeharder: Moving this from static to instance actually breaks the test, since the file is only created and deleted in `globalSetup`, not per-instance `setup`.

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadBlockBlobTest.java
```diff
@@ -3,24 +3,33 @@
...
 import com.azure.storage.blob.perf.core.BlobTestBase;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-public class UploadBlockBlobTest extends BlobTestBase<SizeOptions> {
-    public UploadBlockBlobTest(SizeOptions options) {
+public class UploadBlockBlobTest extends BlobTestBase<PerfStressOptions> {
+    private final InputStream randomInputStream;
```
mikeharder: Could a single `InputStream` or `Flux<ByteBuffer>` be used for all test instances?  If so these should be static.

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadBlobTest.java
mikeharder: Though is it safe to reuse the same flux for each iteration?  Or is the flux "used up" after the first upload so we would need to create a new one every time?

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/DownloadBlobTest.java
```diff
@@ -49,7 +48,9 @@ public void write(int b) throws IOException {
     public Mono<Void> runAsync() {
         return blobAsyncClient.download()
             .map(b -> {
-                b.get(new byte[b.remaining()]);
+                for (int i = 0; i < b.remaining(); i++) {
```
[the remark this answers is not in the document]
g2vinay: yep, memory pressure is reduced.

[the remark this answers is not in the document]
srnagar: Yes, this reduces memory pressure as we don't really need to copy the data, just reading is sufficient.

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadBlobTest.java
srnagar: Yeah, tested it by using the same flux to read multiple times. Same data was returned each time.

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadFromFileTest.java
srnagar: Yes, I was about to update globalSetup to per instance setup but yeah, it makes sense to have a single file. I will switch this back to static.

common/perf-test-core/src/main/java/com/azure/perf/test/core/PerfStressOptions.java
```diff
@@ -38,6 +38,28 @@
     @Parameter(names = { "--sync" }, description = "Runs sync version of test")
     private boolean sync = false;
 
+    @Parameter(names = { "-s", "--size" }, description = "Size of payload (in bytes)")
```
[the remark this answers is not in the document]
srnagar: I was looking to keep all the options in one class so that users can find all available options in one place. Options that don't apply to a test can have appropriate error messages or just simply not use them.

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadBlockBlobTest.java
srnagar: Agree. Will make this static,

common/perf-test-core/src/main/java/com/azure/perf/test/core/PerfStressOptions.java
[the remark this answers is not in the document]
srnagar: Also, if a test needs both size and count options, then a third class has to be created to include both options as the `PerfStressTest` ctor takes a single `PerfTestOptions` instance.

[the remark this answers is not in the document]
mikeharder: Discussed offline, I slightly prefer separate classes to avoid having options which don't apply to all tests, but extra options aren't a big deal.

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadBlobTest.java
mikeharder: Would it also work with a static flux, if multiple instances were reading in parallel?  Or does it need to be an instance flux, so it can be reused across iterations, but each instance needs its own copy?  Shouldn't matter much for perf either way, but I like to scope variables correctly to increase understanding of the minimal test requirements.

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadBlockBlobTest.java
srnagar: Just realized that to create this input stream and flux, we need `PerfStressOptions` which is per-instance. So, can't make this static.

sdk/storage/perf-test/src/main/java/com/azure/storage/blob/perf/UploadBlobTest.java
srnagar: It could have been static if the since of the buffer was fixed. Since the size is determined by the options provided in the constructor, I have made this an instance flux. However, as we discussed, we could make this static and initialize in `globalSetup()` as the options are not going to change for a run and all instances share the same option. I didn't want to make that assumption here as it didn't matter to the performance of the test itself.

PR #9203, opened 2020-03-18
Use custom serialization and deserialization to flatten regexFlags.

sima-zhu:
Working on adding tests.

### DIFF
--- a/sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/implementation/util/CustomPatternTokenizerSerializer.java
+++ b/sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/implementation/util/CustomPatternTokenizerSerializer.java
@@ -49,7 +49,7 @@ public class CustomPatternTokenizerSerializer extends JsonSerializer<PatternToke
     @Override
     public void serialize(final PatternTokenizer tokenizer, final JsonGenerator jsonGenerator,
         final SerializerProvider serializerProvider) throws IOException {
-        if(tokenizer.getName() != null) {
+        if (tokenizer.getName() != null) {
             jsonGenerator.writeStringField("name", tokenizer.getName());
         }
         if (tokenizer.getPattern() != null) {


sdk/search/azure-search/src/main/java/com/azure/search/documents/util/RegexFlagsUtils
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.search.documents.util;
```
alzimmermsft: This isn't marked as a Java file

sima-zhu: I am debugging this for a while

sdk/search/azure-search/src/main/java/com/azure/search/documents/util/RegexFlagsUtils
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * Combines a list of {@link RegexFlags} into one separated by "|".
+     * E.g. {@code "CANON_EQ|CASE_INSENSITIVE|COMMENTS"}
+     *
+     * @param regexFlags A list a {@link RegexFlags} to combine.
+     * @return New combined {@link RegexFlags} separated by vertical bar.
+     */
+    public static RegexFlags combineRegexFlags(RegexFlags... regexFlags) {
+        return RegexFlags.fromString(String.join("|", Arrays.deepToString(regexFlags)));
```
alzimmermsft: This won't work as expected for a few reasons.

- If `regexFlags` is `null` this will return `"null"`.
- If `regexFlags` has values it will result in `"[val1.toString(), val2.toString(), ...]"`.

This would be a better option.

```java
if (CoreUtils.isNullOrEmpty(regexFlags)) {
    return null;
}

return Arrays.stream(regexFlags)
    .map(RegexFlags::toString)
    .collect(Collectors.joining("|"));
```

sdk/search/azure-search/src/main/java/com/azure/search/documents/util/RegexFlagsUtils
```diff
@@ -0,0 +1,24 @@
...
+public final class RegexFlagsUtils {
+
+    /**
+     * Combines a list of {@link RegexFlags} into one separated by "|".
+     * E.g. {@code "CANON_EQ|CASE_INSENSITIVE|COMMENTS"}
+     *
+     * @param regexFlags A list a {@link RegexFlags} to combine.
+     * @return New combined {@link RegexFlags} separated by vertical bar.
+     */
+    public static RegexFlags combineRegexFlags(RegexFlags... regexFlags) {
```
alzimmermsft: I'm a bit worried that `RegexFlags` is an `ExpandableStringEnum` as this may grow unbounded with the number of permutations available and will remain in memory as it is backed by a static dictionary. 

@brjohnstmsft does the ordering of the regex flags matter? If not we should sort before joining as that reduces the number of possible combinations that will be stored in static memory.

alzimmermsft: Would the service return a combined regex flag that a user may want to split up? If so should we offer a `splitRegexFlags` class that has this API shape:

```java
public static RegexFlags[] splitRegexFlags(RegexFlags combinedRegexFlag)
```

sdk/search/azure-search/src/test/java/com/azure/search/documents/CustomAnalyzerSyncTests.java
```diff
@@ -313,6 +314,20 @@ public void canUseAllAnalysisComponentNames() {
         assertCustomAnalysisComponentsEqual(index, createdIndex);
     }
 
+    @Test
+    public void canUseCombinedRegexFlags() {
```
alzimmermsft: Should add some tests that target `RegexFlagsUtils` directly.

sdk/search/azure-search/src/main/java/com/azure/search/documents/util/RegexFlagsUtils
brjohnstmsft: The ordering shouldn't matter, but maybe we should consider modeling the `flags` properties of `PatternAnalyzer` and `PatternTokenizer` as collections instead of introducing this combining primitive? I believe @xirzec is taking this approach in TypeScript.

brjohnstmsft: Nobody has asked for this, but the scenario makes sense. It seems like one more reason to just model the appropriate properties as lists in the first place.

sima-zhu: I have considered to do this. These two classes are final auto-gen classes, we have to do some transformation work if adding this function over there.

sima-zhu: Sync with @xirzec , I will move this function to the best place after discussing with Jeff.

sima-zhu: This is nice to have feature and addon. I would prefer to have it after GA

sima-zhu:
Postpone the PR a bit after I synced with other languages.

sdk/search/azure-search/src/main/java/com/azure/search/documents/util/RegexFlagsUtils
brjohnstmsft: @sima-zhu If you model flags as string collections, you'll need to implement the splitting logic for deserialization anyway. This should be done sooner rather than later.

sdk/search/azure-search/src/main/java/com/azure/search/documents/models/PatternAnalyzer.java
```diff
@@ -112,8 +114,9 @@ public RegexFlags getFlags() {
      * @param flags the flags value to set.
      * @return the PatternAnalyzer object itself.
      */
-    public PatternAnalyzer setFlags(RegexFlags flags) {
-        this.flags = flags;
+    public PatternAnalyzer setFlags(Collection<RegexFlags> flags) {
+        String newRegexFlags = flags.stream().map(RegexFlags::toString).collect(Collectors.joining("|"));
```
alzimmermsft: This won't be `null` safe which is different behavior than before.

sdk/search/azure-search/src/main/java/com/azure/search/documents/models/PatternAnalyzer.java
```diff
@@ -112,8 +114,9 @@ public RegexFlags getFlags() {
      * @param flags the flags value to set.
      * @return the PatternAnalyzer object itself.
      */
-    public PatternAnalyzer setFlags(RegexFlags flags) {
-        this.flags = flags;
+    public PatternAnalyzer setFlags(Collection<RegexFlags> flags) {
```
alzimmermsft: This API may be better as `setFlags(RegexFlags... flags)`.

sdk/search/azure-search/src/main/java/com/azure/search/documents/models/PatternAnalyzer.java
sima-zhu: Yes, it will throw NPE. Is it better to make it silent if pass null?

sdk/search/azure-search/src/main/java/com/azure/search/documents/models/PatternAnalyzer.java
sima-zhu: The reason I choose this because we do not need to convert if passing RegexFlags.values()

alzimmermsft: Would this be a common scenario though? Also, I think this would be likely to fail as `values()` would contain all the unioned `RegexFlags` as well.

alzimmermsft: @brjohnstmsft what does the service do if the same flag is passed multiple times.

sdk/search/azure-search/src/main/java/com/azure/search/documents/models/PatternAnalyzer.java
alzimmermsft: `null` was an allowable configuration before, this changes it so that once this is set you can never clear it out.

sdk/search/azure-search/src/main/java/module-info.java
```diff
@@ -5,12 +5,14 @@
     requires transitive com.azure.core;
     requires com.fasterxml.jackson.datatype.jsr310;
     requires jakarta.activation;
+    requires com.fasterxml.jackson.dataformat.xml;
```
alzimmermsft: Why are we requiring `dataformat.xml`?

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerSerializer.java
```diff
@@ -0,0 +1,79 @@
...
+     * {@inheritDoc}
+     *
+     * @param analyzer The {@link PatternAnalyzer} needs to serialize
+     * @param jsonGenerator Generator used to output resulting Json content
+     * @param serializerProvider Provider that can be used to get serializers for
+     *   serializing Objects value contains, if any.
+     * @throws IOException If IO operation error occurs.
+     */
+    @Override
+    public void serialize(final PatternAnalyzer analyzer, final JsonGenerator jsonGenerator,
```
alzimmermsft: What happens if this is called directly instead of being called through `serializeWithType`? Is it missing the start and end object pieces?

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerSerializer.java
```diff
@@ -0,0 +1,79 @@
...
+     * @throws IOException If IO operation error occurs.
+     */
+    @Override
+    public void serializeWithType(PatternAnalyzer analyzer, JsonGenerator gen, SerializerProvider serializers,
+        TypeSerializer typeSer) throws IOException {
+        gen.writeStartObject();
+        gen.writeStringField(typeSer.getPropertyName(), typeSer.getTypeIdResolver().idFromValue(analyzer));
+        serialize(analyzer, gen, serializers);
+        gen.writeEndObject();
+    }
```
alzimmermsft: Should any logic in here just be added into `serialize` and this acts as a proxy to `serialize`.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public PatternAnalyzer deserialize(final JsonParser p, final DeserializationContext ctxt) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectNode root = mapper.readTree(p);
+        JsonNode flattenFlags = root.get("flags");
+        List<RegexFlags> regexFlags = Arrays.stream(flattenFlags.asText().split(DELIMITER))
```
alzimmermsft: What happens if `flattenFlags` is null?

alzimmermsft: Does this need to live outside of the `fields.hasNext()` loop?

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
```diff
@@ -0,0 +1,62 @@
...
+                analyzer.setName(field.getValue().asText());
+            } else if ("pattern".equals(field.getKey())) {
+                analyzer.setPattern(field.getValue().asText());
+            } else if ("flags".equals(field.getKey())) {
+                analyzer.setFlags(regexFlags);
+            } else if ("lowercase".equals(field.getKey())){
+                analyzer.setLowerCaseTerms(field.getValue().asBoolean());
+            } else if ("stopwords".equals(field.getKey())) {
+                List<String> stopWords = new ArrayList<>();
+                field.getValue().forEach(
```
alzimmermsft: What if `stopwords` is `null`?

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
```diff
@@ -0,0 +1,62 @@
...
+                    jsonNode -> stopWords.add(jsonNode.asText())
+                );
+                analyzer.setStopwords(stopWords);
```
alzimmermsft: Could this be done in a stream?

```java
analyzer.setStopwords(field.getValue().stream()
    .map(node -> node.asText()).collect(Collectors.toList()));
```

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Custom deserializer for {@link PatternAnalyzer}, which convert flatten string to the list of {@link RegexFlags}.
+ */
+public class CustomPatternAnalyzerDeserializer extends JsonDeserializer<PatternAnalyzer> {
+    private static final String DELIMITER = "\\|";
```
alzimmermsft: Does this need to be escaped?

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerSerializer.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param serializerProvider Provider that can be used to get serializers for
+     *   serializing Objects value contains, if any.
+     * @throws IOException If IO operation error occurs.
+     */
+    @Override
+    public void serialize(final PatternAnalyzer analyzer, final JsonGenerator jsonGenerator,
+        final SerializerProvider serializerProvider) throws IOException {
+        if (analyzer == null) {
+            return;
+        }
```
alzimmermsft: What is the output when analyzer is null?

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternTokenizerDeserializer.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public PatternTokenizer deserialize(final JsonParser p, final DeserializationContext ctxt) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectNode root = mapper.readTree(p);
+        JsonNode flattenFlags = root.get("flags");
+        List<RegexFlags> regexFlags = Arrays.stream(flattenFlags.asText().split(DELIMITER))
+            .map(RegexFlags::fromString).collect(Collectors.toList());
```
alzimmermsft: Same questions as the other deserializer about `flags` being `null` and this needing to be outside of the JSON node loop.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternTokenizerDeserializer.java
```diff
@@ -0,0 +1,54 @@
...
+        while (fields.hasNext()) {
+            Map.Entry<String, JsonNode> field = fields.next();
+            if ("name".equals(field.getKey())) {
+                tokenizer.setName(field.getValue().asText());
+            } else if ("pattern".equals(field.getKey())) {
+                tokenizer.setPattern(field.getValue().asText());
+            } else if ("flags".equals(field.getKey())) {
+                tokenizer.setFlags(regexFlags);
+            } else if ("group".equals(field.getKey())){
+                tokenizer.setGroup(field.getValue().asInt());
```
alzimmermsft: How does `asInt()` handle if `getValue()` is `null`? `group` is a nullable `Integer` type so it could come back as `null` from my understanding.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternTokenizerSerializer.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param typeSer Type serializer to use for including type information
+     * @throws IOException if IO operation error occurs.
+     */
+    @Override
+    public void serializeWithType(PatternTokenizer tokenizer, JsonGenerator gen, SerializerProvider serializers,
+        TypeSerializer typeSer) throws IOException {
+        gen.writeStartObject();
+        gen.writeStringField(typeSer.getPropertyName(), typeSer.getTypeIdResolver().idFromValue(tokenizer));
+        serialize(tokenizer, gen, serializers);
+        gen.writeEndObject();
```
alzimmermsft: Same questions.

sdk/search/azure-search/src/main/java/module-info.java
sima-zhu: It is useless. Removed.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerSerializer.java
sima-zhu: ["@odata.type": "#Microsoft.Azure.Search.PatternAnalyzer" ] is added before serialize the rest. So the method is called to add the polymorphism type.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerSerializer.java
sima-zhu: Not really, the typeInfoId and typeInfoName is inside of TypeSerializer.
`["@odata.type": "#Microsoft.Azure.Search.PatternAnalyzer" ]`

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
sima-zhu: That's a miss. Will double check null cases for all.

sima-zhu: No. I forgot to move it into method

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
sima-zhu: Then the check of  "stopwords".equals(field.getKey()) will return false.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
sima-zhu: That's how `String.split` method works.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternTokenizerDeserializer.java
sima-zhu: I don't think the case exists.
If the field value is null, then there is no need to generate field key. If the case exists for required field, then we need to have null or empty checking for the field.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerSerializer.java
sima-zhu: > My main concern is around the serializers and deserializers and some corner cases they may have, could you add tests which specifically target them? Make sure to add guards even in scenarios where the service may not have those returns now since it is always possible to change in the future, especially around default JSON values.

Totally agreed for adding more tests to cover more corner scenarios. Will keep working on this while reviewing the body.

I have several test cases change to cover the major workflow.

sima-zhu:
@alzimmermsft Totally agreed for adding more tests to cover more corner scenarios. Will keep working on this while reviewing the body.

I have several test cases change to cover the major workflow.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
sima-zhu: Double checked code. `field.getVaule` returns iterator.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerSerializer.java
sima-zhu: I cannot have the case. `Index.setAnalyzer` takes list<Analyzer>

There is no way to have a null PatternAnalyzer pass in. 

I plan to remove it.

sdk/search/azure-search/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
alzimmermsft: So if the JSON contains the following we'll never have a `stopwords` field?

```json
{ "stopwords": null }
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/implementation/util/CustomPatternAnalyzerDeserializer.java
```diff
@@ -0,0 +1,61 @@
...
+
+        Iterator<Map.Entry<String, JsonNode>> fields = root.fields();
+        PatternAnalyzer analyzer = new PatternAnalyzer();
+        while (fields.hasNext()) {
+            Map.Entry<String, JsonNode> field = fields.next();
+            if ("name".equals(field.getKey())) {
+                analyzer.setName(field.getValue().asText());
+            } else if ("pattern".equals(field.getKey())) {
+                analyzer.setPattern(field.getValue().asText());
+            } else if ("flags".equals(field.getKey()) && !"null".equals(field.getValue().asText())) {
```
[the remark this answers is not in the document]
sima-zhu: "null" is a string type in json

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/implementation/util/CustomPatternTokenizerDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+        ObjectNode root = mapper.readTree(p);
+        Iterator<Map.Entry<String, JsonNode>> fields = root.fields();
+        PatternTokenizer tokenizer = new PatternTokenizer();
+        while (fields.hasNext()) {
+            Map.Entry<String, JsonNode> field = fields.next();
+            if ("name".equals(field.getKey())) {
+                tokenizer.setName(field.getValue().asText());
+            } else if ("pattern".equals(field.getKey())) {
+                tokenizer.setPattern(field.getValue().asText());
+            } else if ("flags".equals(field.getKey()) && !"null".equals(field.getValue().asText())) {
```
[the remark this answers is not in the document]
sima-zhu: It is a string type in json:

`{"flags": "null"} `

what we parse out here is getValue is "null". That's why I check it like this.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/implementation/util/CustomPatternTokenizerDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+                List<RegexFlags> regexFlags = Arrays.stream(field.getValue().asText().split(DELIMITER))
+                    .map(RegexFlags::fromString).collect(Collectors.toList());
+                tokenizer.setFlags(regexFlags);
+            } else if ("group".equals(field.getKey())){
+                tokenizer.setGroup(field.getValue().asInt());
```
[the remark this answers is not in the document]
sima-zhu: Based on the test of null group, service does not return group back if it is null.

PR #9213, opened 2020-03-19
Cosmos api review iteration

kushagraThapar:
DONE => 
* Updated ChangeFeedProcessorBuilder and CosmosClientBuilder to use builder pattern to set values 
* Moved CosmosPagedIterable and CosmosPagedFlux to new util package
* Moved DatabaseAccount to implementation package
* Removed public constructors from HashIndex and RangeIndex
* Updated CosmosResourceType to capital enum values
* Reduced scope of some public APIs to package-private or protected. 
* Renamed public APIs to correct naming conventions (specially for boolean values)

TODO => 
* JsonSerializable changes (a lot of APIs are still public, which should not be public)
* Check if IndexKind enum is necessary or can be removed
* Why SqlParameterList implements List interface

### DIFF
--- a/eng/pipelines/aggregate-reports.yml
+++ b/eng/pipelines/aggregate-reports.yml
@@ -30,7 +30,7 @@ jobs:
     - task: Maven@3
       displayName: 'Install Module'
       inputs:
-        mavenPomFile: sdk/core/pom.service.xml
+        mavenPomFile: pom.client.xml
         options: '--no-transfer-progress -DskipTests -Dgpg.skip -Dcheckstyle.skip=true -Dspotbugs.skip=true -Dmaven.javadoc.skip=true -Drevapi.skip=true'
         mavenOptions: '$(MemoryOptions)'
         javaHomeOption: 'JDKVersion'
@@ -54,7 +54,7 @@ jobs:
     - task: Maven@3
       displayName: 'Generate aggregate code coverage and dependency reports'
       inputs:
-        mavenPomFile: 'pom.client.xml'
+        mavenPomFile: pom.client.xml
         options: '$(DefaultOptions) -Dinclude-non-shipping-modules -Ddependency-checker -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true'
         mavenOptions: '-Xmx3072m $(LoggingOptions)'
         javaHomeOption: 'JDKVersion'
--- a/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-tests.yml
@@ -33,6 +33,7 @@ parameters:
       OSVmImage: 'windows-2019'
       JavaVersion: '1.11'
       DisplayName: 'Windows - Java 11'
+  MavenTestPomFile: pom.xml
   PreRunSteps: []
   PostRunSteps: []
   TestName: LiveTest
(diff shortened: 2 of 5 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -255,7 +255,7 @@ public CosmosClientBuilder setConsistencyLevel(ConsistencyLevel desiredConsisten
      *
      * @return the connection policy
      */
-    public ConnectionPolicy getConnectionPolicy() {
```
mbhaskar: If these methods are not public on builder, and we are not providing getters on the client, How can enduser get the config of the client after creating it?

kushagraThapar: They don't need to.. We discussed it, and end user should not need to get the config of the client they built. 
Just in case if they need it, we will make them available on the client itself, and not on the builder.

mbhaskar: I strongly feel it will be good to have these getters atleast on the client. I would have missed the discussion, but even .NET sdk has these getters public https://github.com/Azure/azure-cosmos-dotnet-v3/blob/f2e9d97d176fecdbe186c13575db06f7539b75e3/Microsoft.Azure.Cosmos/src/CosmosClientOptions.cs#L161
This can be done separately out of this PR if required but should be tracked.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
```diff
@@ -81,11 +81,11 @@ public void readFeedDocumentsStartFromBeginning() {
...
-            .setHandleChanges(itemConsumer)
-            .setFeedContainer(createdFeedCollection)
-            .setLeaseContainer(createdLeaseCollection)
-            .setOptions(new ChangeFeedProcessorOptions()
+            .hostName(hostName)
+            .handleChanges(itemConsumer)
+            .feedContainer(createdFeedCollection)
+            .leaseContainer(createdLeaseCollection)
+            .options(new ChangeFeedProcessorOptions()
                 .setLeaseRenewInterval(Duration.ofSeconds(20))
```
milismsft: it looks like the ChangeFeedProcessorOptions still has setters. Should we rename those as well to be consistent with the other changes? BTW my preference is to prefix these using "with" but that's probably for a later SDK version :-)

moderakh: I feel we are going back and forth with removing `set` as prefix and then adding it back again and then removing it again. !!

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
kushagraThapar: Yeah, I thought so too, but I couldn't present a solid argument, as in why end user needs them, since end user is already setting these values on the builder. If they need to access it again, they can get it from the place they initially set it in the builder. 
Let me know if you have a strong use case, may be I am missing something.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
milismsft: The resulting ChangeFeedProcessor object is immutable, it cannot be updated after it was built. ChangeFeedProcessorOptions might look like a POJO but in this case it acts almost as a builder; the getters are only used by the internal implementation.

Let's stick with what we have, and get this SDK GA'ed as soon as we can.

And for those curious, some background story about my comment. In the ARM Java Fluent SDK we used with* for methods returning "this", and could be chained in a Fluent/builder like pattern. We also accept setters and getters, the difference was that the setters will have to return "void" and never "this", while for the getters the explicit contract was that a network/long operation call must be made in order to return the value.

kushagraThapar: Sounds good Milis. Thanks for the explanation, appreciate it.

PR #9221, opened 2020-03-19
Fixes for live tests after POM restructure

mitchdenny:
Tweaks to live test execution to build and install dependencies during a live test. Works around an issue created where search doesn't contain references to core modules anymore in ```sdk/search/pom.xml``` (formerly ```pom.service.xml```).

This is experimental right now to quickly see if the fix will work and then I'll scale it out to all the modules that have live tests.


mitchdenny:
Merging this in. The live tests are spinning up correctly, although there are still some failures, but they aren't down to the changes we've got here. Merging this now will help our live test runs over night provide meaningful results so folks can work on test flakiness.

JimSuplizio:
@mitchdenny, we now have duplicated ServiceDirectory and Artifacts entries between ci.yml and tests.yml for some areas, we may want to consider something common.

PR #9222, opened 2020-03-19
Update KV clients to set tracing namespace info

samvaity:
[FEATURE REQ] Distributed tracing: set az.namespace attribute, [here](https://github.com/Azure/azure-sdk-for-java/issues/7049)

Client libraries must provide their respective Microsoft Resource Provider Namespace to the tracing plugin package for setting the namespace attribute on the outgoing spans.

Approach
Add a withContext method overload into FluxUtil that takes in a map to specify tracing attributes.

### DIFF
--- a/sdk/keyvault/azure-security-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/CryptographyAsyncClient.java
+++ b/sdk/keyvault/azure-security-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/CryptographyAsyncClient.java
@@ -30,11 +30,14 @@ import reactor.core.publisher.Mono;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 import static com.azure.core.util.FluxUtil.monoError;
 import static com.azure.core.util.FluxUtil.withContext;
+import static com.azure.core.util.tracing.Tracer.AZ_TRACING_NAMESPACE_KEY;
 import static com.azure.security.keyvault.keys.models.KeyType.EC;
 import static com.azure.security.keyvault.keys.models.KeyType.EC_HSM;
 import static com.azure.security.keyvault.keys.models.KeyType.RSA;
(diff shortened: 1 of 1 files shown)


sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/SecretAsyncClient.java
```diff
@@ -59,6 +62,10 @@
     static final int DEFAULT_MAX_PAGE_RESULTS = 25;
     static final String CONTENT_TYPE_HEADER_VALUE = "application/json";
     static final String KEY_VAULT_SCOPE = "https://vault.azure.net/.default";
+    // Please see <a href=https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-services-resource-providers>here</a>
+    // for more information on Azure resource provider namespaces.
+    public static final Map<String, String> KEYVAULT_TRACING_PROPERTIES =
```
g2vinay: does this need to be public ?

sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
```diff
@@ -98,8 +99,34 @@ private static void accept(ByteArrayOutputStream byteOutputStream, ByteBuffer by
...
+     * </p>
+     *
+     * @param serviceCall serviceCall The lambda function that makes the service call into which azure context
+     * will be passed
+     * @param contextAttributes The map of attributes sent by the calling method to be set on {@link Context}.
+     * @param  <T> The type of response returned from the service call
+     * @return The response from service call
+     */
+    public static <T> Mono<T> withContext(Function<Context, Mono<T>> serviceCall,
+        Map<String, String> contextAttributes) {
```
g2vinay: Just curious, under what scenarios, we would expect this map to contain more than one value ?

samvaity: Yes, it is mostly for future-proofing. So that in future if there are more tracing attribute requirements its easily extendable that way.

sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/SecretAsyncClient.java
samvaity: right, doesn't have to be public.

sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Release History
 
 ## 1.0.0-beta.4 (Unreleased)
+- Keyvault: add az namespace info attribute to all outgoing spans.
```
g2vinay: Is this something good to mention in KV changelogs too ?

samvaity: This would rather be a value add feature to tracing than to KV so we can skip KV changelogs for this one.

PR #9239, opened 2020-03-19
Remove creation of storage resources.

conniey:
Removes creation of storage resources for Event Hubs live tests.

### DIFF
--- a/sdk/eventhubs/test-resources.json
+++ b/sdk/eventhubs/test-resources.json
@@ -8,10 +8,6 @@
                 "description": "Base name to append to Azure resources."
             }
         },
-        "storageEndpointSuffix": {
-            "type": "string",
-            "defaultValue": "core.windows.net"
-        },
         "eventHubName": {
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/test-resources.json
```diff
@@ -8,10 +8,6 @@
                 "description": "Base name to append to Azure resources."
             }
         },
-        "storageEndpointSuffix": {
```
alzimmermsft: Before merging this in and removing the Storage account creation for EventHubs could you double check that Checkpoint Store doesn't use it at all. I know it is built to use Storage Blobs to store checkpointing data, if there are no integration tests than this can be completely removed.

conniey: Yes. We do not have any integration tests for Checkpoint store. It seems wasteful to create something that is unused at the moment. If we add some in, we can do a `git fl` and resurface the code.

PR #9242, opened 2020-03-19
Fixes issue where an error is output from request(0)

conniey:
Fixes issue where, in debug, an error log is output when subscribing to the async clients.

```
java.lang.IllegalArgumentException: Spec. Rule 3.9 - Cannot request a non strictly positive number: 0
at reactor.core.Exceptions.nullOrNegativeRequestException(Exceptions.java:318)
at reactor.core.publisher.Operators.reportBadRequest(Operators.java:905)
at reactor.core.publisher.Operators.validate(Operators.java:1103)
at reactor.core.publisher.Operators$MultiSubscriptionSubscriber.request(Operators.java:1899)
at com.azure.core.amqp.implementation.AmqpChannelProcessor.onSubscribe(AmqpChannelProcessor.java:64)
at reactor.core.publisher.MonoRepeatPredicate.subscribeOrReturn(MonoRepeatPredicate.java:44)
at reactor.core.publisher.Flux.subscribe(Flux.java:8120)
at reactor.core.publisher.Flux.subscribeWith(Flux.java:8298)
```

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
@@ -554,7 +554,9 @@ public class EventHubClientBuilder {
                 if (request == 0) {
                     return;
                 } else if (request > 1) {
-                    logger.warning("Requested more than one connection. Only emitting one. Request: {}", request);
+                    sink.error(logger.logExceptionAsWarning(new IllegalArgumentException(
+                        "Requested more than one connection. Only emitting one. Request: " + request)));
+                    return;
                 }
 
                 final String connectionId = StringUtil.getRandomString("MF");


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
```diff
@@ -552,13 +548,24 @@ private EventHubConnectionProcessor buildConnectionProcessor(MessageSerializer m
...
-        final Flux<EventHubAmqpConnection> connectionFlux = Mono.fromCallable(() -> {
-            final String connectionId = StringUtil.getRandomString("MF");
+        final Flux<EventHubAmqpConnection> connectionFlux = Flux.create(sink -> {
+            sink.onRequest(request -> {
+
+                if (request == 0) {
+                    return;
+                } else if (request > 1) {
+                    logger.warning("Requested more than one connection. Only emitting one. Request: {}", request);
+                }
```
srnagar: Can this happen? We always request 1 connection.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionBasedLoadBalancer.java
```diff
@@ -112,6 +114,12 @@ void loadBalance() {
         final Mono<List<String>> partitionsMono = eventHubAsyncClient
             .getPartitionIds()
             .timeout(Duration.ofMinutes(1))
+            .onErrorResume(TimeoutException.class, error -> {
+                // In the subsequent step where it tries to balance the load, it'll propagate an error to the user.
+                // So it is okay that there is we return an empty Flux.
```
srnagar:
```suggestion
                // So it is okay to return an empty Flux.
```

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
conniey: No. should I throw an error instead if this ever happens? was trying to be defensive if we accidentally introduce a bug.

PR #9243, opened 2020-03-19
Make Test POM Settable

alzimmermsft:
Updates the archetype test YML to allow children YML to override the test target POM.

### DIFF
--- a/sdk/cosmos/tests.yml
+++ b/sdk/cosmos/tests.yml
@@ -58,6 +58,8 @@ jobs:
           PROTOCOLS: '["Https"]'
           DESIRED_CONSISTENCIES: '["Session"]'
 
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/stages/cosmos-sdk-client.yml
```diff
@@ -85,7 +85,7 @@ stages:
             DisplayName: 'Examples Integration Tests'
             PROTOCOLS: '["Https", "Tcp"]'
             DESIRED_CONSISTENCIES: '["Strong", "Session"]'
-
+        MavenTestPomFile: sdk/cosmos/pom.xml
```
mitchdenny: This should be the root POM.

PR #9245, opened 2020-03-19
Skip Version Verification and Update in BOM Releases

alzimmermsft:
Updates the BOM archetype YML to skip version verification and update during the release pipeline. This is done as YMLs are released on a different cadence and won't be using the same versions listed in the versioning scripts.

### DIFF
--- a/eng/pipelines/templates/stages/archetype-sdk-bom.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-bom.yml
@@ -2,9 +2,8 @@ parameters:
   Artifacts: []
   SDKType: bom
   ServiceDirectory: not-specified
-  Skip:
-    UpdatePackageVersion: true
-    VersionVerification: true
+  Skip.UpdatePackageVersion: true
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/stages/archetype-sdk-bom.yml
```diff
@@ -2,6 +2,9 @@ parameters:
   Artifacts: []
   SDKType: bom
   ServiceDirectory: not-specified
+  Skip:
+    UpdatePackageVersion: true
```
weshaggard: This should be:
```
Skip.UpdatePackageVersion: true
Skip.VersionVerification: true
```

alzimmermsft: I was wondering what the syntax on this was 😃

alzimmermsft: Updated to use `Skip.UpdatePackageVersion` and `Skip.VersionVerification`.

PR #9280, opened 2020-03-19
Address Cosmos issues after POM restructure

mitchdenny:
Because Cosmos is quite different in the way it builds the POM restructure didn't work out so well so there will be more changes required to make it behave correctly. This PR is being used to try out a few things to see if we can make it work.


mitchdenny:
@kushagraThapar and @alzimmermsft this should fix up the Cosmos issues after POM restructuring

kushagraThapar:
Thanks @mitchdenny  - I have started the live test CI to test at least one of them.

kushagraThapar:
@mitchdenny  - Looks good from Cosmos live test point.

PR #9281, opened 2020-03-19
Remove Storage Test Dependency on slf4j-simple

### LINKED ISSUE #7801: Remove slf4j-simple dependency
Now that we have DefaultLogger that will be used in the absence of an SLF4J binding, we should now be able to remove the dependency on slf4j-simple.

alzimmermsft:
Fixes #7801 

Removes the last test dependencies on `slf4j-simple`. This is being removed as Azure Core has a default logger which will be used during testing.

Related PRs
#9197 
#9198


alzimmermsft:
@gapra-msft or @rickle-msft mind giving this a quick once over? Should be the removal of an unnecessary test dependency only.

PR #9283, opened 2020-03-20
[Storage] Blob versioning.

kasobol-msft:
See https://review.docs.microsoft.com/en-us/azure/storage/blobs/versioning-overview?branch=pr-en-us-104390&tabs=powershell#blob-versioning-and-soft-delete .

### DIFF
--- a/sdk/storage/tests.yml
+++ b/sdk/storage/tests.yml
@@ -35,3 +35,5 @@ jobs:
         AZURE_TENANT_ID: $(aad-azure-sdk-test-tenant-id)
         AZURE_CLIENT_ID: $(aad-azure-sdk-test-client-id)
         AZURE_CLIENT_SECRET: $(aad-azure-sdk-test-client-secret)
+        VERSIONED_STORAGE_ACCOUNT_NAME: $(azure-sdk-test-versioned-storage-account-name)
+        VERSIONED_STORAGE_ACCOUNT_KEY: $(azure-sdk-test-versioned-storage-account-key)


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -151,7 +152,7 @@ public EncryptedBlobAsyncClient buildEncryptedBlobAsyncClient() {
 
         return new EncryptedBlobAsyncClient(getHttpPipeline(),
             String.format("%s/%s/%s", endpoint, containerName, blobName), serviceVersion, accountName, containerName,
-            blobName, snapshot, customerProvidedKey, keyWrapper, keyWrapAlgorithm);
+            blobName, snapshot, customerProvidedKey, keyWrapper, keyWrapAlgorithm, versionId);
```
rickle-msft: Is it legal for a url to have a snapshot and a versionId? I would guess not, and it may be worth validating that here.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
```diff
@@ -0,0 +1,364 @@
...
+        def blobItemV1 = blobClient.getAppendBlobClient().create();
+        def blobItemV2 = blobClient.getAppendBlobClient().create(true);
+
+        then:
+        blobItemV1.getVersionId() != null
+        blobItemV2.getVersionId() != null
+        !StringUtils.equals(blobItemV1.getVersionId(), blobItemV2.getVersionId())
+    }
+
+    def "Retrieve Block Blob by Version"() {
```
rickle-msft: I think retrieve, properties, and delete are independent of blob type and should share the same code path regardless of blob type, so it may be better to only test that once.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
```diff
@@ -0,0 +1,364 @@
...
+
+        when:
+        blobClient.getVersionClient(blobItemV1.getVersionId()).getAppendBlobClient().delete()
+
+        then:
+        !blobClient.getVersionClient(blobItemV1.getVersionId()).exists()
+        blobClient.getVersionClient(blobItemV2.getVersionId()).exists()
+    }
+
+    def "Retrieve Block Blob Properties by Version"() {
```
rickle-msft: nit: we usually use download instead of retrieve and get properties instead of retrieve properties for test naming

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
```diff
@@ -0,0 +1,364 @@
...
+    def "Copy From Url Blobs with Version"() {
+        given:
+        blobContainerClient.setAccessPolicy(PublicAccessType.CONTAINER, null)
+        def blobItemV1 = blobClient.getBlockBlobClient().upload(defaultInputStream.get(), defaultDataSize)
+        def sourceBlob = blobContainerClient.getBlobClient(generateBlobName())
+        sourceBlob.getBlockBlobClient().upload(defaultInputStream.get(), defaultDataSize)
+
+        when:
+        def response = blobClient.copyFromUrlWithResponse(sourceBlob.getBlobUrl(), null, null, null, null, null, Context.NONE)
+        def versionIdAfterCopy = response.getHeaders().getValue("x-ms-version-id")
```
rickle-msft: We shouldn't have to index into the raw headers to get this value. Maybe swagger needs to be updated to add this field to the deserialized headers?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
```diff
@@ -0,0 +1,364 @@
...
+
+    def "Copy From Url Blobs with Version"() {
+        given:
+        blobContainerClient.setAccessPolicy(PublicAccessType.CONTAINER, null)
+        def blobItemV1 = blobClient.getBlockBlobClient().upload(defaultInputStream.get(), defaultDataSize)
+        def sourceBlob = blobContainerClient.getBlobClient(generateBlobName())
+        sourceBlob.getBlockBlobClient().upload(defaultInputStream.get(), defaultDataSize)
+
+        when:
+        def response = blobClient.copyFromUrlWithResponse(sourceBlob.getBlobUrl(), null, null, null, null, null, Context.NONE)
```
rickle-msft: It may also be worth testing copying from a non-current version. Or at least testing getBlobUrl and ensuring that the version query parameter is present

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
rickle-msft: Looks like you covered that below :)

rickle-msft:
I think we should also have overloads on Container(Async)Client to get a child BlobClient with a version attached.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
kasobol-msft: Correct, that's covered.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
kasobol-msft: Fair point. I'll remove that.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
kasobol-msft: Ok.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
kasobol-msft: The swagger document seems to be correct. We do have access to deserialized headers at some point. However, the cases where I am retrieving raw header cover APIs that doesn't seem to have a good place to expose this value - they return Response<Void> or Response<String>, so deserialized headers get lost in the transformation from impl->blob client. I wonder if there's any other option to expose it.
Other alternative I can see is to shoot get properties request after copy but that's not safe in concurrent access scenarios (what if there's other operation that bumps version executed in the middle?).

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
kasobol-msft: Sounds good. I'll add test case for getBlobUrl.

kasobol-msft:
@rickle-msft Since the version and snapshot ids are mutually exclusive I'd need to add getBlobClient(name, version) but that is going to collide with existing getBlobClient(name, snapshot) as both version and snapshot are Strings. So that would mean getBlobClientByVersion or getBlobClientWithVersion, what do you think should be proper naming for that?
On the other hand having overload for snapshot sounds right as it's functionality you get out of the box. Versioning can be turned on and off, so I'm not sure if we should be exposing such "overload".
What do you think?

rickle-msft:
Since we already have versionId on the builders the same way we have snapshot, I think it makes sense to round out the support for it with a method on the containerClient. Otherwise we're sort of half-heartedly supporting it. Customers that don't have versioning enabled just won't know what it means to pass a version and will ignore the method I imagine. @gapra-msft what do you think?

The naming is a bit unfortunate. Probably getBlobClientWithVersion is the best option I can think of.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
rickle-msft: In that case this is fine. I didn't remember that some operations just return a raw Response header

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
```diff
@@ -355,4 +420,18 @@ public OffsetDateTime getAccessTierChangeTime() {
...
+
+    /**
+     * @return the version identifier the blob.
+     */
+    public String getVersionId() {
+        return versionId;
+    }
+
+    /**
+     * @return the a flag indicating whether version identifier points to current version of the blob.
```
gapra-msft: typo - "the a"

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
```diff
@@ -95,6 +97,67 @@ public BlobProperties(final OffsetDateTime creationTime, final OffsetDateTime la
...
+     * @param isAccessTierInferred Flag indicating if the access tier of the blob was inferred from properties of the
+     * blob.
+     * @param archiveStatus Archive status of the blob.
+     * @param encryptionKeySha256 SHA256 of the customer provided encryption key used to encrypt the blob on the server.
+     * @param accessTierChangeTime Datetime when the access tier of the blob last changed.
+     * @param metadata Metadata associated with the blob.
+     * @param committedBlockCount Number of blocks committed to an append blob, if the blob is a block or page blob
+     * pass {@code null}.
+     * @param versionId The version identifier of the blob.
+     * @param isCurrentVersion Flag indication if version identifier points to current version of the blob.
```
gapra-msft: "Flag indicating"

gapra-msft: Also not 100% sure about naming but I think it would read better to call the property "versionCurrent" and the getter would be like isVersionCurrent.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -121,12 +121,14 @@ class APISpec extends Specification {
...
      acquire so we do not run them in the nightly live run tests. */
     static def MANAGED_DISK_STORAGE = "MANAGED_DISK_STORAGE_"
+    static def VERSIONED_STORAGE = "VERSIONED_STORAGE_"
 
     protected static StorageSharedKeyCredential primaryCredential
     static StorageSharedKeyCredential alternateCredential
     static StorageSharedKeyCredential blobCredential
     static StorageSharedKeyCredential premiumCredential
     static StorageSharedKeyCredential managedDiskCredential
+    static StorageSharedKeyCredential versionedCredential
```
gapra-msft: did we create this new resource just cause versioning is still in stage? Would it eventually be available on any account?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
gapra-msft: We cant call getDeserializedHeaders here?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
rickle-msft: This is because versioning isn't on by default and can't be enabled from our data plane apis. It's similar in that way to a premium account, so we decided it'd be best to have a separate credential for it (though, as with all our credentials, it could be the same as the primary if the primary has this feature turned on).

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/VersioningTest.groovy
rickle-msft: copyFromUrl just returns Response<String>, and Response only has getHeaders, not getDeserializedHeaders

gapra-msft:
> Since we already have versionId on the builders the same way we have snapshot, I think it makes sense to round out the support for it with a method on the containerClient. Otherwise we're sort of half-heartedly supporting it. Customers that don't have versioning enabled just won't know what it means to pass a version and will ignore the method I imagine. @gapra-msft what do you think?
> 
> The naming is a bit unfortunate. Probably getBlobClientWithVersion is the best option I can think of.

I think we definitely should add this method on the container client just because some customers do use this pattern of creating a service client and getting their container/blob clients by calling these getBlob/getContainer methods. Rick's idea for a name sounds ok. For now we can just pick something that sounds the best to us since eventually this will go through Architecture board for approval anyway.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
kasobol-msft: @gapra-msft Regarding naming. I'd rather leave it as is. We have same naming in BlobItem.IsCurrentVersion that we expose elsewhere - naming comes from swagger.
image

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
kasobol-msft: fixed

seanmcc-msft:
Does Blob Version apply to the same APIs as Blob Snapshot?

kasobol-msft:
> Does Blob Version apply to the same APIs as Blob Snapshot?

@seanmcc-msft Mostly yes. Additionally majority of "write" APIs return version-id when feature is enabled. So it works kind of like auto-snapshot.

See https://review.docs.microsoft.com/en-us/azure/storage/blobs/versioning-overview?branch=pr-en-us-104390&tabs=powershell

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobSasImplUtil.java
```diff
@@ -97,7 +104,25 @@ public BlobSasImplUtil(BlobServiceSasSignatureValues sasValues, String container
...
+    /**
+     * Creates a new {@link BlobSasImplUtil} with the specified parameters
+     *
+     * @param sasValues {@link BlobServiceSasSignatureValues}
+     * @param containerName The container name
+     * @param blobName The blob name
+     * @param snapshotId The snapshot id
+     * @param versionId The version id
+     */
+    public BlobSasImplUtil(BlobServiceSasSignatureValues sasValues, String containerName, String blobName,
```
alzimmermsft: This code is in `implementation` so feel free to update the old API as it is non-breaking change.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/AppendBlobItem.java
```diff
@@ -55,6 +56,27 @@ public AppendBlobItem(final String eTag, final OffsetDateTime lastModified, fina
...
+     * @param lastModified Last modified time of the append blob.
+     * @param contentMd5 Content MD5 of the append blob.
+     * @param isServerEncrypted Flag indicating if the page blob is encrypted on the server.
+     * @param encryptionKeySha256 The encryption key used to encrypt the append blob.
+     * @param encryptionScope The encryption scope used to encrypt the append blob.
+     * @param blobAppendOffset The offset at which the block was committed to the append blob.
+     * @param blobCommittedBlockCount The number of committed blocks in the append blob.
+     * @param versionId The version identifier of the append blob.
+     */
+    public AppendBlobItem(final String eTag, final OffsetDateTime lastModified, final byte[] contentMd5,
```
alzimmermsft: Mind filing an issue to investigate a way to make constructors which are capable of handling additional parameters that the service could add in the future. There are a few other classes that need a new constructor every time a new service return is added.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -121,12 +121,14 @@ class APISpec extends Specification {
     /* Unignore any managed disk tests if a managed disk account is available to be tested. They are difficult to
      acquire so we do not run them in the nightly live run tests. */
     static def MANAGED_DISK_STORAGE = "MANAGED_DISK_STORAGE_"
+    static def VERSIONED_STORAGE = "VERSIONED_STORAGE_"
```
alzimmermsft: This is for another Storage Account resource correct? This will need an update to the ARM template which deploys resources before live test runs for these new tests to work in the nightly live test run.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -161,6 +161,22 @@ public BlobAsyncClient getBlobAsyncClient(String blobName, String snapshot) {
             getBlobContainerName(), blobName, snapshot, getCustomerProvidedKey(), encryptionScope);
     }
 
+    /**
+     * Creates a new BlobAsyncClient object by concatenating blobName to the end of ContainerAsyncClient's URL. The new
+     * BlobAsyncClient uses the same request policy pipeline as the ContainerAsyncClient. To change the pipeline, create
+     * the BlobAsyncClient and then call its WithPipeline method passing in the desired pipeline object. Or, call this
```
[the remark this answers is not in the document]
kasobol-msft: This is outdated statement referring to something that existed in previous version. Removed here and places I copied from.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerClient.java
```diff
@@ -98,6 +98,20 @@ public BlobClient getBlobClient(String blobName, String snapshot) {
         return new BlobClient(client.getBlobAsyncClient(blobName, snapshot));
     }
 
+    /**
+     * Initializes a new BlobClient object by concatenating blobName to the end of ContainerAsyncClient's URL. The new
+     * BlobClient uses the same request policy pipeline as the ContainerAsyncClient. To change the pipeline, create the
+     * BlobClient and then call its WithPipeline method passing in the desired pipeline object. Or, call this package's
```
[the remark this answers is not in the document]
kasobol-msft: This is outdated statement referring to something that existed in previous version. Removed here and places I copied from.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -399,6 +401,17 @@ public EncryptedBlobClientBuilder snapshot(String snapshot) {
         return this;
     }
 
+    /**
+     * Sets the version identifier of the blob.
```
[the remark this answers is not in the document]
kasobol-msft: Added here and other places where we pass versionId

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -161,6 +161,22 @@ public BlobAsyncClient getBlobAsyncClient(String blobName, String snapshot) {
...
+     * package's getBlobAsyncClient instead of calling this object's getBlobAsyncClient method.
```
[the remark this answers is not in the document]
kasobol-msft: This is outdated statement referring to something that existed in previous version. Removed.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerClient.java
```diff
@@ -98,6 +98,20 @@ public BlobClient getBlobClient(String blobName, String snapshot) {
...
+     * getBlobAsyncClient instead of calling this object's getBlobAsyncClient method.
+     *
+     * @param blobName A {@code String} representing the name of the blob.
+     * @param versionId the version identifier for the blob.
+     * @return A new {@link BlobClient} object which references the blob with the specified name in this container.
+     */
```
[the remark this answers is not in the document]
kasobol-msft: Fixed in this PR.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobSasImplUtil.java
kasobol-msft: Great! Done.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/AppendBlobItem.java
kasobol-msft: https://github.com/Azure/azure-sdk-for-java/issues/9530

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
kasobol-msft: @alzimmermsft that's correct. Versioning is feature that can be enabled/disabled. Few questions here:
- is there any documentation about how new resource should be added ? I guess it's not only ARM template.
- I'm going to check if versioning can be enabled via ARM at this point. If not then is there any alternative to inject account and it's credentials somehow into the pipeline?

alzimmermsft: If there isn't a way to add this as part of ARM we could resort to using a static testing resource given the limited number of tests that'll be using it. Given this is a new service feature I have a feeling there won't be ARM support for this.

kasobol-msft: Good news is that there is ARM support https://review.docs.microsoft.com/en-us/azure/storage/blobs/versioning-enable?branch=pr-en-us-104390&tabs=template .

Is the `azure-sdk-for-java\sdk\storage\test-resources.json` the only place that needs an update?

kasobol-msft: On the other hand it might take some time before support through ARM is rolled out. 
@alzimmermsft what does it take to define static testing resource? Who does that?

kasobol-msft: @JimSuplizio could you chime in on how to define static resource (hardcoded storage account in this case ) ?

JimSuplizio: @kasobol-msft unfortunately, I'm not the right person for this. @alzimmermsft and @hemanttanwar are the ones that setup the templates and test resources and @danieljurek may also have some input here.

PR #9284, opened 2020-03-20
Configuration Implementation Cleanup

alzimmermsft:
This PR cleans up the internal implementation of `Configuration`. The major changes are the following:

- Removed specialty checks to continuously load specific configurations from the environment.
- Removed loading of default configurations as an instance level method and moved this into the static initializer.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
@@ -130,15 +130,9 @@ public class Configuration implements Cloneable {
     };
 
     /*
-     * The base set of configurations that is loaded at the start of the application. This will be used to pre-populate
-     * all further Configurations created.
-     */
-    private static final Configuration BASE_CONFIGURATIONS = loadBaseConfiguration();
-
-    /*
      * Gets the global configuration shared by all client libraries.
      */
-    private static final Configuration GLOBAL_CONFIGURATION = BASE_CONFIGURATIONS.clone();
+    private static final Configuration GLOBAL_CONFIGURATION = new Configuration();
 
     /**
      * No-op {@link Configuration} object used to opt out of using global configurations when constructing client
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
```diff
@@ -120,22 +129,30 @@
...
+     * The base set of configurations that is loaded at the start of the application. This will be used to pre-populate
+     * all further Configurations created.
+     */
+    private static final Configuration BASE_CONFIGURATIONS = loadBaseConfiguration();
+
+    /*
      * Gets the global configuration shared by all client libraries.
      */
-    private static final Configuration GLOBAL_CONFIGURATION = new Configuration();
+    private static final Configuration GLOBAL_CONFIGURATION = BASE_CONFIGURATIONS.clone();
```
srnagar: Do we need both BASE_CONFIGURATIONS and GLOBAL_CONFIGURATION? We can directly use the global configurations as the base when creating a new instance. Both of these should be immutable anyway.

sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
```diff
@@ -120,22 +129,30 @@
         PROPERTY_AZURE_TRACING_DISABLED,
     };
 
-    /**
+    /*
+     * The base set of configurations that is loaded at the start of the application. This will be used to pre-populate
+     * all further Configurations created.
+     */
+    private static final Configuration BASE_CONFIGURATIONS = loadBaseConfiguration();
```
srnagar: nit: 
```suggestion
    private static final Configuration BASE_CONFIGURATION = loadBaseConfiguration();
```

sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
alzimmermsft: Removed

PR #9320, opened 2020-03-20
Kusto: generate package-2020-02

xseeseesee:


sdk/kusto/mgmt-v2020_02_15/src/main/java/com/microsoft/azure/management/kusto/v2020_02_15/implementation/ClustersInner.java
```diff
@@ -1602,11 +1610,12 @@ public DiagnoseVirtualNetworkResultInner call(ServiceResponse<DiagnoseVirtualNet
      * @throws IllegalArgumentException thrown if parameters fail the validation
      * @return the observable to the List&lt;ClusterInner&gt; object
      */
-    public Observable<List<ClusterInner>> listAsync() {
-        return listWithServiceResponseAsync().map(new Func1<ServiceResponse<List<ClusterInner>>, List<ClusterInner>>() {
-            @Override
-            public List<ClusterInner> call(ServiceResponse<List<ClusterInner>> response) {
-                return response.body();
+    public Observable<Page<ClusterInner>> listAsync() {
```
weidongxu-microsoft: Not sure whether I get it.

Inner usually just reflects the spec.
It seems the spec returns a List, which nothing wrong with it.

The code converting List to PagedList should be in Impl class?

xccc-msft: This difference comes from InnerSupportsListing<ClusterInner>

weidongxu-microsoft: I see. Then it is fine to modify.

PR #9324, opened 2020-03-20
Add EnvVar AZURE_AUTHORITY_HOST

Luyunmt:
Added EnvVar AZURE_AUTHORITY_HOST to azure.identity._constants.
TokenCredentialOptions has an AuthorityHost property.
This will be modified to support EnvVar override with AZURE_AUTHORITY_HOST.
Azure/azure-sdk-for-java/issues/5967


sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
```diff
@@ -83,7 +88,7 @@
     /**
      * Name of the Azure cloud to connect to.
      */
-    public static final String PROPERTY_AZURE_CLOUD = "AZURE_CLOUD";
```
srnagar: Removing/renaming a public env variable that is already released is considered a breaking change. Can this env var be retained and `PROPERTY_AZURE_AUTHORITY_HOST` be added as a new env var?

sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java
```diff
@@ -41,15 +55,22 @@
 
     @Override
     public Mono<AccessToken> getToken(TokenRequestContext request) {
+        String clientId = configuration.get(Configuration.PROPERTY_AZURE_CLIENT_ID);
+        String tenantId = configuration.get(Configuration.PROPERTY_AZURE_TENANT_ID);
+        String clientSecret = configuration.get(Configuration.PROPERTY_AZURE_CLIENT_SECRET);
+        String certPath = configuration.get(Configuration.PROPERTY_AZURE_CLIENT_CERTIFICATE_PATH);
+        String username = configuration.get(Configuration.PROPERTY_AZURE_USERNAME);
+        String password = configuration.get(Configuration.PROPERTY_AZURE_PASSWORD);
```
srnagar: Any reason these have to be initialized synchronously?

PR #9330, opened 2020-03-20
Inject Code Snippets Pre-Build

scbedd:
Getting this out there so we can put it where it actually makes sense.


eng/scripts/inject_codesnippets.py
```diff
@@ -0,0 +1,139 @@
...
+    if not os.path.isdir(file_name):
+        name = os.path.basename(file_name)
+
+        return name in exclusion_array
+
+
+if __name__ == "__main__":
+
+    # walk the codebase, find all java files
+    all_files = glob.glob("**/*.java", recursive=True)
```
alzimmermsft: Right now the only directories that can contain snippets are in `src/samples/java/*` if we want to tighten where we are looking for snippets.

eng/scripts/inject_codesnippets.py
```diff
@@ -0,0 +1,139 @@
+import os
+import glob
+import re
+import json
+from xml.sax.saxutils import escape, unescape
+
+# run this from the root of the repository
+root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))
+
+SNIPPET_BEGIN = r"\s*\/\/\s*BEGIN\:\s*(?P<id>[a-zA-Z0-9\.\#\-\_]*)\s*"
```
srnagar:
```suggestion
SNIPPET_BEGIN = r"\s*\/\/\s*BEGIN\:\s+(?P<id>[a-zA-Z0-9\.\#\-\_]*)\s*"
```

eng/scripts/inject_codesnippets.py
```diff
@@ -0,0 +1,139 @@
...
+SNIPPET_END = r"\s*\/\/\s*END\:\s*(?P<id>[a-zA-Z0-9\.\#\-\_]*)\s*"
```
srnagar:
```suggestion
SNIPPET_END = r"\s*\/\/\s*END\:\s+(?P<id>[a-zA-Z0-9\.\#\-\_]*)\s*"
```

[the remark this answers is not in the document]
srnagar: The space after `BEGIN:` and `END:` tags is mandatory.

eng/scripts/inject_codesnippets.py
scbedd: You're right. Why I choose this method is that I need all the java files _anyway_ to search for @codesnippet CALLS.

Given that I need to find all the java files anyway, I do it once, then pull from that already generated list for the `snippets` files.

I will adjust the targeting in the filter though. Thanks!

eng/scripts/inject_codesnippets.py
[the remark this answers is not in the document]
scbedd: @srnagar Do you mean I should make this `\s` and not `\s*`?

Right here: `END\:\s*`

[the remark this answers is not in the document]
srnagar: `\s*` is 0 or more spaces instead you should have 1 or more spaces `\s+`. See my suggested update to the regex above.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -102,6 +102,11 @@ jobs:
           BuildSHA: $(Build.SourceVersion)
           RepoId: 'Azure/azure-sdk-for-java'
 
+      - script: |
+          python ./eng/scripts/inject_codesnippets.py
+        displayName: 'Replace @codesnippet with embedded pre element'
+        workingDirectory: $(Build.SourcesDirectory)
+
```
JimSuplizio: /CC @weshaggard @mitchdenny 
It's not clear to me if this is the correct place to be calling this or if this should be similar to what I'd done for the generate overview. This does differ from the replace relatively links as that is used across all the language and this is specific to Java. If it used the same approach as the generate overview stuff then it could be tested locally as part of the build with a flag instead of running a script across the entire repo and then building a given library (or submitting a PR and having to download artifacts). The code snippets for a given library are always going to be under that library's directory in the repo (typically under the /samples subdirectory) so processing as part of the library build would effectively work the same way. The changes would minimal and he could follow the same pattern which would be to declare a profile in [pom.client.xml](https://github.com/Azure/azure-sdk-for-java/blob/master/pom.client.xml#L1022) and then pass in that flag inside [archetype-sdk-client.yml](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/templates/jobs/archetype-sdk-client.yml#L117). It would also mean that the python script would have to change to take a single directory as an argument.

scbedd: I'm planning on adjusting to act just like generate-overview after your feedback Jim.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -111,7 +111,7 @@ jobs:
         inputs:
           mavenPomFile: pom.xml
           goals: 'package'
-          options: '$(DefaultOptions) "-DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -DskipTests -Dgenerate-overview -pl $(ProjectList) -am' # We include template-module so we ensure it always builds in CI
+          options: '$(DefaultOptions) "-DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -DskipTests -Dinject-codesnippets -Dgenerate-overview -Dinject-codesnippets -pl $(ProjectList) -am' # We include template-module so we ensure it always builds in CI
```
scbedd: Oh whoops I already did this once. Fixing.

pom.client.xml
```diff
@@ -1073,6 +1073,50 @@
...
+                <goals>
+                  <goal>exec</goal>
+                </goals>
+              </execution>
+            </executions>
+            <configuration>
+              <executable>python</executable>
+              <workingDirectory>${project.basedir}</workingDirectory>
+              <arguments>
+                <argument>../../../eng/pipelines/scripts/inject_codesnippets.py</argument>
```
weshaggard: I'm not a big fan of calling python from maven but we already do it and you made it opt-in so I'll live with it.

PR #9334, opened 2020-03-20
Cosmos API Review Iteration - Part 2

kushagraThapar:
* Moved Indexkind to implementation package
* Reduced scope of JsonSerializable public APIs to protected / package-private APIs.


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/DatabaseAccount.java
```diff
@@ -285,4 +285,9 @@ public String toJson() {
         this.populatePropertyBag();
         return super.toJson();
     }
+
+    @Override
+    public Object get(String propertyName) {
+        return super.get(propertyName);
```
moderakh: why do we need these overridden methods?
the implementation is the same as parent class.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/Offer.java
```diff
@@ -132,4 +132,14 @@ private ObjectNode getContent() {
     private void setContent(ObjectNode offerContent) {
         BridgeInternal.setProperty(this, Constants.Properties.OFFER_CONTENT, offerContent);
     }
+
+    @Override
+    public String getString(String propertyName) {
+        return super.getString(propertyName);
```
moderakh: ditto, here and other places for the new methods, the implementation is the same as parent class why are they needed in every child class?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/PartitionKeyRange.java
```diff
@@ -123,4 +123,9 @@ public void setParents(List<String> parents) {
      * @return a list partition key range ID
      */
     public List<String> getParents() { return this.getList(Constants.Properties.PARENTS, String.class); }
+
+    @Override
+    public String toJson() {
+        return super.toJson();
```
moderakh: ditto

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/Address.java
```diff
@@ -87,4 +87,9 @@ public String getParitionKeyRangeId() {
     public void setPartitionKeyRangeId(String partitionKeyRangeId) {
         BridgeInternal.setProperty(this, Constants.Properties.PARTITION_KEY_RANGE_ID, partitionKeyRangeId);
     }
+
+    @Override
+    public Object get(String propertyName) {
+        return super.get(propertyName);
```
moderakh: ditto

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/DatabaseAccount.java
moderakh: discussed offline.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncUserDefinedFunction.java
```diff
@@ -73,7 +73,7 @@ CosmosAsyncUserDefinedFunction setId(String id) {
     public Mono<CosmosAsyncUserDefinedFunctionResponse> replace(CosmosUserDefinedFunctionProperties udfSettings) {
         return container.getDatabase()
                    .getDocClientWrapper()
-                   .replaceUserDefinedFunction(new UserDefinedFunction(udfSettings.toJson()), null)
+                   .replaceUserDefinedFunction(new UserDefinedFunction(ModelBridgeInternal.toJsonFromJsonSerializable(udfSettings)), null)
```
mbhaskar: (Could be out of scope of this PR) Potential double serialization. Instead of converting toJson and again back from Json, we could just use the propertybag to get UDF by using Resource(ObjectNode) overload.

kushagraThapar: Will discuss offline on this one with you.

moderakh: outside of scope of this PR. we can come back to this later. I fixed double serialization for point read item operation but not the udf, etc.

PR #9337, opened 2020-03-20
Refactor ExpandableStringEnum

alzimmermsft:
Refactoring of `ExpandableStringEnum`'s internal data structure and added unit tests.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/ExpandableStringEnum.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/ExpandableStringEnum.java
@@ -7,10 +7,9 @@ import com.fasterxml.jackson.annotation.JsonValue;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashSet;
-import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
 /**
  * Base implementation for expandable, single string enums.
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/util/ExpandableStringEnum.java
```diff
@@ -7,39 +7,22 @@
...
 
 /**
  * Base implementation for expandable, single string enums.
  *
  * @param <T> a specific expandable enum type
  */
 public abstract class ExpandableStringEnum<T extends ExpandableStringEnum<T>> {
-    private static final ConcurrentMap<String,
-        ? extends ExpandableStringEnum<?>> VALUES_BY_NAME = new ConcurrentHashMap<>();
+    private static final Map<Class<?>, ConcurrentHashMap<String, ? extends ExpandableStringEnum<?>>> VALUES
```
jianghaolu: Are we going to have enum members of different types in the same enum?

jianghaolu: Scratch the above comment. A question: Is there any reason to store all enum values in one map, if we don't ever call `ExpandableEnumString.fromString()`?

alzimmermsft: Shouldn't, we can't currently. This is just simplifying the internal data structure to scope by enum type first then having the name -> enum mapping, before we just mutated the enum name with the class name to make them unique. This work correctly but had the downside in calls to `values` where we ended up duplicating all the `ExpandableStringEnum`s that were ever loaded and checking their type to determine if it was a value of the type we were looking for.

alzimmermsft: The is an example of what we had before:

EnumClass1#value1 -> value1
EnumClass1#value2 -> value2
EnumClass2#value1 -> value1

This is what it would change to:
Class1
  value1 -> value1
  value2 -> value2
Class2
  value1 -> value1

PR #9353, opened 2020-03-21
Move Jetty configuration

mitchdenny:
This PR moves the Jetty configuration file out of the root directory. Just following our general pattern of keeping loose files out of the root if they aren't necessary.


mitchdenny:
> /azp run java - keyvault - data

We've all been there :)

PR #9364, opened 2020-03-23
Add support for listing additional modules to build.

mitchdenny:
This PR aims to address a problem that was introduced with the POM restructuring where the -am switch which targets the shipping artifacts doesn't include some internal only modules (like ```azure-cosmos-benchmark```).

Aiming to get this up and running for Cosmos first since they have the immediate need, but we'll also use this on things like Core which we also want to build downstream dependents.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -107,6 +107,7 @@ jobs:
       - template: ../steps/generate-project-list.yml
         parameters:
           Artifacts: ${{parameters.Artifacts}}
+          AdditionalModules: ${{parameters.AdditionalModules}}
 
       - task: Maven@3
         displayName: 'Build and Package'


mitchdenny:
@kushagraThapar does this look like a regular Cosmos test failure?

kushagraThapar:
> @kushagraThapar does this look like a regular Cosmos test failure?

Yes, I have verified, and we are good to get this in.

sdk/core/ci.yml
```diff
@@ -57,3 +57,6 @@ stages:
         - name: azure-core-tracing-opentelemetry
           groupId: com.azure
           safeName: azurecoretracingopentelemetry
+      AdditionalModules:
+        - name: azure-data-appconfiguration
```
weshaggard: I'm not sure I fully understand why we need to add this here. Doesn't appconfig depend on Core and thus the `-am` should cause it to build?

sdk/cosmos/ci.yml
```diff
@@ -45,3 +45,8 @@ stages:
         - name: azure-cosmos
           groupId: com.azure
           safeName: azurecosmos
+      AdditionalModules:
```
JimSuplizio: @kushagraThapar @mitchdenny @weshaggard 
I realize that we're doing this to allow Cosmos pipelines to pass but the examples and benchmark libraries are supposed to be going away which means that these specific edits should ultimately be going away.

PR #9392, opened 2020-03-23
Add EnvVar AZURE_AUTHORITY_HOST

Luyunmt:
Added EnvVar AZURE_AUTHORITY_HOST to azure.identity._constants.
TokenCredentialOptions has an AuthorityHost property.
This will be modified to support EnvVar override with AZURE_AUTHORITY_HOST.
Azure/azure-sdk-for-java/issues/5967


sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
```diff
@@ -26,7 +27,9 @@
      * Creates an instance of IdentityClientOptions with default settings.
      */
     public IdentityClientOptions() {
-        authorityHost = DEFAULT_AUTHORITY_HOST;
+        Configuration configuration = Configuration.getGlobalConfiguration().clone();
```
alzimmermsft: Is there a need to have a short lived clone of the global configuration? If anything it may be better to use it directly as it could load into static memory the environment setting and will be accessed quicker in the future.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
```diff
@@ -26,7 +27,9 @@
...
+        authorityHost = configuration.contains(configuration.PROPERTY_AZURE_AUTHORITY_HOST)
+        ? configuration.get(configuration.PROPERTY_AZURE_AUTHORITY_HOST) : DEFAULT_AUTHORITY_HOST;
```
alzimmermsft: `Configuration` has a `T get(String configurationName, T defaultValue)` method that could simplify this logic.

```java
authorityHost = configuration.get(Configuration.PROPERTY_AZURE_AUTHORITY_HOST, DEFAULT_AUTHORITY_HOST);
```

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClientOptions.java
Luyunmt: I followed they way in SharedTokenCacheCredential，it cloned a configuration in its constructor.

Luyunmt: Removed clone.

jongio:
Can you please add tests?

g2vinay:
Rebase with master.
and change the core dependency in azure-identity/pom.xml to this:
    <dependency>
      <groupId>com.azure</groupId>
      <version>1.4.0-beta.1</version> 
    </dependency>

PR #9398, opened 2020-03-23
Remove unused Cosmos entries from version_data.txt

JimSuplizio:
V3 sources were removed from master when v4 Cosmos was added. These entries should no longer be in version_data.txt.


JimSuplizio:
@mitchdenny, I'm actually looking at this locally. The original PR's path to get jetty.xml wasn't correct however at least one of my 3 commits should have been. I'm trying to run locally to see what's going on here.

PR #9406, opened 2020-03-23
Update jetty and com:microsoft:azure-keyvault dependencies

conniey:
Update jetty dependencies to latest versions.


conniey:
Because kv version is incorrect.

PR #9410, opened 2020-03-23
Print Azure Core Test Name Before Running

alzimmermsft:
This PR adds a before each test task to print the name of the currently running test. This will enable easier test debugging if a test we to hang in our build and validation pipelines.

### DIFF
--- a/eng/pipelines/templates/variables/globals.yml
+++ b/eng/pipelines/templates/variables/globals.yml
@@ -1,7 +1,7 @@
 variables:
   DocWardenVersion: '0.5.0'
   DefaultOptions: '--batch-mode --fail-at-end --settings eng/settings.xml'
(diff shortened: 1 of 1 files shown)


mitchdenny:
I'm surprised that there isn't a configuration to do this for the test runner?

alzimmermsft:
> 
> 
> I'm surprised that there isn't a configuration to do this for the test runner?

JUnit 4 has `RunListener` and JUnit 5 has a `TestExecutionListener`. The `TestExecutionListener` is an interface where the implementing class could be added to tests via an annotation or presented as an SPI that will be hooked up implicitly. I can take a quick dive into what it would take to get a proper listener for the JUnit 5 tests.

alzimmermsft:
Closing in favor of #10313

PR #9413, opened 2020-03-24
Add AzureKeyCredential to Azure Core

alzimmermsft:
This PR adds `AzureKeyCredential` to Azure Core and updates Search and Text Analytics to use it. The credential replaces the service specific credential, therefore `SearchApiKeyCredential` and `TextAnalyticsApiKeyCredential` have been deleted.

Additionally, this PR moves `TextAnalyticsApiKeyPipelinePolicy` into the public API in the root package.

For reference this is the .NET PR (https://github.com/Azure/azure-sdk-for-net/pull/10668) and the TS PR (https://github.com/Azure/azure-sdk-for-js/pull/7927). @maririos and @willmtemple


sdk/core/azure-core/src/main/java/com/azure/core/credential/AzureKeyCredential.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Rotates the key associated to this credential.
+     *
+     * @param key The new key to associated with this credential.
+     * @return The updated {@code ApiKeyCredential} object.
+     * @throws NullPointerException If {@code key} is {@code null}.
+     * @throws IllegalArgumentException If {@code key} is an empty string.
+     */
+    public AzureKeyCredential updateKey(String key) {
```
maririos: Could we name this method `update` to be consistent across languages?

alzimmermsft: Renamed to `update`.

sdk/search/azure-search/src/main/java/com/azure/search/documents/SearchApiKeyPipelinePolicy.java
```diff
@@ -12,31 +13,37 @@
...
 /**
- * Policy that adds the Cognitive Search Service api-key into the request's Authorization header.
+ * Pipeline policy that uses an {@link AzureKeyCredential} to set the {@code api-key} header which authorizes requests
+ * sent to the Azure Search service.
+ *
+ * <p>
+ * Requests sent with this pipeline policy are required to use {@code HTTPS}. If the request isn't using {@code HTTPS}
+ * an exception will be thrown to prevent leaking the key.
  */
 public final class SearchApiKeyPipelinePolicy implements HttpPipelinePolicy {
```
maririos: Could the policies be private?

maririos: So this means that every client needs to implement it's own policy, is that correct?
Wondering if we could add a generic policy to Azure Core too

alzimmermsft: I was discussing this with @JonathanGiles and we didn't strongly lean on one design or the other. I left the library specific policies in place given they are in preview and adding this into Azure Core would likely require it being part of the public API which could GA it sooner (Java doesn't have a good friends assembly or internal that .NET leverages).

@JonathanGiles any concern around making this a generic Azure Core policy that uses the credential and a header string? We do something conceptually similar with the `RetryPolicy` where we can pass a retry header to look for on the response.

alzimmermsft: Java has a guideline where we're expected to have our pipeline policies in the public API. We allow customers to roll their own pipeline if they wish, so making these inaccessible from the public API would make that more difficult.

@JonathanGiles to point to the guideline since I'm having trouble finding it. (Or to add it if it is missing)

maririos: Thanks for the clarification Alan!

maririos: >I left the library specific policies in place given they are in preview

Note that Text Analytics is going to GA in April, so if we could add the policy to Core now, that will be great for TA.

schaabs: Couldn't the library specific policies be internal to the package, or added under implementation? Then you can always update them to use an implementation exposed by core at some point in the future.

alzimmermsft: Yes, this could be done but goes a bit against the response above around why the policy is in the public API.

mssfang:
Is possible we can think of one generic KeyCredentialPolicy that applies to all Cognitive Service's ApiKeyCredential.

sdk/core/azure-core/src/main/java/com/azure/core/credential/AzureKeyCredential.java
```diff
@@ -0,0 +1,59 @@
...
+     * Rotates the key associated to this credential.
+     *
+     * @param key The new key to associated with this credential.
+     * @return The updated {@code ApiKeyCredential} object.
+     * @throws NullPointerException If {@code key} is {@code null}.
+     * @throws IllegalArgumentException If {@code key} is an empty string.
+     */
+    public AzureKeyCredential update(String key) {
+        Objects.requireNonNull(key);
+        if (key.isEmpty()) {
```
sima-zhu: Could you use one check of CoreUtil.isNullOrEmpty() here?

sdk/search/azure-search/src/main/java/com/azure/search/documents/SearchApiKeyPipelinePolicy.java
JonathanGiles: The wording might be unclear but it's [here](https://azure.github.io/azure-sdk/java_implementation.html#java-auth-policy-impl) - "DO provide a suitable authentication policy that authenticates the HTTP request in the HTTP pipeline when using non-standard credentials. This includes custom connection strings, if supported."

sdk/core/azure-core/src/main/java/com/azure/core/credential/AzureKeyCredential.java
alzimmermsft: Definitely could but our guidelines state if we expect a non-null parameter and we get null a `NullPointerException` should be thrown (the `Objects.requireNonNull` check) and then if we get a value that we don't expect that isn't null throw a `IllegalArgumentException` (the `isEmpty` check).

sdk/search/azure-search/src/main/java/com/azure/search/documents/SearchApiKeyPipelinePolicy.java
schaabs: In this case I would push to move up the shared Azure.Core implementation of a generic auth policy as in the discussion below, so we don't have these types left around as artifacts in libraries that GA before it's available

alzimmermsft:
> 
> 
> Is possible we can think of one generic KeyCredentialPolicy that applies to all Cognitive Service's ApiKeyCredential.

It is possible to create a generic `HttpPipelinePolicy` for this but this requires additional design. I'll work through this with @JonathanGiles and should have an answer before the end of the week.

sdk/search/azure-search/src/main/java/com/azure/search/documents/SearchApiKeyPipelinePolicy.java
alzimmermsft: If it is okay, I'm going to merge this PR as it keeps hitting merge conflicts due to its scope and file an issue to follow-up on a generic policy.

alzimmermsft:
Merging this PR in, issue #9535 tracks adding an `AzureKeyCredentialPolicy` into Azure Core.

PR #9435, opened 2020-03-24
POM restructure phase 2

mitchdenny:
In this PR I am moving the parent POMs from the root of the source tree up into the ```sdk/parents``` folder with a sub-directory for each POM. This gets them out of the root and encourages people to use the ```./pom.xml``` file to build (and they don't really need to think about the parents that much) but more important for this restructuring effort, it will help draw out any areas in our engineering system where we have a hard dependency on the location of the parent POMs and the module list within them (since we shouldn't be using that list).

This will require changes to all of the track 1, track 2 and management plane POMs but we'll start with a few just to find the issues before we scale out and start really impacting the build pool.

### DIFF
--- a/eng/pipelines/aggregate-reports.yml
+++ b/eng/pipelines/aggregate-reports.yml
@@ -16,13 +16,13 @@ jobs:
       displayName: 'Generate Reports'
       inputs:
         mavenPomFile: pom.xml
-        options: '$(DefaultOptions) -Dinclude-non-shipping-modules -Ddependency-checker -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true -pl $(ProjectList) -am'
+        options: '$(DefaultOptions) -Dinclude-non-shipping-modules -Ddependency-checker -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true -pl com.azure:jacoco-test-coverage -am'
         mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
         javaHomeOption: 'JDKVersion'
         jdkVersionOption: '1.11'
         jdkArchitectureOption: 'x64'
         publishJUnitResults: false
-        goals: 'install site:site site:stage verify'
+        goals: 'clean verify'
 
     - task: PublishTestResults@2
       condition: succeededOrFailed()


mitchdenny:
/cc @alzimmermsft @JimSuplizio @JonathanGiles @kushagraThapar @conniey @weshaggard 

FYI. This is the second phase of restructuring. Actually moving the parents :) Not ready to merge yet but figured you might have input on the journey.

sdk/parents/azure-client-sdk-parent/pom.xml
```diff
@@ -629,8 +628,8 @@
         <artifactId>maven-checkstyle-plugin</artifactId>
         <version>3.1.0</version> <!-- {x-version-update;org.apache.maven.plugins:maven-checkstyle-plugin;external_dependency} -->
         <configuration>
-          <configLocation>eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml</configLocation>
-          <suppressionsLocation>eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
+          <configLocation>../../../eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml</configLocation>
+          <suppressionsLocation>../../../eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
```
JimSuplizio: These configLocation and suppressionsLocation entries should be exactly the same, the pathing between the two modified entries is different. Also, by that note there a number of other items that may need to be updated for the pathing
- headerLocation entries [1](https://github.com/Azure/azure-sdk-for-java/blob/f691e525e75585111b3f5e8af693e49e7ff36fb1/sdk/parents/azure-client-sdk-parent/pom.xml#L143) and [2](https://github.com/Azure/azure-sdk-for-java/blob/f691e525e75585111b3f5e8af693e49e7ff36fb1/sdk/parents/azure-client-sdk-parent/pom.xml#L634)
- spotbugs [excludeFilterFile](https://github.com/Azure/azure-sdk-for-java/blob/f691e525e75585111b3f5e8af693e49e7ff36fb1/sdk/parents/azure-client-sdk-parent/pom.xml#L668)
There are also other things that may need to get modified like the [non-shipping-modules](non-shipping-modules). If so then pom.data.xml would also need to get updated for its non-shipping modules.

alzimmermsft:
You'll need to update the `pom.client.xml` usages in pipeline YMLs, one place I know this affects right away is the aggregate report pipeline.

eng/pipelines/aggregate-reports.yml
```diff
@@ -10,11 +10,6 @@ jobs:
     pool:
       vmImage: 'ubuntu-16.04'
     steps:
-    - template: templates/steps/install-reporting-tools.yml
```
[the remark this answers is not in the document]
mitchdenny: I think I'll close the other PR since I have to deal with some of this stuff in this one.

sdk/core/azure-core-amqp/pom.xml
```diff
@@ -8,7 +8,7 @@
     <groupId>com.azure</groupId>
     <artifactId>azure-client-sdk-parent</artifactId>
     <version>1.7.0</version> <!-- {x-version-update;com.azure:azure-client-sdk-parent;current} -->
-    <relativePath>../../../pom.client.xml</relativePath>
+    <relativePath>../../parents/azure-sdk-client-parent</relativePath>
```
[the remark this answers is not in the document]
mitchdenny: No I don't think so. Whenever you specify a path Maven will default try to find the ```pom.xml``` file in that directory, and since we are having one directory per parent POM I think this will fork fine. I did notice that I had my POM name incorrect though (should be ```azure-client-sdk-parent``` instead of ```azure-sdk-client-parent``` :0)

sdk/parents/azure-client-sdk-parent/pom.xml
mitchdenny: Yep. I've got local changes I haven't pushed yet, I'm stepping through all this. It's kinda messy.

mitchdenny:
> You'll need to update the `pom.client.xml` usages in pipeline YMLs, one place I know this affects right away is the aggregate report pipeline.

I was thinking of making the aggregate pipeline just pass in all the modules as additional modules ;)

JimSuplizio:
> > You'll need to update the `pom.client.xml` usages in pipeline YMLs, one place I know this affects right away is the aggregate report pipeline.
> 
> I was thinking of making the aggregate pipeline just pass in all the modules as additional modules ;)

Rather than hard code the list of additional modules (meaning there is one more place to update if a library is added, removed or changed) I would rather this be dynamically done in some way, shape or form.

mitchdenny:
> Rather than hard code the list of additional modules (meaning there is one more place to update if a library is added, removed or changed) I would rather this be dynamically done in some way, shape or form.

Agree in principle. Just need to find a way that we can do it in practice. I don't like that we are having to write all kinds of external helper scripts to do things like readme checks, version checks etc. A lot of the heavy lifting in those scripts is processing and understand the build graph in each repository, however if we leveraged the tools that each ecosystem provides we could probably make things easier on ourselves.

Taking version checking as an example, we could probably write a mojo that runs against each project within the scope of the build graph which then reads the POM file to extract out the version tags that you've got and match them against the actual dependencies and your version list.

That's just an example, but I would rather embrace what the ecosystem provides if we can.

mitchdenny:
@JimSuplizio @JonathanGiles @alzimmermsft I think that this is ready to go. Although this is just moving a few files around I think that there is a chance that various scripts could break later in the release pipeline so I want to delay merging this until after we do the releases to avoid any disruption there.

mitchdenny:
So this breaks the aggregate report, but thankfully I think I can fix it pretty quickly. I'll get this sorted before I merge it in (it may actually speed it up, but will definitely make it simpler).

sdk/storage/azure-storage-blob-cryptography/pom.xml
```diff
@@ -59,12 +59,6 @@
       <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core-test;dependency} -->
       <scope>test</scope>
     </dependency>
-    <dependency>
```
mitchdenny, line 62: @alzimmermsft & @JimSuplizio I'd like to get your opinion on this change. I wasn't sure why we ended up with this dependency being listed twice. So I removed one.

eng/pipelines/aggregate-reports.yml
```diff
@@ -7,61 +7,22 @@ jobs:
     timeoutInMinutes: 180
     variables:
       - template: templates/variables/globals.yml
+      - name: ProjectList
+        value: com.azure:azure-data-appconfiguration,com.azure:azure-core,com.azure:azure-core-amqp,com.azure:azure-core-http-jdk-httpclient,com.azure:azure-core-http-netty,com.azure:azure-core-http-okhttp,com.azure:azure-core-test,com.azure:azure-core-tracing-opentelemetry,com.azure:azure-cosmos,com.azure:azure-messaging-eventhubs,com.azure:azure-messaging-eventhubs-checkpointstore-blob,com.azure:azure-identity,com.azure:azure-security-keyvault-certificates,com.azure:azure-security-keyvault-keys,com.azure:azure-security-keyvault-secrets,com.azure:azure-storage-blob,com.azure:azure-storage-blob-cryptography,com.azure:azure-storage-blob-nio,com.azure:azure-storage-common,com.azure:azure-storage-file-datalake,com.azure:azure-storage-file-share,com.azure:azure-storage-queue,com.azure:azure-sdk-template,com.azure:azure-ai-textanalytics
```
alzimmermsft, line 11: Does this get generated? This concerns me a bit as this seems like another one of those scenarios where we continuously forget to update this as we move libraries around or on-board new libraries.

eng/pipelines/aggregate-reports.yml
```diff
@@ -7,61 +7,22 @@ jobs:
...
-        goals: 'install'
-
     - task: Maven@3
-      displayName: 'Generate Maven project site, including JavaDocs, SpotBugs, and CheckStyle reports'
+      displayName: 'Generate Reports'
       inputs:
-        mavenPomFile: pom.client.xml
-        options: '$(DefaultOptions) -DskipTests -Dgpg.skip'
+        mavenPomFile: pom.xml
+        options: '$(DefaultOptions) -Dinclude-non-shipping-modules -Ddependency-checker -Dgpg.skip -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true -pl com.azure:jacoco-test-coverage -am'
```
alzimmermsft, line 19: Great find by using `-am` on the `jacoco-test-coverage` POM which will include all client libraries that we release.

sdk/storage/azure-storage-blob-cryptography/pom.xml
alzimmermsft: This likely predates some other changes made to our POMs, thank you for removing it 😃

eng/pipelines/aggregate-reports.yml
mitchdenny: At the moment this is just a dead body in the aggregate-reports.yml file. I'll be deleting that line in a subsequent PR but I want to avoid changes if I can so I don't eat up the agent pool again. Getting the aggregate reporting pipeline working correctly is going to be a phase 3 I think :)

eng/pipelines/aggregate-reports.yml
mitchdenny: Yeah ... this is why the project list isn't required anymore. It still isn't working correctly but I suspect that this is just a path issue that I need to work through.

sdk/storage/azure-storage-blob-cryptography/pom.xml
JimSuplizio: @mitchdenny I'd actually had this change in the PR I'm working on as well which is delayed due to ongoing docs fun.

PR #9436, opened 2020-03-24
Remove reporting tools install

mitchdenny:
With the POM restructure we shouldn't need the install step for the reporting tools anymore.


mitchdenny:
Closed because it'll be cleaned up with the next POM restructuring phase.

PR #9467, opened 2020-03-24
Decouple guava 25.0-jre dependency

anuchandy:
Removed azure-cosmos source dependency on the external guava-25.0-jre package.

Created a local package
com.azure.cosmos.implementation.guava25 that contains required guava types. The azure-cosmos source types are updated to use types from this package.

The non-shipping packages - sample, benchmark, and the test will continue to take a dependency on the external guava-25.0-jre package.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
+++ b/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
@@ -1936,6 +1936,7 @@
     UMAC_UNCALLABLE_METHOD_OF_ANONYMOUS_CLASS,
     UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR,
     UWF_UNWRITTEN_FIELD"/>
+    </Match>
     
   <!-- Empty byte array is returned as null to reactor. Reactor transform null from Callable to downstream onComplete  -->
   <Match>


sdk/cosmos/azure-cosmos/pom.xml
```diff
@@ -102,6 +102,19 @@ Licensed under the MIT License.
...
+         which is used in @Nullable annotation in reactor core classes.
+
+         Similar provided dependency exits for "azure-core" as well.
+    -->
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>3.0.2</version> <!-- {x-version-update;com.google.code.findbugs:jsr305;external_dependency} -->
+      <scope>provided</scope>
+    </dependency>
```
anuchandy: azure-core reference: https://github.com/Azure/azure-sdk-for-java/blob/dea52be2efa93c5abecfdd0582b03a9e046cc957/sdk/core/azure-core/pom.xml#L73

mbhaskar:
@anuchandy @JonathanGiles Is this process of copying classes from guava to our implementation a continuous one? I mean lets say we need a new class from guava in future, do we copy it then?

JonathanGiles:
> @anuchandy @JonathanGiles Is this process of copying classes from guava to our implementation a continuous one? I mean lets say we need a new class from guava in future, do we copy it then?

No, my preference would be that we lessen our dependency, not strengthen it.

Issue #9189, opened 2020-03-18
Upload From File Reporter Tests Consistently Fail


alzimmermsft:
Beginning on 03/18 the `Upload from file reporter` tests in Blobs are failing with an unexpected reporting count. These don't appear to be flaky tests as the count returned vs expected is identical across all 6 test runs.

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=312439&view=ms.vss-test-web.build-test-results-tab&runId=9642130&resultId=101680&paneView=debug

alzimmermsft:
@gapra-msft it appears this is still disabled, could we see if this is resolved?

gapra-msft:
Yeah we can see if my recent PR helps this issue.

gapra-msft:
Just linking the PR https://github.com/Azure/azure-sdk-for-java/pull/20964/files

PR #9468, opened 2020-03-24
Disabling Upload From File Reporter

alzimmermsft:
This PR is disabling the `Upload from file reporter` tests which are consistently failing in the live test run with an unexpected reported byte count. Issue #9189 tracks the request to fix these tests and re-enable them in the live test run.


sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
```diff
@@ -779,6 +780,7 @@ class BlockBlobAPITest extends APISpec {
 
     @Unroll
     @Requires({ liveMode() })
+    @Ignore("Failing in live test run with unexpected reported byte count")
```
srnagar: Should this be fixed instead of marking the test as ignored? Incorrectly reported byte count seems like a bug.

PR #9472, opened 2020-03-24
Fixes Service Bus and Event Hubs integration tests

conniey:
* Not creating a new ManagementChannel for every metadata operation
* Deferring the evaluation of renew lock method/

### DIFF
--- a/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
+++ b/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
@@ -145,6 +145,10 @@
   <suppress checks="IllegalImport" files=".*[/\\]com[/\\]azure[/\\]core[/\\]tracing[/\\]opentelemetry[/\\]*"/>
   <suppress checks="IllegalImport" files=".*[/\\]com[/\\]azure[/\\]identity[/\\]*"/>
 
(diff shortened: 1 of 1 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/EventHubReactorAmqpConnection.java
```diff
@@ -87,9 +89,15 @@ public EventHubReactorAmqpConnection(String connectionId, ConnectionOptions conn
                 "connectionId[%s]: Connection is disposed. Cannot get management instance", connectionId))));
         }
 
-        return Mono.defer(() -> getReactorConnection().thenReturn(new ManagementChannel(
-                createRequestResponseChannel(MANAGEMENT_SESSION_NAME, MANAGEMENT_LINK_NAME, MANAGEMENT_ADDRESS),
-                eventHubName, tokenCredential, tokenManagerProvider, this.messageSerializer, scheduler)));
+        return getReactorConnection().then(Mono.fromCallable(() -> {
+            if (managementChannel == null) {
```
srnagar: This should be synchronized or declare `this.managementChannel` as AtomicReference.

PR #9475, opened 2020-03-24
Updated versions to 3.7.1 for cosmos

kushagraThapar:


moderakh:
please also update changelog file.

weshaggard:
The PR validation didn't trigger because the branch name doesn't match https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/cosmos/ci.yml#L29-L32.

PR #9477, opened 2020-03-24
Updated versions to 3.7.1 for cosmos

kushagraThapar:


JimSuplizio:
> is macos version update also needs to be done on v4? master?
@moderakh, the macOS update happened on master a few weeks back, after v4 was merged into master. Since the SHA for v3 was before we'd updated the MacOS version, you guys just had to make the change in your branch.

PR #9478, opened 2020-03-25
Rename module name azure-search to azure-search-documents

sima-zhu:


sdk/search/azure-search-documents/README.md
```diff
@@ -1,7 +1,7 @@
-# Azure cognitive search client library for Java
+# Azure cognitive search documents client library for Java
```
alzimmermsft: The service is still `Azure Search`, so no need to update the README piece which use the service name. The reason for the change from `azure-search` to `azure-search-documents` is that we wanted to preserve the top level package/namespace for `com.azure.search`, additionally it resolves a package issue with Python.

sima-zhu:
Regenerate code. No need to change swagger transformation.

PR #9488, opened 2020-03-25
Remove .azure-pipelines directory

mitchdenny:
This PR removes the ```.azure-pipelines``` directory. @mikeharder @scbedd I think these files aren't used any longer?


mikeharder:
@mitchdenny: I don't know the history of this pipeline.  I defer to @scbedd since this is related to docs.

PR #9494, opened 2020-03-25
Api Management: generate 2019 12

ChenTanyi:


sdk/apimanagement/mgmt-v2019_12_01/src/main/java/com/microsoft/azure/management/apimanagement/v2019_12_01/implementation/EmailTemplateContractImpl.java
```diff
@@ -31,7 +31,7 @@
         super(name, new EmailTemplateContractInner());
         this.manager = manager;
         // Set resource name
-        this.templateName = name;
+        this.templateName = TemplateName.fromString(name);
```
weidongxu-microsoft: I actually not sure whether this works or not ...

ChenTanyi: I see the previous one, 2019_01_01, it also uses such change. This works, but I'm not sure if it is the correct logic. I could only change it like this due to the generation.

PR #9529, opened 2020-03-25
Set New Date Header on Request Retry

alzimmermsft:
This PR resolves an issue where the Date header wouldn't be updated on request retry. If the Date header isn't updated on request retry this could lead to an issue where the Date header becomes stale and is no longer considered valid by the service, correcting the pipelines created by the client builders to do so resolves this issue.


alzimmermsft:
@gapra-msft or @rickle-msft when you get a chance could I get a review on this PR

PR #9532, opened 2020-03-25
SDK Doc change: GetMessageSessions documentation not clear

DorothySun216:
Azure Java SDK doc change: https://msazure.visualstudio.com/DefaultCollection/One/_sprints/taskboard/AzureMessaging/One/AzureMessaging/Manganese/2003/ServiceBusCore-2003?workitem=5766751


DorothySun216:
Should I create a feature branch for this change and create the PR from feature branch? I initially think the change is very small so I didn't create a branch for it. Will update to a feature branch if it is the convention.

PR #9537, opened 2020-03-25
Added AzureKeyCredentialPolicy to Azure Core

### LINKED ISSUE #9535: Investigate Adding AzureKeyCredentialPolicy into Azure Core
PR #9413 added an `AzureKeyCredential` into Azure Core to replace the service specific key credential classes being used in Azure Search and Text Analytics. It should be investigated whether the same could be done for the specific credential policies that these modules are using.

alzimmermsft:
Resolves #9535 

This PR adds `AzureKeyCredentialPolicy` into Azure Core, this policies replaces the service specific `TextAnalyticsApiKeyCredentialPolicy` and `SearchApiKeyCredentialPolicy` classes. `AzureKeyCredentialPolicy` acts as a generic policy that will set the specified header to the key contained in the passed `AzureKeyCredential` object.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AzureKeyCredentialPolicy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AzureKeyCredentialPolicy.java
@@ -13,7 +13,7 @@ import reactor.core.publisher.Mono;
 import java.util.Objects;
 
 /**
- * Pipeline policy that uses an {@link AzureKeyCredential} to set the specified header which authorizes requests.
+ * Pipeline policy that uses an {@link AzureKeyCredential} to set the authorization key for a request.
  *
  * <p>
  * Requests sent with this pipeline policy are required to use {@code HTTPS}. If the request isn't using {@code HTTPS}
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AzureKeyCredentialPolicy.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Creates a policy that uses the passed {@link AzureKeyCredential} to set the specified header.
+     *
+     * @param header The header that will be set to the {@link AzureKeyCredential} key value.
+     * @param credential The {@link AzureKeyCredential} containing the authorization key to use.
+     * @throws NullPointerException If {@code header} or {@code credential} is {@code null}.
+     * @throws IllegalArgumentException If {@code header} is empty.
+     */
+    public AzureKeyCredentialPolicy(String header, AzureKeyCredential credential) {
```
maririos: Alan, on Python land, Johan is suggesting to change the parameter name from `header` to `name` so whenever we have a service that uses query keys instead of header keys, this name will work.
I am considering changing it in .NET too so we are consistent
More info here: https://github.com/Azure/azure-sdk-for-python/pull/10509#discussion_r398744760

What do you think?

alzimmermsft: Works with me, this is a non-breaking change for Java since we don't have named arguments.

alzimmermsft: Though if we have plans to support both headers and query parameters for the passed key Java may want to change this interface a bit to have a flag such as `isHeaderKey` and context switch on that to either add it as a header or a query parameter. This could also be added later as a non-breaking change.

PR #9539, opened 2020-03-25
[Tracing] Add EventHubs info on outgoing Message spans.

samvaity:
Currently, message spans do not have any information about event hub they are being sent to.
Stamp event hub properties on message spans:

`message_bus.destination`: EventHubs entity name
`peer.address`: Fully qualified EventHubs name such as `<name>.servicebus.windows.net/`

image

Ref: #7725


sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventDataBatchIntegrationTest.java
```diff
@@ -83,7 +83,7 @@ public void sendSmallEventsFullBatch() {
     public void sendSmallEventsFullBatchPartitionKey() {
         // Arrange
         final EventDataBatch batch = new EventDataBatch(ClientConstants.MAX_MESSAGE_LENGTH_BYTES, null, PARTITION_KEY, contextProvider,
-            new TracerProvider(Collections.emptyList()));
+            new TracerProvider(Collections.emptyList()), null, null);
```
conniey: Instead of using null, null.. can you use getFullyQualifiedDomainName() and getEventHubName() ? These properties are set in IntegrationTestBase.

samvaity: updated.

Issue #5361, opened 2019-09-12
[FEATURE REQ][Cosmos]Expose current of Change Feed Processor (isStarted/getStatus)


christopheranderson:
We should be able to see whether the ChangeFeedProcessor is started or stopped.

getStatus is most useful. Sugar method for isStarted might be convenient.

kushagraThapar:
@milismsft Do we have a plan for this ?

milismsft:
@kushagraThapar It should be fairly easy to expose such a method like "isStarted()" since we already have a property that keeps track of the overall state.

kushagraThapar:
> @kushagraThapar It should be fairly easy to expose such a method like "isStarted()" since we already have a property that keeps track of the overall state.

Do we plan to do this in this upcoming 3.4.1 release ?

kushagraThapar:
@milismsft , looks like we lost the conversation on this one ?
Can you please update the timeline for this ?

PR #9607, opened 2020-03-26
Cosmos CFP: add API to check if an instance is running

milismsft:
ChangeFeedProcessor: add API to check if an instance is started and currently running.


milismsft:
This change should address https://github.com/Azure/azure-sdk-for-java/issues/5361

PR #8031, opened 2020-02-07
Updates logger from using Log4j 1.x to 2.x

conniey:
* Updates the test logger from using log4j 1.x to log4j 2.x
  * Log4j 1.x went out of support in 2015 and the last release was in 2012.
* Updates and consolidates logger configuration/use in benchmark and examples project. 
* For benchmarks, there are some configuration values that are passed in at runtime. I assumed they are system arguments that are passed: `-Dazure.cosmos.logger.directory=C:\logs -Dazure.cosmos.hostname=test.cosmosdb.com`.  If they are environment variables, I can update the lookup to use `env` instead of `sys`.

PR #9613, opened 2020-03-26
Added module-info.java and removed javax.annotation from RntbdMetrics

kushagraThapar:
* Added module-info.java and removed javax.annotation from RntbdMetrics

### DIFF
--- a/eng/versioning/external_dependencies.txt
+++ b/eng/versioning/external_dependencies.txt
@@ -181,7 +181,7 @@ cosmos_io.netty:netty-handler;4.1.45.Final
 cosmos_io.netty:netty-handler-proxy;4.1.45.Final
 cosmos_io.netty:netty-tcnative;2.0.27.Final
 cosmos_io.netty:netty-transport-native-epoll;4.1.45.Final
-cosmos_io.projectreactor:reactor-core;3.3.3.RELEASE
+cosmos_log4j:log4j;1.2.17
 cosmos_org.mockito:mockito-core;1.10.19
 
 # sdk\eventgrid\microsoft-azure-eventgrid\pom.xml
--- a/sdk/cosmos/azure-cosmos/src/main/java/module-info.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/module-info.java
@@ -3,14 +3,8 @@
 
 module com.azure.cosmos {
 
-    requires transitive reactor.core;
-    requires transitive org.reactivestreams;
+    requires transitive com.azure.core;
 
-    requires transitive com.fasterxml.jackson.annotation;
-    requires transitive com.fasterxml.jackson.core;
-    requires transitive com.fasterxml.jackson.databind;
-
-    requires com.azure.core;
     requires com.fasterxml.jackson.datatype.jsr310;
     requires io.netty.transport;
     requires io.netty.handler;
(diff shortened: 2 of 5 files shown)


sdk/cosmos/azure-cosmos/src/main/java/module-info.java
```diff
@@ -0,0 +1,52 @@
...
+module com.azure.cosmos {
+
+    requires transitive reactor.core;
+    requires transitive org.reactivestreams;
+
+    requires transitive com.fasterxml.jackson.annotation;
+    requires transitive com.fasterxml.jackson.core;
+    requires transitive com.fasterxml.jackson.databind;
+
+    requires com.azure.core;
```
srnagar: This must be transitive

sdk/cosmos/azure-cosmos/src/main/java/module-info.java
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT License.
+
+module com.azure.cosmos {
+
+    requires transitive reactor.core;
+    requires transitive org.reactivestreams;
+
+    requires transitive com.fasterxml.jackson.annotation;
+    requires transitive com.fasterxml.jackson.core;
+    requires transitive com.fasterxml.jackson.databind;
```
srnagar: Don't need any of these as you are requiring com.azure.core which transitively requires these.

sdk/cosmos/azure-cosmos/src/main/java/module-info.java
kushagraThapar: Done

sdk/cosmos/azure-cosmos/src/main/java/module-info.java
kushagraThapar: Done.

eng/versioning/external_dependencies.txt
```diff
@@ -181,7 +181,7 @@ cosmos_io.netty:netty-handler;4.1.45.Final
 cosmos_io.netty:netty-handler-proxy;4.1.45.Final
 cosmos_io.netty:netty-tcnative;2.0.27.Final
 cosmos_io.netty:netty-transport-native-epoll;4.1.45.Final
-cosmos_io.projectreactor:reactor-core;3.3.3.RELEASE
+cosmos_log4j:log4j;1.2.17
```
JimSuplizio: @kushagraThapar We shouldn't be using this version of log4j. This particular version went out of support in 2015 and @conniey had updated these dependencies in early Feb as part of this [PR](https://github.com/Azure/azure-sdk-for-java/pull/8031). I would like to see this cosmos specific version updated to use the 2.x common versions above.

kushagraThapar: @JimSuplizio  - the only reason we had to add log4j is because of this dependency. 
```
<dependency>
      <groupId>com.fasterxml.uuid</groupId>
      <artifactId>java-uuid-generator</artifactId>
      <version>3.2.0</version> <!-- {x-version-update;com.fasterxml.uuid:java-uuid-generator;external_dependency} -->
    </dependency>
```
Cosmos doesn't use this dependency of log4j, but uuid does, and it was complaining for `log4j module not found, required by com.fasterxml.uuid` 
So that is the reason we had to add this.

conniey: Is it possible to move to version 4? It uses slf4j, which would align with our underlying logging framework.

https://github.com/cowtowncoder/java-uuid-generator/blob/master/README.md#dependencies

kushagraThapar: @conniey yes that should be possible, I will give it a shot.

kushagraThapar: Updated the version here: https://github.com/Azure/azure-sdk-for-java/pull/9757

PR #9620, opened 2020-03-27
Added helper function for IndexBatchException

sima-zhu:


sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/IndexAction.java
```diff
@@ -29,7 +29,7 @@
     private Map<String, Object> properties;
 
     @JsonAnyGetter
-    private Map<String, Object> getParamMap() {
+    public Map<String, Object> getParamMap() {
```
alzimmermsft: Did you mean to make this public again?

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/setup/AzureSearchResources.java
```diff
@@ -121,7 +122,7 @@ public void createResourceGroup() {
         String resourceGroupName = Configuration.getGlobalConfiguration().get(AZURE_RESOURCEGROUP_NAME);
```
alzimmermsft: Since there is a default hardcoded test resource group you can get rid of the null or empty check and just use the following here.

```java
String resourcegroupname = Configuration.getGlobalConfiguration().get(AZURE_RESOURCEGROUP_NAME, TEST_RESOURCE_GROUP);
```

This is effectively a get or default.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/IndexBatchException.java
```diff
@@ -0,0 +1,122 @@
...
+
+            case 422:
+            case 409:
+            case 503:
+                return true;    // The above cases might succeed on a subsequent retry.
+
+            default:
+                // If this happens, it's a bug. Safest to assume no retry.
+                return false;
+        }
```
alzimmermsft: This could be simplified a lot.

```java
return statusCode == 422 || statusCode == 409 || statusCode == 503;
```

Definitely add a comment for why those status codes are considered for retry but others aren't, for example 503 is allow since the server failed to process the request properly, etc

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/models/IndexBatchExceptionTests.java
```diff
@@ -0,0 +1,216 @@
...
+
+import static org.mockito.Mockito.mock;
+
+public class IndexBatchExceptionTests {
+    private static final String KEY_FIELD_NAME = "key";
+    private IndexDocumentsResult result;
+
+    @BeforeEach
+    public void setup() {
+        result = mock(IndexDocumentsResult.class);
```
alzimmermsft: You need to initialize mocking.

`MockitoAnnotations.initMocks(this);`

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/models/IndexBatchExceptionTests.java
```diff
@@ -0,0 +1,216 @@
...
+    }
```
alzimmermsft: You also need to cleanup mocking.

```java
@AfterEach
    public void clearMocks() {
        Mockito.framework().clearInlineMocks();
    }
```

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -15,7 +15,7 @@ parameters:
       OSVmImage: 'macOS-10.15'
       JavaVersion: '1.8'
       DisplayName: 'macOS - Java 8'
-    Windows - Java 8:
+    Windows - Java 8:IndexingSyncTests
```
alzimmermsft: Typo?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/IndexAction.java
sima-zhu: For indexAction, we can define either a map (properties) or an object(document).
If we made this private, we cannot access properties only document.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/IndexBatchException.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Finds all index actions in the given batch that failed and need to be retried, and returns them in a new batch.
+     *
+     * @param originalBatch The batch that partially failed indexing.
+     * @param keyFieldName The name of the key field from the index schema.
+     * @return A new batch containing all the actions from the given batch that failed and should be retried.
+     */
+    public IndexBatchBase<SearchDocument> findFailedActionsToRetry(IndexBatchBase<SearchDocument> originalBatch,
+        String keyFieldName) {
```
alzimmermsft: I'm a little confused on why a `keyFieldName` is needed here, I don't see a key field being used in `IndexBatchBase` or in `IndexAction`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/IndexAction.java
alzimmermsft: Could this be package private instead?

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
sima-zhu: Removed.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/IndexBatchException.java
```diff
@@ -0,0 +1,120 @@
...
+        return findFailedActionsToRetry(originalBatch, searchDocument -> searchDocument.get(keyFieldName).toString());
+    }
+
+    /**
+     * Finds all index actions in the given batch that failed and need to be retried, and returns them in a new batch.
+     *
+     * @param originBatch The batch that partially failed indexing.
+     * @param keySelector A lambda that retrieves a key value from a given document of type T.
+     * @param <T> The CLR type that maps to the index schema. Instances of this type can be stored as documents
```
alzimmermsft: By any chance was this copied from .NET? Java doesn't have a CLR 😃

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/IndexBatchException.java
sima-zhu: This is copied from .net. Need to double check

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/IndexBatchException.java
sima-zhu: Yes, it is copied from .net. Will change.

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/models/IndexBatchExceptionTests.java
sima-zhu: Is this needed? As I do not initialize the mock through annotation.

```
Initializes objects annotated with Mockito annotations for given testClass: @Mock, @Spy, @Captor, @InjectMocks
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/IndexAction.java
sima-zhu: It is currently not used by any API. However, based on document, we allowed to define combined operation for batch. 
Even we can do this for now, it might need to be public if we are ready to do "delete|upload" batch operation.

sima-zhu: https://docs.microsoft.com/en-us/rest/api/searchservice/addupdate-or-delete-documents#response

sima-zhu: We can hide behind for now as we do not have batch operation on different action type. Can make it public when we provide the function.

sima-zhu: It cannot the private as we the index batch operation is supported.
`public Mono<IndexDocumentsResult> indexDocuments(IndexDocumentsBatch<?> batch)

sima-zhu: What we can do is to rename it to something make more sense

PR #9623, opened 2020-03-27
Refactor getter and setter methods for resource classes

xseeseesee:
[Feature request from mgmt for Java](https://github.com/Azure/azure-libraries-for-java/issues/1130)
Considering consistency of code migration from track 1 to track 2, we wanted to check if it is acceptable to refactor getter and setter methods for resource classes in azure-core-management. Thanks.

### DIFF
--- a/sdk/core/azure-core-management/src/main/java/com/azure/core/management/Resource.java
+++ b/sdk/core/azure-core-management/src/main/java/com/azure/core/management/Resource.java
@@ -47,7 +47,7 @@ public class Resource extends ProxyResource {
      *
      * @return the tags value
      */
-    public Map<String, String> getTags() {
+    public Map<String, String> tags() {
         return this.tags;
     }
 


sdk/core/azure-core-management/src/main/java/com/azure/core/management/Resource.java
```diff
@@ -47,7 +47,7 @@ public Resource setLocation(String location) {
      *
      * @return the tags value
      */
-    public Map<String, String> getTags() {
+    public Map<String, String> tags() {
```
weidongxu-microsoft: in track1, it is `getTags`

xccc-msft: Fixed according to track 1 style. Though it's strange they put getTags/withTags together.

yungezz: any context on why track1 is getTags(), but not tags? If no specific reason, this is good time for us to fix it for aligment.

weidongxu-microsoft:
@JonathanGiles 

For backward-compatibility, Fluent premium management SDK would prefer to keep method name consistent with track1. As part of it, we'd like e.g. still use `withId` and `id` for POJO.

User could have hundreds of code already using the track1 naming, it is not really convenient for them to manually change them when upgrade, and automatic approach is not very reliable yet.

Moreover, Fluent premium management SDK is not really much about Java language. IMHO, it is more like an internal DSL compiled in Java.

E.g. here is code snippet to create a load balancer https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/network/mgmt/src/test/java/com/azure/management/network/LoadBalancerTests.java#L106
(it works better with indent, e.g. from `defineLoadBalancingRule` to `attach` would configure part of the load balancer, i.e., a load balancing rule)
It is not really making much sense to change `withRegion` to `setRegion` and `withProtocol` to `setProtocol` in the whole fluent flow.

cc PM @nickzhums @selvasingh

JonathanGiles:
As far as I'm aware the design guidelines I have created are only for client libraries and are not 100% applicable to management libraries. Of course, wherever possible I would much prefer to have them be consistent with each other.

weidongxu-microsoft:
@JonathanGiles Thank you for the clarification. Do you aware any other person that we need to keep in loop? Any one on Azure SDK?

Just trying to avoid any surprise on Azure SDK team. Previously we mostly only have internal discussion with PM and engineer managers.

And if we indeed have good idea on how we could change and improve, we can do it before formal release (or even take our chance to have private preview with customer and get feedback).

JonathanGiles:
+ @Petermarcu In case he wants to offer a differing opinion on the extent that management libraries should follow the design guidelines set out for new client libraries.

Petermarcu:
My general belief here is that getting to Track 2 is about following the guidelines. I believe they should be followed unless there is a strong reason not to.

yungezz:
thanks @Petermarcu . The consideration behind the change is to follow fluent style for fluent java sdks. This is actually one of highlight of fluent. User could use fluent client like example @weidongxu-microsoft shared: https://github.com/Azure/azure-libraries-for-java/blob/master/azure-mgmt-network/src/test/java/com/microsoft/azure/management/network/LoadBalancerTests.java#L109. If change it to getter/setter, whole fluent interface need to be changed. 
Let us know if there's any question/concern, thanks.

Petermarcu:
Does this mean that the module is really azure-management-core-fluent? Meaning, this would not be the core you would want to use if you were not doing a fluent pattern?

Petermarcu:
Another idea, perhaps there could be a contribution to the guidelines that say for the management plane, a fluent pattern is preferred. Then you can write the amendments to the guidelines and make them official so everyone is clear where the management plane should be different and where it should be the same?

yungezz:
thanks @Petermarcu for the good suggestion. Yes we should contribute to the guideline to clearly describe what's specific to mgmt/fluent and why. We want to use azure-management-core for mgmt sdks. We're working on evaluate what's best for user using service sdks not covered in fluent premium.

PR #9648, opened 2020-03-27
Fixing NullPointerExceptions in Event Hubs live Tests

conniey:
Live tests are failing because of NPE. The underlying cause of the NPEs is that the connection is already disposed of, and we try to dispose it again.

This happens in our tests because we use the EventHubClient directly to create producers and consumers. Consequently, it incorrectly keeps track of the children created from it.


sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/SetPrefetchCountTest.java
```diff
@@ -20,6 +20,7 @@
 /**
  * Verifies we can use various prefetch options with {@link EventHubConsumerAsyncClient}.
  */
+@Disabled("Set prefetch tests do not work because they try to send very large number of events at once.")
```
srnagar: Do we plan on fixing this? Or should this test be deleted as this will not be supported at all?
If we plan to fix, then please also file an issue and link it here.

PR #9649, opened 2020-03-27
[Storage] Jumbo blob support

kasobol-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
```diff
@@ -88,7 +88,12 @@
      * value will be used.
      */
     public static final int BLOB_DEFAULT_HTBB_UPLOAD_BLOCK_SIZE = 8 * Constants.MB;
+    /**
+     * @deprecated Use {@link #BLOB_MAX_UPLOAD_BLOCK_SIZE_LONG}.
+     */
+    @Deprecated
     static final int BLOB_MAX_UPLOAD_BLOCK_SIZE = 100 * Constants.MB;
```
alzimmermsft, line 95: This is scoped to `package private` shouldn't need deprecation.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
```diff
@@ -26,15 +26,18 @@
...
-                : other.getBlockSize(),
-            other.getNumBuffers() == null ? Integer.valueOf(BlobAsyncClient.BLOB_DEFAULT_NUMBER_OF_BUFFERS)
-                : other.getNumBuffers(),
-            other.getProgressReceiver(),
-            other.getMaxSingleUploadSize() == null ? Integer.valueOf(BlockBlobAsyncClient.MAX_UPLOAD_BLOB_BYTES)
-                : other.getMaxSingleUploadSize());
+        other = other == null ? new ParallelTransferOptions() : other;
+        return new ParallelTransferOptions()
+            .setBlockSize(other.getBlockSizeLong() == null
+                ? Long.valueOf(BlobAsyncClient.BLOB_DEFAULT_UPLOAD_BLOCK_SIZE)
```
alzimmermsft, line 32: Shouldn't need `Long.valueOf`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
```diff
@@ -26,15 +26,18 @@
...
+                : other.getBlockSizeLong())
+            .setNumBuffers(other.getNumBuffers() == null
+                ? Integer.valueOf(BlobAsyncClient.BLOB_DEFAULT_NUMBER_OF_BUFFERS)
```
alzimmermsft, line 35: Shouldn't need `Integer.valueOf`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
```diff
@@ -26,15 +26,18 @@
...
+                : other.getNumBuffers())
+            .setProgressReceiver(other.getProgressReceiver())
+            .setMaxSingleUploadSize(other.getMaxSingleUploadSizeLong() == null
+                ? Long.valueOf(BlockBlobAsyncClient.MAX_UPLOAD_BLOB_BYTES_LONG)
```
alzimmermsft, line 39: Shouldn't need `Long.valueOf`

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/UploadBufferPool.java
```diff
@@ -37,23 +40,23 @@
...
-    private final BlockingQueue<ByteBuffer> buffers;
+    private final BlockingQueue<BufferAggregator> buffers;
 
     private final int maxBuffs;
 
     // The number of buffs we have allocated. We can query the queue for how many are available.
     private int numBuffs;
 
-    private final int buffSize;
+    private long buffSize;
```
alzimmermsft, line 50: Any reason this is no longer final?

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileAsyncClient.java
```diff
@@ -76,8 +77,14 @@
 
     /**
      * Indicates the maximum number of bytes that can be sent in a call to upload.
+     * @deprecated Use {@link #MAX_APPEND_FILE_BYTES_LONG}
      */
+    @Deprecated
     static final int MAX_APPEND_FILE_BYTES = 100 * Constants.MB;
```
alzimmermsft, line 83: Package private, shouldn't require a deprecation

kasobol-msft:
> Changes to the chunking logic look pretty good to me, I'll leave the API changes for `int` to `long` for more review. Should we cherry pick the chunking changes into master after this PR has been vetted?

That's a good idea.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobAsyncClient.java
kasobol-msft: done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
kasobol-msft: I did play with `valueOf` in ModelHelpers. After removing I got quite legit bug report.
image
So I ended up reworking this code little bit.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
kasobol-msft: reworked (see other comment)

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ModelHelper.java
kasobol-msft: reworked (see other comment)

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/UploadBufferPool.java
kasobol-msft: my mistake. fixed.

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileAsyncClient.java
kasobol-msft: done.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
```diff
@@ -71,7 +71,7 @@
 public class EncryptedBlobAsyncClient extends BlobAsyncClient {
 
     static final int BLOB_DEFAULT_UPLOAD_BLOCK_SIZE = 4 * Constants.MB;
-    private static final int BLOB_MAX_UPLOAD_BLOCK_SIZE = 100 * Constants.MB;
+    private static final long BLOB_MAX_UPLOAD_BLOCK_SIZE = 4000L * Constants.MB;
```
gapra-msft, line 74: I think there's a constant for GB that we could use here

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
```diff
@@ -313,7 +313,7 @@
      * @param tier {@link AccessTier} for the destination blob.
      * @param requestConditions {@link BlobRequestConditions}
      * @return An empty response
-     * @throws IllegalArgumentException If {@code blockSize} is less than 0 or greater than 100MB
+     * @throws IllegalArgumentException If {@code blockSize} is less than 0 or greater than 4000MB
```
gapra-msft, line 316: Can we make that 4GB

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
kasobol-msft: Problem is. Jumbo block is up to exactly 4000MB. 4GB is more than that.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
kasobol-msft: Problem is. Jumbo block is up to exactly 4000MB. 4GB is more than that.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobAsyncClient.java
```diff
@@ -58,14 +58,28 @@
 
     /**
      * Indicates the maximum number of bytes that can be sent in a call to upload.
+     * @deprecated Use {@link #MAX_STAGE_BLOCK_BYTES_LONG}
      */
+    @Deprecated
     public static final int MAX_UPLOAD_BLOB_BYTES = 256 * Constants.MB;
```
jaschrep-msft, line 64: Do we want to up these values to max int? I know they're deprecated but there's still a lot of room between 256MB and max int that customers can use if they want and their clamping logic would be misinformed from this.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/ParallelTransferOptions.java
```diff
@@ -60,35 +68,54 @@ public ParallelTransferOptions(Integer blockSize, Integer numBuffers, ProgressRe
...
+     * For upload, The block size is the size of each block that will be staged. This value also determines the number
+     * of requests that need to be made. If block size is large, upload will make fewer network calls, but each
+     * individual call will send more data and will therefore take longer. This parameter also determines the size
+     * that each buffer uses when buffering is required and consequently amount of memory consumed by such methods may
+     * be up to blockSize * numBuffers.
+     *
+     * @param blockSize The block size.
+     * @return The ParallelTransferOptions object itself.
+     */
+    public ParallelTransferOptions setBlockSize(Long blockSize) {
```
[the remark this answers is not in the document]
kasobol-msft: In such case I'd rather add non-null validation - but since it's optional customer should be able to "unset" it - end of day it just became mutable data structure. If I were customer where this thing is result of some computation (weird but possible) then I'd be quite upset about the need of if/else-ing. Besides if some of them are actually passing Integer around that could be null for some reason then autoboxing will blow up on nulls if they don't else/if correctly.

Other thing. Usually in classes like that I used to see symmetry between getter and setter - they usually speak same type. It'd be quite non-idiomatic in Java if we do otherwise.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/Block.java
```diff
@@ -25,7 +25,13 @@
      * The block size in bytes.
      */
     @JsonProperty(value = "Size", required = true)
-    private int size;
+    private long sizeLong;
+
+    /*
+     * The size property.
+     */
```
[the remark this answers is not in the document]
kasobol-msft: That's not harmful remnant of swagger generation. See `swagger/readme.md` change in this PR. In order to transition from int->long I had to generate two Java properties for same field in such a way that one is actually used by Jackson parser to deserialize incoming xml and other to assure backwards compatibility - then remove data binding annotations and body so that extra property doesn't take part in deserialization process.

This comment is remnant of this process. I didn't manage to invent right regex that could get rid of this yet. But I can try harder.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/Block.java
```diff
@@ -48,22 +54,44 @@ public Block setName(String name) {
...
      * @param size the size value to set.
+     * @deprecated Use {@link #setSizeLong(long)}
      * @return the Block object itself.
      */
-    public Block setSize(int size) {
-        this.size = size;
+    @Deprecated public Block setSize(int size) {
+        this.sizeLong = size;
         return this;
     }
```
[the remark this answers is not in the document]
kasobol-msft: I believe Jackson parser needs that to deserialize XML.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
```diff
@@ -159,7 +159,7 @@ void commit() {
         private BlockBlobOutputStream(final BlobAsyncClient client,
             final ParallelTransferOptions parallelTransferOptions, final BlobHttpHeaders headers,
             final Map<String, String> metadata, final AccessTier tier, final BlobRequestConditions requestConditions) {
-            super(BlockBlobClient.MAX_STAGE_BLOCK_BYTES);
+            super(Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
kasobol-msft: It ends up being used here https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageOutputStream.java#L54 .

As far as I can tell from reading inheritance tree of StorageOutputStream - this can't be changed to long because of conversion to Flux of ByteBuffers - these are constrained by int capacity - these conversions happen in subtypes. 

@gapra-msft @rickle-msft  do you guys have any context behind previously chosen value?

I think this could be set to `Math.min(Integer.MAX_VALUE, parallelTransferOptions.getBlockSizeLong())` so that it would result in ByteBuffers already trimmed nice for staging blocks (for small blocks) and largest possible ByteBuffers for jumbo blocks.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobAsyncClient.java
kasobol-msft: That's a good idea. However I'd rather do that only for blocks in this PR. Bumping single upload threshold deserve own PR and testing.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -813,4 +833,37 @@ class APISpec extends Specification {
...
+
+        @Override
+        Mono<HttpResponse> process(HttpPipelineCallContext httpPipelineCallContext, HttpPipelineNextPolicy httpPipelineNextPolicy) {
+            def request = httpPipelineCallContext.httpRequest
+            def key = request.url.toString()
+            if (failureTracker.get(key, false)) {
+                return httpPipelineNextPolicy.process()
+            } else {
+                failureTracker.put(key, true)
+                return request.getBody().flatMap(new Function<ByteBuffer, Publisher<ByteBuffer>>() {
```
rickle-msft, line 849: Can you use closures for these inline functions?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
```diff
@@ -813,4 +833,37 @@ class APISpec extends Specification {
...
+                    @Override
+                    Publisher<ByteBuffer> apply(ByteBuffer byteBuffer) {
+                        byteBuffer.get()
```
rickle-msft, line 852: Is this just to alter the data to inject the failure? In general, can you leave some notes about the intended behavior of this type?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
```diff
@@ -0,0 +1,244 @@
...
+        def blockList = client.getBlockBlobClient().listBlocks(BlockListType.COMMITTED)
+
+        then:
+        blockList.committedBlocks.size() == 1
+        blockList.committedBlocks.get(0).getSizeLong() == maxBlockSize
+    }
+
+    @Requires({ liveMode() })
+    def "Stage Large Blob"() {
+        given:
```
rickle-msft, line 74: Is the only difference between this test and the last one that this one drops the payload and counts it in the pipeline? Can you leave a note about that?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
```diff
@@ -0,0 +1,244 @@
...
+                    def bytesReceived = request.getBody().reduce(0L, new BiFunction<Long, ByteBuffer, Long>() {
+                        @Override
+                        Long apply(Long a, ByteBuffer byteBuffer) {
+                            return a + byteBuffer.remaining()
+                        }
+                    })
+                    putBlockPayloadSizes.add(bytesReceived)
+                }
+                def currentCount = count.incrementAndGet()
+                println(currentCount) // TODO remove this
```
rickle-msft: Remove

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/UploadUtils.java
```diff
@@ -39,72 +37,25 @@
...
-         * Put Block List as the upload mechanism otherwise use Put Blob.
-         */
+        PayloadSizeGate gate = new PayloadSizeGate(parallelTransferOptions.getMaxSingleUploadSizeLong());
+
         return data
             .filter(ByteBuffer::hasRemaining)
-            .windowUntil(buffer -> {
-                if (bufferedDataSize[0] > parallelTransferOptions.getMaxSingleUploadSize()) {
-                    return false;
+            .concatMap(gate::write)
```
rickle-msft, line 45: Can you add some comments here describing the behavior:
The gate will only start emitting data after the threshold has been breached once it has been flushed. Checking the state of the threshold on receiving the first buffer means we can infer what upload operation we should do.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/PayloadSizeGate.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.Queue;
+
+/**
+ * This class provides ability to measure if incoming Flux of ByteBuffers is larger than a threshold.
+ * This answers question if volume of data in bytes is larger than threshold.
+ *
+ * The {@link #write(ByteBuffer)} operation buffers incoming ByteBuffers until threshold is crossed.
+ * After that it's pass-through as fact that data volume exceeds threshold is already determined.
+ *
+ * It is incumbent upon the caller to return the buffers after measurement is completed. It is also the caller's
```
rickle-msft, line 19: I think you can delete this line as there's no concept of returning buffers here.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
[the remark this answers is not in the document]
jaschrep-msft: It is actually 4000MB and not 4096MB.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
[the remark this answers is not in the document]
kasobol-msft: That has been already asked. Problem is. Jumbo block is up to exactly 4000MB. 4GB is more than that.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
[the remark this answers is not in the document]
kasobol-msft: Jumbo block is up to exactly 4000MB.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/Block.java
```diff
@@ -25,7 +25,13 @@
      * The block size in bytes.
      */
     @JsonProperty(value = "Size", required = true)
-    private int size;
+    private long sizeLong;
+
+    /*
+     * The size property.
```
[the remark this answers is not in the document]
kasobol-msft: Yes. I had to generate two bend generator to produce two properties out of one and then string.replace some parts of the code. See swagger/readme.md change in this PR.
I didn't find good regex to get rid of this thing yet.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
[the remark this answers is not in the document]
kasobol-msft: I left long comment on @jaschrep-msft  comment above and tagged you as there's a question. Could you please reply there?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
kasobol-msft: This is to make sure byteBuffers were "touched". I'll leave a comment.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/Block.java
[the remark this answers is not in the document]
kasobol-msft: fixed

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/Block.java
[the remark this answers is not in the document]
kasobol-msft: removed..

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
kasobol-msft: removed.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
gapra-msft: Oh I see

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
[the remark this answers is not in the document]
gapra-msft: Actually, specifically for BlockBlobOutputStream, after my changes to make it write via buffered upload, this this.writeThreshold isn't used. If you look at BlockBlobOutputStream, writeInternal is overridden and dispatchWrite is never called (it just returns an empty mono). Could you just add a note about that if you don't mind?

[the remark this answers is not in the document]
gapra-msft: Basically all the write Threshold logic is handled within buffered upload

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlockBlobAsyncClient.java
kasobol-msft: Sadly. Looks like it would be considered as breaking change.
`[ERROR] Failed to execute goal org.revapi:revapi-maven-plugin:0.11.2:check (default) on project azure-storage-blob: The following API problems caused the build to fail:
[ERROR] java.field.constantValueChanged: field com.azure.storage.blob.specialized.BlockBlobAsyncClient.MAX_STAGE_BLOCK_BYTES: Constant field changed value from '104857600' to '2147483647'. (breaks semantic versioning)
[ERROR] java.field.constantValueChanged: field com.azure.storage.blob.specialized.BlockBlobClient.MAX_STAGE_BLOCK_BYTES: Constant field changed value from '104857600' to '2147483647'. (breaks semantic versioning)
[ERROR]
[ERROR] Consult the plugin output above for suggestions on how to ignore the found problems.`

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/PayloadSizeGate.java
kasobol-msft: deleted

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/UploadUtils.java
kasobol-msft: added comments

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
kasobol-msft: commented

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
kasobol-msft: done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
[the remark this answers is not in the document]
kasobol-msft: added comment after offline discussion. This value has no real meaning for blockbloboutputstream.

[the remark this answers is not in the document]
kasobol-msft: added comment per our discussion.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/APISpec.groovy
kasobol-msft: done

PR #9650, opened 2020-03-27
Add Tests to Azure Core HTTP Response Decoder Classes

alzimmermsft:
This PR adds tests to the HTTP response decoder classes in Azure Core. In addition to adding tests it also removes code paths that weren't reachable due to previous checks happening in either `RestProxy` or within `HttpResponseBodyDecoder`.

An issue with `UnixTime` serialization and deserialization was resolved by adding a `UnixTimeSerializer` which will convert the underlying `OffsetDateTime` into the appropriate epoch and a new constructor for `UnixTime` was added which accepts a `String` representation of the epoch and converts it into an `OffsetDateTime`.


pom.client.xml
```diff
@@ -960,8 +960,10 @@
                 --add-opens com.azure.core/com.azure.core.util.serializer=ALL-UNNAMED
                 --add-opens com.azure.core/com.azure.core.http=ALL-UNNAMED
                 --add-opens com.azure.core/com.azure.core.http.policy=ALL-UNNAMED
+                --add-opens com.azure.core/com.azure.core.http.rest=ALL-UNNAMED
                 --add-opens com.azure.core/com.azure.core.util=ALL-UNNAMED
                 --add-opens com.azure.core/com.azure.core.implementation.http=ALL-UNNAMED
+                --add-opens com.azure.core/com.azure.core.implementation.serializer=ALL-UNNAMED
                 --add-opens com.azure.ai.textanalytics/com.azure.ai.textanalytics=ALL-UNNAMED
                 --add-opens com.azure.data.appconfiguration/com.azure.data.appconfiguration=ALL-UNNAMED
```
alzimmermsft: Not related to the PR but we have a lot of module openings within `pom.client.xml` for module specific tests, should these configurations be moved into their specific module?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/UnixTime.java
```diff
@@ -35,21 +35,27 @@ public UnixTime(long unixSeconds) {
         this.dateTime = OffsetDateTime.ofInstant(Instant.ofEpochSecond(unixSeconds), ZoneOffset.UTC);
     }
 
+    /**
+     * Creates a UnixTime object with the specified epoch.
+     *
+     * @param epoch The epoch value represented as a string.
+     */
+    public UnixTime(String epoch) {
```
alzimmermsft: Should this be replaced with a deserializer module for Jackson?

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -185,60 +162,48 @@ private static Object deserializeBody(String value, Type resultType, Type wireTy
...
      * @param wireType the {@code java.lang.reflect.Type} of entity in REST API response body
      * @return the {@code java.lang.reflect.Type} of REST API response body
      */
     private static Type constructWireResponseType(Type resultType, Type wireType) {
-        Objects.requireNonNull(resultType);
         Objects.requireNonNull(wireType);
-        //
-        Type wireResponseType = resultType;
 
-        if (resultType == byte[].class) {
```
alzimmermsft: Should never get a result type from here, `byte[]` is treated as a non-decodeable type in `isReturnTypeDecodable`.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -185,60 +162,48 @@ private static Object deserializeBody(String value, Type resultType, Type wireTy
...
             }
-        } else {
-            if (TypeUtil.isTypeOrSubTypeOf(resultType, List.class)) {
-                final Type resultElementType = TypeUtil.getTypeArgument(resultType);
-                final Type wireResponseElementType = constructWireResponseType(resultElementType, wireType);
-
-                wireResponseType = TypeUtil.createParameterizedType(
-                    (Class<?>) ((ParameterizedType) resultType).getRawType(), wireResponseElementType);
-            } else if (TypeUtil.isTypeOrSubTypeOf(resultType, Map.class) || TypeUtil.isTypeOrSubTypeOf(resultType,
-                Response.class)) {
```
alzimmermsft: Removed `Response.class` for a few reasons:

- I don't think `Response` would be a sensible `resultType` to get here since we unbox `Response` to its parameterized type in an earlier check, so this would need something like `Response<Response<>>`.
- The `typeArguments` retrieval would fail most of the time here since `Response<T>` only has a single type argument in the interface class, only `ResponseBase<H, T>` or a custom type could have enough type arguments to work here.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -274,90 +234,46 @@ private static Object deserializePage(String value, Type resultType, Type wireTy
      */
     private static Object convertToResultType(Object wireResponse, Type resultType, Type wireType) {
         Object result = wireResponse;
-
-        if (wireResponse != null) {
-            if (resultType == byte[].class) {
```
alzimmermsft: Removed `byte[]` for the same reason mentioned above.

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -274,90 +234,46 @@ private static Object deserializePage(String value, Type resultType, Type wireTy
...
-                if (wireType == Base64Url.class) {
-                    result = ((Base64Url) wireResponse).decodedBytes();
-                }
-            } else if (resultType == OffsetDateTime.class) {
-                if (wireType == DateTimeRfc1123.class) {
-                    result = ((DateTimeRfc1123) wireResponse).getDateTime();
```
alzimmermsft: Not sure why we get the and return the `OffsetDateTime` from `DateTimeRfc1123` and `UnixTime`

sdk/core/azure-core/src/main/java/com/azure/core/implementation/serializer/HttpResponseBodyDecoder.java
```diff
@@ -274,90 +234,46 @@ private static Object deserializePage(String value, Type resultType, Type wireTy
...
-                    for (Map.Entry<String, Object> wireResponseEntry : wireResponseEntries) {
-                        final Object wireResponseValue = wireResponseEntry.getValue();
-                        final Object resultValue = convertToResultType(wireResponseValue, resultValueType, wireType);
-                        if (wireResponseValue != resultValue) {
-                            wireResponseMap.put(wireResponseEntry.getKey(), resultValue);
-                        }
-                    }
-                    //
-                    result = wireResponseMap;
-                } else if (TypeUtil.isTypeOrSubTypeOf(resultType, PagedResponseBase.class)) {
```
alzimmermsft: Removed as this should be caught by the check for `Page.class` and should be deserialized through that code path.

pom.client.xml
srnagar: Initially, I put this here because there were only a handful of them and it was better to just have all these in one place. Now, the list has grown quite big and we may want to consider moving it to its own pom file. That said, we have to configure maven surefire plugin in each of the modules and when we have to make a change, we have to do it in all poms instead of one place in pom.client.xml.

alzimmermsft:
Letting this PR sit until Azure Core is released so this gets bake time to resolve any corner cases that could be introduced.

PR #9652, opened 2020-03-27
OpenTelemetry API version update

samvaity:
Update `io.opentelemetry:opentelemetry-api` to [latest version](https://mvnrepository.com/artifact/io.opentelemetry/opentelemetry-api) and make neccessary changes to azure-core-tracing-opentelemetry package.

### DIFF
--- a/sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
+++ b/sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.0.0-beta.4 (Unreleased)
 - Keyvault: add az namespace info attribute to all outgoing spans.
-
+- `io.opentelemetry` version update to `0.2.4` API changes.
 
 ## 1.0.0-beta.3 (2020-03-06)
 


eng/versioning/external_dependencies.txt
```diff
@@ -63,8 +63,8 @@ com.microsoft.azure:azure-storage;8.0.0
 com.microsoft.azure:msal4j;1.3.0
 com.sun.activation:jakarta.activation;1.2.1
 commons-collections:commons-collections;3.2.2
-io.opentelemetry:opentelemetry-api;0.2.0
-io.opentelemetry:opentelemetry-sdk;0.2.0
+io.opentelemetry:opentelemetry-api;0.2.4
```
samvaity: @JonathanGiles `io.opentelemetry` did release a 0.3.0, but we should not be updating to it, yet. As it introduced a `io.gRPC` dependency on the public exposed API's like [here](https://github.com/open-telemetry/opentelemetry-java/blob/690651b65a938c903c35d65aef3182999fcb1b19/api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java#L79) which they are planning to fix in later releases.
OpenTelemetry github issue to track this - https://github.com/open-telemetry/opentelemetry-java/issues/575

samvaity: Creating an issue on our end to track this - https://github.com/Azure/azure-sdk-for-java/issues/9695

PR #9654, opened 2020-03-27
Adding ScoringParameter class

sima-zhu:


sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Initializes a new instance of the ScoringParameter class with the given name and GeographyPoint value.
+     *
+     * @param name Name of the scoring parameter.
+     * @param value Value of the scoring parameter.
+     */
+    public ScoringParameter(String name, GeoPoint value) {
+        this(name, toLonLatStrings(value));
```
alzimmermsft: Could this or `toLonLatStrings` be changed so the `NullPointerException` thrown from either of these methods instead of the other constructor when `value` is null.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Represents a parameter value to be used in scoring functions (for example, referencePointParameter).
+ */
+public final class ScoringParameter {
+    private final ClientLogger logger = new ClientLogger(ScoringParameter.class);
+    private final String name;
+    private final List<String> values;
+
+    private static final String SEPARATOR = "-";
```
alzimmermsft: nit: Could this be renamed dash to more closely follow the `-` character's name.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
```diff
@@ -0,0 +1,110 @@
...
+        return name;
+    }
+
+    /**
+     * Gets the values of the scoring parameter.
+     *
+     * @return The values of scoring parameter.
+     */
+    public List<String> getValues() {
+        return values;
```
alzimmermsft: Should make sure the constructor and this method follow the same pattern around ensuring the internal values list isn't mutable. Right now the constructor will clone the list, effectively a deep clone based on how Strings work, ensuring it is immutable if the original list changes but this will return a reference to the internal list that could be mutated.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Represents a parameter value to be used in scoring functions (for example, referencePointParameter).
+ */
+public final class ScoringParameter {
```
alzimmermsft: Could you add some tests which target this class specifically without call to the service.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
```diff
@@ -0,0 +1,110 @@
...
+     * @return The values of scoring parameter.
+     */
+    public List<String> getValues() {
+        return values;
+    }
+
+    private static List<String> toLonLatStrings(GeoPoint point) {
+        if (point == null) {
+            return null;
+        }
```
alzimmermsft: Should this just be `Objects.requireNonNull`, if `null` is returned here it will eventually lead to a `NullPointerException` in another, less obvious, location.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
```diff
@@ -0,0 +1,110 @@
...
+    @JsonValue
+    public String toString() {
+        String flattenValue = values.stream().filter(value -> !CoreUtils.isNullOrEmpty(value))
+            .map(this::escapeValue).collect(Collectors.joining(COMMA));
+        if (CoreUtils.isNullOrEmpty(flattenValue)) {
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("There must be at least one valid value for scoring parameter values."));
+        }
+        return name + SEPARATOR + values.stream().filter(value -> !CoreUtils.isNullOrEmpty(value))
+                .map(this::escapeValue).collect(Collectors.joining(COMMA));
```
alzimmermsft: Why do we need to repeat creating `flattenValue`? This could be simplified to the following:

```java
return name + SEPARATOR + flattenValue;
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    private String escapeValue(String value) {
```
alzimmermsft: This could be made `static`

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
sima-zhu: Renamed to DASH

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
sima-zhu: I have added several test cases in SearchSyncTests.
Will add more for  unit testing ScoringParameter only

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
sima-zhu: I planned to remove the null checking here.
Since there are two constructors in this class. User cannot put null to initialize the constructor, it will throw compile time error.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
sima-zhu: Put the null checking here.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
sima-zhu: Good point. Will deep clone the list.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
sima-zhu: switch to null pointer exception check.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ScoringParameter.java
sima-zhu: Put null checking in method at the first place.

PR #9660, opened 2020-03-30
Jetty usage rework

mitchdenny:
As part of our overall POM restructuring work I need to tweak the way that Jetty is invoked in our build system. There is only a small number of modules that actually rely on Jetty so my plan is to move the Jetty configuration into those modules and remove it from the ```azure-sdk-parent``` POM file and also inline the configuration for those specific modules where it is required.

This will hopefully make our build system faster but also will just generally make our Maven usage cleaner.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-data.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-data.yml
@@ -201,19 +201,6 @@ jobs:
           Artifacts: ${{parameters.Artifacts}}
           AdditionalModules: ${{parameters.AdditionalModules}}
 
-      - task: Maven@3
-        displayName: 'Start Jetty'
-        condition: succeeded()
-        inputs:
-          mavenPomFile: pom.data.xml
-          options: '$(DefaultOptions)'
-          mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
-          javaHomeOption: 'JDKVersion'
-          jdkVersionOption: $(JavaVersion)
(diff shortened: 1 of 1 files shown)


eng/versioning/version_client.txt
```diff
@@ -2,7 +2,7 @@
 # <groupId>:<artifactId>;dependency-version;current-version
 
 com.azure:azure-sdk-all;1.0.0;1.0.0
-com.azure:azure-sdk-parent;1.6.0;1.6.0
+com.azure:azure-sdk-parent;1.7.0;1.7.0
```
JimSuplizio: @JonathanGiles Does removal of the maven start jetty task from the parent/pom.xml actually require a pom version increase? It's unclear, to me, if it does however, if we do increase the version then I'm pretty sure we're going to need to publish updated versions during the next release.

mitchdenny: I'll do some validation around this before I merge this because I'd like to avoid it if we can. But I think because that POM is not being used as an aggregate and is not strictly listed as a dependency then Maven will try to pull the version from Maven Central rather than using the local copy.

But I need to test it in isolation because when I experienced the problems that made me bump this version there were lots of other issues going on.

mitchdenny: Looks like we got away with not having to bump the versions!

JimSuplizio: Thank you for confirming @mitchdenny! I'm very cautious about increasing the parent pom versions unless absolutely necessary.

PR #9665, opened 2020-03-30
support legacy PagedList for convenience of migration

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1102

Origin PagedList https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java

Origin test https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-runtime/src/test/java/com/microsoft/azure/PagedListTests.java


sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/utils/PagedList.java
```diff
@@ -0,0 +1,364 @@
...
+
+/**
+ * Defines a list response from a paging operation. The pages are
+ * lazy initialized when an instance of this class is iterated.
+ *
+ * @deprecated Use {@link PagedIterable} instead.
+ *
+ * @param <E> the element type.
+ */
+@Deprecated
```
weidongxu-microsoft: Annotate as deprecated.

sdk/resources/mgmt/src/test/java/com/azure/management/resources/fluentcore/utils/PagedListTests.java
```diff
@@ -0,0 +1,64 @@
...
+                } else {
+                    return Mono.just(new PagedResponseBase<>(null, 200, null,
+                        IntStream.range(nextInt, nextInt + lastPageSize).boxed().collect(Collectors.toList()), null,
+                        (Object) null));
+                }
+            }
+        );
+        final PagedIterable<Integer> pagedIterable = new PagedIterable<>(mockedPagedFlux);
+
+        list = new PagedList<>(pagedIterable);
```
weidongxu-microsoft: Expect user to create it via the standard PagedIterable.

weidongxu-microsoft:
Still adding test cases.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/utils/PagedList.java
yungezz: what's the deprecation mean? Should v2 still use PagedList or should not?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/utils/PagedList.java
```diff
@@ -0,0 +1,367 @@
...
+    protected boolean hasNextPage() {
+        return pagedResponseIterator.hasNext();
+    }
+
+    /**
+     * Loads a page from next page link.
+     * The exceptions are wrapped into Java Runtime exceptions.
+     */
+    protected void loadNextPage() {
+        this.items.addAll(pagedResponseIterator.next().getValue());
```
yungezz: seems cachedpage removed, any concern on performance etc when original cachedpage adding? Or we may revisit it later when there's complain

weidongxu-microsoft: v2 should not use PagedList. This class is provided since return type in v2 is PagedIterable, which is not compatibly with v1 PagedList.

weidongxu-microsoft: PagedIterable/PagedFlux is already doing what need to be done on paged response. This class is only a wrapper to provide interface compatible with v1.

weidongxu-microsoft: With this, user now have 2 choice when migrate.
1. Use PagedIterable, and refactor any code that handling it.
2. Use PagedList, only change the return type. Find time later to do all refactor needed in 1.

yungezz: thanks for clarification. Then in our sample or migration guide, pls use PagedIterable as right and long term way for paging operation.

PR #9666, opened 2020-03-30
update gulp config for mgmt

weidongxu-microsoft:


sdk/management/gulpfile.js
```diff
@@ -150,65 +147,6 @@ function codegen(project, cb) {
     return execa(cmd, [], { shell: true, stdio: "inherit" });
 };
 
-function findTag(stringContainsTag) {
```
yungezz: what're below codes used for?

weidongxu-microsoft: line 112 to 114.

used when autorest v3 does not support `--tag`

PR #9686, opened 2020-03-30
Adding support for auto create and auto delete test container

simplynaveen20:
This PR will create the test db and test container mentioned in the run parameters for all the workload if they doesn't exist.
Also It will delete the container at the end of the test.

### DIFF
--- a/sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsynReadWithMultipleClients.java
+++ b/sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsynReadWithMultipleClients.java
@@ -249,7 +249,7 @@ public class AsynReadWithMultipleClients<T> {
                     } catch (CosmosClientException e) {
                         if (e.getStatusCode() == HttpConstants.StatusCodes.NOTFOUND) {
                             cosmosAsyncContainer =
-                                cosmosAsyncDatabase.createContainer(this.configuration.getCollectionId(), Configuration.DEFAULT_PARTITION_KEY, this.configuration.getThroughput()).block().getContainer();
+                                cosmosAsyncDatabase.createContainer(this.configuration.getCollectionId(), Configuration.DEFAULT_PARTITION_KEY_PATH, this.configuration.getThroughput()).block().getContainer();
                             logger.info("Collection {} is created for this test on host {}", this.configuration.getCollectionId(), endpoint);
                             if(!databaseCreated) {
                                 collectionListToClear.add(cosmosAsyncContainer);
--- a/sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncBenchmark.java
+++ b/sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncBenchmark.java
@@ -87,7 +87,7 @@ abstract class AsyncBenchmark<T> {
         } catch (CosmosClientException e) {
             if (e.getStatusCode() == HttpConstants.StatusCodes.NOTFOUND) {
                 cosmosAsyncContainer =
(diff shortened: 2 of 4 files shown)


sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/Configuration.java
```diff
@@ -32,6 +32,7 @@
 
 class Configuration {
 
+    final static String PARTITION_KEY = "/pk";
```
moderakh: this is the default partition-key if the collection does not exist. But if the collection exist the partition-key path could be different and that's ok. right?

So could you please change the naming to `DEFAULT_PARTITION_KEY_PATH`

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncBenchmark.java
```diff
@@ -130,6 +134,8 @@ protected void init() {
     }
 
     void shutdown() {
+        cosmosAsyncContainer.delete().block();
```
moderakh: In our testing subscriptions we can't create a collection with 1M throughput upfront. The workaround we have is to create a collection with 100K throughput and then scale up. Now with this auto deleting collection it means we have to do this manual process everytime.

we don't want to always delete the collection. We should have a config option called `autoCollectionCreateEnabled` when present we should attempt to 
1. create collection if not exist 
2. delete the collection

when `autoCollectionCreateEnabled=false` we shouldn't attempt to create or delete the collection, and if the collection doesn't exist we should fail.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsynReadWithMultipleClients.java
```diff
@@ -227,33 +212,29 @@ private void createClients() {
...
-                            createDocumentObservables.add(obs);
-                        }
-                        docsToRead = Flux.merge(Flux.fromIterable(createDocumentObservables), 100).collectList().block();
-
-                        logger.info("Client have been initialized with data created for host {}", hostAndKey[0]);
-                    } else {
-                        logger.info("Client have been initialized with host {}", hostAndKey[0]);
+                    CosmosAsyncDatabase cosmosAsyncDatabase = asyncClient.createDatabaseIfNotExists(this.configuration.getDatabaseId()).block().getDatabase();
+                    CosmosAsyncContainer cosmosAsyncContainer =
+                        cosmosAsyncDatabase.createContainerIfNotExists(configuration.getCollectionId(), Configuration.PARTITION_KEY, configuration.getThroughput()).block().getContainer();
```
moderakh: if the application deletes it should create too.

when a user runs an application, the user expects to either 1.  application handles the collection creation/deletion 2. the user provides the collection and the app doesn't create nor deletes.

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/Configuration.java
simplynaveen20: done

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsynReadWithMultipleClients.java
simplynaveen20: Please check current design , it is more seem less in respect to customer expectation , code will only create resource if it is not present , and only delete newly created resource and not the existing

sdk/cosmos/azure-cosmos-benchmark/src/main/java/com/azure/cosmos/benchmark/AsyncBenchmark.java
simplynaveen20: Please check current design , it is more seem less in respect to customer expectation , code will only create resource if it is not present , and only delete newly created resource and not the existing

PR #9689, opened 2020-03-30
Nio directory stream

rickle-msft:


sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureDirectoryStream.java
```diff
@@ -0,0 +1,188 @@
...
+                    throw LoggingUtility.logError(logger, new DirectoryIteratorException(e));
+                }
+            }
+            return false;
+        }
+
+        @Override
+        public Path next() {
+            if (this.bufferedNext == null) {
+                if (!this.hasNext()) { // This will populate cachedNext in the process.
```
gapra-msft: should these comments actually be bufferedNext? I'm assuming you just renamed the variable and forgot to change the comments

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureDirectoryStream.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        @Override
+        public void remove() {
+            throw new UnsupportedOperationException();
+        }
+
+        private Path getNextListResult(BlobItem blobItem) {
+            // Strip the root if it is present so we can relativize the list result, which never has a root.
+            Path withoutRoot = this.path;
```
gapra-msft: is there a reason we can't just do this +next 3 lines once in the constructor for AzureDirectoryIterator?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureResource.java
```diff
@@ -34,10 +34,10 @@
  * It also serves as the interface to Storage clients. Any operation that needs to use a client should first build an
  * AzureResource using a path and then use the getter to access the client.
  */
-class AzureResource {
+final class AzureResource {
     private final ClientLogger logger = new ClientLogger(AzureResource.class);
 
-    static final String DIR_METADATA_MARKER = "is_hdi_folder";
+    static final String DIR_METADATA_MARKER = "hdi_isfolder";
```
gapra-msft: Can we possibly swap all instances of NIOs DIR_METADATA_MARKER for the one in common? I just merged a PR today that added the constant for a datalake feature

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureDirectoryStream.java
```diff
@@ -0,0 +1,188 @@
...
+import java.nio.file.DirectoryIteratorException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Path;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Set;
+
+/**
+ * An object for iterating the contents of a directory.
```
kasobol-msft: nit: an object is an instantiated class. I believe this comment describes class.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureDirectoryStream.java
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * This type does not validate that the given path is a directory; a listing call will be made against the given
+ * resource in either case, and if the resource is a file, it will simply return 0 results (assuming there are no
```
kasobol-msft: Files.newDirectoryStream throws `NotDirectoryException` `if the file could not otherwise be opened because it is not a directory` . Should we be consistent with that?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureDirectoryStream.java
```diff
@@ -0,0 +1,188 @@
...
+
+            /*
+            Search for a new element that passes the filter and buffer it when found. If no such element is found,
+            return false.
+             */
+            while (this.blobIterator.hasNext()) {
+                BlobItem nextBlob = this.blobIterator.next();
+                Path nextPath = getNextListResult(nextBlob);
+                try {
+                    if (this.filter.accept(nextPath) && passesDirectoryDuplicateFilter(nextPath, nextBlob)) {
```
kasobol-msft: nit. should this be called !isDuplicate ? or something like that

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
```diff
@@ -194,12 +193,18 @@ public SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> s
...
+     * Returns an {@link AzureDirectoryStream} for iterating over the contents of a directory.
+     *
      * {@inheritDoc}
      */
     @Override
     public DirectoryStream<Path> newDirectoryStream(Path path, DirectoryStream.Filter<? super Path> filter)
         throws IOException {
-        return null;
+        if (!(path instanceof AzurePath)) {
+            throw LoggingUtility.logError(logger, new IllegalArgumentException("This provider cannot operate on "
```
kasobol-msft: limitation on path type should be mentioned in javadoc I think and what is expected if somebody passes something other than AzurePath.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureDirectoryStream.java
rickle-msft: Ooo probably. Save doing it every time. Good thought!

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureResource.java
rickle-msft: Definitely will do that.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureDirectoryStream.java
rickle-msft: It's listed as an optional specific exception, so I was going to save checking the status, but can make a call to ensure it's a directory.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/AzureFileSystemProvider.java
rickle-msft: So I've been doing this defensively just to be safe, but I don't think under normal circumstances paths usually get passed to providers other than their own. People usually work through the Files type, which gets the appropriate provider for the given path and then passes it to the method on the provider internally. To screw this up, someone would have to query for the provider themselves and make an explicit all. 

The docs also aren't super clear to me on what the behavior should be if a path is used with a provider that is not its own. I suppose instead of asserting the right type, I could always just get the string representation of the path and construct an AzurePath from that. Thoughts?

kasobol-msft: I'm questioning that validation. I'm just saying that we should have `@throws` section in javadoc for this method explaining that behavior - which is not obvious by just reading method signature (it promises to take any `Path`).

Whether to keep validation or try to convert other Paths. I think going with validation is good start. It's easier to loosen it later than make it strict, so let's be strict and wait for feedback.

PR #9694, opened 2020-03-30
Fix receive link processor request upstream logic

srnagar:
When a receive link closes, requestUpstream call in AmqpReceiveLinkProcessor triggers onNext() with already closed link due to how repeat() works and then when another requestUpstream is triggered, the request is not propagated to the upstream. This change is to propagate that request upstream to actually get the new receive link that is not closed.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorClient.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorClient.java
@@ -14,6 +14,9 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventProcessorClient.java
```diff
@@ -122,11 +122,14 @@ public synchronized void start() {
             return;
         }
         logger.info("Starting a new event processor instance with id {}", this.identifier);
-        scheduler.set(Schedulers.newElastic("EventProcessor"));
+
+        ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
```
conniey: Out of curiousity, why the switch to an explicit ScheduledExecutorService over Schedulers.single() and [Scheduler.schedulePeriodically](https://projectreactor.io/docs/core/release/api/reactor/core/scheduler/Scheduler.html#schedulePeriodically-java.lang.Runnable-long-long-java.util.concurrent.TimeUnit-)?

srnagar:
Scheduler.schedulerPeriodically() runs periodically which means if the interval is set to 10 seconds, it'll run at T, T+10, T+20, T+30 and so on. If one of the load balancing runs took say 10 seconds, then the next run is triggered immediately again. ScheduledExecutorService has scheduleAtFixedDelay() which allows you to run after an interval from the time last execution completed.

PR #9697, opened 2020-03-30
ServiceBusClientBuilder updates

conniey:
* Clients implement AutoCloseable
* Receiver throws when operations are called on a disposed one
* Shared connection support
* ClientBuilder API updates.

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
@@ -489,7 +489,8 @@ public final class ServiceBusClientBuilder {
         }
 
         /**
-         * Sets whether or not to automatically complete a received message after it has been processed.
+         * Sets whether or not to automatically complete a received message after it has been processed. Only supported
+         * when using the <b>asynchronous</b> {@link ServiceBusReceiverAsyncClient receiver client}.
          *
          * @param autoComplete {@code true} to automatically complete a received message after it has been
          *     processed; {@code false} otherwise.
@@ -502,7 +503,8 @@ public final class ServiceBusClientBuilder {
         }
 
         /**
-         * Sets if lock should be automatically renewed.
+         * Sets if lock should be automatically renewed. Only supported when using the <b>asynchronous</b>
+         * {@link ServiceBusReceiverAsyncClient receiver client}.
          *
          * @param isLockAutoRenewed {@code true} if the lock should be automatically renewed; {@code false}
          *     otherwise.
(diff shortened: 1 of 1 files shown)


sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
```diff
@@ -423,7 +504,9 @@ public ServiceBusReceiverClientBuilder isAutoComplete(boolean autoComplete) {
         /**
          * Sets if lock should be automatically renewed.
          *
-         * @param isLockAutoRenewed {@code true} if the lock should be automatically renewed; {@code false} otherwise.
+         * @param isLockAutoRenewed {@code true} if the lock should be automatically renewed; {@code false}
+         *     otherwise.
```
hemanttanwar: This should probably also clarify that this will be done for Async client only. And will not autorenew for sync client.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
```diff
@@ -574,14 +644,20 @@ public ServiceBusReceiverAsyncClient buildAsyncClient() {
...
+         * Creates <b>synchronous</b> Service Bus receiver responsible for reading {@link ServiceBusMessage messages}
+         * from a specific queue or topic.
          *
          * @return An new {@link ServiceBusReceiverClient} that receives messages from a queue or topic.
+         * @throws IllegalStateException if {@link #queueName(String) queueName} or {@link #topicName(String)
+         *     topicName} are not set or, both of these fields are set. It is also thrown if the Service Bus {@link
+         *     #connectionString(String) connectionString} contains an {@code EntityPath} that does not match one set in
+         *     {@link #queueName(String) queueName} or {@link #topicName(String) topicName}. Lastly, if a {@link
+         *     #topicName(String) topicName} is set, but {@link #subscriptionName(String) subscriptionName} is not.
+         * @throws IllegalArgumentException if the entity type is not a queue or a topic.
```
hemanttanwar: Line 660: `IllegalArgumentException if the entity type is not a queue or a topic` : You have already explained all the scenario above this. `UNKNOWN` , type is internal to us and if we can avoid using `entity type` , one less thing to learn for user .   
You do throw this exception but this will only happen if user has not set both queue ,topic and connection string has entityName in it.
Or it may say something like `Queue or topic name are not set via queueNme() or topicName() respectively.`

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
```diff
@@ -208,13 +209,23 @@ public ServiceBusReceiverClientBuilder buildReceiverClientBuilder() {
      */
     void onClientClose() {
```
hemanttanwar: Probably not urgent, but should we also add Autocloseableon receiver/sender builder because some time a user might just want to close the connection from builder in order to stop everything ?  We could decide about it later.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
conniey: Fixed.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
conniey: * I added the IllegalArgumentException because checkstyles complained about missing javadocs for the exception.
* I updated the text.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
conniey: This is a design decision, you can create an issue for it.. but it seems weird for a builder to implement autocloseable.

PR #9715, opened 2020-03-31
Fixed behavior of create* methods in file system and directory clients

### LINKED ISSUE #9709: Datalake createFile/subDirectory/Directory methods should not overwrite by default.

gapra-msft:
Resolves #9709


sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileSystemAsyncClient.java
```diff
@@ -466,7 +466,31 @@ public HttpPipeline getHttpPipeline() {
      */
     public Mono<DataLakeFileAsyncClient> createFile(String fileName) {
         try {
-            return createFileWithResponse(fileName, null, null, null, null, null).flatMap(FluxUtil::toMono);
+            return createFile(fileName, false);
+        } catch (RuntimeException ex) {
```
rickle-msft: Do we need to duplicate catching the RuntimeException in every overload?

gapra-msft: Just caught a few in this class - I just pushed a commit removing them.

I left some where we flatMap(FluxUtil::toMono) cause it seems to make sense there? What do you think?

PR #9721, opened 2020-03-31
Updated EncryptedClientBuilder

rickle-msft:
Customers can now specify a key/keyResolver after they specify a pipeline/client on the builder.

The builder will now throw if a pipeline/client was already configured for decryption as the encryption info may conflict.


sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -155,8 +156,28 @@ public EncryptedBlobAsyncClient buildEncryptedBlobAsyncClient() {
...
+            List<HttpPipelinePolicy> policies = new ArrayList<>();
+            // Check that BlobDecryptionPolicy not already present while copying them over to a new policy list.
+            boolean decryptionPolicyPresent = false;
+            for (int i = 0; i < httpPipeline.getPolicyCount(); i++) {
+                HttpPipelinePolicy currPolicy = httpPipeline.getPolicy(i);
+                decryptionPolicyPresent |= currPolicy instanceof BlobDecryptionPolicy;
+                policies.add(currPolicy);
+            }
+            // If a decryption policy is not already present, add it to the front.
+            if (!decryptionPolicyPresent) {
```
gapra-msft: not particularly picky about this: but since we're now throwing on finding the decryption policy, could we just throw an exception in the for loop when we encounter it? That way we dont need to loop over all the rest of the policies if the first policy was a decryption policy?

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClientBuilder.java
```diff
@@ -473,42 +493,20 @@ public EncryptedBlobClientBuilder retryOptions(RequestRetryOptions retryOptions)
 
     /**
      * Sets the {@link HttpPipeline} to use for the service client, and adds a decryption policy if one is not present.
-     *
+     * Note that the underlying pipeline should not already be configured for configured for encryption/decryption.
```
gapra-msft: typo: configured for configured for

gapra-msft: same below in blobClient methods

PR #9722, opened 2020-03-31
Fixed sorting order of shared key headers

### LINKED ISSUE #9244: Storage Metadata not sorted correctly - causes an Authentication Issue
See the Python issue here https://github.com/Azure/azure-sdk-for-python/issues/10277

Potential Fix - Use collator to generate the String to Sign and get the correct ordering
https://docs.oracle.com/javase/7/docs/api/java/text/Collator.html

gapra-msft:
Resolves #9244

### DIFF
--- a/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageSharedKeyCredential.java
+++ b/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageSharedKeyCredential.java
@@ -158,7 +158,7 @@ public final class StorageSharedKeyCredential {
         }
 
         /* Culture-sensitive word sort */
-        Collections.sort(xmsHeaderNameArray, Collator.getInstance());
+        Collections.sort(xmsHeaderNameArray, Collator.getInstance(Locale.ROOT));
 
         final StringBuilder canonicalizedHeaders = new StringBuilder();
         for (final String key : xmsHeaderNameArray) {


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageSharedKeyCredential.java
```diff
@@ -156,7 +157,8 @@ private String getAdditionalXmsHeaders(Map<String, String> headers) {
             return "";
         }
 
-        Collections.sort(xmsHeaderNameArray);
+        /* Culture-sensitive word sort */
+        Collections.sort(xmsHeaderNameArray, Collator.getInstance());
```
kasobol-msft: Should we use Collator.getInstance(Locale)  ? Otherwise this will use local machine or JVM default which might not be what we expect. I guess en_US would be safe choice.

gapra-msft: good point, yeah I wasn't too sure which one to pin it to if I did. There's also Locale.ROOT, Locale.US and Locale.ENGLISH

kasobol-msft: Javadoc of Locale.ROOT looks promising - I'd give that a try. If that doesn't work Locale.US.

PR #9724, opened 2020-03-31
Modified BlobOutputStream to wait on a condition variable until transfer is complete instead of polling

gapra-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
```diff
@@ -183,14 +196,16 @@ void commit() {
...
             // Need to wait until the uploadTask completes
+            lock.lock();
             while (!complete) {
                 try {
-                    Thread.sleep(100L);
+                    transferComplete.await();
                 } catch (InterruptedException e) {
-                    // Does this need to be caught by logger?
-                    e.printStackTrace();
+                    this.lastError = new IOException(e.getMessage());
```
gapra-msft: What am i supposed to do here?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
```diff
@@ -174,7 +182,12 @@ private BlockBlobOutputStream(final BlobAsyncClient client,
                     this.lastError = new IOException(e);
                     return Mono.empty();
                 })
-                .doOnTerminate(() -> complete = true)
+                .doOnTerminate(() -> {
+                    lock.lock();
+                    complete = true;
+                    transferComplete.signal();
```
gapra-msft, line 188: Should I try catch the entire block on the lock and unlock if a generic exception is caught? Or do yall think CI is catching a false positive?

PR #9726, opened 2020-03-31
Remove dependencies related to reflection assertion

sima-zhu:


sdk/search/tests.yml
```diff
@@ -5,9 +5,6 @@ jobs:
     parameters:
       ServiceDirectory: search
       Timeout: 120
-      # Live tests create their own resources for testing. Resources are created against a
-      # finite resource pool, so we limit parallelism to conserve availability in that pool.
-      MaxParallel: 3
```
alzimmermsft, line 10: Did we want to remove this?

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/TestHelpers.java
```diff
@@ -152,4 +96,77 @@ public static MatchConditions generateIfExistsAccessCondition() {
...
+                        assertEquals(a.asText(), b.asText());
+                    }
+                }
+            } else {
+                assertObjectEquals(expectedField.getValue(), actualNode.get(expectedField.getKey()), ignoreDefaults, ignoreFields);
+            }
+        }
+    }
+
+    private static boolean doesFieldNeedCheck(String fieldName, JsonNode fieldValue,
```
alzimmermsft, line 155: nit: Possibly rename this to `shouldCheckField`

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/TestHelpers.java
```diff
@@ -152,4 +96,77 @@ public static MatchConditions generateIfExistsAccessCondition() {
...
+        try {
+            assertEquals(jacksonAdapter.serialize(expected, SerializerEncoding.JSON),
+                jacksonAdapter.serialize(actual, SerializerEncoding.JSON));
+        } catch (IOException ex) {
+            fail("There is something wrong happen in serializer.");
+        }
+    }
+
+    public static void assertObjectEquals(Object expected, Object actual, boolean ignoreDefaults,
+        String ... ignoreFields) {
```
alzimmermsft, line 111: nit: Renamed this to `ignoredFields`

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/TestHelpers.java
```diff
@@ -152,4 +96,77 @@ public static MatchConditions generateIfExistsAccessCondition() {
...
+                    }
+                }
+            } else {
+                assertObjectEquals(expectedField.getValue(), actualNode.get(expectedField.getKey()), ignoreDefaults, ignoreFields);
+            }
+        }
+    }
+
+    private static boolean doesFieldNeedCheck(String fieldName, JsonNode fieldValue,
+        boolean ignoreDefaults, String... ignoreFields) {
```
alzimmermsft, line 156: If this is going to be passed in many times should we convert `ignoredFields` to a `Set` before processing?

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/TestHelpers.java
alzimmermsft: Another possible name is `shouldSkipField` since it's only use is negated.

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/TestHelpers.java
```diff
@@ -152,4 +96,77 @@ public static MatchConditions generateIfExistsAccessCondition() {
...
+
+        if (ignoreDefaults) {
+            if (fieldValue.isNull()) {
+                return false;
+            }
+            if (fieldValue.isBoolean() && !fieldValue.asBoolean()) {
+                return false;
+            }
+            return !fieldValue.isNumber() || fieldValue.asDouble() != 0.0D;
+        }
```
alzimmermsft, line 169: This is pretty risky code as `null`, `false`, and `0` could all be legitimate values being returned. Given that we are validating equality by traversing the JSON tree we could modify the serialization of the objects to always include default values and this section could be removed.

sdk/search/tests.yml
sima-zhu: We have so many tests which timeout for the second round. I want to test the behavior of removing this. If it works not like expected, we can add it back.

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/TestHelpers.java
sima-zhu: This is mimic the Reflection assertion IGNORE_DEFAULTS behavior. 
https://github.com/arteam/unitils/blob/master/unitils-core/src/main/java/org/unitils/reflectionassert/comparator/impl/IgnoreDefaultsComparator.java#L30

sima-zhu: There is a boolean field control this.
`public static void assertObjectEquals(Object expected, Object actual, boolean ignoreDefaults,
        String ... ignoreFields) {`

If we want to test ignoredDefault, we can set it to false.

sdk/search/tests.yml
sima-zhu: There are 20 free quota under subscription. We only use 6 of them. Other languages sequentially executes on the same resources. Quickly finish our tests can minimize the conflicts with other languages.

alzimmermsft: Sounds good, if the bucketing is working correctly it would be better for us to try to finish all our runs as fast as possible.

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/TestHelpers.java
```diff
@@ -152,4 +98,93 @@ public static MatchConditions generateIfExistsAccessCondition() {
...
+        assertOnMapIterator(expectedNode.fields(), actualNode, ignoredDefaults,ignoredFields);
+    }
+
+    private static void assertOnMapIterator(Iterator<Map.Entry<String, JsonNode>> expectedNode,
+        ObjectNode actualNode, boolean ignoredDefaults, String[] ignoredFields) {
+        while (expectedNode.hasNext()) {
+            assertTrue(actualNode.fields().hasNext());
+            Map.Entry<String, JsonNode> expectedField = expectedNode.next();
+            String fieldName = expectedField.getKey();
+            Set<String> ignoredFieldSet = new HashSet<>(Arrays.asList(ignoredFields));
```
alzimmermsft, line 140: You'll want to move this outside of the while loop, right now this will be recreated each time.

alzimmermsft:
Could you do a quick check to see if these dependencies were used anywhere else in our repo, if not could you remove them from the `external_dependencies` file.

PR #9727, opened 2020-04-01
Pluggable Modules for JSON Serialization

alzimmermsft:
This PR adds the `JsonSerializer` interface, with accompanying implmentations using Jackson and GSON, which will allow customers to pass in their own serialization logic that will handle serializing and deserializing their custom types. This interface won't replace the serialization logic in Azure Core that handles service types, such as an AppConfiguration setting, KeyVault secret, or Storage blob properties. This will allow clients that deal with customer object to accept a `JsonSerializer` that will act as a companion to the Azure Core serializer to serialize and deserialize the customer type, it will also allow for non-Jackson annotated models to have full support.

**Example: Search Get Document**

_With Customer Serializer_

```java
SearchIndexClient searchIndexClient = new SearchIndexClientBuilder()
    .indexName(indexName)
    .credential(credential)
    .serializer(jsonSerializer) // This is a potential representation of the change in a client library.
    .buildClient();

CustomerType customerType = searchIndexClient.getDocument(documentKey, CustomerType.class);

// This will still support getting documents as a SearchDocument.
SearchDocument searchDocument = searchIndexClient.getDocument(documentKey);
```

_Without Customer Serializer_

```java
SearchIndexClient searchIndexClient = new SearchIndexClientBuilder()
    .indexName(indexName)
    .credential(credential)
    .buildClient();

SearchDocument searchDocument = searchIndexClient.getDocument(documentKey); // SearchDocument is effectively a `Map<String, Object>`.
```

### DIFF
--- a/eng/versioning/external_dependencies.txt
+++ b/eng/versioning/external_dependencies.txt
@@ -7,7 +7,7 @@ com.fasterxml.jackson.dataformat:jackson-dataformat-xml;2.10.1
 com.fasterxml.jackson.datatype:jackson-datatype-jsr310;2.10.1
 com.github.spotbugs:spotbugs;4.0.0-beta3
 com.github.spotbugs:spotbugs-maven-plugin;3.1.12.2
-com.google.code.gson:gson;2.8.6
+com.google.code.gson:gson;2.8.5
 com.google.guava:guava;24.1.1-jre
 com.h3xstream.findsecbugs:findsecbugs-plugin;1.9.0
 com.microsoft.azure:azure-annotations;1.7.0
--- a/sdk/core/azure-core-serializer-json-gson/pom.xml
+++ b/sdk/core/azure-core-serializer-json-gson/pom.xml
@@ -64,7 +64,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.8.6</version> <!-- {x-version-update;com.google.code.gson:gson;external_dependency} -->
+      <version>2.8.5</version> <!-- {x-version-update;com.google.code.gson:gson;external_dependency} -->
     </dependency>
 
     <dependency>
@@ -92,7 +92,7 @@
             <bannedDependencies>
               <includes>
                 <include>com.azure:*</include>
-                <include>com.google.code.gson:gson:[2.8.6]</include> <!-- {x-include-update;com.google.code.gson:gson;external_dependency} -->
+                <include>com.google.code.gson:gson:[2.8.5]</include> <!-- {x-include-update;com.google.code.gson:gson;external_dependency} -->
               </includes>
             </bannedDependencies>
           </rules>
--- a/sdk/eventhubs/microsoft-azure-eventhubs-eph/pom.xml
+++ b/sdk/eventhubs/microsoft-azure-eventhubs-eph/pom.xml
@@ -45,7 +45,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.8.6</version> <!-- {x-version-update;com.google.code.gson:gson;external_dependency} -->
+      <version>2.8.5</version> <!-- {x-version-update;com.google.code.gson:gson;external_dependency} -->
     </dependency>
 
     <dependency>
--- a/sdk/identity/azure-identity/pom.xml
+++ b/sdk/identity/azure-identity/pom.xml
@@ -90,7 +90,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.8.6</version> <!-- {x-version-update;com.google.code.gson:gson;external_dependency} -->
+      <version>2.8.5</version> <!-- {x-version-update;com.google.code.gson:gson;external_dependency} -->
       <scope>test</scope>
     </dependency>
   </dependencies>


sdk/core/azure-core-serializer-avro-apache/src/main/java/module-info.java
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module azure.core.serializer.avro.apache {
```
JonathanGiles, line 4:
```suggestion
module com.azure.core.serializer.avro.apache {
```

sdk/core/azure-core-serializer-avro-jackson/src/main/java/module-info.java
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module azure.core.serializer.avro.jackson {
```
JonathanGiles, line 4:
```suggestion
module com.azure.core.serializer.avro.jackson {
```

sdk/core/azure-core-serializer-json-gson/src/main/java/module-info.java
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module azure.core.serializer.json.gson {
```
JonathanGiles, line 4:
```suggestion
module com.azure.core.serializer.json.gson {
```

sdk/core/azure-core-serializer-json-jackson/src/main/java/module-info.java
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module azure.core.serializer.json.jackson {
```
JonathanGiles, line 4:
```suggestion
module com.azure.core.serializer.json.jackson {
```

sdk/core/azure-core-serializer-json-gson/src/main/java/com/azure/core/serializer/json/gson/GsonJsonSerializer.java
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.serializer.json.gson;
+
+/*
+ * Thoughts on this class skipping the builder pattern?
```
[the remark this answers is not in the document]
alzimmermsft: I'm wondering if it would be easier/more convenient to just have these classes take the pre-built serializer for a given module, a `Gson` class for gson JSON parsing, `ObjectMapper` for Jackson, etc.

sdk/core/azure-core/src/main/java/com/azure/core/serializer/JsonDeserializer.java
```diff
@@ -0,0 +1,4 @@
+package com.azure.core.serializer;
+
+public interface JsonDeserializer {
```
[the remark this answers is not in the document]
alzimmermsft: I was just throwing interfaces into Azure Core to get a feeling on whether we would want overarching interfaces for serialization, general consensus right now is leaning away from that, whether each interface should be broken down into its serialization and deserialization components.

eng/versioning/external_dependencies.txt
```diff
@@ -3,6 +3,7 @@
 com.fasterxml.jackson.core:jackson-annotations;2.10.1
 com.fasterxml.jackson.core:jackson-core;2.10.1
 com.fasterxml.jackson.core:jackson-databind;2.10.1
+com.fasterxml.jsckson.dataformat:jackson-dataformat-avro;2.10.1
```
joshfree, line 6: typo: jackson

pom.client.xml
```diff
@@ -558,6 +558,22 @@
                 <title>Azure Core - Management</title>
                 <packages>com.azure.core.management*</packages>
               </group>
+              <group>
+                <title>Azure Core - Serializer - Apache Avro</title>
```
JonathanGiles, line 562: I would have this worded slightly differently perhaps - 'Azure Core - Serializer - Avro - Apache', etc.

sdk/core/azure-core-serializer-avro-apache/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-enforcer-plugin;external_dependency} -->
+        <configuration>
+          <rules>
+            <bannedDependencies>
+              <includes>
+                <include>com.azure:*</include>
+                <include>org.apache.avro:avro:[1.9.2]</include>
```
JonathanGiles, line 82: Don't need square brackets around the 1.9.2

sdk/core/azure-core/src/main/java/com/azure/core/serializer/JsonSerializer.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Converts the object into a JSON string and writes it to the {@link OutputStream}.
+     *
+     * @param value The object.
+     * @param stream The {@link OutputStream} where the JSON string will be written.
+     * @param clazz {@link Class} representing the object.
+     * @return An indicator that the object's JSON string has been written to the {@link OutputStream}.
+     */
+    Mono<Void> write(Object value, OutputStream stream, Class<?> clazz);
```
JonathanGiles, line 58: What is the use case for this method with the clazz argument?

sdk/core/azure-core-serializer-avro-jackson/src/main/java/com/azure/core/serializer/avro/jackson/JacksonAvroSerializer.java
```diff
@@ -0,0 +1,69 @@
...
+        this.mapper = new AvroMapper();
+    }
+
+    /**
+     * Constructs a {@link AvroSerializer} using the passed Jackson serializer.
+     *
+     * @param mapper Configured Jackson serializer.
+     */
+    public JacksonAvroSerializer(AvroMapper mapper) {
+        this.mapper = mapper.copy();
```
srnagar, line 33: Can this just used the provided mapper instead of making a copy?

sdk/core/azure-core-serializer-avro-jackson/src/main/java/com/azure/core/serializer/avro/jackson/JacksonAvroSerializer.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public <T> Mono<T> read(byte[] input, AvroSchema schema) {
+        return Mono.defer(() -> {
+            try {
+                return Mono.just(mapper.reader().with(schema).readValue(input));
+            } catch (IOException ex) {
+                return Mono.error(ex);
```
srnagar, line 42: log and return error.

sdk/core/azure-core-serializer-avro-jackson/src/main/java/com/azure/core/serializer/avro/jackson/JacksonAvroSerializer.java
```diff
@@ -0,0 +1,69 @@
...
+                return Mono.just(mapper.writer().with(schema).writeValueAsBytes(value));
+            } catch (IOException ex) {
+                return Mono.error(ex);
+            }
+        });
+    }
+
+    @Override
+    public Mono<Void> write(Object value, AvroSchema schema, OutputStream stream) {
+        return Mono.defer(() -> Mono.fromRunnable(() -> {
```
srnagar, line 60: `Mono.fromRunnable()` should be sufficient. Don't need to wrap it with `Mono.defer()`

sdk/core/azure-core-serializer-avro-apache/src/main/java/com/azure/core/serializer/avro/apache/ApacheAvroSerializer.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Apache based implementation of the {@link AvroSerializer} interface.
+ */
+public class ApacheAvroSerializer implements AvroSerializer {
+    private static final Schema.Parser PARSER = new Schema.Parser();
+    private static final DecoderFactory DECODER_FACTORY = DecoderFactory.get();
+    private static final EncoderFactory ENCODER_FACTORY = EncoderFactory.get();
+
+    @Override
+    public <T> Mono<T> read(byte[] input, String schema) {
```
srnagar, line 30: We should have null checks for input params (in all these read/write methods)

sdk/core/azure-core-serializer-avro-apache/pom.xml
alzimmermsft: Maven Enforcer has a slightly different mechanics for handling version numbers. A plain number means an inclusive range beginning with the listed version, surrounding it with `[]` changes the logic the exclusively be the listed version.

pom.client.xml
alzimmermsft: Updated to use this convention

sdk/core/azure-core-serializer-avro-jackson/src/main/java/com/azure/core/serializer/avro/jackson/JacksonAvroSerializer.java
alzimmermsft: Removed the `copy` call.

sdk/core/azure-core-serializer-avro-jackson/src/main/java/com/azure/core/serializer/avro/jackson/JacksonAvroSerializer.java
alzimmermsft: Done

sdk/core/azure-core-serializer-avro-jackson/src/main/java/com/azure/core/serializer/avro/jackson/JacksonAvroSerializer.java
alzimmermsft: Done

srnagar:
Adding @jsquire as he is also looking into Avro serializers for Event Hubs.

sdk/core/azure-core/src/main/java/com/azure/core/serializer/JsonSerializer.java
```diff
@@ -0,0 +1,59 @@
...
+public interface JsonSerializer {
+    /**
+     * Reads the JSON byte stream into its object representation.
+     *
+     * @param input JSON byte stream.
+     * @param clazz {@link Class} representing the object.
+     * @param <T> Type of the object.
+     * @return The object representing the JSON string.
+     */
+    <T> Mono<T> read(byte[] input, Class<T> clazz);
```
moderakh, line 22: why this needs to be async? You have the byte array and class, this only require some computational work. IMO this doesn't need to be async. same for the other APIs as well.

alzimmermsft: Changed to sync methods.

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,84 @@
...
+            return mapper.writeValueAsBytes(value);
+        } catch (IOException ex) {
+            throw logger.logExceptionAsError(Exceptions.propagate(ex));
+        }
+    }
+
+    @Override
+    public byte[] write(Object value, Class<?> clazz) {
+        try {
+            return mapper.writerFor(clazz).writeValueAsBytes(value);
```
moderakh, line 57: is this more performant than `mapper.writeValueAsBytes(value)`?

alzimmermsft: Nope, actually, I'll be removing this as I've read into this more. `writerFor` is generally meant for scenarios where you want to write the object as one of its super classes or an interface it implements.

alzimmermsft: Removed

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public <T> T read(byte[] input, Class<T> clazz) {
+        if (input == null) {
+            return null;
+        }
+
+        try {
+            return mapper.readValue(input, clazz);
```
moderakh, line 39: I don't remember of top of my head, does this support ObjectNode?

i.e., `mapper.readvalue(input, ObjectNode.class)`

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,64 @@
...
+        } catch (IOException ex) {
+            throw logger.logExceptionAsError(Exceptions.propagate(ex));
+        }
+    }
+
+    @Override
+    public byte[] write(Object value) {
+        try {
+            return mapper.writeValueAsBytes(value);
```
moderakh, line 48: I don't remember of top of my head, does this support ObjectNode?

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializer.java
alzimmermsft: I believe that would be ObjectMapper.readTree, would this be a common scenario for Cosmos?

moderakh: yes, that's a common scenario.

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializer.java
alzimmermsft: This would be for adding a node into a tree being written, correct?

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializer.java
moderakh: how about adding a if/else branch to the same method on the class type?

kirankumarkolli:
Please also cover the error/exception/warning handling semantics.

kirankumarkolli:
Copy from of related comments from [.NET PR](https://github.com/Azure/azure-sdk-for-net/pull/10978#issuecomment-614675354): 
One more scenarios we started recently hearing from CX is per-request serialization customization. 
The new abstraction should explore this scenario possibility.

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializer.java
alzimmermsft: Is the tree reading logic used when dealing with customer related objects or when dealing with service objects? I updated the PR description to better clarify the purpose of the interface added. If the tree reading scenario is common with customer object I'll dive into it more, I'd also rather see it being a separate API than tied into the `read` API which is more data-bindy.

alzimmermsft:
> 
> 
> https://jsoniter.com/ is a famous json lib. They claim it has more than 2x perf improvement comparing to jackson.
> 
> It is worth adding a serializer implementation for jsoniter to make sure the apis we designed works for jsoniter

I'll take a look at jsoniter's APIs around data-bind style serialization and deserialization to see how well the interface conforms to them.

sdk/core/azure-core-serializer-json-gson/src/main/java/com/azure/core/serializer/json/gson/GsonJsonSerializerBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Sets the {@link Gson} that will be used during serialization.
+     * <p>
+     * If this is set to {@code null} the default {@link Gson} will be used.
+     *
+     * @param gson {@link Gson} that will be used during serialization.
+     * @return The updated GsonJsonSerializerBuilder class.
+     */
+    public GsonJsonSerializerBuilder gson(Gson gson) {
```
JonathanGiles, line 33: Let's find a better name for the setter that sets the implementation in these builders.

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializerBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Sets the {@link ObjectMapper} that will be used during serialization.
+     * <p>
+     * If this is set to {@code null} the default {@link ObjectMapper} will be used.
+     *
+     * @param objectMapper {@link ObjectMapper} that will be used during serialization.
+     * @return The updated JacksonJsonSerializerBuilder class.
+     */
+    public JacksonJsonSerializerBuilder objectMapper(ObjectMapper objectMapper) {
```
JonathanGiles, line 33: This name might be more appropriate than `gson` is.

sdk/core/azure-core-serializer-json-gson/src/main/java/com/azure/core/serializer/json/gson/GsonJsonSerializerBuilder.java
alzimmermsft: I'll play with some API names since this will go against the guideline that setters should be named the same as the parameter type.

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializerBuilder.java
alzimmermsft: I would agree, the name here is better.

sdk/core/azure-core-serializer-json-gson/src/main/java/com/azure/core/serializer/json/gson/GsonJsonSerializerBuilder.java
alzimmermsft: Thoughts on converging both the GSON and Jackson builders to use `serializer` when configuring the `Gson` or `ObjectMapper` respectively.

JonathanGiles: yeah, I thought about `serializer` or `implementation`. Neither are great, but it probably makes sense to just have a standard name for all builders.

alzimmermsft: Updated to use `serializer`.

sdk/core/azure-core-serializer-json-jackson/src/main/java/com/azure/core/serializer/json/jackson/JacksonJsonSerializer.java
moderakh: ObjectNode is used for both internal cosmos types, and also for user apis.

as cosmos is a schema less database, we also would like to provide a schema less type, for that reason we also support ObjectNode in the public surface area. The other reason for supporting ObjectNode in the public surface area is perf as deserialization to ObjectNode doesn't require reflection whereas for deserializing to user provivded Pojo, jackson will use reflection which is slow.

As an example all of the following public API invocation currently work:

```java
UserPojo userPojo = cosmosContainer.readItem(id, pk, UserPojo.class).block().getItem();
ObjectNode objectNode = cosmosContainer.readIteam(id, pk, ObjectNode.class).blcok.getItem();
```

moderakh: btw, at least for jackson, ObjectMapper is capable of handling ObjectNode under the same API:

```java
 static class UserPojo {
        @JsonProperty
        String id;

        @JsonProperty
        String pk;
    }

    public static void main(String[] args) throws Exception {

        UserPojo userPojo = new UserPojo();
        userPojo.id = "myId";
        userPojo.pk = "mypk";

        ObjectMapper objectMapper = new ObjectMapper();
        byte[] bytes = objectMapper.writeValueAsBytes(userPojo);

        ObjectNode deserializedObjectNode = objectMapper.readValue(bytes, ObjectNode.class);
        UserPojo deserializedPojo = objectMapper.readValue(bytes, UserPojo.class);

        assert (deserializedObjectNode.get("id").asText().equals(deserializedPojo.id));

        System.out.println("done");
    }
```

alzimmermsft: @moderakh so support for serializing and deserializing `ObjectNode` is needed? I did some digging into the Jackson code and it does have a quick path if `ObjectNode.class` is passed as the `Class<T>` into the deserialize methods. I'll add additional tests to the PR which test this out.

Still digging into what GSON does, they do have a similar class `JsonObject` which I imagine also gets the same treatment.

sdk/core/azure-core-serializer-json-gson/src/main/java/com/azure/core/serializer/json/gson/GsonJsonSerializer.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public <T> T deserialize(byte[] input, Class<T> clazz) {
+        return gson.fromJson(CoreUtils.bomAwareToString(input, null), clazz);
+    }
+
+    @Override
+    public byte[] serialize(Object value) {
+        return gson.toJson(value).getBytes(StandardCharsets.UTF_8);
+    }
```
srnagar, line 38: Should serialize also support other charsets instead of always using UTF 8?

sdk/core/azure-core-serializer-json-gson/src/main/java/module-info.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module com.azure.core.serializer.json.gson {
+    requires transitive com.azure.core;
+    requires transitive gson;
```
srnagar, line 6: This should export `com.azure.core.serializer.json.gson`.

sdk/core/azure-core-serializer-json-jackson/src/main/java/module-info.java
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module com.azure.core.serializer.json.jackson {
+    requires transitive com.azure.core;
```
srnagar, line 5: Export `com.azure.core.serializer.json.jackson`.

sdk/core/azure-core/src/main/java/com/azure/core/serializer/JsonSerializer.java
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.core.serializer;
```
srnagar, line 4: We already have a package named `com.azure.core.util.serializer`. Should we move this interface to that package instead of creating this new package?

sdk/core/azure-core-serializer-json-gson/src/main/java/module-info.java
alzimmermsft: Added

sdk/core/azure-core-serializer-json-jackson/src/main/java/module-info.java
alzimmermsft: Added

sdk/core/azure-core/src/main/java/com/azure/core/serializer/JsonSerializer.java
alzimmermsft: Doesn't hurt to move it into that package, the package currently contains Jackson deserializer/serializer implementations so I don't know if we want to mix this interface into there.

alzimmermsft: Moved into `util.serializer`.

sdk/core/azure-core-serializer-json-gson/src/main/java/com/azure/core/serializer/json/gson/GsonJsonSerializer.java
alzimmermsft: I believe it is safe to default to using UTF-8 when dealing with JSON, looking around in other places in code `HttpRequest` defaults to using `UTF-8` when sending `String` bodies, so at minimum this will fall in line with that.

sdk/core/azure-core-serializer-json-gson/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-enforcer-plugin;external_dependency} -->
+        <configuration>
+          <rules>
+            <bannedDependencies>
+              <includes>
+                <include>com.azure:*</include>
+                <include>com.google.code.gson:gson:[2.8.5]</include> <!-- {x-include-update;com.google.code.gson:gson;external_dependency} -->
```
JonathanGiles, line 95: Our [dependencies report](https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-java/staging/pom.client.html) is a bit messy around GSON. We should get MSAL to upgrade to 2.8.6 and move this to 2.8.6 as well.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/JsonSerializer.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    byte[] serialize(Object value);
+
+    /**
+     * Converts the object into a JSON byte stream and writes it to the {@link OutputStream}.
+     *
+     * @param value The object.
+     * @param stream The {@link OutputStream} where the JSON byte stream will be written.
+     */
+    void serialize(Object value, OutputStream stream);
```
JonathanGiles, line 36: Is there an immediate use case for this API?

sdk/core/azure-core-serializer-json-gson/pom.xml
alzimmermsft: I'll go ahead an update the version, wasn't going to proactively update it without a reason.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/JsonSerializer.java
alzimmermsft: This should be removable and has a viable work around without the existence of the API.

**With API**

```java
serialize(value, stream);
```

**Without API**

```java
stream.write(serialize(value));
```

JonathanGiles: Is there a use case though that the API would enable, in particular with greater perf?

alzimmermsft: There may be a minor performance enhancement but the underlying serialization logic is generally the same when using either `write(Object)` or `write(Object, OutputStream)` APIs. Given the API is for a pattern we generally don't see in our libraries I don't think keeping it adds much worth.

PR #9733, opened 2020-04-01
Added CustomHostname connection string parameter

favoretti:
## Description
In a case where you need to use custom DNS name for a private EventHub
endpoint and do not have the ability to setup up DNS forwarding from
on-premise to Azure DNS you need to split the Endpoint namespace name
from the actual hostname to connect to.

This PR introduces such an ability by adding `CustomHostname`
ConnectionString parameter.

- [x] **I have read the [contribution guidelines](./CONTRIBUTING.md).**
- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR.
- [x] The pull request does not introduce breaking changes (unless a major version change occurs in the assembly and module).
- [x] If applicable, the public code is properly documented.
- [x] Pull request includes test coverage for the included changes.
- [x] The code builds without any errors.

### DIFF
--- a/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ConnectionOptions.java
+++ b/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ConnectionOptions.java
@@ -23,11 +23,12 @@ public class ConnectionOptions {
     private final ProxyOptions proxyOptions;
     private final Scheduler scheduler;
     private final String fullyQualifiedNamespace;
+    private final String customHostName;
     private final CbsAuthorizationType authorizationType;
 
     public ConnectionOptions(String fullyQualifiedNamespace, TokenCredential tokenCredential,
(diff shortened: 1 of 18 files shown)


conniey:
Thanks for contributing, @favoretti! :) I'm trying to understand your scenario a bit more. Looking at the code, does replacing `fullyQualifiedNamespace` in your connection string not work? (ie. `builder.connectionString("sb://10.23.32.34;EntityPath=MyHub;SharedAccessKeyName=root;SharedAccessKey=value")`)

favoretti:
Hi @conniey. It won't. EH needs that FQNS since it uses it as a hash for token and so on. If I change the hostname to, say, my-custom-hub.fedex.com, the client will also try to connect to that namespace on AMQP level and won't find it.

favoretti:
Also namespace name is a fully qualified domain name, so once that's replaced it can't find it in the EH anymore.

favoretti:
Nevermind, this doesn't completely work, still trying to implement it, I'll open this as an issue instead to begin with..

PR #9743, opened 2020-04-01
Enforcing dependency whitelists in Maven

JonathanGiles:
Can be run with `mvn -f pom.client.xml enforcer:enforce`

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -438,28 +438,6 @@
                   </excludes>
                   <includes>
                     <include>com.azure:*</include>
-                    <include>io.projectreactor</include>
-                    <include>io.projectreactor.netty</include>
-                    <include>io.netty</include>
-                    <include>com.squareup.okhttp3</include>
-                    <include>org.slf4j</include>
-                    <include>com.fasterxml.jackson.*</include>
-                    <include>io.opentelemetry</include>
-                    <include>net.java.dev.jna</include>
-
-                    <!-- azure-core-amqp requires the following: -->
-                    <include>org.apache.qpid:proton-j</include>
-                    <include>com.microsoft.azure:qpid-proton-j-extensions</include>
-
-                    <!-- FIXME temporary support for cosmos dependencies -->
-                    <include>io.micrometer:micrometer-core</include>
-                    <include>io.dropwizard.metrics:metrics-core</include>
-                    <include>com.fasterxml.uuid:java-uuid-generator</include>
-
-                    <!-- azure-identity requires the following: -->
-                    <include>com.nimbusds:oauth2-oidc-sdk</include>
-                    <include>com.microsoft.azure:msal4j</include>
-                    <include>org.nanohttpd:nanohttpd</include>
 
                     <!-- Used by many libraries to bring in annotations used by Reactor -->
                     <include>com.google.code.findbugs:jsr305</include>
--- a/sdk/core/azure-core-amqp/pom.xml
+++ b/sdk/core/azure-core-amqp/pom.xml
@@ -102,4 +102,41 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-enforcer-plugin;external_dependency} -->
+        <executions>
+          <execution>
+            <id>default-cli</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <bannedDependencies>
+                  <searchTransitive>false</searchTransitive>
+                  <excludes>
+                    <!-- We exclude everything that is runtime-time, compile-time, or provided scoped (i.e. we allow
+                    dependencies when they are used for test scope or otherwise) -->
+                    <exclude>*:*:*:*:runtime</exclude>
+                    <exclude>*:*:*:*:compile</exclude>
+                    <exclude>*:*:*:*:provided</exclude>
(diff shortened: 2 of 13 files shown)


pom.client.xml
```diff
@@ -408,6 +408,87 @@
...
+                    <exclude>*:*:*:*:runtime</exclude>
+                    <exclude>*:*:*:*:compile</exclude>
+                    <exclude>*:*:*:*:provided</exclude>
+                  </excludes>
+                  <includes>
+                    <include>com.azure:*</include>
+                    <include>io.projectreactor</include>
+                    <include>io.projectreactor.netty</include>
+                    <include>io.netty</include>
+                    <include>com.squareup.okhttp3</include>
```
alzimmermsft, line 444: Do we want to limit these inclusion to the Netty and OkHttp clients?

pom.client.xml
```diff
@@ -408,6 +408,87 @@
...
+                    <include>org.slf4j</include>
```
alzimmermsft, line 445: Could we specifically scope this down to `org.slf4j:slf4j-api`, I don't want use seeing `slf4j-simple` getting included again.

pom.client.xml
```diff
@@ -408,6 +408,87 @@
...
+                    <include>com.fasterxml.jackson.*</include>
+                    <include>io.opentelemetry</include>
```
alzimmermsft, line 447: Should this be scoped to the tracing library only?

pom.client.xml
```diff
@@ -408,6 +408,87 @@
...
+                    <include>net.java.dev.jna</include>
```
alzimmermsft, line 448: Mind adding a comment for what this is and where it is being used.

pom.client.xml
```diff
@@ -408,6 +408,87 @@
...
+
+                    <!-- FIXME temporary support for cosmos dependencies -->
+                    <include>io.micrometer:micrometer-core</include>
+                    <include>io.dropwizard.metrics:metrics-core</include>
+                    <include>com.fasterxml.uuid:java-uuid-generator</include>
+
+                    <!-- azure-identity requires the following: -->
+                    <include>com.nimbusds:oauth2-oidc-sdk</include>
+                    <include>com.microsoft.azure:msal4j</include>
+                    <include>org.nanohttpd:nanohttpd</include>
```
alzimmermsft, line 462: Could these be moved into their library specific POMs?

pom.client.xml
```diff
@@ -408,6 +408,87 @@
...
+                </requireReleaseDeps>
+
+                <!--
+                Rule: https://maven.apache.org/enforcer/enforcer-rules/requireSameVersions.html
+                This rule is used to enforce that all modules depend on the same version of particular dependencies or
+                plugins.
+                -->
+                <requireSameVersions>
+                  <dependencies>
+                    <dependency>com.azure.*</dependency>
```
alzimmermsft, line 483: I don't know if this is going to work all the time as we allow for unreleased dependencies when a module needs it.

pom.client.xml
```diff
@@ -408,6 +408,87 @@
...
+                <requireReleaseDeps>
+                  <message>No Snapshots Allowed!</message>
+                </requireReleaseDeps>
+
+                <!--
+                Rule: https://maven.apache.org/enforcer/enforcer-rules/requireSameVersions.html
+                This rule is used to enforce that all modules depend on the same version of particular dependencies or
+                plugins.
+                -->
+                <requireSameVersions>
```
alzimmermsft, line 481: Could we also add the following checks?

https://maven.apache.org/enforcer/enforcer-rules/requireReleaseVersion.html
https://maven.apache.org/enforcer/enforcer-rules/requirePluginVersions.html
https://maven.apache.org/enforcer/enforcer-rules/requirePrerequisite.html
https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html

pom.client.xml
```diff
@@ -408,6 +408,87 @@
...
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-enforcer-plugin;external_dependency} -->
+        <executions>
+          <execution>
+            <id>default-cli</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
```
alzimmermsft, line 422: Would moving this configuration outside of the execution section, into the root of the plugin allow for easier overriding in the inheriting POMs?

pom.client.xml
JimSuplizio, line 483: @JonathanGiles This is the reason that the analyze step is failing. The configuration for the plugin overloads both dependency and plugins for the requireSameVersion configuration. I've fixed the pom file version scanner as part of [this PR](https://github.com/Azure/azure-sdk-for-java/pull/9787) which has been reviewed and merged into master. Once you update your branch and resubmit the analyze step should pass.

pom.client.xml
JonathanGiles: Good point

pom.client.xml
JonathanGiles: Done

pom.client.xml
JonathanGiles: I didn't really want too many special cases, but this is fair - I'll move some of the special cases down into the specific client library to minimise their use outside of that library.

pom.client.xml
JonathanGiles: I've moved it into the azure-identity whitelist

pom.client.xml
JonathanGiles: Everything has been relocated now.

pom.client.xml
JonathanGiles: Yes, I don't particularly care about these other rules - they can be removed when they become a nuisance.

PR #9747, opened 2020-04-01
Update ci yaml for style check

ChenTanyi:
@weidongxu-microsoft @xccc-msft Since the error in pipeline doesn't show out. You could use this to run style check in root folder.
```
mvn -pl com.azure:azure-parent -am -amd install
mvn -pl com.azure:azure-parent -amd checkstyle:chek
```

### DIFF
--- a/sdk/management/pipelines/fluentJavaMgmt.yml
+++ b/sdk/management/pipelines/fluentJavaMgmt.yml
@@ -84,7 +84,7 @@ jobs:
         displayName: 'install'
         inputs:
           mavenPomFile: sdk/management/pom.xml
-          options: '$(DefaultOptions) $(ProfileFlag) -pl !samples javadoc:aggregate -DskipTests=true -DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -am'
+          options: '$(DefaultOptions) $(ProfileFlag) -pl !samples javadoc:aggregate -DskipTests=true -Dcheckstyle.skip=true -DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -am'
           mavenOptions: '$(LoggingOptions)'
           javaHomeOption: 'JDKVersion'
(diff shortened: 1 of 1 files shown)


sdk/management/pipelines/fluentJavaMgmt.yml
```diff
@@ -84,8 +82,8 @@ jobs:
       - task: Maven@3
         displayName: 'install'
         inputs:
-          mavenPomFile: sdk/management/pom.xml
-          options: '--batch-mode $(DefaultOptions) $(ProfileFlag) -pl !samples javadoc:aggregate -DskipTests=true -DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -am'
+          mavenPomFile: pom.xml
+          options: '$(DefaultOptions) $(ProfileFlag) -pl com.azure:azure-parent javadoc:aggregate -DskipTests=true -DpackageOutputDirectory=$(Build.ArtifactStagingDirectory) -Dcheckstyle.skip -am -amd'
```
yungezz, line 86: you'll set checkstyle back right

ChenTanyi: Yes, but stylecheck is not in here, is at the last commit I commented.

This checkstyle skip will be always, since the checkstyle tool is incompatible for its checkstyle

ChenTanyi:
After communicate with weidong, I'll first remove the management from root folder. You will need to install tool manually.
```
cd eng/code-quality-reports/
mvn install
cd ../../sdk/management/
mvn checkstyle:check
```

sdk/management/pipelines/fluentJavaMgmt.yml
yungezz: ok

ChenTanyi:
Fix for the install error, I think it will pass CI now. I'll fix header first.

PR #9749, opened 2020-04-01
Support authentication based on Azure Identity adoption

xseeseesee:
Fix [Azure/azure-libraries-for-java#1111](https://github.com/Azure/azure-libraries-for-java/issues/1111)

### DIFF
--- a/sdk/appservice/mgmt/src/test/resources/session-records/canCRUDFunctionApp.json
+++ b/sdk/appservice/mgmt/src/test/resources/session-records/canCRUDFunctionApp.json
@@ -1,252 +1,245 @@
 {
   "networkCallRecords" : [ {
     "Method" : "PUT",
-    "Uri" : "http://localhost:1234/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/javacsmrgb34761444?api-version=2019-08-01",
+    "Uri" : "http://localhost:1234/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/javacsmrgf0d368500?api-version=2019-08-01",
     "Headers" : {
-      "User-Agent" : "azsdk-java-com.azure.management/null (1.8.0_231; Windows 10 10.0)",
+      "User-Agent" : "azsdk-java-com.azure.management.resources/2.0.0 (1.8.0_221; Windows 10 10.0)",
       "Content-Type" : "application/json"
     },
     "Response" : {
-      "date" : "Mon, 02 Mar 2020 05:30:41 GMT",
+      "date" : "Wed, 08 Apr 2020 09:14:32 GMT",
       "content-length" : "317",
       "expires" : "-1",
-      "x-ms-ratelimit-remaining-subscription-writes" : "1175",
+      "x-ms-ratelimit-remaining-subscription-writes" : "1199",
       "retry-after" : "0",
       "StatusCode" : "201",
-      "pragma" : "no-cache",
-      "x-ms-correlation-request-id" : "95a518cb-74d3-475a-bc03-9c75f269787c",
       "strict-transport-security" : "max-age=31536000; includeSubDomains",
+      "pragma" : "no-cache",
+      "x-ms-correlation-request-id" : "0b4920ca-31b9-4294-8467-e5b8beada2db",
       "x-content-type-options" : "nosniff",
-      "x-ms-routing-request-id" : "KOREASOUTH:20200302T053041Z:95a518cb-74d3-475a-bc03-9c75f269787c",
-      "connection" : "keep-alive",
+      "x-ms-routing-request-id" : "SOUTHEASTASIA:20200408T091433Z:0b4920ca-31b9-4294-8467-e5b8beada2db",
       "content-type" : "application/json; charset=utf-8",
       "cache-control" : "no-cache",
-      "x-ms-request-id" : "95a518cb-74d3-475a-bc03-9c75f269787c",
-      "Body" : "{\"id\":\"/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/javacsmrgb34761444\",\"name\":\"javacsmrgb34761444\",\"type\":\"Microsoft.Resources/resourceGroups\",\"location\":\"westus\",\"tags\":{\"date\":\"2020-03-02T05:30:39.407Z\",\"product\":\"javasdk\",\"cause\":\"automation\"},\"properties\":{\"provisioningState\":\"Succeeded\"}}"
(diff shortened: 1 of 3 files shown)


xccc-msft:
Not sure if it would be more friendly by setting method `fromServicePrincipalWithPemCertificate`/`fromServicePrincipalWithPfxCertificate`. Or we just set them as overloading of `fromServicePrincipal`.

weidongxu-microsoft:
There is a `DefaultAzureCredential` in identity, can it be the recommended one?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/authentication/AzureTokenCredentialBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        if (this.value == null) {
+            if (this.configuration.get(Configuration.PROPERTY_AZURE_CLIENT_ID) == null) {
+                throw new IllegalArgumentException("Please set credential values either by credential builder or environment variables.");
+            }
+            this.value = new EnvironmentCredentialBuilder()
+                .authorityHost(this.environment.getActiveDirectoryEndpoint())
+                .build();
+        }
+        return new AzureTokenCredential(this.domain, this.defaultSubscriptionId, this.clientId, this.environment, this.value);
```
weidongxu-microsoft, line 108: Does it need validation before/after this `build` that necessary fields are not null/empty?

sdk/resources/mgmt/src/main/java/com/azure/management/RestClientBuilder.java
```diff
@@ -117,8 +117,9 @@ public RestClientBuilder withBaseUrl(AzureEnvironment environment, AzureEnvironm
      * @param credential the credential.
      * @return builder
      */
-    public RestClientBuilder withCredential(TokenCredential credential) {
+    public RestClientBuilder withCredential(AzureTokenCredential credential) {
```
weidongxu-microsoft, line 120: Does this means all other credential e.g. (`ManagedIdentityCredential`) need to be wrapped in `AzureTokenCredential`?

What will be the `domain` or `clientId` if this is the case (another credential wrapped)?

Not sure about it. Why do we had to use a subclass instead of a widely used base class `TokenCredential`?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/authentication/AzureTokenCredential.java
```diff
@@ -0,0 +1,71 @@
...
+     * @return the Azure environment.
+     */
+    public AzureEnvironment getEnvironment() {
+        return this.environment;
+    }
+
+    /**
+     * @return the credential value.
+     */
+    public TokenCredential getValue() {
```
weidongxu-microsoft, line 68: Not sure `value` be a good name. Not sure from doc what it is.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/authentication/AzureCredentialFactory.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private AzureCredentialFactory() {
+    }
+
+    /**
+     * Creates the credential by environment variables
+     * @return the credential.
+     */
+    public static AzureTokenCredential fromEnvironment() {
+        return new AzureTokenCredentialBuilder()
```
weidongxu-microsoft, line 22: My suggestion is not merge credential from environment to AzureTokenCredentialBuilder. This factory class should be the place that make the AzureTokenCredential from environment variables.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/authentication/AzureCredentialFactory.java
```diff
@@ -0,0 +1,42 @@
...
+
+package com.azure.management.resources.fluentcore.authentication;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Utility class for authentication.
+ */
+public class AzureCredentialFactory {
```
yungezz, line 12: it's strange to have factory for credential..

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/authentication/AzureTokenCredential.java
```diff
@@ -0,0 +1,71 @@
...
+package com.azure.management.resources.fluentcore.authentication;
+
+import com.azure.core.credential.TokenCredential;
+import com.azure.core.management.AzureEnvironment;
+
+/**
+ * A credential provider for Azure authentication
+ * before using the APIs provided in the library.
+ */
+public class AzureTokenCredential {
```
yungezz, line 13: is token here a refresh token or access token?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/authentication/AzureTokenCredential.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private String domain;
```
yungezz, line 15: how about tenant for multiple-tenant?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/authentication/AzureTokenCredential.java
ChenTanyi: I think the `AzureTokenCredential` should implement `TokenCredential` rather than use `getValue`.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/authentication/AzureTokenCredential.java
ChenTanyi: @yungezz multi-tenant should use `AuxiliaryAuthenticationPolicy`. https://github.com/azure/azure-sdk-for-java/blob/036b9f6e6ff54a0e497f1b433f67f0eb4c32625d/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/policy/AuxiliaryAuthenticationPolicy.java#L27

xccc-msft:
Close this as we decide to prepare #10492  with RestClient refactoring together.

PR #9760, opened 2020-04-01
Track 1 Event Hubs: fix null handling in application properties when sending

JamesBirdsall:
1) Throw a useful exception if an application property key is null.
2) Accept application properties with null value.

### DIFF
--- a/sdk/eventhubs/microsoft-azure-eventhubs/src/main/java/com/microsoft/azure/eventhubs/impl/AmqpUtil.java
+++ b/sdk/eventhubs/microsoft-azure-eventhubs/src/main/java/com/microsoft/azure/eventhubs/impl/AmqpUtil.java
@@ -14,6 +14,8 @@ import org.apache.qpid.proton.message.Message;
 import java.util.Date;
 import java.util.Locale;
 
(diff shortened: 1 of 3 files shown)


sdk/eventhubs/microsoft-azure-eventhubs/src/main/java/com/microsoft/azure/eventhubs/impl/AmqpUtil.java
```diff
@@ -90,6 +95,11 @@ public static int getDataSerializedSize(Message amqpMessage) {
     }
 
     private static int sizeof(Object obj) {
+        if (obj == null) {
+            // In AMQP, null is a special type which has no data beyond the encoding byte.
+            return 0;
+        }
+
```
sjkwak, line 102: For null value scenario, is the sizeof API returning illegal argument error?

JamesBirdsall: Yes, that's where the NPE was throwing from: the first instanceof was trying to reference the object and throwing.

PR #9765, opened 2020-04-02
Fix path filters

mitchdenny:
Due to my ignorance I didn't realize that path filters in Azure Pipelines (when defined in YAML) don't patch partial path segments. So we had some instances where we had an exclude filter such as ```sdk/keyvault/microsoft-azure-``` assuming it would exclude changes in ```sdk/keyvault/microsoft-azure-keyvault``` when in fact it won't.

Unfortunately that means that we need to explicitly specify the full path segment for the directories that we want to exclude. That is what this PR is.

### DIFF
--- a/sdk/search/ci.mgmt.yml
+++ b/sdk/search/ci.mgmt.yml
@@ -15,7 +15,7 @@ trigger:
     include:
       - sdk/search/
     exclude:
-      - sdk/search/azure-search
+      - sdk/search/azure-search-documents
 
 pr:
   branches:
@@ -28,7 +28,7 @@ pr:
     include:
       - sdk/search/
     exclude:
-      - sdk/search/azure-search
+      - sdk/search/azure-search-documents
 
 stages:
   - template: ../../eng/pipelines/templates/stages/archetype-sdk-management.yml


sdk/search/ci.mgmt.yml
```diff
@@ -15,9 +15,7 @@ trigger:
     include:
       - sdk/search/
     exclude:
-      - sdk/search/azure
-      - sdk/search/microsoft-azure
-      - sdk/search/ms-azure
+      - sdk/search/azure-search
```
sima-zhu, line 18: Search rename the model to azure-search-documents

mitchdenny: Fixed.

weshaggard:
This seems like something we might need to repeat sometimes and I suspect you probably wrote a script to do it and if so can you check that script in so we can update these easier in the future?

mitchdenny:
I didn't script it (much) I just had some PowerShell that generated the include lines. I think it would be possible to automate however. But its probably a discrete task. I think we already have an issue to automate (partially) the creation of ci.yml files.

PR #9766, opened 2020-04-02
Eventgrid: regenerate 2020 04 pre

ChenTanyi:


ChenTanyi:
@weidongxu-microsoft @xccc-msft I'm not sure the fix for override function is OK or not.

xccc-msft:
It looks the variable instance `utags` is used. Can we replace `with inner().tags` if you remove this override?

ChenTanyi:
Done, remove the useless utags.

PR #9775, opened 2020-04-02
feat: update license header for mgmt

ChenTanyi:


ChenTanyi:
@weidongxu-microsoft I haven't check whether it is a generated header or not. I think all generated header is not in this format now.

weidongxu-microsoft:
@ChenTanyi Newly generated code is in this style + `Code generated ...`

https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/resources/mgmt/src/main/java/com/azure/management/resources/models/ResourceGroupsInner.java#L1

You are free to update all, as long as `Code generated ...` is kept.

ChenTanyi:
> @ChenTanyi Newly generated code is in this style + `Code generated ...`
> 
> https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/resources/mgmt/src/main/java/com/azure/management/resources/models/ResourceGroupsInner.java#L1
> 
> You are free to update all, as long as `Code generated ...` is kept.

This is the actual problem I ask for, since my simple script has not checked for the `Code generated ...`, I 'll check it later.

weidongxu-microsoft:
AFAIK, it only affect the clean-up stage when regen code. Though still better to keep if not too much trouble.

ChenTanyi:
Done for it. Not sure why these 6 files have old header.

PR #9803, opened 2020-04-02
Added support for timeout on BlobClient.upload

rickle-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
```diff
@@ -164,25 +165,58 @@ public void upload(InputStream data, long length, boolean overwrite) {
      * @param requestConditions {@link BlobRequestConditions}
      * @param timeout An optional timeout value beyond which a {@link RuntimeException} will be raised.
      * @param context Additional context that is passed through the Http pipeline during the service call.
+     * @deprecated
```
gapra-msft, line 168: Do we need to also deprecate in EncryptedBlobClient?

gapra-msft: I think it would also be helpful to link the method to use instead of this method as part of the deprecation message.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
```diff
@@ -164,25 +165,58 @@ public void upload(InputStream data, long length, boolean overwrite) {
...
      */
+    @Deprecated
     public void uploadWithResponse(InputStream data, long length, ParallelTransferOptions parallelTransferOptions,
```
alzimmermsft, line 171: Could it be possible to leave this API around and hook up taking `Context` in `BlobOutputStream`? I feel that would be the best path moving forwards as using `BlobOutputStream` is lacking functionality to support tracing.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
```diff
@@ -164,25 +165,58 @@ public void upload(InputStream data, long length, boolean overwrite) {
...
-        try {
-            StorageImplUtils.copyToOutputStream(data, length, blobOutputStream);
-            blobOutputStream.close();
-        } catch (IOException e) {
-            Throwable cause = e.getCause();
-            if (cause instanceof BlobStorageException) {
-                throw logger.logExceptionAsError((BlobStorageException) cause);
-            } else {
-                throw logger.logExceptionAsError(new UncheckedIOException(e));
+        Mono<Object> upload = Mono.fromCallable(() -> {
```
alzimmermsft, line 197: With this change should we just have this call into `BlockBlobClient`'s upload method?

alzimmermsft:
We can remove the `APIChange` label based on the latest changes as there is no longer any API changes happening.

PR #9805, opened 2020-04-02
Fix template build.

mitchdenny:
This PR addresses some clean-up that needed to be done to make sure that the template library was building successfully. It has been broken for a while.

/cc @chidozieononiwu


mitchdenny:
@chidozieononiwu this change is in now.

PR #9806, opened 2020-04-02
Added base avro parser package

gapra-msft:


sdk/storage/azure-storage-internal-avro/src/main/java/module-info.java
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module azure.storage.internal.avro {
```
JonathanGiles, line 4: This should be `com.azure.*`

pom.client.xml
```diff
@@ -522,6 +522,10 @@
                 <title>Azure Storage Files - Data Lake</title>
                 <packages>com.azure.storage.file.datalake*</packages>
               </group>
+              <group>
+                <title>Azure Storage Internal Avro</title>
+                <packages>com.azure.storage.internal.avro*</packages>
+              </group>
```
JonathanGiles, line 528: Don't add JavaDoc grouping here - add it to the list of excluded packages

sdk/storage/azure-storage-internal-avro/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+  <properties>
+    <jacoco.min.linecoverage>0.10</jacoco.min.linecoverage>
+    <jacoco.min.branchcoverage>0.10</jacoco.min.branchcoverage>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-storage-common</artifactId>
+      <version>12.6.0-beta.1</version> <!-- {x-version-update;com.azure:azure-storage-common;dependency} -->
```
JimSuplizio, line 44: There is one change that needs to be done here. This dependency tag needs to be **current** not **dependency**, this is why the Analyze step is failing. Items built within the same pipeline use the current versions of each other, not the dependency versions.

gapra-msft: Gotcha, thanks for catching that!

sdk/storage/azure-storage-internal-avro/src/main/java/module-info.java
gapra-msft: fixed

PR #9818, opened 2020-04-03
fix: add new line in file end and remove unused import for checkstyle

ChenTanyi:
script in https://github.com/ChenTanyi/work-script/blob/master/checkstyle.py


sdk/appservice/mgmt/src/main/java/com/azure/management/appservice/AppServiceCertificate.java
```diff
@@ -176,4 +176,4 @@
                 Creatable<AppServiceCertificate> {
         }
     }
-}
\ No newline at end of file
+}
```
xccc-msft, line 179: Can we avoid this kind of change?

weidongxu-microsoft: checkstyle

PR #9845, opened 2020-04-03
Refactor etag handling using ETagState

xseeseesee:
Fix [Azure/azure-libraries-for-java#1105](https://github.com/Azure/azure-libraries-for-java/issues/1105)

Main changes:
- Provide overloading for default unconditional request. E.g. adding `deleteById(id)` for existing `deleteById(id, etag)`. Similar for other delete/lock/extend operations if applicable.
- Make ETagState available for general services instead of only for DNS RP.
- Rename `withIfMatch` in Storage/BlobContainers/ImmutabilityPolicy by `withETagCheck` to align with etag style in DNS.
- Remove `withETagCheck` from Create phase. Support both `withETagCheck()`and `withETagCheck(etag)`. It's due to ImmutabilityPolicy service APIs only contains If-Match header.


sdk/dns/mgmt/src/main/java/com/azure/management/dns/implementation/DnsZoneImpl.java
```diff
@@ -394,34 +395,25 @@ public DnsZoneImpl withoutTxtRecordSet(String name, String eTag) {
 
     @Override
     public DnsZoneImpl withETagCheck() {
-        if (isInCreateMode()) {
-            this.dnsZoneETag = "*";
-            return this;
-        }
-        return this.withETagCheck(this.inner().etag());
+        this.eTagState.withImplicitETagCheckOnCreate();
+        this.eTagState.withImplicitETagCheckOnUpdate();
```
weidongxu-microsoft, line 399: Why set both? Will it affect line 413 when you call both regardless create or update?

xccc-msft: `withETagCheck()` could be available both Create phase and Update phase. So it's required to set both here. 

In Create phase, line#413 `self.inner().etag()` is NULL so it will send `If-Match: null` and `If-None-Match:*` during creation. In Update phase, it would send `If-Match: etag` and `If-None-Match: *`. This is the scenario applied by RecordSet in track 1. I tested it's working when sending `If-Match: etag` and `If-None-Match: *` for Update.

weidongxu-microsoft: The problem is why does `If-None-Match: *` work on UPDATE? The condition specified by it is that there being not any match, isn't it? Unless you can find doc elaborating it is acceptable, don't do it.
The test pass might be just service behavior, which might not work on another service.

xccc-msft: Agree. I just commit changes for this. Now it will only set one of them by verifying inner.id().

weidongxu-microsoft:
Also remember to add javadoc before merge.

PR #9847, opened 2020-04-03
fix checkstyle error message

weidongxu-microsoft:
In comment and in regex it is 20.

```
        // 'name' should not be empty, no Space allowed and the length should less than or equal to 20 characters
        Pattern serviceNamePattern = Pattern.compile("^[a-zA-Z0-9]{1,20}$");
```


weidongxu-microsoft:
BTW, does it matter if different ServiceInterface having same name?

Generated code could have `ResourceManagementClientResourceGroups`, `ResourceManagementClientDeploymentOperations`, `ResourceManagementClientOperations`, etc.
When truncated they all gives same name.

mssfang:
I think it matters. We should keep name of `ServiceInterface` varies. 
But i am 100% sure. Want confirmation from @JonathanGiles and @srnagar

weidongxu-microsoft:
@mssfang Thanks. If it does matter, we would try a better approach for shorten the name.

PR #9848, opened 2020-04-03
mgmt regen resources module for checkstyle fix

weidongxu-microsoft:


weidongxu-microsoft:
autorest.java PR https://github.com/Azure/autorest.java/pull/566

ChenTanyi:
What happen for the manual fix? Could it be done by generator?

weidongxu-microsoft:
@ChenTanyi 

Currently major thing is `build()` -> `buildClient()`, which I am still checking with Redmond whether they are OK with the change. Since it is only 1 per service, I figure it is not urgent.

The rest is either on conflict class always in resources (since it got 4 clients, instead of 1 for usual services), or on Features (which is strange). I will check Features later.

Generator is on `v4_fix_checkstyle1` branch. Not merged.

xccc-msft:
> @ChenTanyi
> 
> Currently major thing is `build()` -> `buildClient()`, which I am still checking with Redmond whether they are OK with the change. Since it is only 1 per service, I figure it is not urgent.
> 
> The rest is either on conflict class always in resources (since it got 4 clients, instead of 1 for usual services), or on Features (which is strange). I will check Features later.
> 
> Generator is on `v4_fix_checkstyle1` branch. Not merged.

Is this for consistency? Normally it should be `build()` for Builder class.

weidongxu-microsoft:
@xccc-msft 

checkstyle requires it be `buildClient` for class annotated with `@ServiceClientBuilder`.

xccc-msft:
> @xccc-msft
> 
> checkstyle requires it be `buildClient` for class annotated with `@ServiceClientBuilder`.
Got it. Yes, resources has more clients as it contains different RPs.

PR #9849, opened 2020-04-03
checkstyle ignore single word in comment

weidongxu-microsoft:
Quite often mgmt swagger would have this:
```
     * @param resourceId The fully qualified ID of the resource, including the resource name and resource type. Use the
     *     format,
     *     /subscriptions/{guid}/resourceGroups/{resource-group-name}/{resource-provider-namespace}/{resource-type}/{resource-name}.
```

IMO there is no good option to break this long URL, to fit into 120 length.

https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/sizes/LineLengthCheck.html

> To configure the check to ignore lines that begin with " * " code, followed by just one word, such as within a Javadoc comment:
> 
>  <module name="LineLength">
>    <property name="ignorePattern" value="^ *\* *[^ ]+$"/>
>  </module>

### DIFF
--- a/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml
+++ b/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle.xml
@@ -54,7 +54,8 @@ page at http://checkstyle.sourceforge.net/config.html -->
   <module name="LineLength">
     <property name="max" value="120"/>
     <!-- Allow codeSnippet to override max line length -->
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft:
Hello Jonathan @JonathanGiles 
Would you let me know your opinion on this checkstyle ignore case. Personally I think it is valid for ignore.

If I cannot ignore this, I would had to break the url or delete it. Either approach would only convert a ugly javadoc to a wrong javadoc.

PR #9854, opened 2020-04-03
Make README and CHANGELOG changes for new guidelines

sima-zhu:


eng/versioning/version_client.txt
```diff
@@ -23,7 +23,7 @@ com.azure:azure-identity;1.0.4;1.1.0-beta.3
 com.azure:azure-messaging-eventhubs;5.0.2;5.1.0-beta.1
 com.azure:azure-messaging-eventhubs-checkpointstore-blob;1.0.2;1.1.0-beta.1
 com.azure:azure-messaging-servicebus;1.0.0-beta.1;1.0.0-beta.1
-com.azure:azure-search-documents;11.0.0-beta.1;11.0.0-beta.2
+com.azure:azure-search-documents;11.0.0-beta.1;1.0.0-beta.2
```
JimSuplizio, line 26: @sima-zhu I'm curious about these versions. com.azure:azure-search-documents is a renamed com.azure:azure-search and [com.azure:azure-search was released as 11.0.0-beta.1](https://search.maven.org/search?q=g:com.azure%20AND%20a:azure-search) however there is no released com.azure:azure-search-documents so the current and dependency versions should be same. (Obviously nothing in the tree currently uses the dependency version of this otherwise we would have seen compilation errors.) My recommendation here is to set both the current and dependency versions to 1.0.0-beta.1.

sdk/search/azure-search-documents/CHANGELOG.md
```diff
@@ -1,10 +1,19 @@
...
+- Moved `AzureKeyCredentialPolicy` to Azure Core.
 - Fixed a bug where the Date header wouldn't be updated with a new value on request retry.
+- Changed the field type of `CustomAnalyzer`.
+- Made `RangeFacetResult` and `ValueFacetResult` object strongly typed.
+- Added helper function for IndexBatchException. 
+- Added ScoringParameter class.
+- Refactored some boolean field getter.
+- Made `IndexDocumentsBatch` APIs plurality.
 
 ## 11.0.0-beta.1 (2020-03-10)
```
JimSuplizio, line 18: @alzimmermsft what should happen here, with the versions in the changelog, since this library has been renamed? Also, if this is being re-versioned and the previous release was only a beta should we even keep the 11* version stuff in here?

eng/versioning/version_client.txt
sima-zhu: The version name aligns with all languages. 
@tg-msft might have further explanation on the decision.

tg-msft: This was something we discussed in the Search crew and decided to go with beta.2 to indicate to customers that it's the second preview of this library.  The hope was to make it clearer to anyone who goes digging through history that something came before this.  Sima is correct that all the other languages are aligning to Preview.2, beta.2, b2, etc.

sdk/search/azure-search-documents/CHANGELOG.md
sima-zhu: Synced offline. Leave the change log as it is so that customer is aware of the version history.

sima-zhu: Added notes to make it clear.

PR #9856, opened 2020-04-03
handling < and > the 'javadoc' way

scbedd:
We removed encoding of `&amp`. Updated the encoding for `<` and `>`. Real html isn't a javadoc thing.


eng/pipelines/scripts/inject_codesnippets.py
```diff
@@ -21,8 +21,8 @@
 HTML_ESCAPE_TABLE = {
     "&": "&amp;",
     '"': "&quot;",
-    ">": "&gt;",
-    "<": "&lt;",
+    ">": "&amp;gt;",
```
weshaggard, line 24: Would the other entries need this double escaping as well?

scbedd: Nope. This ended up being a javadoc intricacy/bug. 

I'm surrounding with `@code` and NOT replacing it now.

PR #9857, opened 2020-04-03
Support token offset for cached tokens

jianghaolu:
When we query the token cache for tokens, our mechanism of preemptive token expiration is not honored. Thus we have to check the expiration ourselves and force the refresh.


jianghaolu:
This is being redesigned. Closing this PR.

PR #9858, opened 2020-04-03
Cosmos CFP: fix for handling partition splits or partition not found

milismsft:
Bug fix when handling partition splits or partition not found. The cause of the bug is a typo in the call to log the current partition information such as the partition ID.
Bug fix when syncronizing lease updates across different threads.
Add test case for handling partition splits.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
@@ -464,11 +464,11 @@ public class ChangeFeedProcessorTest extends TestSuiteBase {
         assertThat(remainingWork >= 0).as("Failed to receive all the feed documents").isTrue();
     }
 
-    @BeforeMethod(groups = { "emulator" }, timeOut = 2 * SETUP_TIMEOUT, alwaysRun = true)
+    @BeforeMethod(groups = { "emulator", "simple" }, timeOut = 2 * SETUP_TIMEOUT, alwaysRun = true)
      public void beforeMethod() {
      }
 
-    @BeforeClass(groups = { "emulator" }, timeOut = SETUP_TIMEOUT, alwaysRun = true)
+    @BeforeClass(groups = { "emulator", "simple" }, timeOut = SETUP_TIMEOUT, alwaysRun = true)
     public void before_ChangeFeedProcessorTest() {
         client = getClientBuilder().buildAsyncClient();
         createdDatabase = getSharedCosmosDatabase(client);
@@ -495,11 +495,11 @@ public class ChangeFeedProcessorTest extends TestSuiteBase {
 //        createdDatabase = createDatabase(client, databaseId);
     }
 
-    @AfterMethod(groups = { "emulator" }, timeOut = 3 * SHUTDOWN_TIMEOUT, alwaysRun = true)
+    @AfterMethod(groups = { "emulator", "simple" }, timeOut = 3 * SHUTDOWN_TIMEOUT, alwaysRun = true)
     public void afterMethod() {
     }
 
-    @AfterClass(groups = { "emulator" }, timeOut = 2 * SHUTDOWN_TIMEOUT, alwaysRun = true)
+    @AfterClass(groups = { "emulator", "simple" }, timeOut = 2 * SHUTDOWN_TIMEOUT, alwaysRun = true)
     public void afterClass() {
 //        try {
 //            client.readAllDatabases()


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
```diff
@@ -63,7 +63,8 @@ public ServiceItemLease withId(String id) {
         return this;
     }
 
-    @JsonIgnore
+//    @JsonIgnore
+    @JsonProperty("_etag")
```
moderakh, line 67: when a class is annotated with jackson annotation, jackson uses reflection for serialization/deserialization to json.

reflection is expensive and that being said, If there are pre existing classes in the code which needs to be serialized/deserialized to json we should ideally use a jackson serializer rather than annotating the class with jackson annotation.

(not necessarily in the scope of this PR, but something to be addressed outside of this PR).

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
```diff
@@ -332,10 +315,136 @@ public void staledLeaseAcquiring() {
...
+            })
+            .doOnSuccess(aVoid -> {
+                // Wait for the feed processor to receive and process the first batch of documents.
+                waitToReceiveDocuments(2 * CHANGE_FEED_PROCESSOR_TIMEOUT, FEED_COUNT);
+            })
+            .then(
+                // increase throughput to force a single partition collection to go through a split
+                createdFeedCollectionForSplit.readProvisionedThroughput().subscribeOn(Schedulers.elastic())
+                    .flatMap(currentThroughput ->
+                        createdFeedCollectionForSplit.replaceProvisionedThroughput(FEED_COLLECTION_THROUGHPUT).subscribeOn(Schedulers.elastic())
```
moderakh, line 351: partition split normally takes 10-20 min to complete. how are we ensuring that the test wait long enough to observe a partition split?

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
```diff
@@ -332,10 +315,136 @@ public void staledLeaseAcquiring() {
...
+                    )
+                .then()
+            )
+            .subscribe();
+
+        // Wait for the feed processor to receive and process the first batch of documents and apply throughput change.
+        try {
+            Thread.sleep(4 * CHANGE_FEED_PROCESSOR_TIMEOUT);
+        } catch (InterruptedException e) {
+            log.error(e.getMessage());
```
moderakh, line 361: don't catch `InterruptedException` if that's thrown, we can fail the test.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
```diff
@@ -332,10 +315,136 @@ public void staledLeaseAcquiring() {
...
+                    throw new RuntimeException("Throughput change is not done.");
+                }
+                return count;
+            })
+            // this will timeout approximately after 3 minutes
+            .retry(40, throwable -> {
+                try {
+                    log.warn("Retrying...");
+                    Thread.sleep(CHANGE_FEED_PROCESSOR_TIMEOUT);
+                } catch (InterruptedException e) {
```
moderakh, line 390: don't catch `InterruptedException` if that's thrown, we can fail the test.

moderakh: here and elsewhere please. we don't need to catch this, let it get thrown

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
milismsft: n this case the "_etag" field was not passed through the serialization and that resulted in "IF MATCH" condition to fail when updating the lease document.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
milismsft: There's a check later to ensure that we see at least 2 partitions. In regular case it takes couple minutes (not 10 minutes) to detect the split.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
milismsft: The try/catch is needed to avoid compilation errors such as: "Error:(440, 29) java: unreported exception java.lang.InterruptedException; must be caught or declared to be thrown"

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
milismsft: We need to in this case, in order to avoid compilation errors...

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
moderakh: The ask is to add something similar to this:
https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/routing/PartitionKeyInternal.java#L35-L36

The internal types should not rely on reflection for serialization/deserialization.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
moderakh: why can't we add exception to the test method signature?

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
moderakh: makes sense here.
so please use this instead to translate checked to unchecked exception.
https://projectreactor.io/docs/core/release/api/reactor/core/Exceptions.html#propagate-java.lang.Throwable-

if `InterruptedException` we should error out.

milismsft: Good suggestion, but it goes way beyond what we need to address at this time which is to release a fix for the CFP ASAP.

Feel free to open an issue and assign it to me to refactor these tests and remove the handling of the exception.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
milismsft: replied on a similar comment below...

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
moderakh: @milismsft I have already signed off on v3, and I agree with you on v3 as this is hotfix for v3 it should go out as soon as possible.

on v4 we should do the right thing (this is not in prod yet) rather than introducing technical debt.

milismsft: V4 latest release is currently used by couple customers a preview.

milismsft: I'll deal with the debt later; the changes you're requesting go beyond what this PR/work item should address. IMHO et's keep things simple enough in case we need to review this later and use a different PR to address  other things.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
milismsft: Fixed. We only need a serializer here, the de-serialization is done at the Document/CosmosItemProperties level and that instance it is later used to create a Lease instance.

moderakh: If you have a serializer you need a deserializer too. 

Document/CosmosItemProperties should call to the custom deserializer for deserialization. no?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
```diff
@@ -20,6 +22,7 @@
 /**
  * Document service lease.
  */
+@JsonSerialize(using = ServiceItemLease.ServiceItemLeaseJsonSerializer.class)
```
moderakh, line 25: Please add unit test scoped to serialization.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
```diff
@@ -181,34 +182,17 @@ public ServiceItemLease withTimestamp(ZonedDateTime timestamp) {
         return this;
     }
 
-    @JsonIgnore
+//    @JsonIgnore
```
moderakh, line 185: stale code. remove

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
```diff
@@ -181,34 +182,17 @@ public ServiceItemLease withTimestamp(ZonedDateTime timestamp) {
...
     public String getExplicitTimestamp() {
         return this.timestamp;
     }
 
-    @JsonIgnore
+//    @JsonIgnore
```
moderakh, line 190: ditto

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
```diff
@@ -236,4 +235,32 @@ public String toString() {
...
+        @Override
+        public void serialize(ServiceItemLease lease, JsonGenerator writer, SerializerProvider serializerProvider) {
+            try {
+                writer.writeStartObject();
+                writer.writeStringField(Constants.Properties.ID, lease.getId());
+                writer.writeStringField(Constants.Properties.E_TAG, lease.getETag());
+                writer.writeStringField("LeaseToken", lease.getLeaseToken());
+                writer.writeStringField("ContinuationToken", lease.getContinuationToken());
+                writer.writeStringField("timestamp", lease.getTimestamp());
+                writer.writeStringField("Owner", lease.getOwner());
```
moderakh, line 259: LeaseToken, ContinuationToken, Owner are Camel case, but timestamp, id, etag are not is that expected?

also for Owner, timestamp, ContinuationToken, LeaseToken we probably should define constants.

milismsft: The current naming convention follows .Net and the content of a lease document as written from that code. The idea behind this design is that you can have CFP instances resuming work from both platforms, Java and .NET.
The ServiceItemLease is internal only (not exposed to the user); for the easy of debugging keeping the same names for members and the content of the lease document as written in Cosmos is actually beneficial.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
```diff
@@ -424,6 +427,46 @@ public void readFeedDocumentsAfterSplit() {
         receivedDocuments.clear();
     }
 
+    @Test(groups = { "simple" }, timeOut = CHANGE_FEED_PROCESSOR_TIMEOUT)
+    public void testServiceItemLeaseSerialization() {
```
moderakh, line 431: thank you for the test. as a code style rule we don't add "test" as prefix.

Please drop the "test" prefix to be consistent with other tests.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
milismsft: Test added.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
```diff
@@ -424,6 +427,46 @@ public void readFeedDocumentsAfterSplit() {
...
+                String.format("%s%s%s",
+                    "{\"id\":\"id1\",\"_etag\":\"etag1\",\"LeaseToken\":\"1\",\"ContinuationToken\":\"12\",\"timestamp\":\"",
+                    timeNowValue,
+                    "\",\"Owner\":\"Owner1\"}"));
+            assertThat(mapper.writeValueAsString(lease2)).isEqualTo(
+                String.format("%s%s%s",
+                    "{\"id\":\"id2\",\"_etag\":\"etag2\",\"LeaseToken\":\"2\",\"ContinuationToken\":\"22\",\"timestamp\":\"",
+                    timeNowValue,
+                    "\",\"Owner\":null}"));
+        } catch(Exception ex) {
```
moderakh, line 465: please drop the try/catch block here. if there is a failure the failure should get thrown and the exception needs to be added to the test method signature.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
milismsft: removed

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
milismsft: removed

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
milismsft: done

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
milismsft: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
```diff
@@ -236,4 +234,32 @@ public String toString() {
             this.getTimestamp(),
             UNIX_START_TIME.plusSeconds(Long.parseLong(this.getTs())));
     }
+
+    @SuppressWarnings("serial")
+    static final class ServiceItemLeaseJsonSerializer extends StdSerializer<ServiceItemLease> {
```
moderakh, line 239: we should add a deserializer as well. Document/CosmosItemProperties can use the custom deserializer.

You can take a look at `PartitionKeyInternal.java` as example.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
```diff
@@ -332,10 +325,172 @@ public void staledLeaseAcquiring() {
...
+        // Wait for the feed processor to shutdown.
+        try {
+            Thread.sleep(CHANGE_FEED_PROCESSOR_TIMEOUT);
+        } catch (InterruptedException e) {
+            log.error(e.getMessage());
+        }
+        receivedDocuments.clear();
+    }
+
+    @Test(groups = { "simple" }, timeOut = CHANGE_FEED_PROCESSOR_TIMEOUT)
```
moderakh, line 430: "simple" test group is only invoked as integration test in the presence of a cosmosdb endpoint.
 
as this is unit test, not integration test, the group needs to be "unit". 

also the unit tests should go in a separate class, e.g.,  please move the test to  the new class ServiceItemLeaseTest.java

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
milismsft: Won't fix.

moderakh: > Won't fix.

why?

milismsft: already explained "why"; you're asking to add dead code because there's no direct Json deserialization that is needed for ServiceItemLease at this time. The "deserialization" is done via fromDocument() method call.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
milismsft: Won't fix.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/changefeed/ServiceItemLease.java
moderakh: the explanation on fromDocument() makes sense. so that acts as the deserializer, that's what needed. thanks.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
moderakh: > Won't fix.

why?

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/ChangeFeedProcessorTest.java
```diff
@@ -360,27 +493,18 @@ public void before_ChangeFeedProcessorTest() {
...
-        safeDeleteCollection(createdLeaseCollection);
-
-        // Allow some time for the collections and the database to be deleted before exiting.
-        try {
-            Thread.sleep(500);
-        } catch (Exception e){ }
     }
 
     @AfterClass(groups = { "emulator" }, timeOut = 2 * SHUTDOWN_TIMEOUT, alwaysRun = true)
     public void afterClass() {
```
moderakh, line 503: I suspect you have client leak in your test class. this might cause CI test issues later.

the afterClass is annotated as part of emulator test group. But your partition split test is in a different test group. Due to that the afterClass may not get invoked.

Our test pattern is to not mix different test groups in the same class for simplicity.
Different test group go to different test classes, otherwise you need to be very cautious in the cleaning logic.

moderakh: @kushagraThapar FYI

milismsft: Fixed

PR #9863, opened 2020-04-04
Support for receiving events in batches in processor client

### LINKED ISSUE #9878: [Event Hubs] Add support for receiving events in batches.
Event Processor supports receiving single events today. Add a new API in `EventProcessorClientBuilder`  that allows for receiving events in batches.

srnagar:
This PR is to add support for receiving events in batches in event processor.

Fixes #9878

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionPumpManager.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionPumpManager.java
@@ -207,8 +207,8 @@ class PartitionPumpManager {
     }
 
     private void processEvent(PartitionContext partitionContext, PartitionProcessor partitionProcessor,
-        EventHubConsumerAsyncClient eventHubConsumer, PartitionEvent partitionEvent) {
-        EventData eventData = partitionEvent.getData();
+        EventHubConsumerAsyncClient eventHubConsumer, EventContext eventContext) {
+        EventData eventData = eventContext.getEventData();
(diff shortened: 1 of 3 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionPumpManager.java
```diff
@@ -146,17 +160,38 @@ void startPartitionPump(PartitionOwnership claimedOwnership, Checkpoint checkpoi
...
-                .subscribe(partitionEvent -> processEvent(partitionContext, partitionProcessor, eventHubConsumer,
-                    partitionEvent),
-                    /* EventHubConsumer receive() returned an error */
-                    ex -> handleError(claimedOwnership, eventHubConsumer, partitionProcessor, ex, partitionContext),
-                    () -> {
-                        partitionProcessor.close(new CloseContext(partitionContext,
-                            CloseReason.EVENT_PROCESSOR_SHUTDOWN));
-                        cleanup(claimedOwnership, eventHubConsumer);
-                    });
+            if (maxWaitTime != null) {
```
conniey, line 163: What happens if they set Duration.ZERO? Would we try the window timeout approach?

srnagar:
@conniey added code snippets and samples

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/PartitionPumpManager.java
srnagar: Added validation to check that duration is not zero and also switched to window timeout as that seems like the more suited operator here.

PR #9864, opened 2020-04-04
Cosmos CFP: port V4 fix for handling partition splits 

milismsft:
Bug fix when handling partition splits or partition not found. The cause of the bug is a typo in the call to log the current partition information such as the partition ID.
Bug fix when synchronizing lease updates across different threads.

### DIFF
--- a/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/ServiceItemLease.java
+++ b/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/ServiceItemLease.java
@@ -62,7 +62,7 @@ public class ServiceItemLease implements Lease {
         return this;
     }
 
-    @JsonIgnore
+    @JsonProperty("_etag")
     public String getEtag() {
         return this._etag;
     }
@@ -132,6 +132,10 @@ public class ServiceItemLease implements Lease {
         this.withTimestamp(date.toInstant().atZone(zoneId));
     }
 
+    public void setTimestamp(String timestamp) {
+        this.timestamp = timestamp;
+    }
+
     @Override
     public void setId(String id) {
         this.withId(id);
@@ -191,7 +195,7 @@ public class ServiceItemLease implements Lease {
         return this.getEtag();
     }
 
-    public static ServiceItemLease fromDocument(Document document) {
+    public static ServiceItemLease fromDocument(CosmosItemProperties document) {
         ServiceItemLease lease = new ServiceItemLease()
             .withId(document.id())
             .withEtag(document.etag())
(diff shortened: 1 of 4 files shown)


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/ServiceItemLease.java
```diff
@@ -62,7 +62,7 @@ public ServiceItemLease withId(String id) {
         return this;
     }
 
-    @JsonIgnore
+    @JsonProperty("_etag")
```
moderakh, line 65: when a class is annotated with jackson annotation, jackson uses reflection for serialization/deserialization to json.

reflection is expensive and that being said, If there are pre existing classes in the code which needs to be serialized/deserialized to json we should ideally use a jackson serializer rather than annotating the class with jackson annotation.

(not necessarily in the scope of this PR, but something to be addressed outside of this PR).

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
```diff
@@ -124,18 +124,21 @@ public PartitionProcessorImpl(ChangeFeedObserver observer, ChangeFeedContextClie
...
+                        this.settings.getPartitionKeyRangeId(), Thread.currentThread().getId(), clientException);
                     StatusCodeErrorType docDbError = ExceptionClassifier.classifyClientException(clientException);
 
                     switch (docDbError) {
                         case PARTITION_NOT_FOUND: {
                             this.resultException = new PartitionNotFoundException("Partition not found.", this.lastContinuation);
+                            break;
                         }
                         case PARTITION_SPLIT: {
                             this.resultException = new PartitionSplitException("Partition split.", this.lastContinuation);
```
moderakh, line 136: @milismsft we need to have explicit automated tests for partition split.

We have some tests for partition split that you can use as a sample:
DocumentProducerTest (unit test)
ReadMyWritesConsistencyTest (integration test)

milismsft: I've added a specific test for splits in V4 respective change. The challenge is that public emulator does not support splitting so the test must be run against the cloud service.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/ServiceItemLease.java
milismsft: In this case the "_etag" field was not passed through the serialization and that resulted in "IF MATCH" condition to fail when updating the lease document.

sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/changefeed/implementation/PartitionProcessorImpl.java
moderakh: if we add tests on v4 that should be good enough. thanks.

PR #9865, opened 2020-04-04
SB Track2 SyncReceiverClient: add maxWaitTime in receive()

hemanttanwar:
Use Duration maximumWaitTime in receive() functionality

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
@@ -16,8 +16,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import static com.azure.core.util.FluxUtil.monoError;
-
 /**
  * A <b>synchronous</b> receiver responsible for receiving {@link ServiceBusReceivedMessage} from a specific queue or
  * topic on Azure Service Bus.
--- a/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientTest.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientTest.java
@@ -3,10 +3,8 @@
 
 package com.azure.messaging.servicebus;
 
-import com.azure.core.amqp.implementation.AmqpSendLink;
 import com.azure.core.util.IterableStream;
 import com.azure.core.util.logging.ClientLogger;
-import com.azure.messaging.servicebus.models.CreateBatchOptions;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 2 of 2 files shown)


sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -28,6 +28,8 @@
  */
 @ServiceClient(builder = ServiceBusClientBuilder.class)
 public class ServiceBusReceiverClient implements AutoCloseable {
+    private static final Duration DEFAULT_RECEIVE_WAIT_TIME = Duration.ofMinutes(1);
```
hemanttanwar, line 31: This was discussed during review that instead of using tryTimeout, we should be using some default value 1 minute or less.

conniey: The default AmqpRetryOptions has the timeout already set. This isn't necessary. Or if they pass their own default tryTimeout.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -250,18 +252,47 @@ public ServiceBusReceivedMessage peekAt(long sequenceNumber) {
...
+     * The receive operation will wait for a default 1 minute for receiving a message before it times out. You can it
+     * override by using {@link #receive(int, Duration)}.
      *
      * @param maxMessages The maximum number of messages to receive.
      * @return An {@link IterableStream} of at most {@code maxMessages} messages from the Service Bus entity.
+     *
+     * @throws IllegalArgumentException if {@code maxMessages} is zero or a negative value.
      */
     public IterableStream<ServiceBusReceivedMessage> receive(int maxMessages) {
         if (maxMessages <= 0) {
```
conniey, line 265: To save duplicated logic, this method should call the maximal overload.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -250,18 +252,47 @@ public ServiceBusReceivedMessage peekAt(long sequenceNumber) {
...
+     * @param maxWaitTime The time the client waits for receiving a message before it times out.
+     * @return An {@link IterableStream} of at most {@code maxMessages} messages from the Service Bus entity.
+     *
+     * @throws IllegalArgumentException if {@code maxMessages} or {@code maxWaitTime} is zero or a negative value.
+     */
+    public IterableStream<ServiceBusReceivedMessage> receive(int maxMessages, Duration maxWaitTime) {
+        if (maxMessages <= 0) {
+            throw logger.logExceptionAsError(new IllegalArgumentException(
+                "'maxMessages' cannot be less than or equal to 0. maxMessages: " + maxMessages));
+        } else if (maxWaitTime.isNegative() || maxWaitTime.isZero()) {
```
conniey, line 289: Possible NullPointerException if maxWaitTime is null

PR #9884, opened 2020-04-06
Cleanup Enforcer Configurations

alzimmermsft:
This PR updates the Maven Enforcer plugin's configuration in `pom.client.xml` to allow for easier overriding by children POMs. The configuration was moved outside of the execution it was in to become the global configuration for the plugin, this allows for children POMs to set the specific configurations they need without knowing which execution the configuration will be ran in. In the future we should consider doing this in places where we have many children POMs that override an execution configuration or when we have multiple executions for the same plugin which are similar as the global configuration can be overridden by executions as well.


common/perf-test-core/pom.xml
```diff
@@ -37,7 +37,8 @@
       <plugin>
         <groupId>com.github.spotbugs</groupId>
         <artifactId>spotbugs-maven-plugin</artifactId>
-        <version>3.1.12.2</version> <!-- {x-version-update;com.github.spotbugs:spotbugs-maven-plugin;external_dependency} -->
+        <version>3.1.12.2
+        </version> <!-- {x-version-update;com.github.spotbugs:spotbugs-maven-plugin;external_dependency} -->
```
srnagar, line 41: This change can be reverted to keep the version tags in the same line.

common/perf-test-core/pom.xml
```diff
@@ -46,37 +47,22 @@
...
-                    <exclude>*:*:*:*:runtime</exclude>
-                    <exclude>*:*:*:*:compile</exclude>
-                    <exclude>*:*:*:*:provided</exclude>
-                  </excludes>
-                  <includes>
-                    <include>com.azure:*</include>
-                    <include>org.slf4j</include>
-                    <include>com.fasterxml.jackson.*</include>
+        <version>3.0.0-M3
+        </version> <!-- {x-version-update;org.apache.maven.plugins:maven-enforcer-plugin;external_dependency} -->
```
srnagar, line 51: This change can be reverted to keep the version tags in the same line.

PR #9909, opened 2020-04-07
Generate dependency whitelist report

srnagar:
This PR is to include a new maven plugin that generates a report of all whitelisted dependencies used in Track 2 SDKs.

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -438,6 +438,8 @@
                 <param>com.azure:azure-cosmos-benchmark</param>
                 <param>com.azure:azure-core-test</param>
                 <param>com.azure:azure-e2e</param>
+                <param>com.azure:azure-storage-perf</param>
+                <param>com.azure:perf-test-core</param>
               </ignoredGA>
             </configuration>
           </execution>


srnagar:
azp run java-aggregate-report

pom.client.xml
```diff
@@ -419,6 +419,31 @@
...
+              <goal>report</goal>
+            </goals>
+            <configuration>
+              <baseDir>.</baseDir>
+              <reportFile>dependency-whitelist.json</reportFile>
+              <ignoredGA>
+                <param>com.azure:azure-cosmos-examples</param>
+                <param>com.azure:azure-cosmos-benchmark</param>
+                <param>com.azure:azure-core-test</param>
+                <param>com.azure:azure-e2e</param>
```
JimSuplizio, line 440: You're going to want to add com.azure:azure-storage-perf and com.azure:perf-test-core to this list. Right now, they're not being built as part of any pipeline but when I do the versioning work for the whitelist includes they will be.

srnagar: Added both these modules to ignored list.

PR #9911, opened 2020-04-07
SQL: fix checkstyle

ChenTanyi:


sdk/sql/mgmt/src/main/java/com/azure/management/sql/AdministratorType.java
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
yungezz, line 1: what's this file from? new generaton?

ChenTanyi: Yes, all header with `Code generated ...` is from generator.

PR #9912, opened 2020-04-07
Update whitelist includes

JimSuplizio:
The first commit is relatively basic and consists of the following:
1. Update the whitelist includes to all be of the form <groupdId>:<artifactId> removing any wildcards and any unnecessary includes.
2. Add com.azure:perf-test-core to the core\ci.yml's AdditionalModules. There was currently nothing actually building this. While this module isn't under core, its only non-external dependency is azure-core-http-netty so it makes sense to build this with core.
3. Add azure-storage-perf to the storage\ci.yml's AdditionalModules. Again, there was nothing ever building this. This is a storage module used for perf testing so it makes sense to build this with storage.

The second commit consists of
1. Lock the whitelist includes to a specific version. Note: while the syntax of the version, itself, looks a little odd it needs the braces. [1.2.3] locks to 1.2.3. Without the braces, 1.2.3 would be the equivalent of [1.2.3,) which is that version or higher.
2. Add tooling to the update_versions.py script to handle the include entries. Since these are external_dependencies they'll only be updated when the update type is external_dependency or all.
3. The pom_file_version_scanner will scan the entries and tags for both existence and correctness.

### DIFF
--- a/eng/versioning/pom_file_version_scanner.ps1
+++ b/eng/versioning/pom_file_version_scanner.ps1
@@ -524,7 +524,7 @@ Get-ChildItem -Path $Path -Filter pom*.xml -Recurse -File | ForEach-Object {
                     $script:FoundError = $true
                     Write-Error-With-Color "Error: <include> version update tag for $($includeNode.InnerText) should be <!-- {x-include-update;$($groupId):$($artifactId);external_dependency} -->"
                 }
-                else 
+                else
                 {
                     # verify that the version is formatted correctly
                     if (!$version.StartsWith("[") -or !$version.EndsWith("]"))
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core-http-netty/pom.xml
```diff
@@ -193,16 +193,14 @@
...
-                    <include>io.projectreactor.netty</include>
-
                     <include>io.netty:netty-buffer</include>
                     <include>io.netty:netty-codec-http</include>
                     <include>io.netty:netty-codec-http2</include>
                     <include>io.netty:netty-handler</include>
                     <include>io.netty:netty-handler-proxy</include>
                     <include>io.netty:netty-transport-native-unix-common</include>
                     <include>io.netty:netty-transport-native-epoll</include>
+                    <include>io.projectreactor.netty:reactor-netty </include>
```
alzimmermsft, line 203: nit
```suggestion
                    <include>io.projectreactor.netty:reactor-netty</include>
```

JimSuplizio: That space is gone.

JimSuplizio:
It's worth noting that the storage perf checkstyle suppression are being fixed by https://github.com/Azure/azure-sdk-for-java/pull/10030 (thanks @alzimmermsft ) and rebase with those changes once that is checked in.

PR #9918, opened 2020-04-07
Make ActiveTokenManager tests more reliable.

conniey:
Use StepVerifier.withVirtualTime to make tests more consistent.


sdk/core/azure-core-amqp/src/test/java/com/azure/core/amqp/implementation/ActiveClientTokenManagerTest.java
```diff
@@ -22,18 +24,29 @@
...
     private static final String SCOPES = "scopes-test";
-    private static final Duration TIMEOUT = Duration.ofSeconds(4);
+    private static final Duration DEFAULT_DURATION = Duration.ofSeconds(20);
 
     @Mock
     private ClaimsBasedSecurityNode cbsNode;
 
+    @BeforeAll
+    static void beforeAll() {
+        StepVerifier.setDefaultTimeout(Duration.ofSeconds(30));
```
srnagar, line 42: Does setting the default timeout change the timeout for all tests running outside this class too?

conniey: No. I have an `AfterAll` notation to reset the timeout.

PR #9920, opened 2020-04-07
Better Handling for Blob Batch Tests in Playback

alzimmermsft:
This PR updates the Blob Batch tests where some operations fail, but not all the operations fail. The underlying request generation system is non-deterministic as it is able to generated multiple operations at once which may result in a different operation order each time the test is ran. This work properly when running against the live service as the request and response are tied together using a `Content-ID` marker but running the tests in playback this is an issue as it isn't capable of handling variance in ordering.

Tests where some operations are expected to fail were updated to use an order invariant check to determine correctness. All responses are checked for either having the correct response code or throwing the expected exception, they are then aggregated to validate that the number of successful operations matched what was expected. When tests are running against the live service the tests will continue to check that the operation succeeded or failed as expected in a strict ordering.

### DIFF
--- a/sdk/storage/azure-storage-blob-batch/src/test/java/com/azure/storage/blob/batch/BatchAPITest.groovy
+++ b/sdk/storage/azure-storage-blob-batch/src/test/java/com/azure/storage/blob/batch/BatchAPITest.groovy
@@ -38,10 +38,10 @@ class BatchAPITest extends APISpec {
         try {
             def statusCode = response.getStatusCode()
             assert statusCode == expectedStatusCode
-            return true
+            return 1
         } catch (def exception) {
             assert exception instanceof BlobStorageException
-            return false
+            return 0
         }
     }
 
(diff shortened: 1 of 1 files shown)


alzimmermsft:
@gapra-msft @rickle-msft, could you take a look at this PR, should resolve the flaky Blob Batch test failures happening during PR validation.

sdk/storage/azure-storage-blob-batch/src/test/java/com/azure/storage/blob/batch/BatchAPITest.groovy
```diff
@@ -128,14 +152,19 @@ class BatchAPITest extends APISpec {
...
+        notThrown(BlobBatchStorageException)
 
-        when:
-        response2.getStatusCode()
-
-        then:
-        thrown(BlobStorageException)
+        // In PLAYBACK check responses in an order invariant fashion.
+        if (testMode == TestMode.PLAYBACK) {
+            assertExpectedOrException(response1, 200) || assertExpectedOrException(response2, 200)
```
sima-zhu, line 159: If there is definitely one will fail, it is better to check fail at &.
```
!assertExpectedOrException(response1, 200) && assertExpectedOrException(response2, 200)
```

alzimmermsft: The code would need to be the following, which is equivalent to what is already being used.

```java
!(assert... && assert...)
```

To be more pedantic on validation I could explicitly check each of these for being `true` and asserting the count of trues to be one. Another option is having these return `0` for failure and `1` for success and adding and checking it equals `1`.

alzimmermsft: Updated to doing a count validation on successful responses.

PR #9930, opened 2020-04-07
Loosen README Title Requirements

alzimmermsft:
This PR loosens the title requirements for README titles by allowing additional library modifiers. This adds allowance for `plugin` and `shared` to go along with the current `client`, valid optional are now `client library`, `plugin library`, and `shared library`.


jianghaolu:
To me a plugin is a software add-on that is installed on a program, enhancing its capabilities, which means a plugin should not be required for a system to work. Like plugins are not required for maven, or IntelliJ IDE work. An HTTP client module is required for any SDK to work, it doesn't feel like a "plugin" to me.

As to identity, I'm okay but not thrilled with it being a "plugin" because it's technically not a must to use an SDK as one can always provide a token directly. However even though it's not a technical must to use the identity library it is a "logical must". If one can acquire a token and refresh it on his/her own without the need of azure-identity library, why are they using our other SDKs anyway?

Issue #9802, opened 2020-04-02
[BUG] Cosmos SDK 3.7.1 io.netty.util.IllegalReferenceCountException


allenhumphreys:
**Describe the bug**
Sporadically we're receiving exceptions from Cosmos due to an illegal referent count on a netty byte buffer. We are using direct connectivity.

***Exception or Stack Trace***
<details><summary>Stack Trace</summary>
<p>

```
io.netty.util.IllegalReferenceCountException: refCnt: 0
	at io.netty.buffer.AbstractByteBuf.ensureAccessible(AbstractByteBuf.java:1489)
	at io.netty.buffer.AbstractByteBuf.slice(AbstractByteBuf.java:1221)
	at io.netty.buffer.AbstractByteBuf.slice(AbstractByteBuf.java:1211)
	at io.netty.buffer.Unpooled.wrappedBuffer(Unpooled.java:229)
	at reactor.core.publisher.FluxMap$MapSubscriber.onNext(FluxMap.java:100)
	at jdk.internal.reflect.GeneratedMethodAccessor99.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at <redacted>.common.logging.hooks.CoreSubscriberProxyInvocationHandler.wrappedExceptionInvoke(CoreSubscriberProxyInvocationHandler.java:61)
	at <redacted>.common.logging.hooks.CoreSubscriberProxyInvocationHandler.lambda$handleInvocation$0(CoreSubscriberProxyInvocationHandler.java:48)
	at <redacted>.common.logging.hooks.SubscriberProxyingInterceptor$proxySubscribe$proxy$1.accept(SubscriberProxyingInterceptor.kt:25)
	at <redacted>.common.logging.hooks.SubscriberProxyingInterceptor$proxySubscribe$proxy$1.accept(SubscriberProxyingInterceptor.kt:11)
	at <redacted>.common.logging.hooks.CoreSubscriberProxyInvocationHandler.handleInvocation(CoreSubscriberProxyInvocationHandler.java:45)
	at com.google.common.reflect.AbstractInvocationHandler.invoke(AbstractInvocationHandler.java:86)
	at reactor.core.publisher.$Proxy217.onNext(Unknown Source)
[...]

PR #9934, opened 2020-04-07
Fixed re-wrapping of ByteBuf when sending request body to reactor netty

kushagraThapar:
No need to re-wrap the byteBuf as we already wrap it in RxGatewayStoreModel.java while creating request: https://github.com/Azure/azure-sdk-for-java/blob/release/cosmos_v3/sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/RxGatewayStoreModel.java#L147

Addresses this issue: https://github.com/Azure/azure-sdk-for-java/issues/9802

### DIFF
--- a/sdk/cosmos/tests.yml
+++ b/sdk/cosmos/tests.yml
@@ -20,28 +20,28 @@ jobs:
         # 59m 59s, timed out
         Http_E2E:
           DisplayName: ReadMyWrites Integration Tests
-          OSVmImage: 'windows-2019'
+          OSVmImage: 'ubuntu-18.04'
           ProfileFlag: '-Pe2e'
           AdditionalArgs: '-DargLine="-Dcosmos.directModeProtocol=Https"'
           DESIRED_CONSISTENCY: 'Session'
         # 09m 23s, expect passed
         Single_Region_Fast:
           DisplayName: Single Region Fast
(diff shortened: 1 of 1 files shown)


sdk/cosmos/microsoft-azure-cosmos/src/main/java/com/azure/data/cosmos/internal/http/ReactorNettyClient.java
```diff
@@ -118,8 +117,7 @@ private void configureChannelPipelineHandlers() {
                 reactorNettyRequest.header(header.name(), header.value());
             }
             if (restRequest.body() != null) {
-                Flux<ByteBuf> nettyByteBufFlux = restRequest.body().map(Unpooled::wrappedBuffer);
-                return reactorNettyOutbound.send(nettyByteBufFlux);
+                return reactorNettyOutbound.send(restRequest.body());
```
moderakh, line 120: interesting. we missed this :-)

great finding.

kushagraThapar: Thanks Mo :) 
Thanks to @anuchandy  as well for the help so far :)

PR #9942, opened 2020-04-08
Update missing module

conniey:
Fix module name for azure-storage-perf from perf-test

### DIFF
--- a/pom.client.xml
+++ b/pom.client.xml
@@ -1479,8 +1479,8 @@
     <module>sdk/storage/azure-storage-common</module>
     <module>sdk/storage/azure-storage-file-share</module>
     <module>sdk/storage/azure-storage-file-datalake</module>
+    <module>sdk/storage/azure-storage-perf</module>
     <module>sdk/storage/azure-storage-queue</module>
-    <module>sdk/storage/perf-test</module>
     <module>sdk/textanalytics/azure-ai-textanalytics</module>
   </modules>
 </project>


mitchdenny:
Why do we need this module referenced at all? I'm planning to delete all those module references from pom.client.xml very soon (part of the POM restructuring work).

conniey:
@mitchdenny I fixed it because maven couldn't build pom.client.xml because perf-test doesn't exist. I have no problems if you want to remove it instead.

mitchdenny:
Is there somewhere that is building pom.client.xml? Or is this for your local dev? Probably best to cut over to building against the root POM with a pl switch. I'm approving this for now so anyone still building with pom.client.xml are temporarily unblocked, but folks won't be able to build with pom.client.xml anymore moving forward because it won't be an aggregator.

mitchdenny:
... there will be CONTRIBUTING.md changes to that effect.

conniey:
@mitchdenny Yes. I had tried to get set-up on a new machine and was trying to find the quickest way to get all the random current packages installed.

mitchdenny:
Moving forward you can do this (in the root):

```
mvn install -pl com.azure:project-you-are-interested-in -am

cd sdk/service-you-are-interested-in
mvn verify
```

The first invocation will install all the relevant dependencies you need (incl. sdk-build-tools). Then from that point on, provided you don't modify the dependencies you can just run verify on the aggregate POM in the service directory.

mitchdenny:
Another way of doing it is this:

```
mvn install -pl com.azure:azure-messaging-eventhubs -am -DskipTests -Dgpg.skip
mvn verify -pl com.azure:azure-messaging-eventhubs
```

Assuming you are interested in event hubs for example, This will install all the deps for the event hubs package and from then on you are just working with your module.

conniey:
Event Hubs and Service Bus are ones I'm interested in but I also wanted to compile azure-core-amqp because they depend on an unreleased dependency. It was a lot of manual typing.

Issue #9248, opened 2020-03-19
[Feature] Auto detect content type from input streams


samvaity:
beginAnalyzeForm, beginAnalayzeReceipt, and beginAnalyzeLayout all require a FormContentType to be passed in to indicate whether the stream contains a JPG, PNG, etc. We should auto-detect this and possibly allow clients to override (although decide based on our implementation whether they need to do this at all).

PR #9953, opened 2020-04-08
[FR] implements auto-content-detection method getContentType() 

mssfang:
fixes: #9248

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
@@ -5,6 +5,7 @@ package com.azure.ai.formrecognizer;
 
 import com.azure.ai.formrecognizer.implementation.FormRecognizerClientImpl;
 import com.azure.ai.formrecognizer.implementation.models.AnalyzeOperationResult;
+import com.azure.ai.formrecognizer.implementation.models.ContentType;
 import com.azure.ai.formrecognizer.implementation.models.OperationStatus;
 import com.azure.ai.formrecognizer.implementation.models.SourcePath;
 import com.azure.ai.formrecognizer.models.ErrorInformation;
@@ -41,7 +42,7 @@ import static com.azure.ai.formrecognizer.FormRecognizerClientBuilder.DEFAULT_DU
 import static com.azure.ai.formrecognizer.Transforms.toReceipt;
 import static com.azure.ai.formrecognizer.Transforms.toRecognizedForm;
 import static com.azure.ai.formrecognizer.Transforms.toRecognizedLayout;
-import static com.azure.ai.formrecognizer.implementation.Utility.getContentType;
+import static com.azure.ai.formrecognizer.implementation.Utility.detectContentType;
 import static com.azure.ai.formrecognizer.implementation.Utility.parseModelId;
 import static com.azure.core.util.FluxUtil.monoError;
 
@@ -382,17 +383,22 @@ public final class FormRecognizerAsyncClient {
     private Function<PollingContext<OperationResult>, Mono<OperationResult>> receiptStreamActivationOperation(
         Flux<ByteBuffer> data, long length, FormContentType formContentType, boolean includeTextDetails) {
         Objects.requireNonNull(data, "'data' is required and cannot be null.");
-
         return pollingContext -> {
-            try {
-                return data.next()
-                    .map(byteBuffer -> getContentType(byteBuffer, formContentType))
+            if (formContentType != null) {
+                try {
+                    return service.analyzeReceiptAsyncWithResponseAsync(
+                        ContentType.fromString(formContentType.toString()), data, length, includeTextDetails)
+                        .map(response -> new OperationResult(
+                            parseModelId(response.getDeserializedHeaders().getOperationLocation())));
+                } catch (RuntimeException ex) {
+                    return monoError(logger, ex);
+                }
+            } else {
+                return detectContentType(data)
                     .flatMap(contentType ->
                         service.analyzeReceiptAsyncWithResponseAsync(contentType, data, length, includeTextDetails)
                             .map(response -> new OperationResult(
                                 parseModelId(response.getDeserializedHeaders().getOperationLocation()))));
-            } catch (RuntimeException ex) {
-                return monoError(logger, ex);
             }
         };
     }
@@ -444,16 +450,22 @@ public final class FormRecognizerAsyncClient {
     private Function<PollingContext<OperationResult>, Mono<OperationResult>> contentStreamActivationOperation(
         Flux<ByteBuffer> data, long length, FormContentType formContentType) {
         Objects.requireNonNull(data, "'data' is required and cannot be null.");
-
         return pollingContext -> {
-            try {
-                return data.next()
-                    .map(byteBuffer -> getContentType(byteBuffer, formContentType))
-                    .flatMap(contentType -> service.analyzeLayoutAsyncWithResponseAsync(contentType, data, length)
+            if (formContentType != null) {
+                try {
+                    return service.analyzeLayoutAsyncWithResponseAsync(
+                        ContentType.fromString(formContentType.toString()), data, length)
                         .map(response -> new OperationResult(parseModelId(
-                            response.getDeserializedHeaders().getOperationLocation()))));
-            } catch (RuntimeException ex) {
-                return monoError(logger, ex);
+                            response.getDeserializedHeaders().getOperationLocation())));
+                } catch (RuntimeException ex) {
+                    return monoError(logger, ex);
+                }
+            } else {
+                return detectContentType(data)
+                    .flatMap(contentType ->
+                        service.analyzeLayoutAsyncWithResponseAsync(contentType, data, length)
+                            .map(response -> new OperationResult(
+                                parseModelId(response.getDeserializedHeaders().getOperationLocation()))));
             }
         };
     }
@@ -560,17 +572,23 @@ public final class FormRecognizerAsyncClient {
         FormContentType formContentType, boolean includeTextDetails) {
         Objects.requireNonNull(data, "'data' is required and cannot be null.");
         Objects.requireNonNull(modelId, "'modelId' is required and cannot be null.");
-
         return pollingContext -> {
-            try {
-                return data.next()
-                    .map(byteBuffer -> getContentType(byteBuffer, formContentType))
-                    .flatMap(contentType -> service.analyzeWithCustomModelWithResponseAsync(UUID.fromString(modelId),
-                        contentType, data, length, includeTextDetails)
-                        .map(response -> new OperationResult(
-                            parseModelId(response.getDeserializedHeaders().getOperationLocation()))));
-            } catch (RuntimeException ex) {
-                return monoError(logger, ex);
+            if (formContentType != null) {
+                try {
+                    return service.analyzeWithCustomModelWithResponseAsync(UUID.fromString(modelId),
+                        ContentType.fromString(formContentType.toString()), data, length, includeTextDetails)
+                        .map(response -> new OperationResult(parseModelId(
+                            response.getDeserializedHeaders().getOperationLocation())));
+                } catch (RuntimeException ex) {
+                    return monoError(logger, ex);
+                }
+            } else {
+                return detectContentType(data)
+                    .flatMap(contentType ->
+                        service.analyzeWithCustomModelWithResponseAsync(UUID.fromString(modelId), contentType, data,
+                            length, includeTextDetails)
+                            .map(response -> new OperationResult(
+                                parseModelId(response.getDeserializedHeaders().getOperationLocation()))));
             }
         };
     }
--- a/sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
@@ -4,10 +4,10 @@
 package com.azure.ai.formrecognizer.implementation;
 
 import com.azure.ai.formrecognizer.implementation.models.ContentType;
-import com.azure.ai.formrecognizer.models.FormContentType;
 import com.azure.core.util.CoreUtils;
 import com.azure.core.util.logging.ClientLogger;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -25,46 +25,81 @@ public final class Utility {
     }
 
     /**
-     * Automatically detect byte buffer's content type if the {@code formContentType} is not given. Otherwise,
-     * use the given content type value as the result.
+     * Automatically detect byte buffer's content type.
      *
      * Given the source: <a href="https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files"/>.
      *
      * @param buffer The byte buffer input.
-     * @param formContentType The user specified content type.
      *
-     * @return The {@link ContentType} content type.
+     * @return The {@link Mono} of {@link ContentType} content type.
      */
-    public static ContentType getContentType(ByteBuffer buffer, FormContentType formContentType) {
-        // if the form content type is given, use it as return result, otherwise, do the guessing based on the buffer
-        if (formContentType != null) {
-            return ContentType.fromString(formContentType.toString());
-        }
+    public static Mono<ContentType> detectContentType(Flux<ByteBuffer> buffer) {
+        byte[] header = new byte[4];
+        int[] written = new int[] {0};
+        ContentType[] contentType = { ContentType.fromString("none")};
+        return buffer.map(chunk -> {
+            final int len = chunk.remaining();
+            for (int i = 0; i < len; i++) {
+                header[written[0]] = chunk.get(i);
+                written[0]++;
 
-        final byte[] bytes = buffer.array();
-        if (bytes.length < 4) {
-            throw LOGGER.logExceptionAsError(
-                new IllegalArgumentException("Invalid input. Expect more than 4 bytes of data"));
-        }
+                if (written[0] == 4) {
+                    if (isJpeg(header)) {
+                        contentType[0] = ContentType.IMAGE_JPEG;
+                    } else if (isPdf(header)) {
+                        contentType[0] = ContentType.APPLICATION_PDF;
+                    } else if (isPng(header)) {
+                        contentType[0] = ContentType.IMAGE_PNG;
+                    } else if (isTiff(header)) {
+                        contentType[0] = ContentType.IMAGE_TIFF;
+                    }
+                    // Got a four bytes matching or not, either way no need to read more byte return false
+                    // so that takeWhile can cut the subscription on data
+                    return false;
+                }
+            }
+            // current chunk don't have enough bytes so return true to get next Chunk if there is one.
+            return true;
+        })
+        .takeWhile(doContinue -> doContinue)
+        .then(Mono.defer(() -> {
+            if (contentType[0] != null) {
+                return Mono.just(contentType[0]);
+            } else {
+                return Mono.error(new RuntimeException("Content type could not be detected. "
+                    + "Should use other overload API that takes content type."));
+            }
+        }));
+    }
 
-        if (bytes[0] == (byte) 0x25 && bytes[1] == (byte) 0x50 && bytes[2] == (byte) 0x44 && bytes[3] == (byte) 0x46) {
-            return ContentType.APPLICATION_PDF;
-        } else if (bytes[0] == (byte) 0xff && bytes[1] == (byte) 0xd8) {
-            return ContentType.IMAGE_JPEG;
-        } else if (bytes[0] == (byte) 0x89 &&  bytes[1] == (byte) 0x50 && bytes[2] == (byte) 0x4e
-            && bytes[3] == (byte) 0x47) {
-            return ContentType.IMAGE_PNG;
-        } else if (
-            // little-endian
-            (bytes[0] == (byte) 0x49 && bytes[1] == (byte) 0x49 && bytes[2] == (byte) 0x2a && bytes[3] == (byte) 0x0)
-                // big-endian
-                || (bytes[0] == (byte) 0x4d && bytes[1] == (byte) 0x4d && bytes[2] == (byte) 0x0
-                    && bytes[3] == (byte) 0x2a)) {
-            return ContentType.IMAGE_TIFF;
-        } else {
-            throw LOGGER.logExceptionAsError(new IllegalArgumentException(
-                "Content type could not be detected. Should use other overload API that takes content type."));
-        }
+    private static boolean isJpeg(byte[] header) {
+        return (header[0] == (byte) 0xff && header[1] == (byte) 0xd8);
+    }
+
+    private static boolean isPdf(byte[] header) {
+        return header[0] == (byte) 0x25
+            && header[1] == (byte) 0x50
+            && header[2] == (byte) 0x44
+            && header[3] == (byte) 0x46;
+    }
+
+    private static boolean isPng(byte[] header) {
+        return header[0] == (byte) 0x89
+            &&  header[1] == (byte) 0x50
+            && header[2] == (byte) 0x4e
+            && header[3] == (byte) 0x47;
+    }
+
+    private static boolean isTiff(byte[] header) {
+        return (header[0] == (byte) 0x49
+            && header[1] == (byte) 0x49
+            && header[2] == (byte) 0x2a
+            && header[3] == (byte) 0x0)
+            // big-endian
+            || (header[0] == (byte) 0x4d
+            && header[1] == (byte) 0x4d
+            && header[2] == (byte) 0x0
+            && header[3] == (byte) 0x2a);
     }
 
     /**
--- a/sdk/formrecognizer/azure-ai-formrecognizer/src/test/java/com/azure/ai/formrecognizer/ContentTypeDetectionTest.java
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/src/test/java/com/azure/ai/formrecognizer/ContentTypeDetectionTest.java
@@ -5,7 +5,6 @@ package com.azure.ai.formrecognizer;
 
 import com.azure.ai.formrecognizer.implementation.Utility;
 import com.azure.ai.formrecognizer.implementation.models.ContentType;
-import com.azure.ai.formrecognizer.models.FormContentType;
 import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Flux;
 
@@ -15,7 +14,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
 
-import static com.azure.ai.formrecognizer.implementation.Utility.getContentType;
+import static com.azure.ai.formrecognizer.implementation.Utility.detectContentType;
 import static com.azure.ai.formrecognizer.implementation.Utility.toFluxByteBuffer;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
@@ -26,7 +25,7 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 public class ContentTypeDetectionTest {
 
     /**
-     * Test for JPG file content type detection for {@link Utility#getContentType} method.
+     * Test for JPG file content type detection for {@link Utility#detectContentType} method.
      *
      * @throws  IOException if an I/O error occurs reading from the stream
      */
@@ -34,11 +33,11 @@ public class ContentTypeDetectionTest {
     public void jpgContentDetectionTest() throws IOException {
         File sourceFile = new File("src/test/resources/sample-files/contoso-allinone.jpg");
         Flux<ByteBuffer> buffer = toFluxByteBuffer(new ByteArrayInputStream(Files.readAllBytes(sourceFile.toPath())));
-        assertEquals(ContentType.IMAGE_JPEG, getContentType(buffer.blockFirst(), FormContentType.IMAGE_JPEG));
+        assertEquals(ContentType.IMAGE_JPEG, detectContentType(buffer).block());
     }
 
     /**
-     * Test for PDF file content type detection for {@link Utility#getContentType} method.
+     * Test for PDF file content type detection for {@link Utility#detectContentType} method.
      *
      * @throws  IOException if an I/O error occurs reading from the stream
      */
@@ -46,11 +45,11 @@ public class ContentTypeDetectionTest {
     public void pdfContentDetectionTest() throws IOException {
         File sourceFile = new File("src/test/resources/sample-files/sample.pdf");
         Flux<ByteBuffer> buffer = toFluxByteBuffer(new ByteArrayInputStream(Files.readAllBytes(sourceFile.toPath())));
-        assertEquals(ContentType.APPLICATION_PDF, getContentType(buffer.blockFirst(), FormContentType.APPLICATION_PDF));
+        assertEquals(ContentType.APPLICATION_PDF, detectContentType(buffer).block());
     }
 
     /**
-     * Test for PNG file content type detection for {@link Utility#getContentType} method.
+     * Test for PNG file content type detection for {@link Utility#detectContentType} method.
      *
      * @throws  IOException if an I/O error occurs reading from the stream
      */
@@ -58,11 +57,11 @@ public class ContentTypeDetectionTest {
     public void pngContentDetectionTest() throws IOException {
         File sourceFile = new File("src/test/resources/sample-files/pngFile.png");
         Flux<ByteBuffer> buffer = toFluxByteBuffer(new ByteArrayInputStream(Files.readAllBytes(sourceFile.toPath())));
-        assertEquals(ContentType.IMAGE_PNG, getContentType(buffer.blockFirst(), FormContentType.IMAGE_PNG));
+        assertEquals(ContentType.IMAGE_PNG, detectContentType(buffer).block());
     }
 
     /**
-     * Test for TIFF file content type detection for {@link Utility#getContentType} method.
+     * Test for TIFF file content type detection for {@link Utility#detectContentType} method.
      *
      * @throws  IOException if an I/O error occurs reading from the stream
      */
@@ -70,11 +69,11 @@ public class ContentTypeDetectionTest {
     public void tiffContentDetectionTest() throws IOException {
         File sourceFile = new File("src/test/resources/sample-files/cell.tif");
         Flux<ByteBuffer> buffer = toFluxByteBuffer(new ByteArrayInputStream(Files.readAllBytes(sourceFile.toPath())));
-        assertEquals(ContentType.IMAGE_TIFF, getContentType(buffer.blockFirst(), FormContentType.IMAGE_TIFF));
+        assertEquals(ContentType.IMAGE_TIFF, detectContentType(buffer).block());
     }
 
     /**
-     * Test for not supported file content type detection for {@link Utility#getContentType} method.
+     * Test for not supported file content type detection for {@link Utility#detectContentType} method.
      *
      * @throws  IOException if an I/O error occurs reading from the stream
      */
@@ -82,6 +81,6 @@ public class ContentTypeDetectionTest {
     public void notSupportContentDetectionTest() throws IOException {
         File sourceFile = new File("src/test/resources/sample-files/docFile.doc");
         Flux<ByteBuffer> buffer = toFluxByteBuffer(new ByteArrayInputStream(Files.readAllBytes(sourceFile.toPath())));
-        assertThrows(IllegalArgumentException.class, () -> getContentType(buffer.blockFirst(), null));
+        assertThrows(RuntimeException.class, () -> detectContentType(buffer).block());
     }
 }
--- a/sdk/formrecognizer/azure-ai-formrecognizer/src/samples/java/com/azure/ai/formrecognizer/AdvancedDiffLabeledUnlabeledDataAsync.java
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/src/samples/java/com/azure/ai/formrecognizer/AdvancedDiffLabeledUnlabeledDataAsync.java
@@ -3,7 +3,6 @@
 
 package com.azure.ai.formrecognizer;
 
-import com.azure.ai.formrecognizer.implementation.Utility;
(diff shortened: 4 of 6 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/Transforms.java
```diff
@@ -183,6 +185,43 @@ private static PageMetadata getPageInfo(ReadResult readResultItem) {
...
+        if (bytes.length < 4) {
+            throw new IllegalArgumentException("Invalid input. Expect more than 4 bytes of data");
+        }
+
+        final int byte0 = Byte.toUnsignedInt(bytes[0]);
+        final int byte1 = Byte.toUnsignedInt(bytes[1]);
+        final int byte2 = Byte.toUnsignedInt(bytes[2]);
+        final int byte3 = Byte.toUnsignedInt(bytes[3]);
+
+        if (byte0 == Byte.toUnsignedInt((byte) 0x25) && byte1 == Byte.toUnsignedInt((byte) 0x50) &&
```
iscai-msft, line 207: It could be nice to pull out the Byte.toUnsignedInt((byte)) into a separate function

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/Transforms.java
```diff
@@ -183,6 +185,43 @@ private static PageMetadata getPageInfo(ReadResult readResultItem) {
...
+            byte2 == Byte.toUnsignedInt((byte) 0x44) && byte3 == Byte.toUnsignedInt((byte) 0x46)) {
+            return ContentType.APPLICATION_PDF;
+        } else if (byte0 == Byte.toUnsignedInt((byte) 0xff) && byte1 == Byte.toUnsignedInt((byte) 0xd8)) {
```
iscai-msft, line 210: For Jpeg, I think you also need to check the last two bytes are 0xFF and 0xD9

sdk/formrecognizer/azure-ai-formrecognizer/src/test/java/com/azure/ai/formrecognizer/FormRecognizerClientTestBase.java
```diff
@@ -218,7 +218,7 @@ private static void validatePageMetadata(PageMetadata expectedPageInfo, PageMeta
      * Create a client builder with endpoint and API key credential.
      *
      * @param endpoint the given endpoint
-     * @param credential the given {@link FormRecognizerApiKeyCredential} credential
+     * @param credential the given {@link AzureKeyCredential} credential
```
iscai-msft, line 221: Just making sure that this little edit is all that needs to be changed in the key credential transition

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/Transforms.java
```diff
@@ -183,6 +185,43 @@ private static PageMetadata getPageInfo(ReadResult readResultItem) {
...
+        } else if (byte0 == Byte.toUnsignedInt((byte) 0x89) && byte1 == Byte.toUnsignedInt((byte) 0x50) &&
+            byte2 == Byte.toUnsignedInt((byte) 0x4e) && byte3 == Byte.toUnsignedInt((byte) 0x47)) {
+            return ContentType.IMAGE_PNG;
+        } else if ((byte0 == Byte.toUnsignedInt((byte) 0x49) && byte1 == Byte.toUnsignedInt((byte) 0x49) &&
+                byte2 == Byte.toUnsignedInt((byte) 0x2a) && byte3 == Byte.toUnsignedInt((byte) 0x0)) || // little-endian
+            (byte0 == Byte.toUnsignedInt((byte) 0x4d) && byte1 == Byte.toUnsignedInt((byte) 0x4d) &&
+                byte2 == Byte.toUnsignedInt((byte) 0x0) && byte3 == Byte.toUnsignedInt((byte) 0x2a))) {  // big-endian
+            return ContentType.IMAGE_TIFF;
+        } else {
+            throw new IllegalArgumentException ("content type could not be detected");
```
kristapratico, line 221: If content type cannot be auto-detected we still need to let the user pass content type themselves. Perhaps include that in this exception message?

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/Transforms.java
kristapratico: I removed this for python

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
```diff
@@ -16,37 +16,61 @@
...
+                byte[] buffer = new byte[BYTE_BUFFER_CHUNK_SIZE];
+                try {
+                    int numBytes = inputStream.read(buffer);
+                    if (numBytes > 0) {
+                        return pair.buffer(ByteBuffer.wrap(buffer, 0, numBytes)).readBytes(numBytes);
+                    } else {
+                        return pair.buffer(null).readBytes(numBytes);
+                    }
+                } catch (IOException ioe) {
+                    throw Exceptions.propagate(ioe);
```
samvaity, line 46: Should this be logger throw?

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
```diff
@@ -16,37 +16,61 @@
...
-     *
-     * @return {@link ByteBuffer} which contains the input data.
-     * @throws RuntimeException When I/O error occurs.
+     * @param inputStream InputStream to back the Flux
+     * @return Flux of ByteBuffer backed by the InputStream
      */
-    public static Flux<ByteBuffer> convertStreamToByteBuffer(InputStream data) {
-        return Flux.just(toByteArray(data))
-            .doOnError(error -> LOGGER.warning("Failed to convert stream to byte array - {}", error));
+    public static Flux<ByteBuffer> toFluxByteBuffer(InputStream inputStream) {
```
samvaity, line 32: Should we consider moving this to core?
cc: @srnagar

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/Transforms.java
```diff
@@ -183,6 +185,43 @@ private static PageMetadata getPageInfo(ReadResult readResultItem) {
...
 
+    /**
+     * Automatically detect byte buffer's content type.
+     * Given the source: <a href="https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files"/>.
+     *
+     * @param buffer The byte buffer input.
+     *
+     * @return The {@link ContentType} content type.
+     */
+    static ContentType getContentType(ByteBuffer buffer) {
```
samvaity, line 196: Need tests?

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -30,6 +29,7 @@
 import java.util.UUID;
 import java.util.function.Function;
 
+import static com.azure.ai.formrecognizer.Transforms.getContentType;
```
jeremymeng, line 32: Feels like it shouldn't belong to `Trasnform`?

samvaity: Yeah, could move this to Utility @mssfang ?

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/Transforms.java
iscai-msft: ok ignore this

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
mssfang: Will move to Utility class

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
mssfang: FYI, I talked to @alzimmermsft  a couple days ago. I will finish this implementation in Form Recognizer first, and then he will decide to move this to core or not.

sdk/formrecognizer/azure-ai-formrecognizer/src/test/java/com/azure/ai/formrecognizer/FormRecognizerClientTestBase.java
mssfang: FormRecognizerApiKeyCredential is a deprecated type

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/Transforms.java
mssfang: updated

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
mssfang: updated

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/Transforms.java
mssfang: added a few unit tests

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/Transforms.java
mssfang: updated

sdk/formrecognizer/azure-ai-formrecognizer/src/test/java/com/azure/ai/formrecognizer/FormRecognizerClientTestBase.java
maririos: Sounds more like Isa was calling out if there were other places where the change needed to happen.
Looks like this was the only remaining though

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
```diff
@@ -22,6 +23,41 @@
...
+     * @return The {@link ContentType} content type.
+     */
+    public static ContentType getContentType(ByteBuffer buffer) {
+        final byte[] bytes = buffer.array();
+        if (bytes.length < 4) {
+            throw LOGGER.logExceptionAsError(
+                new IllegalArgumentException("Invalid input. Expect more than 4 bytes of data"));
+        }
+
+        if (isEqual(bytes[0], 0x25) && isEqual(bytes[1], 0x50) && isEqual(bytes[2], 0x44) && isEqual(bytes[3], 0x46)) {
```
iscai-msft, line 41: nitpick (feel free to ignore)
You could make the function isEqual take in a list of bytes and what they should be, so the code could be more like
`isEqual(bytes.subList(0, 4), [0x25, 0x50, 0x44, 0x46])`

sdk/formrecognizer/azure-ai-formrecognizer/src/test/java/com/azure/ai/formrecognizer/ContentTypeDetectionTest.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+public class ContentTypeDetectionTest {
+
+    /**
+     * Test for JPG file content type detection.
+     *
+     * @throws  IOException if an I/O error occurs reading from the stream
+     */
+    @Test
+    public void jpgContentDetectionTest() throws IOException {
```
iscai-msft, line 31: Consider adding a test where the user passes in the content type

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -173,11 +170,11 @@ public FormRecognizerServiceVersion getServiceVersion() {
     }
 
     private Function<PollingContext<OperationResult>, Mono<OperationResult>> receiptStreamActivationOperation(
-        Flux<ByteBuffer> buffer, long length, FormContentType formContentType, boolean includeTextDetails) {
+        Flux<ByteBuffer> buffer, long length, boolean includeTextDetails) {
         return (pollingContext) -> {
             try {
                 return service.analyzeReceiptAsyncWithResponseAsync(includeTextDetails,
-                    ContentType.fromString(formContentType.toString()), buffer, length)
+                    getContentType(buffer.blockFirst()), buffer, length)
```
alzimmermsft, line 177: We generally don't want to block in Reactive code as this could cause a RuntimeException for blocking in a non-blocking thread. We have some control on whether non-blocking threads are being used but not complete control.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
alzimmermsft: This is something we'll want to add into `FluxUtil` eventually, mind filing an issue for this.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
```diff
@@ -16,37 +17,107 @@
...
-        return Flux.just(toByteArray(data))
-            .doOnError(error -> LOGGER.warning("Failed to convert stream to byte array - {}", error));
+    public static ContentType getContentType(ByteBuffer buffer) {
+        final byte[] bytes = buffer.array();
+        if (bytes.length < 4) {
+            throw LOGGER.logExceptionAsError(
+                new IllegalArgumentException("Invalid input. Expect more than 4 bytes of data"));
+        }
+
+        if (isEqual(bytes[0], 0x25) && isEqual(bytes[1], 0x50) && isEqual(bytes[2], 0x44) && isEqual(bytes[3], 0x46)) {
```
alzimmermsft, line 41: Instead of casting the byte to an int we could just use the following checks:

```java
bytes[0] == (byte) 0x25...
```

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
```diff
@@ -16,37 +17,107 @@
...
+
+    /**
+     * Compare if a byte value equals to a hex type value.
+     *
+     * @param byteValue the byte type value
+     * @param hexValue the hex type value
+     * @return true if two type's values are equal in unsigned int comparision.
+     */
+    private static boolean isEqual(byte byteValue, int hexValue) {
+        return Byte.toUnsignedInt(byteValue) == Byte.toUnsignedInt((byte) hexValue);
```
alzimmermsft, line 121: If we do want to end up keeping this we shouldn't cast and re-cast the hex value.

`byteValue == (byte) hexValue` should be sufficient.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
```diff
@@ -16,37 +17,107 @@
...
+            return ContentType.IMAGE_PNG;
+        } else if (
+            // little-endian
+            (isEqual(bytes[0], 0x49) && isEqual(bytes[1], 0x49) && isEqual(bytes[2], 0x2a) && isEqual(bytes[3], 0x0))
+            // big-endian
+            || (isEqual(bytes[0], 0x4d) && isEqual(bytes[1], 0x4d) && isEqual(bytes[2], 0x0)
+                && isEqual(bytes[3], 0x2a))) {
+            return ContentType.IMAGE_TIFF;
+        } else {
+            throw new IllegalArgumentException(
```
alzimmermsft, line 56: Since we have the static logger we should log this.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
```diff
@@ -16,37 +17,107 @@
...
-     * @return {@link ByteBuffer} which contains the input data.
-     * @throws RuntimeException When I/O error occurs.
+     * @return The {@link ContentType} content type.
      */
-    public static Flux<ByteBuffer> convertStreamToByteBuffer(InputStream data) {
-        return Flux.just(toByteArray(data))
-            .doOnError(error -> LOGGER.warning("Failed to convert stream to byte array - {}", error));
+    public static ContentType getContentType(ByteBuffer buffer) {
+        final byte[] bytes = buffer.array();
+        if (bytes.length < 4) {
```
alzimmermsft, line 36: Since we are only checking the first `ByteBuffer` what would happen if the first buffer is empty due to a disk or networking latency but then the next one contains the byte marking which would properly determine the file type.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
mssfang: https://github.com/Azure/azure-sdk-for-java/issues/10337

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -382,18 +384,50 @@ public FormRecognizerServiceVersion getServiceVersion() {
...
+                cachedBuffers.add(cachedBuffer);
+
+                if (bufferedDataSize[0] >= length) {
+                    return true;
+                }
+                return false;
+            }, true, Integer.MAX_VALUE)
+            .next()
+            .flatMap(newData -> {
+                // if content type is not given, do guessing the content type.
```
alzimmermsft, line 424: If Content-Type is given we can skip trying to guess the type.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
```diff
@@ -17,37 +17,121 @@
...
+    }
+
+    /**
+     * Compare if a byte value equals to a hex type value.
+     *
+     * @param byteValueList An array of byte type values.
+     * @param hexValueList An array of hex type values.
+     * @return true if two type's values are equal in unsigned int comparision, otherwise return false.
+     */
+    private static boolean isEqual(byte[] byteValueList, int[] hexValueList) {
```
alzimmermsft, line 120: I don't feel this abstraction is really worth the overhead, both in maintenance and performance.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -556,19 +623,53 @@ private void throwIfAnalyzeStatusInvalid(SimpleResponse<AnalyzeOperationResult>
...
+        // Size of buffered data, it used to collect all data from input Flux<ByteBuffer>
+        final long[] bufferedDataSize = {0};
+        // Cached byte buffer
+        final List<ByteBuffer> cachedBuffers = new LinkedList<>();
+        // A boolean value that only used once for guessing content type on input data
+        final AtomicBoolean[] isGuessed = {new AtomicBoolean(true)};
+        // Content type on input data
+        final ContentType[] contentType = new ContentType[1];
+
+        return pollingContext -> data.filter(ByteBuffer::hasRemaining)
```
alzimmermsft, line 638: This logic appears to be duplicated three times, any chance this could be abstracted? Possibly a method that takes `Flux<ByteBuffer>` and returns `Mono<Tuple2<ContentType, Flux<ByteBuffer>>`.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -442,18 +476,51 @@ public FormRecognizerServiceVersion getServiceVersion() {
...
```
alzimmermsft, line 490: The logic here is close, but a little off. Right now this will cache the entire data stream, this could be modified to cache until we get the number of bytes required to determine the content type, then join the cached buffers with the remaining buffers.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -382,18 +384,50 @@ public FormRecognizerServiceVersion getServiceVersion() {
...
+                ByteBuffer cachedBuffer = ByteBuffer.allocate(buffer.remaining()).put(buffer);
+                cachedBuffer.flip();
+                cachedBuffers.add(cachedBuffer);
+
+                if (bufferedDataSize[0] >= length) {
+                    return true;
+                }
+                return false;
+            }, true, Integer.MAX_VALUE)
+            .next()
```
anuchandy, line 422: If I understand it correctly - the `data` here can be in one of the two formats: 

data = content-type + content
data = content

We've this `windowUntil` here to detect the content-type but that comes with a cost i.e. we need to allocate a new ByteBuffer (or multiple ByteBuffers if data emits byte-by-byte), we do that even if user provide us the contentType as an argument. I think when contentType is given we can call the underlying API and give it data as it is, if contentType is not given then apply the `windowUntil` path.  This way we can also simplify the logic, we don't the var  `isGuessed` or check like `formContentType == null` 

Coming to the current `windowUntil` path - I think it may not work correctly if data comes as chunks i.e.

```
data = ByteBuffer_chunk1 (containing contentType and some data),  ByteBuffer_chunk2, ByteBuffer_chunk3
```
Looks like by doing `next()` after `windowUntil`, the newData will be just ByteBuffer_chunk1. Good to double check.

You might want to apply buffer operator after windowUntil
```
.buffer(2)
.next()
.flatMap(fluxes -> {
    var contentType = getContentType(cachedBuffers);
    if (fluxes.size() == 1) {
        return callAPI(Flux.fromIterable(cachedBuffers));
    } else {
        return callAPI(Flux.fromIterable(cachedBuffers).concatWith(fluxes.get(1)));
    }
})
```

anuchandy: Also with explicit contentType check

```java
if (contentType == null) {
   return callApi(data, ..);
} else {
   return data.windowUntil(..).flatMap(fluxes -> {
      var contentType = getContentType(cachedBuffers);
      if (fluxes.size() == 1) {
          return callAPI(Flux.fromIterable(cachedBuffers));
      } else {
          return callAPI(Flux.fromIterable(cachedBuffers).concatWith(fluxes.get(1)));
      }
    });
}
```

I think we just have to cache in the `windowUntil` operator i.e. no contentType check in `windowUntil`, the check can be deferred to flatmap operator.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -382,18 +384,50 @@ public FormRecognizerServiceVersion getServiceVersion() {
...
-                    data, length, includeTextDetails)
-                    .map(response -> new OperationResult(
+        Objects.requireNonNull(data, "'data' is required and cannot be null.");
+
+        // Size of buffered data, it used to collect all data from input Flux<ByteBuffer>
+        final long[] bufferedDataSize = {0};
+        // Cached byte buffer
+        final List<ByteBuffer> cachedBuffers = new LinkedList<>();
+        // A boolean value that only used once for guessing content type on input data
+        final AtomicBoolean[] isGuessed = {new AtomicBoolean(true)};
```
srnagar, line 394: Why do you need an array? This can just be `AtomicBoolean`.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -382,18 +384,50 @@ public FormRecognizerServiceVersion getServiceVersion() {
...
+                // Cache the data to a list of byte buffers
+                ByteBuffer cachedBuffer = ByteBuffer.allocate(buffer.remaining()).put(buffer);
+                cachedBuffer.flip();
+                cachedBuffers.add(cachedBuffer);
+
+                if (bufferedDataSize[0] >= length) {
+                    return true;
+                }
+                return false;
+            }, true, Integer.MAX_VALUE)
```
srnagar, line 421: Is this done to support hot fluxes? I thought this API only supported cold fluxes. If we only support cold fluxes, this can be simplified.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/implementation/Utility.java
```diff
@@ -17,37 +17,121 @@
...
-        return Flux.just(toByteArray(data))
-            .doOnError(error -> LOGGER.warning("Failed to convert stream to byte array - {}", error));
+    public static ContentType getContentType(ByteBuffer buffer) {
+        final byte[] bytes = buffer.array();
+        if (bytes.length < 2) {
+            throw LOGGER.logExceptionAsError(
+                new IllegalArgumentException("Invalid input. Expect more than 2 bytes of data"));
+        }
+
+        final byte[] firstFourBytesArray = new byte[] {bytes[0], bytes[1], bytes[2], bytes[3]};
```
srnagar, line 41: The byte array length is checked for `< 2` but the first four bytes are accessed. Should the length check be `< 4`?

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -382,18 +384,50 @@ public FormRecognizerServiceVersion getServiceVersion() {
...
+                    isGuessed[0].compareAndSet(true, false);
+                }
+                // Cache the data to a list of byte buffers
+                ByteBuffer cachedBuffer = ByteBuffer.allocate(buffer.remaining()).put(buffer);
+                cachedBuffer.flip();
+                cachedBuffers.add(cachedBuffer);
+
+                if (bufferedDataSize[0] >= length) {
+                    return true;
+                }
```
srnagar, line 419: If the buffer data is greater than length, we should stop and throw an exception as the length provided by user is incorrect.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
anuchandy: right but like we discussed offline this could be a cold + self-destructed flux, so cloning is required for scanning of content type, and we don't know upfront that its self-destructed flux or not :( .. that's why i thought we should do windowUntil (and the cloning comes with it) only if contentType is missing.

anuchandy: @srnagar - Regarding cold (replay-able) Flux, was thinking a bit more on it. Seeing an issue with self-destructible Flux assumption - though we'll manage to read the header here (for content type) with a single subscription, there is still an issue with retry since each retry is a new subscription unless we do ByteBuffer pooling + back-pressure internally. I don't know this API supports chunked upload where pooling can be applied.

Maybe chunked upload (if supported) could be a future improvement. Keeping that chunk/pooling aside - I think it's better if we make it clear in the doc that this API expects replay-able Flux and go with double subscription, which will anyway happen with retry. Thoughts?

srnagar: My assumption with this API was always that it's going to be a replayable flux. A non-replayable flux would require us to deal with what happens when we want to retry? Storage blob provides a mechanism to have chunk uploads that can be committed once the whole flux is uploaded. Form Recognizer doesn't have this capability. 

So, having a large media file upload requires us to cache the entire flux in-memory or use storage to upload chunks. Also, I am not sure if a non-replayable stream is supported in other languages. If it's not, then we should also make it clear in our documentation that only cold fluxes are supported.

anuchandy: yep, I think good to have this assumption added in our Java guidelines, i.e. all APIs that takes `Flux<BuyteBuffer>` requires the stream to be replay-able. 

If form recognizer support recognizing content of a blob then its good to have a note in the FR api that takes `Flux<BuyteBuffer>`that - for large files use storage optimized-reliable upload and use relevant FR api that takes blob uri.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -382,14 +381,16 @@ public FormRecognizerServiceVersion getServiceVersion() {
...
+
+        return pollingContext -> {
             try {
-                Objects.requireNonNull(data, "'data' is required and cannot be null.");
-                Objects.requireNonNull(formContentType, "'formContentType' is required and cannot be null.");
-                return service.analyzeReceiptAsyncWithResponseAsync(ContentType.fromString(formContentType.toString()),
-                    data, length, includeTextDetails)
-                    .map(response -> new OperationResult(
-                        parseModelId(response.getDeserializedHeaders().getOperationLocation())));
+                return data.next()
```
anuchandy, line 388: Thanks, Shawn, for simplifying this.

This looks better, but it assumes that the first ByteBuffer from data is at least 2 or 4 bytes, which may not always the case. For example, if the data arrives in following structure, then next() operator is not sufficient:

```java
Flux<ByteBuffer> data = Flux.fromArray(new ByteBuffer [] {
    ByteBuffer.wrap(new byte[]{(byte) 0xff}),
    ByteBuffer.wrap(new byte[]{(byte) 0xd8}),
    ByteBuffer.wrap(new byte[]{3}),
    ByteBuffer.wrap(new byte[]{4})
});
```

may be something like below which looks for header even if the header bytes are split across chunks, (please test it, also return type should be ContentType)

```java
private static Mono<String> detectContentType(Flux<ByteBuffer> data) {
    byte[] header = new byte[4];
    int[] written = new int[] {0};
    String[] contentType = new String[] {null};
    return data.map(chunk -> {
        final int len = chunk.remaining();
        for (int i = 0 ;i < len; i++) {
            header[written[0]] = chunk.get(i);
            written[0]++;
            if (written[0] == 2) {
                if (isJpeg(header)) {
                    contentType[0] = "IMAGE_JPEG";
                    // Got a matching two bytes, return false so that takeWhile
                    // can cut the subscription on data
                    return false;
                } else {
                    // two bytes didn't match, continue to read next byte
                    continue;
                }
            }
            if (written[0] == 4) {
                if (isPdf(header)) {
                    contentType[0] = "APPLICATION_PDF";
                } else if (isPng(header)) {
                    contentType[0] = "IMAGE_PNG";
                } else if (isTiff(header)) {
                    contentType[0] = "IMAGE_TIFF";
                }
                // Got a four bytes matching or not, either way no need to read more byte return false 
               // so that takeWhile can cut the subscription on data
                return false;
            }
        }
        // current chunk don't have enough bytes so return true to get next Chunk if there is one.
        return true;
    })
    .takeWhile(doContinue -> doContinue == true)
    .then(Mono.defer(() -> {
        if (contentType[0] != null) {
            return Mono.just(contentType[0]);
        } else {
            return Mono.error(new RuntimeException("Content type could not be detected. " +
                "Should use other overload API that takes content type."));
        }
    }));
}

private static boolean isJpeg(byte[] header) {
    return (header[0] == (byte) 0xff && header[1] == (byte) 0xd8);
}

private static boolean isPdf(byte[] header) {
    return header[0] == (byte) 0x25
        && header[1] == (byte) 0x50
        && header[2] == (byte) 0x44
        && header[3] == (byte) 0x46;
}

private static boolean isPng(byte[] header) {
   return header[0] == (byte) 0x89
        &&  header[1] == (byte) 0x50
        && header[2] == (byte) 0x4e
        && header[3] == (byte) 0x47;
}

private static boolean isTiff(byte[] header) {
    return (header[0] == (byte) 0x49
        && header[1] == (byte) 0x49
        && header[2] == (byte) 0x2a
        && header[3] == (byte) 0x0)
        // big-endian
        || (header[0] == (byte) 0x4d
        && header[1] == (byte) 0x4d
        && header[2] == (byte) 0x0
        && header[3] == (byte) 0x2a);
}
```

anuchandy: also, it may be more clean/readable to have this scanning only if user didn't provide the contentType, so that we can avoid double subscription if not needed.

```java
if (contentType !=) {
   return callAPI(data, contentType);
} else {
  return detectContentType(data)
     .flatmap(contentType -> {
         return callAPI(data, contentType);
     }); 
}
```

srnagar: We can further simplify the JPEG section as a valid JPEG contains a minimum of 4 bytes: https://en.wikipedia.org/wiki/JPEG#Syntax_and_structure

```java
for (int i = 0; i < len; i++) {
	header[written[0]] = chunk.get(i);
	written[0]++;
	if (written[0] == 4) {
		if (isJpeg(header)) {
			contentType[0] = "IMAGE_JPEG";
		} else if (isPdf(header)) {
			contentType[0] = "APPLICATION_PDF";
		} else if (isPng(header)) {
			contentType[0] = "IMAGE_PNG";
		} else if (isTiff(header)) {
			contentType[0] = "IMAGE_TIFF";
		}
		// Got a four bytes matching or not, either way no need to read more byte return false 
	   // so that takeWhile can cut the subscription on data
		return false;
	}
}      
```

anuchandy: @srnagar interesting, I just copied from the existing code already in this PR. I wonder the two byte check was agreed across all language sdks, if so we should communicate correct look up (as you suggested) to others.

srnagar: To find the content type, the first 2 bytes (Start of image) are sufficient. However, a valid jpeg will also have 2 trailing bytes (End of image). So, reading 4 bytes before determining the content type is okay and doesn't need special logic to stop after reading 2 bytes.

anuchandy: synced with @srnagar, @mssfang -  so JPEG has 2 leading bytes that mark the start of image and 2 trailing bytes that mark the end of image, we can read 4 bytes but have to check only the first two bytes, which simplifies the code like Srikanta pointed.

sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -382,16 +382,23 @@ public FormRecognizerServiceVersion getServiceVersion() {
...
-            try {
-                Objects.requireNonNull(data, "'data' is required and cannot be null.");
-                Objects.requireNonNull(formContentType, "'formContentType' is required and cannot be null.");
-                return service.analyzeReceiptAsyncWithResponseAsync(ContentType.fromString(formContentType.toString()),
-                    data, length, includeTextDetails)
-                    .map(response -> new OperationResult(
-                        parseModelId(response.getDeserializedHeaders().getOperationLocation())));
-            } catch (RuntimeException ex) {
-                return monoError(logger, ex);
+        Objects.requireNonNull(data, "'data' is required and cannot be null.");
```
srnagar, line 385: This will throw an error that will cause the public async API to also throw an error. As per reactive guidelines, you should instead catch this exception in the public API method and return an error. See `PagedFlux` APIs for reference. You will have to do something similar for PollerFlux too.

File an issue for this and can be fixed post preview release.

mssfang: Thanks. Srikanta. #10408

PR #9955, opened 2020-04-08
feat: do not suppress management cosmos sdk

ChenTanyi:
The mgmt cosmos sdk don't want to be suppressed, so I add more path in it. Let me know your opinion if better to add `mgmt` in the negative lookahead regex.


ChenTanyi:
@mbhaskar It seems the suppression is added by you. Could you take a look at whether this change affects your suppression or not?

PR #9962, opened 2020-04-08
mgmt keyvault fix checkstyle

weidongxu-microsoft:


sdk/keyvault/mgmt/src/main/java/com/azure/management/keyvault/Secret.java
```diff
@@ -121,16 +121,16 @@
...
+//        /** The stage of a secret update allowing to set the secret to a different version. */
+//        interface WithVersion {
+//            /**
+//             * Specifies the version the secret show use.
+//             *
+//             * @param version the version of the secret
+//             * @return the next stage of the secret update
+//             */
+//            Update withVersion(String version);
+//        }
```
ChenTanyi, line 133: Does it could be removed rather than commented?

weidongxu-microsoft: Removed.

PR #10022, opened 2020-04-09
Add support for receiving from Subscription

### LINKED ISSUE #9883: Add support for receiving from Topics and a Subscription
* Currently we have not implemented a way to receive messages via a subscription from a topic.
* Add tests
* Add samples
* Make sure it works for the synchronous client.

conniey:
* Adds support for receiving from subscriptions.
* Adds a clean up phase for tests. In the case that they fail, we should clear the queue.. else subsequent tests will fail.

Fixes #9883

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
@@ -316,14 +316,14 @@ class ServiceBusReceiverAsyncClientIntegrationTest extends IntegrationTestBase {
     /**
      * Verifies that we can renew message lock.
      */
-    @MethodSource("messagingEntityWithSessions")
+    @MethodSource("messagingEntityProvider")
     @ParameterizedTest
-    void receiveAndRenewLock(MessagingEntityType entityType, boolean isSessionEnabled) {
+    void receiveAndRenewLock(MessagingEntityType entityType) {
         // Arrange
-        setSenderAndReceiver(entityType, isSessionEnabled);
+        setSenderAndReceiver(entityType, false);
 
         final String messageId = UUID.randomUUID().toString();
-        final ServiceBusMessage message = getMessage(messageId, isSessionEnabled);
+        final ServiceBusMessage message = getMessage(messageId, false);
 
         final ReceiveAsyncOptions options = new ReceiveAsyncOptions()
             .setEnableAutoComplete(false);
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClient.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClient.java
@@ -473,18 +473,17 @@ public final class ServiceBusReceiverAsyncClient implements AutoCloseable {
             return monoError(logger, new IllegalArgumentException("'message.lockToken' cannot be empty."));
         }
 
-        UUID lockTokenUuid = null;
+        final UUID lockTokenUuid;
         try {
             lockTokenUuid = UUID.fromString(lockToken.getLockToken());
         } catch (IllegalArgumentException ex) {
-            monoError(logger, ex);
+            return monoError(logger, ex);
         }
 
-        UUID finalLockTokenUuid = lockTokenUuid;
         return connectionProcessor
             .flatMap(connection -> connection.getManagementNode(entityPath, entityType, sessionId))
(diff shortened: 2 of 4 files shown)


sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
```diff
@@ -432,7 +433,7 @@ public ServiceBusSenderAsyncClient buildAsyncClient() {
                 case QUEUE:
                     entityName = queueName;
                     break;
-                case TOPIC:
+                case SUBSCRIPTION:
                     entityName = topicName;
```
hemanttanwar, line 437: It is confusing to call it `SUBSCRIPTION` in sender. But I understand why you have done it.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
```diff
@@ -476,11 +523,14 @@ void sendReceiveMessageWithVariousPropertyTypes() {
         sentProperties.put("UUIDProperty", UUID.randomUUID());
         sentProperties.put("StringProperty", "string");
 
-        sender.send(messageToSend).block(TIMEOUT);
+        sender.send(messageToSend);
```
hemanttanwar, line 527: Did you miss using `sendMessage(messageToSend)` here  and  adding timeout?

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusSenderClientIntegrationTest.java
```diff
@@ -62,6 +96,47 @@ void nonSessionMessageBatch() {
...
         }
+
         sender.send(batch);
+
+        for (int i = 0; i < messages.size(); i++) {
+            messagesPending.incrementAndGet();
+        }
+    }
+
+    void setSenderAndReceiver(MessagingEntityType entityType) {
```
hemanttanwar, line 107: I can see that quickly we will have to change this to include other input parameters, for example
queuename / topic names , sessionId for receivers . So instead of taking taking just one parameters `entityType`, it should take a bag of things and  use that  bag to create receivers.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
```diff
@@ -22,178 +24,165 @@
...
-        final ServiceBusMessage message2 = TestUtils.getServiceBusMessage(CONTENTS, messageId2, 0);
-        final Duration duration = Duration.ofSeconds(10);
-        final Instant scheduledEnqueueTime = Instant.now().plus(duration);
         final ReceiveAsyncOptions options = new ReceiveAsyncOptions().setEnableAutoComplete(false);
 
-        sender.scheduleMessage(message1, scheduledEnqueueTime)
-            .block(TIMEOUT);
-        sender.scheduleMessage(message2, scheduledEnqueueTime)
-            .block(TIMEOUT);
+        sender.scheduleMessage(message, scheduledEnqueueTime).block();
```
hemanttanwar, line 165: Why we are not using TIMEOUT ?

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusSenderClientIntegrationTest.java
conniey: There is an overload that takes: `Function<ServiceBusReceiverClientBuilder, ServiceBusReceiverClientBuilder> onCreate`, where you can apply any other custom modifications.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClientIntegrationTest.java
conniey: In IntegrationTestBase, we set the `StepVerifier.setDefaultTimeout()`, so it'll use that. Practically, we can remove all the .block(TIMEOUT).

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
conniey: Yes. To match the entity type passed to Service Bus.

PR #10023, opened 2020-04-09
Remove account name from host

sima-zhu:
The task is to hide the account name from playback jsons.

I have checked the record and playback in appConfig. (Both passed)
Record host like below:
```
"Uri" : "https://REDACTED.azconfig.io/kv/key2899449606b?api-version=1.0"
```

First phase is to update the record policy.
Then we can re-record for all client libraries.

### DIFF
--- a/sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
+++ b/sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
@@ -82,7 +82,7 @@ public class RecordNetworkCallPolicy implements HttpPipelinePolicy {
 
         // Remove sensitive information such as SAS token signatures from the recording.
         UrlBuilder urlBuilder = UrlBuilder.parse(context.getHttpRequest().getUrl());
-        removeAccountName(urlBuilder);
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core-test/src/main/java/com/azure/core/test/policy/RecordNetworkCallPolicy.java
```diff
@@ -112,6 +113,13 @@ public RecordNetworkCallPolicy(RecordedData recordedData) {
             });
     }
 
+    private void redactedAccountName(UrlBuilder urlBuilder) {
+        String[] hostParts = urlBuilder.getHost().split("\\.");
+        hostParts[0] = "REDACTED";
```
alzimmermsft, line 118: From what I remember when we are associating requests to responses when using the playback records we explicitly ignore the entire host string. Should be able to simplify the logic to redact the entire host string.

sima-zhu: That's true. 
My opinion is to leave insensitive data over there as Json file is also a good and quick reference of what the request and response are composed of. Also, it is cheap to have the information over there.

I am flexible of making change to replace all.

Issue #9982, opened 2020-04-08
HTTP respoonse has a malformed body [BUG]


bryandmc:
**Describe the bug**
Version 12.5.0 of the azure-storage-blob library throws this exception (below) upon simple usage of the SDK. It's somewhat unclear because of the stack below which of 2 possible calls to the SDK caused this. It's either after calling `blobContainerClient.getBlobContainerurl()` or `blobServiceClient.generateAccountSas(..)`. These calls all worked under the previous version we used (12.3.0). We have changed our dependency back to that in the short-term to deal with the issue. Unfortunately the code in question is closed-source so I can't provide long detailed snippets from our codebase, but it should be naively reproducible by just using that version with one of these calls. It happens every time in our situation. 

***Exception or Stack Trace***
```
com.azure.core.exception.HttpResponseException: HTTP response has a malformed body.
	at com.azure.core.implementation.serializer.HttpResponseBodyDecoder.lambda$decode$1(HttpResponseBodyDecoder.java:95) ~[factory.jar:master]
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:118) ~[factory.jar:master]
	at reactor.core.publisher.FluxDoFinally$DoFinallySubscriber.onNext(FluxDoFinally.java:123) ~[factory.jar:master]
	at reactor.core.publisher.FluxHandle$HandleSubscriber.onNext(FluxHandle.java:112) ~[factory.jar:master]
	at reactor.core.publisher.FluxMap$MapConditionalSubscriber.onNext(FluxMap.java:213) ~[factory.jar:master]
	at reactor.core.publisher.FluxDoFinally$DoFinallySubscriber.onNext(FluxDoFinally.java:123) ~[factory.jar:master]
	at reactor.core.publisher.FluxHandleFuseable$HandleFuseableSubscriber.onNext(FluxHandleFuseable.java:178) ~[factory.jar:master]
	at reactor.core.publisher.FluxContextStart$ContextStartSubscriber.onNext(FluxContextStart.java:103) ~[factory.jar:master]
	at reactor.core.publisher.Operators$MonoSubscriber.complete(Operators.java:1592) ~[factory.jar:master]
	at reactor.core.publisher.MonoCollectList$MonoCollectListSubscriber.onComplete(MonoCollectList.java:121) ~[factory.jar:master]
[...]

PR #10027, opened 2020-04-09
Update HttpResponse Default getBodyAsString Charset

### LINKED ISSUE #8710: Review HttpResponse Handling of getBodyAsString
In a recent review it was determined that `BufferedHttpResponse` assumes that the `Charset` is `UTF-8` when using the `getBodyAsString` method. Digging into this further the Netty and OkHttp implementations of the `HttpResponse` interface will default to what the underlying library uses. These implementations should be reviewed and standardized to prevent bugs from occurring if a different `HttpResponse` is returned.

alzimmermsft:
Fixes #8710 #9982

This PR updates how the Netty and JDK `HttpClient`s handle encoding the `byte` stream to a `String` when a `Charset` isn't passed. The logic change is the following:

1. Check the byte stream for a byte order mark (BOM). If a BOM exists use the specified encoding and strip those leading bytes from the returned string.
2. Check for a `Content-Encoding` response header. If a `Content-Encoding` header is set use that encoding.
3. Default to using `UTF-8`.

The OkHttp `HttpClient` didn't need to be updated as OkHttp handles this when getting the body as String.

### DIFF
--- a/sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
+++ b/sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java
@@ -9,6 +9,7 @@ import com.azure.core.http.HttpHeaders;
 import com.azure.core.http.HttpMethod;
 import com.azure.core.http.HttpRequest;
 import com.azure.core.http.HttpResponse;
+import com.azure.core.util.CoreUtils;
 import okhttp3.Call;
 import okhttp3.Headers;
 import okhttp3.MediaType;
@@ -250,21 +251,8 @@ class OkHttpAsyncHttpClient implements HttpClient {
 
         @Override
         public Mono<String> getBodyAsString() {
-            return Mono.fromCallable(() -> {
-                // Reactor: The fromCallable operator treats a null from the Callable
-                // as completion signal.
-                if (responseBody == null) {
-                    return null;
-                }
-                String content = responseBody.string();
-                // Consistent with GAed behaviour.
-                if (content.length() == 0) {
-                    return null;
-                }
-                // OkHttp: When calling ResponseBody::string() the underlying stream automatically closed.
-                // https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response-body/#the-response-body-must-be-closed
-                return content;
-            });
+            return getBodyAsByteArray()
+                .map(bytes -> CoreUtils.bomAwareToString(bytes, headers.getValue("Content-Type")));
         }
 
         @Override
--- a/sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/HttpClientTests.java
+++ b/sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/HttpClientTests.java
@@ -48,6 +48,9 @@ public abstract class HttpClientTests {
      */
     protected abstract int getWireMockPort();
 
+    /**
+     * Tests that a response without a byte order mark or a 'Content-Type' header encodes using UTF-8.
+     */
     @Test
     public void plainResponse() {
         String expected = new String(EXPECTED_RETURN_BYTES, StandardCharsets.UTF_8);
@@ -57,6 +60,9 @@ public abstract class HttpClientTests {
             .verifyComplete();
     }
 
+    /**
+     * Tests that a response with a 'Content-Type' header encodes using the specified charset.
+     */
     @Test
     public void headerResponse() {
         String expected = new String(EXPECTED_RETURN_BYTES, StandardCharsets.UTF_16BE);
@@ -66,6 +72,9 @@ public abstract class HttpClientTests {
             .verifyComplete();
     }
 
+    /**
+     * Tests that a response with a 'Content-Type' containing an invalid or unsupported charset encodes using UTF-8.
+     */
     @Test
     public void invalidHeaderResponse() {
         String expected = new String(EXPECTED_RETURN_BYTES, StandardCharsets.UTF_8);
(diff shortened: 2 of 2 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/JacksonAdapter.java
```diff
@@ -66,7 +65,7 @@
      * BOM header from some response bodies. To be removed in deserialization.
      */
     private static final String BOM = "\uFEFF";
-    private static final String BOM_STRING = new String(BOM.getBytes(StandardCharsets.UTF_8), Charset.defaultCharset());
+    private static final String BOM_STRING = new String(BOM.getBytes(StandardCharsets.UTF_8), StandardCharsets.UTF_8);
```
anuchandy, line 68: Do we need this BOM stripping on already stripped-and-decoded string anymore?

alzimmermsft: Shouldn't need it anymore, will add tests into the share `RestProxyTests` that will validate encoding handling for all clients.

sdk/core/azure-core-http-jdk-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
```diff
@@ -216,14 +216,36 @@ public HttpHeaders getHeaders() {
...
         public Mono<String> getBodyAsString() {
-            return getBodyAsByteArray()
-                .map(bytes -> new String(bytes, StandardCharsets.UTF_8));
+            return getBodyAsByteArray().map(bytes -> {
+                if (bytes.length >= 3 && bytes[0] == (byte) 239 && bytes[1] == (byte) 187 && bytes[2] == (byte) 191) {
+                    return new String(bytes, 3, bytes.length - 3, StandardCharsets.UTF_8);
+                } else if (bytes.length >= 2 && bytes[0] == (byte) 254 && bytes[1] == (byte) 255) {
+                    return new String(bytes, 2, bytes.length - 2, StandardCharsets.UTF_16BE);
+                } else if (bytes.length >= 2 && bytes[0] == (byte) 255 && bytes[1] == (byte) 254) {
+                    return new String(bytes, 2, bytes.length - 2, StandardCharsets.UTF_16LE);
```
srnagar, line 225: Any reason why we look for only these 3 charsets and not others? Adding a comment here to explain that would be good.

anuchandy:
Do we need to apply a similar BOM lookup here as well?

https://github.com/Azure/azure-sdk-for-java/blob/88adf363b4a073b9823bfa7fb939c18fffb68278/sdk/core/azure-core/src/main/java/com/azure/core/implementation/http/BufferedHttpResponse.java#L64

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/JacksonAdapter.java
anuchandy: Do we need to apply a similar BOM lookup here as well?

https://github.com/Azure/azure-sdk-for-java/blob/88adf363b4a073b9823bfa7fb939c18fffb68278/sdk/core/azure-core/src/main/java/com/azure/core/implementation/http/BufferedHttpResponse.java#L63

anuchandy: If so, wondering whether it makes sense to move this to FluxUtils?  something like: 

```java
Mono<String> FluxUtils.toBomAwarePublisher(
    Flux<ByteBuffer> stream,  
    String contentEncoding)
```

where `contentEncoding` is the value of Content-Encoding header, if no BOM prefix in-stream and if `contentEncoding` is null then method uses utf8 as default.

sdk/core/azure-core-http-netty/src/main/java/com/azure/core/http/netty/NettyAsyncHttpClient.java
```diff
@@ -211,9 +214,31 @@ public HttpHeaders getHeaders() {
 
         @Override
         public Mono<String> getBodyAsString() {
-            return bodyIntern().aggregate().asString().doFinally(s -> {
-                if (!reactorNettyConnection.isDisposed()) {
-                    reactorNettyConnection.channel().eventLoop().execute(reactorNettyConnection::dispose);
+            return getBodyAsByteArray().map(bytes -> {
```
srnagar, line 217: Both JDK client and Netty client has to do the same logic for converting response byte array to string. It would be better to put this in Core utils somewhere to reduce duplication and if there are any fixes or updates to this logic, we don't need to update in two places.

alzimmermsft: Yeah I was thinking about that, also would give the opportunity to test is more directly and thoroughly.

sdk/core/azure-core-http-jdk-httpclient/src/main/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClient.java
alzimmermsft: Added support for `UTF-32BE` and `UTF-32LE`

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/JacksonAdapter.java
alzimmermsft: Added helper into `CoreUtils`

PR #10030, opened 2020-04-09
Fix Checkstyle Suppressions for azure-storage-perf

alzimmermsft:
Fixes the Checkstyle suppressions for `azure-storage-perf` that weren't updated when the module was renamed to following naming conventions.


kasobol-msft:
Thank you!

PR #10031, opened 2020-04-09
Initial proposal for processor model for multi session

hemanttanwar:
This PR  support following use cases. This PR only have API and no implementation.

**Use case to support**

1. Receive from one named or un-known session  ( No session roll over)
  (Gist : https://gist.github.com/hemanttanwar/92721fb10882f17d4f322572a2073a7b )
  **How user would implement it** : Example is here https://github.com/Azure/azure-sdk-for-java/pull/10031/files#diff-cf4589dfe9d1010fcde96f905f061e5cR30

2. Multi Session roll-over Support  
    We are proposing to follow Event Hub processor model.
( Gist : https://gist.github.com/hemanttanwar/5e7fdb7a28b615c209b4188a16012e74 )

**Requirement** :

  1. User should be able to receive message from multiple unknown sessions.
  2. User should be able to settle the message.
  3. There is no requirement for user to renew the session lock, as long as SDK will renew in background.
**How user would implement it** : Example is here https://github.com/Azure/azure-sdk-for-java/pull/10031/files#diff-32e2cd46278018bdd216c9b605324ea0R17
More alternative option to support Multiple Session and  their Pro/Con are discussed  here ( https://gist.github.com/hemanttanwar/5e7fdb7a28b615c209b4188a16012e74)

**Why not merging single named session receiver with multi-session rollover use case in one client**

If we merge "single named session receive" with Multi session rollover EH Processor model, We will not be able to satisfy following use case for single named session-id.
1. Other features like normal receive example peek, peekBatch, receiveDeferred ( These API not applicable for multi-session)
2. Sync pull API
3. We also do not want to pollute Multi session rollover like EH Processor model with feature for "single named session-id".

So I still think we should not merge "Single named session" use case with Multisession-rollover use case.

**Does single unknown session and roll-over to next available session fall in this use case?**

No. Here is why .. 

1. As soon as we start rolling-over to next session, user need to understand that it is different scenario, because renew lock on different session-id. 
2. And if one session fails and we roll over user need to know it is not terminating situation. All the above make case for keep single session id to strictly single session and not rolling over.

Same comment in JS for this.

https://github.com/Azure/azure-sdk-for-js/pull/8255/files#r408466922

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ServiceBusMultiSessionProcessorSample.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ServiceBusMultiSessionProcessorSample.java
@@ -34,10 +34,11 @@ public class ServiceBusMultiSessionProcessorSample {
         Disposable subscription = multiSessionReceiverAsyncClient.receiveMultiSession(options)
             .flatMap(receivedMessage -> {
                /* System.out.println("Session State : " + multiSessionReceiverAsyncClient.getSessionState(
-                    receivedMessage.getSessionId()));*/
+                    receivedMessage.getSessionId())); */
                 System.out.println("!!! Received Message seq no  "+ receivedMessage.getSequenceNumber() +" SessionId " +  receivedMessage.getSessionId() + " Data : " +  new String (receivedMessage.getBody()));
                 if (receivedMessage.isSessionError()) {
-                    myMessageProcessor.processError(receivedMessage.getServiceBusErrorContext());
+                    System.out.println("!!! Error Message session id" +  receivedMessage.getServiceBusErrorContext().getSessionId());
+                    //myMessageProcessor.processError(receivedMessage.getServiceBusErrorContext());
                     return Mono.empty();
                 }
                 return Mono.empty();
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceivedMessage.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceivedMessage.java
@@ -3,7 +3,6 @@
 
 package com.azure.messaging.servicebus;
 
-import com.azure.core.util.CoreUtils;
 import com.azure.messaging.servicebus.models.ReceiveMode;
 import com.azure.messaging.servicebus.models.ServiceBusErrorContext;
 
@@ -46,6 +45,12 @@ public final class ServiceBusReceivedMessage implements MessageLockToken {
     ServiceBusReceivedMessage(byte[] body) {
         this.body = Objects.requireNonNull(body, "'body' cannot be null.");
         this.properties = new HashMap<>();
+    }
+
+    ServiceBusReceivedMessage(ServiceBusErrorContext errorContext) {
+        this.serviceBusErrorContext = Objects.requireNonNull(errorContext, "'errorContext' cannot be null.");
+        this.properties = new HashMap<>();
+        this.body = null;
     }
 
     /**
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClient.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverAsyncClient.java
@@ -25,6 +25,7 @@ import com.azure.messaging.servicebus.implementation.ServiceBusManagementNode;
 import com.azure.messaging.servicebus.implementation.ServiceBusReceiveLinkProcessor;
 import com.azure.messaging.servicebus.models.ReceiveAsyncOptions;
 import com.azure.messaging.servicebus.models.ReceiveMode;
+import com.azure.messaging.servicebus.models.ServiceBusErrorContext;
 import reactor.core.publisher.BaseSubscriber;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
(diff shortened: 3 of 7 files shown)


sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ServiceBusMultiSessionProcessorSample.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class ServiceBusMultiSessionProcessorSample {
+
+    public static void main(String[] args) {
+        final MyMessageProcessor myMessageProcessor = new MyMessageProcessor();
+        ServiceBusMultiSessionProcessorClient multiSessionProcessorClient = new ServiceBusClientBuilder()
+            .connectionString("connectionString")
+            .multiSessionProcessor()
```
KieranBrantnerMagee, line 21: Why do you need to specify multisession here, and then again when you build() at the end?

hemanttanwar: I certainly have not put enough time on naming this class.
This duplication should be avoided, I should change `multiSessionProcessor()` to `buildProcessor()`

sdk/servicebus/azure-messaging-servicebus/src/samples/java/com/azure/messaging/servicebus/ServiceBusMultiSessionProcessorSample.java
```diff
@@ -0,0 +1,56 @@
...
+    static class MyMessageProcessor implements AutoCloseable {
+        private final Logger logger = LoggerFactory.getLogger(MyMessageProcessor.class);
+        private final AtomicBoolean isDisposed = new AtomicBoolean();
+
+        void onMessage(ServiceBusReceivedMessage message, SessionMessageManager manager) {
+            System.out.println("Process message here. Session id : " + message.getSessionId()
+                + ", message Id :" + message.getMessageId());
+            manager.complete(message);
+        }
+        void onError(ServiceBusErrorContext errorContext) {
```
KieranBrantnerMagee, line 44: Does this give enough granularity for various failure modes, or for the user to take the desired action?  Failure on a specific session, (where they may need session details) vs failure all up, transient vs persistent, etc?

hemanttanwar: Will add more granularity , I see dotnet have error code for different errors. I have not gone through all the error scenario.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
```diff
@@ -376,6 +400,170 @@ private static MessagingEntityType validateEntityPaths(ClientLogger logger, Stri
...
+         * @return {@link ServiceBusSessionReceiverAsyncClient} instance.
+         */
+        public ServiceBusSessionReceiverAsyncClient buildAsyncClient() {
+            return null;
+        }
+        public ServiceBusSessionReceiverAsyncClientBuilder receiveMode(ReceiveMode receiveMode) {
+            this.receiveMode = receiveMode;
+            return this;
+        }
+        public ServiceBusSessionReceiverAsyncClientBuilder sessionId(String sessionId) {
```
richardpark-msft, line 425: null would be acceptable here? And then it's just "get next available session" behavior?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
```diff
@@ -376,6 +400,170 @@ private static MessagingEntityType validateEntityPaths(ClientLogger logger, Stri
...
+
+        /**
+         * Sets the name of the topic.
+         *
+         * @param topicName Name of the topic.
+         *
+         * @return The modified {@link ServiceBusMultiSessionProcessorClientBuilder} object.
+         * @see #subscriptionName A subscription name should be set as well.
+         */
+        public ServiceBusSessionReceiverAsyncClientBuilder topicName(String topicName) {
```
richardpark-msft, line 463: I think I've asked you this before but can you not just have a builder method that takes (topicName, subscriptionName) so you don't ever omit one of them on accident?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
```diff
@@ -376,6 +400,170 @@ private static MessagingEntityType validateEntityPaths(ClientLogger logger, Stri
...
+        }
+
+        /**
+         * The function that is called when an error occurs while processing messages. The input contains exception
+         * when the error happened.
+         *
+         * @param processError The callback that's called when an error occurs while processing messages.
+         * @return The updated {@link ServiceBusMultiSessionProcessorClientBuilder} instance.
+         */
+        public ServiceBusMultiSessionProcessorClientBuilder processError(Consumer<ServiceBusErrorContext> processError) {
```
richardpark-msft, line 561: Would `ServiceBusErrorContext` contain information like the session ID or something of that nature?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMultiSessionProcessorClient.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+    public  boolean isRunning() { return isRunning;
+    }
+
+    /**
+     * Disposes of the {@link ServiceBusMultiSessionProcessorClient}. If the client had a dedicated connection, the underlying
+     * connection is also closed.
+     */
+    @Override
+    public void close() {
```
richardpark-msft: Can I 'undo' close() with 'start()'?

hemanttanwar:
Closing this as we have decided not to implement processor model.

PR #10044, opened 2020-04-10
Smoke Tests: install cosmos

danieljurek:


danieljurek:
/azp run 
java - smoke-tests

PR #10049, opened 2020-04-10
fix: resource test checkstyle

ChenTanyi:


sdk/resources/mgmt/src/test/java/com/azure/management/resources/DeploymentsTests.java
```diff
@@ -116,7 +116,7 @@ public void canPostDeploymentWhatIfOnResourceGroup() throws Exception {
                 .whatIf();
 
         Assertions.assertEquals("Succeeded", result.status());
-        // FIXME: regression?
+        // TODO(not known): regression?
         // Assertions.assertEquals(3, result.changes().size());
```
weidongxu-microsoft, line 120: Could you try in PLAYBACK whether these are now OK?

ChenTanyi: Done, also resolved another TODO

PR #10050, opened 2020-04-10
Revision to CONTRIBUTING.md after POM restructure.

mitchdenny:
This PR revises the instructions in ```CONTRIBUTING.md``` to be accurate now that we have restructured the POM files. Folks should no longer be invoking Maven via the ```pom.client.xml``` file and instead should use the root aggregate and the ```-pl``` switch, or if they prefer install dependencies via the root POM an use the ```pom.xml``` in each service directory.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,7 +80,12 @@ mvn -Dgpg.skip -DskipTests -pl com.azure:azure-messaging-eventhubs -am
 ```
 
 The ```-pl``` switch takes a comma seperated list of fully qualified module names and the ```-am```
-switch tells Maven to _also make_ all of the dependencies within the repository that it depends on.
+switch tells Maven to _also make_ all of the dependencies within the repository that it depends on. If
+you are making changes across two modules you can tell Maven to build them both like this:
+
+```
+mvn -Dgpg.skip -DskipTests -pl com.azure:azure-messaging-eventhubs,com.azure:azure-core-amqp -am
+```
 
 Some parts of the SDK have multiple modules for a particular service, if you want to avoid typing
 in all those module names for the ```-pl``` switch you can just target the appropriate POM file


CONTRIBUTING.md
```diff
@@ -97,7 +116,7 @@ running live tests.
 To run live tests against a service after deploying live resources:
 
 ```
-mvn -f sdk/keyvault/pom.service.xml -Dmaven.wagon.http.pool=false --batch-mode --fail-at-end --settings eng/settings.xml test
+mvn -f sdk/keyvault/pom.xml -Dmaven.wagon.http.pool=false --batch-mode --fail-at-end --settings eng/settings.xml test
```
sima-zhu, line 119: Why don't we use -pl -am for live tests here? Or mention dependencies need to cache in local before run this?

CONTRIBUTING.md
```diff
@@ -56,27 +56,46 @@ Merging Pull Requests (for project contributors with write access)
...
 
-The easiest way to build is by running the following command from the root folder:
+>**Note**: Build and test instructions have recently changed. The ```pom.client.xml``` file is no
+> longer in the root of the repository and should be not be used as an aggregator moving forward.
+
+This repository uses Maven to build the various modules that make up the Azure SDK for Java. You can build the entire set of modules by running the following command from the root of the repository:
+
 ```
-mvn -f pom.client.xml -Dgpg.skip -DskipTests -Dinclude-non-shipping-modules clean install
+mvn -Dgpg.skip clean install
```
sima-zhu, line 65: It is much cleaner! Hooray!

CONTRIBUTING.md
```diff
@@ -56,27 +56,46 @@ Merging Pull Requests (for project contributors with write access)
...
 
-### Compiling one project only
+```
+mvn -Dgpg.skip -DskipTests clean install
+```
 
+If you only want to work on one module, you may opt to build just that module and its dependencies (e.g. com.azure:azure-messaging-eventhubs):
+
+```
+mvn -Dgpg.skip -DskipTests -pl com.azure:azure-messaging-eventhubs -am
```
sima-zhu, line 79: Is it better to have another example on how to build two modules at once?
e.g.
-pl com.azure:azure-storage-blob, com.azure:azure-storage-file

CONTRIBUTING.md
mitchdenny: Let's say you are working on EventHubs ... do you really want to spin up all the storage resources for your EventHubs test (because of checkpoint store?). Nothing here is stopping you do it but I'm trying to make it simple for folks.

Of course if you are making changes across storage and event hubs then you'll probably want to do just that. Although in that case you'd probably run Maven against the root and specify both sets of modules on the -pl switch.

CONTRIBUTING.md
mitchdenny: Added an example.

PR #10056, opened 2020-04-10
mgmt config loglevel

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1074

For samples, currently I set them uniformly to BASIC.

### DIFF
--- a/sdk/resources/mgmt/src/main/java/com/azure/management/RestClientBuilder.java
+++ b/sdk/resources/mgmt/src/main/java/com/azure/management/RestClientBuilder.java
@@ -133,6 +133,7 @@ public final class RestClientBuilder {
      * @return builder
      */
     public RestClientBuilder withHttpLogOptions(HttpLogOptions logOptions) {
+        Objects.requireNonNull(logOptions);
         httpLogOptions = logOptions;
         return this;
     }


sdk/resources/mgmt/src/main/java/com/azure/management/RestClientBuilder.java
```diff
@@ -50,7 +51,7 @@
      */
     public RestClientBuilder() {
         retryPolicy = new RetryPolicy();
-        httpLogOptions = new HttpLogOptions().setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS);
+        httpLogOptions = new HttpLogOptions().setLogLevel(HttpLogDetailLevel.NONE);
```
weidongxu-microsoft, line 54: I've changed the default level to NONE.

1. it is consistent with track1
2. I think NONE should be the default.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/AzureConfigurable.java
```diff
@@ -18,12 +19,23 @@
...
+     * @param logOptions the HttpLogDetailLevel logging options
      * @return the configurable object itself
      */
-    T withLogOptions(HttpLogOptions level);
+    T withLogOptions(HttpLogOptions logOptions);
+
+    /**
+     * Sets the logging detail level on the HTTP client.
+     *
+     * If set, this configure will override {@link HttpLogOptions#setLogLevel(HttpLogDetailLevel)} configure of
```
weidongxu-microsoft, line 32: Javadoc to notify that this will take precedence if set.

sdk/resources/mgmt/src/main/java/com/azure/management/RestClientBuilder.java
yungezz: interesting that track1 default NONE..

weidongxu-microsoft: Always a potential security risk if print out too many things. It is less risky for track2 log, since by default only a few header/query is white-listed.

PR #10082, opened 2020-04-10
fix checkstyle for management authorization

xseeseesee:


sdk/authorization/mgmt/src/main/java/com/azure/management/graphrbac/AddOwnerParameters.java
```diff
@@ -13,7 +13,9 @@
 public final class AddOwnerParameters {
     /*
      * A owner object URL, such as
-     * "https://graph.windows.net/0b1f9851-1bf0-433f-aec3-cb9272f093dc/directoryObjects/f260bbc4-c254-447b-94cf-293b5ec434dd", where "0b1f9851-1bf0-433f-aec3-cb9272f093dc" is the tenantId and "f260bbc4-c254-447b-94cf-293b5ec434dd" is the objectId of the owner (user, application, servicePrincipal, group) to be added.
```
weidongxu-microsoft, line 20: Em... I think an updated generator has already fixed this... Let me know if yo still see this when doing other services.

PR #10088, opened 2020-04-10
Moved permission type to implementation package

kushagraThapar:
* Moved permission type to implementation package
* Using CosmosPermissionProperties.java on public surface

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
@@ -4,8 +4,9 @@ package com.azure.cosmos;
 
 import com.azure.core.annotation.ServiceClientBuilder;
 import com.azure.cosmos.implementation.Configs;
-import com.azure.cosmos.models.Permission;
+import com.azure.cosmos.implementation.Permission;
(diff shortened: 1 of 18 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -225,7 +226,7 @@ public CosmosClientBuilder resourceToken(String resourceToken) {
      * @param permissions Permission list for authentication.
      * @return current Builder.
      */
-    public CosmosClientBuilder permissions(List<Permission> permissions) {
+    public CosmosClientBuilder permissions(List<CosmosPermissionProperties> permissions) {
```
moderakh, line 229: The existing tests all are wired up against Permission (internal type) not CosmosPermissionProperties, So I I think we don't have any tests covering this API. This can be added as separate independent PR.

kushagraThapar: Yes, I am already working on those. Will come up with a new PR for those.

kushagraThapar: @moderakh - Here is the PR which fixes all the tests: https://github.com/Azure/azure-sdk-for-java/pull/10100

PR #10089, opened 2020-04-10
Fix Incorrect Module Paths

alzimmermsft:


alzimmermsft:
@mitchdenny @JimSuplizio @JonathanGiles 

I don't think there is a pipeline associated to the client parent POM, nor any of the parent POMs, should we configure one?

PR #10114, opened 2020-04-13
mgmt checkstyle azure

weidongxu-microsoft:


sdk/management/azure/src/main/java/com/azure/management/Azure.java
```diff
@@ -405,575 +405,474 @@ public Azure withDefaultSubscription() throws CloudException, IOException {
...
+
+    /** @return entry point to managing web apps. */
     public WebApps webApps() {
         return appServiceManager.webApps();
     }
 
-    /**
-     * @return entry point to managing app services.
-     */
-    public AppServiceManager appServices() {
```
weidongxu-microsoft, line 774: Class in implementation is not allowed to be exposed. So use common approach of exposing all its sub interface instead.

PR #10116, opened 2020-04-13
Network: fix checkstyle

ChenTanyi:


sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkSecurityRuleImpl.java
```diff
@@ -162,7 +162,7 @@ public NetworkSecurityRuleImpl withProtocol(SecurityRuleProtocol protocol) {
 
     @Override
     public NetworkSecurityRuleImpl withAnyProtocol() {
-        return this.withProtocol(SecurityRuleProtocol._);
+        return this.withProtocol(SecurityRuleProtocol.fromString("*"));
```
weidongxu-microsoft, line 155: Better still use the enum

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkInterfaceImpl.java
```diff
@@ -69,7 +69,7 @@
 
     @Override
     protected Mono<NetworkInterfaceInner> getInnerAsync() {
-        // FIXME: 3rd parameter
+        // TODO(not known): 3rd parameter
```
weidongxu-microsoft, line 72: This TODO can be solved now

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkSecurityRuleImpl.java
ChenTanyi: I think from string will not change at all. Instead, the enum name is still not ready after discussed.

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkSecurityGroupImpl.java
```diff
@@ -80,7 +80,7 @@ public NetworkSecurityRuleImpl defineRule(String name) {
 
     @Override
     protected Mono<NetworkSecurityGroupInner> getInnerAsync() {
-        // FIXME: parameter
+        // TODO(not known): parameter
```
weidongxu-microsoft, line 83: Should be solved.

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkInterfaceImpl.java
ChenTanyi: Should all expand TODO remain `null`?

ChenTanyi:
All comments are done.

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkInterfaceImpl.java
weidongxu-microsoft: If the call is using `null` it could now be deleted.

This is yaohai adding these `null`s when method call with defaults not ready.

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkSecurityRuleImpl.java
weidongxu-microsoft: The point is when generator changes it, no compiler will tell anyone anything about this line.

ChenTanyi: Yes, I change it back. But I think the `any` should not be other char than '*'.

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkInterfaceImpl.java
ChenTanyi: done

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkSecurityGroupImpl.java
ChenTanyi: done

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/LoadBalancerImpl.java
```diff
@@ -79,7 +79,6 @@
 
     @Override
     protected Mono<LoadBalancerInner> getInnerAsync() {
-        // TODO(not known): Extra parameter
```
weidongxu-microsoft: Eh, I mean, you could try delete the `null` in `.getByResourceGroupAsync(this.resourceGroupName(), this.name(), null);`. If it compiles then all is good. If not let me know. It should works same on ``.getByResourceGroupAsync(this.resourceGroupName(), this.name());`

ChenTanyi: Done

PR #10119, opened 2020-04-13
Sample: fix checkstyle

ChenTanyi:


sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppStorageAccountConnection.java
```diff
@@ -86,7 +86,7 @@ public static boolean runSample(Azure azure) {
 
             System.out.println("Creating web app " + app1Name + "...");
 
-            // FIXME the env variable will not work in linux since dot is not allowed in env variable name
+            // TODO: the env variable will not work in linux since dot is not allowed in env variable name
```
weidongxu-microsoft, line 89: I think you can change the TODO to just a note. I've updated the WAR to fallback to env var with "_" if "." not found.

sdk/management/samples/src/main/java/com/azure/management/samples/DockerUtils.java
```diff
@@ -68,7 +69,8 @@ public DockerSSLConfig(String caPem, String keyPem, String certPem) {
                 sslConfig.keyStorePassword("docker");
                 sslConfig.trustStore(CertificateUtils.createTrustStore(caPem));
             } catch (Exception e) {
-                throw new DockerClientException(e.getMessage(), e);
+                ClientLogger logger = new ClientLogger(getClass());
```
weidongxu-microsoft, line 72: might not be the best place to define `logger`?

ChenTanyi: I have thought logger should not add in class since it is just a sample. Done, change it to comment rather than logger.

sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppStorageAccountConnection.java
ChenTanyi: Done

PR #10137, opened 2020-04-13
Fixed the failure of search nightly CI build

sima-zhu:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=354178&view=logs&j=3d7a0212-45d0-5080-1981-6d6ef159b344&t=14d7dc25-187c-516a-cf22-fdfcecf8f884

Steps on retry on search account creation:
1. Create service account.
1. Ping the service, and run test if succeeds, or ping the service again if fails. There is an 10 sec interval between every service ping which retry 3 times. 
1. If service ping still fail, then retry on search creation again. If succeeds, then run test. If fails on already exist exception, silent on exception, continue on running tests. 
1. If the creation API retry more than 3 times, then fail the test with message.

Statistics:
1. Usually, it will succeed on first creation. 
2. In worst case, it will use 90 sec before test start.

### DIFF
--- a/sdk/search/azure-search-documents/pom.xml
+++ b/sdk/search/azure-search-documents/pom.xml
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.5.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.4.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-test</artifactId>
-      <version>1.3.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core-test;dependency} -->
+      <version>1.2.0</version> <!-- {x-version-update;com.azure:azure-core-test;dependency} -->
       <scope>test</scope>
     </dependency>
     <dependency>


sdk/search/azure-search-documents/pom.xml
```diff
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.4.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+      <version>1.5.0-beta.1</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
```
alzimmermsft, line 32: Would rather not eagerly move to an unreleased version of Azure Core, is this being done to make testing the build from source issue more easily?

sima-zhu: I am not sure why. The search ci nightly build failed because of new azure core change. 
I don't quite get why the build test on new changes.

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=354178&view=logs&j=3d7a0212-45d0-5080-1981-6d6ef159b344&t=14d7dc25-187c-516a-cf22-fdfcecf8f884

sima-zhu: Search changes here is to test over the azure-core-test changes. Will revert back to azure-core and azure-core released version after testing.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/MockHttpResponse.java
```diff
@@ -182,7 +182,7 @@ public HttpHeaders getHeaders() {
 
         return bodyBytes == null
                 ? Mono.empty()
-                : Mono.just(new String(bodyBytes, charset));
+                : Mono.just(CoreUtils.bomAwareToString(bodyBytes, getHeaderValue("Content-Type")));
```
alzimmermsft, line 185: You'll want to move this into the `getBodyAsString` overload without parameters, nothing I've found in Azure Core leverages the `Charset` overload.

alzimmermsft:
```java
public Mono<String> getBodyAsString() {
  return (bodyBytes == null)
    ? Mono.empty()
    : Mono.just(CoreUtils.bomAwareToString(bodyBytes, getHeaderValue("Content-Type"));
}

File: sdk/search/azure-search-documents/pom.xml (hunk 1 above)
Comment (srnagar) [inline sdk/search/azure-search-documents/pom.xml:32]: Shouldn't the version tag also be updated when we switch to unreleased dependency?

File: sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/setup/AzureSearchResources.java [hunk 3]
```diff
@@ -114,13 +115,21 @@ public void createService(TestResourceNamer testResourceNamer) {
...
-                .withFreeSku()
-                .create();
-            recreateCount += 1;
+            try {
+                searchService = azure.searchServices()
+                    .define(searchServiceName)
+                    .withRegion(location)
+                    .withExistingResourceGroup(resourceGroup)
+                    .withFreeSku()
+                    .create();
```
srnagar, line 124: If create succeeded, then you can break out of the `do-while` loop.

sdk/search/azure-search-documents/pom.xml
sima-zhu: This is for test purpose. I am going to revert back to the released version as we do not depend on any new Version functionality for now.

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/setup/AzureSearchResources.java
sima-zhu: The problem is the management plan does not provide us the status of create. I have to ping the service to confirm. The newly added code is to avoid failure of already exists.

alzimmermsft: Could we check the `status` of the `SearchService` object?

sima-zhu: Just checked over the API. Find the status one this time. Missing from the initial investigate.
Will try to switch to status() checking

sima-zhu: I have tried the API they provided. Seems not that stable as service ping. In order to make our live tests more stable. I decided to switch back to ping machenism.

sima-zhu: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=356026&view=results

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/setup/AzureSearchResources.java
```diff
@@ -114,13 +115,21 @@ public void createService(TestResourceNamer testResourceNamer) {
...
         do {
-            searchService = azure.searchServices()
-                .define(searchServiceName)
-                .withRegion(location)
-                .withExistingResourceGroup(resourceGroup)
-                .withFreeSku()
-                .create();
-            recreateCount += 1;
+            try {
+                searchService = azure.searchServices()
```
srnagar, line 119: We should add some delay between each retry instead of retrying with no delay.

sima-zhu: There is delay of 30 seconds on every .create() operation.
Put it in shouldRetry() helper.

Issue #10096, opened 2020-04-10
EventHubClientBuilder parses wrongly HTTP_PROXY URL


stan-antyufeev-8451:
**Describe the bug**
If HTTP_PROXY sets to an URL format of  "http://host:port" then the EventHubClientBuilder class  will fail of creation  of an instance of EventHubProducerClient type

***Exception or Stack Trace***
[ERROR] 2020-04-10 16:53:32.353 [main] MessagingNotificationService -  
java.lang.NumberFormatException: For input string: "//proxy-host"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65) ~[?:?]
	at java.lang.Integer.parseInt(Integer.java:638) ~[?:?]
	at java.lang.Integer.parseInt(Integer.java:770) ~[?:?]
	at com.azure.messaging.eventhubs.EventHubClientBuilder.getDefaultProxyConfiguration(EventHubClientBuilder.java:633) ~[azure-messaging-eventhubs-5.0.3.jar:?]
	at com.azure.messaging.eventhubs.EventHubClientBuilder.getConnectionOptions(EventHubClientBuilder.java:596) ~[azure-messaging-eventhubs-5.0.3.jar:?]
	at com.azure.messaging.eventhubs.EventHubClientBuilder.buildConnectionProcessor(EventHubClientBuilder.java:543) ~[azure-messaging-eventhubs-5.0.3.jar:?]
	at com.azure.messaging.eventhubs.EventHubClientBuilder.buildAsyncClient(EventHubClientBuilder.java:480) ~[azure-messaging-eventhubs-5.0.3.jar:?]
	at com.azure.messaging.eventhubs.EventHubClientBuilder.buildClient(EventHubClientBuilder.java:514) ~[azure-messaging-eventhubs-5.0.3.jar:?]
	at com.azure.messaging.eventhubs.EventHubClientBuilder.buildProducerClient(EventHubClientBuilder.java:429) ~[azure-messaging-eventhubs-5.0.3.jar:?]

**To Reproduce**
Set environment variable HTTP_PROXY to any proxy URL in format of "http://hostname:port" and create EventHubProducerClient instance.

***Code Snippet***
None, use code snippets example from documentation

**Expected behavior**
Azure Event Hub client library should correctly parse HTTP_PROXY and HTTPS_PROXY URLs.

**Setup (please complete the following information):**
 - OS: MAC OS
 - IDE : Intellij
[...]

PR #10145, opened 2020-04-13
Fix default proxy configuration format

srnagar:
Currently, default proxy configuration string parsing logic expects the env variable to be of the format `hostname:port` but typically, the configuration in env variable is of the format `protocol://hostname:port`. 

This PR changes the parsing logic to check for either of these formats.

Fixes https://github.com/Azure/azure-sdk-for-java/issues/10096

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
@@ -117,6 +117,21 @@ public final class ServiceBusClientBuilder {
     }
 
     /**
+     * Sets the configuration store that is used during construction of the service client.
+     *
+     * If not specified, the default configuration store is used to configure Service Bus clients. Use
+     * {@link Configuration#NONE} to bypass using configuration settings during construction.
+     *
+     * @param configuration The configuration store used to configure Service Bus clients.
+     *
+     * @return The updated {@link ServiceBusClientBuilder} object.
+     */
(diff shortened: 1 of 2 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubClientBuilder.java
```diff
@@ -624,17 +627,26 @@ private ProxyOptions getDefaultProxyConfiguration(Configuration configuration) {
...
-        final String username = configuration.get(ProxyOptions.PROXY_USERNAME);
-        final String password = configuration.get(ProxyOptions.PROXY_PASSWORD);
+        return getProxyOptions(authentication, proxyAddress);
+    }
 
-        return new ProxyOptions(authentication, proxy, username, password);
+    private ProxyOptions getProxyOptions(ProxyAuthenticationType authentication, String proxyAddress) {
+        String host;
+        int port;
+        if (HOST_PORT_PATTERN.matcher(proxyAddress.trim()).find()) {
```
conniey, line 636: This logic doesn't exist in SB either.... would you mind copy and paste it there? :D Kill two birds with 3 stones?

srnagar: Sure, let me fix SB too.

conniey: You're the best!

PR #10146, opened 2020-04-14
Removed public APIs from FeedOptions for page size and continuation token

kushagraThapar:
* Reduced scope of public APIs for page size and continuation token to package-private. 
* Updated live tests to use CosmosPagedFlux `byPage(continuationToken, pageSize)` APIs

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/SinglePartitionDocumentQueryTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/SinglePartitionDocumentQueryTest.java
@@ -173,7 +173,7 @@ public class SinglePartitionDocumentQueryTest extends TestSuiteBase {
                 .requestChargeGreaterThanOrEqualTo(1.0).build())
             .build();
 
-        validateQuerySuccess(queryObservable.byPage(), validator);
+        validateQuerySuccess(queryObservable.byPage(maxItemCount), validator);
     }
 
     @Test(groups = { "simple" }, timeOut = TIMEOUT)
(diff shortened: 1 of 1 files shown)


sdk/cosmos/azure-cosmos-examples/src/main/java/com/azure/cosmos/rx/examples/multimaster/samples/Worker.java
```diff
@@ -85,7 +86,7 @@ public Worker(AsyncDocumentClient client, String databaseName, String collection
                 do {
 
                     FeedOptions options = new FeedOptions();
-                    options.setRequestContinuation(response != null ? response.getContinuationToken() : null);
+                    ModelBridgeInternal.setFeedOptionsContinuationToken(options, response != null ? response.getContinuationToken() : null);
```
milismsft, line 89: Here and in all the other tests and samples; can we avoid callingModelBridgeInternal.setFeedOptionsMaxItemCount() and instead do the byPage() conversion?

kushagraThapar: We cannot, because these ones use `AsyncDocumentClient` - which is an internal type. Public APIs `byPage()` are not available here. So that's why we used ModelBridgeInternal class. 
See this: https://github.com/Azure/azure-sdk-for-java/pull/10146/files/9c29922a7eea87190f77de4548325a4890364a9b#diff-6ca349d07cccd84853fafeaecacab723R30
But I agree, going forward, we should change the examples itself to use `CosmosClient` and public APIs.  But that is a much bigger change across SDK.

milismsft: understood

PR #10171, opened 2020-04-14
Turn on -amd switch to test downstream dependents.

mitchdenny:
This PR adds the ```-amd``` switch to the build and test jobs. This will trigger them to build and test downstream dependents. The goal here is that if you change something in the repo that someone else depends on, then your PR will validate that you didn't break them (for playback test scenarios).

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -114,7 +114,7 @@ jobs:
         inputs:
           mavenPomFile: pom.xml
           goals: 'package'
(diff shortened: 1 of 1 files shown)


JimSuplizio:
//CC @alzimmermsft @srnagar @weshaggard 
@mitchdenny I don't think that the -amd switch enabled for everything is going work. The main problem we're going to have will stem from Jacoco which uses the current versions of everything and that will cause every build/run except the From Source ones. With that being said, based upon the discussion that you, Alan and I had on Tuesday, the only times we're really going to see benefits from the -amd switch are in the From Source runs anyways. What if we only enable the -amd switch only for the From Source test runs? This would have a the desired effect, the version manipulation would allow everything to actually build and because of the -amd switch it should run all of the appropriate tests for that area and downstream dependencies. It would make the From Source runs the long pole in terms of pipeline runs but that's only 1 of the 7 test runs.

akennihan:
Test comment to help with triaging Azure Dev Ops Pipelines / GitHub Integration issue.

mitchdenny:
Closing this PR down since Jim has got a good approach for this.

PR #10172, opened 2020-04-14
fix checkstyle for management compute

xseeseesee:


sdk/compute/mgmt/src/main/java/com/azure/management/compute/ApiError.java
```diff
@@ -66,17 +62,17 @@ public ApiError withDetails(List<ApiErrorBase> details) {
...
     public InnerError innererror() {
         return this.innererror;
     }
 
     /**
-     * Set the innererror property: Inner error details.
-     * 
+     * Set the innererror property: The Api inner error.
+     *
      * @param innererror the innererror value to set.
```
joshfree, line 76: nit: "innererror the InnerError value to set."

PR #10182, opened 2020-04-14
Added isHidden API and make retrievable API private

sima-zhu:


sima-zhu:
#10184

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
```diff
@@ -224,6 +224,17 @@
...
 
+    /*
+     * A value indicating whether the field will be returned in a search
+     * result. This property must be true for key fields, and must be null for
+     * complex fields. You can hide a field from search results if you want to
+     * use it only as a filter, for sorting, or for scoring. This property can
+     * also be changed on existing fields and enabling it does not cause an
+     * increase in index storage requirements.
+     */
+    @JsonProperty(value = "hidden")
```
alzimmermsft, line 235: This may cause issues if the deserializer on the service side doesn't ignore unknown properties. Is there a way this change could be done which doesn't include adding a new property to the model?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
```diff
@@ -759,4 +770,35 @@ public Field setFields(List<Field> fields) {
         this.fields = fields;
         return this;
     }
+
+    /**
+     * Get the hidden property: A value indicating whether the field will be
+     * returned in a search result. This property must be true for key fields,
```
alzimmermsft, line 776: You'll want to fix the documentation to be negated, I have a feeling the `key` field shouldn't be hidden.

```suggestion
     * returned in a search result. This property must be false for key fields,
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
```diff
@@ -759,4 +770,35 @@ public Field setFields(List<Field> fields) {
...
+     * and must be null for complex fields. You can hide a field from search
```
alzimmermsft, line 777:
```suggestion
     * and must be true for complex fields. You can hide a field from search
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/SearchServiceVersion.java
```diff
@@ -9,7 +9,8 @@
  * The versions of Azure Search supported by this client library.
  */
 public enum SearchServiceVersion implements ServiceVersion {
-    V2019_05_06("2019-05-06");
+    V2019_05_06("2019-05-06"),
```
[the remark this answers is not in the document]
sima-zhu: They all have one service version. Will double check

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/SearchServiceVersion.java
```diff
@@ -9,7 +9,8 @@
...
+    V2019_05_06_Preview("2019-05-06-preview");
```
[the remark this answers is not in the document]
sima-zhu: Yes. Good catch

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
sima-zhu: Can use @JsonIgnore

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
sima-zhu: Most of the javadoc is the same as .net. Will double check

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
sima-zhu: This one keep the same as retrievable comments.

> This property must be true for key fields, and it must be null for complex fields.

alzimmermsft: Shouldn't keep the Javadocs the same between `hidden` and `retrievable` since they are inverses of each other. So, if the key field needs to be `retrievable = true` it must be `hidden = false/null` and the same applies for complex fields.

alzimmermsft: Scratch this comment, missed the logic where `hidden == null` means `retrievable == null`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
```diff
@@ -759,4 +771,35 @@ public Field setFields(List<Field> fields) {
...
+     *
+     * @return the hidden value.
+     */
+    public Boolean isHidden() {
+        return retrievable == null ? null : !retrievable;
+    }
+
+    /**
+     * Set the hidden property: A value indicating whether the field will be
+     * returned in a search result. This property must be true for key fields,
```
alzimmermsft, line 791: Need to update this spot as well 😄 

```suggestion
     * returned in a search result. This property must be false for key fields,
```

sdk/search/azure-search-documents/swagger/readme.md
```diff
@@ -428,4 +428,25 @@ directive:
...
         .replace(/public Boolean isShouldDetectOrientation\(\) \{/g, "public Boolean shouldDetectOrientation() {")
+
+    # Changed isRetrievable to isHidden
+    - from: swagger-document
+      where: $.definitions.Field.properties
+      transform: >
+         $.hidden = $.retrievable;
+         $.hidden = {
+            "type": "boolean", 
+            "description": "A value indicating whether the field will be returned in a search result. This property must be true for key fields, and must be null for complex fields. You can hide a field from search results if you want to use it only as a filter, for sorting, or for scoring. This property can also be changed on existing fields and enabling it does not cause an increase in index storage requirements."
```
alzimmermsft, line 439: Should update this so the Javadocs in the class generate with this must be false for key fields.

sima-zhu: I have merged your comments but regen the code. That's why the merge changes are gone. Done with transformation change

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
```diff
@@ -314,7 +326,7 @@ public Field setKey(Boolean key) {
      *
      * @return the retrievable value.
      */
-    public Boolean isRetrievable() {
+    private Boolean isRetrievable() {
```
conniey, line 329: Do we need these private getters and setters at all? We can just set the member variable.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
```diff
@@ -759,4 +771,35 @@ public Field setFields(List<Field> fields) {
...
+     * and must be null for complex fields. You can hide a field from search
+     * results if you want to use it only as a filter, for sorting, or for
+     * scoring. This property can also be changed on existing fields and
+     * enabling it does not cause an increase in index storage requirements.
+     *
+     * @param hidden the hidden value to set.
+     * @return the Field object itself.
+     */
+    public Field setHidden(Boolean hidden) {
+        retrievable = this.hidden == null ? null : !this.hidden;
```
conniey, line 801: This would be easier to understand if you do:
```java
if (this.hidden != null) {
    this.retrievable = !this.hidden;
}
```
Shouldn't the retrievable logic be done after setting the member variable? not before?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
conniey: Oh. I see. This is generated

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
sima-zhu: setHidden is the only entry to set `retrievable` field.
The logic needs to change to 
```
if (hidden != null) {
    this.retrievable = hidden;
}
this.hidden = hidden;
```
Good catch

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
sima-zhu: This is auto-gen code. I hope there is an easy way to remove getter and setter. Unfortunately, we have to write tons of codes in transformation. It is not a huge burden to leave it here. Try to balance the maintenance of transformation file and model structure here.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/Field.java
sima-zhu: One line change is easy to do in transformer. :)

I would like to switch to 
```
this.hidden = hidden;
retrievable = this.hidden == null ? null : !this.hidden;
```

PR #10220, opened 2020-04-15
Fixing aggregate reports

mitchdenny:
This PR aims to fix the aggregate reporting pipeline that broke after the POM restructuring. I think that after Alan fixed the pathing issues some of the other issues that I was having with this fell into place.

I'm not quite done yet, as I need to verify that the pipeline is picking up all the right things and there might be some configuration that I put back in now that I understand better how things are hanging together.

### DIFF
--- a/eng/pipelines/aggregate-reports.yml
+++ b/eng/pipelines/aggregate-reports.yml
@@ -59,7 +59,6 @@ jobs:
         # copy output/pom.client.html (Join-Path $(Build.ArtifactStagingDirectory) "staging")
 
         Copy-Item eng/common/InterdependencyGraph.html $(Build.ArtifactStagingDirectory)/staging
-        Copy-Item eng/reporting/dependency-whitelist.html $(Build.ArtifactStagingDirectory)/staging
         Copy-Item data.js $(Build.ArtifactStagingDirectory)/staging
       displayName: 'Copy reports to artifact staging'
 


sdk/parents/azure-client-sdk-parent/pom.xml
```diff
@@ -316,73 +316,31 @@
...
-              <goal>report</goal>
-            </goals>
-            <configuration>
-              <!-- Sets the output directory for the code coverage report. -->
-              <outputDirectory>${project.reporting.outputDirectory}/test-coverage</outputDirectory>
-              <excludes>
-                <exclude>**/com/azure/cosmos/implementation/apachecommons/**/*</exclude>
-                <exclude>**/com/azure/cosmos/implementation/guava25/**/*</exclude>
-                <exclude>**/com/azure/cosmos/implementation/guava27/**/*</exclude>
-              </excludes>
```
srnagar, line 351: We want to keep these exclusions.

sdk/parents/azure-client-sdk-parent/pom.xml
```diff
@@ -316,73 +316,31 @@
...
-                      <minimum>${jacoco.min.linecoverage}</minimum>
-                    </limit>
-                    <limit>
-                      <counter>BRANCH</counter>
-                      <value>COVEREDRATIO</value>
-                      <minimum>${jacoco.min.branchcoverage}</minimum>
-                    </limit>
-                  </limits>
-                </rule>
-              </rules>
```
srnagar, line 376: This is required to ensure we don't fall below coverage threshold. The PR should fail if the coverage is below this threshold.

eng/jacoco-test-coverage/pom.xml
```diff
@@ -188,14 +188,11 @@
...
           <execution>
-            <phase>prepare-package</phase>
+            <id>report-aggregate</id>
+            <phase>verify</phase>
             <goals>
               <goal>report-aggregate</goal>
             </goals>
-            <configuration>
-              <!-- Sets the output directory for the code coverage report. -->
-              <outputDirectory>${project.reporting.outputDirectory}/test-coverage</outputDirectory>
```
srnagar, line 197: This will require a corresponding update to [index.html](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/code-quality-reports/src/main/resources/index.html#L520)

eng/pipelines/aggregate-reports.yml
```diff
@@ -52,13 +50,16 @@ jobs:
...
-        copy output/pom.client.html (Join-Path $(Build.ArtifactStagingDirectory) "staging")
-        copy eng/common/InterdependencyGraph.html (Join-Path $(Build.ArtifactStagingDirectory) "staging")
-        copy eng/common/dependency-whitelist.html (Join-Path $(Build.ArtifactStagingDirectory) "staging")
-        copy data.js (Join-Path $(Build.ArtifactStagingDirectory) "staging")
+        Copy-Item eng/jacoco-test-coverage/target/staging $(Build.ArtifactStagingDirectory) -Recurse
+        Copy-Item eng/jacoco-test-coverage/target/dependency-whitelist.json $(Build.ArtifactStagingDirectory)\staging
+        Copy-Item eng/code-quality-reports/src/main/resources/index.html $(Build.ArtifactStagingDirectory)
+        
+        # What is this?
+        # copy output/pom.client.html (Join-Path $(Build.ArtifactStagingDirectory) "staging")
```
alzimmermsft, line 58: This is the dependency report that gets generated, but this won't generate using the new build structure.

mitchdenny: OK so how do we work around this one? Do we need to move the dependency checker logic into jacoco-test-coverage?

PR #10265, opened 2020-04-15
Tracer support in cosmosdb

simplynaveen20:
Added tracer support on all the public api

Sample view from the JaegerUi
image

**Following below points**

End to end public api capture

No nested spans will be captured only capturing parent call

Span name "<method name>.<container name>"
For query "method name>.<container name>.<query>”

Span attributes
	db.type (cosmosdb)
    db.instance(Database name)
    db.statement(exact same as span name)
    db.url(cosmos account uri)
      	 
In case of exception
     error.msg(throwable.getMessage())
     error.type(throwable.getClass().getName())
    error.stack(throwable.printStackTrace())

Span status(both in succes and failure) - Comos status code (HttpBased)

### DIFF
--- a/sdk/cosmos/azure-cosmos/pom.xml
+++ b/sdk/cosmos/azure-cosmos/pom.xml
@@ -131,7 +131,7 @@ Licensed under the MIT License.
       <groupId>com.azure</groupId>
       <artifactId>azure-core-tracing-opentelemetry</artifactId>
       <scope>test</scope>
-      <version>1.0.0-beta.4</version> <!-- {x-version-update;com.azure:azure-core-tracing-opentelemetry;dependency} -->
+      <version>1.0.0-beta.5</version> <!-- {x-version-update;com.azure:azure-core-tracing-opentelemetry;dependency} -->
       <exclusions>
         <exclusion>
           <groupId>io.netty</groupId>


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -223,10 +221,14 @@ CosmosKeyCredential cosmosKeyCredential() {
         }
         Database wrappedDatabase = new Database();
         wrappedDatabase.setId(databaseSettings.getId());
-        return asyncDocumentClient.createDatabase(wrappedDatabase, ModelBridgeInternal.toRequestOptions(options))
-                   .map(databaseResourceResponse -> ModelBridgeInternal.createCosmosAsyncDatabaseResponse(databaseResourceResponse,
-                       this))
-                   .single();
+        final CosmosDatabaseRequestOptions requestOptions = options;
+        return withContext(context -> createDatabase(wrappedDatabase, requestOptions, context)).subscriberContext(reactorContext -> {
```
simplynaveen20, line 225: As per conversation with Sameekhsa and Anu, core team will provide withContext, fluxContext api with map attributes to to handle nested calls on sdk more cleanly

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
```diff
@@ -84,4 +117,28 @@ public void subscribe(CoreSubscriber<? super T> coreSubscriber) {
...
+            if (pagedFluxOptions.getTracerProvider().isEnabled()) {
+                reactor.util.context.Context reactorContext = FluxUtil.toReactorContext(context);
+                Objects.requireNonNull(reactorContext.hasKey(TracerProvider.MASTER_CALL));
+                Optional<Object> callerFunc = reactorContext.getOrEmpty(TracerProvider.NESTED_CALL);
+                if (!callerFunc.isPresent()) {
+                    parentContext.set(pagedFluxOptions.getTracerProvider().startSpan(pagedFluxOptions.getTracerSpanName(),
+                        context.addData(TracerProvider.ATTRIBUTE_MAP, pagedFluxOptions.getTracingAttributes()), ProcessKind.DATABASE));
+                }
+            }
+        }).doOnEach(responseSignal -> {
```
simplynaveen20, line 134: Ending span on very first element of pagedFlux , avoiding accountability of any user side delay. Please let me know if anyone think over wise and want it on OnComplete()

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -411,4 +421,101 @@ public CosmosAsyncDatabase getDatabase(String id) {
...
+            }).doOnSuccess(response -> {
+                if (isTracingEnabled) {
+                    tracerProvider.endSpan(parentContext.get(), Signal.complete(), response.getStatusCode());
+                }
+            }).doOnError(throwable -> {
+                if (isTracingEnabled) {
+                    tracerProvider.endSpan(parentContext.get(), Signal.error(throwable), 0);
+                }
+            });
+    }
```
anuchandy, line 520: @simplynaveen20 - It seems the trace specific `doOnSubscribe | doOnSuccess | doOnError `  handlers need to be repeated in all APIs. Wondering it makes sense to centralize it

With that this method looks something like:

```java
private Mono<CosmosAsyncTriggerResponse> read(Context context) {
    Mono<CosmosAsyncDatabaseResponse> resultPublisher = asyncDocumentClient.createDatabase(database, ModelBridgeInternal.toRequestOptions(options))
        .map(databaseResourceResponse -> ModelBridgeInternal.createCosmosAsyncDatabaseResponse(databaseResourceResponse,
            this))
        .single()
    if (!tracerProvider.isEnabled()) {
        return resultPublisher;
    } else {
        final String spanName = "createDatabase." + database.getId();
        Map<String, String> tracingAttributes = new HashMap<String, String>() {{
            put(TracerProvider.DB_TYPE, TracerProvider.DB_TYPE_VALUE);
            put(TracerProvider.DB_INSTANCE, database.getId());
            put(TracerProvider.DB_URL, serviceEndpoint);
            put(TracerProvider.DB_STATEMENT, spanName);
        }};
    return traceEnabledPublisher(resultPublisher, tracingAttributes);
    }
}

// Reusable code 
//
private <T> Mono<T> traceEnabledPublisher(Mono<T> resultPublisher, Map<String, String> tracingAttributes) {
    return resultPublisher
        .doOnSubscribe(ignoredValue -> {
                reactor.util.context.Context reactorContext = FluxUtil.toReactorContext(context);
                Objects.requireNonNull(reactorContext.hasKey(TracerProvider.MASTER_CALL));
                Optional<Object> callerFunc = reactorContext.getOrEmpty(TracerProvider.NESTED_CALL);
                if (!callerFunc.isPresent()) {
                    parentContext.set(tracerProvider.startSpan(spanName,
                        context.addData(TracerProvider.ATTRIBUTE_MAP, tracingAttributes), ProcessKind.DATABASE));
                }
        }).doOnSuccess(response -> {
            tracerProvider.endSpan(parentContext.get(), Signal.complete(), response.getStatusCode());
        }).doOnError(throwable -> {
            tracerProvider.endSpan(parentContext.get(), Signal.error(throwable), 0);
        });
}

```

simplynaveen20: Yea actually I am thinking of cleaning this , and some other ones too  .Thank for suggestion . Will send it in next iteration

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
samvaity: This should already be available to use on latest core i.e ` <version>1.4.0-beta.1</version> `

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -411,4 +421,101 @@ public CosmosAsyncDatabase getDatabase(String id) {
     public void close() {
         asyncDocumentClient.close();
     }
+
+    TracerProvider getTracerProvider(){
```
samvaity, line 425: Do we need this?
We can directly use `this.tracerProvider` wherever needed. Given that the `CosmosAsynClient(Builder) `constructor is the only entry point to access these API's?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
samvaity: @trask Would like to know your thoughts on the expected behavior, if we should end the span of the first element received.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
simplynaveen20: Checked offline with Sameeksha 1.4.0-beta.1 does not have the requested overload

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
trask: @simplynaveen20 @samvaity That's a great question! I think that ideally it would capture until complete, while at the same time somehow capturing the internal time that was spent waiting on cosmosdb. Maybe the diagnostic data returned from the cosmosdb server could represent the internal time, as event(s) on the overarching span? This doesn't have to be part of the initial work though, and it's ok if you want to end the span on first element until we have a way to separate cosmosdb internal time from (external) user time while iterating over results.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @param context Additional metadata that is passed through the call stack.
+     * @return An updated context object.
+     */
+    public Context startSpan(String methodName, Context context, ProcessKind processKind) {
+        Context local = Objects.requireNonNull(context, "'context' cannot be null.");
+        Objects.requireNonNull(processKind, "'processKind' cannot be null.");
+
+        for (Tracer tracer : tracers) {
+            local = tracer.start(methodName, local, processKind);
```
samvaity, line 66: Given the requirements to set these specific attributes on every span it would help if we could use 
```java
local = tracer.start(methodName, local); // will start the span
tracer.setAttribute(DB_INSTANCE, "myDatabaseName", local); // use the current span to set attributes on it
```

This way we don't have to add a new ProcessKind and [this](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java#L45) `startSpan` overload can be used as it.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -386,6 +389,8 @@ CosmosKeyCredential cosmosKeyCredential() {
      */
     public CosmosPagedFlux<CosmosDatabaseProperties> queryDatabases(SqlQuerySpec querySpec, FeedOptions options) {
         return UtilBridgeInternal.createCosmosPagedFlux(pagedFluxOptions -> {
+            String spanName = "queryDatabases." + querySpec.getQueryText();
```
moderakh, line 392: this logs the query text? If the query has private info, this will get logged? is that ok?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -156,9 +192,34 @@ public String getId() {
...
-                   .replaceCollection(ModelBridgeInternal.getV2Collection(containerProperties), ModelBridgeInternal.toRequestOptions(options))
-                   .map(response -> ModelBridgeInternal.createCosmosAsyncContainerResponse(response, database)).single();
+
+        final CosmosContainerRequestOptions requestOptions = options;
+        return withContext(context -> replace(containerProperties, requestOptions, context)).subscriberContext(reactorContext -> {
+            Optional<String> master = reactorContext.getOrEmpty(TracerProvider.MASTER_CALL);
+            if (master.isPresent()) {
+                reactorContext = reactorContext.put(TracerProvider.NESTED_CALL, true);
+            }
+            return reactorContext.put(TracerProvider.MASTER_CALL, true);
```
moderakh, line 202: seems this code is repeated exactly, can this be factored out as a function and passed here?
\liked this
```java
return withContext(context -> replace(containerProperties, requestOptions, context))
    .subscriberContext(reactorContext -> factoredOutFunc())
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -216,16 +276,37 @@ public String getId() {
...
             options = new CosmosItemRequestOptions();
         }
+
+        final CosmosItemRequestOptions requestOptions = options;
+        return withContext(context -> createItem(item, requestOptions, context)).subscriberContext(reactorContext -> {
+            Optional<String> master = reactorContext.getOrEmpty(TracerProvider.MASTER_CALL);
+            if (master.isPresent()) {
+                reactorContext = reactorContext.put(TracerProvider.NESTED_CALL, true);
+            }
+            return reactorContext.put(TracerProvider.MASTER_CALL, true);
```
moderakh, line 286: exactly the same code, can this be factored out

```java
return withContext(context -> replace(containerProperties, requestOptions, context))
    .subscriberContext(reactorContext -> factoredOutFunc())
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -379,6 +464,9 @@ public String getId() {
     private <T> CosmosPagedFlux<T> queryItemsInternal(
         SqlQuerySpec sqlQuerySpec, FeedOptions feedOptions, Class<T> classType) {
         return UtilBridgeInternal.createCosmosPagedFlux(pagedFluxOptions -> {
+            String spanName = "queryItems." + this.getId() + "." + sqlQuerySpec.getQueryText();
```
moderakh, line 467: does this log the user query? is that ok? user query may have private info?
```java
"queryItems." + this.getId() + "." + sqlQuerySpec.getQueryText();
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -481,11 +572,14 @@ public String getId() {
...
-                   .replaceDocument(getItemLink(itemId), doc, ModelBridgeInternal.toRequestOptions(options))
-                   .map(response -> ModelBridgeInternal.createCosmosAsyncItemResponse(response, itemType))
-                   .single();
+        final CosmosItemRequestOptions requestOptions = options;
+        return withContext(context -> replaceItem(itemType, itemId, doc, requestOptions,context)).subscriberContext(reactorContext -> {
+            Optional<String> master = reactorContext.getOrEmpty(TracerProvider.MASTER_CALL);
+            if (master.isPresent()) {
+                reactorContext = reactorContext.put(TracerProvider.NESTED_CALL, true);
+            }
+            return reactorContext.put(TracerProvider.MASTER_CALL, true);
```
moderakh, line 581: can we factor out this code, here and other places?

e.g,
```java
return withContext(context -> replaceItem(itemType, itemId, doc, requestOptions,context))
    .subscriberContext(reactorContext -> factoredOutFunc())
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -681,4 +763,168 @@ String getParentLink() {
...
+        Document doc,
+        CosmosItemRequestOptions options,
+        Context context) {
+        String spanName = "replaceItem." + this.getId();
+        Map<String, String> tracingAttributes = new HashMap<String, String>() {{
+            put(TracerProvider.DB_TYPE, TracerProvider.DB_TYPE_VALUE);
+            put(TracerProvider.DB_INSTANCE, getId());
+            put(TracerProvider.DB_URL, database.getClient().getServiceEndpoint());
+            put(TracerProvider.DB_STATEMENT, spanName);
+        }};
```
moderakh, line 801: exactly same code as above, can we factor out?

```java
Map<String, String> createTracingMap(...) {
   ...
}

and then invoke createTracingMap() here
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -681,4 +763,168 @@ String getParentLink() {
...
+    }
+
+    private <T> Mono<CosmosAsyncItemResponse<T>> upsertItem(T item, CosmosItemRequestOptions options, Context context) {
+        String spanName = "upsertItem." + this.getId();
+        Map<String, String> tracingAttributes = new HashMap<String, String>() {{
+            put(TracerProvider.DB_TYPE, TracerProvider.DB_TYPE_VALUE);
+            put(TracerProvider.DB_INSTANCE, getId());
+            put(TracerProvider.DB_URL, database.getClient().getServiceEndpoint());
+            put(TracerProvider.DB_STATEMENT, spanName);
+        }};
```
moderakh, line 820: ditto, here and other places.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
simplynaveen20: Discussed with Sameeksha offline , as of now cosmos sdk wont need it

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -281,10 +278,14 @@ CosmosKeyCredential cosmosKeyCredential() {
...
-                       this))
-                   .single();
+        final CosmosDatabaseRequestOptions requestOptions = options;
+        return withContext(context -> createDatabase(wrappedDatabase, requestOptions, context)).subscriberContext(reactorContext -> {
+            Optional<String> master = reactorContext.getOrEmpty(TracerProvider.MASTER_CALL);
+            if (master.isPresent()) {
+                reactorContext = reactorContext.put(TracerProvider.NESTED_CALL, true);
+            }
+            return reactorContext.put(TracerProvider.MASTER_CALL, true);
+        });
```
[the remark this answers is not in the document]
simplynaveen20: Moved the PR out draft version, did bunch of refactoring , and I like your third option i.e using
CallDepth (more intuitive for code readability)  , thanks for advice :)

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,209 @@
...
+        final AtomicReference<Context> parentContext = isTracingEnabled
+            ? new AtomicReference<>(Context.NONE)
+            : null;
+        return resultPublisher
+            .doOnSubscribe(ignoredValue -> {
+                if (isTracingEnabled) {
+                    reactor.util.context.Context reactorContext = FluxUtil.toReactorContext(context);
+                    Objects.requireNonNull(reactorContext.hasKey(TracerProvider.MASTER_CALL));
+                    Optional<Object> callerFunc = reactorContext.getOrEmpty(TracerProvider.NESTED_CALL);
+                    if (!callerFunc.isPresent()) {
```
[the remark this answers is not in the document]
simplynaveen20: Moved to CallDepth enum, all set on this

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,209 @@
...
+    public final static String DB_TYPE_VALUE = "cosmosdb";
+    public final static String DB_TYPE = "db.type";
+    public final static String DB_INSTANCE = "db.instance";
+    public final static String DB_URL = "db.url";
+    public static final String DB_STATEMENT = "db.statement";
+    public static final String ERROR_MSG = "error.msg";
+    public static final String ERROR_TYPE = "error.type";
+    public static final String ERROR_STACK = "error.stack";
+    public static final String MASTER_CALL = "masterCall";
+    public static final String NESTED_CALL = "nestedCall";
```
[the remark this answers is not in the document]
simplynaveen20: Not needed anymore

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
simplynaveen20: Removed ProcessKind dependency.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: I was waiting on final handshake between cosmos and core for withContext api, did the refactoring now.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,209 @@
...
+                                                                                        String spanName) {
+        final boolean isTracingEnabled = this.isEnabled();
+        final AtomicReference<Context> parentContext = isTracingEnabled
+            ? new AtomicReference<>(Context.NONE)
+            : null;
+        return resultPublisher
+            .doOnSubscribe(ignoredValue -> {
+                if (isTracingEnabled) {
+                    reactor.util.context.Context reactorContext = FluxUtil.toReactorContext(context);
+                    Objects.requireNonNull(reactorContext.hasKey(TracerProvider.MASTER_CALL));
```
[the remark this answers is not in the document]
simplynaveen20: Removed this method usage

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,209 @@
...
+            }).doOnSuccess(response -> {
+                if (isTracingEnabled) {
+                    this.endSpan(parentContext.get(), Signal.complete(), response.getStatusCode());
+                }
+            }).doOnError(throwable -> {
+                if (isTracingEnabled) {
+                    this.endSpan(parentContext.get(), Signal.error(throwable), 0);
+                }
+            });
+    }
```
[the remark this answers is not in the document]
simplynaveen20: Done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
trask: For jdbc, we only capture PreparedStatement query templates, and not the actual bind parameters provided by the user. We do something similar for MongoDB and other databases. is `querytext` a parameterized query template? or are bind parameters already embedded into it?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
simplynaveen20: Moved to to doOnComplete()(It will include both sdk + user side time). However we have diagnostics to find out exact time taken by SDK

trask: oh, nice, will that be part of the diagnostics that we capture in the next phase?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,206 @@
...
+    public final static String DB_TYPE_VALUE = "cosmosdb";
+    public final static String DB_TYPE = "db.type";
+    public final static String DB_INSTANCE = "db.instance";
+    public final static String DB_URL = "db.url";
+    public static final String DB_STATEMENT = "db.statement";
+    public static final String ERROR_MSG = "error.msg";
+    public static final String ERROR_TYPE = "error.type";
+    public static final String ERROR_STACK = "error.stack";
+    public static final String COSMOS_CALL_DEPTH = "cosmosCallDepth";
+    public static final String MICROSOFT_DOCOMENTDB = "Microsoft.DocumentDB";
```
moderakh, line 39: Does documentDB refer to us? if so, that name is deprecated.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+
+import static com.azure.core.util.tracing.Tracer.AZ_TRACING_NAMESPACE_KEY;
+
+public class TracerProvider {
+    private final ClientLogger logger = new ClientLogger(TracerProvider.class);
+    private final List<Tracer> tracers = new ArrayList<>();
+    public final static String DB_TYPE_VALUE = "cosmosdb";
```
moderakh, line 30: shouldn't this be "cosmos" instead of "cosmosdb"

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: Cosmos have two api for query , one take plain text with parameter embedded into it and second takes SqlQuerySpec , which is template based. For plain text query, code itself create SqlQuerySpec and use common code flow as above.
We have two option here - 
1. Only log query if user uses SqlQuerySpec  api , but this will be inconsistent.
2. Log for both as it is happening now , assuming for all the sensitive query user will use SqlQuerySpec  api.
3. Don't log query text any of the apis.

I would go with option 1 , as I can take inconsistency over possibility of leaking private info. Option 3 will be my second choice
Thoughts ?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
samvaity: The resource provider name is being provided from this [list of Azure services](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-services-resource-providers).
Feature request to include this on all Track 2 SDK's - https://github.com/Azure/azure-sdk-for-java/issues/7049

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
trask: yes, #1 sounds good. we still want to capture a span for plain text queries, but use something more generic for the span name and `db.statement` attribute, e.g. the same value that you put into `db.instance` would be acceptable

simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
simplynaveen20: @trask  ?

trask: both are good from my perspective 👍

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -199,7 +201,12 @@ boolean isContentResponseOnWriteEnabled() {
      * an error.
      */
     public Mono<CosmosAsyncDatabaseResponse> createDatabaseIfNotExists(CosmosDatabaseProperties databaseSettings) {
-        return createDatabaseIfNotExistsInternal(getDatabase(databaseSettings.getId()));
+        if(!getTracerProvider().isEnabled()) {
```
samvaity, line 204: The `withContext` overload would be used for context passing and not just tracing context-related so should not be opted in for regardless of tracing presence.
@srnagar

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -199,7 +201,12 @@ boolean isContentResponseOnWriteEnabled() {
...
+            return createDatabaseIfNotExistsInternal(getDatabase(databaseSettings.getId()));
+        }
+
+        return withContext(context -> createDatabaseIfNotExistsInternal(getDatabase(databaseSettings.getId()),
```
samvaity, line 208: The `withContext` overload would be used for context passing and not just tracing context-related so should not be opted in for regardless of tracing presence.
@srnagar

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -444,4 +447,77 @@ public CosmosAsyncDatabase getDatabase(String id) {
...
+                .map(response -> BridgeInternal.createFeedResponse(
+                    ModelBridgeInternal.getCosmosDatabasePropertiesFromV2Results(response.getResults()),
+                    response.getResponseHeaders()));
+        });
+    }
+
+
+    private Mono<CosmosAsyncDatabaseResponse> createDatabaseIfNotExistsInternal(CosmosAsyncDatabase database,
+                                                                                Context context) {
+        String spanName = "createDatabaseIfNotExistsInternal." + database.getId();
```
samvaity, line 476: Should this span name be `createDatabaseIfNotExists` and not internal?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
simplynaveen20: We don't want any perf hit due to tracing code flow if user has not opt in the tracing. Unless we have any requirement which need with context , and that will be out of scope of this PR

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,197 @@
...
+    public static final String ERROR_TYPE = "error.type";
+    public static final String ERROR_STACK = "error.stack";
+    public static final String COSMOS_CALL_DEPTH = "cosmosCallDepth";
+    public static final String MICROSOFT_DOCOMENTDB = "Microsoft.DocumentDB";
+
+
+    public final static Function<reactor.util.context.Context, reactor.util.context.Context> callDepthAttributeFunc = (
+        reactor.util.context.Context reactorContext) -> {
+        CallDepth callDepth = reactorContext.getOrDefault(COSMOS_CALL_DEPTH, CallDepth.ZERO);
+        assert callDepth != null;
```
srnagar, line 43: `callDepth` cannot be null as you are doing `getOrDefault` in the previous line. 

Also, it's better to use `if (callDepth == null) { throw exception; }` instead of using `assert`. If the assertion fails, it throws `AssertionError` extends from `Error` which is considered fatal and the program cannot recover.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -444,4 +447,77 @@ public CosmosAsyncDatabase getDatabase(String id) {
...
+            final Throwable unwrappedException = Exceptions.unwrap(exception);
+            if (unwrappedException instanceof CosmosClientException) {
+                final CosmosClientException cosmosClientException = (CosmosClientException) unwrappedException;
+                if (cosmosClientException.getStatusCode() == HttpConstants.StatusCodes.NOTFOUND) {
+                    return createDatabase(new CosmosDatabaseProperties(database.getId()),
+                        new CosmosDatabaseRequestOptions());
+                }
+            }
+            return Mono.error(unwrappedException);
+        });
```
srnagar, line 487: This code segment is common for both overloads of `createDatabaseIfNotExistsInternal`. Consider extracting into a common method instead of repeating it.

sdk/cosmos/azure-cosmos/pom.xml
```diff
@@ -132,6 +132,19 @@ Licensed under the MIT License.
...
 
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-tracing-opentelemetry</artifactId>
+      <scope>test</scope>
+      <version>1.0.0-beta.4</version>
+      <exclusions>
+        <exclusion>
+          <groupId>io.netty</groupId>
+          <artifactId>*</artifactId>
```
moderakh, line 143: excluding only boring ssl should be sufficient.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
moderakh: The brand of our service, our package name, etc, is cosmos. So this should match that.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,197 @@
...
+    public final static String DB_URL = "db.url";
+    public static final String DB_STATEMENT = "db.statement";
+    public static final String ERROR_MSG = "error.msg";
+    public static final String ERROR_TYPE = "error.type";
+    public static final String ERROR_STACK = "error.stack";
+    public static final String COSMOS_CALL_DEPTH = "cosmosCallDepth";
+    public static final String MICROSOFT_DOCOMENTDB = "Microsoft.DocumentDB";
+
+
+    public final static Function<reactor.util.context.Context, reactor.util.context.Context> callDepthAttributeFunc = (
```
moderakh, line 40: code style: callDepthAttributeFunc is a static field. CALL_DEPTH_ATTRIBUTE_FUNC ?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,197 @@
...
+    public final static String DB_TYPE_VALUE = "cosmosdb";
+    public final static String DB_TYPE = "db.type";
+    public final static String DB_INSTANCE = "db.instance";
+    public final static String DB_URL = "db.url";
+    public static final String DB_STATEMENT = "db.statement";
+    public static final String ERROR_MSG = "error.msg";
+    public static final String ERROR_TYPE = "error.type";
+    public static final String ERROR_STACK = "error.stack";
+    public static final String COSMOS_CALL_DEPTH = "cosmosCallDepth";
+    public static final String MICROSOFT_DOCOMENTDB = "Microsoft.DocumentDB";
```
moderakh, line 37: MICROSOFT_DOCOMENTDB -> RESOURCE_PROVIDER_NAME

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
```diff
@@ -37,24 +44,23 @@
     @Override
     public Flux<FeedResponse<T>> byPage() {
         CosmosPagedFluxOptions cosmosPagedFluxOptions = new CosmosPagedFluxOptions();
-
-        return this.optionsFluxFunction.apply(cosmosPagedFluxOptions);
+        return FluxUtil.fluxContext(context ->  byPage(cosmosPagedFluxOptions, context)).subscriberContext(TracerProvider.callDepthAttributeFunc);
```
moderakh, line 47: regardless of if tracing is enabled or not we are passing the Pageflux to the tracing stack, and recreating some internal Map in `FluxUtil::toAzureContext`

This should sit behind a tracing enabled flag.

```suggestion
if (tracing.enabled) {
 return FluxUtil.fluxContext(context ->  byPage(cosmosPagedFluxOptions, context)).subscriberContext(TracerProvider.callDepthAttributeFunc);
} else {
 return this.optionsFluxFunction.apply(cosmosPagedFluxOptions);
}
```

You should be able to factor out the common if/else code into a method and just call that here, but it seems to me we should have a short circurt for when tracing is not enabled.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
```diff
@@ -84,4 +90,26 @@ public void subscribe(CoreSubscriber<? super T> coreSubscriber) {
...
+        assert callDepth != null;
+        final boolean isNestedCall = callDepth.equals(TracerProvider.CallDepth.TWO_OR_MORE);
+        return this.optionsFluxFunction.apply(pagedFluxOptions).doOnSubscribe(ignoredValue -> {
+            if ( pagedFluxOptions.getTracerProvider().isEnabled() && !isNestedCall) {
+                    parentContext.set(pagedFluxOptions.getTracerProvider().startSpan(pagedFluxOptions.getTracerSpanName(), pagedFluxOptions.getDatabaseId(), pagedFluxOptions.getServiceEndpoint(),
+                        context));
+            }
+        }).doOnComplete(() -> {
+            if ( pagedFluxOptions.getTracerProvider().isEnabled() && !isNestedCall) {
+                pagedFluxOptions.getTracerProvider().endSpan(parentContext.get(), Signal.complete(), 200);
```
moderakh, line 107: magic number 200 should be a constant.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosTracerTest.java
```diff
@@ -0,0 +1,251 @@
+package com.azure.cosmos;
```
moderakh, line 1: license header.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -85,8 +88,13 @@ public String getId() {
...
         }
-        return database.getDocClientWrapper().readCollection(getLink(), ModelBridgeInternal.toRequestOptions(options))
-                   .map(response -> ModelBridgeInternal.createCosmosAsyncContainerResponse(response, database)).single();
+
+        final CosmosContainerRequestOptions requestOptions = options;
+        if(!database.getClient().getTracerProvider().isEnabled()){
+            return readInternal(options);
+        }
+
+        return withContext(context -> read(requestOptions, context)).subscriberContext(TracerProvider.callDepthAttributeFunc);
```
moderakh, line 97: this will have additional cost even if tracer not enabled.
i see in the code withContext is doing Map creation etc.

we should short circuit this if tracer is not enabled.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -104,8 +112,13 @@ public String getId() {
...
         }
-        return database.getDocClientWrapper().deleteCollection(getLink(), ModelBridgeInternal.toRequestOptions(options))
-                   .map(response -> ModelBridgeInternal.createCosmosAsyncContainerResponse(response, database)).single();
+
+        if (!database.getClient().getTracerProvider().isEnabled()) {
+            return deleteInternal(options);
+        }
+
+        final CosmosContainerRequestOptions requestOptions = options;
+        return withContext(context -> deleteInternal(requestOptions, context)).subscriberContext(TracerProvider.callDepthAttributeFunc);
```
moderakh, line 121: ditto, here and in other places.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -218,16 +234,33 @@ public String getId() {
...
+        if (!database.getClient().getTracerProvider().isEnabled()) {
+            return createItemInternal(item, options);
+        }
+
+        final CosmosItemRequestOptions requestOptions = options;
+        return withContext(context -> createItemInternal(item, requestOptions, context)).subscriberContext(TracerProvider.callDepthAttributeFunc);
+    }
+
+    private <T> Mono<CosmosAsyncItemResponse<T>> createItemInternal(T item, CosmosItemRequestOptions options, Context context) {
+        String spanName = "createItem." + this.getId();
```
moderakh, line 247: id is the id of the container. so it means in the scope of this container `"createItem." + this.getId()` can be pre-computed for the container, rather than re-computing it each time.

Could you pre-compute this in the container level.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -302,6 +334,9 @@ public String getId() {
      */
     public <T> CosmosPagedFlux<T> readAllItems(FeedOptions options, Class<T> classType) {
         return UtilBridgeInternal.createCosmosPagedFlux(pagedFluxOptions -> {
+            String spanName = "readAllItems." + this.getId();
```
moderakh, line 337: id is the id of the container. so it means in the scope of this container "readAllItems." + this.getId() can be pre-computed for the container, rather than re-computing it each time.

Could you pre-compute this in the container level.

here and other places.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -747,4 +776,182 @@ String getParentLink() {
...
+                true)
+            .map(response -> ModelBridgeInternal.createCosmosAsyncItemResponse(response, itemType))
+            .single();
+    }
+
+    private <T> Mono<CosmosAsyncItemResponse<T>> readItemInternal(
+        String itemId, PartitionKey partitionKey,
+        RequestOptions requestOptions, Class<T> itemType,
+        Context context) {
+        String spanName = "readItem." + this.getId();
```
moderakh, line 846: `"readItem." + this.getId()` can be pre-computed in container rather than re-computing each time.

please pre-compute in the container.

and for similar other operations.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
moderakh: @samvaity it seems as per the link you shared the service provider name of our service is not updated from Microsoft.DocumentDB to Microsoft.Cosmos.

My question to you is if for tracer work, this is merge implementation details or whether in the application insight the customers will see the term "DocumentDB" ?

moderakh: FYI @kirankumarkolli

samvaity: The customers would see the term "DocumentDB".
I think the ask was from the Azure Monitor team to be able to group spans/trace information for a particular Azure service and wanted to follow consistent naming across different components (i. Azure Monitor). 

>Azure Monitor exporter will use this information to navigate to the corresponding Azure resource from AzMon UX and show proper icons. #7049

I agree we should not follow deprecated naming but then is this listed incorrectly on the [list](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-services-resource-providers) too?
@trask Do you think if Cosmos should still be referred to as `Microsoft.DocumentDB`?

kirankumarkolli: Our SDK names already deviate from the RP name. 
Just for context: "Azure Cosmos DB" is the new branding (and RP will rebrand eventually). 

For all Track2 libraries: How about using the package name 'Azure.Cosmos" in our specific case?

/cc: @JonathanGiles

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
kirankumarkolli: For consistency on SDK lets use "Cosmos"

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: its already after the tracer enable check

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: its already after the tracer enable check

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
simplynaveen20: Cosmos page flux does not hold any state , so we don't have tracing.enabled flag till optionsFluxFunction.apply is called, please suggest

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
simplynaveen20: as discussed offline , it is due to spot bug

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
simplynaveen20: as discussed offline will introduce this on demand

sdk/cosmos/azure-cosmos/pom.xml
```diff
@@ -132,6 +132,18 @@ Licensed under the MIT License.
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-tracing-opentelemetry</artifactId> <!-- {x-version-update;com.azure:azure-core-tracing-opentelemetry;dependency} -->
```
srnagar, line 137: The tag should be associated with the `<version>`.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
```diff
@@ -84,4 +103,18 @@ public void subscribe(CoreSubscriber<? super T> coreSubscriber) {
             return Flux.fromIterable(elements);
         }).subscribe(coreSubscriber);
     }
+
+    private Flux<FeedResponse<T>> byPage(CosmosPagedFluxOptions pagedFluxOptions, Context context) {
```
moderakh, line 107: `CosmosPagedFlux` allows both paginated iteration and also item by item iteration.

I see that we covered the pageByPage iteration through adding `CosmosPagedFlux#byPage(.)` what about the item by item iteration shoudldn't we cover that as well?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -768,4 +782,244 @@ String getParentLink() {
...
+            .single();
+    }
+
+    private <T> Mono<CosmosAsyncItemResponse<T>> readItemInternal(
+        String itemId, PartitionKey partitionKey,
+        RequestOptions requestOptions, Class<T> itemType,
+        Context context) {
+        Mono<CosmosAsyncItemResponse<T>> responseMono = readItemInternal(itemId, partitionKey, requestOptions, itemType);
+        return database.getClient().getTracerProvider().traceEnabledCosmosItemResponsePublisher(responseMono,
+            context, this.readContainerSpanName, database.getId(), database.getClient().getServiceEndpoint());
```
moderakh, line 851: readContainerSpanName -> readItemSpanName. Please check other ones too.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -768,4 +782,244 @@ String getParentLink() {
...
+    }
+
+    private <T> Mono<CosmosAsyncItemResponse<T>> readItemInternal(
+        String itemId, PartitionKey partitionKey,
+        RequestOptions requestOptions, Class<T> itemType) {
+        return this.getDatabase().getDocClientWrapper()
+            .readDocument(getItemLink(itemId), requestOptions)
+            .map(response -> ModelBridgeInternal.createCosmosAsyncItemResponse(response, itemType))
```
moderakh, line 859: partitionKey is not passed here.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
moderakh: discussed offline.

sdk/cosmos/azure-cosmos/src/main/java/module-info.java
```diff
@@ -24,13 +24,18 @@
     requires micrometer.core;
     //  This is only required by guava shaded libraries
     requires java.logging;
+    requires reactor.core;
+    requires org.reactivestreams;
+    requires com.fasterxml.jackson.databind;
+    requires org.slf4j;
```
srnagar, line 30: These are not required as `com.azure.core` module already requires them transitively.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -164,9 +207,13 @@ public String getId() {
         if (options == null) {
             options = new CosmosContainerRequestOptions();
         }
-        return database.getDocClientWrapper()
-                   .replaceCollection(ModelBridgeInternal.getV2Collection(containerProperties), ModelBridgeInternal.toRequestOptions(options))
-                   .map(response -> ModelBridgeInternal.createCosmosAsyncContainerResponse(response, database)).single();
+
+        if(!database.getClient().getTracerProvider().isEnabled()) {
```
samvaity, line 211: Seeing this is used frequently throughout the class, can this be extracted to somewhere common?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncUserDefinedFunction.java
```diff
@@ -115,4 +121,45 @@ String getLink() {
         builder.append(getId());
         return builder.toString();
     }
+
+    private Mono<CosmosAsyncUserDefinedFunctionResponse> readInternal(Context context) {
+        String spanName = "readUDF." + container.getId();
```
samvaity, line 126: The span name should match the method name which is `read` here

samvaity: Same for methods `replace` and `delete` below.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,175 @@
...
+     * The {@code context} will be checked for containing information about a parent span. If a parent span is found the
+     * new span will be added as a child, otherwise the span will be created and added to the context and any downstream
+     * start calls will use the created span as the parent.
+     *
+     * @param context Additional metadata that is passed through the call stack.
+     * @return An updated context object.
+     */
+    public Context startSpan(String methodName, String databaseId, String endpoint, Context context) {
+        Context local = Objects.requireNonNull(context, "'context' cannot be null.");
+        for (Tracer tracer : tracers) {
```
samvaity, line 61: We don't need to support multiple tracer implementation. 
Refer here - https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/TracerProvider.java#L40

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,175 @@
...
+            local = tracer.start(methodName, local); // start the span and return the started span
+            if (databaseId != null) {
+                tracer.setAttribute(TracerProvider.DB_INSTANCE, databaseId, local);
+            }
+
+            tracer.setAttribute(AZ_TRACING_NAMESPACE_KEY, RESOURCE_PROVIDER_NAME, local);
```
samvaity, line 67: AZ_TRACING_NAMESPACE_KEY needs to be passed using the context as the tracer expects it in the context [here](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core-tracing-opentelemetry/src/main/java/com/azure/core/tracing/opentelemetry/OpenTelemetryTracer.java#L52) and aligns with other SDK behavior.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,175 @@
...
+                break;
+            case ON_ERROR:
+                Throwable throwable = null;
+                if (signal.hasError()) {
+                    // The last status available is on error, this contains the thrown error.
+                    throwable = signal.getThrowable();
+
+                    if (throwable instanceof CosmosException) {
+                        CosmosException exception = (CosmosException) throwable;
+                        // confirm ?
```
samvaity, line 99: remove this comment?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,175 @@
...
+        return traceEnabledPublisher(resultPublisher,  context, spanName,databaseId, endpoint,
+            (T response) -> response.getStatusCode());
+    }
+
+    public <T> Mono<T> traceEnabledNonCosmosResponsePublisher(Mono<T> resultPublisher,
+                                                              Context context,
+                                                              String spanName,
+                                                              String databaseId,
+                                                              String endpoint) {
+        return traceEnabledPublisher(resultPublisher,  context, spanName, databaseId, endpoint, (T response) -> HttpConstants.StatusCodes.OK);
```
samvaity, line 125: Why are we always sending a `200 OK` status here?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,175 @@
...
+            });
+    }
+
+    private void end(int statusCode, Throwable throwable, Context context) {
+        for (Tracer tracer : tracers) {
+            if (throwable != null) {
+                tracer.setAttribute(TracerProvider.ERROR_MSG, throwable.getMessage(), context);
+                tracer.setAttribute(TracerProvider.ERROR_TYPE, throwable.getClass().getName(), context);
+                StringWriter errorStack = new StringWriter();
+                throwable.printStackTrace(new PrintWriter(errorStack));
```
samvaity, line 169: do we think this is a good idea?
especially since `Throwable.printStackTrace()` isn't thread-safe and could give out confusing logs/statements?
cc: @srnagar

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
```diff
@@ -29,41 +34,55 @@
 public final class CosmosPagedFlux<T> extends ContinuablePagedFlux<String, T, FeedResponse<T>> {
 
     private final Function<CosmosPagedFluxOptions, Flux<FeedResponse<T>>> optionsFluxFunction;
+    private final boolean isTracerEnable;
```
samvaity, line 37: isTracerEnable --> isTracerEnabled

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
```diff
@@ -84,4 +103,18 @@ public void subscribe(CoreSubscriber<? super T> coreSubscriber) {
...
+        final AtomicReference<Context> parentContext = new AtomicReference<>(Context.NONE);
+        return this.optionsFluxFunction.apply(pagedFluxOptions).doOnSubscribe(ignoredValue -> {
+            parentContext.set(pagedFluxOptions.getTracerProvider().startSpan(pagedFluxOptions.getTracerSpanName(),
+                pagedFluxOptions.getDatabaseId(), pagedFluxOptions.getServiceEndpoint(),
+                context));
+        }).doOnComplete(() -> {
+            pagedFluxOptions.getTracerProvider().endSpan(parentContext.get(), Signal.complete(),
+                HttpConstants.StatusCodes.OK);
+        }).doOnError(throwable -> {
+            pagedFluxOptions.getTracerProvider().endSpan(parentContext.get(), Signal.error(throwable), 0);
```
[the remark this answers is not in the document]
samvaity: +1 we should pass the failing status code.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/UtilBridgeInternal.java
```diff
@@ -24,10 +24,11 @@
     private UtilBridgeInternal() {}
 
     @Warning(value = INTERNAL_USE_ONLY_WARNING)
-    public static <T> CosmosPagedFlux<T> createCosmosPagedFlux(Function<CosmosPagedFluxOptions, Flux<FeedResponse<T>>> pagedFluxOptionsFluxFunction) {
-        return new CosmosPagedFlux<>(pagedFluxOptionsFluxFunction);
+    public static <T> CosmosPagedFlux<T> createCosmosPagedFlux(Function<CosmosPagedFluxOptions, Flux<FeedResponse<T>>> pagedFluxOptionsFluxFunction, boolean isTracerEnable) {
```
samvaity, line 27: isTraceeEnable --> isTracerEnabled

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosTracerTest.java
```diff
@@ -0,0 +1,251 @@
...
+        cosmosAsyncContainer = getSharedMultiPartitionCosmosContainer(client);
+
+    }
+
+    @Test(groups = {"emulator"}, timeOut = TIMEOUT)
+    public void cosmosAsyncClient() {
+        TracerProvider tracer = Mockito.spy(new TracerProvider(ServiceLoader.load(Tracer.class)));
+        ReflectionUtils.setTracerProvider(client, tracer);
+
+        client.createDatabaseIfNotExists(cosmosAsyncDatabase.getId()).block();
```
samvaity, line 53: use `StepVerifier` instead of `block` for testing async calls?

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosTracerTest.java
```diff
@@ -0,0 +1,251 @@
...
+        Mockito.verify(tracer, Mockito.times(1)).startSpan(Matchers.anyString(), Matchers.anyString(),
```
samvaity, line 54: Rather than any string could we validate the span names and attributes being set on the span?

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/TestSuiteBase.java
```diff
@@ -469,8 +469,8 @@ public static CosmosItemProperties createDocument(CosmosAsyncContainer cosmosCon
     }
```
samvaity, line 469: Should consider adding tests for `TracerProvider`.

samvaity: Also, consider adding a nesting call example to validate the behavior of spans.

sdk/cosmos/azure-cosmos/pom.xml
```diff
@@ -127,7 +127,18 @@ Licensed under the MIT License.
         </exclusion>
       </exclusions>
     </dependency>
-
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-tracing-opentelemetry</artifactId>
```
samvaity, line 132: Is this test dependency being used even when you are using mockito?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
JonathanGiles: I would like this to be resolved within this PR - Context is general purpose and not specifically tied to tracing. That it is treated as such here means that we are not following convention and users who are using Context for other purposes, e.g. in their own http policies, will get a surprise when the context is not properly propagating for them.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
srnagar: I don't think we should have the entire stacktrace in tracing. It's important to capture that there was an error but the stracktrace is likely not that useful. Stacktrace should be available in logs, if that's necessary. It's also a security issue to capture stacktrace since the user cannot control or turn it off.

sdk/cosmos/azure-cosmos/pom.xml
simplynaveen20: Yes

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/TestSuiteBase.java
simplynaveen20: This is cover by start span count assert , test will fail if there is nesting . Will add explicit test in next pr covering event

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosTracerTest.java
simplynaveen20: Will address all testing format comment in next  tracer event pr as discussed offline

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/CosmosTracerTest.java
simplynaveen20: Will address testing format comment in next  tracer event pr as discussed offline

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
simplynaveen20: removed setting AZ_TRACING_NAMESPACE_KEY explicitly on tracer

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncUserDefinedFunction.java
simplynaveen20: Discussed offline , we need resource name on read/delete/replace as we have hierarchy model and all the resource have same api name for these operation, so we need to differentiate them

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: Eventually this check will go away when we will go via withContext all the time.

sdk/cosmos/azure-cosmos/src/main/java/module-info.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -46,12 +49,42 @@
...
+        this.readThroughputSpanName = "readThroughput." + this.id;
+        this.readContainerSpanName = "readContainer." + this.id;
+        this.upsertItemSpanName = "upsertItem." + this.id;
+        this.deleteItemSpanName = "deleteItem." + this.id;
+        this.replaceItemSpanName = "replaceItem." + this.id;
+        this.createItemSpanName = "createItem." + this.id;
+        this.readAllItemsSpanName = "readAllItems." + this.id;
+        this.queryItemsSpanName = "queryItems." + this.id;
+        this.readAllConflictsSpanName = "readAllConflicts." + this.id;
+        this.queryConflictsSpanName = "queryConflicts." + this.id;
```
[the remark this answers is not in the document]
simplynaveen20: done

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
```diff
@@ -0,0 +1,175 @@
...
+                    parentContext.set(this.startSpan(spanName, databaseId, endpoint,
+                        context));
+                }
+            }).doOnSuccess(response -> {
+                if (!isNestedCall) {
+                    this.endSpan(parentContext.get(), Signal.complete(), statusCodeFunc.apply(response));
+                }
+            }).doOnError(throwable -> {
+                if (!isNestedCall) {
+                    this.endSpan(parentContext.get(), Signal.error(throwable), 0);
```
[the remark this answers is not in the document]
simplynaveen20: end span method have the logic of parsing cosmos exception from throwable , and sending status code. If exception is non cosmos then only this 0 will get entertain

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/util/CosmosPagedFlux.java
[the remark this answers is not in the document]
simplynaveen20: end span method have the logic of parsing cosmos exception from throwable , and sending status code. If exception is non cosmos then only this 0 will get entertain

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
simplynaveen20: Discussed with Sameeksha , once the documentation [here ](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-services-resource-providers) updates with azure.cosmosdb, we will change in the code too. Other azure services are also mentioned as Microsoft currently.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
simplynaveen20: With context having perf hit , if we do it on all, then non tracer user will get affected , already communicated to Sameeksha and Srikanta. We will track it as separate item and will introduce it on demand and we will then revisit perf again.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
simplynaveen20: As discussed keeping it the same i.e. AZ_TRACING_NAMESPACE_KEY on attributes, user will get extra warning log if he don't pass this. Will revisit it again in next tracing PR

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
simplynaveen20: Initially we had non cosmosResponse api returning Mono<Integer>, now we removed those api , so removing this method altogether

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/TracerProvider.java
simplynaveen20: Removed the error stacktrace

PR #10286, opened 2020-04-16
copy spotbugs and revapi from sdk parent

ChenTanyi:
After my test, there are a lot of errors for spotbugs, we need to fix it like checkstyle. For revapi, it seems to detect api change. We haven't released any package, so it do nothing.

run:
```
mvn spotbugs:check
```

### DIFF
--- a/sdk/management/pipelines/fluentJavaMgmt.yml
+++ b/sdk/management/pipelines/fluentJavaMgmt.yml
@@ -84,7 +84,7 @@ jobs:
         displayName: 'install'
         inputs:
           mavenPomFile: sdk/management/pom.xml
-          options: '$(DefaultOptions) $(ProfileFlag) -pl !samples javadoc:aggregate -DskipTests=true -Dcheckstyle.skip=true -DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -am'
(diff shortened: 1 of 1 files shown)


sdk/management/pom.xml
```diff
@@ -199,6 +199,79 @@
...
+                <dependencies>
+                <dependency>
+                    <groupId>com.azure</groupId>
+                    <artifactId>sdk-build-tools</artifactId>
+                    <version>1.0.0</version> <!-- {x-version-update;com.azure:sdk-build-tools;external_dependency} -->
+                </dependency>
+                <dependency>
+                    <groupId>com.github.spotbugs</groupId>
+                    <artifactId>spotbugs</artifactId>
+                    <version>4.0.0-beta3</version> <!-- {x-version-update;com.github.spotbugs:spotbugs;external_dependency} -->
```
yungezz, line 215: is the dependency version managed by parent pom?

ChenTanyi: Yes. For data-plane sdk, all plugins are managed by parent pom. https://github.com/azure/azure-sdk-for-java/blob/340ce8d5da791d4f21fa5ec35ece8832b490f8bf/sdk/parents/azure-client-sdk-parent/pom.xml#L176

weidongxu-microsoft:
Spotbugs on generated code is mostly on NPE protection. I will start fix it. Meanwhile if on might want to add ignore to "models" locally so one can focus on manual code.

PR #10299, opened 2020-04-16
Identity credential unavailable exception updates

g2vinay:


sdk/identity/azure-identity/src/main/java/com/azure/identity/ChainedTokenCredential.java
```diff
@@ -36,18 +38,29 @@
...
+                       return Mono.empty();
+                   }), 1)
+                   .next()
+                   .switchIfEmpty(Mono.defer(() -> {
+
+                       StringBuilder message = new StringBuilder("Tried "
+                             + credentials.stream().map(c -> c.getClass().getSimpleName())
+                                   .collect(Collectors.joining(", "))
+                             + " but failed to acquire a token for any of them. Please verify the"
+                             + " environment for either of them"
```
schaabs, line 54: Should we remove "either of them" since the list of credentials we try is more than 2?

sdk/identity/azure-identity/src/main/java/com/azure/identity/ChainedTokenCredential.java
```diff
@@ -36,18 +38,29 @@
...
-            }), 1)
-            .next()
-            .switchIfEmpty(Mono.defer(() -> Mono.error(new RuntimeException("Tried "
-                + credentials.stream().map(c -> c.getClass().getSimpleName()).collect(Collectors.joining(", "))
-                + " but failed to acquire a token for any of them. Please verify the environment for either of them"
-                + " and see more details in the causes below.", cause.get()))));
+                   .flatMap(p -> p.getToken(request).onErrorResume(CredentialUnavailableException.class, t -> {
+                       exceptions.add(t);
+                       return Mono.empty();
+                   }), 1)
```
srnagar, line 46: nit: don't need to specify concurrency as 1 for flatmap. It's 1 by default.

g2vinay: That's not true.
Default concurrency is computed via:
Math.max(16, Integer.parseInt(System.getProperty("reactor.bufferSize.small", "256")));

PR #10309, opened 2020-04-16
Field builder implementation

sima-zhu:
This is Skeleton design of FieldBuilder as we have no concrete saying of the implementation.

The basic idea is from [.Net track 1](https://github.com/Azure/azure-sdk-for-net/blob/feature/search/preview/sdk/search/Microsoft.Azure.Search.Service/src/Customizations/Indexes/FieldBuilder/FieldBuilder.cs) and [Bruce's prototype](https://github.com/brjohnstmsft/fsharp-doodles/blob/master/FieldPrototype/FieldPrototype/Field.fs)

Will seek feedback first before completing the code.

### DIFF
--- a/sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
+++ b/sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
@@ -1,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.search.documents;
+
+import com.azure.search.documents.models.ComplexSearchField;
+import com.azure.search.documents.models.SimpleDataType;
+import com.azure.search.documents.models.SimpleSearchField;
+
+/**
+ * Helper to build search field
+ */
+public class FieldBuilder {
+    public static SimpleSearchField getSimpleSearchField() {
+        return new SimpleSearchField();
+    }
+
+    public static ComplexSearchField getComplexSearchField() {
+        return new ComplexSearchField();
+    }
+}
+
--- a/sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexSearchField.java
+++ b/sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexSearchField.java
@@ -3,8 +3,50 @@
 
 package com.azure.search.documents.models;
 
+import java.util.List;
+import java.util.stream.Collectors;
+
 /**
  * A helper Field model to build a simple search field.
  */
 public class ComplexSearchField {
+    private String name;
+    private ComplexDataType dataType;
+    private List<Field> fieldList;
+
+    public String getName() {
+        return name;
+    }
+
+    public ComplexSearchField setName(String name) {
+        this.name = name;
+        return this;
+    }
+
+    public ComplexDataType getDataType() {
+        return dataType;
+    }
+
+    public ComplexSearchField setDataType(ComplexDataType dataType) {
+        this.dataType = dataType;
+        return this;
+    }
+
+    public List<Field> getFieldList() {
+        return fieldList;
+    }
+
+    public ComplexSearchField setSimpleFieldList(List<SimpleSearchField> fieldList) {
+        this.fieldList = fieldList.stream().map(SimpleSearchField::build).collect(Collectors.toList());
+        return this;
+    }
+
+    public ComplexSearchField setComplexFieldList(List<ComplexSearchField> fieldList) {
+        this.fieldList = fieldList.stream().map(ComplexSearchField::build).collect(Collectors.toList());
+        return this;
+    }
+
+    public Field build() {
+        return new Field().setName(name).setType(dataType.toDataType()).setFields(fieldList);
+    }
 }
--- a/sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/SimpleSearchField.java
+++ b/sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/SimpleSearchField.java
@@ -9,5 +9,107 @@ package com.azure.search.documents.models;
 public class SimpleSearchField {
     private String name;
     private PrimitiveType dataType;
-    private boolean searchable;
+//    private Boolean searchable;
+//    private Boolean filterable;
+//    private Boolean sortable;
+//    private Boolean facetable;
+//    private Boolean hidden;
+//    private AnalyzerName analyzer;
+//    private AnalyzerName searchAnalyzer;
+//    private AnalyzerName indexAnalyzer;
+
+    public String getName() {
+        return name;
+    }
+
+    public SimpleSearchField setName(String name) {
+        this.name = name;
+        return this;
+    }
+
+    public PrimitiveType getDataType() {
+        return dataType;
+    }
+
+    public SimpleSearchField setDataType(PrimitiveType dataType) {
+        this.dataType = dataType;
+        return this;
+    }
+//
+//    public Boolean getSearchable() {
+//        return searchable;
+//    }
+//
+//    public SimpleSearchField setSearchable(Boolean searchable) {
+//        this.searchable = searchable;
+//        return this;
+//    }
+//
+//    public Boolean getFilterable() {
+//        return filterable;
+//    }
+//
+//    public SimpleSearchField setFilterable(Boolean filterable) {
+//        this.filterable = filterable;
+//        return this;
+//    }
+//
+//    public Boolean getSortable() {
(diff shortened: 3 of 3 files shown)


sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexDataType.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.azure.search.documents.models;
+
+import com.azure.core.util.ExpandableStringEnum;
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+/**
+ * Simple DataType is used to create SimpleSearchField.
+ */
+public class ComplexDataType extends ExpandableStringEnum<ComplexDataType> {
```
alzimmermsft, line 12: Don't feel great with using an expandable enum here, I feel this will lead to a bad direction. Instead could this be a simple model class where it is basically wrapping a string?

```java
public class ComplexDataType {
  private final String type;
  
  private ComplexDataType(String type) {
    this.type = type;
  }

  public static final ComplexDataType EDM_COMPLEX_TYPE = new ComplexDataType(DataType.EDM_COMPLEX_TYPE.toString());

  public static ComplexDataType COLLECTION = new ComplexDataType(String.format("Collection(%s)", EDM_COMPLEX_TYPE.type));

  public toString() {
    return type;
  }
}
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/PrimitiveType.java
```diff
@@ -0,0 +1,63 @@
...
+
+package com.azure.search.documents.models;
+
+import com.azure.core.util.ExpandableStringEnum;
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+/**
+ * Primitive DataType for building SearchField.
+ */
+public class PrimitiveType extends ExpandableStringEnum<PrimitiveType>{
```
alzimmermsft, line 12: I believe this would be better off as a "enum" class of static strings.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexDataType.java
sima-zhu: Yes, redundant of using extensible. Agreed on this

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.azure.search.documents.models.ComplexSearchField;
+import com.azure.search.documents.models.SimpleDataType;
+import com.azure.search.documents.models.SimpleSearchField;
+
+/**
+ * Helper to build search field
+ */
+public class FieldBuilder {
+    public static SimpleSearchField getSimpleSearchField() {
```
conniey, line 14: Why does this builder exist when we can just run `new SimpleSearchField()`?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexSearchField.java
```diff
@@ -0,0 +1,52 @@
...
+    public ComplexSearchField setDataType(ComplexDataType dataType) {
+        this.dataType = dataType;
+        return this;
+    }
+
+    public List<Field> getFieldList() {
+        return fieldList;
+    }
+
+    public ComplexSearchField setSimpleFieldList(List<SimpleSearchField> fieldList) {
```
conniey, line 39: Why the suffixes with "List"? I think it's implied by the parameter name. Same with `getFieldList`. I think:

```java
public List<Field> getFields() {}
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: I am having this like a factory.
Just want to let customer know the type of Field they can use. Naming of the class is flexible to change.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexSearchField.java
sima-zhu: > Is there a reason this can't be done in Java? It looks like some normal reflection where you probe the properties for a JSON object and then set it.

Sorry, which part cannot be done in Java? The major purpose of the new field type is to help user build correct field (as much as possible) in local before sending out to the service.

The structure here is not all scenarios we have to cover. Will collect more points from .Net and service team.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,22 @@
...
+package com.azure.search.documents;
+
+import com.azure.search.documents.models.ComplexSearchField;
+import com.azure.search.documents.models.SimpleDataType;
+import com.azure.search.documents.models.SimpleSearchField;
+
+/**
+ * Helper to build search field
+ */
+public class FieldBuilder {
```
heaths, line 13: Where's the `build method that takes a type (class) and uses attributes to construct a bunch of fields, or even the whole index? That's what track 1 in .NET was doing.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.search.documents;
+
+import com.azure.search.documents.models.ComplexSearchField;
```
heaths, line 6: We had settled on `SimpleField`, `ComplexField`, and `SearchableField`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexDataType.java
heaths, line 12: Why is this exposed? There's only 1 type of complex, or a collection of them. Whether you want a collection could just be a boolean (that's what I did for .NET).

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexSearchField.java
heaths, line 39: Complex fields can also have complex fields. This method name doesn't work. I'd just keep the property `fields`, ie. `setFields`. That's what it is in track 1 and track 2 in .NET.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/SimpleDataType.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.azure.core.util.ExpandableStringEnum;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+/**
+ * Simple DataType is used to create SimpleSearchField.
+ */
+public class SimpleDataType extends ExpandableStringEnum<SimpleDataType> {
```
heaths, line 14: Why not just use the `SearchDataType` we already agreed on?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexSearchField.java
conniey: Heath's comment suggests that the behaviour in .NET is what this code should look like.
https://github.com/Azure/azure-sdk-for-java/pull/10309#pullrequestreview-395132635

sima-zhu: Thanks!

Will switch the field builder for taking the search document type.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
JonathanGiles: This class shouldn't be kept - as Connie notes it doesn't appear to offer any additional value to the user.

sima-zhu: I am working on rewrite the FieldBuilder. Thanks for comments!

sima-zhu:
Close the PR. Will open another one for FieldBuilder

Issue #10217, opened 2020-04-15
[BUG] Discriminator with "." in json is not handled correctly with FlatteningSerializer


weidongxu-microsoft:
**Describe the bug**
If a discriminator value with "." e.g. `odata.type` from monitor, and one of the class using it has `@JsonFlatten` annotation, the FlatteningSerializer would break that ".".

If I instead escape it as `property = "odata\\.type"`. The serialization is good, but de-serialization would then fail (subclass not serialized correctly).

***Exception or Stack Trace***
Add the exception log and stack trace if available

**To Reproduce**
```
    @JsonFlatten
    @Fluent
    public static class Foo {
        @JsonProperty(value = "properties.criteria", required = true)
        private MetricAlertCriteria criteria;

        public MetricAlertCriteria criteria() {
            return this.criteria;
        }

        public Foo withCriteria(MetricAlertCriteria criteria) {
            this.criteria = criteria;
            return this;
        }
    }

    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "odata.type", defaultImpl = com.azure.management.monitor.MetricAlertCriteria.class)
    @JsonTypeName("MetricAlertCriteria")
    @JsonSubTypes({
        @JsonSubTypes.Type(name = "Microsoft.Azure.Monitor.SingleResourceMultipleMetricCriteria", value = MetricAlertSingleResourceMultipleMetricCriteria.class),
        @JsonSubTypes.Type(name = "Microsoft.Azure.Monitor.MultipleResourceMultipleMetricCriteria", value = MetricAlertMultipleResourceMultipleMetricCriteria.class)
    })
    @Fluent
    public static class MetricAlertCriteria {

    }

    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "odata.type")
    @JsonTypeName("Microsoft.Azure.Monitor.SingleResourceMultipleMetricCriteria")
    @Fluent
    public static final class MetricAlertSingleResourceMultipleMetricCriteria extends MetricAlertCriteria {

    }

    @Test
    public void testSerialization() throws Exception {
        SerializerAdapter adapter = new AzureJacksonAdapter();

        Foo foo = new Foo();
        foo.withCriteria(new MetricAlertSingleResourceMultipleMetricCriteria());
        String fooJson = adapter.serialize(foo, SerializerEncoding.JSON);
[...]

Issue #8372, opened 2020-02-21
[QUERY] AzureJacksonAdapter fails to serialize map with key of resource ID


xseeseesee:
**Query/Question**
We use AzureJacksonAdapter to serialize the [VirtualMachineInner](https://github.com/Azure/azure-libraries-for-java/blob/vnext-compute-fix-tests/azure-mgmt-compute/src/main/java/com/azure/management/compute/models/VirtualMachineInner.java). It doesn't work as expected when parsing the map inside the property of `identity`. 

***Why is this not a Bug or a feature Request?***
Not sure if we missed adding some specific annotation in the object.

Code snippet to reproduce:
```java
        VirtualMachineInner inner = new VirtualMachineInner();
        VirtualMachineIdentity identity = new VirtualMachineIdentity();
        Map<String, VirtualMachineIdentityUserAssignedIdentities> map = new HashMap<>();
        map.put("/subscriptions/0-0-0-0-0/resourcegroups/0/providers/Microsoft.ManagedIdentity/userAssignedIdentities/0",
                new VirtualMachineIdentityUserAssignedIdentities());
        identity.withType(ResourceIdentityType.SYSTEM_ASSIGNED_USER_ASSIGNED);
        identity.withUserAssignedIdentities(map);
        inner.withIdentity(identity);
```
If we only serialize identity, we will get what we need
```json
{
  "type": "SystemAssigned, UserAssigned",
  "userAssignedIdentities": {
    "/subscriptions/0-0-0-0-0/resourcegroups/0/providers/Microsoft.ManagedIdentity/userAssignedIdentities/0": {}
  }
}
```
But the actual what we get from inner is
```json
  "identity": {
    "type": "SystemAssigned, UserAssigned",
    "userAssignedIdentities": {
      "/subscriptions/0-0-0-0-0/resourcegroups/0/providers/Microsoft": {
        "ManagedIdentity/userAssignedIdentities/0": {}
      }
    }
  },
```

**Setup (please complete the following information if applicable):**
 - OS: Windows
 - IDE : IntelliJ
 - Version of the Library used
 
 **Information Checklist**
 Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ ] Query Added
- [ ] Setup information Added

xccc-msft:
[...]

PR #10310, opened 2020-04-16
Serialization and deserialization handling of types with discriminator

anuchandy:
This PR brings in complete support for serialization and deserialization of discriminator types. Discriminator support and flattening bug fixes were the work I already did for the track1 runtime. But that runtime work and my initial work on azure-core overlapped, and the feature never gets sync-ed. 

The main improvements are - 
  (1. handling of type-Id containing dots that were colliding with flattening dots. 
  (2. Flattening of nested types
  (3. Increase flattening code coverage.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
@@ -3,6 +3,7 @@
 
 package com.azure.core.util.serializer;
 
+import com.azure.core.annotation.JsonFlatten;
 import com.azure.core.implementation.TypeUtil;
 import com.azure.core.util.CoreUtils;
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -16,7 +17,6 @@ import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.deser.BeanDeserializer;
 import com.fasterxml.jackson.databind.deser.BeanDeserializerModifier;
 import com.fasterxml.jackson.databind.deser.ResolvableDeserializer;
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
@@ -71,7 +71,7 @@ final class FlatteningDeserializer extends StdDeserializer<Object> implements Re
             public JsonDeserializer<?> modifyDeserializer(DeserializationConfig config,
                                                           BeanDescription beanDesc,
                                                           JsonDeserializer<?> deserializer) {
-                if (BeanDeserializer.class.isAssignableFrom(deserializer.getClass())) {
+                if (beanDesc.getBeanClass().getAnnotation(JsonFlatten.class) != null) {
                     // Register 'FlatteningDeserializer' for complex type so that 'deserializeWithType'
                     // will get called for complex types and it can analyze typeId discriminator.
                     return new FlatteningDeserializer(beanDesc.getBeanClass(), deserializer, mapper);


sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
```diff
@@ -65,58 +68,185 @@ public static SimpleModule getModule(final ObjectMapper mapper) {
...
+    public Object deserializeWithType(JsonParser jp,
+                                      DeserializationContext cxt,
+                                      TypeDeserializer tDeserializer) throws IOException {
+        // This method will be called from Jackson for each "Json object with TypeId" as it
+        // process the input data. This enable us to pre-process then give it to the next
+        // deserializer in the Jackson pipeline.
+        //
+        // The parameter 'jp' is the reader to read "Json object with TypeId"
+        //
+        JsonNode currentJsonNode = mapper.readTree(jp);
```
alzimmermsft, line 97: Should we just cast this to `ObjectNode` ahead of time so we don't need to cast all the time later in the method?

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
```diff
@@ -65,58 +68,185 @@ public static SimpleModule getModule(final ObjectMapper mapper) {
...
+    /**
+     * Given a field of a POJO class and JsonNode corresponds to the same POJO class,
+     * check field's {@link JsonProperty} has flattening dots in it if so
+     * flatten the nested child JsonNode corresponds to the field in the given JsonNode.
+     *
+     * @param classField the field in a POJO class
+     * @param jsonNode the json node corresponds to POJO class that field belongs to
+     */
+    @SuppressWarnings("unchecked")
+    private static void handleFlatteningForField(Field classField, JsonNode jsonNode) {
```
alzimmermsft, line 157: Should we just use `ObjectNode`?

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
```diff
@@ -65,58 +68,185 @@ public static SimpleModule getModule(final ObjectMapper mapper) {
         SimpleModule module = new SimpleModule();
         module.setDeserializerModifier(new BeanDeserializerModifier() {
             @Override
-            public JsonDeserializer<?> modifyDeserializer(DeserializationConfig config, BeanDescription beanDesc,
+            public JsonDeserializer<?> modifyDeserializer(DeserializationConfig config,
+                                                          BeanDescription beanDesc,
                                                           JsonDeserializer<?> deserializer) {
-                if (beanDesc.getBeanClass().getAnnotation(JsonFlatten.class) != null) {
+                if (BeanDeserializer.class.isAssignableFrom(deserializer.getClass())) {
```
weidongxu-microsoft, line 74: I am not sure if opening it up is good or bad at this point.

Here is a bug in track1 that kind of *caused* by open-up if there is any bug in Flattening (not really, but previous class without "@JsonFlatten" will not be affected, but now it will)
https://github.com/Azure/autorest-clientruntime-for-java/issues/663

And we got this one there
https://github.com/Azure/azure-sdk-for-java/issues/8372

Just saying it could be risky if there is unfixed or even unknown bug.

anuchandy: I think we're safe to do this in "deserializer" because in deserialization path we always check for `JsonProperty` value and if it has un-escaped dots then it means "do flatten", right?

\\cc @jianghaolu - please chime if this interpretation in "deserialization" context is wrong

weidongxu-microsoft: Sorry, I think I mean the same place in FlatteningSerializer (it is opened-up as well). Both bug referred is happening in serialization stage.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
anuchandy: I'm not sure we should do the upcasting ahead - w.r.t `jsonProperty` lookup and modification in this method we're trying not to move away from the current logic (in released versions) to mutate ObjectNode. The `ObjectNode` is type extending from `JsonNode` and today we do the upcasting only if we find it has `jsonProperty`, but if we have to take `ObjectNode` here then we had to do the upcasting for all nodes from the caller. 

An example is - In calling side, when we attempted to assign the value of readTree to `ObjectNode`, there happened a runtime cast exception where readTree returned `NullNode` which is not an `ObjectNode` but `JsonNode`, and auto-casting based on receiving variable type failed.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
anuchandy: Please see the below [comment](https://github.com/Azure/azure-sdk-for-java/pull/10310#discussion_r410349470) on upcasting based on the presence of `jsonProperty`, similar constraint applied here but based on the presence of `JsonTypeInfo`.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
```diff
@@ -65,58 +68,185 @@ public static SimpleModule getModule(final ObjectMapper mapper) {
...
-                                break;
-                            }
-                        }
-                        // If additional properties have a conflicting key, escape the additional property's key
-                        if (root.has(value)) {
-                            String escapedValue = value.replace(".", "\\.");
-                            ((ObjectNode) root).set(escapedValue, root.get(value));
+            } else {
+                final JsonTypeInfo typeInfo = c.getAnnotation(com.fasterxml.jackson.annotation.JsonTypeInfo.class);
+                if (typeInfo != null) {
```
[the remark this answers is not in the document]
anuchandy: cool, yes we can invert this and make more readable.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
anuchandy: For reference: upcasting in released versions [here](https://github.com/Azure/azure-sdk-for-java/blob/4a22178c595320eb1fb658d13dae0591154a9b73/sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java#L106), see line-91  where we do check for annotation presence before the upcast, the same pattern followed in this PR.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
anuchandy: Right, this PR is to address the discriminated type "deserialization" issue, https://github.com/Azure/azure-sdk-for-java/issues/10217.

Thanks for the heads-up on the "serialization" path. Reverted the  FlatteningSerializer, so there won't be any change in the `FlatteningSerializer` and "serialization" logic stay as it is. This means the track1 issue with template "serialization" "663" will not happen when using azure-core. The "8372" is a different issue, and this PR does not address it.

anuchandy: @weidongxu-microsoft  - [reg](https://github.com/Azure/azure-sdk-for-java/pull/10310#discussion_r409970231) -  I had a chat with Jinaghao on AutoRest logic associated with escaping dots in `JsonProperty`. Such escaping will be done only if the properties in the parent type have to include non-escaped dots, this means we should limit the application of Deserializer only to JsonFlatten annotated types.

anuchandy:
For updated Serializer and De-Serializer to pick up discriminator with dot value, the following constraints must be met:

1. Any dot "." in the discriminator in `JsonTypeInfo` annotation must be escaped. E.g. `odata.type` in swagger become `odata\\.type` in java.
2. If a type is annotated with `JsonTypeInfo` with discriminator containing dot then it also needs to have `JsonFlatten` annotation.


#### Example:

```java
// ---------------------------------------------------------
//
@JsonFlatten
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME,
    include = JsonTypeInfo.As.PROPERTY,
    property = "@odata\\.type",
    defaultImpl = Animal.class)
@JsonTypeName("Animal")
@JsonSubTypes({
    @JsonSubTypes.Type(name = "#Favourite.Pet.Dog", value = Animal.class)
})
public class Animal {
}
//
// ---------------------------------------------------------
//
@JsonFlatten
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME,
    include = JsonTypeInfo.As.PROPERTY,
    property = "@odata\\.type",
    defaultImpl = Dog.class)
@JsonTypeName("#Favourite.Pet.Dog")
public class Dog extends Animal {
    @JsonProperty(value = "breed")
    private String breed;

    // Flattenable property
    @JsonProperty(value = "properties.cuteLevel")
    private Integer cuteLevel;

    public String breed() {
        return this.breed;
    }

    public DogWithTypeIdContainingDot withBreed(String audioLanguage) {
        this.breed = audioLanguage;
        return this;
    }

    public Integer cuteLevel() {
        return this.cuteLevel;
    }

    public DogWithTypeIdContainingDot withCuteLevel(Integer cuteLevel) {
        this.cuteLevel = cuteLevel;
        return this;
    }
}
//
// ---------------------------------------------------------
```

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningDeserializer.java
weidongxu-microsoft: @anuchandy Got it, as well as the required "@JsonFlatten" on there classes as well.

One question, at least for the ARM template (which should not have any flatten in there), should we always use `readValue` to convert json to an `Object` (as `LinkedHashMap`), instead of using `readTree` to `ObjectNode`, so that if FlattenSerializer got involved, it could still escape the "." as it is now a `Map` with nested `Map`s?

PR #10313, opened 2020-04-17
Add TestWatcher to Print Testing Information

alzimmermsft:
This PR adds implementations of JUnit 5's before and after test extensions and creates an SPI that will be loaded during the test run. This follows the documentation here: https://junit.org/junit5/docs/5.0.1/user-guide/#extensions.


sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -105,6 +111,14 @@
...
             <usePhrasedTestCaseMethodName>true</usePhrasedTestCaseMethodName>
           </statelessTestsetReporter>
+          <properties>
+            <configurationParameters>
+              junit.jupiter.conditions.deactivate = *
+              junit.jupiter.extensions.autodetection.enabled = true
+              junit.jupiter.testinstance.lifecycle.default = per_class
+              junit.jupiter.execution.parallel.enabled = true
+            </configurationParameters>
+          </properties>
```
alzimmermsft, line 121: Let's aim to remove these on-off configurations from children POMs where possible.

srnagar: Yes, ideally, this surefire plugin should not be in these modules. It should be configured only in parent.

alzimmermsft: This was removed.

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -93,20 +93,11 @@
...
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-test-watcher</artifactId>
+      <version>1.0.0-beta.1</version> <!-- {x-version-update;com.azure:azure-test-watcher;dependency} -->
+      <scope>test</scope>
+    </dependency>
   </dependencies>
-
```
sima-zhu, line 97: Could you port the surefire extension change to check the display in both console and report?

alzimmermsft: Is this what you were looking for in the console output?

image

sima-zhu: IDE is working always. Need to check the surefire report under target.
image
image

eng/azure-test-watcher/src/main/java/com/azure/testing/package-info.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Package containing JUnit 5 testing extensions.
+ */
+package com.azure.testing;
```
conniey, line 7: nit: add new line.

pom.xml
```diff
@@ -8,6 +8,7 @@
   <packaging>pom</packaging>
   <version>1.0.0</version><!-- Need not change for every release-->
   <modules>
+    <module>eng/azure-test-watcher</module>
```
conniey, line 11: Is there a reason this can't be lumped into azure-core-test? It's only used for tests. It'll give azure-core-test a reason to live.

alzimmermsft: Won't be usable in Azure Core due to circular dependencies.

JimSuplizio: @alzimmermsft this should probably be in the common with perf-test-core and smoke-tests, instead of eng.

alzimmermsft: Moving this into `common` makes sense to me, will move it in the next commit.

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
alzimmermsft: This still has `surefire-reports` in target.

eng/azure-test-watcher/src/main/java/com/azure/testing/package-info.java
alzimmermsft: Added

pom.xml
alzimmermsft: Moved into common

common/azure-test-watcher/src/main/java/com/azure/testing/AzureTestWatcher.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    @Override
+    public void afterTestExecution(ExtensionContext context) {
+        Method testMethod = context.getRequiredTestMethod();
+        long start = getStore(context).remove(testMethod, long.class);
+        long duration = System.currentTimeMillis() - start;
+
+        System.out.printf("completed in %d ms.%n", duration);
```
conniey, line 44: nit: completed -> Completed

common/azure-test-watcher/src/main/java/com/azure/testing/AzureTestWatcher.java
```diff
@@ -0,0 +1,50 @@
...
+        String testName = "";
+        String fullyQualifiedTestName = "";
+        if (extensionContext.getTestMethod().isPresent()) {
+            Method method = extensionContext.getTestMethod().get();
+            testName = method.getName();
+            fullyQualifiedTestName = method.getDeclaringClass().getName() + "." + testName;
+        }
+
+        if (!Objects.equals(displayName, testName)) {
+            System.out.printf("Starting test %s (%s), ", fullyQualifiedTestName, displayName);
```
conniey, line 32: Do you need a `%n` for these too?

common/azure-test-watcher/pom.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.azure</groupId>
+  <artifactId>azure-test-watcher</artifactId>
+  <version>1.0.0-beta.1</version>
```
conniey, line 9: Should this line have that x-version-update metadata?

sdk/parents/azure-client-sdk-parent/pom.xml
```diff
@@ -1458,6 +1464,7 @@
 
   <modules>
     <module>../../../common/perf-test-core</module>
+    <module>../../../common/azure-test-watcher</module>
```
conniey, line 1467: nit: alphabetise

common/azure-test-watcher/src/main/java/com/azure/testing/AzureTestWatcher.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public void afterTestExecution(ExtensionContext context) {
+        Method testMethod = context.getRequiredTestMethod();
+        long start = getStore(context).remove(testMethod, long.class);
+        long duration = System.currentTimeMillis() - start;
+
+        System.out.printf("completed in %d ms.%n", duration);
+    }
+
+    private ExtensionContext.Store getStore(ExtensionContext context) {
```
conniey, line 47: this could be a static method because it doesn't rely on any state from its parent class.

common/azure-test-watcher/src/main/java/com/azure/testing/AzureTestWatcher.java
alzimmermsft: The idea was that this and the other print statement would form a single line.

Test begins, console view: 

`Starting test {qualified name} ({display name})`

Test completes, console view:

`Starting test {qualified name} ({display name}), completed in {millis} ms.{newline}`.

common/azure-test-watcher/pom.xml
alzimmermsft: This won't be managed by the versioning script, it should be treated the same as the code-quality-reports and jacoco projects.

common/azure-test-watcher/src/main/java/com/azure/testing/AzureTestWatcher.java
alzimmermsft: Good point, will replace `getClass()` with `AzureTestWatcher.class` and make this method static.

sdk/parents/azure-client-sdk-parent/pom.xml
alzimmermsft: Done

PR #10315, opened 2020-04-17
azure-core-management add unit tests

weidongxu-microsoft:


sdk/core/azure-core-management/src/test/java/com/azure/core/management/CloudExceptionTests.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.io.IOException;
+
+public class CloudExceptionTests {
+
+    @Test
+    public void deserialization() throws IOException {
+        // actually a test on CloudErrorDeserializer
+
+        final String errorBody = "{\"error\":{\"code\":\"ResourceGroupNotFound\",\"message\":\"Resource group 'rg-not-exist' could not be found.\"}}";
```
yungezz, line 19: constants code?

weidongxu-microsoft: A sample response from ARM.

Java does not have constants, so `final` should be all?

PR #10324, opened 2020-04-17
Resource: fix spotbugs

ChenTanyi:
Remaining auto-generated issue.

### DIFF
--- a/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/dag/TaskGroup.java
+++ b/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/dag/TaskGroup.java
@@ -648,7 +648,7 @@ public class TaskGroup
      * "post run" dependents are those TaskGroup which need to be invoked as part of invocation
      * of actual TaskGroup.
      */
-    static final class ProxyTaskGroupWrapper {
+    protected static final class ProxyTaskGroupWrapper {
         // The "proxy TaskGroup"
         private TaskGroup proxyTaskGroup;
         // The "actual TaskGroup" for which above TaskGroup act as proxy
--- a/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/dag/Graph.java
+++ b/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/dag/Graph.java
@@ -165,7 +165,7 @@ public class Graph<DataT, NodeT extends Node<DataT, NodeT>> {
     /**
      * The edge types in a graph.
      */
-    enum EdgeType {
+    protected enum EdgeType {
         /**
          * An edge (u, v) is a tree edge if v is visited the first time.
          */
@@ -190,7 +190,7 @@ public class Graph<DataT, NodeT extends Node<DataT, NodeT>> {
      *
      * @param <U> the type of the node
      */
-    interface Visitor<U> {
+    protected interface Visitor<U> {
         /**
          * visit a node.
          *
--- a/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/dag/TaskGroupEntry.java
+++ b/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/dag/TaskGroupEntry.java
@@ -14,7 +14,7 @@ import reactor.core.publisher.Mono;
  *
  * @param <TaskT> the task type that can return a value
  */
-final class TaskGroupEntry<TaskT extends TaskItem>
+public final class TaskGroupEntry<TaskT extends TaskItem>
         extends DAGNode<TaskT, TaskGroupEntry<TaskT>> {
     /**
      * The proxy entry for this entry if exists.


sdk/management/pom.xml
```diff
@@ -91,6 +91,12 @@
                 <artifactId>azure-storage-blob</artifactId>
                 <version>12.4.0</version>
             </dependency>
+            <dependency>
+                <groupId>com.github.spotbugs</groupId>
+                <artifactId>spotbugs-annotations</artifactId>
+                <version>4.0.0-beta3</version>
```
xccc-msft, line 97: Any reason we pick this beta version? The latest is [4.0.2](https://mvnrepository.com/artifact/com.github.spotbugs/spotbugs-annotations/4.0.2).

sdk/resources/mgmt/src/main/java/com/azure/management/RestClientBuilder.java
```diff
@@ -29,7 +29,7 @@
  * A rest client builder.
  */
 @ServiceClientBuilder(serviceClients = RestClient.class)
-public final class RestClientBuilder {
+public final class RestClientBuilder implements Cloneable {
```
weidongxu-microsoft, line 32: Maybe we should change the name of `clone` method. Maybe `newBuilder`. (Because I didn't really want a `Cloneable` when adding that method.)

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/ExpandableStringEnum.java
```diff
@@ -15,36 +16,34 @@
  * @param <T> a specific expandable enum type
  */
 public abstract class ExpandableStringEnum<T extends ExpandableStringEnum<T>> {
```
weidongxu-microsoft, line 18: There should be a same class in core. May be we can remove this later.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/collection/implementation/CreatableResourcesImpl.java
```diff
@@ -79,12 +79,12 @@ protected CreatableResourcesImpl() {
      *
      * @param <ResourceT> the type of the resources in the batch.
      */
-    private class CreatedResourcesImpl<ResourceT extends Indexable>
+    private static class CreatedResourcesImpl<ResourceT extends Indexable>
             extends HashMap<String, ResourceT>
             implements CreatedResources<ResourceT> {
-        private final ClientLogger logger = new ClientLogger(this.getClass());
+        private final transient ClientLogger logger = new ClientLogger(this.getClass());
```
weidongxu-microsoft, line 85: Please detail on what happens here. `transient` should not be used lightly.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/collection/implementation/IndependentChildrenImpl.java
```diff
@@ -81,7 +79,7 @@ public T getById(String id) {
     @Override
     public Mono<T> getByIdAsync(String id) {
         ResourceId resourceId = ResourceId.fromString(id);
-        if (resourceId == null) {
+        if (resourceId.parent() == null) {
```
weidongxu-microsoft, line 82: why `parent()`? maybe only delete this line?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/models/implementation/IndependentChildImpl.java
```diff
@@ -68,6 +68,10 @@ public String resourceGroupName() {
         }
     }
 
+    protected String parentName() {
```
weidongxu-microsoft, line 71: What error message associates with this?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/dag/TaskGroup.java
```diff
@@ -337,6 +339,7 @@ public void addPostRunDependentTaskGroup(TaskGroup dependentTaskGroup) {
     /**
      * @return list with current task entries in this task group
      */
+    @SuppressFBWarnings(value = "BC_UNCONFIRMED_CAST_OF_RETURN_VALUE", justification = "Incorrect spot bugs")
```
weidongxu-microsoft, line 342: Currently the norm in sdk is to add it to suppression
```
  <Match>
    <Class name="~com\.microsoft\.azure\.keyvault\.requests\.\w+Request"/>
    <Bug pattern="BC_UNCONFIRMED_CAST_OF_RETURN_VALUE"/>
  </Match>
```

We could keep the annotation, and see what to do after more services get checked.

sdk/management/pom.xml
ChenTanyi: This is aligned with the plugin version. https://github.com/azure/azure-sdk-for-java/blob/5810f30086c93a1431deeae7f0a10f890b0012d6/sdk/management/pom.xml#L221, I'm not sure if these should be aligned or not. Maybe we could discuss about.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/collection/implementation/CreatableResourcesImpl.java
ChenTanyi: The spotbugs said the CreatableResourcesImpl has two variable that cannot be serialized. After my search, it seems add transient can fix it.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/collection/implementation/IndependentChildrenImpl.java
ChenTanyi: Since the next line has `resourceId.parent().name()`, so I add `parent()` in it. Is it better not to check this NPE?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/models/implementation/IndependentChildImpl.java
ChenTanyi: The error message is we could not expose a protected / public variable that hasn't been used. So I add this function to use the variable.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/dag/TaskGroup.java
ChenTanyi: Agree, if much more services meet such an error spotbugs, we could add it to suppression.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/collection/implementation/CreatableResourcesImpl.java
weidongxu-microsoft: Could you see if it is possible to convert the class to composition (use a HashMap var, instead of extends HashMap, then delegate other Map method to that var)? This way we should be able to get rid of `Serializable` interface.

sdk/management/pom.xml
ChenTanyi: I've change it to latest version

sdk/resources/mgmt/src/main/java/com/azure/management/RestClientBuilder.java
ChenTanyi: done

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/ExpandableStringEnum.java
ChenTanyi: removed

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/collection/implementation/CreatableResourcesImpl.java
ChenTanyi: Changed extends to private var. Done

PR #10340, opened 2020-04-17
Updates azure-sdk-bom Version in Preparation for Release

alzimmermsft:


sdk/boms/azure-sdk-bom/pom.xml
```diff
@@ -5,7 +5,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.azure</groupId>
   <artifactId>azure-sdk-bom</artifactId>
-  <version>1.0.0-beta.1</version>
+  <version>1.0.0</version>
```
srnagar, line 8: Should this have a version tag? Makes it easy for e2e and spring modules to refer to the bom and update when this gets updated.

alzimmermsft: I can add one, the BOM isn't version managed in the same way the other modules are, @JimSuplizio thoughts?

alzimmermsft:
> Should we consider adding our direct dependencies to the BOM (JacksonXML, etc), at least temporarily?

https://github.com/Azure/azure-sdk-for-java/issues/10342

sdk/boms/azure-sdk-bom/pom.xml
JimSuplizio: The answer is no, there will be no tags in the bom file. First, the bom file was only supposed to be released periodically (every 3-6 months according to the original conversation) and second, my automation specifically disallows dependencyManagement sections which, you know, is pretty much what a bom file is.

PR #10341, opened 2020-04-17
Move surefire extension to parent

sima-zhu:


sima-zhu:
> Also cleanup the KeyVault POMs that are using this configuration.

KeyVault has rurunOnFailure configuration. @g2vinay can rerunFailingTestsCount remove from surefire configuration?

PR #10349, opened 2020-04-20
Transitively require all jackson modules

srnagar:
In azure-core module-info.java 3 out of 5 jackson modules were transitively required. This PR updates the module-info.java to require all jackson modules transitively. This will allow all client libraries to just require on com.azure.core module to get all jackson modules. 

Note: I have not updated any client libraries that now require jackson modules directly as that would require taking a dependency on unreleased version of azure-core. This update can wait until after next version of azure-core is released or any other significant change that warrants taking an unreleased dependency on azure-core.

### DIFF
--- a/sdk/core/azure-core/src/main/java/module-info.java
+++ b/sdk/core/azure-core/src/main/java/module-info.java
@@ -10,8 +10,8 @@ module com.azure.core {
     requires transitive com.fasterxml.jackson.core;
     requires transitive com.fasterxml.jackson.databind;
 
-    requires com.fasterxml.jackson.dataformat.xml;
-    requires com.fasterxml.jackson.datatype.jsr310;
+    requires transitive com.fasterxml.jackson.dataformat.xml;
+    requires transitive com.fasterxml.jackson.datatype.jsr310;
 
     // public API surface area
     exports com.azure.core.annotation;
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/main/java/module-info.java
```diff
@@ -10,8 +10,8 @@
     requires transitive com.fasterxml.jackson.core;
     requires transitive com.fasterxml.jackson.databind;
 
-    requires com.fasterxml.jackson.dataformat.xml;
-    requires com.fasterxml.jackson.datatype.jsr310;
+    requires transitive com.fasterxml.jackson.dataformat.xml;
+    requires transitive com.fasterxml.jackson.datatype.jsr310;
```
alzimmermsft, line 14: I thought requires transitive was only needed when the module is exposed in public API?

srnagar: No, `requires transitive` allows client libraries to access it if they just add `requires com.azure.core` to their respective module-info.java. A few client libraries like Cosmos, Storage and Search are not explicitly adding `requires com.fasterxml.jackson.dataformat.xml` or `requires  com.fasterxml.jackson.datatype.jsr310` which can be avoided with this change.

PR #10350, opened 2020-04-20
Restore dependency-whitelist.html

srnagar:
Restore dependency-whitelist.html and move full whitelist section to the bottom.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/dependency-whitelist.html
+++ b/eng/code-quality-reports/src/main/resources/dependency-whitelist.html
@@ -1,0 +1,90 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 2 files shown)


mitchdenny:
Is there any particular reason why these files are listed as resources in this module? The module doesn't do anything with the resource? All of the assembly of the aggregate report is done by a PowerShell script.

srnagar:
>Is there any particular reason why these files are listed as resources in this module? The module doesn't do anything with the resource? All of the assembly of the aggregate report is done by a PowerShell script.

These are not directly used by the `code-quality-reports` modules but they are html pages that render the reports. So, these html pages are added as resources to that module. I had initially put them in eng/common but that got deleted as it syncs with sdk-tools. I didn't find any other appropriate directory to host these files. I am open for suggestions to move to another directory or create a new directory.

PR #10351, opened 2020-04-20
Exclude NP_LOAD_OF_KNOWN_NULL_VALUE from the files generated by autorest

weidongxu-microsoft:
I only exclude NP_LOAD_OF_KNOWN_NULL_VALUE for generated code.

---

Generated code for method overload would be this

```
    public PagedFlux<ResourceGroupInner> listAsync() {
        final String filter = null;
        final Integer top = null;
        return new PagedFlux<>(() -> listSinglePageAsync(filter, top), nextLink -> listNextSinglePageAsync(nextLink));
    }
```
which is a overload of `PagedFlux<ResourceGroupInner> listAsync(String filter, Integer top)`.

There is not much point to move the `null` to parameter, for generated code.


weidongxu-microsoft:
Add `SKIPPED_CLASS_TOO_BIG` to generated code. Not much to be done.

PR #10364, opened 2020-04-20
fix: cosmos spotbugs

ChenTanyi:


sdk/cosmos/mgmt/src/main/java/com/azure/management/cosmosdb/implementation/CosmosDBAccountImpl.java
```diff
@@ -622,7 +603,7 @@ private void addLocationsForParameters(HasLocations locationParameters, List<Fai
                                         .flatMap(
                                             index -> {
                                                 data.set(0, data.get(0) + 30);
-                                                return Mono.delay(manager().getSdkContext().getLroRetryDuration());
+                                                return Mono.delay(SdkContext.getLroRetryDuration());
```
weidongxu-microsoft, line 606: Why it using lro regry duration?

ChenTanyi: The cosmos multi location prevision is async. It will return succeeded when first location is done. This func I think is to ensure all location is succeeded.

weidongxu-microsoft: Yes, but why use `getLroRetryDuration`? It should be only used for LRO, but this is not.

weidongxu-microsoft: You can use `getDelayDuration` with a specified delay.

ChenTanyi: This is a LRO situation, just isn't a common LRO situation.

ChenTanyi: Long running operation to wait all location prevision complete.

weidongxu-microsoft: An operation takes long time does not means that it is LRO.

The specific var is used here for the `PollerFlux`.

```
   public <T, U> PollerFlux<PollResult<T>, U> getLroResultAsync(Mono<SimpleResponse<Flux<ByteBuffer>>> lroInit,
                                                                 HttpPipeline httpPipeline,
                                                                 Type pollResultType, Type finalResultType) {
        return PollerFactory.create(
            getSerializerAdapter(),
            httpPipeline,
            pollResultType,
            finalResultType,
            SdkContext.getLroRetryDuration(),
            activationOperation(lroInit)
        );
    }
```

weidongxu-microsoft: Track1 is such

```
                        ).repeatWhen(new Func1<Observable<? extends java.lang.Void>, Observable<?>>() {
                            @Override
                            public Observable<?> call(Observable<? extends Void> observable) {
                                data.set(0, data.get(0) + 5);
                                return observable.delay(5, TimeUnit.SECONDS);
                            }
                        })
```

ChenTanyi: Done, change to `getDelayDuration`.

PR #10368, opened 2020-04-20
Vs Code Credential

g2vinay:


sdk/identity/azure-identity/pom.xml
```diff
@@ -93,6 +93,12 @@
       <version>2.8.5</version> <!-- {x-version-update;com.google.code.gson:gson;external_dependency} -->
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>com.microsoft.azure</groupId>
+      <artifactId>msal4j-persistence-extension</artifactId>
+      <version>0.1</version>
```
g2vinay, line 100: add x version, and update external dep file.

sdk/identity/azure-identity/src/main/java/com/azure/identity/VisualStudioCodeCredential.java
```diff
@@ -0,0 +1,56 @@
...
+import com.azure.identity.implementation.IdentityClientOptions;
+import com.azure.identity.implementation.MsalToken;
+import reactor.core.publisher.Mono;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Enables authentication to Azure Active Directory using data from Visual Studio Code
+ */
+public class VisualStudioCodeCredential implements TokenCredential {
```
schaabs, line 20: This type shouldn't be public. For now we are adding these credential types to the `DefaultAzureCredential` authentication flow but not exposing the credential types directly.

sdk/identity/azure-identity/src/main/java/com/azure/identity/VisualStudioCodeCredentialBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT License.
+
+package com.azure.identity;
+
+/**
+ * Fluent credential builder for instantiating a {@link VisualStudioCodeCredential}.
+ *
+ * @see VisualStudioCodeCredential
+ */
+public class VisualStudioCodeCredentialBuilder extends CredentialBuilderBase<DefaultAzureCredentialBuilder> {
```
schaabs, line 11: Remove this builder type.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -429,6 +434,85 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+
+            if (userSettings.has("azure.cloud")) {
+                cloud = userSettings.get("azure.cloud").asText();
+            }
+        }
+
+        String credential = accessor.getCredentials("VS Code Azure", cloud);
+
+        String authority = KnownAuthorityHosts.AZURE_CLOUD.replaceAll("/+$", "")
+            + "/organizations/" + tenant;
```
schaabs, line 468: The authority has to be configurable why is this hard-coded to public cloud? Also I think it should append EITHER '/organizations/' OR tenantId. At least that's what we do in other languages. Have you confirmed that /orginizations/<tenantId> actually works?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -429,6 +434,85 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+            JsonNode parsedCredentials = mapper.readTree(credential);
+            String redirectUri = parsedCredentials.get("redirectionUrl").asText();
+            String authorizationCode = parsedCredentials.get("code").asText();
+
+            AuthorizationCodeParameters authorizationCodeParameters = AuthorizationCodeParameters
+                .builder(authorizationCode, new URI(redirectUri))
+                .scopes(new HashSet<>(request.getScopes()))
+                .build();
+
+            PublicClientApplication application = publicClientApplicationBuilder.authority(authority).build();
```
schaabs, line 487: In .NET we don't handle the case where the authorization code is stored, only the case where the stored value is a refresh token. We should align this.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
schaabs: To be more precise here the authority should be determined by the azure.cloud vscode setting, and the tenantId should use whatever was passed in through code, or if not specified default to the vscode setting.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/WindowsCredentialApi.java
```diff
@@ -0,0 +1,294 @@
...
+     *
+     * typedef struct _CREDENTIAL_ATTRIBUTE {
+     *   LPTSTR Keyword;
+     *   DWORD  Flags;
+     *   DWORD  ValueSize;
+     *   LPBYTE Value;
+     * } CREDENTIAL_ATTRIBUTE, *PCREDENTIAL_ATTRIBUTE;
+     *
+     */
+    class CREDENTIAL_ATTRIBUTE extends Structure {
```
srnagar, line 39: Why are classnames uppercase?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/WindowsCredentialApi.java
```diff
@@ -0,0 +1,294 @@
...
+     *   DWORD                 CredentialBlobSize;
+     *   LPBYTE                CredentialBlob;
+     *   DWORD                 Persist;
+     *   DWORD                 AttributeCount;
+     *   PCREDENTIAL_ATTRIBUTE Attributes;
+     *   LPTSTR                TargetAlias;
+     *   LPTSTR                UserName;
+     * } CREDENTIAL, *PCREDENTIAL;
+     */
+    class CREDENTIAL extends Structure {
```
srnagar, line 98: Same here - class names should use Pascal case.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/WindowsCredentialApi.java
```diff
@@ -0,0 +1,294 @@
...
+            read();
+        }
+
+        /**
+         *   A bit member that identifies characteristics of the credential. Undefined bits should be initialized
+         *   as zero and not otherwise altered to permit future enhancement.
+         *
+         *   See MSDN doc for all possible flags
+         */
+        public int Flags;
```
srnagar, line 135: variable names should start from lower case.

sdk/identity/azure-identity/src/main/java/com/azure/identity/VisualStudioCodeCredentialBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private String tenantId;
+
+    /**
+     * Sets the tenant ID of the application.
+     *
+     * @param tenantId the tenant ID of the application.
+     * @return An updated instance of this builder with the tenant id set as specified.
+     */
+    @SuppressWarnings("unchecked")
```
srnagar, line 21: I guess this whole builder class will be removed as suggested by @schaabs but if it's not removed then this annotation should be removed.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -429,6 +434,85 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+
+            return Mono.fromFuture(() -> application.acquireToken(
+                authorizationCodeParameters))
+                .map(ar -> new MsalToken(ar, options));
```
anuchandy, line 491: just curious, some questions:

1. if there are two ways to acquire token (firstPath:authCode, secondPath:refreshToken) then is it by design to try the second path only if we fail to parse the local configuration? 
2. If it is not always guaranteed to have authCode in the config, does it makes sense to try the second path? 
3. Is it make sense to try out the second path if there is an async error in the first path?

```java
Mono<MsalToken> mono1 = tokenByAuthCode();
Mono<MsalToken> mono2 = tokenByRefreshToken();
mono1.onErrorResume(t -> mono2);
```

sdk/identity/azure-identity/src/main/java/com/azure/identity/VisualStudioCodeCredential.java
g2vinay: Yeah, made it private.

sdk/identity/azure-identity/src/main/java/com/azure/identity/VisualStudioCodeCredentialBuilder.java
g2vinay: removed.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: Yep, added the logic to support multiple clouds in both VScode + IntelliJ creds.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: removed

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: WE don't need to support auth code, that logic is removed now.

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredentialBuilder.java
```diff
@@ -20,9 +20,24 @@
     private boolean excludeManagedIdentityCredential;
     private boolean excludeSharedTokenCacheCredential;
     private boolean excludeAzureCliCredential;
+    private boolean excludeVsCodeCredential;
+    private String tenantId;
     private final ClientLogger logger = new ClientLogger(DefaultAzureCredentialBuilder.class);
 
 
+    /**
+     * Sets the tenant ID of the application.
```
JonathanGiles, line 29: Need more text here.

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredentialBuilder.java
```diff
@@ -20,9 +20,24 @@
...
+     *
+     * @param tenantId the tenant ID of the application.
+     * @return An updated instance of this builder with the tenant id set as specified.
+     */
+    @SuppressWarnings("unchecked")
```
JonathanGiles, line 34: Why suppress warnings?

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredentialBuilder.java
```diff
@@ -79,6 +94,16 @@ public DefaultAzureCredentialBuilder excludeAzureCliCredential() {
         return this;
     }
 
+    /**
+     * Excludes the {@link VisualStudioCodeCredential} from the {@link DefaultAzureCredential} authentication flow.
+     *
+     * @return An updated instance of this builder with the Visual Studio Code credential exclusion set as specified.
+     */
+    public DefaultAzureCredentialBuilder excludeVisualStudioCodeCredential() {
```
JonathanGiles, line 102: Should this just be 'excludeVSCodeCredential()`?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/VisualStudioCacheAccessor.java
```diff
@@ -0,0 +1,128 @@
...
+            if (Platform.isWindows()) {
+                settingsPath = Paths.get(System.getenv("APPDATA"), "Code", "User", "settings.json")
+                        .toString();
+            } else if (Platform.isMac()) {
+                settingsPath = Paths.get(homeDir, "Library",
+                        "Application Support", "Code", "User", "settings.json").toString();
+            } else if (Platform.isLinux()) {
+                settingsPath = Paths.get(homeDir, ".config", "Code", "User", "settings.json")
+                        .toString();
+            }
```
JonathanGiles, line 48: Handle the situation better where a different OS is being used.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/VisualStudioCacheAccessor.java
```diff
@@ -0,0 +1,128 @@
...
+            LinuxKeyRingAccessor keyRingAccessor = new LinuxKeyRingAccessor(
+                    "org.freedesktop.Secret.Generic", "service",
+                    serviceName, "account", accountName);
+
+            byte[] readCreds = keyRingAccessor.read();
+            credential = new String(readCreds, StandardCharsets.UTF_8);
+        } else {
+            throw logger.logExceptionAsError(
+                new CredentialUnavailableException("Platform could not be determined for VsCode"
+                                                           + " Credential authentication."));
```
JonathanGiles, line 87:
```suggestion
                                                           + " credential authentication."));
```

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/VisualStudioCacheAccessor.java
```diff
@@ -0,0 +1,128 @@
...
+
+            LinuxKeyRingAccessor keyRingAccessor = new LinuxKeyRingAccessor(
+                    "org.freedesktop.Secret.Generic", "service",
+                    serviceName, "account", accountName);
+
+            byte[] readCreds = keyRingAccessor.read();
+            credential = new String(readCreds, StandardCharsets.UTF_8);
+        } else {
+            throw logger.logExceptionAsError(
+                new CredentialUnavailableException("Platform could not be determined for VsCode"
```
JonathanGiles, line 86:
```suggestion
                new CredentialUnavailableException("Platform could not be determined for VS Code"
```

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/WindowsCredentialApi.java
g2vinay: maps to native C structs fields.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/WindowsCredentialApi.java
g2vinay: Maps to native C structs.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/WindowsCredentialApi.java
g2vinay: These classes map to native C code via Jna.
This is a general best practice when using Native classes to distinguish them.

sdk/identity/azure-identity/src/main/java/com/azure/identity/VisualStudioCodeCredential.java
```diff
@@ -0,0 +1,56 @@
...
+                             .build();
+        this.cachedToken = new AtomicReference<>();
+    }
+
+    @Override
+    public Mono<AccessToken> getToken(TokenRequestContext request) {
+        return Mono.defer(() -> {
+            if (cachedToken.get() != null) {
+                return identityClient.authenticateWithUserRefreshToken(request, cachedToken.get())
+                           .onErrorResume(t -> Mono.empty());
```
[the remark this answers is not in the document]
g2vinay: This will be possibly covered, when adding Identity wide logging.

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredentialBuilder.java
g2vinay: This is removed.

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredentialBuilder.java
g2vinay: Yeah that's more identifiable as a user.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/VisualStudioCacheAccessor.java
g2vinay: handled.

sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredentialBuilder.java
g2vinay: updated

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: /organizations/ + tenantid 
works well.
tenantId is either user passed or "common"
Should I follow Either strategy ? to align with other languages ?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -429,6 +431,62 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+            if (userSettings.has("azure.cloud")) {
+                cloud = userSettings.get("azure.cloud").asText();
+            }
+        }
+
+        String credential = accessor.getCredentials("VS Code Azure", cloud);
+
+        String authority = accessor.getAzureAuthHost(cloud).replaceAll("/+$", "")
+            + "/organizations/" + tenant;
+        PublicClientApplication.Builder publicClientApplicationBuilder = PublicClientApplication.builder(clientId);
```
schaabs, line 466: If you create a public client application every time we request a token there won't be any caching and each call to this method will result in a call to AAD.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -429,6 +431,62 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+
+            if (userSettings.has("azure.cloud")) {
+                cloud = userSettings.get("azure.cloud").asText();
+            }
+        }
+
+        String credential = accessor.getCredentials("VS Code Azure", cloud);
+
+        String authority = accessor.getAzureAuthHost(cloud).replaceAll("/+$", "")
+            + "/organizations/" + tenant;
```
schaabs, line 465: The value in the vscode settings azure.cloud azure.tenant need to be overridden if the user specified a different AuthorityHost or TenantId when constructing the credential. This doesn't appear to allow for that as it seems to prefer the value in the settings file.

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -257,6 +257,9 @@
   <suppress checks="com.azure.tools.checkstyle.checks.GoodLoggingCheck"
             files="com.azure.endtoend.identity.WebJobsIdentityTest.java"/>
 
+  <suppress checks="MethodName|MemberName|VisibilityModifier|TypeName"
```
JonathanGiles, line 260: Comment on why this suppression is necessary please.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
```diff
@@ -429,6 +431,62 @@ private HttpPipeline setupPipeline(HttpClient httpClient) {
...
+
+            RefreshTokenParameters parameters = RefreshTokenParameters
+                                                    .builder(new HashSet<>(request.getScopes()), credential)
+                                                    .build();
+
+            return Mono.defer(() -> Mono.fromFuture(clientApplication.acquireToken(parameters))
+                                        .map(ar -> new MsalToken(ar, options)));
+
+        } catch (MalformedURLException e) {
+            return Mono.error(logger.logExceptionAsError(new IllegalStateException(e)));
```
JonathanGiles, line 486: Why did you choose to wrap the MalformedURLException in an IllegalStateException?

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: Updated

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: We may need to add this logic in .NET too:  (I maybe wrong, can you double check in .NET below)
https://github.com/Azure/azure-sdk-for-net/blob/a2e2de0ac6aa695182aeda7c44964d6c05935ed6/sdk/identity/Azure.Identity/src/VisualStudioCodeCredential.cs#L117

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
g2vinay: updated

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java
g2vinay: Added the logic to give priority to user-specified tenant id.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/VisualStudioCacheAccessor.java
```diff
@@ -0,0 +1,185 @@
...
+
+    private boolean isRefreshTokenString(String str) {
+        for (int i = 0; i < str.length(); i++) {
+            char ch = str.charAt(i);
+            if ((ch < '0' || ch > '9') && (ch < 'A' || ch > 'Z') && (ch < 'a' || ch > 'z')
+                        && ch != '_' && ch != '-' && ch != '.') {
+                return false;
+            }
+        }
+        return true;
```
srnagar, line 161: This code can be replaced by `return str.matches("^[-_.a-zA-Z0-9]+$")`

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/VisualStudioCacheAccessor.java
```diff
@@ -0,0 +1,185 @@
...
+ */
+public class VisualStudioCacheAccessor {
+    private static final String PLATFORM_NOT_SUPPORTED_ERROR = "Platform could not be determined for VS Code"
+                                               + " credential authentication.";
+    private final ClientLogger logger = new ClientLogger(VisualStudioCacheAccessor.class);
+
+    /**
+     * Creates an instance of {@link VisualStudioCacheAccessor}
+     */
+    public VisualStudioCacheAccessor() { }
```
srnagar, line 32: This ctor is not necessary.

sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/VisualStudioCacheAccessor.java
g2vinay: String.matches() way of matching regex to String has a good perf cost.
String matches forces the input regex to always first get compiled via a Pattern.compile() and doing it on every call comes with a perf cost.
There is an alternative to String.matches and retain the perf, updated with that logic.

PR #10377, opened 2020-04-20
Implemented a push based Avro parser

gapra-msft:


sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/AvroSchema.java
```diff
@@ -0,0 +1,153 @@
...
+     * Gets the schema associated with the type.
+     *
+     * @param type The {@link AvroType type} that defines the schema.
+     * @param state {@link AvroParserState}
+     * @param onResult {@link Consumer}
+     * @return {@link AvroSchema}
+     *
+     * @see AvroType
+     */
+    public static AvroSchema getSchema(AvroType type, AvroParserState state, Consumer onResult) {
```
gapra-msft, line 106: Can I just add @SuppressWarnings("unchecked") here?
The build is in general super mad at me for casting unchecked

gapra-msft: Talked to Rick about it. I'm going to make AvroSchema ungeneric since it doesnt really fit the semantics and make the Consumers all Consumers of Objects and cast accordingly as I need Objects in Consumers.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParser.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Parses as many objects as possible in the buffer.
+     *
+     * Caches buffer remaining as part of state in case the entire buffer is not used.
+     * @param buffer {@link ByteBuffer} that is part of an Avro file.
+     * @return A reactive stream of Objects found in this
+     */
+    public Flux<Object> parse(ByteBuffer buffer) {
```
jaschrep-msft, line 82: Why are we returning a flux here? Or anything for that matter? My understanding is we are pushing results through `schema.publish()`.

gapra-msft: schema.publish will just call the result handler for the specific schema on top of the stack (in our case, it will add it to the global list of objects in this class)

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParserState.java
```diff
@@ -0,0 +1,130 @@
...
+        this.stack.pop();
+    }
+
+    /**
+     * Whether or not the state is ready to emit sizeRequired bytes.
+     *
+     * @param sizeRequired The number of bytes required.
+     * @return Whether or not the state is ready to emit sizeRequired bytes
+     */
+    public boolean contains(Long sizeRequired) {
```
kasobol-msft, line 77: could we use `long` instead of `Long`?

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroConstants.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.Set;
+
+/**
+ * Constants for Avro.
+ */
+public class AvroConstants {
+
+    public static final int SYNC_MARKER_SIZE = 16;
+    public static final byte[] MAGIC_BYTES = { (byte) 'O', (byte) 'b', (byte) 'j', (byte) 1 };
```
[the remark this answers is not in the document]
gapra-msft: It's 00000001

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParserState.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Consumes bytes from the state's internal cache of buffers.
+     * Meant for use by AvroSchema objects (specifically ones that represent primitive types, since complex types
+     * are just a combination of primitive types)
+     *
+     * @param size The number of bytes to consume.
+     * @return A List of ByteBuffers with the number of bytes requested.
+     */
+    public List<ByteBuffer> consume(Long size) {
```
kasobol-msft, line 89: small `long`?

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/util/AvroUtils.java
```diff
@@ -0,0 +1,32 @@
...
+                bytesArr.add(i.get());
+            }
+        });
+
+        byte[] ret = new byte[bytesArr.size()];
+        for (int i = 0; i < bytesArr.size(); i++) {
+            ret [i] = bytesArr.get(i);
+        }
+
+        return ret;
```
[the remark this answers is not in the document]
gapra-msft: Good idea. I'll change that

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/complex/AvroArraySchema.java
```diff
@@ -0,0 +1,138 @@
...
+    @Override
+    public void progress() {
+        /* Progress is defined by progress on the sub-type schemas. */
+    }
+
+    @Override
+    public boolean canProgress() {
+        /* Can always make progress since it is defined by the progress on the sub-type schemas. */
+        return true;
+    }
```
kasobol-msft, line 137: I think these are effectively not needed for complex types (like we discussed the other day). However, I'd rather not leave them no-op like that - in case we have bug in top level logic this could lead to infinite loop.

I'd rather make them throw OR remove from this interface. Consider such options:
- create PrivmitiveSchema interface where these are defined and not define them for complex types at all (this would require casting/instanceof when popping stack)
- create ComplexSchema abstract class and define them there in one place. Also make them throw some not supported exception so that they break loop in case there's some bug elsewhere.. (not ideal contract).

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/primitive/AvroLongSchema.java
```diff
@@ -0,0 +1,92 @@
...
+        super(state, onResult);
+    }
+
+    @Override
+    public void add() {
+        this.state.push(this);
+    }
+
+    /**
+     * Consuming bytes written into AvroParserState
```
kasobol-msft, line 48: Might we worth describing little bit more about "variable lenght zig-zag encoding" per Avro official doc (or reference there from here)

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/primitive/AvroNullSchema.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    @Override
+    public boolean canProgress() {
+        return true;
+    }
+
+    /* We use a custom type to return null since null cannot be emitted in a Flux.
+       Users of the AvroParser must transform all NullSchema.Null objects to null if necessary. */
+    public static class Null {
+    }
```
kasobol-msft, line 46: So we expect customer to see this type. For change feed that probably doesn't matter. but for QQ customer's are going to see that, so this should go to some model (or there should be mapping from this to some model when it comes to QQ implementation). What do you think?

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/util/AvroUtils.java
kasobol-msft, line 30: This should throw some meaningful exception if one array is not enough.

sdk/storage/azure-storage-internal-avro/src/test/java/com/azure/storage/internal/avro/implementation/AvroParserTest.groovy
```diff
@@ -0,0 +1,121 @@
...
+            .expectNextMatches(predicate)
+            .expectNextMatches(predicate)
+            .expectNextMatches(predicate)
+            .expectNextMatches(predicate)
+            .expectNextMatches(predicate)
+            .verifyComplete()
+    }
+
+    /**
+     * These are the schemas that are being tested.
```
kasobol-msft, line 41: Deja vu :-)

kasobol-msft:
This looks very nice.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParser.java
```diff
@@ -0,0 +1,112 @@
...
+import reactor.core.publisher.Flux;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A class that represents a push based AvroParser that can parse avro data from a reactive stream.
+ *
+ * The parser stores the {@link AvroParserState current state}, the sync marker (parsed from the header),
```
rickle-msft, line 16: Can you elaborate on what the sync marker and file type are either here or on their field declaration?

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParser.java
```diff
@@ -0,0 +1,112 @@
...
+ * the file type (parsed from the header metadata), and the list of records collected so far.
+ *
+ * For each ByteBuffer that is emitted from the stream, we first add it to the state's cache, while we can make
```
rickle-msft, line 19: I think this would be more clear if you take a sentence to define what the cache and the schema are and how/why we add bytes to the cache but query the schema for determining whether it's done.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParser.java
rickle-msft: To elaborate, if I understand correctly, this is sort of the same pattern as the UploadBufferPool. You essentially flatMap all the data from your source into calls to this method, and it'll return a bunch of empty fluxes as it's reading stuff in and then periodically emit one that has an actual item.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParserState.java
```diff
@@ -0,0 +1,130 @@
...
+        this.cache = new LinkedList<>();
+        this.size = 0;
+    }
+
+    /**
+     * Writes to the state's internal cache of buffers.
+     *
+     * @param buffer The buffer to add to the cache.
+     */
+    void write(ByteBuffer buffer) {
```
rickle-msft, line 40: Wondering if naming these like writeBuffer (cacheBuffer?), pushSchema(), peekSchema() would help make clear that these objects are being tracked differently? Although arguably I guess the types of the parameters/return values do that for you.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParserState.java
rickle-msft: I also usually think of contains as like a string operation. Maybe like containsSufficientBytes? or even like sizeGreaterThan

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParserState.java
rickle-msft: Is there any reason we can't use read which I think is more idiomatic?

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParserState.java
```diff
@@ -0,0 +1,130 @@
...
+     * @param size The number of bytes to consume.
+     * @return A List of ByteBuffers with the number of bytes requested.
+     */
+    public List<ByteBuffer> consume(Long size) {
+        List<ByteBuffer> result = new LinkedList<>();
+        Iterator<ByteBuffer> bufferIterator = this.cache.iterator();
+        long needed = size;
+        while (needed > 0) {
+            ByteBuffer current = bufferIterator.next();
+            /* Buffer can wholly satisfy this request. */
```
rickle-msft, line 95: I think this comment is wrong. This buffer can't wholly satisfy the request because it's less than the size we need. Alternatively you meant that this whole buffer is needed to satisfy the request?

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParser.java
gapra-msft: What Rick is saying is right, but say a ByteBuffer were to contain enough data to emit like say 5 different Objects, it'll emit them all together (after processing as many schemas as possible).

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParserState.java
gapra-msft: Yeah you're right - my bad. The whole buffer can satisfy at least part of the request

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParserState.java
gapra-msft: Yeah I can change them to be more descriptive. It may help improve readability.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/primitive/AvroNullSchema.java
gapra-msft: Since the Parser is internal, we can handle the conversion of this custom type to real null. I'll do that in both CF and QQ

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/util/AvroUtils.java
```diff
@@ -0,0 +1,119 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Utility methods for Avro.
+ */
+public class AvroUtils {
```
kasobol-msft, line 14: I'd rather break it down into more concrete bags of functionality. Otherwise it just becomes a bag do dump totally unrelated code and becomes a mess over time.

Reference: http://ralin.io/blog/oop-anti-patterns-utility-or-helper-classes.html

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/util/AvroUtils.java
```diff
@@ -0,0 +1,119 @@
...
+
+        return ret;
+    }
+
+    /**
+     * Checks whether an Object is of type Long.
+     * @param name The name of the variable
+     * @param variable The variable.
+     */
+    public static void checkLong(String name, Object variable) {
```
kasobol-msft, line 53: Instead of defining method per each type you could pass `Class type` and use `isAssignableFrom` instead of `instanceof` .

See https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#isAssignableFrom-java.lang.Class-

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/AvroType.java
```diff
@@ -0,0 +1,366 @@
...
+     */
+    public String getType() {
+        return type;
+    }
+
+    /**
+     * An avro primitive type.
+     * No additional data is required to parse a primitive.
+     */
+    public static class AvroPrimitiveType extends AvroType {
```
rickle-msft, line 57: I think nested classes is not super idiomatic and maybe we should create different independent types for each of these

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParser.java
```diff
@@ -0,0 +1,112 @@
...
+    private AvroParserState state; /* State of the Avro Parser. */
+    private byte[] syncMarker; /* Sync marker read in the header. */
+    private AvroType objectType; /* Type of objects in this Avro file. */
+    private List<Object> records; /* Holds records collected so far. */
+
+    public AvroParser() {
+        this.state = new AvroParserState();
+        this.records = new ArrayList<>();
+
+        /* TODO (gapra): Make this a record schema since it essentially does the same stuff. */
```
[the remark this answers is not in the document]
gapra-msft: I'll do this in a separate PR after I integrate CF/QQ

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/complex/AvroArraySchema.java
gapra-msft: The cleanest way to do this would be to just throw from complex types if isDone is false.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/primitive/AvroNullSchema.java
gapra-msft: This essentially means I don't expect a customer to see this

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/util/AvroUtils.java
gapra-msft: ooh that's really nice. I was wondering how I could make a helper method to do this

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroParser.java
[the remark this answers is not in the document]
gapra-msft: https://github.com/Azure/azure-sdk-for-java/issues/10463

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/util/AvroUtils.java
gapra-msft: https://github.com/Azure/azure-sdk-for-java/issues/10463

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/util/AvroUtils.java
gapra-msft: https://github.com/Azure/azure-sdk-for-java/issues/10463

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/complex/AvroArraySchema.java
gapra-msft: added to issues https://github.com/Azure/azure-sdk-for-java/issues/10463

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/primitive/AvroLongSchema.java
gapra-msft: I'll add a doc.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/complex/AvroArraySchema.java
kasobol-msft: @gapra-msft . Yet other option to consider:
In "complex" types we execute logic to listen to next part of that type in onXXX() listener methods. Maybe we should reduce their responsibility and push that part of logic into "progress" and "canProgress". So that instead of "returning" from child element we'd use "progress" to trigger listening to next subelement. Then "progress" would have meaning here. Additionally it would decrease complexity and resposibility of onXXX handlers with is a good thing to do.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/schema/AvroType.java
gapra-msft: Discussed offline. Nested classes seem ok in this use case as per Oracle docs. 
https://docs.oracle.com/javase/tutorial/java/javaOO/nested.html

PR #10382, opened 2020-04-20
Blob tags

rickle-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/BlobFlatListSegment.java
```diff
@@ -21,14 +21,14 @@
      * The blobItems property.
      */
     @JsonProperty("Blob")
-    private List<BlobItem> blobItems = new ArrayList<>();
+    private List<BlobItemInternal> blobItems = new ArrayList<>();
```
rickle-msft, line 24: Need to consider if this is a breaking change and possibly revert

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/BlobHierarchyListSegment.java
```diff
@@ -57,7 +57,7 @@ public BlobHierarchyListSegment setBlobPrefixes(List<BlobPrefix> blobPrefixes) {
      *
      * @return the blobItems value.
      */
-    public List<BlobItem> getBlobItems() {
+    public List<BlobItemInternal> getBlobItems() {
```
rickle-msft, line 60: Need to consider if this is a breaking change and possibly revert

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1226,8 +1317,9 @@ private void downloadToFileCleanup(AsynchronousFileChannel channel, String fileP
                     hd.getCopyCompletionTime(), hd.getCopyStatusDescription(), hd.isServerEncrypted(),
                     hd.isIncrementalCopy(), hd.getDestinationSnapshot(), AccessTier.fromString(hd.getAccessTier()),
                     hd.isAccessTierInferred(), ArchiveStatus.fromString(hd.getArchiveStatus()),
-                    hd.getEncryptionKeySha256(), hd.getAccessTierChangeTime(), hd.getMetadata(),
-                    hd.getBlobCommittedBlockCount(), hd.getVersionId(), hd.isCurrentVersion());
+                    hd.getEncryptionKeySha256(), hd.getEncryptionScope(), hd.getAccessTierChangeTime(),
+                        hd.getMetadata(), hd.getBlobCommittedBlockCount(), hd.getVersionId(), hd.isCurrentVersion(),
```
rickle-msft, line 1321: Formatting?

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/BlobClientJavaDocCodeSnippets.java
```diff
@@ -429,4 +429,34 @@ public void uploadFromFile2() throws IOException {
...
+            .setContentMd5("data".getBytes(StandardCharsets.UTF_8))
+            .setContentLanguage("en-US")
+            .setContentType("binary");
+
+        Map<String, String> metadata = Collections.singletonMap("metadata", "value");
+        Map<String, String> tags = Collections.singletonMap("tag", "value");
+        BlobRequestConditions requestConditions = new BlobRequestConditions()
+            .setLeaseId(leaseId)
+            .setIfUnmodifiedSince(OffsetDateTime.now().minusDays(3));
+        Integer blockSize = 100 * 1024 * 1024; // 100 MB;
```
kasobol-msft, line 450: Please use long as block size and don't use deprecated constructor below.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
```diff
@@ -270,8 +295,34 @@ public void uploadFromFile(String filePath, boolean overwrite) {
     public void uploadFromFile(String filePath, ParallelTransferOptions parallelTransferOptions,
         BlobHttpHeaders headers, Map<String, String> metadata, AccessTier tier, BlobRequestConditions requestConditions,
         Duration timeout) {
+
```
gapra-msft, line 298: Is this method empty now?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -329,6 +331,72 @@ public String getAccountUrl() {
...
     }
 
+
+    /**
+     * Returns a reactive Publisher emitting the blobs in this account whose tags match the query expression. For more
+     * information, including information on the query syntax, see the <a href="DOC LINK NEEDED">Azure Docs</a>.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.BlobServiceAsyncClient.filterBlobs#String}
```
gapra-msft, line 341: rename code snippet

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -329,6 +331,72 @@ public String getAccountUrl() {
...
+     */
+    public PagedFlux<FilterBlobItem> findBlobsByTags(String query, Integer maxResultsPerPage) {
+        try {
+            return findBlobsByTags(query, maxResultsPerPage, null);
+        } catch (RuntimeException ex) {
+            return pagedFluxError(logger, ex);
+        }
+    }
+
+    // TODO: options instead of maxResults?
```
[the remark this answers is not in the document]
gapra-msft: Yeah I like the idea of making it an Options since that matches what we do with listing in general

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -329,6 +331,72 @@ public String getAccountUrl() {
...
+    // TODO: Return type?
```
gapra-msft, line 371: could we return like BlobItem? What are the properties we want to expose on this?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobItem.java
```diff
@@ -15,6 +18,35 @@
...
+        this.name = blobItemInternal.getName();
+        this.deleted = blobItemInternal.isDeleted();
+        this.snapshot = blobItemInternal.getSnapshot();
+        this.properties = blobItemInternal.getProperties();
+        this.metadata = blobItemInternal.getMetadata();
+        this.versionId = blobItemInternal.getVersionId();
+        this.isCurrentVersion = blobItemInternal.isCurrentVersion();
+        this.isPrefix = blobItemInternal.isPrefix();
+
+        this.tags = new HashMap<>();
```
gapra-msft, line 42: How do the tags get populated in swagger? Is it similar to metadata? Or completely different. Why do we need to do this set thing?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobItemInternal.java
```diff
@@ -0,0 +1,251 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+
+package com.azure.storage.blob.models;
```
gapra-msft, line 5: If this is internal only, it can go in implementation.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobItem.java
```diff
@@ -124,82 +154,91 @@ public BlobItem setSnapshot(String snapshot) {
...
+
+    /**
+     * Set the versionId property: The versionId property.
+     *
+     * @param versionId the versionId value to set.
      * @return the BlobItem object itself.
      */
-    public BlobItem setProperties(BlobItemProperties properties) {
-        this.properties = properties;
+    public BlobItem setVersionId(String versionId) {
```
gapra-msft, line 220: Do we want to consider deprecating the setters if we no longer want people to use them?

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/specialized/BlobAsyncClientBaseJavaDocCodeSnippets.java
```diff
@@ -198,6 +198,25 @@ public void setMetadataCodeSnippet() {
         // END: com.azure.storage.blob.specialized.BlobAsyncClientBase.setMetadata#Map
     }
 
+    /**
+     * Code snippets for {@link BlobAsyncClientBase#getTags()}
```
gapra-msft, line 202: do we need a snippet for the max overloads too?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1348,6 +1440,108 @@ private void downloadToFileCleanup(AsynchronousFileChannel channel, String fileP
...
+        List<BlobTag> tagList = null;
+        if (tags != null) {
+            tagList = new ArrayList<>();
+            for (String tag : tags.keySet()) {
+                tagList.add(new BlobTag().setKey(tag).setValue(tags.get(tag)));
+            }
+        }
+        BlobTags t = new BlobTags().setBlobTagSet(tagList);
+        return this.azureBlobStorage.blobs().setTagsWithRestResponseAsync(null, null, null, snapshot,
+            null /* versionId */, null, null, null, t, context)
```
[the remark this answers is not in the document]
rickle-msft: The spec says you can set tags on a version. It's unclear if it bumps the version number. It is, however, not allowed on a snapshot.

Do you think we should throw if the client has a snapshot value so that customers don't think we're setting tags on their snapshot when we're really just ignoring it?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
rickle-msft: Yikes good catch.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
rickle-msft: Unfortunately the results here only come back with container name, blob name, and tag key/value. Because BlobItem has so many properties, it seemed weird to return it when most of them would be null, also there's no place for the tag key/value to go. But it's also kinda lame that we have a whole different list result type for this. 

I don't really think there's a good alternative, but I left it open in case other people could think of one and because I can't

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobItem.java
rickle-msft: Metadata looks like this:
<Metadata>     
        <Name>value</Name>  
</Metadata>  

Tags looks like this:
<Tags> 
    <TagSet> 
        <Tag> 
            <Key>key1</Key> 
            <Value>val1</Value> 
        </Tag> 
        <Tag> 
            <Key>key2</Key> 
            <Value>val2</Value> 
        </Tag> 
    </TagSet> 
</Tags> 
 So unfortunately there's a lot more hierarchy in the tags results, which gets generates as this gross object model that you need to call like Tags.getTagSet.getTag(i).getValue as opposed to being able to index into a map.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/BlobFlatListSegment.java
rickle-msft: It is not because we process this type internally and convert the elements into a PagedFlux. Even byPage on the paged flux doesn't use this.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/BlobHierarchyListSegment.java
rickle-msft: See comment on FlatListSegment.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
```diff
@@ -241,11 +241,55 @@
...
+     * @param headers {@link BlobHttpHeaders}
+     * @param metadata Metadata to associate with the blob.
+     * @param tags Tags to associate with the destination blob.
+     * @param tier {@link AccessTier} for the destination blob.
+     * @param requestConditions {@link BlobRequestConditions}
+     * @return A reactive response containing the information of the uploaded block blob.
+     */
+    public Mono<Response<BlockBlobItem>> uploadWithResponse(Flux<ByteBuffer> data,
+        ParallelTransferOptions parallelTransferOptions, BlobHttpHeaders headers, Map<String, String> metadata,
+        Map<String, String> tags, AccessTier tier, BlobRequestConditions requestConditions) {
```
alzimmermsft, line 286: Would we want to reorder the arguments so `tags` is at the end to promote an easier upgrade path for current SDK users?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -329,6 +331,72 @@ public String getAccountUrl() {
...
+    PagedFlux<FilterBlobItem> findBlobsByTags(String query, Integer maxResults, Duration timeout) {
+        throwOnAnonymousAccess();
+        Function<String, Mono<PagedResponse<FilterBlobItem>>> func =
+            marker -> findBlobsByTags(query, marker, maxResults, timeout)
+                .map(response -> {
+                    List<FilterBlobItem> value = response.getValue().getBlobs() == null
+                        ? new ArrayList<>(0)
```
alzimmermsft, line 378: Collection.singletonList?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -329,6 +331,72 @@ public String getAccountUrl() {
...
+     * information, including information on the query syntax, see the <a href="DOC LINK NEEDED">Azure Docs</a>.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.BlobServiceAsyncClient.filterBlobs#String}
+     *
+     * @param query Filters the results to return only blobs whose tags match the specified expression.
+     * @return A reactive response emitting the list of blobs.
+     */
+    public PagedFlux<FilterBlobItem> findBlobsByTags(String query) {
```
alzimmermsft, line 346: Thoughts on adding some convenience method to build a query?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClient.java
```diff
@@ -193,6 +194,40 @@ public String getAccountUrl() {
...
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.BlobServiceClient.filterBlobs#String-Integer-Duration}
+     *
+     * @param query Filters the results to return only blobs whose tags match the specified expression.
+     * @param maxResultsPerPage The maximum number of results to return in a given page.
+     * @param timeout An optional timeout value beyond which a {@link RuntimeException} will be raised.
+     * @return The list of blobs.
+     */
+    public PagedIterable<FilterBlobItem> findBlobsByTags(String query, Integer maxResultsPerPage, Duration timeout) {
```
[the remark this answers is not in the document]
rickle-msft: Waiting on confirmation from Jonathan for this one.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1348,6 +1440,108 @@ private void downloadToFileCleanup(AsynchronousFileChannel channel, String fileP
...
+        try {
+            return withContext(this::getTagsWithResponse);
+        } catch (RuntimeException ex) {
+            return monoError(logger, ex);
+        }
+    }
+
+    Mono<Response<Map<String, String>>> getTagsWithResponse(Context context) {
+        return this.azureBlobStorage.blobs().getTagsWithRestResponseAsync(null, null, null, null, snapshot,
+            null /* versionId */, context)
```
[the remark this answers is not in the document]
rickle-msft: Container name, blob name, request id, timeout. I'm pretty sure we set those to null on all calls to the protocol layer

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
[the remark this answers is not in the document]
kasobol-msft: then we should be passing versionId here (should be available by now).

I think we can let service do the validation for snapshot (in case they change their mind).

sdk/storage/azure-storage-blob/src/samples/java/com/azure/storage/blob/specialized/BlobAsyncClientBaseJavaDocCodeSnippets.java
rickle-msft: There is one on line 463. This file is just structured such that all the min overloads are up top

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
rickle-msft: I get that, but I also think I don't agree with it. I think it's cleaner to have things sort of logically/consistently grouped, and requestConditions are always at the end. And tags and metadata are conceptually similar. 

I can be persuaded to the contrary if other folks agree we should always add new parameters to the end.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
rickle-msft: I think this is a part of an ongoing email thread about what sort of object model we want to offer, if any, to support building queries.

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
gapra-msft: I think its fine to change groupings since we didnt deprecate the old method. If we forced everyone to bump to this new overload, then I could see a case for trying to make it easier to upgrade

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceClient.java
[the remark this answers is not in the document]
rickle-msft: Done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/CustomHierarchicalListingDeserializer.java
```diff
@@ -17,11 +16,11 @@
 
     @Override
     public BlobHierarchyListSegment deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
-        ArrayList<BlobItem> blobItems = new ArrayList<>();
+        ArrayList<BlobItemInternal> blobItems = new ArrayList<>();
```
kasobol-msft, line 19: Why this rename is required?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobOutputStream.java
```diff
@@ -186,9 +204,9 @@ void commit() {
 
         private BlockBlobOutputStream(final BlobAsyncClient client,
             final ParallelTransferOptions parallelTransferOptions, final BlobHttpHeaders headers,
-            final Map<String, String> metadata, final AccessTier tier, final BlobRequestConditions requestConditions,
-            Context context) {
-            super(Integer.MAX_VALUE); // writeThreshold is effectively not used by BlockBlobOutputStream.
```
kasobol-msft, line 191: we should preserve this comment.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/CustomHierarchicalListingDeserializer.java
rickle-msft: Because the BlobItem type produced by the generator has a really bad interface for interacting with tags resulting from how it's structured in the payload/swagger. We want to get rid of that and just return a map, which required renaming the generated type to be internal

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobClient.java
```diff
@@ -270,8 +297,27 @@ public void uploadFromFile(String filePath, boolean overwrite) {
     public void uploadFromFile(String filePath, ParallelTransferOptions parallelTransferOptions,
```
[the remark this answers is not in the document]
rickle-msft: I don't think we need to deprecate. There's nothing really wrong with them unless we want to be proactive about trimming older options

PR #10383, opened 2020-04-20
[Storage][Jumbo blobs] Bump single upload max threshold. Preserve defaults.

kasobol-msft:
Max single upload threshold goes to 5000MB for blobs (and 4000MB for datalake).
Keeping previous defaults 256MB for blob and 100MB for datalake if not specified, so that customers leveraging parallel uploads for payloads larger than former defaults won't see performance degradation.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
+++ b/sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
@@ -286,12 +286,12 @@ class LargeBlobTest extends APISpec {
     private Flux<ByteBuffer> createLargeBuffer(long size, int bufferSize) {
         def bytes = getRandomByteArray(bufferSize)
         long numberOfSubBuffers = (long) (size / bufferSize)
-        int reminder = (int) (size - numberOfSubBuffers * bufferSize)
+        int remainder = (int) (size - numberOfSubBuffers * bufferSize)
         Flux<ByteBuffer> result =  Flux.just(ByteBuffer.wrap(bytes))
             .map{buffer -> buffer.duplicate()}
             .repeat(numberOfSubBuffers - 1)
-        if (reminder > 0) {
-            def extraBytes = getRandomByteArray(reminder)
+        if (remainder > 0) {
+            def extraBytes = getRandomByteArray(remainder)
             result = Flux.concat(result, Flux.just(ByteBuffer.wrap(extraBytes)))
         }
         return result
@@ -315,6 +315,10 @@ class LargeBlobTest extends APISpec {
         return file
     }
 
+    /**
+     * This class is intended for large payload test cases only and reports directly into this test class's
+     * state members.
+     */
     private class PayloadDroppingPolicy implements HttpPipelinePolicy {
         @Override
         Mono<HttpResponse> process(HttpPipelineCallContext httpPipelineCallContext, HttpPipelineNextPolicy httpPipelineNextPolicy) {
@@ -364,12 +368,12 @@ class LargeBlobTest extends APISpec {
         }
 
         private boolean isPutBlockRequest(HttpRequest request) {
-            return request.url.getQuery() != null && request.url.getQuery().endsWith("comp=block");
+            return request.url.getQuery() != null && request.url.getQuery().contains("comp=block");
         }
 
         private boolean isSinglePutBlobRequest(HttpRequest request) {
             return request.getHttpMethod().equals(HttpMethod.PUT) &&
-                request.getUrl().toString().endsWith(blobName) &&
+                request.getUrl().getPath().endsWith(blobName) &&
                 request.getUrl().getQuery() == null
         }
     }
--- a/sdk/storage/azure-storage-file-datalake/src/test/java/com/azure/storage/file/datalake/LargeFileTest.groovy
+++ b/sdk/storage/azure-storage-file-datalake/src/test/java/com/azure/storage/file/datalake/LargeFileTest.groovy
@@ -181,12 +181,12 @@ class LargeFileTest extends APISpec{
     private Flux<ByteBuffer> createLargeBuffer(long size, int bufferSize) {
         def bytes = getRandomByteArray(bufferSize)
         long numberOfSubBuffers = (long) (size / bufferSize)
-        int reminder = (int) (size - numberOfSubBuffers * bufferSize)
+        int remainder = (int) (size - numberOfSubBuffers * bufferSize)
         Flux<ByteBuffer> result =  Flux.just(ByteBuffer.wrap(bytes))
             .map{buffer -> buffer.duplicate()}
             .repeat(numberOfSubBuffers - 1)
-        if (reminder > 0) {
-            def extraBytes = getRandomByteArray(reminder)
+        if (remainder > 0) {
+            def extraBytes = getRandomByteArray(remainder)
             result = Flux.concat(result, Flux.just(ByteBuffer.wrap(extraBytes)))
         }
         return result
@@ -224,6 +224,10 @@ class LargeFileTest extends APISpec{
         return file
     }
 
+    /**
+     * This class is intended for large payload test cases only and reports directly into this test class's
+     * state members.
+     */
     private class PayloadDroppingPolicy implements HttpPipelinePolicy {
         @Override
         Mono<HttpResponse> process(HttpPipelineCallContext httpPipelineCallContext, HttpPipelineNextPolicy httpPipelineNextPolicy) {


sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
```diff
@@ -70,6 +77,84 @@ class LargeBlobTest extends APISpec {
         blockList.committedBlocks.get(0).getSizeLong() == maxBlockSize
     }
 
+    @Requires({ liveMode() })
+    @Ignore("Takes really long time")
+    // This test sends payload over the wire
+    def "Upload Real Large Blob in Single Upload"() {
+        given:
+        // TODO (kasobol-msft) Bump this to 5000MB.
```
gapra-msft, line 85: Is this TODO meant to be part of this PR?

gapra-msft: And do we need to modify the tests for datalake too? Could we add a test that somehow makes sure the default max size is adhered to?

kasobol-msft: It seems to be test account/server side issue. I'm following up on that, but I'd rather not block on this now.

kasobol-msft: I'll take a look on datalake tests.

kasobol-msft: @gapra-msft  I added test to `LargeFileTest.groovy` where default single upload for data lake is triggerd.

BTW. thank you for this suggestion. This prompted me to run full blown live tests on datalake file client and caught one nasty bug. I added fix for that in `DataLakeFileAsyncClient`. The confusing `remaining` method from `BufferAggregator` is also hidden and renamed to avoid this in the future.

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileAsyncClient.java
```diff
@@ -307,7 +307,7 @@ public String getFileName() {
         return chunkedSource.concatMap(pool::write)
             .concatWith(Flux.defer(pool::flush))
             /* Map the data to a tuple 3, of buffer, buffer length, buffer offset */
-            .map(buffer -> Tuples.of(buffer, buffer.remaining(), 0L))
+            .map(bufferAggregator -> Tuples.of(bufferAggregator, bufferAggregator.length(), 0L))
```
kasobol-msft, line 310: Should have used .lenght() everywhere in this  method after transition from ByteBuffer to BufferAggregator.
Confusing method has been hidden and renamed...

gapra-msft: Sweet great to hear we found something!

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
```diff
@@ -28,8 +32,11 @@ class LargeBlobTest extends APISpec {
     BlobClient blobClient
     BlobAsyncClient blobAsyncClient
     String blobName
-    List<Mono<Long>> putBlockPayloadSizes = Collections.synchronizedList(new ArrayList<>())
-    AtomicLong count = new AtomicLong()
+    List<Long> putBlockPayloadSizes = Collections.synchronizedList(new ArrayList<>())
+    AtomicLong blocksCount = new AtomicLong()
+    List<Long> putBlobPayloadSizes = Collections.synchronizedList(new ArrayList<>())
+    AtomicLong singleUploadCount = new AtomicLong()
+    ConcurrentHashMap<String, Boolean> retryTracker = new ConcurrentHashMap<>()
```
alzimmermsft, line 39: Should these be test class level properties or be migrated into the pipeline policy being used to test this functionality? I don't know when in the test loop, if ever, these would be reset to the default value again.

sdk/storage/azure-storage-file-datalake/src/test/java/com/azure/storage/file/datalake/LargeFileTest.groovy
```diff
@@ -148,16 +152,44 @@ class LargeFileTest extends APISpec{
...
+
     private Flux<ByteBuffer> createLargeBuffer(long size) {
         return createLargeBuffer(size, Constants.MB)
     }
 
     private Flux<ByteBuffer> createLargeBuffer(long size, int bufferSize) {
         def bytes = getRandomByteArray(bufferSize)
         long numberOfSubBuffers = (long) (size / bufferSize)
-        return Flux.just(ByteBuffer.wrap(bytes))
+        int reminder = (int) (size - numberOfSubBuffers * bufferSize)
```
alzimmermsft, line 184:
```suggestion
        int remainder = (int) (size - numberOfSubBuffers * bufferSize)
```

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
```diff
@@ -204,9 +286,15 @@ class LargeBlobTest extends APISpec {
     private Flux<ByteBuffer> createLargeBuffer(long size, int bufferSize) {
         def bytes = getRandomByteArray(bufferSize)
         long numberOfSubBuffers = (long) (size / bufferSize)
-        return Flux.just(ByteBuffer.wrap(bytes))
+        int reminder = (int) (size - numberOfSubBuffers * bufferSize)
```
alzimmermsft, line 289:
```suggestion
        int remainder = (int) (size - numberOfSubBuffers * bufferSize)
```

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
```diff
@@ -232,20 +320,57 @@ class LargeBlobTest extends APISpec {
...
+        }
+
+        private boolean isPutBlockRequest(HttpRequest request) {
+            return request.url.getQuery() != null && request.url.getQuery().endsWith("comp=block");
+        }
+
+        private boolean isSinglePutBlobRequest(HttpRequest request) {
+            return request.getHttpMethod().equals(HttpMethod.PUT) &&
+                request.getUrl().toString().endsWith(blobName) &&
+                request.getUrl().getQuery() == null
```
alzimmermsft, line 373: Would it be better to check for certain parts of the query string being null in case this gets used by a client authorizing with a SAS token.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
```diff
@@ -232,20 +320,57 @@ class LargeBlobTest extends APISpec {
...
+            return result
+        }
+
+        private boolean isPutBlockRequest(HttpRequest request) {
+            return request.url.getQuery() != null && request.url.getQuery().endsWith("comp=block");
+        }
+
+        private boolean isSinglePutBlobRequest(HttpRequest request) {
+            return request.getHttpMethod().equals(HttpMethod.PUT) &&
+                request.getUrl().toString().endsWith(blobName) &&
```
alzimmermsft, line 372: I think this should check that the path ends with the blob name.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
```diff
@@ -232,20 +320,57 @@ class LargeBlobTest extends APISpec {
...
+                        return a + byteBuffer.remaining()
+                    }
+                })
+            }
+            request.setBody("dummyBody")
+            return result
+        }
+
+        private boolean isPutBlockRequest(HttpRequest request) {
+            return request.url.getQuery() != null && request.url.getQuery().endsWith("comp=block");
```
alzimmermsft, line 367: Should this check that the query contains `comp=block`.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
kasobol-msft: Every test framework I worked with JUnit, TestNG (Java) or NUnit, XUnit, MSTest (.Net) instantiates new instance of test class for each test case. Therefore these are reset for each test.
I could potentially put them into policy but then I'd need to keep reference to policy in test case.

I'd rather keep them in test class (they represent test case state) rather than in policy (that is used to update test case state).

Reference: https://stackoverflow.com/questions/19381352/does-junit-reinitialize-the-class-with-each-test-method-invocation

sdk/storage/azure-storage-file-datalake/src/test/java/com/azure/storage/file/datalake/LargeFileTest.groovy
kasobol-msft: Ups.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
kasobol-msft: Ups.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
kasobol-msft: I agree. However, I don't expect adding SAS tokens scenarios here (orthogonal feature). This Policy is not intended to be used outside of large payload testing. Therefore, I'd rather keep it simple until we need more functionality here.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
kasobol-msft: Make sense. Thank you!

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
kasobol-msft: Make sense. Thank you!

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
alzimmermsft: Good with me, was verifying that no bad state would be reached with these in the class scope.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/LargeBlobTest.groovy
alzimmermsft: Sounds good

PR #10385, opened 2020-04-21
Adding support for completing messages on receiver link. Moving logic out of azure-core-amqp. Adding more status codes

### LINKED ISSUE #10033: Settle messages on receive link
Currently, we only settle messages through the management link. Settlement of messages should also be done by changing the disposition of the delivery.

conniey:
## Service Bus changes
* Settles (abandon, complete, defer, dead-letter) messages on the receive link rather than always going to the management node
* Various fixes in fetching a named session.
* Move ManagementLockContainer out of receive link. These are only kept updated by messages received from management link
  * Makes lock container an expiring set. (Adapted from .NET implementation.)

## azure-core-amqp changes
* Moves translation of management responses into their respective services.
  * Event Hubs translates `RequestResponse` messages differently than Service Bus does. Which makes the logic in azure-core-amqp very complex
* Cleans up the endpoint status management
* Adds flexibility for translation of AMQP error codes, instead of throwing an IllegalArgumetnException. They are an expandable set.
* Adding missing AmqpResponseCodes. The ones taken from Event Hubs are incomplete.

## Tests
* Changes in Event Hubs was tested by running live integration tests.
* Adds tests for azure-core-amqp classes like RequestResponseUtils and RequestResponseChannel.

Fixes #10033

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusMessageProcessor.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusMessageProcessor.java
@@ -433,9 +433,9 @@ class ServiceBusMessageProcessor extends FluxProcessor<ServiceBusReceivedMessage
                 return instant;
             })
             .subscribe(lockedUntil -> {
-                    logger.verbose("lockToken[{}]. lockedUntil[{}]. Lock renewal successful.",
-                        sequenceNumber, lockToken, lockedUntil);
-                },
+                logger.verbose("lockToken[{}]. lockedUntil[{}]. Lock renewal successful.",
+                    sequenceNumber, lockToken, lockedUntil);
+            },
                 error -> {
                     logger.error("Error occurred while renewing lock token.", error);
                     hasError.set(true);


sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/exception/AmqpResponseCode.java
```diff
@@ -10,13 +10,54 @@
  * Error response codes returned from AMQP.
  */
 public enum AmqpResponseCode {
-    ACCEPTED(0xca),
+    ACCEPTED(202),
```
conniey, line 13: Changed from hex to dec values for readability.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -86,19 +84,11 @@
...
+                    ManagementConstants.OPERATION_CANCEL_SCHEDULED_MESSAGE, null);
 
                 requestMessage.setBody(new AmqpValue(Collections.singletonMap(ManagementConstants.SEQUENCE_NUMBERS,
                     new Long[]{sequenceNumber})));
                 return channel.sendWithAck(requestMessage);
-            }).map(responseMessage -> {
-                int statusCode = RequestResponseUtils.getResponseStatusCode(responseMessage);
-
-                if (statusCode == AmqpResponseCode.OK.getValue()) {
-                    return Mono.empty();
```
hemanttanwar, line 98: How are we checking the error condition and throw error when service did not send OK status code ?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
```diff
@@ -417,17 +383,17 @@ private Message createDispositionMessage(UUID[] lockTokens, DispositionStatus di
...
-    private Message createManagementMessage(String operation, String linkName) {
+    private Message createManagementMessage(String operation, String associatedLinkName) {
         final Duration serverTimeout = MessageUtils.adjustServerTimeout(operationTimeout);
         final Map<String, Object> applicationProperties = new HashMap<>();
         applicationProperties.put(ManagementConstants.MANAGEMENT_OPERATION_KEY, operation);
         applicationProperties.put(ManagementConstants.SERVER_TIMEOUT, serverTimeout.toMillis());
 
-        if (linkName != null && !linkName.isEmpty()) {
-            applicationProperties.put(ManagementConstants.ASSOCIATED_LINK_NAME_KEY, linkName);
+        if (associatedLinkName != null && !associatedLinkName.isEmpty()) {
```
hemanttanwar, line 395: Should we use CoreUtils.isNullOrEmpty(associatedLinkName) here ?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ManagementChannel.java
conniey: Resolved

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/MessageLockContainer.java
```diff
@@ -3,14 +3,41 @@
...
+            }
+
+            final Instant now = Instant.now();
+            final boolean removed = lockTokens.entrySet().removeIf(entry -> {
+                final Instant expiration = entry.getValue();
+                return expiration != null && expiration.isBefore(now);
+            });
+
+            if (removed) {
+                logger.verbose("{}: Removed expired entries.", e);
```
srnagar, line 37: Maybe worth making this log `info` as removing expired entries should be uncommon and users should be able to see it at info level? Also, consider adding a count of expired entries removed in the log.

conniey: makes sense. I updated it!

PR #10391, opened 2020-04-21
mgmt fix appservice spotbugs

weidongxu-microsoft:


sdk/appservice/mgmt/src/main/java/com/azure/management/appservice/implementation/WebAppBaseImpl.java
```diff
@@ -1113,9 +1114,8 @@ private static IndexableTaskItem sequentialTask(IndexableTaskItem taskItem1, Fun
 
     WebAppBaseImpl<FluentT, FluentImplT> withNewHostNameSslBinding(
         final HostNameSslBindingImpl<FluentT, FluentImplT> hostNameSslBinding) {
-        if (hostNameSslBinding.newCertificate() != null) {
-            sslBindingsToCreate.put(hostNameSslBinding.name(), hostNameSslBinding);
-        }
+        hostNameSslBinding.newCertificate();
```
ChenTanyi, line 1117: I think the mono will never run. Does it need a `block()`?

weidongxu-microsoft: Thanks for find it.

I think it does not need to be called.
It will be not `null` for every conditional branch that get to this `attach`.
And here it is not needed.

PR #10392, opened 2020-04-21
fix: storage spotbugs

ChenTanyi:


sdk/storage/mgmt/src/main/java/com/azure/management/storage/implementation/BlobContainersImpl.java
```diff
@@ -54,7 +54,7 @@ private ImmutabilityPolicyImpl wrapImmutabilityPolicyModel(ImmutabilityPolicyInn
         return new ImmutabilityPolicyImpl(inner, manager());
     }
 
-    private Mono<ImmutabilityPolicyInner> getImmutabilityPolicyInnerUsingBlobContainersInnerAsync(String id) {
+    public Mono<ImmutabilityPolicyInner> getImmutabilityPolicyInnerUsingBlobContainersInnerAsync(String id) {
```
weidongxu-microsoft, line 57: why public?

ChenTanyi: It isn't used as a private function. If we would not make it public / protected, we need to delete it. Since the impl is not public, I think it ok to public the function in order for future usage.

PR #10400, opened 2020-04-21
Revert Surefire Version Change, Misc Configuration Cleanup

alzimmermsft:
This PR reverts the Maven surefire version rev (3.0.0-m4 -> 3.0.0-m3) due to issues introduced in the new version where some tests weren't running properly (or at all). Additionally, this PR cleans up some configurations in the parent POMs that weren't in the proper location, data library environment variable configurations in the parent POM instead of the data parent, etc.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/Configuration.java
@@ -3,8 +3,6 @@
 
 package com.azure.core.util;
 
-import java.util.Arrays;
-import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
@@ -84,12 +82,12 @@ public class Configuration implements Cloneable {
      * Name of the Azure cloud to connect to.
      */
     public static final String PROPERTY_AZURE_CLOUD = "AZURE_CLOUD";
-    
+
     /**
      * The Azure Active Directory endpoint to connect to.
-     */  
+     */
     public static final String PROPERTY_AZURE_AUTHORITY_HOST = "AZURE_AUTHORITY_HOST";
-    
+
     /**
      * Disables telemetry collection.
      */
@@ -104,18 +102,6 @@ public class Configuration implements Cloneable {
      * Disables tracing.
      */
     public static final String PROPERTY_AZURE_TRACING_DISABLED = "AZURE_TRACING_DISABLED";
-
-    /*
-     * System property loader.
-     */
-    private static final Function<String, String> PROPERTY_LOADER = System::getProperty;
-
-    /*
-     * Environment variable loader.
-     */
-    private static final Function<String, String> ENV_VAR_LOADER = System::getenv;
-
-    private static final List<Function<String, String>> LOADERS = Arrays.asList(PROPERTY_LOADER, ENV_VAR_LOADER);
 
     /*
      * Configurations that are loaded into the global configuration store when the application starts.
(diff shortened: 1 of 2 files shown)


sdk/parents/azure-data-sdk-parent/pom.xml
```diff
@@ -441,6 +443,18 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-surefire-plugin;external_dependency} -->
+        <configuration>
+          <environmentVariables>
+            <test.mode>${testMode}</test.mode>
+            <KEYVAULT_VAULTURI>https://azure-keyvault-3.vault.azure.net</KEYVAULT_VAULTURI>
+            <KEYVAULT_VAULTURI_ALT>https://azure-keyvault-2.vault.azure.net</KEYVAULT_VAULTURI_ALT>
+          </environmentVariables>
```
srnagar, line 455: The parent pom had these two additional lines. Those should be copied to this pom as well. 

```
<configurationAPIEnvironmentFound>environmentConfiguration</configurationAPIEnvironmentFound>	
<configurationAPIUseRuntimeFirst>environmentConfiguration</configurationAPIUseRuntimeFirst>
```

conniey:
> tests weren't running properly (or at all)
Is this an issue with the plugin, or our tests? Is there a corresponding issue in the surefire repository?

alzimmermsft:
> 
> 
> > tests weren't running properly (or at all)
> 
> Is this an issue with the plugin, or our tests? Is there a corresponding issue in the surefire repository?

Not completely certain on this, but reverting to the older version fixed the tests running which leads me to believe it is an issue with the plugin.

sima-zhu:
If this only applies to Storage, can we check over the groovy eclipse plugin in Storage?

conniey:
> > > tests weren't running properly (or at all)
> > 
> > 
> > Is this an issue with the plugin, or our tests? Is there a corresponding issue in the surefire repository?
> 
> Not completely certain on this, but reverting to the older version fixed the tests running which leads me to believe it is an issue with the plugin.

I think it would be worth creating an issue in SUREFIRE with some repro, so we can at least track this issue... or it may not be fixed in future releases.

https://issues.apache.org/jira/projects/SUREFIRE/issues

sdk/parents/azure-data-sdk-parent/pom.xml
alzimmermsft: These were being used in tests for Azure Core, will update those tests to no longer require these configurations.

alzimmermsft:
> 
> 
> If this only applies to Storage, can we check over the groovy eclipse plugin in Storage?

For now it'll be easier to revert back to a known working version rather than finding the magic combination of surefire and the eclipse versions that work, especially since the newer version of surefire doesn't offer any clear benefits once the test printing PR is merged in.

alzimmermsft:
> 
> 
> > > > tests weren't running properly (or at all)
> > > 
> > > 
> > > Is this an issue with the plugin, or our tests? Is there a corresponding issue in the surefire repository?
> > 
> > 
> > Not completely certain on this, but reverting to the older version fixed the tests running which leads me to believe it is an issue with the plugin.
> 
> I think it would be worth creating an issue in SUREFIRE with some repro, so we can at least track this issue... or it may not be fixed in future releases.
> 
> https://issues.apache.org/jira/projects/SUREFIRE/issues

I'll file an issue with steps to repro.

alzimmermsft:
Filed this issue against Surefire, https://issues.apache.org/jira/browse/SUREFIRE-1779.

PR #10413, opened 2020-04-22
New Constructors for AzureException, HttpRequestException, and HttpResponseException

### LINKED ISSUE #10339: [FEATURE REQ] [core]  com.azure.core.exception.AzureException should support stackless constructor for improved performance
Exceptions with stacks are slow in Java and sometimes the stack adds no value. This is true, for example, when mapping status codes to exceptions in a transport client (e.g., the Cosmos Direct TCP transport client). In these cases the value of the exception is the error that it reports. The stack does not help to understand the problem. The exception message is enough to understand the problem. Java's `Throwable` class adds such a constructor to cover this use case:
```java
protected | Throwable​(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace)
``` 

This is a request for a protected `AzureException` constructor for creating a stackless exception hierarchy based on `AzureException`. 

The constructor should have this signature:
```java
protected AzureException​(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace)
```

It should construct a new throwable with the specified detail message, cause, suppression enabled or disabled, and writable stack trace enabled or disabled.

## References
1. [The Exceptional Performance of Lil' Exception](https://shipilev.net/blog/2014/exceptional-performance/)
2. [Throwable class](https://docs.oracle.com/javase/9/docs/api/java/lang/Throwable.html)

alzimmermsft:
Fixes #10339

This PR adds a new constructors to `AzureException`, `HttpRequestException`, and `HttpResponseException` which mirror the super class constructors in `RuntimeException`. The other exception types in Azure Core weren't updated as they didn't appear to be used, if they are in the future these additional constructors can be added as a non-breaking change.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/exception/AzureException.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/exception/AzureException.java
@@ -4,7 +4,7 @@
 package com.azure.core.exception;
 
 /**
- * The base exception type for all Azure-related exceptions.
+ * The base Azure exception.
  *
  * @see HttpRequestException
  * @see ServiceResponseException
@@ -14,11 +14,27 @@ public class AzureException extends RuntimeException {
 
     /**
      * Initializes a new instance of the AzureException class.
+     */
+    public AzureException() {
+        super();
+    }
+
+    /**
+     * Initializes a new instance of the AzureException class.
      *
      * @param message The exception message.
      */
     public AzureException(final String message) {
         super(message);
+    }
+
+    /**
+     * Initializes a new instance of the AzureException class.
+     *
+     * @param cause The {@link Throwable} which caused the creation of this AzureException.
+     */
+    public AzureException(final Throwable cause) {
+        super(cause);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/exception/AzureException.java
```diff
@@ -25,9 +25,22 @@ public AzureException(final String message) {
...
+
+    /**
+     * Initializes a new instance of the AzureException class.
+     *
+     * @param message The exception message.
+     * @param cause The {@link Throwable} which caused the creation of this AzureException.
+     * @param enableSuppression Whether suppression is enabled or disabled.
+     * @param writableStackTrace Whether the exception stack trace will be filled in.
+     */
+    public AzureException(final String message, final Throwable cause, final boolean enableSuppression,
```
srnagar, line 42: Add the other 2 overloads too.
`public AzureException()`
`public AzureException(Throwable cause)`.

alzimmermsft: Should these overloads also be added into the other based exception types such as `HttpResponseException` and `HttpRequestException`?

srnagar: It's better to provide all constructor overloads and let the caller decide which ctor they want to use. Since all we are doing is delegating to the super ctor, I think it's better to include them in all exception types we create.

alzimmermsft: Added additional constructors to `AzureException`, `HttpRequestException`, and `HttpResponseException` as they were the only exception types in Azure Core which that have usage.

PR #10419, opened 2020-04-22
POM-only release pipeline

mitchdenny:
This PR adds the ability to release artifacts that are just a POM file (e.g. the packages in ```sdk/parents``` and ```sdk/boms```. At the moment the pipeline is not capable of releasing ```com.microsoft.azure:azure-arm-parent``` because of an old dependency. When that dependency is brought up this pipeline will also be able to release the arm parent (/cc @weidongxu-microsoft).

The layout of artifacts in this pipeline are different to our regular pipelines (its basically a snapshot of a locally deployed repository). We'll probably end up moving to a similar structure for all of the Java pipelines.

I wanted to get @scbedd and @JimSuplizio's input on what I've done here since they need to pass through the craggy nooks of version update land and the valley of doc publishing despair.

I've done testing up to the point of releasing a staged repository using the ```java - parents``` pipeline and also validated that the tweaks I made to the Maven publishing script didn't break the existing pipelines with the ```java - template``` pipeline.

### DIFF
--- a/sdk/parents/azure-client-sdk-parent/pom.xml
+++ b/sdk/parents/azure-client-sdk-parent/pom.xml
@@ -1470,7 +1470,7 @@
     <module>../../core/azure-core-tracing-opentelemetry</module>
     <module>../../cosmos/azure-cosmos</module>
     <module>../../cosmos/azure-cosmos-benchmark</module> <!-- CosmosSkip - Benchmark will go away, this is temporary -->
-    <module>../../cosmos/azure-cosmos-examples</module> <!-- CosmosSkip - Benchmark will go away, this is temporary -->
+    <module>../../cosmos/azure-cosmos-examples</module> <!-- CosmosSkip - Examples will go away, this is temporary -->
     <module>../../e2e</module>
     <module>../../eventhubs/azure-messaging-eventhubs</module>
     <module>../../eventhubs/azure-messaging-eventhubs-checkpointstore-blob</module>


eng/scripts/Publish-MavenPackages.ps1
```diff
@@ -261,7 +266,7 @@ foreach ($packageDetail in $packageDetails) {
 
   if ($RepositoryUrl -like "https://pkgs.dev.azure.com/azure-sdk/public/*") {
     Write-Information "GPG Signing and deploying package in one step to: $RepositoryUrl"
-    mvn gpg:sign-and-deploy-file "--batch-mode" "$pomOption" "$fileOption" "$javadocOption" "$sourcesOption" "$filesOption" $classifiersOption "$typesOption" "-Durl=$RepositoryUrl" "$gpgexeOption" "-DrepositoryId=target-repo" "-Drepo.password=$RepositoryPassword" "--settings=$PSScriptRoot\..\maven.publish.settings.xml"
+    # mvn gpg:sign-and-deploy-file "--batch-mode" "$pomOption" "$fileOption" "$javadocOption" "$sourcesOption" "$filesOption" $classifiersOption "$typesOption" "-Durl=$RepositoryUrl" "$gpgexeOption" "-DrepositoryId=target-repo" "-Drepo.password=$RepositoryPassword" "--settings=$PSScriptRoot\..\maven.publish.settings.xml"
```
JimSuplizio, line 269: Should this be commented out or removed?

sdk/parents/azure-client-sdk-parent/pom.xml
```diff
@@ -1468,8 +1468,8 @@
     <module>../../core/azure-core-test</module>
     <module>../../core/azure-core-tracing-opentelemetry</module>
     <module>../../cosmos/azure-cosmos</module>
-    <module>../../cosmos/azure-cosmos-benchmark</module> <!-- CosmosSkip - Benchmark will go away, this is temporary -->
-    <module>../../cosmos/azure-cosmos-examples</module> <!-- CosmosSkip - Examples will go away, this is temporary -->
+    <module>../../cosmos/azure-cosmos-benchmark</module>
+    <module>../../cosmos/azure-cosmos-examples</module>
```
JimSuplizio, line 1472: @mitchdenny please do not remove these comments from the pom file. These are the responsibility of the Cosmos team to remove before GA and I want them kept here.

sdk/parents/ci.yml
```diff
@@ -31,11 +31,18 @@ pr:
...
-
+  
 stages:
-  - stage: Placeholder
-    jobs:
-      - job: Placeholder
-        steps:
-          - script: |
-              echo Hello, world!
+  - template: ../../eng/pipelines/templates/stages/archetype-sdk-pom-only.yml
```
JimSuplizio, line 36: I'm under the impression that this is mainly for management track right now so why isn't that artifact listed here? In theory, we're done releasing client and data parent poms and have been for months.

eng/scripts/Publish-MavenPackages.ps1
mitchdenny: A yes .. that was a safety mechanism I put in whilst testing, will take that out. Thanks for spotting that I thought I got them all.

sdk/parents/ci.yml
mitchdenny: There is a challenge with management plane right now, specifically it is pointing to an old parent POM which has a distribution management section. As part of this PR I remove the distribution management section from ```azure-sdk-parent``` since it gets in the road (and is kinda pointless anyway in our context).

I think there are some changes coming that will result in the parent POMs being updated (I think @alzimmermsft  was working on something recently).

mitchdenny: ... oh and this same pipeline will also be used by BOM deployments.

sdk/parents/azure-client-sdk-parent/pom.xml
mitchdenny: Put them back in, although for the most part the module list in the parent POMs can be completely deleted (unless we have a script somewhere that is still depending on them). They aren't considered at all for the build itself anymore. Only the ```sdk/cosmos/pom.xml``` defines the modules that are used in the Cosmos build.

mitchdenny:
This should turn green after this PR goes in: #10582

mitchdenny:
@JimSuplizio when you get a chance could you pass a critical eye over where this PR is now. I'd like to get it merged soon so we've got a unified pipeline for parent POMs and BOMs.

sdk/parents/azure-client-sdk-parent/pom.xml
```diff
@@ -1470,7 +1470,7 @@
     <module>../../core/azure-core-tracing-opentelemetry</module>
     <module>../../cosmos/azure-cosmos</module>
     <module>../../cosmos/azure-cosmos-benchmark</module> <!-- CosmosSkip - Benchmark will go away, this is temporary -->
-    <module>../../cosmos/azure-cosmos-examples</module> <!-- CosmosSkip - Examples will go away, this is temporary -->
+    <module>../../cosmos/azure-cosmos-examples</module> <!-- CosmosSkip - Benchmark will go away, this is temporary -->
```
JimSuplizio, line 1473: nit: This change is neither correct nor necessary, please undo the changes to this file and remove from this PR.

sdk/parents/ci.yml
JimSuplizio: To be 100% clear: the client and data parent poms should really not need to be released ever again. The only parent pom that we know we're guaranteed to have to release at some point and time in the future is azure-arm-parent which management's parent. That should definitely be one of the artifacts listed below.

sdk/parents/azure-client-sdk-parent/pom.xml
mitchdenny: Ah cut and paste error. I had originally removed those module lines altogether but put them back in at your request but didn't adjust the comment.

mitchdenny:
@JimSuplizio thanks for the review. I'm less optimistic about never having to release the parent POMs again :) Regardless its an artifact that we have had to release in the past so I wanted to make sure we had it covered in our unified pipelines. We will definitely want to release BOMs and they use the same pipeline archetype.

I cannot add the ```azure-arm-parent``` to the list of artifacts until they address this (from the POM file):

```xml
  <parent>
    <groupId>com.azure</groupId>
    <version>1.5.0</version> <!-- {x-version-exempt;com.azure:azure-sdk-parent;all of management needs to be updated in order for this to be at the latest} -->
    <relativePath>../azure-sdk-parent/pom.xml</relativePath>
  </parent>```

If they can't go to the latest local parent they can't pick up the config changes that were necessary there.

Issue #10408, opened 2020-04-21
[FR] make sure  all async API has retuen error but not throw error.


mssfang:
Make sure all async API has retuen error but not throw error.
As per reactive guidelines, you should instead catch this exception in the public API method and return an error. See PagedFlux APIs for reference. You will have to do something similar for PollerFlux too.

For example:
https://github.com/Azure/azure-sdk-for-java/pull/9953/files/14a9a779ce0e22b7ccb5cc1fae52c127ac4c7236#r412509995

Issue #9716, opened 2020-03-31
Add utility method to convert incoming stream to Flux<ByteBuffer> 


samvaity:
Add utility methods into Azure Core to support the conversion of the incoming input stream to `Flux<ByteBuffer>`

anuchandy:
@samvaity in case it help, I wrote one something similar for okhttp https://github.com/Azure/azure-sdk-for-java/blob/04f5f53a5b1b5b26b3a8b521b15d25b6403a1bfe/sdk/core/azure-core-http-okhttp/src/main/java/com/azure/core/http/okhttp/OkHttpAsyncHttpClient.java#L305

You still need to take care of closing the stream, see the caller part of this method.

srnagar:
Since this is required in multiple places, I suggest we move the utility method @anuchandy mentioned to azure-core utils.

@samvaity, @mssfang and @anuchandy

anuchandy:
@alzimmermsft when we enable this in azure-core, maybe its good not updating OkHttpClient to use it. 
Though okio exposes the response stream as InputStream, it's a special stream with internal housekeeping to ensure it's methods (read, close) are not called concurrently. I remember running into a rare concurrency edge-case when canceling resulting Flux. Unfortunately, no idea how it can be addressed without incurring locks, so keeping OkHttpClient as it is may save some effort later.

PR #10420, opened 2020-04-22
[FR] Async API should always return error but not throw error

mssfang:
fixes: #10408

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
@@ -151,6 +151,9 @@ public final class FormRecognizerAsyncClient {
      * <p>The service does not support cancellation of the long running operation and returns with an
      * error message indicating absence of cancellation support.</p>
      *
+     * Note that the {@code data} passed must be replayable if retries are enabled (the default). In other words, the
+     * {@code Flux} must produce the same data each time it is subscribed to.
+     *
      * <p><strong>Code sample</strong></p>
      * {@codesnippet com.azure.ai.formrecognizer.FormRecognizerAsyncClient.beginRecognizeCustomForms#Flux-string-long-FormContentType}
      *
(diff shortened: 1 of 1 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClient.java
```diff
@@ -424,21 +424,21 @@ public FormRecognizerServiceVersion getServiceVersion() {
         Flux<ByteBuffer> data, long length, FormContentType formContentType, boolean includeTextDetails) {
         Objects.requireNonNull(data, "'data' is required and cannot be null.");
```
samvaity, line 425: Should this be in try to be caught and not thrown?

sdk/formrecognizer/azure-ai-formrecognizer/src/samples/java/com/azure/ai/formrecognizer/FormRecognizerAsyncClientJavaDocCodeSnippets.java
```diff
@@ -333,7 +334,7 @@ public void beginRecognizeReceiptsFromUrlWithOptions() {
     public void beginRecognizeReceipts() throws IOException {
         // BEGIN: com.azure.ai.formrecognizer.FormRecognizerAsyncClient.beginRecognizeReceipts#Flux-long-FormContentType
         File sourceFile = new File("{file_source_url}");
-        Flux<ByteBuffer> buffer = Utility.toFluxByteBuffer(
+        Flux<ByteBuffer> buffer = toFluxByteBuffer(
             new ByteArrayInputStream(Files.readAllBytes(sourceFile.toPath())));
```
anuchandy, line 338: @mssfang

Following sample code will work:

```java
toFluxByteBuffer(new ByteArrayInputStream(Files.readAllBytes(sourceFile.toPath())));
```

but wondering it should be re-playable to support any retry:

```java
toFluxByteBuffer(Mono.fromCallable(() -> new ByteArrayInputStream(Files.readAllBytes(sourceFile.toPath()))));
```

Not high priority, we can create an issue and link to 9716, we should followup post release, I think scope is bigger here and needs to be discussed in this work item https://github.com/Azure/azure-sdk-for-java/issues/9716

PR #10421, opened 2020-04-22
fix: mgmt network spotbugs

ChenTanyi:


sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/ApplicationGatewayImpl.java
```diff
@@ -578,30 +578,30 @@ private static ApplicationGatewayFrontendImpl useSubnetFromIPConfigForFrontend(
      * @param byName object found by name
      * @param byPort object found by port
      * @param name the desired name of the object
-     * @return true if already found, false if ok to create, null if conflict
+     * @return positive if already found, 0 if ok to create, negative if conflict
      */
-    <T> Boolean needToCreate(T byName, T byPort, String name) {
+    <T> int needToCreate(T byName, T byPort, String name) {
```
ChenTanyi, line 583: Could be reviewed carefully.

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/ExpressRouteCircuitPeeringImpl.java
```diff
@@ -10,21 +10,28 @@
...
+import com.azure.management.resources.fluentcore.model.Refreshable;
 import com.azure.management.resources.fluentcore.model.implementation.CreatableUpdatableImpl;
 import com.azure.management.resources.fluentcore.utils.Utils;
 import java.util.Arrays;
 import reactor.core.publisher.Mono;
 
-class ExpressRouteCircuitPeeringImpl
+interface ExpressRouteCircuitPeeringParent<ModelT, InnerT>
+    extends GroupableResource<NetworkManager, InnerT>,
+    Refreshable<ModelT> {}
```
ChenTanyi, line 22: Could be reviewed carefully.

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/VirtualNetworkGatewayImpl.java
```diff
@@ -414,6 +414,7 @@ VirtualNetworkGatewayIPConfiguration defaultIPConfiguration() {
             // ...and no need to create VNet
             networkObservable = Mono.empty(); // ...and don't create another VNet
         } else {
+            assert creatableNetwork != null;
```
ChenTanyi, line 417: Not sure if it better use if instead of assert

sdk/network/mgmt/src/main/java/com/azure/management/network/ExpressRouteCircuitPeering.java
```diff
@@ -121,16 +121,16 @@
...
         /** The stage of Express Route Circuit Peering definition allowing to specify AS number for peering. */
-        interface WithPeerAsn {
+        interface WithPeerASN {
             /**
-             * @param peerAsn AS number for peering. Both 2-byte and 4-byte AS numbers can be used
+             * @param peerASN AS number for peering. Both 2-byte and 4-byte AS numbers can be used
              * @return next stage of definition
              */
-            WithCreate withPeerAsn(long peerAsn);
+            WithCreate withPeerASN(long peerASN);
```
xccc-msft, line 133: Does this pass checkstyle?

ChenTanyi: Yes, the checkstyle doesn't add ASN in blacklist.

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/ApplicationGatewayImpl.java
weidongxu-microsoft: Em, what does spotbugs say on Boolean?

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkPeeringImpl.java
```diff
@@ -24,8 +24,7 @@
         NetworkPeering, Network, VirtualNetworkPeeringInner, NetworkPeeringImpl, NetworkManager>
     implements NetworkPeering,
         NetworkPeering.Definition,
-        NetworkPeering.Update,
-        IndependentChild.DefinitionStages.WithParentResource<NetworkPeering, Network> {
```
weidongxu-microsoft, line 28: Any effect on this deleted interface?

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/VirtualNetworkGatewayImpl.java
weidongxu-microsoft: Maybe just check not null, throw IllegalState if null?

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/ApplicationGatewayImpl.java
ChenTanyi: If return null for Boolean, it will cause NPE for boolean as return type.

A method that returns either Boolean.TRUE, Boolean.FALSE or null is an accident waiting to happen. This method can be invoked as though it returned a value of type boolean, and the compiler will insert automatic unboxing of the Boolean value. If a null value is returned, this will result in a NullPointerException.

https://github.com/codacy/codacy-findbugs/blob/master/src/main/resources/docs/description/NP_BOOLEAN_RETURN_NULL.md

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/NetworkPeeringImpl.java
ChenTanyi: No, the deletion is just because the super class has such implements

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/ApplicationGatewayImpl.java
weidongxu-microsoft: Agree. But `int` does not make it clearer as well (at least for human). Who would know a integer only can take 3 values? If you do like clarity, do enum.

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/VirtualNetworkGatewayImpl.java
ChenTanyi: done

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/ExpressRouteCircuitPeeringImpl.java
weidongxu-microsoft: Any detail about this?

ChenTanyi: In order to deal with the peering inner in route filter. https://github.com/Azure/azure-sdk-for-java/pull/10421/files#diff-54b158e5bec26f0a0848f449f42c704cR55

Maybe I think I could remove the interface and use in ExpressRouteCircuitPeering template.

ChenTanyi: I have removed it and add template in ExpressRouteCircuitPeeringImpl

sdk/network/mgmt/src/main/java/com/azure/management/network/implementation/ApplicationGatewayImpl.java
ChenTanyi: done

ChenTanyi: done

ChenTanyi:
Also fix for azure-samples revapi error

PR #10429, opened 2020-04-22
Update Client Parent POM to Better Leverage Maven Inheritance

alzimmermsft:
This PR updates `sdk/parents/azure-client-sdk-parent/pom.xml` to better leverage Maven's configuration inheritance, additionally it removes some configurations that was setting default values.


sdk/core/azure-core-test/src/test/java/com/azure/core/test/implementation/RestProxyWithMockTests.java
```diff
@@ -166,14 +161,6 @@ public void service1GetDateTimeRfc1123() {
         assertEquals(OffsetDateTime.ofInstant(Instant.ofEpochMilli(0), ZoneOffset.UTC), dateTime);
     }
 
-    @Test
```
mitchdenny, line 169: Tests not required anymore?

alzimmermsft: I can add this test back, it was just relying on an illegal access to implementation in Azure Core since it was configured as Java 8.

mitchdenny: I've seen quite a few of these illegal access warnings. Is there a plan within the Java team to track them down and eliminate them? I'm assuming we don't want to disable all the tests where they occur, and instead just need to find another route to get the same test coverage?

joshfree (approved):

sdk/core/azure-core/src/main/java/module-info.java
```diff
@@ -41,6 +41,9 @@
     // Exports HttpProviders#getAllHttpClients API to azure-core-test module
     exports com.azure.core.implementation.http to com.azure.core.test;
 
+    // Exports UnixTime for testing.
+    exports com.azure.core.implementation to com.azure.core.test;
```
srnagar, line 45: I don't see any changes in `com.azure.core.test` module. Is this required?

srnagar (approved):

sdk/core/azure-core/src/main/java/module-info.java
alzimmermsft: PR is a bit stale at this point, I'll get around to getting it up to date.

alzimmermsft:
Closing PR as it is out of date and investigation needs to take another angle that doesn't affect as many downstream clients.

PR #10430, opened 2020-04-22
ServiceBusReceiverClient : Setting up Integration test

hemanttanwar:
ServiceBusReceiverClient : Setting up Integration test

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
@@ -88,11 +88,38 @@ class ServiceBusReceiverClientIntegrationTest extends IntegrationTestBase {
                 logger.warning("Error occurred when draining deferred messages Entity: {} ", receiver.getEntityPath(), e);
             }
 
-            // set empty list
-            messagesDeferred.set(new ArrayList<>());
         }
 
         dispose(receiver, sender, receiveAndDeleteReceiver);
+    }
+
+    /**
+     * Verifies that we can only call receive() once only.
+     */
+    @MethodSource("messagingEntityWithSessions")
+    @ParameterizedTest
+    void receiveByTwoSubscriber(MessagingEntityType entityType, boolean isSessionEnabled) {
+        // Arrange
+        setSenderAndReceiver(entityType, isSessionEnabled);
+        final int maxMessages = 1;
+        final Duration shortTimeOut = Duration.ofSeconds(5);
+
+        // Act & Assert
+        IterableStream<ServiceBusReceivedMessage> messages = receiver.receive(maxMessages, shortTimeOut);
+
+        int receivedMessages = messages.stream().collect(Collectors.toList()).size();
+        assertEquals(0, receivedMessages);
+
+        // Second time user try to receive, it should throw exception.
+        try {
+            messages = receiver.receive(maxMessages, shortTimeOut);
+            messages.stream().collect(Collectors.toList()).size();
+
+            //We can not have second subscriber.
+            assertTrue(false, "Should have thrown IllegalStateException.");
+        } catch (Exception ex) {
+            assertTrue(ex instanceof IllegalStateException);
+        }
     }
 
     /**
@@ -115,14 +142,15 @@ class ServiceBusReceiverClientIntegrationTest extends IntegrationTestBase {
         IterableStream<ServiceBusReceivedMessage> messages = receiver.receive(maxMessages, TIMEOUT);
 
         // Assert
-        final List<ServiceBusReceivedMessage> asList = messages.stream().collect(Collectors.toList());
-        assertEquals(maxMessages, asList.size());
-
-        for (ServiceBusReceivedMessage receivedMessage : asList) {
+        int receivedMessageCount = 0;
+        for (ServiceBusReceivedMessage receivedMessage : messages) {
             assertMessageEquals(receivedMessage, messageId, isSessionEnabled);
             receiver.complete(receivedMessage);
             messagesPending.decrementAndGet();
+            ++receivedMessageCount;
         }
+
+        assertEquals(maxMessages, receivedMessageCount);
 
     }
 
@@ -164,14 +192,15 @@ class ServiceBusReceiverClientIntegrationTest extends IntegrationTestBase {
         IterableStream<ServiceBusReceivedMessage> messages = receiver.receive(maxMessages, TIMEOUT);
 
         // Assert
-        final List<ServiceBusReceivedMessage> asList = messages.stream().collect(Collectors.toList());
-        assertEquals(maxMessages, asList.size());
 
-        for (ServiceBusReceivedMessage receivedMessage : asList) {
+        int receivedMessageCount = 0;
+        for (ServiceBusReceivedMessage receivedMessage : messages) {
             assertMessageEquals(receivedMessage, messageId, isSessionEnabled);
             receiver.complete(receivedMessage);
             messagesPending.decrementAndGet();
+            ++receivedMessageCount;
         }
+        assertEquals(maxMessages, receivedMessageCount);
 
     }
 


sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/IntegrationTestBase.java
```diff
@@ -224,4 +238,38 @@ protected void dispose(AutoCloseable... closeables) {
             }
         }
     }
+
+    static Stream<Arguments> messagingEntityProvider() {
```
hemanttanwar, line 242: I am moving common ,between sync and async client,   part in IntegrationTestBase

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/IntegrationTestBase.java
```diff
@@ -50,6 +62,9 @@
     private String testName;
     private final Scheduler scheduler = Schedulers.parallel();
 
+    static final byte[] CONTENTS_BYTES = "Some-contents".getBytes(StandardCharsets.UTF_8);
+    String sessionId;
```
conniey, line 66: private modifier for both.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,477 @@
...
+    final AtomicInteger messagesPending = new AtomicInteger();
+
+
+    protected ServiceBusReceiverClientIntegrationTest() {
+        super(new ClientLogger(ServiceBusReceiverClientIntegrationTest.class));
+    }
+
+    @BeforeAll
+    static void beforeAll() {
+        StepVerifier.setDefaultTimeout(Duration.ofSeconds(30));
```
conniey, line 57: You could probably move this BeforeAll and AfterAll setup into the base class because we want there to be a timeout.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,477 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Integration tests for {@link ServiceBusReceiverClient} from queues or subscriptions.
+ */
+public class ServiceBusReceiverClientIntegrationTest extends IntegrationTestBase {
+
+    private final ClientLogger logger = new ClientLogger(ServiceBusReceiverClientIntegrationTest.class);
+
+    ServiceBusReceiverClient receiver;
```
conniey, line 41: private

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,477 @@
...
+        final IterableStream<ServiceBusReceivedMessage> messageIte = receiver.receive(1, TIMEOUT);
+        Assertions.assertNotNull(messageIte);
+
+        final List<ServiceBusReceivedMessage> asList = messageIte.stream().collect(Collectors.toList());
+        ServiceBusReceivedMessage receivedMessage = asList.get(0);
+
+        assertNotNull(receivedMessage);
+
+        // Assert & Act
+        ServiceBusReceivedMessage receivedPeekMessage = receiver.peekAt(receivedMessage.getSequenceNumber());
```
conniey, line 171: This would be Act. And the line after would be Assert.

conniey: It's Act and then Assert. For all the other ones.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,477 @@
...
+            }
+        }
+    }
+
+    private void setSenderAndReceiver(MessagingEntityType entityType, boolean isSessionEnabled) {
+        setSenderAndReceiver(entityType, isSessionEnabled, Function.identity());
+    }
+
+    private void setSenderAndReceiver(MessagingEntityType entityType, boolean isSessionEnabled,
+                                      Function<ServiceBusClientBuilder.ServiceBusReceiverClientBuilder, ServiceBusClientBuilder.ServiceBusReceiverClientBuilder> onReceiverCreate) {
```
conniey, line 421: Indent 4 spaces rather than matching it with the method name.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,477 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
conniey, line 1: What about cases where no messages are received in that time out period? Or if you try to receive two batches, one after the other?

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/IntegrationTestBase.java
hemanttanwar: sessionId : This will be needed in `ServiceBusReceiverClientIntegrationTest`  to create receiver and sender.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+
+        sendMessage(message);
+        sendMessage(message);
+
+        // Act
+        Iterable<ServiceBusReceivedMessage> iterableMessages = receiver.receive(howManyMessage, TIMEOUT);
+
+        // Assert
+        for (ServiceBusReceivedMessage receivedMessage: iterableMessages) {
+            assertMessageEquals(receivedMessage, messageId, isSessionEnabled);
```
conniey, line 121: You didn't assert that you had two messages. If there were 0 messages, this would still pass.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+
+        dispose(receiver, sender, receiveAndDeleteReceiver);
+    }
+
+    /**
+     * Verifies that we can send and receive two messages.
+     */
+    @MethodSource("messagingEntityWithSessions")
+    @ParameterizedTest
+    void receiveTwoMessagesAutoComplete(MessagingEntityType entityType, boolean isSessionEnabled) {
```
conniey, line 105: Auto-complete is not supported in the synchronous scenario. I would expect you to be completing these manually.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+
+    /* Sometime not all the messages are cleaned-up. This is buffer to ensure all the messages are cleaned-up.*/
+    private static final int BUFFER_MESSAGES_TO_REMOVE = 10;
+
+    /**
+     * Receiver used to clean up resources in {@link #afterTest()}.
+     */
+    private ServiceBusReceiverClient receiveAndDeleteReceiver;
+
+    final AtomicInteger messagesPending = new AtomicInteger();
```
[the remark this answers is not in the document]
conniey: https://www.jetbrains.com/help/idea/reformat-and-rearrange-code.html

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+        final ServiceBusMessage message = getMessage(messageId, isSessionEnabled);
+
+        sendMessage(message);
+
+        // Act
+        Iterable<ServiceBusReceivedMessage> iterableMessages = receiver.receive(howManyMessage, TIMEOUT);
+
+        // Assert
+        for (ServiceBusReceivedMessage receivedMessage: iterableMessages) {
+            assertMessageEquals(receivedMessage, messageId, isSessionEnabled);
```
conniey, line 167: This would still pass with 0 items in the collection.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+        final String messageId = UUID.randomUUID().toString();
+        final ServiceBusMessage message = getMessage(messageId, isSessionEnabled);
+
+        sendMessage(message);
+
+        // Act
+        Iterable<ServiceBusReceivedMessage> iterableMessages = receiver.receive(howManyMessage, TIMEOUT);
+
+        // Assert
+        for (ServiceBusReceivedMessage receivedMessage: iterableMessages) {
```
conniey, line 166: There should be a space before the colon.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+            assertMessageEquals(receivedMessage, messageId, isSessionEnabled);
+        }
+
+        messagesPending.decrementAndGet();
+        messagesPending.decrementAndGet();
```
conniey, line 171: You only received one message and sent one message, this should be decremented once.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+
+        // Act
+        Iterable<ServiceBusReceivedMessage> iterableMessages = receiver.receive(howManyMessage, TIMEOUT);
+
+        // Assert
+        for (ServiceBusReceivedMessage receivedMessage: iterableMessages) {
+            assertMessageEquals(receivedMessage, messageId, isSessionEnabled);
+        }
+
+        messagesPending.decrementAndGet();
```
conniey, line 124: The reason this is here is that when we successfully receive a message, we don't need to clean it up in after test case scenario. This should be in the foreach block. Same with any uses below.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
conniey: This should be in the foreach loop as well.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+    void peekFromSequenceNumberMessage(MessagingEntityType entityType, boolean isSessionEnabled) {
+        // Arrange
+        setSenderAndReceiver(entityType, isSessionEnabled);
+
+        final String messageId = UUID.randomUUID().toString();
+        final ServiceBusMessage message = getMessage(messageId, isSessionEnabled);
+
+        sendMessage(message);
+
+        final IterableStream<ServiceBusReceivedMessage> messageIte = receiver.receive(1, TIMEOUT);
```
conniey, line 209: nit: variable name: Ite?

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+        Assertions.assertNotNull(messageIte);
+
+        final List<ServiceBusReceivedMessage> asList = messageIte.stream().collect(Collectors.toList());
+        ServiceBusReceivedMessage receivedMessage = asList.get(0);
+
+        assertNotNull(receivedMessage);
```
conniey, line 215: Is this assertion necessary? If the list was empty, the .get(0) would fail with an out of range exception. The assertion is probably `Assertions.assertEquals(1, asList.size())` before you get the first item.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+        final int fromSequenceNumber = 1;
+
+        sendMessage(message);
+        sendMessage(message);
+
+        // Act
+        IterableStream<ServiceBusReceivedMessage> iterableMessages = receiver.peekBatchAt(maxMessages, fromSequenceNumber);
+
+        // Assert
+        Assertions.assertEquals(maxMessages, iterableMessages.stream().collect(Collectors.toList()).size());
```
conniey, line 268: You should be asserting that the messages also start from that sequence number.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+        final List<ServiceBusReceivedMessage> asList = messageIte.stream().collect(Collectors.toList());
+        ServiceBusReceivedMessage receivedMessage = asList.get(0);
+
+        assertNotNull(receivedMessage);
+
+        // Act
+        ServiceBusReceivedMessage receivedPeekMessage = receiver.peekAt(receivedMessage.getSequenceNumber());
+
+        // Assert
+        assertMessageEquals(receivedPeekMessage, messageId, isSessionEnabled);
```
conniey, line 221: You should also assert that the message is from that sequence number.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,578 @@
...
+
+        sendMessage(message);
+
+        final IterableStream<ServiceBusReceivedMessage> messageIte = receiver.receive(1, TIMEOUT);
+        Assertions.assertNotNull(messageIte);
+
+        final List<ServiceBusReceivedMessage> asList = messageIte.stream().collect(Collectors.toList());
+        ServiceBusReceivedMessage receivedMessage = asList.get(0);
+
+        assertNotNull(receivedMessage);
```
conniey, line 291: Same with the comment above. If you did .get(0) and the list was empty, it would throw. This check is incorrect.

conniey: And all other instances of it.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,590 @@
...
+    /* Sometime not all the messages are cleaned-up. This is buffer to ensure all the messages are cleaned-up.*/
+    private static final int BUFFER_MESSAGES_TO_REMOVE = 10;
+
+    private final AtomicInteger messagesPending = new AtomicInteger();
+    private final AtomicReference<List<Long>> messagesDeferred = new AtomicReference<>(new ArrayList<>());
+    private final ClientLogger logger = new ClientLogger(ServiceBusReceiverClientIntegrationTest.class);
+
+    private ServiceBusReceiverClient receiver;
+    private ServiceBusSenderClient sender;
+
```
conniey, line 45: nit: extra line.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,590 @@
...
+            dispose(receiver, sender, receiveAndDeleteReceiver);
+            return;
+        }
+        // In the case that this test failed... we're going to drain the queue or subscription.
+        if (pending > 0) {
+            try {
+                IterableStream<ServiceBusReceivedMessage> removedMessage = receiveAndDeleteReceiver.receive(
+                    pending + BUFFER_MESSAGES_TO_REMOVE);
+                removedMessage.stream().forEach(receivedMessage -> {
+                    System.out.println("Removed Message sequence : " + receivedMessage.getSequenceNumber());
```
conniey, line 74: Use logger.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,590 @@
...
+
+        if (messagesDeferred.get().size() > 0) {
+            try {
+                List<Long> listOfDeferredMessages = messagesDeferred.get();
+                for (Long seqNumber : listOfDeferredMessages) {
+                    ServiceBusReceivedMessage deferredMessages = receiver.receiveDeferredMessage(seqNumber);
+                    receiver.complete(deferredMessages);
+                }
+            } catch (Exception e) {
+                logger.warning("Error occurred when draining deferred messages Entity : " + receiver.getEntityPath(), e);
```
conniey, line 89: Should use placeholders so it can be lazily evaluated.
http://www.slf4j.org/faq.html#logging_performance

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,595 @@
...
+                for (Long seqNumber : listOfDeferredMessages) {
+                    ServiceBusReceivedMessage deferredMessages = receiver.receiveDeferredMessage(seqNumber);
+                    receiver.complete(deferredMessages);
+                }
+            } catch (Exception e) {
+                logger.warning("Error occurred when draining deferred messages Entity: {} ", receiver.getEntityPath(), e);
+            }
+
+            // set empty list
+            messagesDeferred.set(new ArrayList<>());
```
conniey, line 92: Do we need the empty set? This is done during clean up. The class will be recreated after every method.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,595 @@
...
+        sendMessage(message);
+
+        // Act
+        IterableStream<ServiceBusReceivedMessage> messages = receiver.receive(maxMessages, TIMEOUT);
+
+        // Assert
+        final List<ServiceBusReceivedMessage> asList = messages.stream().collect(Collectors.toList());
+        assertEquals(maxMessages, asList.size());
+
+        for (ServiceBusReceivedMessage receivedMessage : asList) {
```
conniey, line 121: A better way is to count the number of time you go into the loop and assert at the end rather than creating the list eagerly.

```java
int count = 0;
for (Message m : messages) {
    count++;
}
assertEquals(2, count);
```

conniey: Same with other instances.

sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/ServiceBusReceiverClientIntegrationTest.java
```diff
@@ -0,0 +1,595 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Integration tests for {@link ServiceBusReceiverClient} from queues or subscriptions.
+ */
+class ServiceBusReceiverClientIntegrationTest extends IntegrationTestBase {
```
conniey, line 34: One test case you're missing is what happens if we receive() twice. Customers would probably try this in a loop several times rather than once.

PR #10431, opened 2020-04-22
Add a convenience API to create poller flux error instance

srnagar:
When creating poller flux, care has to be taken to ensure all activation, cancellation, polling and getFinalResult operations do not throw exceptions directly. This requires having try-catch block in all these methods for even null checks. This PR introduces a convenient API to create a PollerFlux error instance that simply returns an error on subscription. This simplifies the client library code.

[Reference PR](https://github.com/Azure/azure-sdk-for-java/pull/10420/files#diff-f3e13c2bd95cecb8a90329a3c1e1f6db)

Sample use:

```java
public PollerFlux<OperationResult, IterableStream<RecognizedForm>>
	beginRecognizeCustomFormsFromUrl(String fileSourceUrl, String modelId, boolean includeTextDetails,
	Duration pollInterval) {
	try {
		final Duration interval = pollInterval != null ? pollInterval : DEFAULT_DURATION;
		return new PollerFlux<OperationResult, IterableStream<RecognizedForm>>(
			interval,
			analyzeFormActivationOperation(fileSourceUrl, modelId, includeTextDetails),
			createAnalyzeFormPollOperation(modelId),
			(activationResponse, context) -> Mono.error(new RuntimeException("Cancellation is not supported")),
			fetchAnalyzeFormResultOperation(modelId, includeTextDetails));
	} catch (Exception ex) {
		 return PollerFlux.error(ex);
	}
}
```

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
@@ -41,6 +41,7 @@ import java.util.function.Function;
  * @param <U> The type of the final result of long running operation.
  */
 public final class PollerFlux<T, U> extends Flux<AsyncPollResponse<T, U>> {
+
     private final ClientLogger logger = new ClientLogger(PollerFlux.class);
     private final PollingContext<T> rootContext = new PollingContext<>();
     private final Duration defaultPollInterval;
@@ -163,15 +164,15 @@ public final class PollerFlux<T, U> extends Flux<AsyncPollResponse<T, U>> {
     }
 
     /**
-     * Creates a poller flux instance that returns an error on subscription.
+     * Creates a PollerFlux instance that returns an error on subscription.
      *
      * @param ex The exception to be returned on subscription of this {@link PollerFlux}.
      * @param <T> The type of poll response value.
      * @param <U> The type of the final result of long running operation.
      * @return A poller flux instance that returns an error without emitting any data.
      *
-     * @see {@link Mono#error(Throwable)}
-     * @see {@link Flux#error(Throwable)}
+     * @see Mono#error(Throwable)
+     * @see Flux#error(Throwable)
      */
     public static <T, U> PollerFlux<T, U> error(Exception ex) {
         return new PollerFlux<>(Duration.ofMillis(1L), context -> Mono.error(ex), context -> Mono.error(ex),


sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -162,6 +162,22 @@ private PollerFlux(Duration pollInterval,
...
+     * @param <U> The type of the final result of long running operation.
+     * @return A poller flux instance that returns an error without emitting any data.
+     *
+     * @see {@link Mono#error(Throwable)}
+     * @see {@link Flux#error(Throwable)}
+     */
+    public static <T, U> PollerFlux<T, U> error(Exception ex) {
+        return new PollerFlux<>(Duration.ofMillis(1L), context -> Mono.error(ex), context -> Mono.error(ex),
+            (context, response) -> Mono.error(ex), context -> Mono.error(ex));
+    }
```
anuchandy, line 179: so this is the equivalent of `Mono.error` or `Flux.error` in `PollerFlux`.  The activation will never succeeded so all subscriptions are going to get the error from the activation [stage](https://github.com/Azure/azure-sdk-for-java/blob/51fdb024fad315a0c8273f0a9d84a5c3cf241b76/sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java#L323). The other functions (finalResult, Poll) will never get called. lgtm.

srnagar: Yes, that's correct! This will return the error from activation function. Since poll, finalResult and cancel functions cannot be null, I just add the same error to all of them.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -162,6 +162,22 @@ private PollerFlux(Duration pollInterval,
...
+     *
+     * @param ex The exception to be returned on subscription of this {@link PollerFlux}.
+     * @param <T> The type of poll response value.
+     * @param <U> The type of the final result of long running operation.
+     * @return A poller flux instance that returns an error without emitting any data.
+     *
+     * @see {@link Mono#error(Throwable)}
+     * @see {@link Flux#error(Throwable)}
+     */
+    public static <T, U> PollerFlux<T, U> error(Exception ex) {
```
alzimmermsft, line 176: Should this use `Throwable` instead?

alzimmermsft, line 176: Do we want this to take a `ClientLogger` so the exception can be logged? Or is it expected for the caller to log this exception?

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
```diff
@@ -162,6 +162,22 @@ private PollerFlux(Duration pollInterval,
         this.syncActivationOperation = cxt -> activationOperation.apply(cxt).block();
     }
 
+    /**
+     * Creates a poller flux instance that returns an error on subscription.
```
alzimmermsft, line 166: nit: Consistency with the rest of the Javadocs

```suggestion
     * Creates a PollerFlux instance that returns an error on subscription.
```

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
alzimmermsft: This would follow the pattern when dealing with `Mono.error`, `Flux.error`, and `PagedFlux.error`.

srnagar: Initially, I started with Throwable but it conflicts with Flux's `error` method. 
So, cannot use `Throwable` here:
`'error(Throwable)' in 'com.azure.core.util.polling.PollerFlux' clashes with 'error(Throwable)' in 'reactor.core.publisher.Flux'; both methods have same erasure, yet neither hides the other`

srnagar: Logging is expected to be done by the caller. This is just like `Mono.error()`. 

What you are referring to is `FluxUtils.monoError()`.

sdk/core/azure-core/src/main/java/com/azure/core/util/polling/PollerFlux.java
anuchandy: was thinking a little more about this. So if we create the PollerFlux like below then parameter validations are delayed until consumer subscribe. 

```java
new PollerFlux<Foo, Void>(Duration.ofMillis(3),
    // The Activation Function
    new Function<PollingContext<Foo>, Mono<Foo>>() {
        @Override
        public Mono<Foo> apply(PollingContext<Foo> pollingContext) {
            Mono<Foo> initMono = startLro(start_params);
            return initMono;
        }
    },
    // The Poll Function
    new Function<PollingContext<Foo>, Mono<PollResponse<Foo>>>() {
        @Override
        public Mono<PollResponse<Foo>> apply(PollingContext<Foo> fooPollingContext) {
            Mono<PollResponse<Foo>> pollMono = doOnePoll(poll_params);
            return pollMono;
        }
    },
    // The Cancel Function
    new BiFunction<PollingContext<Foo>, PollResponse<Foo>, Mono<Foo>>() {
        @Override
        public Mono<Foo> apply(PollingContext<Foo> fooPollingContext,
                               PollResponse<Foo> fooPollResponse) {
            return Mono.error(new RuntimeException("not supported"));
        }
    },
    // The final result function
    new Function<PollingContext<Foo>, Mono<Void>>() {
        @Override
        public Mono<Void> apply(PollingContext<Foo> fooPollingContext) {
            return Mono.error(new RuntimeException("not supported"));
        }
    });
```


1. Any such delayed param validation error from the activation function will be bubbled up. 
2. But any such delayed param validation error from poll function will not bubbled up since the polling loop continue to poll on error.

The above behavior of two functions are same for both local param validation and server returned an error. The PollerFlux.error cannot be used in this pattern.
On the other hand, if we create the PollerFlux like below where we do eagerly do param validation then yes we can use PollerFlux.error to catch and convert to async error.

But this will not handle the error that poller function produce from server returned response.


```java

new PollerFlux<Foo, Void>(Duration.ofMillis(3),
    // The Activation Function
    getStartMonoFunction(start_params),
    // The Poll Function
    getPollMonoFunction(poll_params),
    // The Cancel Function
    new BiFunction<PollingContext<Foo>, PollResponse<Foo>, Mono<Foo>>() {
        @Override
        public Mono<Foo> apply(PollingContext<Foo> fooPollingContext,
                               PollResponse<Foo> fooPollResponse) {
            return Mono.error(new RuntimeException("not supported"));
        }
    },
    // The final result function
    new Function<PollingContext<Foo>, Mono<Void>>() {
        @Override
        public Mono<Void> apply(PollingContext<Foo> fooPollingContext) {
            return Mono.error(new RuntimeException("not supported"));
        }
    });

static Function<PollingContext<Foo>, Mono<Foo>> getStartMonoFunction(start_params) {
    // validation of "start_params" parameters and throwing of error outside reactor path
    return new Function<PollingContext<Foo>, Mono<Foo>>() {
        @Override
        public Mono<Foo> apply(PollingContext<Foo> pollingContext) {
            Mono<Foo> initMono = startLro();
            return initMono;
        }
    };
}

Function<PollingContext<Foo>, Mono<PollResponse<Foo>>> getPollMonoFunction(poll_params) {
     // validation of "poll_params" and throwing of error outside reactor path
    return new Function<PollingContext<Foo>, Mono<PollResponse<Foo>>>() {
        @Override
        public Mono<PollResponse<Foo>> apply(PollingContext<Foo> fooPollingContext) {
            Mono<PollResponse<Foo>> pollMono = doOnePoll();
            return pollMono;
        }
    };
}
```

srnagar: Yes, @anuchandy, you are right. The intention of creating this method is to do input validation in our public APIs. For e.g.

```java
public PollerFlux<Foo, Bar> beginPollingOperation(String data, Duration interval) {
    try {
        // All these input validations need to be done right here but will be surfaced to the user through proper error channel by returning PollerFlux.error()
         Objects.requireNonNull(data);
         Objects.requireNonNull(interval);
         if(interval.isZero()) {
               throw IllegalArgumentException("'interval' cannot be zero");
         }
         return new PollerFlux(.....);
    }
    catch (Exception ex) {
          return PollerFlux.error(ex);
    }
}
```

PR #10436, opened 2020-04-22
Minor updates for Form Recognizer samples and tests

samvaity:
- Update tests recordings to include correct User agent string.
- Update sample files , file path.


sdk/formrecognizer/azure-ai-formrecognizer/src/main/java/com/azure/ai/formrecognizer/FormTrainingClient.java
```diff
@@ -192,6 +193,7 @@ public void deleteModel(String modelId) {
      *
      * @return A {@link Mono} containing containing status code and HTTP headers
```
mssfang, line 194:
```suggestion
     * @return A {@link Response} containing containing status code and HTTP headers
```

PR #10437, opened 2020-04-22
Remove module references from parent POMs.

mitchdenny:
This PR removes all ```<module />``` references from parent POMs. I've been a little over eager here with my deletions to trigger some conversation about modules that appear to have been added recently (e.g. ```azure-test-watcher```). Basically if there is a module that I've removed from the parent POMs we need to figure out how they get attached to the root ```./pom.xml``` either direction or via an intermediate aggregate.

### DIFF
--- a/sdk/parents/azure-client-sdk-parent/pom.xml
+++ b/sdk/parents/azure-client-sdk-parent/pom.xml
@@ -1128,34 +1128,6 @@
       </build>
     </profile>
 
-    <profile>
-      <id>non-shipping-modules</id>
-      <activation>
-        <property>
-          <name>include-non-shipping-modules</name>
-        </property>
-      </activation>
-      <modules>
-        <module>../../../common/azure-test-watcher</module>
-        <module>../../../eng/spotbugs-aggregate-report</module>
-        <module>../../../eng/code-quality-reports</module>
-        <module>../../../eng/jacoco-test-coverage</module>
-      </modules>
-    </profile>
-
-    <profile>
-      <id>template-module</id>
-      <activation>
-        <property>
-          <name>env.ENABLETEMPLATEDOCS</name>
-          <value>true</value>
-        </property>
-      </activation>
-      <modules>
-        <module>../../template/azure-sdk-template</module>
-      </modules>
-    </profile>
-
     <!-- Skip overview generation unless -Dgenerate-overview is passed as an argument.
       This is specifically done to prevent requiring python as a pre-requisite since
       this is only needed by the build system. -->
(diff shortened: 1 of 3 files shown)


sdk/parents/azure-client-sdk-parent/pom.xml
```diff
@@ -1455,41 +1427,4 @@
       </build>
     </profile>
   </profiles>
-
-  <modules>
```
alzimmermsft, line 1459: Have the aggregate-reports been fixed where this is no longer needed?

mitchdenny: Aggregate reports runs against the root ```./pom.xml``` file and uses that module list. The parent POM isn't even in the graph directly except by parent references (which don't cause module evaluation).

mitchdenny: ... unless I misunderstand the question/concern :)

mitchdenny: Couldn't we just use this goal?

http://maven.apache.org/plugins/maven-checkstyle-plugin/checkstyle-aggregate-mojo.html

Then do something similar to what we do with the code coverage reports and have a module which runs this goal and does the aggregation?

alzimmermsft: So we do have that goal configured in the client parent POM, but the configuration isn't allowed to be inherited by children POMs. I think we could allow it to be inherited to resolve the Checkstyle aggregate generation issue.

JonathanGiles:
@mitchdenny Does this change mean that there is no way in the IDE to only open the track two client libraries any longer, and that I need to have open in the IDE track one and track two client libraries and all management libraries?

mitchdenny:
It doesn't load all management libraries. But it will load track 1 libraries. It is possible that it will load all management libraries in the future as the engineering systems converge. The constraint that we are living with here is that Maven enforces that a single project cannot be referenced multiple times within the POM hierarchy.

So you can have a model where you have a single POM aggregator hierarchy (this is what we have now) or multiple hierarchies (what we had). The downside with the single hierarchy is that its potentially big. The downside of the multiple hierarchies is that if you have common tooling that you want to share you have to build and install that tooling separately BEFORE you build your target hierarchy.

That said it might be feasible (for the time being) to introduce an additional layer of aggregates to segment the tree. In effect it would be the same as pom.client.xml but without all the project inheritence stuff, its purely a structural piece to isolate track 1, track 2 and management plane (in the future).

However, as management plane starts to adopt our track 2 core, that will be less palatable, and in fact we'll want management plane to be part of the track 2 project model to drive compat testing etc.

mitchdenny:
That said ```mvn compile``` only takes about 11 seconds on my machine (including all track 1 and track 2 modules).

PR #10456, opened 2020-04-23
Exclude exception spotbugs in management samples

ChenTanyi:


eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
```diff
@@ -1965,4 +1965,10 @@
     <Bug pattern="NP_LOAD_OF_KNOWN_NULL_VALUE,SKIPPED_CLASS_TOO_BIG"/>
   </Match>
 
+  <!-- Exclude REC_CATCH_EXCEPTION for management samples. All samples in management catch Exception instead of throw it. -->
+  <Match>
+    <Class name="~com\.azure\.management(\.[^.]+)*\.samples(\.[^.]+)*"/>
```
weidongxu-microsoft, line 1970: `<Package>` might be better?

ChenTanyi: done

PR #10459, opened 2020-04-23
Added jdk httpclient to live tests

sima-zhu:
The jdk httpclient needs to setup on Java 11+. 
Issue link: https://github.com/Azure/azure-sdk-for-java/issues/10465
Jdk httpclient has some blacklist headers (like Date) which affect the results of live tests, so it cannot be tested over Java 11.
Think about to add another Java version (like 13, MTS) to test Jdk httpclient.
Refer this build:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=368178&view=logs&j=d4b67cd2-6572-561e-035e-54acb01c3f6e
This PR is still in discussion. Welcome to review, but this is not ready to merge yet. Thanks!

### DIFF
--- a/sdk/keyvault/azure-security-keyvault-certificates/pom.xml
+++ b/sdk/keyvault/azure-security-keyvault-certificates/pom.xml
@@ -122,7 +122,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
-            <version>2.22.0</version>
+            <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-surefire-plugin;external_dependency} -->
             <configuration>
               <systemPropertyVariables>
                 <jdk.httpclient.allowRestrictedHeaders>Content-Length</jdk.httpclient.allowRestrictedHeaders>
--- a/sdk/keyvault/azure-security-keyvault-keys/pom.xml
+++ b/sdk/keyvault/azure-security-keyvault-keys/pom.xml
@@ -127,7 +127,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
-            <version>2.22.0</version>
+            <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-surefire-plugin;external_dependency} -->
             <configuration>
               <systemPropertyVariables>
                 <jdk.httpclient.allowRestrictedHeaders>Content-Length</jdk.httpclient.allowRestrictedHeaders>
--- a/sdk/keyvault/azure-security-keyvault-secrets/pom.xml
+++ b/sdk/keyvault/azure-security-keyvault-secrets/pom.xml
@@ -128,7 +128,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
-            <version>2.22.0</version>
+            <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-surefire-plugin;external_dependency} -->
             <configuration>
               <systemPropertyVariables>
                 <jdk.httpclient.allowRestrictedHeaders>Content-Length</jdk.httpclient.allowRestrictedHeaders>


eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -25,7 +25,7 @@ parameters:
       JavaVersion: '1.8'
       DisplayName: 'Windows - Java 8'
       CloudType: AzureCloud
-      AZURE_TEST_HTTP_CLIENTS: netty
+      AZURE_TEST_HTTP_CLIENTS: jdk
```
alzimmermsft, line 28: This won't work, the JDK client only works in Java 11+

eng/versioning/version_client.txt
```diff
@@ -45,6 +45,7 @@ com.azure:azure-test-watcher;1.0.0-beta.1;1.0.0-beta.1
 # Format;
 # unreleased_<groupId>:<artifactId>;dependency-version
 # note: The unreleased dependencies will not be manipulated with the automatic PR creation code.
+unreleased_com.azure:azure-core-http-jdk-httpclient;1.0.0-beta.1
```
alzimmermsft, line 48: We don't need this unreleased tag since the current and dev version are the same, the libraries taking a dependency on this can just use the dependency tag.

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -57,6 +57,12 @@
       <version>1.2.1</version> <!-- {x-version-update;com.azure:azure-core-http-okhttp;dependency} -->
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-http-jdk-httpclient</artifactId>
+      <version>1.0.0-beta.1</version> <!-- {x-version-update;unreleased_com.azure:azure-core-http-jdk-httpclient;dependency} -->
+      <scope>test</scope>
+    </dependency>
```
alzimmermsft, line 65: Having this as a general include will fail build on Java 10 and below, I think this dependency will need to be added through a profile to gate it on Java 11+.

@anuchandy @srnagar what did you do to work around this issue when building the repo in Java 8?

anuchandy: had to override the profile inherited from the parent pom, like this https://github.com/Azure/azure-sdk-for-java/blob/cd464cfe7c217332e1d02a29b7a78c3ff04ef9b6/sdk/core/azure-core-http-jdk-httpclient/pom.xml#L131

sima-zhu: I have tried in parent pom. It triggers dependency cycle problem.

sima-zhu: https://dev.azure.com/azure-sdk/public/_build/results?buildId=368008&view=logs&j=9d510ff3-53a8-5192-bd2d-466ce830bc13&t=01de8c6b-9f4b-5017-894e-951524cf562e

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -100,4 +100,20 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <profiles>
```
alzimmermsft, line 103: Don't all HTTP based clients need to be updated with this profile before this change is merged? Otherwise won't there be failing live tests.

sdk/core/azure-core-test/src/main/java/com/azure/core/test/TestBase.java
```diff
@@ -164,7 +164,7 @@ protected void afterTest() {
     public static boolean shouldClientBeTested(HttpClient client) {
         String configuredHttpClientToTest = Configuration.getGlobalConfiguration().get(AZURE_TEST_HTTP_CLIENTS);
         if (CoreUtils.isNullOrEmpty(configuredHttpClientToTest)) {
-            return client.getClass().getSimpleName().equals(AZURE_TEST_HTTP_CLIENTS_VALUE_NETTY);
+            return client.getClass().getSimpleName().toLowerCase().contains(AZURE_TEST_HTTP_CLIENTS_VALUE_NETTY);
```
alzimmermsft, line 167: Could this use `equalsCaseIgnore`?

sima-zhu: We check contains instead equal

sdk/appconfiguration/azure-data-appconfiguration/pom.xml
```diff
@@ -100,4 +100,20 @@
...
+    <profile>
+      <id>java-lts</id>
+      <activation>
+        <jdk>[11,)</jdk>
```
alzimmermsft, line 107: This activation profile can be scoped more by adding in a check for environment variable `AZURE_TEST_HTTP_CLIENTS` being `jdk`. This may be the better way to go, because in that scenario we could potentially remove the Netty implementation from the class path, which would allow us to fully use Maven to handle which HTTP client will be used and we no longer need to hook this logic into JUnit as we do now.

alzimmermsft: https://maven.apache.org/surefire/maven-surefire-plugin/examples/configuring-classpath.html

alzimmermsft: @srnagar @anuchandy @JonathanGiles Thoughts on this logic change

srnagar: We will have a full matrix of tests enabled for all clients and all service versions that run less frequently than the nightly live tests. So, this profile should be activated when the env variable is set to either `jdk` or `all`. So, we should not scope it down to just `jdk`. And when it's `all`, letting maven handle classpath is going to be tricky.

sdk/keyvault/azure-security-keyvault-certificates/pom.xml
```diff
@@ -103,4 +103,20 @@
...
 
   </dependencies>
+  <profiles>
+    <profile>
+      <id>java-lts</id>
+      <activation>
+        <jdk>[11,)</jdk>
+      </activation>
+      <dependencies>
+        <dependency>
```
mitchdenny, line 113: Why does this dependency need to be added to KeyVault specifically? Isn't this a core thing, why is KeyVault special?

JimSuplizio: @mitchdenny I think this may be for the additional test dependency however there's a discussion above in the sdk/appconfiguration/azure-data-appconfiguration/pom.xml changes.

kurtzeborn:
@conniey, as the Java SME for EngSys related concerns, would you take a look at this PR and decide whether this is still a direction worth continuing in.  If this is still desirable to add, please work with @JimSuplizio and @mitchdenny to get this into a state where it can be merged.  Sima is no longer working on Java related work (though she certainly can help explain anything in the PR that doesn't make sense) and has moved on to other tasks.

sdk/core/azure-core-http-jdk-httpclient/src/test/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilderTests.java
```diff
@@ -231,4 +242,104 @@ public void buildWithNonProxyConfigurationProxy() {
...
+
+        Set<String> expectedRestrictedHeaders = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        expectedRestrictedHeaders.addAll(JdkAsyncHttpClientBuilder.DEFAULT_RESTRICTED_HEADERS);
+        expectedRestrictedHeaders.removeAll(Arrays.asList("content-length", "upgrade"));
+
+        validateRestrictedHeaders(jdkAsyncHttpClientBuilder, expectedRestrictedHeaders, 3);
+    }
+
+    @Test
+    void testCaseInsensitivity() {
```
conniey, line 321: Why not use parameterised test for this?

sdk/core/azure-core-http-jdk-httpclient/src/test/java/com/azure/core/http/jdk/httpclient/JdkAsyncHttpClientBuilderTests.java
```diff
@@ -231,4 +242,104 @@ public void buildWithNonProxyConfigurationProxy() {
...
+        System.setProperty("jdk.httpclient.allowRestrictedHeaders", "content-LENGTH");
```
conniey, line 322: If you set this system property, after the test it should be returned to its original state.

mitchdenny:
Should this now be closed?

mitchdenny:
In fact, I am pre-emptively closing it ... if anyone needs to pick it up they can re-open.

PR #10464, opened 2020-04-23
Object replication

rickle-msft:


rickle-msft:
I gotta test download, too.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobDownloadHeaders.java
```diff
@@ -342,6 +364,70 @@ public BlobDownloadHeaders setMetadata(Map<String, String> metadata) {
...
+                    objectReplicationSourcePolicies.get(policyId)
+                        .putRuleAndStatus(ruleId, this.objectReplicationRuleStatus.get(str));
+                } else {
+                    ObjectReplicationPolicy policy = new ObjectReplicationPolicy(policyId);
+                    policy.putRuleAndStatus(ruleId, objectReplicationRuleStatus.get(str));
+                    objectReplicationSourcePolicies.put(policyId, policy);
+                }
+            }
+        }
+        return objectReplicationSourcePolicies;    }
```
gapra-msft, line 416: nit : add a new line to the end in swagger README

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobItem.java
```diff
@@ -15,6 +17,30 @@
...
+    public BlobItem() {
+
+    }
+
+    /**
+     * Initializes a new blob item.
+     *
+     * @param blobItemInternal The internal structure from which to pull state.
+     */
+    public BlobItem(BlobItemInternal blobItemInternal) {
```
gapra-msft, line 33: Is this ok to do since BlobItemInternal is "private" and BlobItem is public. Would it be better to like extract the pieces in a private method wherever you use this constructor and make this constructor like take in all the properties?

gapra-msft: Though this is technically cleaner

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
```diff
@@ -191,6 +194,116 @@ public BlobProperties(final OffsetDateTime creationTime, final OffsetDateTime la
         this.committedBlockCount = committedBlockCount;
         this.versionId = versionId;
         this.isCurrentVersion = isCurrentVersion;
+
+        this.objectReplicationSourcePolicies = new HashMap<>();
+        this.objectReplicationDestinationPolicyId = objectReplicationStatus.getOrDefault("policy-id", null);
```
gapra-msft, line 199: Is there a clean way we could unduplicate this code? I'm not sure where it could live though.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
```diff
@@ -191,6 +194,116 @@ public BlobProperties(final OffsetDateTime creationTime, final OffsetDateTime la
...
+     * @param isAccessTierInferred Flag indicating if the access tier of the blob was inferred from properties of the
+     * blob.
+     * @param archiveStatus Archive status of the blob.
+     * @param encryptionKeySha256 SHA256 of the customer provided encryption key used to encrypt the blob on the server.
+     * @param accessTierChangeTime Datetime when the access tier of the blob last changed.
+     * @param metadata Metadata associated with the blob.
+     * @param committedBlockCount Number of blocks committed to an append blob, if the blob is a block or page blob
+     * pass {@code null}.
+     * @param versionId The version identifier of the blob.
+     * @param isCurrentVersion Flag indicating if version identifier points to current version of the blob.
```
gapra-msft, line 255: Why are there 2 similar seeming constructors added? Is it to make the OR logic separate from the max constructor?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
```diff
@@ -926,6 +926,41 @@ class BlobAPITest extends APISpec {
         null     | null       | null        | null         | garbageLeaseID
     }
 
+    /*
+    This test requires two accounts that are configured in a very specific way. It is not feasible to setup that
+    relationship programmatically, so we have recorded a successful interaction and only test recordings.
+     */
+    @Requires( {playbackMode()})
```
gapra-msft, line 933: Nice, I think this is a great way to test features that need to be configured a specific way

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobItem.java
rickle-msft: Hmm. Interesting. I guess that is what we do with BlobProperties. Though I don't remember the reason why. There must've been a reason other than we needed to construct it from different sources (download, getProperties, etc.) or we would've just had multiple constructors.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
rickle-msft: I don't think so. One lives in the readme and the other lives in BlobProperties. And they aren't totally identical because the one here saves this as the actual property rather than constructing it in the getter, and that would be way more complicated to do in swagger.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
rickle-msft: It's because sometimes we construct this type from downloadHeaders, which has already deserialized the ORS stuff (second constructor) and sometimes we construct this type from getPropertiesHeaders, which hasn't already done this.

Our story for returning these properties is a mess.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobDownloadHeaders.java
```diff
@@ -12,27 +12,107 @@
...
  */
 @JacksonXmlRootElement(localName = "Blob-Download-Headers")
 @Fluent
 public final class BlobDownloadHeaders {
+
+    public BlobDownloadHeaders() {
+
+    }
+
+    public BlobDownloadHeaders(com.azure.storage.blob.implementation.models.BlobDownloadHeaders headers) {
```
gapra-msft, line 28: Can we add a comment somewhere just for future reference for why we have these two classes?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
gapra-msft: Is there a way we could get the above constructor to call into this one (it could still do that deserialization) in the above constructor

rickle-msft: Yea that's probably more concise. I can do the deserialization and then call into the last constructor. Good call.

rickle-msft: Turns out this isn't possible. The compiler needs the call into another constructor to be the first line :(

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobProperties.java
rickle-msft: Even with the update of moving this out of the readme, I don't know that there's a great place for us to put this code to get rid of the duplication. There's not really a common code path between them, and we're trying to move away from using utility methods.

Issue #10467, opened 2020-04-23
[BUG] Revisit ReactorNettyHttpResponse::close()


anuchandy:
It looks like there are cases where when `ReactorNettyHttpResponse::close()` is called the connection is in the disposed state but the underlying body is not drained, which can result in leak warnings from netty. 

Need to followup with the reactor-netty team to understand the behavior of connection disposal.

This is not impacting the majority of the cases since almost all cases either we subscribe to the body in RestPoxy or by the user when downloading as raw ByteBuf. One case in key-vault revealed this issue where it was not disposing the body inside the policy and was expecting close() to do it

krishgeeneuron:
Hi,

We are facing the same issue with azure keyvault 4.1.5. Please could someone help us identify the right version the below leak issue is fixed. Is there any azure key vault version below 4.2.0 which has the fix ?

{
"timestamp": 1625580889173,
"status": 500,
"error": "Internal Server Error",
"message": "Request processing failed; nested exception is java.lang.RuntimeException: Max retries 3 times exceeded. Error Details: failed to allocate 16777216 byte(s) of direct memory (used: 486539271, max: 501743616)"
}

trask:
We ran across this when porting from from Apache HttpClient to Azure SDK HttpClient.

Here's a small repro in case it helps with investigation: https://github.com/trask/netty-bytebuf-leak-repro

In the meantime, always consuming the response (even in cases where we don't require it) prevents the issue from occurring.

alzimmermsft:
Thank you for the reproduction of this @trask!

alzimmermsft:
@krishgeeneuron, unfortunately there are no versions of KeyVault Keys `4.1.x` that contain of fix for the error you've seen. Any version later than `4.2.3` should contain a fix where retried exceptions consume the error response body before retrying, resolving the `ByteBuf` leak issue.

alzimmermsft:
@anuchandy, is the goal of this issue to resolve `ByteBuf` leaks when using Azure SDKs with the Netty `HttpClient` implementation or to resolve any usage of the Netty `HttpClient` implementation? If it is the former this has already be resolved as `azure-core` will handle consuming the response body on an error retry attempt, if it is the latter the Netty `HttpClient` implementation assumes that the caller will handle the consumption of the body.

anuchandy:
Hi @alzimmermsft - At that time, it was that - 

[...]

Issue #6096, opened 2019-10-30
Investigate ByteBuf resource leak in Azure Core Netty


g2vinay:
Steps to reproduce:

1. Set up Block Blob Client in Storage
2. Call Download method in a loop atleast 10 times.

Oct 30, 2019 4:37:12 PM io.netty.util.ResourceLeakDetector reportTracedLeak
SEVERE: LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records:
Created at:
    io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:349)
    io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:187)
    io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:178)
    io.netty.buffer.AbstractByteBufAllocator.buffer(AbstractByteBufAllocator.java:115)
    io.netty.handler.ssl.SslHandler.allocate(SslHandler.java:2143)
    io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1324)
    io.netty.handler.ssl.SslHandler.decodeJdkCompatible(SslHandler.java:1224)
    io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1271)
    io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:505)
    io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:444)
    io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:283)
    io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:374)
    io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:360)
    io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:352)
    io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1422)
    io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:374)
[...]

PR #10468, opened 2020-04-23
Addressing leak in KeyVaultCredentailPolicy.

anuchandy:
Related issues:

  https://github.com/Azure/azure-sdk-for-java/issues/7934
  https://github.com/Azure/azure-sdk-for-java/issues/6096

The `HttpResponse::close` in azure-core needs to be revisited to see whether similar logic is needed there as well, due to the bigger scope we require more time to understand why current disposal of connection is not sufficient. Tracking issue on core https://github.com/Azure/azure-sdk-for-java/issues/10467

### DIFF
--- a/sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/implementation/KeyVaultCredentialPolicy.java
+++ b/sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/implementation/KeyVaultCredentialPolicy.java
@@ -10,6 +10,7 @@ import com.azure.core.http.HttpPipelineNextPolicy;
 import com.azure.core.http.HttpResponse;
 import com.azure.core.http.policy.HttpPipelinePolicy;
 import com.azure.core.util.CoreUtils;
+import com.azure.core.util.logging.ClientLogger;
 import reactor.core.publisher.Mono;
 
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


sdk/keyvault/azure-security-keyvault-secrets/src/main/java/com/azure/security/keyvault/secrets/implementation/KeyVaultCredentialPolicy.java
```diff
@@ -52,8 +52,20 @@ public KeyVaultCredentialPolicy(TokenCredential credential) {
...
-            .doOnNext(HttpResponse::close)
+            .doOnNext(httpResponse -> {
+                // KV follows challenge based auth. Currently every service
+                // call hit the endpoint for challenge and then resend the
+                // request with token. The challenge response body is not
+                // consumed, not draining/closing the body will result in leak.
+                // Ref: https://github.com/Azure/azure-sdk-for-java/issues/7934
+                try {
+                    httpResponse.getBody().subscribe().dispose();
+                } catch (Throwable ignored) {
```
srnagar, line 63: Should this throwable be logged as warning?

anuchandy: yes, we should, will update the PR. It's not throwing now but I put it as a safeguard to ensure this version of key vault lib works with the future version of reactor-netty HTTP client when it gets addressed in the core level.

krishgeeneuron:
Hi,

We are facing the same issue with azure keyvault 4.1.5. Please could someone help us identify the right version the below leak issue is fixed.

{
    "timestamp": 1625580889173,
    "status": 500,
    "error": "Internal Server Error",
    "message": "Request processing failed; nested exception is java.lang.RuntimeException: Max retries 3 times exceeded. Error Details: failed to allocate 16777216 byte(s) of direct memory (used: 486539271, max: 501743616)",
    "path": "/siig-0.1/rest/sii/oss"
}

vcolin7:
Hello @krishgeeneuron, I think you should be able to find this on version `4.2.8` and above, but I would recommend you use `4.3.0` as it includes some new APIs and key bug fixes.

PR #10470, opened 2020-04-23
Add InputStream to Flux<ByteBuffer> Utility Method

### LINKED ISSUE #9716: Add utility method to convert incoming stream to Flux<ByteBuffer>

alzimmermsft:
Fixes #9716

This PR adds an `InputStream` to `Flux<ByteBuffer>` utility method into `FluxUtil`. There are two overloads where one defaults a chunk size and the other takes it as a parameter. The `chunkSize` will be used to fill the `ByteBuffer`s with a certain number of bytes, this is useful in scenarios such as Storage where such a function may be used to parallelize block uploads of a certain size. This utility function **DOES NOT** guarantee replayability of the stream given that `InputStream`s aren't guaranteed to be replayable either. This could be changed by caching the buffer stream but this runs the risk of exhausting memory if a large file is being converted.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
@@ -545,7 +545,7 @@ public final class FluxUtil {
                     doRead();
                 }
                 int missed = 1;
-                for (; ; ) {
+                while (true) {
                     if (cancelled) {
                         return;
                     }


sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
```diff
@@ -83,6 +84,74 @@ private static void accept(ByteArrayOutputStream byteOutputStream, ByteBuffer by
...
+    public static Flux<ByteBuffer> toFluxByteBuffer(InputStream inputStream) {
+        return toFluxByteBuffer(inputStream, 4096);
+    }
+
+    /**
+     * Converts an {@link InputStream} into a {@link Flux} of {@link ByteBuffer}.
+     * <p>
+     * Given that {@link InputStream} is not guaranteed to be replayable the returned {@link Flux} should be considered
+     * non-replayable as well.
+     * <p>
```
anuchandy, line 107: Should we document that the caller is responsible for closing the provided stream? (if the user provides the produced Flux to our async API then they should dispose it once async API completes)

alzimmermsft: Good question, I don't think in any other location where we are converting an `InputStream` into a `Flux<ByteBuffer>` we close the stream.

anuchandy: yes, the Flux.generate is not closing stream and it shouldn't. I mean consumer of this API is in charge of dealing with it.  

```java
InputStream is = // open-file
FormRecognizer.someAsyncAPI(FluxUtils.toByteBufferFlux(is))
    .doFinally(s -> {
       is.close();
    }).block();
```

srnagar: The lifecyle of the inputstream should be managed by the user. They might decide to reset and reuse the inputstream if they want to. So, we shouldn't be closing user provided stream.

sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
```diff
@@ -83,6 +84,74 @@ private static void accept(ByteArrayOutputStream byteOutputStream, ByteBuffer by
...
+     *
+     * @param inputStream The {@link InputStream} to convert into a {@link Flux}.
+     * @param chunkSize The requested size for each {@link ByteBuffer}.
+     * @return A {@link Flux} of {@link ByteBuffer ByteBuffers} that contains the contents of the stream.
+     * @throws IllegalArgumentException If {@code chunkSize} is less than or equal to {@code 0}.
+     */
+    public static Flux<ByteBuffer> toFluxByteBuffer(InputStream inputStream, int chunkSize) {
+        if (chunkSize <= 0) {
+            throw new IllegalArgumentException("'chunkSize' must be greater than 0.");
+        }
```
srnagar, line 118: A method returning a reactive publisher shouldn't throw exceptions. It makes chaining of reactive operators difficult as the caller will have to handle errors in 2 ways. Return Flux.error() instead.

sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
anuchandy: yes.

Also, any thoughts on the replayability of such Fluxes?  I see this https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#markSupported() API to check the stream can be marked and reset. Not sure how feasible it with File streams.

In FR APIs we do two subscriptions - one for contentType detection and another one when sending the content. The sync API takes `InputStream` from the user, convert that to Flux and call async API, so those APIs won't work as expected.

alzimmermsft: @anuchandy I think it would be safer for the FormRecognizer API to not subscribe twice in case the Flux being sent isn't replayable, this goes back to the problem where we need to assume `Flux` isn't replayable since treating as non-replayable will work for replayable but not the other way around.

sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
alzimmermsft: Changed to `Flux.error`

sdk/core/azure-core/src/main/java/com/azure/core/util/FluxUtil.java
anuchandy: I thought, in general, our SDKs assume Flux is "replayable" :), otherwise, retry is not going to work (retry requires multi-subscription). Storage is a special case where they are committed to supporting "non-replyable" Flux, that makes sense because of server-side chunk support.

We had some discussion on the same topic, in this thread https://github.com/Azure/azure-sdk-for-java/pull/9953#discussion_r411589700

sima-zhu:
Storage use case is to take another expected length parameter. If length are not the same throw UnexpectedLengthException. Can you have another overload?

alzimmermsft:
> 
> 
> Storage use case is to take another expected length parameter. If length are not the same throw UnexpectedLengthException. Can you have another overload?

@sima-zhu I don't think that is necessary since there is code in Azure Core when sending a request which will throw the exception if it sees too many bytes compared to the `Content-Length` header. https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/RestProxy.java#L134

PR #10480, opened 2020-04-24
mgmt: Deprecate getAccessProfile in aks and change some api

ChenTanyi:
close https://github.com/Azure/azure-libraries-for-java/issues/1144

### DIFF
--- a/sdk/containerservice/mgmt/src/main/java/com/azure/management/containerservice/KubernetesCluster.java
+++ b/sdk/containerservice/mgmt/src/main/java/com/azure/management/containerservice/KubernetesCluster.java
@@ -43,10 +43,10 @@ public interface KubernetesCluster
     byte[] userKubeConfigContent();
 
     /** @return the Kubernetes credentials with administrative privileges to the cluster */
-    List<CredentialResult> adminKubeConfigList();
+    List<CredentialResult> adminKubeConfigs();
 
     /** @return the Kubernetes credentials with user-level privileges to the cluster */
-    List<CredentialResult> userKubeConfigList();
+    List<CredentialResult> userKubeConfigs();
 
     /** @return the service principal client ID */
     String servicePrincipalClientId();
--- a/sdk/containerservice/mgmt/src/main/java/com/azure/management/containerservice/implementation/KubernetesClusterImpl.java
+++ b/sdk/containerservice/mgmt/src/main/java/com/azure/management/containerservice/implementation/KubernetesClusterImpl.java
@@ -64,7 +64,7 @@ public class KubernetesClusterImpl
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


sdk/containerservice/mgmt/src/main/java/com/azure/management/containerservice/KubernetesCluster.java
```diff
@@ -40,6 +42,12 @@
     /** @return the Kubernetes configuration file content with user-level privileges to the cluster */
     byte[] userKubeConfigContent();
 
+    /** @return the Kubernetes credentials with administrative privileges to the cluster */
+    List<CredentialResult> listadminKubeConfigs();
```
weidongxu-microsoft, line 46: Just `adminKubeConfigs`?

ChenTanyi: Is it confused with above one `adminKubeConfigContent`?

weidongxu-microsoft: Better to have a name which matches the functionality. User your expertise :-)

Personally I prefer complete name like `adminKubernetesConfigures` or `adminKubernetesConfigureCredentials` (return type is `CredentialResult`, so it is config or credential?)

ChenTanyi: Maybe `KubeConfig` is better than `KubernetesConfigurate` since the default file name is `kube.config`. So I think the final name could be `adminKubeConfigCredentials`?

weidongxu-microsoft: I am still not sure whether it is credential or config...

If it is credential, this name should be fine. If it is config, maybe we can name it `adminKubeConfigList`?

```
CredentialResult

    /**
     * The name of the credential.
     */
    @JsonProperty(value = "name", access = JsonProperty.Access.WRITE_ONLY)
    private String name;

    /**
     * Base64-encoded Kubernetes configuration file.
     */
    @JsonProperty(value = "value", access = JsonProperty.Access.WRITE_ONLY)
    private byte[] value;
```

ChenTanyi: In kubernetes, the cluster credential is stored in a config file named `kube.config`, which will be read by `kubectl`. This is what the `value` is. I'm not sure it is said as credential or config.

ChenTanyi: Finally, I think naming to `KubeConfigs` is better.

Issue #10415, opened 2020-04-22
[FEATURE REQ] core-mgmt - remove CloudException type


weidongxu-microsoft:
Remove CloudException type from azure-core-management
Move CloudError into .exception package

Issue #10414, opened 2020-04-22
[FEATURE REQ] core-mgmt - remove serializer package


weidongxu-microsoft:
Remove serializer package

PR #10481, opened 2020-04-24
prelimiary refactor on CloudError for mgmt

weidongxu-microsoft:
Fix https://github.com/Azure/azure-sdk-for-java/issues/10415
Fix https://github.com/Azure/azure-sdk-for-java/issues/10414

Detail of the implementation would depend on how azure-core would handle the util to help create the e.g. `CloudError` from `HttpResponseException`.

Also the flatten of `error` property would also require revisit.

Plan is to has `CloudError` as base class. If any mgmt service defines error type having more properties, generate a new class that extends `CloudError` class.

### DIFF
--- a/sdk/core/azure-core-management/src/test/java/com/azure/core/management/exception/CloudErrorTests.java
+++ b/sdk/core/azure-core-management/src/test/java/com/azure/core/management/exception/CloudErrorTests.java
@@ -14,6 +14,7 @@ import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Objects;
 
 public class CloudErrorTests {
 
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core-management/src/test/java/com/azure/core/management/exception/CloudErrorTests.java
```diff
@@ -0,0 +1,89 @@
...
+            }.getType(), SerializerEncoding.JSON);
+        WebError webError = errorResponse.getError();
+        Assertions.assertEquals("WepAppError", webError.getCode());
+        Assertions.assertEquals("Deployment error.", webError.getInnererror());
+    }
+
+    @Immutable
+    private static class WebError extends CloudError {
+        @JsonProperty(value = "innererror", access = JsonProperty.Access.WRITE_ONLY)
+        private String innererror;
```
weidongxu-microsoft, line 57: This happens to web service, which have an additional `innererror` field.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/exception/CloudError.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * @return the additional info for the error.
+     */
+    public List<CloudErrorAdditionalInfo> getAdditionalInfo() {
+        return additionalInfo;
+    }
+
+    @Override
+    public String toString() {
+        return message == null ? super.toString() : message;
```
ChenTanyi, line 101: Is it better to return the whole object rather than message? Since there is one part: detail.

weidongxu-microsoft: `details` is the list, so...

PR #10490, opened 2020-04-24
Artifact segregation

mitchdenny:
This PR changes the main build step in Maven to do a _deploy_ instead of _package_. But instead of installing into the ```.m2``` cache we install into the artifact staging directory. Basically lifting a repository instead of trying to roll our own segregation.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -114,7 +114,7 @@ jobs:
         inputs:
           mavenPomFile: pom.xml
           goals: 'deploy'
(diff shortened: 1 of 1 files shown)


mitchdenny:
This pipeline shows what an upload Maven repo looks like as an artifact:

https://dev.azure.com/azure-sdk/public/_build/results?buildId=371058&view=artifacts&type=publishedArtifacts

JimSuplizio:
@mitchdenny, since this does change the artifacts directory I would highly recommend verifying that the [publish package](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/templates/stages/archetype-java-release.yml#L104) and [publish docs](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/templates/stages/archetype-java-release.yml#L143) steps still work. The publish docs ultimately ends up calling [copy-docs-to-blobstorage.ps1](https://github.com/Azure/azure-sdk-tools/blob/master/eng/common/scripts/copy-docs-to-blobstorage.ps1) and it would be good to know that this process still works. I'd rather not have surprises happen during the next release.

mitchdenny:
Yeah ... good call I'm merging this into a feature branch on upstream to do some experimentation.

PR #10492, opened 2020-04-24
refactor mgmt rest client and integrate azure identity

xseeseesee:
Fix [Azure/azure-libraries-for-java#1111](https://github.com/Azure/azure-libraries-for-java/issues/1111)
Fix [Azure/azure-libraries-for-java#1168](https://github.com/Azure/azure-libraries-for-java/issues/1168)

- [x] refactor rest client
- [x] integrate azure identity
- [x] move auth file and related classes to test package
- [x] resources - code changes, playback tests, living tests
- [x] rest modules - code changes, playback tests, living tests
- [x] for classes in package `com.azure.management`, delete the classes not required before merging this PR and move the other classes into expected location

### DIFF
--- a/sdk/management/azure/src/main/java/com/azure/management/Azure.java
+++ b/sdk/management/azure/src/main/java/com/azure/management/Azure.java
@@ -205,9 +205,19 @@ public final class Azure {
         Authenticated withSdkContext(SdkContext sdkContext);
 
         /**
+         * Specifies a specific tenant for azure.
+         *
+         * <p>Only Graph RBAC APIs require a tenant to be selected.</p>
+         *
+         * @param tenantId the ID of the tenant
+         * @return the authenticated itself for chaining
+         */
+        Authenticated withTenantId(String tenantId);
+
+        /**
          * Selects a specific subscription for the APIs to work with.
          *
-         * <p>Most Azure APIs require a specific subscription to be selected.
+         * <p>Most Azure APIs require a specific subscription to be selected.</p>
          *
          * @param subscriptionId the ID of the subscription
          * @return an authenticated Azure client configured to work with the specified subscription
@@ -219,7 +229,7 @@ public final class Azure {
          *
          * <p>The default subscription can be specified inside the Azure profile using {@link
          * AzureProfile}. If no default subscription has been previously provided, the first subscription as
-         * returned by {@link Authenticated#subscriptions()} will be selected.
+         * returned by {@link Authenticated#subscriptions()} will be selected.</p>
          *
          * @return an authenticated Azure client configured to work with the default subscription
          */
@@ -297,6 +307,12 @@ public final class Azure {
         @Override
         public SdkContext sdkContext() {
             return this.sdkContext;
+        }
+
+        @Override
+        public Authenticated withTenantId(String tenantId) {
+            profile.withTenantId(tenantId);
+            return this;
         }
 
         @Override
--- a/sdk/management/samples/src/main/java/com/azure/management/containerregistry/samples/ManageContainerRegistryWithWebhooks.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/containerregistry/samples/ManageContainerRegistryWithWebhooks.java
@@ -204,7 +204,7 @@ public class ManageContainerRegistryWithWebhooks {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/AzureConfigurable.java
+++ b/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/AzureConfigurable.java
@@ -11,6 +11,8 @@ import com.azure.core.http.policy.HttpPipelinePolicy;
 import com.azure.core.http.policy.RetryPolicy;
 import com.azure.core.util.Configuration;
 
+import java.util.List;
+
 /**
  * The base interface for allowing configurations to be made on the HTTP client.
  *
@@ -48,10 +50,18 @@ public interface AzureConfigurable<T extends AzureConfigurable<T>> {
     /**
      * Set the cross-tenant auxiliary credentials for Azure which can hold up to three.
      *
-     * @param tokens the AzureTokenCredential list
+     * @param token the auxiliary credential
      * @return the configurable object itself
      */
-    T withAuxiliaryCredentials(TokenCredential... tokens);
+    T withAuxiliaryCredential(TokenCredential token);
+
+    /**
+     * Set the cross-tenant auxiliary credentials for Azure which can hold up to three.
+     *
+     * @param tokens the auxiliary credentials
+     * @return the configurable object itself
+     */
+    T withAuxiliaryCredentials(List<TokenCredential> tokens);
 
     /**
      * Sets the retry policy used in http pipeline.
@@ -68,6 +78,14 @@ public interface AzureConfigurable<T extends AzureConfigurable<T>> {
      * @return the configurable object itself for chaining
      */
     T withScope(String scope);
+
+    /**
+     * Sets the credential scopes.
+     *
+     * @param scopes the credential scope
+     * @return the configurable object itself for chaining
+     */
+    T withScopes(List<String> scopes);
 
     /**
      * Sets the http client.
--- a/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/implementation/AzureConfigurableImpl.java
+++ b/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/implementation/AzureConfigurableImpl.java
@@ -34,11 +34,12 @@ public class AzureConfigurableImpl<T extends AzureConfigurable<T>>
     private List<String> scopes;
     private RetryPolicy retryPolicy;
     private Configuration configuration;
-    private TokenCredential[] tokens;
+    private List<TokenCredential> tokens;
 
     protected AzureConfigurableImpl() {
         policies = new ArrayList<>();
         scopes = new ArrayList<>();
+        tokens = new ArrayList<>();
         retryPolicy = new RetryPolicy();
         httpLogOptions = new HttpLogOptions().setLogLevel(HttpLogDetailLevel.NONE);
     }
@@ -65,9 +66,16 @@ public class AzureConfigurableImpl<T extends AzureConfigurable<T>>
     }
 
     @Override
-    public T withAuxiliaryCredentials(TokenCredential... tokens) {
+    public T withAuxiliaryCredential(TokenCredential token) {
+        Objects.requireNonNull(token);
+        this.tokens.add(token);
+        return (T) this;
+    }
+
+    @Override
+    public T withAuxiliaryCredentials(List<TokenCredential> tokens) {
         Objects.requireNonNull(tokens);
-        this.tokens = tokens;
+        this.tokens.addAll(tokens);
         return (T) this;
     }
 
@@ -82,6 +90,13 @@ public class AzureConfigurableImpl<T extends AzureConfigurable<T>>
     public T withScope(String scope) {
         Objects.requireNonNull(scope);
         this.scopes.add(scope);
+        return (T) this;
+    }
+
+    @Override
+    public T withScopes(List<String> scopes) {
+        Objects.requireNonNull(scopes);
+        this.scopes.addAll(scopes);
         return (T) this;
     }
 
@@ -101,8 +116,9 @@ public class AzureConfigurableImpl<T extends AzureConfigurable<T>>
 
     protected HttpPipeline buildHttpPipeline(TokenCredential credential, AzureProfile profile) {
         Objects.requireNonNull(credential);
-        if (tokens != null) {
-            policies.add(new AuxiliaryAuthenticationPolicy(profile.environment(), tokens));
+        if (!tokens.isEmpty()) {
+            policies.add(
+                new AuxiliaryAuthenticationPolicy(profile.environment(), tokens.toArray(new TokenCredential[0])));
         }
         return HttpPipelineProvider.buildHttpPipeline(credential, profile, scopes(), httpLogOptions, configuration,
             retryPolicy, policies, httpClient);
--- a/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
+++ b/sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
@@ -15,7 +15,7 @@ public class AzureProfile {
 
     private String tenantId;
     private String subscriptionId;
-    private AzureEnvironment environment;
+    private final AzureEnvironment environment;
 
     /**
      * Creates AzureProfile instance with Azure environment. The global environment is {@link AzureEnvironment#AZURE}.
@@ -31,8 +31,9 @@ public class AzureProfile {
      */
     public AzureProfile(AzureEnvironment environment, boolean loadEnvironmentVariables) {
         Objects.requireNonNull(environment);
+        this.environment = environment;
         if (loadEnvironmentVariables) {
-            Configuration configuration = Configuration.getGlobalConfiguration().clone();
+            Configuration configuration = Configuration.getGlobalConfiguration();
             this.tenantId = configuration.get(Configuration.PROPERTY_AZURE_TENANT_ID);
             this.subscriptionId = configuration.get(Configuration.PROPERTY_AZURE_SUBSCRIPTION_ID);
         }
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppBasic.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppBasic.java
@@ -156,7 +156,7 @@ public final class ManageFunctionAppBasic {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppLogs.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppLogs.java
@@ -165,7 +165,7 @@ public final class ManageFunctionAppLogs {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppSourceControl.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppSourceControl.java
@@ -269,7 +269,7 @@ public final class ManageFunctionAppSourceControl {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppWithAuthentication.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppWithAuthentication.java
@@ -219,7 +219,7 @@ public final class ManageFunctionAppWithAuthentication {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppWithDomainSsl.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageFunctionAppWithDomainSsl.java
@@ -187,7 +187,7 @@ public final class ManageFunctionAppWithDomainSsl {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxFunctionAppSourceControl.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxFunctionAppSourceControl.java
@@ -148,7 +148,7 @@ public class ManageLinuxFunctionAppSourceControl {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppBasic.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppBasic.java
@@ -172,7 +172,7 @@ public final class ManageLinuxWebAppBasic {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppCosmosDbByMsi.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppCosmosDbByMsi.java
@@ -194,7 +194,7 @@ public final class ManageLinuxWebAppCosmosDbByMsi {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppSourceControl.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppSourceControl.java
@@ -210,7 +210,7 @@ public final class ManageLinuxWebAppSourceControl {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppSqlConnection.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppSqlConnection.java
@@ -141,7 +141,7 @@ public final class ManageLinuxWebAppSqlConnection {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppStorageAccountConnection.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppStorageAccountConnection.java
@@ -149,7 +149,7 @@ public final class ManageLinuxWebAppStorageAccountConnection {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
             final TokenCredential credential = new DefaultAzureCredentialBuilder()
                 .authorityHost(profile.environment().getActiveDirectoryEndpoint())
                 .build();
--- a/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppWithContainerRegistry.java
+++ b/sdk/management/samples/src/main/java/com/azure/management/appservice/samples/ManageLinuxWebAppWithContainerRegistry.java
@@ -182,7 +182,7 @@ public class ManageLinuxWebAppWithContainerRegistry {
             //=============================================================
             // Authenticate
 
-            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE, true);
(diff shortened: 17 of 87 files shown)


sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/implementation/AzureConfigurableImpl.java
```diff
@@ -92,21 +113,49 @@ public T withProxy(Proxy proxy) {
...
+    protected HttpPipeline buildHttpPipeline(TokenCredential credential) {
+        Objects.requireNonNull(credential);
+
+        List<HttpPipelinePolicy> policies = new ArrayList<>();
+        policies.add(new UserAgentPolicy(httpLogOptions, configuration));
+
+        List<HttpPipelinePolicy> retryPolicies = new ArrayList<>();
+        retryPolicies.add(retryPolicy);
+        retryPolicies.add(new AuthenticationPolicy(credential, environment, scopes()));
+        retryPolicies.add(new HttpLoggingPolicy(httpLogOptions));
```
ChenTanyi, line 153: Logging should be the last one. If we want user to see the network log.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/policy/ProviderRegistrationPolicy.java
```diff
@@ -69,13 +67,8 @@ private boolean isResponseSuccessful(HttpResponse response) {
                             if (cloudError != null && MISSING_SUBSCRIPTION_REGISTRATION.equals(cloudError.getCode())) {
                                 String subscriptionId = ResourceUtils.extractFromResourceId(
                                     request.getUrl().getPath(), "subscriptions");
-                                RestClient restClient = new RestClientBuilder()
-                                        .withBaseUrl(String.format("%s://%s",
-                                            request.getUrl().getProtocol(), request.getUrl().getHost()))
-                                        .withCredential(credential)
-                                        .withSerializerAdapter(jacksonAdapter).buildClient();
                                 // TODO: add proxy in rest client
```
ChenTanyi, line 70: I think this todo could be removed, since the proxy is inside http client now.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Azure profile for client.
+ */
+public class AzureProfile {
+
+    private String tenantId;
+    private String subscriptionId;
+    private AzureEnvironment environment;
+    private final Configuration configuration = Configuration.getGlobalConfiguration().clone();
```
ChenTanyi, line 17: Don't clone the global config.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
```diff
@@ -0,0 +1,98 @@
...
+
+package com.azure.management.resources.fluentcore.profile;
+
+import com.azure.core.management.AzureEnvironment;
+import com.azure.core.util.Configuration;
+
+/**
+ * Azure profile for client.
+ */
+public class AzureProfile {
```
ChenTanyi, line 12: Does it need `withEnvironment`?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/utils/HttpPipelineProvider.java
```diff
@@ -0,0 +1,75 @@
...
+     * Creates http pipeline with policies and http client.
+     *
+     * @param policies the policies not required in retry strategy
+     * @param retryPolicies the policies required in retry strategy
+     * @param httpClient the http client
+     * @return the http pipeline
+     */
+    public static HttpPipeline buildHttpPipeline(
+        List<HttpPipelinePolicy> policies,
+        List<HttpPipelinePolicy> retryPolicies,
```
ChenTanyi, line 62: I think this name is little confused. You even add any `retryPolicy` in pipeline. Not sure we need such function or not.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/implementation/ResourceManager.java
```diff
@@ -57,24 +55,19 @@
...
-        return new AuthenticatedImpl(new RestClientBuilder()
-                .withBaseUrl(credential.getEnvironment(), AzureEnvironment.Endpoint.RESOURCE_MANAGER)
-                .withCredential(credential)
-                .withSerializerAdapter(new AzureJacksonAdapter())
-                .withPolicy(new ProviderRegistrationPolicy(credential))
-                .withPolicy(new ResourceManagerThrottlingPolicy())
-                .buildClient());
+    public static ResourceManager.Authenticated authenticate(TokenCredential credential) {
+        return new AuthenticatedImpl(HttpPipelineProvider.buildHttpPipeline(credential),
+            new AzureProfile(AzureEnvironment.AZURE));
```
ChenTanyi, line 60: Most of user will use this simple `authenticate`, I think maybe `AzureProfile` is needed. Not sure whether they need to set `tenantId` or not.

sdk/resources/mgmt/src/test/java/com/azure/management/resources/core/TestBase.java
```diff
@@ -213,30 +240,39 @@ public void write(int b) throws IOException {
...
+            HttpLogOptions httpLogOptions = new HttpLogOptions().setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS);
+            List<HttpPipelinePolicy> policies = new ArrayList<>();
+            policies.add(new UserAgentPolicy(httpLogOptions, null));
+            List<HttpPipelinePolicy> retryPolicies = new ArrayList<>();
+            retryPolicies.add(new RetryPolicy());
+            retryPolicies.add(new ResourceGroupTaggingPolicy());
+            retryPolicies.add(new TimeoutPolicy(Duration.ofMinutes(1)));
+            retryPolicies.add(new CookiePolicy());
+            retryPolicies.add(new AuthenticationPolicy(credential, profile.environment(), null));
+            retryPolicies.add(new HttpLoggingPolicy(httpLogOptions));
```
ChenTanyi, line 262: I think pipeline builder is more clear

sdk/resources/mgmt/src/main/java/com/azure/management/UserAgentPolicy.java
```diff
@@ -63,7 +63,7 @@ public UserAgentPolicy(HttpLogOptions httpLogOptions, Configuration configuratio
         }
 
         if (configuration == null) {
-            this.configuration = Configuration.getGlobalConfiguration();
+            this.configuration = Configuration.getGlobalConfiguration().clone();
```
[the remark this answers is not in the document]
xccc-msft: Okay, make sense.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/utils/HttpPipelineProvider.java
xccc-msft: This is mainly about how we handle the `HttpPolicyProviders.addBeforeRetryPolicies` and `HttpPolicyProviders.addAfterRetryPolicies`. We will have policies required to add before RetryPolicies as per my understanding, which is existing behavior before refactoring.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/implementation/ResourceManager.java
xccc-msft: Agree with this, I will change it to apply `AzureProfile` together

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
xccc-msft: Hmmm looks withEnvironment can be removed after I checked. The environment is prerequisite before creating `Authenticated` instance.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/utils/HttpPipelineProvider.java
ChenTanyi: The mainly concern is what you want to do by this function. If you want customer to use, I think the pipeline builder will be more clear than such a function. If you only want to use it internally, I think the parameters should be separated.

sdk/resources/mgmt/src/main/java/com/azure/management/UserAgentPolicy.java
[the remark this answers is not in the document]
xccc-msft: This might be a question whether we support the users to change configuration when running the client. It's not allowed by credential from Azure Identity. Maybe we should consider to align with the same rule.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/implementation/AzureConfigurableImpl.java
```diff
@@ -24,33 +32,46 @@
...
-        this.restClientBuilder = this.restClientBuilder.withPolicy(policy);
+        Objects.requireNonNull(policy);
+        policies.add(policy);
         return (T) this;
     }
 
     @Override
-    public T withAuxiliaryCredentials(AzureTokenCredential... tokens) {
+    public T withAuxiliaryCredentials(TokenCredential... tokens) {
         return null;
```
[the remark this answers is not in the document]
xccc-msft: Added.

sdk/resources/mgmt/src/test/java/com/azure/management/resources/core/AuthFile.java
```diff
@@ -0,0 +1,227 @@
...
+package com.azure.management.resources.core;
+
+import com.azure.core.credential.TokenCredential;
+import com.azure.core.implementation.TypeUtil;
+import com.azure.core.management.AzureEnvironment;
+import com.azure.core.management.serializer.AzureJacksonAdapter;
+import com.azure.core.util.logging.ClientLogger;
+import com.azure.core.util.serializer.SerializerAdapter;
+import com.azure.core.util.serializer.SerializerEncoding;
+import com.azure.identity.ClientCertificateCredentialBuilder;
```
yungezz, line 13: is this necessary?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/AzureConfigurable.java
```diff
@@ -96,4 +99,28 @@
...
      */
     T withProxy(Proxy proxy);
+
+    /**
+     * Sets the credential scope.
+     *
+     * @param scope the credential scope
+     * @return the configurable object itself for chaining
+     */
+    T withScope(String scope);
```
yungezz, line 109: what's this?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/implementation/AzureConfigurableImpl.java
```diff
@@ -92,21 +114,45 @@ public T withProxy(Proxy proxy) {
...
-//                .withPolicy(new ProviderRegistrationPolicy())
-//                .withPolicy(new ResourceManagerThrottlingPolicy())
-                .buildClient();
-        // TODO: Add proxy support
-//        if (client.httpClient().proxy() != null) {
-//            credentials.withProxy(client.httpClient().proxy());
-//        }
-        return client;
+    @Override
+    public T withScope(String scope) {
```
yungezz, line 118: what's scope used for

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/implementation/ManagerBase.java
```diff
@@ -14,43 +16,52 @@
...
-        this.restClient = restClient;
-        if (restClient != null) {
-            this.resourceManager = ResourceManager.authenticate(restClient)
+    protected ManagerBase(HttpPipeline httpPipeline, AzureProfile profile, SdkContext sdkContext) {
+        this.httpPipeline = httpPipeline;
+        if (httpPipeline != null) {
+            this.resourceManager = ResourceManager.authenticate(httpPipeline, profile)
                     .withSdkContext(sdkContext)
-                    .withSubscription(subscriptionId);
+                    .withDefaultSubscription();
```
yungezz, line 28: why resourcemanager use default subs, not profile.subsid?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/policy/AuxiliaryAuthenticationPolicy.java
```diff
@@ -30,14 +31,17 @@
...
+    private final AzureEnvironment environment;
 
     /**
      * Initialize an auxiliary authentication policy with the list of AzureTokenCredentials.
      *
+     * @param environment the Azure environment
      * @param credentials the AzureTokenCredentials list
      */
-    public AuxiliaryAuthenticationPolicy(AzureTokenCredential... credentials) {
+    public AuxiliaryAuthenticationPolicy(AzureEnvironment environment, TokenCredential... credentials) {
```
yungezz, line 43: what's this class used for? I think we don't need this anymore

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/policy/ProviderRegistrationPolicy.java
```diff
@@ -67,16 +69,9 @@ private boolean isResponseSuccessful(HttpResponse response) {
...
-                                ResourceManager resourceManager = ResourceManager.authenticate(restClient)
-                                        .withSubscription(subscriptionId);
+                                ResourceManager resourceManager = ResourceManager.authenticate(credential, profile)
+                                        .withDefaultSubscription();
```
yungezz, line 74: same as above, if profile, why defaultsubs? what's defaultsubs?

sdk/resources/mgmt/src/main/java/com/azure/management/AzureTokenCredential.java
```diff
@@ -1,94 +1,94 @@
...
-
-
-    /**
-     * @return the proxy being used for accessing Active Directory.
-     */
-    public Proxy getProxy() {
-        return proxy;
-    }
-}
+//
```
yungezz, line 3: this file will be removed right?

sdk/resources/mgmt/src/main/java/com/azure/management/ApplicationTokenCredential.java
```diff
@@ -1,140 +1,140 @@
...
-        String digest = String.join(" ", scopes);
-
-        Function<String, SimpleTokenCache> computeSimpleTokenCache = key ->
-                new SimpleTokenCache(() -> clientSecretCredential.getToken(request));
-
-        return Mono.just(cache.computeIfAbsent(digest, computeSimpleTokenCache))
-                .flatMap(simpleTokenCache -> simpleTokenCache.getToken());
-    }
-}
+//
```
yungezz, line 3: this file will be removed right

yungezz:
thanks @xccc-msft for the change. A minor suggestion, it will be great to separate the change into 2 prs. one for core change, then referencing to code change.

sdk/resources/mgmt/src/test/java/com/azure/management/resources/core/AuthFile.java
xccc-msft: AuthFile is used to  support file parsing.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/AzureConfigurable.java
xccc-msft: The credential scope can be used in AuthenticationPolicy. @ChenTanyi should be able to suggest whether we plan to keep this.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/implementation/ManagerBase.java
xccc-msft: This is the step to create instance of `ResourceManager` rather than instance of `ResourceManager.Authenticated`. It will take subscription id from profile.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/policy/AuxiliaryAuthenticationPolicy.java
xccc-msft: This is the policy used for [cross tenant authentication](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/authenticate-multi-tenant)

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/policy/ProviderRegistrationPolicy.java
xccc-msft: At first step, the users can create instance for `ResourceManager.Authenticated` without subscription id. Next step, they can use `withSubscription(id)` pass subs id into profile. In another case, they can use `withDefaultSubscription()` if profile already contains subs id.

sdk/resources/mgmt/src/main/java/com/azure/management/AzureTokenCredential.java
xccc-msft: I will remove all redundant before merging.

xccc-msft:
> thanks @xccc-msft for the change. A minor suggestion, it will be great to separate the change into 2 prs. one for core change, then referencing to code change.

The first two commits are core change. And the rest are code changes after core change. If we seperate into 2 PRs, the concern is CI failure according to compilation.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/utils/HttpPipelineProvider.java
xccc-msft: The purpose for this provider is to reuse these pieces of code when creating internal http pipeline. Maybe I should make them private.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/AzureConfigurable.java
weidongxu-microsoft: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-scheme-object

Currently we are using hardcoded `/.default`

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
```diff
@@ -0,0 +1,98 @@
...
+    private String subscriptionId;
+    private AzureEnvironment environment;
+    private final Configuration configuration = Configuration.getGlobalConfiguration();
+
+    /**
+     * Creates AzureProfile instance with Azure environment.
+     *
+     * @param environment the Azure environment
+     */
+    public AzureProfile(AzureEnvironment environment) {
```
weidongxu-microsoft, line 24: Maybe improve to either 1. document in javadoc that AzureEnvironment.AZURE can be used as azure global; 2. or an overloaded constructor

Prefer `Objects.requireNonNull`

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Gets Azure environment.
+     *
+     * @return the Azure environment
+     */
+    public AzureEnvironment environment() {
+        if (this.environment == null) {
+            this.environment = AzureEnvironment.AZURE;
+        }
```
weidongxu-microsoft, line 95: Prefer not to have a `null` environment from start.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Azure profile for client.
+ */
+public class AzureProfile {
+
+    private String tenantId;
+    private String subscriptionId;
+    private AzureEnvironment environment;
+    private final Configuration configuration = Configuration.getGlobalConfiguration();
```
weidongxu-microsoft, line 17: Prefer this be settable, so user can choose to use or not use the configuration.

In this code they do not have the choice at all.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
weidongxu-microsoft: Should we use a Builder to make this class immutable? Do we expect user be able to change `subscriptionId` or `tenantId` in the middle of running the code.

sdk/appservice/mgmt/src/main/java/com/azure/management/appservice/implementation/FunctionAppImpl.java
```diff
@@ -92,8 +96,9 @@
         SiteLogsConfigInner logConfig,
         AppServiceManager manager) {
         super(name, innerObject, siteConfig, logConfig, manager);
-        functionAppKeyServiceHost = manager.restClient().getBaseUrl().toString();
-        functionAppKeyService = RestProxy.create(FunctionAppKeyService.class, manager.restClient().getHttpPipeline());
+        // TODO: confirm host url
```
weidongxu-microsoft, line 99: host url should be correct, same resource manager endpoint..

Prefer re-run some live appservice sample tests before merge.

sdk/appservice/mgmt/src/main/java/com/azure/management/appservice/implementation/FunctionAppImpl.java
```diff
@@ -111,16 +116,20 @@ private void initializeFunctionService() {
...
+            for (int i = 0, count = manager().httpPipeline().getPolicyCount(); i < count; ++i) {
+                policies.add(manager().httpPipeline().getPolicy(i));
+            }
+            policies.add(new FunctionAuthenticationPolicy(this));
+            HttpPipeline httpPipeline= new HttpPipelineBuilder()
+                .policies(policies.toArray(new HttpPipelinePolicy[0]))
+                .httpClient(manager().httpPipeline().getHttpClient())
+                .build();
+            //TODO: confirm host url
+            functionServiceHost = manager().environment().getResourceManagerEndpoint();
```
weidongxu-microsoft, line 130: This is not correct. Should be the `baseUrl` above.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
xccc-msft: Agree, I will prefer javadoc.

sdk/resources/mgmt/src/main/java/com/azure/management/UserAgentPolicy.java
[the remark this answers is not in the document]
ChenTanyi: I think configuration should be decided by user no matter for any purpose.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/implementation/AzureConfigurableImpl.java
```diff
@@ -92,21 +114,45 @@ public T withProxy(Proxy proxy) {
...
+    }
+
+    @Override
+    public T withConfiguration(Configuration configuration) {
+        Objects.requireNonNull(configuration);
+        this.configuration = configuration;
+        return (T) this;
+    }
+
+    protected HttpPipeline buildHttpPipeline(TokenCredential credential, AzureProfile profile) {
```
weidongxu-microsoft, line 138: Does this need to be in sync with `HttpPipelineProvider.buildHttpPipeline(TokenCredential credential, AzureProfile profile)`?

`ProviderRegistrationPolicy` and `ResourceManagerThrottlingPolicy` seems not here?

xccc-msft: This is a TODO item I marked in provider class. I will discuss with @ChenTanyi for the basic policies we should provide if the users don't make customization.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
xccc-msft: I prefer to have boolean flag/method like `allowGobalConfiguration` if we want to make it settable. I was thinking about to have a builder class, just not sure if we should apply the style of  `with*()` to builder class here.

sdk/management/azure/src/main/java/com/azure/management/Azure.java
```diff
@@ -124,76 +123,69 @@
...
+//     *                        client= #client id<br>
+//     *                        key= #client key<br>
+//     *                        managementURI= #management URI<br>
+//     *                        baseURL= #base URL<br>
+//     *                        authURL= #authentication URL<br>
+//     *                        </code>
+//     * @return authenticated Azure client
+//     * @throws IOException exception thrown from file access
+//     */
+//    public static Authenticated authenticate(File credentialFile) throws IOException {
```
weidongxu-microsoft, line 153: You can delete it.

sdk/management/azure/src/main/java/com/azure/management/Azure.java
```diff
@@ -384,51 +369,43 @@ public SdkContext sdkContext() {
...
-            } else {
-                PagedIterable<Subscription> subs = this.subscriptions().list();
-                if (subs.iterator().hasNext()) {
-                    return withSubscription(subs.iterator().next().subscriptionId());
-                } else {
-                    return withSubscription(null);
-                }
-            }
+        public Azure withDefaultSubscription() throws CloudException {
+            return new Azure(httpPipeline, profile, this);
```
weidongxu-microsoft, line 378: Maybe do a throw if `subscriptionId` is null in profile? If user really want to provide a null `subscriptionId`, we would like them to do it explicitly in `withSubscription(String subscriptionId)`.

Also when will `CloudException` happen? It is scheduled to be deleted from core-mgmt.

sdk/appservice/mgmt/src/main/java/com/azure/management/appservice/implementation/FunctionAppImpl.java
xccc-msft: For my understanding, it will be resource manager endpoint too as per the flow we start from `TestBase` to `AppServiceManager` to `FunctionAppImpl`. It seems there isn't any code snippet to change baseUrl after we set resouce manager endpoint initially.

weidongxu-microsoft: I wrote this so I am very sure. It is not mgmt client. It is like kudu client, host of which is `xxx.azurewebsite.net`

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/utils/Utils.java
```diff
@@ -193,34 +188,34 @@ public static void removeFromList(List<String> list, String value) {
...
-        return environment;
-    }
+//    /**
+//     * Try to extract the environment the client is authenticated to based
+//     * on the information on the rest client.
+//     *
+//     * @param restClient the RestClient instance
+//     * @return the non-null AzureEnvironment
+//     */
+//    public static AzureEnvironment extractAzureEnvironment(RestClient restClient) {
```
weidongxu-microsoft, line 198: Just delete it. You can have a backup on your forked repo if needed.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
xccc-msft: The user might want to set `subscriptionId` or `tenantId` after listing subs/tenants via `Authenticated`. Inside each manager, we just keep the `subscriptionId` and `environment` instead of profile. Profile is only used in the constructor of manager.

weidongxu-microsoft: That is method inside AzureConfigure, which does not related to immutable or immutable of this class. You can always create a new AzureProfile there, like what you did to `HttpPipeline`, which is immutable.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
weidongxu-microsoft: Update?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
```diff
@@ -20,6 +20,12 @@
 
     /**
      * Creates AzureProfile instance with Azure environment. The global environment is {@link AzureEnvironment#AZURE}.
+     * The tenant ID and subscription ID can be set via environment variables. The environment variables are expected
```
xccc-msft, line 23: @weidongxu-microsoft The `configuration` is used when the user might prefer [EnvironmentCredential](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/identity/azure-identity/src/main/java/com/azure/identity/EnvironmentCredential.java#L48). Instead of making configuration settable, I prefer following the way environment credential handle configuration. At least for credential related, the user only need to set environment variables in their system.

weidongxu-microsoft: Like I said, I prefer user knows when configure from env is used, when it is not. That is the reason of the existence of `EnvironmentCredential` class at first place. It let user use the credential from env consciously.

sdk/management/samples/src/main/java/com/azure/management/containerregistry/samples/ManageContainerRegistryWithWebhooks.java
```diff
@@ -201,11 +204,15 @@ public static void main(String[] args) {
             //=============================================================
             // Authenticate
 
-            final File credFile = new File(System.getenv("AZURE_AUTH_LOCATION"));
+            final AzureProfile profile = new AzureProfile(AzureEnvironment.AZURE);
```
ChenTanyi, line 207: Where do you add subId and tenantId in samples?

sdk/management/azure/src/main/java/com/azure/management/Azure.java
```diff
@@ -384,51 +301,47 @@ public SdkContext sdkContext() {
...
-        public Azure withDefaultSubscription() throws CloudException, IOException {
-            if (this.defaultSubscription != null) {
-                return withSubscription(this.defaultSubscription);
-            } else {
-                PagedIterable<Subscription> subs = this.subscriptions().list();
-                if (subs.iterator().hasNext()) {
-                    return withSubscription(subs.iterator().next().subscriptionId());
-                } else {
-                    return withSubscription(null);
-                }
```
ChenTanyi, line 400: Keep this in order for default subs. Meanwhile add a with default tenant like it. Since these two could be null in profile. I think it more friendly than all need user to add.

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/arm/implementation/AzureConfigurableImpl.java
```diff
@@ -23,90 +27,88 @@
...
-//        }
-//        return (T) this;
-//    }
-
-    @Override
-    public T withUserAgent(String userAgent) {
-        this.restClientBuilder = this.restClientBuilder.withUserAgent(userAgent);
+    public T withAuxiliaryCredentials(TokenCredential... tokens) {
+        Objects.requireNonNull(tokens);
+        this.tokens = tokens;
```
ChenTanyi, line 70: I think addAll may be more reasonable? Like scopes

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/profile/AzureProfile.java
```diff
@@ -28,9 +27,15 @@
...
      *
      * @param environment the Azure environment
+     * @param loadEnvironmentVariables the boolean flag indicates whether the environment variables are set
      */
-    public AzureProfile(AzureEnvironment environment) {
-        this(null, null, environment);
+    public AzureProfile(AzureEnvironment environment, boolean loadEnvironmentVariables) {
+        Objects.requireNonNull(environment);
+        if (loadEnvironmentVariables) {
+            Configuration configuration = Configuration.getGlobalConfiguration().clone();
```
weidongxu-microsoft, line 35: Seems no need to clone it?

sdk/resources/mgmt/src/main/java/com/azure/management/resources/fluentcore/policy/ProviderRegistrationPolicy.java
ChenTanyi: Why don't remove it?

xccc-msft:
> Please run a few function app live samples before merge.
> 
> Also try spotbugs on resources before merge.

I plan to run after adopting the http client changes applying to dataplane clients. There are appservice samples involved too. What's your comments?

sdk/management/azure/src/main/java/com/azure/management/Azure.java
weidongxu-microsoft: Personally I kind of against this code. It might just pick a wrong subscription id and mess things up.

ChenTanyi: em, I'm not sure how many user will use only one subs for every credential. Like me, my auth will always be only one subs.

Ok to remove it as I think it may be many beginner use the code in sample and will not set subs when call the default subs.

weidongxu-microsoft:
> > Please run a few function app live samples before merge.
> > Also try spotbugs on resources before merge.
> 
> I plan to run after adopting the http client changes applying to dataplane clients. There are appservice samples involved too. What's your comments?

sure

weidongxu-microsoft:
> It seems you read subsId, tenantId from env. Then should we keep a single parameter `TokenCredential` for custom in azure. Like what I wrote in design doc, other parameter are all optional or default.

Personally I against it. E.g. if coder has no idea on which azureEnvironment, tenentId or subscriptionId they are operating upon, they should not do anything. For credential, they still need to clearly wrote `DefaultAzureCredential` if they want to use default option, instead of not providing anything. I would vote similar for `AzureProfile`, that they need to know they are using default.

sdk/management/azure/src/main/java/com/azure/management/Azure.java
weidongxu-microsoft: Yes, file usually only have one, but that is already coded in that file (and we kind of deprecate file for now). Our tenant actually got 10+ subIds.

Maybe we can add e.g. `withSingleSubscription` if requirement arise (but still prefer only do this if there is only 1 subscriptionId avaliable, hence the mock name).

ChenTanyi: No, My auth will always list one subs from azure, since I only add one subs permission to one service principal, it will not list other subs. 

I think we could use `withProfileSubscription` and `withFirstSubscriptionFetched` to distinguish if we want. Sometimes the `withDefaultSubscription` is a little strange.

PR #10500, opened 2020-04-25
Field builder and Field annotation.

sima-zhu:
Initial version of Field builder.

Will add more test cases and comments.


sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldIgnore.java
```diff
@@ -0,0 +1,14 @@
+package com.azure.search.annotation;
```
conniey, line 1: Is this auto-generated? Trying to understand why we would use this annotation without any documentation.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldProperty.java
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
conniey, line 1: Some documentation for this would be nice?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
```diff
@@ -0,0 +1,25 @@
+package com.azure.search.annotation;
```
conniey, line 1: Curious why this doesn't have default values like the FieldProperty?

conniey: Why doesn't this extend from the one above?

conniey: Looking at the properties on FieldProperty, couldn't we just use "isSearchable" to be true? I don't see the extra value in adding this annotation.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SimpleFieldProperty.java
```diff
@@ -0,0 +1,21 @@
+package com.azure.search.annotation;
```
conniey, line 1: Same here... Not sure of the value.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    private static void buildField(Field searchField, FieldProperty annotation) {
+        searchField.setSearchable(annotation.isSearchable());
+        searchField.setSortable(annotation.isSortable());
+        searchField.setFilterable(annotation.isFilterable());
+        searchField.setFacetable(annotation.isFacetable());
+        searchField.setKey(annotation.isKey());
+        if (!"null".equals(annotation.analyzer())) {
```
conniey, line 108: This should be a string constant "null". But why is the default "null" rather than just a null object?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldIgnore.java
```diff
@@ -0,0 +1,14 @@
...
+import com.fasterxml.jackson.annotation.JacksonAnnotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface FieldIgnore {
```
heaths, line 12: What about `IgnoreField`? Not only does it read better, but fits word order better (comments below).

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldProperty.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Retention(RUNTIME)
+@Target({FIELD})
+public @interface FieldProperty {
```
heaths, line 14: I thought we had agreed on using attribute names and properties that match the convenience fields, e.g. `SimpleField`, `SearchableField`, and `ComplexField`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SimpleFieldProperty.java
```diff
@@ -0,0 +1,21 @@
...
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SimpleFieldProperty {
+    boolean isKey();
+    boolean isRetrievable();
+    boolean isFacetable();
+    boolean isSearchable();
+    boolean isSortable();
+    boolean isFilterable();
+    String analyzer();
```
heaths, line 17: Some of these can only be set on `SearchableField` and shouldn't appear here. Make sure they don't in the convenience classes as well.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+                    buildSimpleField(searchField, (SimpleFieldProperty) annotation);
+                    searchFieldList.add(searchField);
+                } else if (annotation instanceof SearchableFieldProperty) {
+                    buildSearchableField(searchField, (SearchableFieldProperty) annotation);
+                    searchFieldList.add(searchField);
+                } else if (annotation instanceof FieldProperty) {
+                    buildField(searchField, (FieldProperty) annotation);
+                    searchFieldList.add(searchField);
+                }
+                logger.info("The property {} is not a search field.", field.getName());
```
heaths, line 56: What about `ComplexField`? Just not implemented yet?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+        return searchFieldList;
+    }
+
+    private static void buildSimpleField(Field searchField, SimpleFieldProperty annotation) {
+        searchField.setSearchable(true);
+        searchField.setSortable(true);
+        searchField.setFilterable(true);
+        searchField.setFacetable(true);
+        searchField.setKey(annotation.isKey());
+        if (!"null".equals(annotation.analyzer())) {
```
heaths, line 68: Mentioned above, but just to reiterate here (and applies below): analyzers, synonymMaps, etc., are only settable on searchable fields and shouldn't be here.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexDataType.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.azure.search.documents.models;
+
+import com.azure.core.util.ExpandableStringEnum;
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+/**
+ * Simple DataType is used to create SimpleSearchField.
+ */
+public class ComplexDataType extends ExpandableStringEnum<ComplexDataType> {
```
heaths, line 12: Couldn't you just use the same one as for the convenience classes?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexField.java
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.search.documents.models;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * A helper Field model to build a simple search field.
```
heaths, line 10: You mention "search field" here. Copy/paste error?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/SimpleDataType.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.azure.core.util.ExpandableStringEnum;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+/**
+ * Simple DataType is used to create SimpleSearchField.
+ */
+public class SimpleDataType extends ExpandableStringEnum<SimpleDataType> {
```
heaths, line 14: You can't use the same one as for the convenience type? Less chance of breaking across generations.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldIgnore.java
sima-zhu: I am working with search team to get clear about how customer uses the annotation. I will add doc after major workflow works.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/SimpleField.java
```diff
@@ -0,0 +1,76 @@
...
+    public Boolean getFacetable() {
+        return facetable;
+    }
+
+    public SimpleField setFacetable(Boolean facetable) {
+        this.facetable = facetable;
+        return this;
+    }
+
+    public AnalyzerName getAnalyzer() {
```
heaths, line 45: This an others below are not supported on `SimpleField`. Only `SearchableField`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldIgnore.java
sima-zhu: Mimic Jackson package naming. 
They use @JsonIgnore if they don't want to serialize the property.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
brjohnstmsft: This is for the "static" case where we're reflecting over a type, right? In that case, you shouldn't need an attribute to identify complex types -- the type of the property/accessor should indicate that (by not being a primitive, string, or geo-point or collection of those).

heaths: That's implying that non-attributed fields should be reflected and turned into fields. That doesn't seem to follow what track 1 did, though I only took a cursory glance. If we should reflect non-attributed fields, then that means all attributes' properties' values would be default, correct?

brjohnstmsft: @heaths Track 1 reflected all public properties, even those without attributes. One reason is that complex fields don't require any additional attributes (because they already have a name and type). As for defaults, in Track 1 it was the client's defaults (everything disabled by default). In Track 2, that would be a safe default to have, but I'm not sure if that's different than the defaults for explicitly-constructed `SearchFields`.

heaths: Okay, so long as we're agreed that since we're following convenience classes' defaults for these convenience attributes (or lack thereof), then we will have a consistent experience. E.g. if a string field/property has no attributes, it will be equivalent to a `SimpleField` of "Edm.String" and no other traits enabled, including that `IsHidden` is `false` so it will appear in results.

sima-zhu: Thank for comments! Will treat non-attributed field to complex field if it is non-primitive.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: Annotation for string method cannot default to null object. The field is not required, so I made it default to null string.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
sima-zhu: The namespace should go for com.azure.search.documents.indexes.annotation.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+            return arrayOrListType.getClass().getComponentType();
+        }
+        if (isList(arrayOrListType)) {
+            ParameterizedType pt = (ParameterizedType) arrayOrListType;
+            return pt.getActualTypeArguments()[0];
+        }
+        throw logger.logExceptionAsError(new RuntimeException("Should not be there"));
+    }
+
+    private static Field convertToBasicSearchField(java.lang.reflect.Field classField,
```
heaths, line 172: `Field` should be `SearchField` per mode renames.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+        ClientLogger logger) {
+        Field searchField = new Field();
+        searchField.setName(classField.getName());
+        DataType dataType = covertToDataType(classField.getGenericType(), false, logger);
+        searchField.setType(dataType);
+        searchField.setKey(false);
```
heaths, line 178: If these all default to `false`, why set them just to potentially override them later with `enrichWithAnnotations` ?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+            }
+            SearchableFieldProperty searchableFieldPropertyAnnotation =
+                classField.getDeclaredAnnotation(SearchableFieldProperty.class);
+            searchField.setSearchable(true);
+            searchField.setSortable(searchableFieldPropertyAnnotation.isSortable());
+            searchField.setFilterable(searchableFieldPropertyAnnotation.isFilterable());
+            searchField.setFacetable(searchableFieldPropertyAnnotation.isFacetable());
+            searchField.setKey(searchableFieldPropertyAnnotation.isKey());
+            searchField.setHidden(searchableFieldPropertyAnnotation.isHidden());
+            if (!"null".equals(searchableFieldPropertyAnnotation.analyzer())) {
```
heaths, line 220: Should just be a simple `null` check if not defined as a string, right? What if they set it to `null` explicitly when declaring?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexField.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class ComplexField extends FieldBase {
```
heaths, line 12: Don't use `FieldBase` (at least in .NET, "Base" is not allowed for base classes). See the model rename issue for the suggestion (I forget off hand).

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexField.java
```diff
@@ -0,0 +1,55 @@
...
+    public ComplexField(String name, boolean collection) {
+        super(name, collection ? DataType.collection(DataType.EDM_COMPLEX_TYPE) : DataType.EDM_COMPLEX_TYPE);
+    }
+
+    /**
+     * Gets a collection of {@link SimpleField} or {@link ComplexField} child fields.
+     *
+     * @return The list of sub-fields.
+     */
+    public List<Field> getSubFields() {
```
heaths, line 30: Just `getFields`. "Sub" is redundant and atypical.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexField.java
```diff
@@ -0,0 +1,55 @@
...
+        return subFields;
+    }
+
+    /**
+     * Sets a collection of {@link SimpleField} or {@link ComplexField} child fields.
+     *
+     * @param subFields The list of sub-fields.
+     * @return The {@link ComplexField} object itself.
+     */
+    public ComplexField setSubFields(List<Field> subFields) {
```
heaths, line 40: Just `setFields`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/SimpleField.java
```diff
@@ -0,0 +1,150 @@
...
+    private boolean hidden;
+
+    /**
+     * Initializes a new instance of the {@link SimpleField} class.
+     *
+     * @param name The name of the field, which must be unique within the index or parent field.
+     * @param dataType The {@link DataType} of the {@link Field}.
+     * @throws NullPointerException when {@code name} is null.
+     */
+    public SimpleField(String name, DataType dataType) {
```
heaths, line 27: Need an overload for whether this is also a collection (like you did in `ComplexField`).

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/FieldBuilderTest.java
```diff
@@ -0,0 +1,98 @@
...
+    @Test
+    public void hotelListFieldSearchableThrowException() {
+        Exception exception = assertThrows(RuntimeException.class, () -> {
+            FieldBuilder.build(HotelSearchableExceptionOnList.class);
+        });
+        assertTrue(exception.getMessage().contains("passcode"));
+        assertTrue(exception.getMessage().contains(DataType.collection(DataType.EDM_INT32).toString()));
+    }
+
+    private List<Field> buildHotelFieldsFromModel() {
```
heaths, line 51: Nit: wouldn't "build...FromModel" mean using the field builder against a model type? This seems more like manual index creation / defining of fields.

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/models/HotelSearchException.java
```diff
@@ -0,0 +1,33 @@
+package com.azure.search.documents.test.environment.models;
```
heaths, line 1: Might be good to have the cycle a level deeper just to make sure the recursive stack push/check is working correctly (and continues to do so when changed).

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
```diff
@@ -0,0 +1,79 @@
...
+     * @return True if it is not retrievable, and false by default for retrievable field.
+     */
+    boolean isHidden() default false;
+
+    /**
+     * Optional arguments defines whether the field is facetable or not.
+     *
+     * @return True if it is facetable, and false by default for non-facetable field.
+     */
+    boolean isFacetable() default false;
```
alzimmermsft, line 36: Shouldn't this be `true` to match the service documentation?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Optional arguments defines whether the field is sortable or not.
+     *
+     * @return True if it is sortable, and false by default for non-sortable field.
+     */
+    boolean isSortable() default false;
```
alzimmermsft, line 43: Shouldn't this be `true` to match the service documentation?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Optional arguments defines whether the field is filterable or not.
+     *
+     * @return True if it is filterable, and false by default for non-filterable field.
+     */
+    boolean isFilterable() default false;
```
alzimmermsft, line 50: Shouldn't this be `true` to match the service documentation?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
```diff
@@ -0,0 +1,79 @@
...
+     * @return {@link AnalyzerName} String value. Or default to "null" String type.
+     */
+    String analyzer() default "null";
+
+    /**
+     * Optional arguments defines the name of the search analyzer used for the field.
+     *
+     * @return {@link AnalyzerName} String value. Or default to "null" String type.
+     */
+    String searchAnalyzer() default "null";
```
alzimmermsft, line 64: This should be `null` right?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Optional arguments defines the name of the analyzer used for the field.
+     *
+     * @return {@link AnalyzerName} String value. Or default to "null" String type.
+     */
+    String indexAnalyzer() default "null";
```
alzimmermsft, line 71: This should be `null` right?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Optional arguments defines the array of synonymMaps used for the field.
+     *
+     * @return An array of synonym map values. Or default to empty string array.
+     */
+    String[] synonymMaps() default {};
```
alzimmermsft, line 78: This should be `null` right?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SimpleFieldProperty.java
```diff
@@ -0,0 +1,51 @@
...
+     * @return True if it is not retrievable, and false by default for retrievable field.
+     */
+    boolean isHidden() default false;
+
+    /**
+     * Optional arguments defines whether the field is facetable or not.
+     *
+     * @return True if it is facetable, and false by default for non-facetable field.
+     */
+    boolean isFacetable() default false;
```
alzimmermsft, line 36: Shouldn't this be `true` to match the service documentation?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SimpleFieldProperty.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Optional arguments defines whether the field is sortable or not.
+     *
+     * @return True if it is sortable, and false by default for non-sortable field.
+     */
+    boolean isSortable() default false;
```
alzimmermsft, line 43: Shouldn't this be `true` to match the service documentation?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SimpleFieldProperty.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Optional arguments defines whether the field is filterable or not.
+     *
+     * @return True if it is filterable, and false by default for non-filterable field.
+     */
+    boolean isFilterable() default false;
```
alzimmermsft, line 50: Shouldn't this be `true` to match the service documentation?

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/setup/AzureSearchResources.java
```diff
@@ -22,6 +23,11 @@
 import java.io.IOException;
 import java.net.InetAddress;
 import java.security.SecureRandom;
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
```
alzimmermsft, line 29: Are these imports being used?

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/setup/AzureSearchResources.java
```diff
@@ -22,6 +23,11 @@
 import java.io.IOException;
 import java.net.InetAddress;
 import java.security.SecureRandom;
+import java.time.LocalDateTime;
```
alzimmermsft, line 26: Is this import being used?

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/setup/AzureSearchResources.java
```diff
@@ -6,6 +6,7 @@
 import com.azure.core.test.utils.TestResourceNamer;
 import com.azure.core.util.Configuration;
 import com.azure.core.util.CoreUtils;
+import com.azure.core.util.DateTimeRfc1123;
```
alzimmermsft, line 9: Is this import being used?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldIgnore.java
```diff
@@ -0,0 +1,20 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marker annotation that indicates the field is to be ignored by converting to SearchField.
+ */
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface FieldIgnore {
```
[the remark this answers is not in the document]
heaths: Talking with @brjohnstmsft yesterday, it should be opt-out since that's how track 1 works. That said, by *not* attributing/annotating properties, you don't get much by default.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
```diff
@@ -0,0 +1,79 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * The annotation is to indicate whether the field is a searchable field. The boolean field of isSearchable
+ * defaults to true if use the annotation.
+ */
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SearchableFieldProperty {
```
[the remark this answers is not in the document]
heaths: We intentionally default everything to false for consistency.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
heaths: No. We pass false by default. All values default to their defaults (bool false, int 0, etc.).

heaths: Same for similar questions elsewhere.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
sima-zhu: I am worrying customer has trouble to find this. I will move to com.azure.search.documents.indexes first. Will check with UX study

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldIgnore.java
sima-zhu: Do you have any name suggestion?
@IgnoreField ?

sdk/search/azure-search-documents/CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Release History
 
 ## 1.0.0-beta.3 (Unreleased)
-- Change `createOrUpdate*` and `delete*` APIs in `SearchServiceClient` to use boolean `onlyIfUnchanged` instead of `MatchConditioans`.
+- Changed `createOrUpdate*` and `delete*` APIs in `SearchServiceClient` to use boolean `onlyIfUnchanged` instead of `MatchConditioans`.
+- Added helper class `FieldBuilder` which coverts strongly-typed model class to `List<Field>`.
```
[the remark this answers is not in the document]
sima-zhu: This is what I want to address for the next phase. Don't want to involve too many changes within the same PR.

sdk/search/azure-search-documents/CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Release History
 
 ## 1.0.0-beta.3 (Unreleased)
-- Change `createOrUpdate*` and `delete*` APIs in `SearchServiceClient` to use boolean `onlyIfUnchanged` instead of `MatchConditioans`.
+- Changed `createOrUpdate*` and `delete*` APIs in `SearchServiceClient` to use boolean `onlyIfUnchanged` instead of `MatchConditioans`.
+- Added helper class `FieldBuilder` which coverts strongly-typed model class to `List<Field>`. 
+- Added annotations `FieldIgnore`, `SimpleFieldProperty`, `SearchableFieldProperty` to define the `Field` on model properties.
```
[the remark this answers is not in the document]
sima-zhu: Then the convenience class and the annotation will name the same `SimpleField`.
Does it confuse user?

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldIgnore.java
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * Optional argument that defines whether this annotation is active or not.
+     *
+     * @return True if annotation is enabled (by default); false if it is to be ignored.
+     */
+    boolean value() default true;
```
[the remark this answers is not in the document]
sima-zhu: Right. I also have the same thought. Will remove.

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
```diff
@@ -0,0 +1,79 @@
...
+     * @return True if it is filterable, and false by default for non-filterable field.
+     */
+    boolean isFilterable() default false;
+
+    /**
+     * Optional arguments defines the name of the analyzer used for the field.
+     *
+     * @return {@link AnalyzerName} String value. Or default to "null" String type.
+     */
+    String analyzer() default "null";
```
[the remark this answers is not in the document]
sima-zhu: Switch to empty, as Java annotation has no way to default to null object.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Stack;
+import java.util.stream.Collectors;
+
+/**
+ * Helper to covert model class to search field {@link Field}.
+ */
+public class FieldBuilder {
+    private static final int MAX_DEPTH = 10;
```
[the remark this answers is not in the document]
sima-zhu: This is kind of protection to avoid deep calling. We just recommend customer to redesign their index if it is call path is too deep. Like complexField 1-> 2 -> 3 ... -> 10 -> 11 (Fail here). I don't have reference for the number(willing to change). Also, it is just a good to have validation.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: It is checked in recursive call.
```
       List<Field> searchFields = classFields.stream()
            .filter(classField -> !classField.isAnnotationPresent(FieldIgnore.class))
            .map(classField -> buildField(classField, classChain, logger))
            .collect(Collectors.toList());

File: sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java (hunk 16 above)
Comment (sima-zhu) [reply sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java:178]: This is good for some field without annotation. 
e.g.
```
private String id;
```

We convert this to Field with Field (name=id, type=EDM.String, isSortable=true, etc)
This is align with the one with annotation.

File: sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java (hunk 17 above)
Comment (sima-zhu) [reply sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java:219]: Switch to empty

File: sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexField.java (hunk 18 above)
Comment (sima-zhu) [reply sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/ComplexField.java:12]: name and dataType initialization and validation apply to all three `SimpleField`, `SearchableField`, `ComplexField`
Don't want to repeat them again

File: sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/FieldBuilderTest.java (hunk 22 above)
Comment (sima-zhu) [reply sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/FieldBuilderTest.java:54]: Will rename

File: sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java (hunk 42 above)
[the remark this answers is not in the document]
Comment (brjohnstmsft) [reply sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java:31]: Depth validation should be left to the service. The depth check done by the REST API can vary based on pricing tier, so having a hard-coded value here is likely to cause breakage for at least some customers, unless you pick a very large value.

If you're checking for recursive types, then unbounded stack growth isn't possible. If you're just concerned about a very deep tree of other types, pick a much larger value, like 10000, that is likely to always be greater than the REST API's own limits.

File: sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldIgnore.java (hunk 40 above)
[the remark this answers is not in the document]
Comment (sima-zhu) [reply sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/FieldIgnore.java]: Removed.

File: sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/setup/AzureSearchResources.java (hunk 35 above)
Comment (sima-zhu) [reply sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/test/environment/setup/AzureSearchResources.java]: It is not supposed to have any changes in this class. I will try to remove the diff here.

File: sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java [hunk 43]
```diff
@@ -0,0 +1,268 @@
...
+                String.format("There is circular dependencies %s, %s", classChain, curClass)));
+        }
+        if (classChain.size() > MAX_DEPTH) {
+            throw logger.logExceptionAsError(
+                new RuntimeException("The dependency graph is too deep. Please review your schema."));
+        }
+        classChain.push(curClass);
+        List<java.lang.reflect.Field> classFields = Arrays.asList(curClass.getDeclaredFields());
+        List<Field> searchFields = classFields.stream()
+            .filter(classField -> !classField.isAnnotationPresent(FieldIgnore.class))
```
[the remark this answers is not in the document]
sima-zhu: I have removed value setting.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: We have two Field in builder, one is Java reflection, one is our search field. 
Will have rename changes afterwards to make this clear

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+     * Recursive class to build complex data type.
+     *
+     * @param curClass Current class to be built.
+     * @param classChain A class chain from {@code modelClass} to prior of {@code curClass}.
+     * @param logger {@link ClientLogger}.
+     * @return A list of {@link Field} that curClass is built to.
+     */
+    private static List<Field> build(Class<?> curClass, Stack<Class<?>> classChain, ClientLogger logger) {
+        if (classChain.contains(curClass)) {
+            logger.warning(String.format("There is circular dependencies %s, %s", classChain, curClass));
```
alzimmermsft, line 85: Do we want to log a warning for a circular dependency? There are a lot of data structures that will have this, for example a binary tree. Would it be better to just return `null` to break the cycle?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+        Field searchField = new Field();
+        searchField.setName(classField.getName());
+        DataType dataType = covertToDataType(classField.getGenericType(), false, logger);
+        searchField.setType(dataType);
+        searchField.setKey(false);
+        searchField.setSearchable(false);
+        searchField.setFacetable(false);
+        searchField.setHidden(false);
+        searchField.setFilterable(false);
+        searchField.setSortable(false);
```
srnagar, line 177: Use the fluent pattern instead.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Stack;
+import java.util.stream.Collectors;
+
+/**
+ * Helper to covert model class to search field {@link Field}.
+ */
+public class FieldBuilder {
```
srnagar, line 34: Since this class is a collection of static helper methods, it should be `final` and also add a private constructor.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Stack;
+import java.util.stream.Collectors;
+
+/**
+ * Helper to covert model class to search field {@link Field}.
```
srnagar, line 32:
```suggestion
 * Helper to convert model class to search field {@link Field}.
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/package-info.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Package containing the annotations used by FieldBuilder which help covert
```
srnagar, line 5:
```suggestion
 * Package containing the annotations used by FieldBuilder which help convert
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/SearchableField.java
```diff
@@ -0,0 +1,138 @@
...
+     * Convert SearchableField to {@link Field}.
+     *
+     * @return The {@link Field} object.
+     */
+    public Field build() {
+        return new Field().setName(super.getName()).setType(super.getDataType()).setSearchable(true)
+            .setKey(super.isKey()).setSortable(super.isSortable()).setFilterable(super.isFilterable())
+            .setHidden(super.isHidden()).setFacetable(super.isFacetable()).setAnalyzer(this.analyzer)
+            .setSearchAnalyzer(this.searchAnalyzer).setIndexAnalyzer(this.indexAnalyzer)
+            .setSynonymMaps(this.synonymMaps);
```
srnagar, line 136: It might be more readable if you have one setter per line.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: I have returned null below.
```
if (classChain.contains(curClass)) {
    logger.warning(...);
    return null;
}
```

This is decided by team, which we ignore any property of a type that's already been "seen".

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * Creates a collection of {@link Field} objects corresponding to the properties of the type supplied.
+     *
+     * @param modelClass The class for which fields will be created, based on its properties.
+     * @param <T> The generic type of the model class.
+     * @return A collection of fields.
+     */
+    public static <T> List<Field> build(Class<T> modelClass) {
+        ClientLogger logger = new ClientLogger(FieldBuilder.class);
```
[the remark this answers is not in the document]
sima-zhu: This is a helper for building index schema for search. The operation will not happen frequently. Probably only when customer wants to update the index (rarely happen, a breaking change).

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
heaths: No. As discussed with @brjohnstmsft and in other meetings we've had, we want the defaults for model properties and parameters to match the language defaults for those types (e.g. boolean is false). Those should be sent to the service as-is to avoid confusion and force customers to look what the docs currently list as defaults.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+
+
+    private static boolean isArrayOrList(Type type) {
+        if (type.getClass().isArray()) {
+            return true;
+        }
+        return isList(type);
+    }
+
+    private static boolean isList(Type type) {
```
alzimmermsft, line 134: Should this check `Collection` as that is the super type for `List` and will capture `Set`, `Queue`, etc.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+import java.util.Map;
+import java.util.Stack;
+import java.util.stream.Collectors;
+
+/**
+ * Helper to convert model class to search field {@link Field}.
+ */
+public final class FieldBuilder {
+    private static final int MAX_DEPTH = 10000;
+    private static final Map<Class<?>, DataType> SUPPORTED_NONE_PARAMETERIZED_TYPE = new HashMap<>();
```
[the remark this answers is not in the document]
alzimmermsft: After looking at the PR further could we rename this `SUPPORTED_SIMPLE_TYPES`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+        if (isArrayOrList(type)) {
+            return buildCollectionField(classField, classChain, logger);
+        }
+        List<Field> childFields = build((Class<?>) type, classChain, logger);
+        Field searchField = convertToBasicSearchField(classField, logger);
+        searchField.setFields(childFields);
+        return searchField;
+    }
+
+    private static boolean isSupportedNoneParameterizedType(Type type) {
```
alzimmermsft, line 116: Is this helper method needed?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+
+    private static Type getComponentOrElementType(Type arrayOrListType, ClientLogger logger) {
+        if (arrayOrListType.getClass().isArray()) {
+            return arrayOrListType.getClass().getComponentType();
+        }
+        if (isList(arrayOrListType)) {
+            ParameterizedType pt = (ParameterizedType) arrayOrListType;
+            return pt.getActualTypeArguments()[0];
+        }
+        throw logger.logExceptionAsError(new RuntimeException("Should not be there"));
```
alzimmermsft, line 163: Can we add an actual exception message if we ever run into this issue in production.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+            .setFilterable(false)
+            .setSortable(false);
+        return searchField;
+    }
+
+    private static Field enrichWithAnnotation(Field searchField, java.lang.reflect.Field classField,
+        ClientLogger logger) {
+        if (classField.isAnnotationPresent(SimpleFieldProperty.class)
+            && classField.isAnnotationPresent(SearchableFieldProperty.class)) {
+            throw logger.logExceptionAsError(new IllegalArgumentException(
```
alzimmermsft, line 185: Doesn't needed to be changed now, but does this make more sense as an `IllegalStateException`?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+                searchField.setAnalyzer(AnalyzerName.fromString((searchableFieldPropertyAnnotation.analyzer())));
+            }
+            if (!searchableFieldPropertyAnnotation.searchAnalyzer().isEmpty()) {
+                searchField.setAnalyzer(AnalyzerName.fromString((searchableFieldPropertyAnnotation.searchAnalyzer())));
+            }
+            if (!searchableFieldPropertyAnnotation.indexAnalyzer().isEmpty()) {
+                searchField.setAnalyzer(AnalyzerName.fromString((searchableFieldPropertyAnnotation.indexAnalyzer())));
+            }
+            if (searchableFieldPropertyAnnotation.synonymMaps().length != 0) {
+                searchField.setSynonymMaps(Arrays.asList(searchableFieldPropertyAnnotation.synonymMaps()));
```
alzimmermsft, line 223: Should we attempt to cleanse any potential `null` values in the annotation value? Ignore this if Java doesn't allow `{ "aSynonym", null, "anotherSynonym" }`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/FieldBase.java
```diff
@@ -0,0 +1,47 @@
...
+    private final DataType dataType;
+
+    /**
+     * Initializes a new instance of the {@link FieldBase} class.
+     * @param name The name of the field, which must be unique within the index or parent field.
+     * @param dataType The data type of the field.
+     */
+    protected FieldBase(String name, DataType dataType) {
+        if (CoreUtils.isNullOrEmpty(name)) {
+            throw logger.logExceptionAsError(new IllegalArgumentException("The name of field cannot be null"));
```
alzimmermsft, line 24: Should we also validate that `DataType` != `null`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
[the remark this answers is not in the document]
sima-zhu: That's off a bit.
SimpleType includes (primitive, boxing of primitive, array/list of boxing primitive)

What I want to put here is: (primitive, boxing of primitive)

This is java class reflection term.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+    static {
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(Integer.class, DataType.EDM_INT32);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(int.class, DataType.EDM_INT32);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(Long.class, DataType.EDM_INT64);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(long.class, DataType.EDM_INT64);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(Double.class, DataType.EDM_DOUBLE);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(double.class, DataType.EDM_DOUBLE);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(Boolean.class, DataType.EDM_BOOLEAN);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(boolean.class, DataType.EDM_BOOLEAN);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(String.class, DataType.EDM_STRING);
```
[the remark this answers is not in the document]
sima-zhu: First of all, I think we can force people to use String instead of `CharSequence` since I cannot think of why String cannot support customer needs. I would prefer to put it into `UNSUPPORTED_TYPE`.

If customer request to have `CharSequence`, I prefer to put into map directly instead of using `instanceof` as it is not a heavy load and there is not too much such cases in the real world, still constant cost. Adding instanceof add extra cost for most of the cases.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: @brjohnstmsft 

What collection type should we support for `Collection (DataType) ` in search? Does service support `Set`, `Queue` etc? I currently only support user to have array or list as collection type.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: @brjohnstmsft I think this is valid. 
Do you think we need extra step of removing null out of map or throw error here?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: It is used for distinguish whether it is Supported type or complex one. It is necessary to have in top level checking.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: This means user uses both annotation at the same time. This can be fixed by customer.

sima-zhu: `IllegalStateException` sounds like not able to fix by customer.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
brjohnstmsft: @sima-zhu This seems more like a question about what's idiomatic for Java. As long as it serializes to a JSON array, the service can work with it.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
brjohnstmsft: @sima-zhu I don't recall if the service ignores or rejects null in the list of synonym maps. Nulls are no meaningful in that context, so I think it would be fine to ignore them.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: Based on service side only take JSON array, I would leave array/list as the choice of collection DataType. 
First, they are enough for most of the use case.
Moreover, as java sdk is not the only entry point of search service, I don't think it is necessary to leave customer the flexibility of `set`, `queue` to manipulate the data which service cannot maintain.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: Will remove null from the list.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
alzimmermsft: The reason I think it isn't needed as it is checking something we can do without calling another method in a single line, just replacing all instances of this being called with:

```java
SUPPORTED_NONE_PARAMETERIZED_TYPE.containsKey(type);
```

alzimmermsft: Does not need to be removed now.

sima-zhu: Make sense. Will simplify

sdk/search/azure-search-documents/src/main/java/com/azure/search/annotation/SearchableFieldProperty.java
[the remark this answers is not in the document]
sima-zhu: I think the name can retrieve from the attribute name. This is also what I get from track 1. If name is also we want to config here. I can create an issue to track this.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(Date.class, DataType.EDM_DATE_TIME_OFFSET);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(OffsetDateTime.class, DataType.EDM_DATE_TIME_OFFSET);
+        SUPPORTED_NONE_PARAMETERIZED_TYPE.put(GeoPoint.class, DataType.EDM_GEOGRAPHY_POINT);
+    }
+
+    private static final List<Class<?>> UNSUPPORTED_TYPES = Arrays.asList(Byte.class,
```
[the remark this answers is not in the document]
sima-zhu: https://github.com/Azure/azure-sdk-for-java/issues/10647

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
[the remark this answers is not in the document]
sima-zhu: Change it to 10000

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+        }
+        return isList(type);
+    }
+
+    private static boolean isList(Type type) {
+        if (type instanceof ParameterizedType) {
+            Type rawType = ((ParameterizedType) type).getRawType();
+            return List.class.isAssignableFrom((Class<?>) rawType);
+        }
+        return false;
```
conniey, line 136: I'm a fan of bailing out of a method early. I think it makes it easier to read and the actual logic isn't shifted to the right with a ton of spaces if there are nested if-statements. It's a personal preference. You can leave as-is if you want.
```
if (!(type instanceof ParameterizedType)) {
    return false;
}

Type rawType = ((ParameterizedType) type).getRawType();
return List.class.isAssignableFrom((Class<?>) rawType);
```

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+
+    private static void validateType(Class<?> type, boolean hasArrayOrCollectionWrapped, ClientLogger logger) {
+        if (Map.class.isAssignableFrom(type)) {
+            throw logger.logExceptionAsError(new IllegalArgumentException("Map and its subclasses are not supported"));
+        }
+        if (UNSUPPORTED_TYPES.contains(type)) {
+            throw logger.logExceptionAsError(new IllegalArgumentException(String.format("%s is not supported",
+                    type.getName())));
+        }
+        if (Collection.class.isAssignableFrom(type)) {
```
conniey, line 237: I'd invert this if to avoid the nested ifs.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/FieldBase.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+        Objects.requireNonNull(dataType, "DataType cannot be null.");
+        this.name = name;
+        this.dataType = dataType;
```
conniey, line 30: You can save a line by using:
```java
this.dataType = Objects.requireNonNull(dataType, "'dataType' cannot be null.");
```

sdk/search/azure-search-documents/src/test/java/com/azure/search/documents/FieldBuilderTest.java
```diff
@@ -0,0 +1,138 @@
...
+        Field baseRate = new SimpleField("baseRate", DataType.EDM_DOUBLE, false).build();
+        Field bedOptions = new SimpleField("bedOptions", DataType.EDM_STRING, false).build();
+        Field sleepsCount = new SimpleField("sleepsCount", DataType.EDM_INT32, false).build();
+        Field smokingAllowed = new SimpleField("smokingAllowed", DataType.EDM_BOOLEAN, false).build();
+        Field tags = new SimpleField("tags", DataType.EDM_STRING, true).build();
+        return Arrays.asList(description, descriptionFr, type, baseRate, bedOptions, sleepsCount, smokingAllowed, tags);
+    }
+
+    private List<Field> sortByFieldName(List<Field> fields) {
+        return fields.stream().sorted((o1, o2) -> o1.getName().compareTo(o2.getName())).collect(Collectors.toList());
```
conniey, line 136: Have you tried `Collections.sort(fields, (o1, o2) -> {})`? It'll save you from having to turn this from a list to a stream, then back into a list.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @param curClass Current class to be built.
+     * @param classChain A class chain from {@code modelClass} to prior of {@code curClass}.
+     * @param logger {@link ClientLogger}.
+     * @return A list of {@link Field} that curClass is built to.
+     */
+    private static List<Field> build(Class<?> curClass, Stack<Class<?>> classChain, ClientLogger logger) {
+        if (classChain.contains(curClass)) {
+            logger.warning(String.format("There is circular dependencies %s, %s", classChain, curClass));
+            return null;
```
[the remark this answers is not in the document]
brjohnstmsft: We shouldn’t throw because the user might not intend for all properties of the type to be mapped to fields. Throwing an exception prevents the type from being used at all. Better to ignore properties we can’t handle.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,256 @@
...
+            return pt.getActualTypeArguments()[0];
+        }
+        throw logger.logExceptionAsError(new RuntimeException(String.format(
+            "Collection type %s is not supported.", arrayOrListType.getTypeName())));
+    }
+
+    private static Field convertToBasicSearchField(java.lang.reflect.Field classField,
+        ClientLogger logger) {
+        Field searchField = new Field();
+        searchField.setName(classField.getName());
```
heaths, line 161: The field name needs to be overridable. Imagine a key named "hid" but the field is `hotelId`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,256 @@
...
+                .setHidden(searchableFieldPropertyAnnotation.isHidden());
+            if (!searchableFieldPropertyAnnotation.analyzer().isEmpty()) {
+                searchField.setAnalyzer(AnalyzerName.fromString((searchableFieldPropertyAnnotation.analyzer())));
+            }
+            if (!searchableFieldPropertyAnnotation.searchAnalyzer().isEmpty()) {
+                searchField.setAnalyzer(AnalyzerName.fromString((searchableFieldPropertyAnnotation.searchAnalyzer())));
+            }
+            if (!searchableFieldPropertyAnnotation.indexAnalyzer().isEmpty()) {
+                searchField.setAnalyzer(AnalyzerName.fromString((searchableFieldPropertyAnnotation.indexAnalyzer())));
+            }
```
heaths, line 213: After this, you should assert that only analyzer or both searchAnalyzer and indexAnalyzer are specified.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/SearchableField.java
```diff
@@ -0,0 +1,146 @@
...
+     * Sets a list of names of synonym maps to associate with this field.
+     * Currently, only one synonym map per field is supported.
+     *
+     * Assigning a synonym map to a field ensures that query terms targeting that field are expanded at query-time using
+     * the rules in the synonym map. This attribute can be changed on existing fields.
+     *
+     * @param synonymMaps list of names of synonym maps to associate with this field.
+     * @return The SearchableField object itself.
+     */
+    public SearchableField setSynonymMaps(List<String> synonymMaps) {
```
heaths, line 121: I'd call this `synonymMapNames` since it's the names of synonymMaps. IIRC, that's what I did for that reason.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/FieldBase.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.azure.core.util.CoreUtils;
+import com.azure.core.util.logging.ClientLogger;
+
+import java.util.Objects;
+
+/**
+ * Base field type for helper classes to more easily create a {@link Index}.
+ */
+public abstract class FieldBase {
```
heaths, line 14: When you do the renames, you might call this `SearchFieldBase` given our discussion internally.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/FieldBase.java
```diff
@@ -0,0 +1,49 @@
...
+    private final DataType dataType;
+
+    /**
+     * Initializes a new instance of the {@link FieldBase} class.
+     * @param name The name of the field, which must be unique within the index or parent field.
+     * @param dataType The data type of the field.
+     */
+    protected FieldBase(String name, DataType dataType) {
+        if (CoreUtils.isNullOrEmpty(name)) {
+            throw logger.logExceptionAsError(new IllegalArgumentException("The name of field cannot be null"));
```
heaths, line 26: Nit: "The name of *the* field cannot be null"

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchableFieldProperty.java
```diff
@@ -0,0 +1,82 @@
...
+     * Optional arguments defines whether the field is filterable or not.
+     *
+     * @return True if it is filterable, and false by default for non-filterable field.
+     */
+    boolean isFilterable() default false;
+
+    /**
+     * Optional arguments defines the name of the analyzer used for the field.
+     *
+     * @return {@link AnalyzerName} String value. Or default to "null" String type.
```
heaths, line 58: 1. The comment is off, since the default is not "null". Also, why can't the default just be `null` (null reference, not a string with contents "null")? Is that not typical in Java?

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchableFieldProperty.java
```diff
@@ -0,0 +1,82 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * The annotation is to indicate whether the field is a searchable field. The boolean field of isSearchable
+ * defaults to true if use the annotation.
+ */
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SearchableFieldProperty {
```
heaths, line 19: Missing `name`. The name must be overridable. It can default to the property name, but that isn't always desirable. Same for all other attribute classes below.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
```diff
@@ -0,0 +1,256 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Stack;
+import java.util.stream.Collectors;
+
+/**
+ * Helper to convert model class to Search {@link Field fields}.
+ */
+public final class FieldBuilder {
+    private static final int MAX_DEPTH = 10000;
```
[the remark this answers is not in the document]
brjohnstmsft: Yes. Had this conversation with Sima already. The limits in the service are not static. This hard-coded value is orders of magnitude higher than the highest server-side limit on purpose so that the client doesn’t disallow valid field definitions.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchableFieldProperty.java
sima-zhu: Will change JavaDoc. Good catch!

Java annotation does not allow to take default of null reference.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/models/FieldBase.java
sima-zhu: Will do. Thanks for reminding! Will add you as review for renaming changes.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchableFieldProperty.java
sima-zhu: I have added name for Searchable, Simple attributes.

heaths: This leaves a problem for complex fields since we don't have an attribute.

Maybe we need to punt field renaming to preview 4. Come to think of it, in .NET different attributes (from Json.NET, though in track 2 we'd use the built-in ones) could rename it. You could do the same with your JSON serialization framework (I forget the name off hand).

/cc @brjohnstmsft

sima-zhu: https://github.com/Azure/azure-sdk-for-java/issues/10647

This is a summery issue for all open questions not going to address for this PR.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: Same one as below.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/FieldBuilder.java
sima-zhu: Several ways to achieve this:
1. Throw error if they both categories exist. 
1. Default to analyzer when both categories exist.

sima-zhu: To make it simple, I will throw error for now. Can come back to this we want different behavior.

PR #10523, opened 2020-04-27
Update receive batch API to provide option to get partition context

srnagar:
This PR is to change the receive batch consumer to accept an `EventBatchContext` object instead of `List<EventContext>`. This change allows the receive batch consumer to get the partition context even when the batch is empty. 

Previously, returning a list of `EventContext` would have required at least 1 event to be present in the batch for the user to know the partition context. A batch can be empty if the max wait time is set and there are no events received for that duration. When this happens, the `processEventBatch` handler is called with an empty list and the user will not be able to get the partition context. So, this change wraps the list of events in a context container that always guarantees the presence of partition context even when there are no events in the batch.

https://apiview.dev/Assemblies/Review/c2a7c6bbfd9c433aa407740a72fb24ec?diffRevisionId=315a6ac83318411bb00b36dc31578c1f

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionProcessor.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionProcessor.java
@@ -5,12 +5,14 @@ package com.azure.messaging.eventhubs.implementation;
 
 import com.azure.core.util.logging.ClientLogger;
 import com.azure.messaging.eventhubs.EventProcessorClient;
+import com.azure.messaging.eventhubs.EventProcessorClientBuilder;
 import com.azure.messaging.eventhubs.models.CloseContext;
 import com.azure.messaging.eventhubs.models.CloseReason;
 import com.azure.messaging.eventhubs.models.ErrorContext;
 import com.azure.messaging.eventhubs.models.EventBatchContext;
 import com.azure.messaging.eventhubs.models.EventContext;
 import com.azure.messaging.eventhubs.models.InitializationContext;
+import java.util.function.Consumer;
 
 /**
  * An abstract class defining all the operations that a partition processor can perform. Users of {@link
@@ -52,8 +54,9 @@ public abstract class PartitionProcessor {
     public abstract void processEvent(EventContext eventContext);
 
     /**
-     * This method is called when a batch of events is received for this partition. The default behavior will be to
-     * forward the individual events in this batch to {@link #processEvent(EventContext)}.
+     * This method is called when a batch of events is received for this partition. To receive events in batches,
+     * {@link EventProcessorClientBuilder#processEventBatch(Consumer, int) processEventBatch} has to be
+     * setup when creating {@link EventProcessorClient} instance.
      *
      * @param eventBatchContext The event batch context containing the batch of events along with partition information.
      */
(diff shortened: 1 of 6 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/EventBatchContext.java
```diff
@@ -0,0 +1,108 @@
...
+    public PartitionContext getPartitionContext() {
+        return partitionContext;
+    }
+
+    /**
+     * Returns a list of event data received from Event Hub.
+     *
+     * @return The list of event data received from Event Hub.
+     */
+    public List<EventData> getEventDataList() {
```
conniey, line 59: I may have tapped out during this API review, but `getEvents()` seems less cumbersome than getEventDataList.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionProcessor.java
```diff
@@ -45,15 +45,20 @@ public void initialize(InitializationContext initializationContext) {
...
-    public void processEventBatch(List<EventContext> eventContextBatch) {
-        eventContextBatch.forEach(this::processEvent);
+    /**
+     * This method is called when a batch of events is received for this partition. The default behavior will be to
+     * forward the individual events in this batch to {@link #processEvent(EventContext)}.
+     *
+     * @param eventBatchContext The event batch context containing the batch of events along with partition information.
+     */
+    public void processEventBatch(EventBatchContext eventBatchContext) {
+        throw logger.logExceptionAsError(new UnsupportedOperationException("Processing event batch not implemented"));
```
JonathanGiles, line 61: The default behavior mentioned in the JavaDoc does not match the code?

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/EventBatchContext.java
```diff
@@ -0,0 +1,108 @@
...
+import com.azure.messaging.eventhubs.CheckpointStore;
+import com.azure.messaging.eventhubs.EventData;
+import com.azure.messaging.eventhubs.EventProcessorClientBuilder;
+import java.util.List;
+import java.util.function.Consumer;
+import reactor.core.publisher.Mono;
+
+/**
+ * A class that contains a batch of {@link EventData} and the partition information the event batch belongs to. This is
+ * given to the {@link EventProcessorClientBuilder#processEventBatch(Consumer, int)}  processEventBatch} handler each
```
JonathanGiles, line 15: You have a curly brace mismatch here

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/EventBatchContext.java
JonathanGiles: I agree

srnagar: Updated name to `getEvents()`

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/PartitionProcessor.java
srnagar: Updated the javadoc.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/models/EventBatchContext.java
srnagar: Fixed.

PR #10528, opened 2020-04-27
Add send API for sending an iterable of events

srnagar:
This PR is to re-enable the send API for sending an iterable of events to Event Hub as a single batch from sync and async producer clients.

Arch board review - https://github.com/Azure/azure-sdk/issues/1259

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerClient.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerClient.java
@@ -3,6 +3,7 @@
 
 package com.azure.messaging.eventhubs;
 
+import com.azure.core.amqp.exception.AmqpException;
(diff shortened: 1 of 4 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/EventHubProducerClient.java
```diff
@@ -202,7 +202,7 @@ void send(Iterable<EventData> events) {
      * @param events Events to send to the service.
      * @param options The set of options to consider when sending this batch.
      */
-    void send(Iterable<EventData> events, SendOptions options) {
+    public void send(Iterable<EventData> events, SendOptions options) {
```
JonathanGiles, line 205: These need appropriate documentation of the exception that will be triggered, and any useful sample code.

srnagar: Updated documentation and added sample code.

PR #10564, opened 2020-04-28
Artifact grouping

mitchdenny:
This PR replaces #10490 ... I merged that one onto a feature branch so that I could do some end-to-end testing of how the pipeline copes with the segregated artifacts. There were a few tweaks that needed to be made which caused me to break my artifact staging logic out into a reusable template.

@JimSuplizio this should address your concerns. Here is the template artifact released:

https://repo.maven.apache.org/maven2/com/azure/azure-sdk-template/1.0.4-beta.14/

### DIFF
--- a/eng/pipelines/templates/steps/stage-artifacts.yml
+++ b/eng/pipelines/templates/steps/stage-artifacts.yml
@@ -8,7 +8,7 @@ steps:
       New-Item -Force -Type Directory -Name ${{parameters.TargetFolder}} -Path $(Pipeline.Workspace)
       $items = Get-ChildItem -Recurse -Path $(Pipeline.Workspace)/${{parameters.SourceFolder}} 
       Write-Host "Found $($items.Count) total items in source folder"
-      $filteredItems = $items | Where-Object -FilterScript { $_.Name -like "azure-sdk-template-[0-9]*.[0-9]*.[0-9]*-*" }      
+      $filteredItems = $items | Where-Object -FilterScript { $_.Name -like "${{parameters.PackageName}}-[0-9]*.[0-9]*.[0-9]*-*" }      
       Write-Host "Found $($filteredItems.Count) items which match filter criteria"
       $filteredItems | Copy-Item -Destination $(Pipeline.Workspace)/${{parameters.TargetFolder}}
       Get-ChildItem $(Pipeline.Workspace)/${{parameters.TargetFolder}}


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -113,8 +113,8 @@ jobs:
         displayName: 'Build and Package'
         inputs:
           mavenPomFile: pom.xml
-          goals: 'package'
-          options: '$(DefaultOptions) "-DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -DskipTests -Dinject-codesnippets -Dgenerate-overview -pl $(ProjectList) -am' # We include template-module so we ensure it always builds in CI
+          goals: 'deploy'
+          options: '$(DefaultOptions) -DskipTests -Dinject-codesnippets -Dgenerate-overview -Dspotbugs.skip=true -Dcheckstyle.skip=true -Drevapi.skip=true -pl $(ProjectList) -am -DaltDeploymentRepository=id::default::file://$(Build.ArtifactStagingDirectory)' # We include template-module so we ensure it always builds in CI
```
weshaggard, line 117: comment on the end can be deleted as it no longer applies.

weshaggard: Can we run deploy on just the main project so we can do something 
 like `file://$(Build.ArtifactStagingDirectory)/$(ServiceDirectory)`? Wouldn't that give us the easy segregation of the artifacts so we don't have to do any crazy filename matching in our release templates?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -113,8 +113,8 @@ jobs:
         displayName: 'Build and Package'
         inputs:
           mavenPomFile: pom.xml
-          goals: 'package'
-          options: '$(DefaultOptions) "-DpackageOutputDirectory=$(Build.ArtifactStagingDirectory)" -DskipTests -Dinject-codesnippets -Dgenerate-overview -pl $(ProjectList) -am' # We include template-module so we ensure it always builds in CI
```
weshaggard, line 117: We probably should also remove the packageOutputDirectory property and the ant copy step after we make this switch to deploy.

JimSuplizio: @weshaggard do we really want to remove that? I know when I'm building locally I typically set the packageOutputDirectory on my command line, it's more of an ease of use thing.

weshaggard: If we want deploy to the be the main goal for producing output then maybe we do want to remove it.

mitchdenny: I know for me it was one of those things that confused me when I first started working with Maven in depth on these team. I assumed that it was something that Maven did automatically until I realized we had this bit of Ant magic in our parent POM.

I'm all for magic if it is truly useful, but given Maven provides an OOB solution for spitting packages into a repository I'm not sure its that useful, and is possibly confusing.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
mitchdenny: I don't think that would help unfortunately since all projects in scope of the reactor would get put into that directory anyway ... unless I am missing something.

weshaggard: As long as the output deployment directory doesn't have any thing other then the assets that are going to be published (i.e. they don't contain dependencies and referenced stuff) then I don't think we have to worry about the version number in the directory. We can turn all the logic in our yml files to always take everything under the name of the artifact and eliminated the need for version wildcards and extra copying.

mitchdenny:
@weshaggard and @JimSuplizio I'm looking to get this one merged in this week so we've got a chance to try and iron out any issues that may not be obvious.

mitchdenny:
OK ... this is ready to go. I spent a fair amount of time today debugging issues with the artifact staging so that it worked with doc publishing. @scbedd would appreciate your input on my changes to Stage Artifacts.

eng/pipelines/templates/steps/stage-artifacts.yml
```diff
@@ -6,5 +6,10 @@ parameters:
 steps:
   - pwsh: |
       New-Item -Force -Type Directory -Name ${{parameters.TargetFolder}} -Path $(Pipeline.Workspace)
-      Copy-Item $(Pipeline.Workspace)/${{parameters.SourceFolder}}/${{parameters.PackageName}}-[0-9]*.[0-9]*.[0-9]* $(Pipeline.Workspace)/${{parameters.TargetFolder}}
-    displayName: Stage artifacts
\ No newline at end of file
+      $items = Get-ChildItem -Recurse -Path $(Pipeline.Workspace)/${{parameters.SourceFolder}} 
+      Write-Host "Found $($items.Count) total items in source folder"
+      $filteredItems = $items | Where-Object -FilterScript { $_.Name -like "azure-sdk-template-[0-9]*.[0-9]*.[0-9]*-*" }
```
scbedd, line 11: we've hardcoded template here.

scbedd: Probably `{{ params.PackageName }}` instead of `azure-template` right?

eng/pipelines/templates/steps/stage-artifacts.yml
```diff
@@ -6,5 +6,10 @@ parameters:
 steps:
   - pwsh: |
       New-Item -Force -Type Directory -Name ${{parameters.TargetFolder}} -Path $(Pipeline.Workspace)
-      Copy-Item $(Pipeline.Workspace)/${{parameters.SourceFolder}}/${{parameters.PackageName}}-[0-9]*.[0-9]*.[0-9]* $(Pipeline.Workspace)/${{parameters.TargetFolder}}
-    displayName: Stage artifacts
\ No newline at end of file
+      $items = Get-ChildItem -Recurse -Path $(Pipeline.Workspace)/${{parameters.SourceFolder}} 
+      Write-Host "Found $($items.Count) total items in source folder"
+      $filteredItems = $items | Where-Object -FilterScript { $_.Name -like "${{parameters.PackageName}}-[0-9]*.[0-9]*.[0-9]*-*" }
```
scbedd, line 11: 👍

eng/pipelines/templates/jobs/archetype-sdk-client.yml
weshaggard: @mitchdenny did we give up on the idea of having each artifact set published to its own folder in the "DevOps artifact"?  We have been seeing too many issues with doing the copying in various release tasks and I'd much prefer we avoid those issues by having the exact set of artifacts for a given package under a folder with that package name. Even if we have to do this copying once in the build stage that will make our lives so much easier in the various release stages that need to be correctly configured to copy the correct files.

PR #10565, opened 2020-04-28
support configured http client inherited by various clients

xseeseesee:
Fix: [Azure/azure-libraries-for-java#1133](https://github.com/Azure/azure-libraries-for-java/issues/1133)

Run living test for below samples:
- [x] app service
- [x] cosmos
- [x] sql
- [x] key vault

Run living test for below test cases:
- [x] app service
- [x] key vault

### DIFF
--- a/sdk/appservice/mgmt/src/test/java/com/azure/management/appservice/WarDeployTests.java
+++ b/sdk/appservice/mgmt/src/test/java/com/azure/management/appservice/WarDeployTests.java
@@ -9,8 +9,6 @@ import com.azure.management.resources.fluentcore.arm.Region;
 import java.io.File;
 
 import com.azure.management.resources.fluentcore.profile.AzureProfile;
-import com.azure.management.resources.fluentcore.utils.DelayProvider;
-import com.azure.management.resources.fluentcore.utils.SdkContext;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 


sdk/management/samples/src/main/java/com/azure/management/cosmosdb/samples/CreateCosmosDBWithEventualConsistency.java
```diff
@@ -37,9 +37,10 @@
     /**
      * Main function which runs the actual sample.
      * @param azure instance of the azure client
+     * @param connectionPolicy connection policy used for cosmos client
      * @return true if sample runs successfully
      */
-    public static boolean runSample(Azure azure) {
+    public static boolean runSample(Azure azure, ConnectionPolicy connectionPolicy) {
```
ChenTanyi, line 43: It seems useless to add this parameter in sample.

xccc-msft: As sample, we should provide this just in case the user requires to set proxy settings. At least, they will know CosmosClient need additional settings if they follow the sample and work with mgmt SDK & CosmosClient for their project.

ChenTanyi: But it is a sample, it is not the API. I think sample should not expose parameter if not necessary. Sample just provides minimal scripts to run, it is not to help user distinguish every parameter. Otherwise, I think lots of parameters should be exposed.

Like, container registry will need to create docker client. if in such situation, we need to expose docker proxy settings, docker tls settings and so on to user.

weidongxu-microsoft: What is in that `connectionPolicy`?

Alternatively you can provide an overload `runSample(Azure azure)` with default connectionPolicy, if this one is not very important for user.

Issue #10463, opened 2020-04-23
Avro finishing touches


gapra-msft:
Stuff from PR that I will get to once I fix QQ/CF
- Make header schema into record schema
- Deal with check methods in AvroUtils. Instead of defining method per each type you could pass Class type and use isAssignableFrom instead of instanceof . See https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#isAssignableFrom-java.lang.Class-
- Separate AvroUtils I'd rather break it down into more concrete bags of functionality. Otherwise it just becomes a bag do dump totally unrelated code and becomes a mess over time.
Reference: http://ralin.io/blog/oop-anti-patterns-utility-or-helper-classes.html

gapra-msft:
Alternative to rework some of the AvroSchema - split into Primitive and Complex Schemas

[11:48 AM] Kamil Sobol
    maybe schema should have isPrimitive()
​[11:50 AM] Kamil Sobol
    while ((schema instanceof ComplexSchema /* OR !schema.isPrimitve() */ ) || ((PrimitiveSchema)schema.canProgress()) {
    if (schema.isPrimitive()) {
        schema.progress();
    }
    /* If schema is done, pop yourself off the stack and publish.
       It is important to publish after popping yourself off because the parent may need to complete/add
       other schemas as needed. */
    if (schema.isDone()) {
        this.state.popOffStack();
        schema.publishResult();
    } else {
        if (!schema.isPrimitive()){
            KABOOM
        }
    }
    schema = this.state.peekFromStack();
}

PR #10579, opened 2020-04-28
Implemented an Avro Parser

gapra-msft:
Incorporates PRs https://github.com/Azure/azure-sdk-for-java/pull/10377 and https://github.com/Azure/azure-sdk-for-java/pull/9806.
Addresses comments from both. Note: this PR is against stg73, whereas the other two were against the changefeed branch.

Resolves https://github.com/Azure/azure-sdk-for-java/issues/10463

### DIFF
--- a/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
+++ b/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
@@ -1971,7 +1971,8 @@
 
   <!-- Internal Avro Parser does check the types before casting. Seems to be a false positive.  -->
   <Match>
-    <Class name="com.azure.storage.internal.avro.implementation.schema.complex.AvroSchema"/>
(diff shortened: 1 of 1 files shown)


eng/jacoco-test-coverage/pom.xml
```diff
@@ -163,6 +163,11 @@
       <artifactId>azure-storage-file-datalake</artifactId>
       <version>12.2.0-beta.1</version> <!-- {x-version-update;com.azure:azure-storage-file-datalake;current} -->
     </dependency>
+    <dependency>
```
JimSuplizio, line 166: If this is an internal only implementation do we need coverage numbers?

gapra-msft: I have plans to add a few more tests as I create PRs for the features that depend on this, so I'll bump the coverage numbers and see where we're at then. For now, I think this is the level of testing that we've been doing in .NET and Python as well

PR #10582, opened 2020-04-29
Fix azure-test-watcher for scheduled builds.

mitchdenny:
This PR adds the versioning tags to the azure-test-watcher module.


mitchdenny:
@alzimmermsft or @JimSuplizio could I get your eyes on this. It is to fix scheduled build failures.

PR #10594, opened 2020-04-29
Fixes backpressure for AmqpReceiveLink and Service Bus equivalent

### LINKED ISSUE #9792: Investigate backpressure, its interaction with prefetchCount, and link credits, when receiving through azure-core-amqp
Currently, we use prefetchCount as the number of credits to add to the link at any given time. The intention of prefetchCount was the "initial" number of credits that should be dumped on the link. Afterwards, credits would be added to the queue one at a time.
In Reactor, we use backpressure to ask for how many events we want. Say, we only want 8 events, we would only add enough credits on the link until 8 events are fetched. Afterwards, we shouldn't fetch more until another request is made. In the case that backpressure is equal to long.MAX_VALUE, we should continue adding credits to the link to fetch more messages.

### LINKED ISSUE #10520: ServiceBusReceiveLinkProcessor should not be using `request(0L)`
According to reactive spec, we should be requesting at minimum, 1 item on `Subscription.request(long)`. Currently, in debug mode, reactor will log an error message that looks like something failed. `onSubscribe(Subscription)`, we are requesting 0L, which results in this error.

conniey:
* Previously, we did not take into account backpressure properly and would continue to emit even after the requested number was satisfied. In this model, if the value is Long.MAX_VALUE, there is no back pressure and we always add credits to the link. Otherwise, we will only add the requested amount.
* Closes the reactor connection on close
* Fixes ClaimsBasedChannel to properly check the error status on unsuccessful authorization.

Fixes #10520 
Fixes #9792

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/implementation/AmqpReceiveLinkProcessorTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/implementation/AmqpReceiveLinkProcessorTest.java
@@ -32,6 +32,7 @@ import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
@@ -570,6 +571,40 @@ class AmqpReceiveLinkProcessorTest {
         final Integer creditValue = value.get();
         // Expecting 1 because it is Long.MAX_VALUE.
         Assertions.assertEquals(1, creditValue);
+    }
+    /**
+     * Verifies that when we request back pressure amounts, if it only requests a certain number of events, only
+     * that number is consumed.
+     */
+    @Test
+    void backpressureRequestOnlyEmitsThatAmount() {
+        // Arrange
+        final int backpressure = 10;
+        AmqpReceiveLinkProcessor processor = Flux.just(link1).subscribeWith(linkProcessor);
+        FluxSink<AmqpEndpointState> sink = endpointProcessor.sink();
+
+        when(link1.getCredits()).thenReturn(1);
+
+        // Act & Assert
+        StepVerifier.create(processor, backpressure)
+            .then(() -> {
+                sink.next(AmqpEndpointState.ACTIVE);
+                final int emitted = backpressure + 5;
+                for (int i = 0; i < emitted; i++) {
+                    messageProcessorSink.next(mock(Message.class));
+                }
+            })
+            .expectNextCount(backpressure)
+            .thenAwait(Duration.ofSeconds(1))
+            .thenCancel()
+            .verify();
+
+        Assertions.assertTrue(processor.isTerminated());
+        Assertions.assertFalse(processor.hasError());
+        Assertions.assertNull(processor.getError());
+
+        verify(link1).addCredits(eq(PREFETCH));
+        verify(link1).setEmptyCreditListener(any());
     }
 
     private static Flux<AmqpReceiveLink> createSink(AmqpReceiveLink[] links) {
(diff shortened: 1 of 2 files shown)


conniey:
java - eventhubs - tests

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ClaimsBasedSecurityChannel.java
```diff
@@ -37,7 +45,7 @@ public ClaimsBasedSecurityChannel(Mono<RequestResponseChannel> responseChannelMo
         this.authorizationType = Objects.requireNonNull(authorizationType, "'authorizationType' cannot be null.");
         this.retryOptions = Objects.requireNonNull(retryOptions, "'retryOptions' cannot be null.");
         this.credential = Objects.requireNonNull(tokenCredential, "'tokenCredential' cannot be null.");
-        this.cbsChannelMono = Objects.requireNonNull(responseChannelMono, "'responseChannelMono' cannot be null.");
+        this.cbsChannelMono = Objects.requireNonNull(responseChannelMono, "'cbsChannelMono' cannot be null.");
```
srnagar, line 48: This didn't have to change. The error message should use the name of the param instead of the instance variable name.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/AmqpReceiveLinkProcessor.java
```diff
@@ -25,40 +24,50 @@
...
+    private volatile boolean isCancelled;
     private volatile AmqpReceiveLink currentLink;
     private volatile Disposable currentLinkSubscriptions;
     private volatile Disposable retrySubscription;
 
-    volatile int wip;
-    static final AtomicIntegerFieldUpdater<AmqpReceiveLinkProcessor> WIP =
+    private volatile CoreSubscriber<? super Message> downstream;
+    @SuppressWarnings("rawtypes")
+    private static final AtomicReferenceFieldUpdater<AmqpReceiveLinkProcessor, CoreSubscriber> DOWNSTREAM =
```
srnagar, line 54: Can we use `AtomicReference` here instead of `AtomicReferenceFieldUpdater`? I don't think we'll have too many instances of AmqpReceiveLinkProcessor for memory being an issue. The reflection-based update can cause issues if the name of the volatile var is changed and the field updater is not updated.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusReceiveLinkProcessor.java
```diff
@@ -14,47 +14,69 @@
...
+    private volatile boolean isCancelled;
     private volatile AmqpReceiveLink currentLink;
     private volatile Disposable currentLinkSubscriptions;
     private volatile Disposable retrySubscription;
 
+    private volatile CoreSubscriber<? super Message> downstream;
+    @SuppressWarnings("rawtypes")
+    private static final AtomicReferenceFieldUpdater<ServiceBusReceiveLinkProcessor, CoreSubscriber> DOWNSTREAM =
+        AtomicReferenceFieldUpdater.newUpdater(ServiceBusReceiveLinkProcessor.class, CoreSubscriber.class,
+            "downstream");
```
srnagar, line 65: Same here - prefer to use `AtomicReference`.

sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/ClaimsBasedSecurityChannel.java
conniey: Good catch. I didn't check the actual parameter name.

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/com/azure/messaging/eventhubs/implementation/AmqpReceiveLinkProcessor.java
conniey: Makes sense. I was thinking of those EHs that have 1k partitions. I can revert it.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusReceiveLinkProcessor.java
conniey: For SB, I'm not sure, because they create a separate receive link for each session and non-session receiver. But I guess at any given time, they'll have (maxConcurrentSessions). 🤔

PR #10608, opened 2020-04-29
Fixed a bug in upload where we did not copy the buffer given to us so data could be overwritten upstream. 

gapra-msft:


sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/PayloadSizeGate.java
```diff
@@ -38,12 +38,17 @@
      * @return Buffered data or incoming data depending on threshold condition.
      */
     Flux<ByteBuffer> write(ByteBuffer buf) {
+        /* TODO (gapra): Investigate if there is way to avoid copying the data twice since this may result in lower perf. */
```
rickle-msft, line 41: Can you add that the other copy is when we write it to the pool? Just so we don't forget what we mean later?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/PayloadSizeGate.java
```diff
@@ -38,12 +38,18 @@
      * @return Buffered data or incoming data depending on threshold condition.
      */
     Flux<ByteBuffer> write(ByteBuffer buf) {
+        /* TODO (gapra): Investigate if there is way to avoid copying the data twice since this may result in lower
+            perf. The other copy is when we write the buffer to the pool */
+        /* Deep copy the buffer. This is required to ensure integrity of data. */
+        ByteBuffer cachedBuffer = ByteBuffer.allocate(buf.remaining()).put(buf);
+        cachedBuffer.flip();
```
kasobol-msft, line 45: Ups. My bad.
I think there isn't double copy now as copying was removed buffer pool - hence regression.

For future. Maybe it would be good to extract copying part to separate component that's inserted into Flux transform early in the process - so that it's more explicit (as we apparently quite depend on that).

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
```diff
@@ -80,6 +80,31 @@ class BlobAPITest extends APISpec {
...
+        stream.toByteArray() == randomData
+
+        where:
+        size                || _
+        1 * Constants.KB    || _  /* Less than copyToOutputStream buffer size, Less than maxSingleUploadSize */
+        8 * Constants.KB    || _  /* Equal to copyToOutputStream buffer size, Less than maxSingleUploadSize */
+        20 * Constants.KB   || _  /* Greater than copyToOutputStream buffer size, Less than maxSingleUploadSize */
+    }
+
+    /* TODO (gapra): Add more tests to test large data sizes. */
```
kasobol-msft, line 106: Thanks for the test. There wasn't anything that could catch that regression.

I'm not sure if adding larger data sizes would bring value in this test.

gapra-msft: Just a future thought to improve the condition of the upload tests :)

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/PayloadSizeGate.java
gapra-msft: https://github.com/Azure/azure-sdk-for-java/issues/10639

PR #10615, opened 2020-04-30
Add Readme Artifact to Output Folder if Present

scbedd:
This supports my efforts to bring docs.ms automation into our release pipelines.

@JimSuplizio tyvm for suggesting this route. Should be good!

@alzimmermsft this does **NOT** add it to the artifact itself, merely as an additional item that's present in the output directory.

### DIFF
--- a/sdk/parents/azure-client-sdk-parent/pom.xml
+++ b/sdk/parents/azure-client-sdk-parent/pom.xml
@@ -1311,7 +1311,7 @@
                     <echo>Copying ${project.basedir}/README.md to
                       ${packageOutputDirectory}/${project.build.finalName}-readme.md
                     </echo>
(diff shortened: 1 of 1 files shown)


JimSuplizio:
@scbedd, one thing to be aware of is that the README.md file you're copying will have had its relative links replaced as [that step is done before the build and package](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/templates/jobs/archetype-sdk-client.yml#L102).

scbedd:
@JimSuplizio exactly what we want.

scbedd:
Confirmed that this should not affect our publish to Maven due to our filtering in [this script.](https://github.com/Azure/azure-sdk-for-java/blob/d77fba6595e0159a545b7de3b1ac9f2255580f3e/eng/scripts/Publish-MavenPackages.ps1#L51)

PR #10628, opened 2020-04-30
draft for mgmt README

weidongxu-microsoft:
Fix https://github.com/Azure/azure-libraries-for-java/issues/1131

First draft for README

Let me know if there is more worth putting to key concepts and examples.

### DIFF
--- a/sdk/management/README.md
+++ b/sdk/management/README.md
@@ -27,8 +27,9 @@ that is optimized for ease of use, succinctness and consistency.
 
 ### Include the recommended packages
 
-It requires a `TokenCredential` implementation for authentication and an `HttpClient` implementation for HTTP client.
-`azure-identity` package and `azure-core-http-netty` package provide recommended implementation.
+Azure Management Libraries require a `TokenCredential` implementation for authentication and an `HttpClient` implementation for HTTP client.
+
+`azure-identity` package and `azure-core-http-netty` package provide the default implementation.
 
 [Azure Identity][azure_identity] provides Azure Active Directory token authentication support across the Azure SDK.
 
@@ -86,7 +87,7 @@ See [Samples][sample] for code snippets and samples.
 
 ## Key concepts
 
-The key concepts of Azure Management Library includes:
+The key concepts of Azure Management Libraries includes:
 
 - Fluent interface to manage Azure resources.
 - Asynchronous operations with [Reactor][reactor].
@@ -168,7 +169,7 @@ Azure azure = Azure
 
 ### Include single package
 
-Instead of include the complete Azure Management Library, you can choose to include a single service package.
+Instead of include the complete Azure Management Libraries, you can choose to include a single service package.
 
 For example, here is sample maven dependency for Compute package.
 
@@ -204,7 +205,7 @@ Azure SDKs for Java offer a consistent logging story to help aid in troubleshoot
 their resolution. The logs produced will capture the flow of an application before reaching the terminal state to help
 locate the root issue. View the [logging][logging] wiki for guidance about enabling logging.
 
-Sample code to enable logging in Azure Management Library.
+Sample code to enable logging in Azure Management Libraries.
 ```java
 Azure azure = Azure
     .configure()


sdk/management/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+By default, Azure Active Directory token authentication depends on correct configure of following environment variables.
+
+- `AZURE_CLIENT_ID` for Azure client ID.
+- `AZURE_TENANT_ID` for Azure tenant ID.
+- `AZURE_CLIENT_SECRET` or `AZURE_CLIENT_CERTIFICATE_PATH` for client secret or client certificate.
+
+In addition, Azure subscription ID can be configured via environment variable `AZURE_SUBSCRIPTION_ID`.
+
+See [Authenticate][authenticate] for more options.
```
weidongxu-microsoft, line 180: Point to AUTH.md later.

sdk/management/azure/pom.xml
```diff
@@ -12,7 +12,7 @@
     <relativePath>../pom.xml</relativePath>
   </parent>
 
-  <artifactId>azure</artifactId>
+  <artifactId>azure-management</artifactId>
```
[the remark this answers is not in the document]
ChenTanyi: Does the azure-parent pom also need to be changed?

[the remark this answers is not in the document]
weidongxu-microsoft: Yes. I think so.

sdk/management/README.md
```diff
@@ -0,0 +1,221 @@
+# Azure management client library for Java
+
+The Azure Management Library for Java is a higher-level, object-oriented API for *managing* Azure resources,
```
xccc-msft, line 3: Which should be the expected for us in this repo, `Azure management client library for Java`, `Azure Management Library for Java`? Or using `libraries` as we did in previous repo? To distinguish the dataplane SDKs and auto-generated Lite SDKs.

weidongxu-microsoft: The title is due to this https://github.com/Azure/azure-sdk-for-java/blob/master/eng/.docsettings.yml#L16
Currently not want to touch that yet.

Given this README is located for all mgmt sdks, yes, I agree `libraries` would be better.

sdk/management/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Include optional packages
+
+[Azure Identity][azure_identity] provides Azure Active Directory token authentication support across the Azure SDK.
+
+[//]: # ({x-version-update-start;com.azure:azure-identity;current})
+```xml
+<dependency>
+  <groupId>com.azure</groupId>
+  <artifactId>azure-identity</artifactId>
```
yungezz, line 36: is azure-identity optional package? I can get the point our mgmt libs not depends on it. but from user's perspective, how to make it clear to them that they need azure-identity or TokenCredential before performing any mgmt operation. Maybe link to auth.md somewhere?

sdk/management/README.md
```diff
@@ -0,0 +1,221 @@
...
+  <version>1.1.0-beta.3</version>
+</dependency>
+```
+[//]: # ({x-version-update-end})
+
+[Azure Core Netty HTTP client][azure_core_http_netty] is a plugin for [Azure Core][azure_core] HTTP client API.
```
yungezz, line 42: i think it's worth to a separate section to let user know they need choose from 2 http client by themselves, and reference to each of them

sdk/management/README.md
```diff
@@ -0,0 +1,221 @@
...
+or checkout [StackOverflow for Azure Java SDK](http://stackoverflow.com/questions/tagged/azure-java-sdk).
+
+### HTTP client
+
+An `HttpClient` implementation must exist on the classpath.
+See [Include optional packages](#include-optional-packages).
+
+### Authentication
+
+By default, Azure Active Directory token authentication depends on correct configure of following environment variables.
```
yungezz, line 172: how about move this earlier, my point is that as a user, when i read or follow major steps in the readme, I can have a rough end to end idea of complete a mgmt task, from auth to config, to service features, then logging etc advanced features.

yungezz:
is it necessary to have a list of java mgmt sdks, then reference to it from here for users' easy search?

sdk/management/README.md
weidongxu-microsoft: It is optional. User can do their own OAUTH, without any lib. I think this is the exact reason that requires us not to include azure-identity module.

sdk/management/README.md
weidongxu-microsoft: We can optionally refer to https://github.com/Azure/azure-sdk-for-java/wiki/HTTP-clients

I've checked other libs when writing this, they generally only mention it https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/storage/azure-storage-common#default-http-client

sdk/management/README.md
weidongxu-microsoft: Currently following other client README structure dictated by https://github.com/Azure/azure-sdk-for-java/blob/master/eng/.docsettings.yml#L16

When AUTH.md is ready, we can refer it in first line.

weidongxu-microsoft: I've moved this section to "Getting started" at top. Also added sample code.

sdk/management/azure/pom.xml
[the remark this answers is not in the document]
weidongxu-microsoft: Also change parent to "azure-management-parent".

Issue #10183, opened 2020-04-14
Implement multisession receive


hemanttanwar:
ServiceBusMultiSessionProcessorClient : In order to support multiple session, implementation feature similar to  Track-1 `IMessageAndSessionPump`  for ServiceBus.
Something like ServiceBusMultiSessionPumpManager 

  Should be able to find available session.
  Should be able to move to next available session upon completion of previous session.
  Inform user of errors in session ( Example Session lock lost and other similar errors).

PR #10637, opened 2020-04-30
Add support for associated link name, session id, and clean up ManagementNode

conniey:
## Feature changes
* Removes `sessionId` from `createManagementNode`. Multi-session deals with many session ids, we don't want to create a different management node for each session id.
* Adds support for associating the link name.
* Moves `lastPeekedSequence` into receiver. The receiver can peek different messages depending on its sessionId, this will result in the incorrectly mapped sequence number if we use `peek()` or `peekBatch(int)`.
* Cleans up ManagementNode by using overloads and reducing the number of interface methods.
* Adds tests.

## Bug fixes
* Fixes bug where management node by its key was not properly returned.
* Fixes bug where session'd entity operations were not correct on management node.

Part of a larger PR for multi-sessioned support. Related to #10183

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusReceiveLinkProcessor.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/implementation/ServiceBusReceiveLinkProcessor.java
@@ -326,7 +326,8 @@ public class ServiceBusReceiveLinkProcessor extends FluxProcessor<AmqpReceiveLin
             logger.info("Parent connection is disposed. Not reopening on error.");
         }
 
-        logger.warning("linkName[{}] entityPath[{}]. Non-retryable error occurred in AMQP receive link.", throwable);
+        logger.warning("linkName[{}] entityPath[{}]. Non-retryable error occurred in AMQP receive link.",
+            linkName, entityPath, throwable);
         lastError = throwable;
 
         isTerminated.set(true);


sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/implementation/ManagementChannelTests.java
```diff
@@ -117,11 +137,9 @@ void teardown(TestInfo testInfo) {
     void setsSessionState(byte[] state) {
         // Arrange
         final String sessionId = "A session-id";
-        managementChannel = new ManagementChannel(Mono.just(requestResponseChannel), NAMESPACE, ENTITY_PATH, sessionId,
-            tokenManager, messageSerializer, Duration.ofSeconds(10));
 
         // Act
-        StepVerifier.create(managementChannel.setSessionState(state))
+        StepVerifier.create(managementChannel.setSessionState(sessionId, state, LINK_NAME))
```
hemanttanwar, line 142: Another error case : Where user pass in a session id which is not equal to current active session id.

conniey: This is done on the management channel. This id doesn't have to be the active one. It could be one that was passed in from another application that is also renewing the lock.

conniey: This is done on the managememt channel. If another application passes you this id, you can complete it as long as someone is renewing that lock. the .NET code doesn't check that the session is not lost.

conniey: If this is done on the actual receive link, then it checks to see if it is not lost.

PR #10641, opened 2020-04-30
Implemented query for blob and datalake

gapra-msft:
Initial PR - https://github.com/Azure/azure-sdk-for-java/pull/8688

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobClientBase.java
@@ -1099,7 +1099,7 @@ public class BlobClientBase {
 
         // Create input stream from the data.
         if (response == null) {
-            throw new IllegalStateException("Query response cannot be null");
+            throw logger.logExceptionAsError(new IllegalStateException("Query response cannot be null"));
         }
         return new FluxInputStream(response.getValue());
     }
--- a/sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileClient.java
+++ b/sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/DataLakeFileClient.java
@@ -594,7 +594,7 @@ public class DataLakeFileClient extends DataLakePathClient {
 
         // Create input stream from the data.
         if (response == null) {
-            throw new IllegalStateException("Query response cannot be null");
+            throw logger.logExceptionAsError(new IllegalStateException("Query response cannot be null"));
         }
         return new FluxInputStream(response.getValue());
     }


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobQueryAsyncResponse.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.azure.core.http.HttpHeaders;
+import com.azure.core.http.HttpRequest;
+import com.azure.core.http.rest.ResponseBase;
+import reactor.core.publisher.Flux;
+
+import java.nio.ByteBuffer;
+
+/**
+ * This class contains the response information returned from the server when running a quick query a blob.
```
gapra-msft, line 14: fix doc

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1227,7 +1247,8 @@ private void downloadToFileCleanup(AsynchronousFileChannel channel, String fileP
                     hd.isIncrementalCopy(), hd.getDestinationSnapshot(), AccessTier.fromString(hd.getAccessTier()),
                     hd.isAccessTierInferred(), ArchiveStatus.fromString(hd.getArchiveStatus()),
                     hd.getEncryptionKeySha256(), hd.getAccessTierChangeTime(), hd.getMetadata(),
-                    hd.getBlobCommittedBlockCount(), hd.getVersionId(), hd.isCurrentVersion());
+                    hd.getBlobCommittedBlockCount(), null, null);
```
gapra-msft, line 1250: todo: i think i generated off a slightly older version of swagger, so I need to undo this

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1599,4 +1620,293 @@ public String generateSas(BlobServiceSasSignatureValues blobServiceSasSignatureV
...
+            .setInputSerialization(in)
+            .setOutputSerialization(out);
+
+        return this.azureBlobStorage.blobs().quickQueryWithRestResponseAsync(null, null, qr,
+            getSnapshotId(), null, requestConditions.getLeaseId(), requestConditions.getIfModifiedSince(),
+            requestConditions.getIfUnmodifiedSince(), requestConditions.getIfMatch(),
+            requestConditions.getIfNoneMatch(), null, getCustomerProvidedKey(), context)
+            .map(response -> new BlobQueryAsyncResponse(response.getRequest(), response.getStatusCode(),
+                response.getHeaders(),
+                parse(response.getValue(), o -> this.parseRecord(o, finalQueryOptions.getErrorReceiver(),
```
gapra-msft, line 1686: todo: comment this part

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1599,4 +1620,293 @@ public String generateSas(BlobServiceSasSignatureValues blobServiceSasSignatureV
...
+     * Parses a quick query record.
+     *
+     * @param quickQueryRecord The quick query record.
+     * @param errorReceiver The error receiver.
+     * @param progressReceiver The progress receiver.
+     * @return The optional data in the record.
+     */
+    private Mono<ByteBuffer> parseRecord(Object quickQueryRecord, ErrorReceiver<BlobQueryError> errorReceiver,
+        ProgressReceiver progressReceiver) {
+        if (!(quickQueryRecord instanceof Map)) {
```
gapra-msft, line 1715: remove : this is redundant

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1599,4 +1620,293 @@ public String generateSas(BlobServiceSasSignatureValues blobServiceSasSignatureV
...
+            case "resultData":
+                return parseResultData(record);
+            case "end":
+                return parseEnd(record, progressReceiver);
+            case "progress":
+                return parseProgress(record, progressReceiver);
+            case "error":
+                return parseError(record, errorReceiver);
+            default:
+                return Mono.error(new UncheckedIOException(new IOException(String.format("Unknown record type %s "
```
gapra-msft, line 1732: This can be just a IllegalStateException or something?

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/BlobAPITest.groovy
```diff
@@ -3,30 +3,11 @@
...
-import com.azure.storage.blob.models.CustomerProvidedKey
-import com.azure.storage.blob.models.DeleteSnapshotsOptionType
-import com.azure.storage.blob.models.DownloadRetryOptions
-import com.azure.storage.blob.models.LeaseStateType
-import com.azure.storage.blob.models.LeaseStatusType
-import com.azure.storage.blob.models.ParallelTransferOptions
-import com.azure.storage.blob.models.PublicAccessType
-import com.azure.storage.blob.models.RehydratePriority
-import com.azure.storage.blob.models.SyncCopyStatusType
+import com.azure.storage.blob.models.*
```
gapra-msft, line 10: uh oh - need to revert this

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlockBlobAPITest.groovy
```diff
@@ -11,22 +11,8 @@ import com.azure.core.http.HttpRequest
...
-import com.azure.storage.blob.models.BlobHttpHeaders
-import com.azure.storage.blob.models.BlobRange
-import com.azure.storage.blob.models.BlobRequestConditions
-import com.azure.storage.blob.models.BlobStorageException
-import com.azure.storage.blob.models.BlockListType
-import com.azure.storage.blob.models.CustomerProvidedKey
-import com.azure.storage.blob.models.ParallelTransferOptions
-import com.azure.storage.blob.models.PublicAccessType
+import com.azure.storage.blob.*
+import com.azure.storage.blob.models.*
```
gapra-msft, line 15: revert

sdk/storage/azure-storage-blob/swagger/README.md
```diff
@@ -15,7 +15,8 @@ autorest --use=@microsoft.azure/autorest.java@3.0.4 --use=jianghaolu/autorest.mo
 
 ### Code generation settings
 ``` yaml
-input-file: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/storage-dataplane-preview/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-12-12/blob.json
+# input-file: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/storage-dataplane-preview/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-12-12/blob.json
+input-file: https://github.com/Azure/azure-rest-api-specs/blob/6cd6f02d824569dccb2427cf1611fe1cd8cc4350/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-12-12/blob.json
```
gapra-msft, line 19: pointed at specific version to make generating easy

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1605,4 +1623,293 @@ public String generateSas(BlobServiceSasSignatureValues blobServiceSasSignatureV
...
+    }
+
+    /**
+     * Avro parses a quick query reactive stream.
+     *
+     * @param avro The reactive stream.
+     * @param objectHandler The function to handle objects.
+     * @return The parsed quick query reactive stream.
+     */
+    private Flux<ByteBuffer> parse(Flux<ByteBuffer> avro, Function<Object, Mono<ByteBuffer>> objectHandler) {
```
rickle-msft, line 1701: Can you add a description of what's happening here and what the data looks like? Like we get a stream of bytes that's avro. It's expected to have this schema. Once we get the data out of it, we need to apply these transformations, etc.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlobBaseAPITest.groovy
```diff
@@ -0,0 +1,609 @@
...
+
+        then:
+        notThrown(BlobStorageException)
+        for (int j = 0; j < downloadedData.length; j++) {
+            assert osData[j] == downloadedData[j]
+        }
+
+        where:
+        numCopies | recordSeparator || _
+        0         | '\n'            || _
```
rickle-msft, line 175: Seems like the recordSeparator variable can just be a constant.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlobBaseAPITest.groovy
```diff
@@ -0,0 +1,604 @@
...
+        BlobQueryOptions options = new BlobQueryOptions().setInputSerialization(ser).setOutputSerialization(ser)
+
+        /* Input Stream. */
+        when:
+        InputStream qqStream = bc.openQueryInputStream(expression, options)
+        byte[] queryData = readFromInputStream(qqStream, downloadedData.length)
+
+        then:
+        notThrown(IOException)
+        for (int j = 0; j < downloadedData.length; j++) {
```
rickle-msft, line 157: I think in groovy you can just do a == between arrays and it'll compare them for you

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlobBaseAPITest.groovy
```diff
@@ -0,0 +1,604 @@
...
+
+        where:
+        modified | unmodified | match       | noneMatch    | leaseID
+        newDate  | null       | null        | null         | null
+        null     | oldDate    | null        | null         | null
+        null     | null       | garbageEtag | null         | null
+        null     | null       | null        | receivedEtag | null
+        null     | null       | null        | null         | garbageLeaseID
+    }
+}
```
rickle-msft, line 604: Do you test all the options like escape character and line delimiter? I don't remember seeing tests for that. Maybe you can make your upload data methods more generic to accept these options.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
```diff
@@ -0,0 +1,242 @@
...
+        /* At this point, buffer should not be null. If it is, that indicates either an error or completion event
+           was emitted by the Flux. */
+        if (this.buffer == null) { // Only executed on first subscription.
+            if (this.lastError != null) {
+                throw logger.logThrowableAsError(this.lastError);
+            }
+            if (this.fluxComplete) {
+                return -1;
+            }
+            this.buffer = new ByteArrayInputStream(new byte[0]);
```
rickle-msft, line 95: This line feels spurious. Since the comment says it should only be null if there was an error or completion, and both tof those cases exit the method, we shouldn't need this line, right?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
```diff
@@ -0,0 +1,242 @@
...
+           the value -1 is returned; otherwise, at least one byte is read and stored into b. */
+
+        /* Not subscribed? subscribe and block for data */
+        if (!subscribed) {
+            blockForData();
+        }
+        /* Now, we have subscribed. */
+        /* At this point, buffer should not be null. If it is, that indicates either an error or completion event
+           was emitted by the Flux. */
+        if (this.buffer == null) { // Only executed on first subscription.
```
rickle-msft, line 88: I think you may want to lock around checking if buffer is null and assigning to it here. You lock around that in your onNext below, so modifying it here without locking will give you a race condition, i think

rickle-msft: Same thing wherever you access this.buffer

---
In reply to: [425432427](https://github.com/Azure/azure-sdk-for-java/pull/10641#discussion_r425432427) [](ancestors = 425432427)

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
```diff
@@ -0,0 +1,242 @@
...
+            return -1;
+        } else {
+            throw logger.logExceptionAsError(new IllegalStateException("An unexpected error occurred. No data was "
+                + "read from the stream but the stream did not indicate completion."));
+        }
+    }
+
+    @Override
+    public void close() throws IOException {
+        subscription.cancel();
```
rickle-msft, line 125: Is there any value in checking if the flux has completed before cancelling the subscription? Or is this a no-op in that case?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                },
+                // Error consumer
+                throwable -> {
+                    // Signal the consumer in case an error occurs (indicates we completed without data).
+                    signalOnCompleteOrError();
+                    if (throwable instanceof HttpResponseException) {
+                        this.lastError = new IOException(throwable);
+                    } else if (throwable instanceof IllegalArgumentException) {
+                        this.lastError = new IOException(throwable);
```
rickle-msft, line 189: I don't think we need to wrap IllegalArgumentException in an IOException. And I don't think we need to wrap ioExceptions in another IOException.

gapra-msft: at least for IllegalArgument - is the read API allowed to throw anything other than IOexception? I'm not certain

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
gapra-msft: I think I added this since I wanted buffer to be null only before the subscription. But now that I think about it, that doesnt need to be a semantic I follow since I have the subscribed boolean.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
gapra-msft: Do you think it'd be ok if I just move the buffer population above the locking logic? The intent was for the waitingForData variable to be protected by the lock moreso than the buffer

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
gapra-msft: I don't think it would matter if the flux completed or not because the user is done with the stream regardless

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobQueryOptions.java
```diff
@@ -0,0 +1,125 @@
...
+
+package com.azure.storage.blob.models;
+
+import com.azure.storage.common.ErrorReceiver;
+import com.azure.storage.common.ProgressReceiver;
+
+/**
+ * Optional parameters for Blob Query.
+ */
+public class BlobQueryOptions {
```
kasobol-msft, line 12: This should be annotated with `@Fluent`

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobQuerySerialization.java
```diff
@@ -0,0 +1,24 @@
...
+
+package com.azure.storage.blob.models;
+
+/**
+ * Defines the input and output serialization for a blob quick query request.
+ * either {@link BlobQueryJsonSerialization} or {@link BlobQueryDelimitedSerialization}
+ */
+public abstract class BlobQuerySerialization {
+
+    protected char recordSeparator;
```
kasobol-msft, line 12: @seanmcc-msft discovered that .NET XMLSerializer transform line endings. We should check if Java's XML engine does the same thing and if it does then we'd probably need to ask to extend generator.
For example `\n` might be converted into `\r\n`. We should make sure whitespaces are sent to server side verbatim.

Other thing to check is column separator in delimited serialization. For example adding test that would query a [TSV](https://en.wikipedia.org/wiki/Tab-separated_values) file.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1605,4 +1625,292 @@ public String generateSas(BlobServiceSasSignatureValues blobServiceSasSignatureV
...
+            }
+        } else {
+            return Mono.error(new IllegalArgumentException("Failed to parse error record from "
+                + "query response stream."));
+        }
+        return Mono.empty();
+    }
+
+    /**
+     * Validates that all parameters are non-null. Throws IOException if any of them are.
```
kasobol-msft, line 1836: This doesn't seem to throw anything.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1605,4 +1625,292 @@ public String generateSas(BlobServiceSasSignatureValues blobServiceSasSignatureV
...
+    private static JsonTextConfiguration transformJson(BlobQueryJsonSerialization jsonSerialization) {
+        if (jsonSerialization == null) {
+            return null;
+        }
+        return new JsonTextConfiguration()
+            .setRecordSeparator(charToString(jsonSerialization.getRecordSeparator()));
+    }
+
+    private static String charToString(char c) {
+        return c == '\0' ? "" : Character.toString(c);
```
kasobol-msft, line 1914: Why do we treat \0 specially? Can customer choose \0 as record separator (for whatever weird reason)?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobQueryInputStream.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.azure.storage.common.implementation.FluxInputStream;
+import reactor.core.publisher.Flux;
+
+import java.nio.ByteBuffer;
+
+/**
+ * An <code>InputStream</code> interface that represents the stream to use for reading the query response.
+ */
+public class BlobQueryInputStream extends FluxInputStream {
```
kasobol-msft, line 14: This type doesn't seem to bring any value. Should we just use `FluxInputStream` or even just `InputStream` as return type from `query` method family ? I think if we should declare most abstract type as return type as we can.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlobBaseAPITest.groovy
```diff
@@ -0,0 +1,602 @@
...
+
+        then:
+        // Make sure theyre all increasingly bigger
+        for (long progress : mockReceiver.progressList) {
+            assert progress >= temp
+            temp = progress
+        }
+    }
+
+    class MockProgressReceiver implements ProgressReceiver {
```
kasobol-msft, line 418: I'd move these classes to the bottom of the files so that we don't mix tests with test aids.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ErrorReceiver.java
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT License.
+
+package com.azure.storage.common;
+
+/**
+ * An {@code ErrorReceiver} is a class that can be used to report errors on transfers. When specified on transfer
+ * operations, the {@code reportError} method will be called whenever errors are encountered. The user may configure
+ * this method to report errors in whatever format desired.
+ */
+public interface ErrorReceiver<T> {
```
kasobol-msft, line 11: I think we could replace this with https://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html in respective API and just call parameters `errorConsumer`.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlobBaseAPITest.groovy
gapra-msft: I'll add some tests testing other ones

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
```diff
@@ -0,0 +1,243 @@
...
+            lock.unlock();
+        }
+    }
+
+    /**
+     * Subscribes to the data with a special subscriber.
+     */
+    private void subscribeToData() {
+        this.data
+            .onBackpressureBuffer(Constants.GB)
```
kasobol-msft, line 170: What does it do? 1GB is quite high threshold .

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlobBaseAPITest.groovy
gapra-msft: ooh yeah

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
kasobol-msft: IllegalArgumentException sounds like validation failed somewhere. That usually should bubble up. Isn't it runtimeexception?

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/models/FileQueryAsyncResponse.java
```diff
@@ -0,0 +1,30 @@
...
+import com.azure.core.http.HttpRequest;
+import com.azure.core.http.rest.ResponseBase;
+import reactor.core.publisher.Flux;
+
+import java.nio.ByteBuffer;
+
+/**
+ * This class contains the response information returned from the server when running a query on a file.
+ */
+public final class FileQueryAsyncResponse extends ResponseBase<FileQueryHeaders, Flux<ByteBuffer>> {
```
kasobol-msft, line 16: Shouldn't we just use ResponseBase<> ? this type doesn't seem to do anything new. If we need we can always introduce this in the future without breaking change.
Or is this generated code?

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/models/FileQueryInputStream.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.azure.storage.common.implementation.FluxInputStream;
+import reactor.core.publisher.Flux;
+
+import java.nio.ByteBuffer;
+
+/**
+ * An <code>InputStream</code> interface that represents the stream to use for reading the query response.
+ */
+public class FileQueryInputStream extends FluxInputStream {
```
kasobol-msft, line 14: Same feedback here. I think APIs should just return InputStream. We don't need this type.

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/models/FileQuerySerialization.java
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Gets the record separator.
+     *
+     * @return the record separator.
+     */
+    public char getRecordSeparator() {
+        return recordSeparator;
+    }
+
+    abstract FileQuerySerialization setRecordSeparator(char recordSeparator);
```
kasobol-msft: I think we should just have normal setter here. Derivatives shouldn't be concerned about this field - I assume we're using inheritance to share this property.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobQuerySerialization.java
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Gets the record separator.
+     *
+     * @return the record separator.
+     */
+    public char getRecordSeparator() {
+        return recordSeparator;
+    }
+
+    abstract BlobQuerySerialization setRecordSeparator(char recordSeparator);
```
kasobol-msft: Same here. I think this should be just usual setter.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobQuerySerialization.java
gapra-msft: Yep! Just added a few tests for this. It seems like \t works. Some that don't work however, include \r, \n. I'll bring this up in standup

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
gapra-msft: That's poor porting on my part. It used to throw. Will fix the doc

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/models/FileQueryAsyncResponse.java
gapra-msft: This is more just following the pattern of what we return on downloads. FileQueryHeaders is a generated type

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/models/FileQuerySerialization.java
gapra-msft: I originally wanted to do that but the issue I had with that was we want to preserve the specific type to keep things  Fluent. Otherwise I wouldnt be able to do something like this to make a BlobQuerySerialization. 

blobQuerySerialization = new BlobQueryDelimitedSerialization().setRecordSeparator('\n').setColumnSep.....

kasobol-msft: Take a look on this creative solution https://stackoverflow.com/questions/14105275/fluent-setters-with-inheritance-in-java .

Other option would be to move implementation here and just call `super.setRecordSeparator` + cast in derivatives to avoid repeating implementation.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
```diff
@@ -1605,4 +1625,292 @@ public String generateSas(BlobServiceSasSignatureValues blobServiceSasSignatureV
...
+     * <a href="https://docs.microsoft.com/en-us/rest/api/">Azure Docs</a></p>
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.specialized.BlobAsyncClientBase.query#String}
+     *
+     * @param expression The query expression.
+     * @return A reactive response containing the queried data.
+     */
+    public Flux<ByteBuffer> query(String expression) {
```
seanmcc-msft, line 1643: I put the Query API on BlockBlobClient, since it currently only works with block blobs.  If other blobs, we can move it to BaseBlobClient (assuming you have a BaseBlobClient in Java)

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
seanmcc-msft: Also, if this is the main BlobClient, I think it might be best to put the Query private helper methods somewhere else, so we don't clutter it.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ErrorReceiver.java
seanmcc-msft: Ya in .NET we replaced this interface with a lamda.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
gapra-msft: I had to do this since you can't get a snapshot block blob client. We figured this is fine since there are some other APIs on Blob Base that don't work for specific types of blobs

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ErrorReceiver.java
gapra-msft: Do you think it's not important to follow the pattern of Progress Receiver?

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
rickle-msft: Yea I'd rather throw IllegalState in that case so that it's consistent with the comments and with the intended behavior.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
rickle-msft: Yea true. I more was just worried about the flux throwing an exception if you cancel a completed subscription.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ErrorReceiver.java
kasobol-msft: I think so. In next major rev we could think about using standard Java functional interfaces for Progress receiver as well. 
Let's not create types where standard JDK types can be used.

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
rickle-msft: Uhh. It feels safest to put the buffer population in the lock because what if you get two items emitted really quickly and they have different amounts of data or one of them is the end of the stream and is actually empty. Then we may have entered an if block based on one buffer but are operating on a different buffer for which the conditions are no longer true. But we can also walk through this offline and think about possibilities if you think it's safe to put the buffer stuff outside the lock

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
gapra-msft: Yeah I can move these into a BlobQueryUtil in implementation

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ErrorReceiver.java
seanmcc-msft: @KrzysztofCwalina wasn't a fan of the interface in .NET:

https://github.com/Azure/azure-sdk-for-net/pull/11586#discussion_r423212648

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java
gapra-msft: I thought that Karim told me that null should be "" but I have sent an email reconfirming.

gapra-msft: null is the default param for field quote and escape char according to the QQ folks.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobQueryInputStream.java
gapra-msft: I think that's a good idea

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/ErrorReceiver.java
gapra-msft: Ok I'm cool with using a consumer interface.

gapra-msft: I also changed the ProgressReceiver and made a custom BlobQueryProgress class

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/models/FileQuerySerialization.java
gapra-msft: Yeah I've seen the one where you suppress warnings before, but I didnt like the fact that I would need to suppress warnings.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlobBaseAPITest.groovy
gapra-msft: done

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
gapra-msft: done

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
gapra-msft: We only request when we blockForData though so I'm not sure that could happen. We can talk about this next week

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
gapra-msft: I don;t see any documentation about it throwing if the subscription completed. https://www.reactive-streams.org/reactive-streams-1.0.0-javadoc/org/reactivestreams/Subscription.html#cancel--

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
gapra-msft: Here actually should I just wrap any type of exception except IOException in an IOEXception? I'm forgetting why I even checked for specific types of throwables.

gapra-msft: Like this 
                    if (throwable instanceof IOException) {
                        this.lastError = (IOException) throwable;
                    } else {
                        this.lastError = new IOException(throwable);
                    }

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
```diff
@@ -447,4 +449,22 @@ This should be the only place we allocate memory in encryptBlob(). Although ther
...
         }
     }
+
+    /**
+     * Unsupported.
+     */
+    @Override
+    public Flux<ByteBuffer> query(String expression) {
+        throw logger.logExceptionAsError(new UnsupportedOperationException(
+            "Cannot query a client side encrypted client"));
```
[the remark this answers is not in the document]
gapra-msft: done

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobAsyncClient.java
```diff
@@ -447,4 +449,22 @@ This should be the only place we allocate memory in encryptBlob(). Although ther
             throw logger.logExceptionAsError(Exceptions.propagate(e));
         }
     }
+
+    /**
+     * Unsupported.
```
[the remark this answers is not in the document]
gapra-msft: added more info

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobQueryOptions.java
gapra-msft: done

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlobBaseAPITest.groovy
gapra-msft: done

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
gapra-msft: This is in case downstream is slower for some reason than the actual upstream flow of data. We dont want to drop any bytebuffers so I think this is the way we tell Reactor don't drop anything. Perhaps the default onBackpressureBuffer() may be more appropriate here than setting it to a GB

sdk/storage/azure-storage-blob-cryptography/src/main/java/com/azure/storage/blob/specialized/cryptography/EncryptedBlobClient.java
```diff
@@ -202,4 +207,41 @@ public PageBlobClient getPageBlobClient() {
...
+            "Cannot query data encrypted on client side."));
+    }
+
+    /**
+     * Unsupported. Cannot query data encrypted on client side.
+     */
+    @Override
+    public void query(OutputStream stream, String expression) {
+        throw logger.logExceptionAsError(new UnsupportedOperationException(
+            "Cannot query a client side encrypted client"));
```
gapra-msft, line 234: just noticed this was missed.  will fix

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobQueryDelimitedSerialization.java
```diff
@@ -0,0 +1,95 @@
...
+
+package com.azure.storage.blob.models;
+
+/**
+ * Defines the input or output delimited (CSV) serialization for a blob quick query request.
+ */
+public class BlobQueryDelimitedSerialization extends
+    BlobQuerySerialization {
+
+    /* TODO(gapra) : Consider diff datatype in case future break. */
```
gapra-msft, line 12: I think this will continue to be char from what I can tell from the discussions we've had with the QQ team

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobQuickQueryHeaders.java
```diff
@@ -0,0 +1,1151 @@
...
+import com.azure.core.util.DateTimeRfc1123;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
+import java.time.OffsetDateTime;
+import java.util.Map;
+
+/**
+ * Defines headers for QuickQuery operation.
+ */
+@JacksonXmlRootElement(localName = "Blob-QuickQuery-Headers")
```
gapra-msft, line 19: TODO : rename to BlobQueryHeaders

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/BlobBaseAPITest.groovy
```diff
@@ -0,0 +1,667 @@
...
+        '\n'            | ','             | '\0'       | '\0'       | false          || _ /* Default. */
+        '\n'            | ','             | '\0'       | '\0'       | true           || _ /* Headers. */
+        '\t'            | ','             | '\0'       | '\0'       | false          || _ /* Record Separators. */
+        '\r'            | ','             | '\0'       | '\0'       | false          || _
+        '<'             | ','             | '\0'       | '\0'       | false          || _
+        '>'             | ','             | '\0'       | '\0'       | false          || _
+        '&'             | ','             | '\0'       | '\0'       | false          || _
+        '\\'            | ','             | '\0'       | '\0'       | false          || _
+        ','             | '.'             | '\0'       | '\0'       | false          || _ /* Column separator. */
+//        | ','             | '\0'       | '\\'       | false          || _ /* Field quote. */
```
gapra-msft, line 197: i'm not sure why or how some of my other cases got deleted. I'll add them back. Also forgot to add to file

sdk/storage/azure-storage-file-datalake/src/main/java/com/azure/storage/file/datalake/models/FileQuerySerialization.java
gapra-msft: done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobQuerySerialization.java
gapra-msft: done

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/implementation/FluxInputStream.java
rickle-msft: I don't think we want to wrap IllegalArgumentException, etc. Probably shouldn't wrap any RuntimeExceptions, so rethrow IOExceptions and RuntimeExceptions, wrap HttpResponseExceptions. How does that sound?

gapra-msft: If we do that exactly though we would be silently ignoring any other Exception which would be really bad too

kasobol-msft: https://stackoverflow.com/questions/52195112/what-is-a-ioexception-and-how-do-i-fix-it/52195182
https://docs.oracle.com/javase/7/docs/api/java/io/IOException.html
https://stackoverflow.com/questions/13216148/java-what-throws-an-ioexception

The IOException is quite ambiguous. Do we know what are scenarios when RuntimeException is thrown here?

gapra-msft: I think it's dependent on the behavior of the Flux. In our primary use case, the Flux we care about is the one we return in blob.query - the code that handles that would take in the inputs, and it could return some type of error if the query failed (that would be IOException), or the input validation failed (IllegalArgument), etc

kasobol-msft: Summarizing our offline discussion. Anything thrown during pulling data is legit IOException. For IllegalArgument and it's family we should find a way to invoke validation eagerly (i.e. these exceptions should come from `query`/`queryWithResponse` and shouldn't be deferred until someone tries to read from stream).

gapra-msft: Resolveed by moving input validation further up by blocking on the network request

PR #10643, opened 2020-04-30
Nio input stream

rickle-msft:


sdk/storage/azure-storage-blob-nio/src/test/java/com/azure/storage/blob/nio/NioBlobInputStreamTest.groovy
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @Unroll
+    def "Read buff"() {
+        setup:
+        def nioBytes = new byte[size]
+        def fileBytes = new byte[size]
+
+        when:
+        nioStream.read(nioBytes)
```
gapra-msft, line 49: Does our implementation of BlobInputStream always give back the user as much data as they requested? Cause isnt it possible for an InputStream to return less than expected?

rickle-msft: You're right that the docs say it could return less, but our read method will always return the amount asked for unless there's an error or the end of the blob is hit. 

```
int chunks = (int) (Math.ceil((double) len / (double) this.chunkSize));
        int numOfBytesRead = 0;
        for (int i = 0; i < chunks; i++) {
            int results = this.readInternal(b, off + numOfBytesRead, len);
            if (results == -1) {
                return -1;
            }
            numOfBytesRead += results;
        }
        return numOfBytesRead;
```

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/NioBlobInputStream.java
```diff
@@ -0,0 +1,201 @@
...
+import com.azure.storage.blob.specialized.BlobInputStream;
+import com.azure.storage.common.implementation.Constants;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Provides an InputStream to read a file stored as an Azure Blob.
+ */
+public class NioBlobInputStream extends InputStream {
```
gapra-msft, line 16: Are you wrapping each method since you want to log using the NioBlobInputStream class? Just wondering

kasobol-msft: +1 I wonder what additional functionality `NioBlobInputStream` provides over `BlobInputStream`.

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/NioBlobInputStream.java
```diff
@@ -0,0 +1,201 @@
...
+     */
+    @Override
+    public int read() throws IOException {
+        try {
+            return this.blobInputStream.read();
+        } catch (RuntimeException e) {
+            if (e.getMessage().equals(Constants.STREAM_CLOSED)) {
+                throw LoggingUtility.logError(logger, new IOException(e));
+            }
+            throw LoggingUtility.logError(logger, e);
```
kasobol-msft, line 86: Shoundn't this already happen inside BlobInputStream ? If not would it be possible to push it down there instead of creating new type ?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/NioBlobInputStream.java
rickle-msft: There's no additional functionality right now, but I can't say for sure that usage patterns will be the same between nio and the sdk, and they may require swapping out the implementation for different performance characteristics or adding features we don't want in the sdk. Because of that, I think it's safer to define a type specific to nio that we can update independent of the underlying BlobInputStream

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/NioBlobInputStream.java
rickle-msft: BlobInputStream throws RuntimeExceptions, and the contract specifies that these methods are supposed to throw an IOException. I suspect that BlobInputStream was changed to throw RuntimeExceptions to work better with Reactor, but in an environment that isn't touched by async stuff, I think it's best to conform to the actual api conctract. 

I can look into skipping the double logging if you don't like that. It'll probably require a checkstyle suppression since CI doesn't usually let you use a raw throw statement without somehow passing it through a logger.

kasobol-msft: I'm not sure if BlobInputStream throwin RuntimeException is/was intentional. As you said the InputStream contract forces caller to handle IOException, so reactive code we have that's calling it should be handling/convering IOExceptions (otherwise compiler wouldn't be happy) - if that's the case I think we could safely make underlying stream confirming to the contract instead (e.g. throw IOException on certain response codes).

If we keep error translation here then I think we have more to cover . I.e. underlying stream can throw `new RuntimeException(UNEXPECTED_STREAM_READ_ERROR)` . Then, if the list of messages indicating IOException grows then we'll have a challage of maintaining parity between this class and underlying infra (which would be another argument to push this responsibility to underlying stream).

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/NioBlobInputStream.java
kasobol-msft: Hmm. Then the only reason this exist would be to reserve publicly available class name (+/- error translation depending on other discussion). Otherwise this could be [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) .

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
```diff
@@ -286,7 +286,7 @@ public int read(final byte[] b, final int off, final int len) throws IOException
         int chunks = (int) (Math.ceil((double) len / (double) this.chunkSize));
         int numOfBytesRead = 0;
         for (int i = 0; i < chunks; i++) {
-            int results = this.readInternal(b, off + numOfBytesRead, len);
+            int results = this.readInternal(b, off + numOfBytesRead, len - numOfBytesRead);
```
gapra-msft, line 289: nice catch!

gapra-msft: may not be relevant to this PR, but do we want to add tests for this fix?

sdk/storage/azure-storage-blob-nio/src/main/java/com/azure/storage/blob/nio/NioBlobInputStream.java
rickle-msft: I think changing the kinds of exceptions we throw is a breaking change, even if the customer theoretically should be handling it already, no?

kasobol-msft: Since we're talking about `RuntimeException` vs `IOException` I think it all boils down to how did we document `@throws` in existing streams.

For example [here](https://github.com/Azure/azure-sdk-for-java/blob/895bf6f96ca5926e6071f5055e95101e2cb133dc/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java#L228-L229) or [here](https://github.com/Azure/azure-sdk-for-java/blob/895bf6f96ca5926e6071f5055e95101e2cb133dc/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java#L193) the described contract is quite open. Moreover I'd say that if we throw `RuntimeExceptions` there on something that's "I/O related" then that should be bug (since we don't conform to our own documentation).

sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/StorageInputStream.java
rickle-msft: I added one to NioBlobInputStream. It looks like that may get moved into regular BlobInputStream, though.

rickle-msft:
@kasobol-msft #10740 can be used to track fixing BlobInputStream and hopefully removing NioBlobInputStream

PR #10646, opened 2020-04-30
Update CHECKENFORCER

mitchdenny:


eng/CHECKENFORCER
```diff
@@ -1,2 +1,2 @@
 format: v0.1-alpha
-minimumCheckRuns: 2
+minimumCheckRuns: 1
```
weshaggard, line 2: what is special about java that it only needs 1?

JimSuplizio: /CC @mitchdenny 
I believe this was initially set to 2 before it was counting the license/cla. Does the same change need to be made to the other languages?

PR #10195, opened 2020-04-14
Pluggable Modules for Avro Serialization

alzimmermsft:
This PR adds the `AvroSerializer` interface, with accompanying implementations using Apache and Jackson, which allows for abstraction of the Avro serialization implementations, and its dependencies, in our client libraries. In cases where a client library needs to interact with Avro serialization and de-serialization it should add an option to its builder to accept `AvroSerializer`, this will allow for client libraries to be agnostic to the underlying implementation. It is still undecided whether `AvroSerializer` will have a service provider interface (SPI) or libraries using Avro will take a dependency on one of the implementing libraries, which is overrideable if a customer wants to use another implementation.

**Example: EventHubs**

```java
EventHubClient eventHubClient = new EventHubClientBuilder()
    .connectionString(connectionString)
    .serializer(avroSerializer) // This is a potential representation of the change in a client library.
    .buildClient();
```

PR #10660, opened 2020-05-01
Add Schema Registry Client SDK

arerlend:
Hey Azure SDK folks - here's the initial add of the Schema Registry SDK.  Currently does not contain authorization component.

This PR contains no changes to EH Track 2 SDK, and has no references or changes to any Azure core code.

Will be working with Srikanta to get required infrastructure set up.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
+++ b/eng/code-quality-reports/src/main/resources/spotbugs/spotbugs-exclude.xml
@@ -1901,7 +1901,7 @@
     consistently instead of injecting null payload behavior from the payload. -->
   <Match>
     <Class name="com.azure.data.schemaregistry.avro.SchemaRegistryAvroSerializer"/>
-    <Method name="serializeSync"/>
+    <Method name="serialize"/>
     <Bug pattern="PZLA_PREFER_ZERO_LENGTH_ARRAYS"/>
   </Match>
 


sdk/schemaregistry/client/src/main/java/com/azure/schemaregistry/client/SchemaRegistryClient.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.io.IOException;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ *  Interface that defines operation for registering and fetching schemas and schema information to and from a
+ *  schema registry store.
+ */
+public interface SchemaRegistryClient {
```
srnagar, line 16: Clients have to be concrete classes instantiated from `ClientBuilders`. See [guidelines](https://azure.github.io/azure-sdk/java_design.html#service-clients) on creating service clients. Also, add the necessary annotations.

sdk/schemaregistry/client/src/main/java/com/azure/schemaregistry/client/rest/entities/responses/RegisterSchemaResponse.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.azure.schemaregistry.client.rest.entities.responses;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.IOException;
+
+public class RegisterSchemaResponse {
```
srnagar, line 14: Do you have a service swagger that can be used to generate clients using autorest?

arerlend: @hmlam

sdk/schemaregistry/client/src/main/java/com/azure/schemaregistry/client/rest/RestService.java
```diff
@@ -0,0 +1,376 @@
...
+import com.azure.schemaregistry.client.rest.entities.responses.SchemaObjectResponse;
+import com.azure.schemaregistry.client.rest.exceptions.RestClientException;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLSocketFactory;
```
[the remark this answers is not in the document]
arerlend: will be resolved by autorest generated HTTP client

sdk/schemaregistry/client/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+
+  <parent>
+    <groupId>com.azure</groupId>
+    <artifactId>schemaregistry-parent</artifactId>
+    <version>0.5-preview</version>
+  </parent>
+
+  <artifactId>schemaregistry-client</artifactId>
+  <packaging>jar</packaging>
+  <version>0.5-preview</version>
```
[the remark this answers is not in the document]
arerlend: updated

sdk/schemaregistry/client/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+  <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+  </properties>
+  
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>3.8.1</version>
```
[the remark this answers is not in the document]
JimSuplizio: To add to what @srnagar said. The [analyze step is failing](https://dev.azure.com/azure-sdk/public/_build/results?buildId=397954&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=bdeefc16-b669-5ebd-ad94-a2c19ade53b0&l=75) because of the following client libraries need to be defined in [version_client.txt](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/version_client.txt):
azure-schemaregistry-client
azure-schemaregistry-serde-avro
azure-schemaregistry-serde-common
The pom files need the appropriate tags and for the above libraries, within the sdk/schemaregistry they'll all be the current version. You'll also need to add the libraries to the [eng/jacoco-test-coverage](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/jacoco-test-coverage/pom.xml
)
The following [external_dependencies](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/external_dependencies.txt) need some work.
You'd added org.apache.avro:avro and tagged the bannedDependencies correctly in the pom files however the version tags for the dependencies, themselves are missing. 
Similarly, org.easymock:easymock needs to get added to external_dependencies and correctly tagged within the pom files. 
Last but not least, junit:junit dependencies need to be tagged but you're using version 4.12 and it would be great if, instead of declaring a one off, you could use the common 4.13-beta-3 version that's already defined.

Fixing the above issues should allow the analyze step to pass as well as the From Source runs. Note: You can run the [pom_file_version_scanner.ps1](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/versioning/pom_file_version_scanner.ps1#L30) locally from a pwsh or powershell window. The line I linked to has the exact command and it takes like 5 seconds to run.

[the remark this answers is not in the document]
arerlend: Thanks for the comment Jim.  I'm still in the process of fixing all of Srikanta's comments but this will be helpful.

I'll also be replacing any easymock use with mockito as Srikanta requested.  I've also decided to just convert all junit 4 classes to use junit-jupiter-api.

sdk/schemaregistry/client/src/main/java/com/azure/schemaregistry/client/MockSchemaRegistryClient.java
```diff
@@ -0,0 +1,75 @@
...
+
+package com.azure.schemaregistry.client;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.function.Function;
+
+public class MockSchemaRegistryClient implements SchemaRegistryClient {
```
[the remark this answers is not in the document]
arerlend: Moved MockSchemaRegistryClient to test directory of serde-common, where it's actually used.

sdk/schemaregistry/client/src/main/java/com/azure/schemaregistry/client/SRObject.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.nio.charset.Charset;
+import java.util.function.Function;
+
+/**
+ * Stores all relevant information returned from SchemaRegistryClient layer.
+ *
+ * @param <T> is derived from the parser function that is passed in the constructor.
+ */
+public class SRObject<T> {
```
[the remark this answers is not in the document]
arerlend: Renamed.  Also took out generic parameter since it is unnecessary.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,281 @@
...
+    public synchronized void reset() {
+        guidCache.clear();
+        schemaStringCache.clear();
+        typeParserDictionary.clear();
+    }
+
+    /**
+     * Checks if caches should be reinitialized to satisfy initial configuration
+     */
+    private synchronized void resetIfNeeded() {
```
alzimmermsft, line 255: Has a least recently used cache been thought about? Instead of clearing everything when we hit the limit just begin prune stale entries. Current logic will work but for hot path schemas they'll need to go make a network call when this case happens. 

Overall, this doesn't need to be prioritized now as 1000 entries, by default, seems to be a lot of head space.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,281 @@
...
+public class CachedSchemaRegistryClient implements SchemaRegistryClient {
+    private final ClientLogger logger = new ClientLogger(CachedSchemaRegistryClient.class);
+
+    public static final int MAX_SCHEMA_MAP_SIZE_DEFAULT = 1000;
+    public static final Charset SCHEMA_REGISTRY_SERVICE_ENCODING = StandardCharsets.UTF_8;
+    static final int MAX_SCHEMA_MAP_SIZE_MINIMUM = 10;
+
+    private final AzureSchemaRegistryRestService restService;
+    private final int maxSchemaMapSize;
+    private final HashMap<String, Function<String, Object>> typeParserDictionary;
```
alzimmermsft, line 50: I see that `typeParserDictionary` generally has it's key lower cased to allow for case insensitive matching, I believe using a `TreeMap` should cover that case without needing to manually manage key casing on accesses. Using `new TreeMap(String::equalsIgnoreCase)` should produce the same effect.

https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html#TreeMap-java.util.Comparator-

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClientBuilder.java
```diff
@@ -0,0 +1,284 @@
...
+            final List<HttpPipelinePolicy> policies = new ArrayList<>();
+
+            policies.add(new UserAgentPolicy(httpLogOptions.getApplicationId(), clientName, clientVersion,
+                Configuration.getGlobalConfiguration().clone()));
+            policies.add(new RequestIdPolicy());
+            policies.add(new AddHeadersPolicy(this.headers));
+
+            HttpPolicyProviders.addBeforeRetryPolicies(policies);
+
+            policies.add(retryPolicy);
```
alzimmermsft, line 254: Need to `null` check the `retryPolicy` here and use the default. Another option is `null` checking when setting the retry policy.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClientBuilder.java
```diff
@@ -0,0 +1,284 @@
...
+        this.typeParserDictionary = new HashMap<>();
+        this.httpClient = null;
+        this.credential = null;
+        this.retryPolicy = new RetryPolicy("retry-after-ms", ChronoUnit.MILLIS);
+
+        Map<String, String> properties = CoreUtils.getProperties(CLIENT_PROPERTIES);
+        clientName = properties.getOrDefault(NAME, "UnknownName");
+        clientVersion = properties.getOrDefault(VERSION, "UnknownVersion");
+
+        this.headers = new HttpHeaders();
```
alzimmermsft, line 100: Is this ever mutated? If there are plans to add static headers let's leave it, but otherwise we can just remove it.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClientBuilder.java
```diff
@@ -0,0 +1,284 @@
...
+
+            policies.add(retryPolicy);
+
+            policies.add(new AddDatePolicy());
+            // Authentications
+            if (credential != null) {
+                // User token based policy
+                policies.add(new BearerTokenAuthenticationPolicy(credential, DEFAULT_SCOPE));
+            } else {
+                // Throw exception that credential and tokenCredential cannot be null
```
alzimmermsft, line 262: Let's check for this when starting the construction of the `HttpPipeline`, no need to waste time to just throw.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/SchemaRegistryObject.java
```diff
@@ -0,0 +1,79 @@
...
+                    this.schemaType, this.schemaId)));
+        }
+
+        if (this.deserialized == null) {
+            logger.verbose(
+                String.format("Deserializing schema, id: %s, schema string %s",
+                    this.schemaId,
+                    new String(this.schemaByteArray, CachedSchemaRegistryClient.SCHEMA_REGISTRY_SERVICE_ENCODING)));
+            this.deserialized = parseMethod.apply(
+                new String(this.schemaByteArray, CachedSchemaRegistryClient.SCHEMA_REGISTRY_SERVICE_ENCODING));
```
alzimmermsft, line 75: Let's initialize this String once before any potential logging.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,281 @@
...
+ * @see SchemaRegistryClient Implements SchemaRegistryClient interface to allow for testing with mock
+ * @see CachedSchemaRegistryClientBuilder Follows builder pattern for object instantiation
+ */
+@ServiceClient(
+    builder = CachedSchemaRegistryClientBuilder.class,
+    serviceInterfaces = AzureSchemaRegistryRestService.class)
+public class CachedSchemaRegistryClient implements SchemaRegistryClient {
+    private final ClientLogger logger = new ClientLogger(CachedSchemaRegistryClient.class);
+
+    public static final int MAX_SCHEMA_MAP_SIZE_DEFAULT = 1000;
```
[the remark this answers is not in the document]
arerlend: Changed to default

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,281 @@
...
+        if (registryUrl == null || registryUrl.isEmpty()) {
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("Schema Registry URL cannot be null or empty."));
+        }
+
+        if (maxSchemaMapSize < MAX_SCHEMA_MAP_SIZE_MINIMUM) {
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException(
+                    String.format("Max schema map size must be greater than %d schemas", MAX_SCHEMA_MAP_SIZE_MINIMUM)));
+        }
```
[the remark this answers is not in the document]
arerlend: Yeah this is a redundant check, already in the builder.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClientBuilder.java
arerlend: Ok let's take it out for now.  We had some ideas on adding headers for internal testing, but we don't have any real use cases for clients adding their own headers to HTTP calls.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/SchemaRegistryObject.java
```diff
@@ -0,0 +1,79 @@
...
+     *  object is stored and returned.
+     *
+     *  @return schema object, deserialized using stored schema parser method.
+     */
+    public Object deserialize() {
+        if (parseMethod == null) {
+            throw logger.logExceptionAsError(new SchemaRegistryClientException(
+                String.format("No loaded parser for %s format. Schema guid: %s",
+                    this.schemaType, this.schemaId)));
+        }
```
srnagar, line 67: Shouldn't this check happen in the constructor itself?

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/SchemaRegistryObject.java
```diff
@@ -0,0 +1,79 @@
...
+
+        if (this.deserialized == null) {
```
srnagar, line 69: You might want to synchronize access to this.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/rest/AzureSchemaRegistryRestServiceClientBuilder.java
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.schemaregistry.client.rest;
```
srnagar, line 4: This should be in `implementation` package as it's the convention we follow to denote packages that users shouldn't use.
https://azure.github.io/azure-sdk/java_design.html#example-namespaces

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/rest/AzureSchemaRegistryRestService.java
```diff
@@ -0,0 +1,1411 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.schemaregistry.client.rest;
```
srnagar, line 4: This should not be used directly by the user. So, move this to `implementation` package. Same for any other classes that shouldn't in public API view.

sdk/schemaregistry/azure-schemaregistry-client/src/main/module-info.java
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module com.azure.schemaregistry.client {
+    requires transitive com.azure.core;
+
+    exports com.azure.schemaregistry.client;
+    exports com.azure.schemaregistry.client.rest;
```
srnagar, line 8: Does this package have to be exported? It contains codgen rest service client and builder which shouldn't be used anywhere outside this module.

sdk/schemaregistry/azure-schemaregistry-client/swagger.yaml
```diff
@@ -0,0 +1,378 @@
+openapi: 3.0.0
```
srnagar, line 1: Move this inside a directory called `swagger`.

sdk/schemaregistry/azure-schemaregistry-client/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+    <jacoco.min.linecoverage>0.10</jacoco.min.linecoverage>
+    <jacoco.min.branchcoverage>0.01</jacoco.min.branchcoverage>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core</artifactId>
+      <version>1.5.0</version> <!-- {x-version-update;com.azure:azure-core;dependency} -->
+    </dependency>
```
srnagar, line 51: We make all HTTP based clients use `azure-core-http-netty` as the default http client plugin. So, add that dependency as well so the service loaded can automatically create the http client.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
arerlend: Yeah I thought about this but figured it wasn't worth adding.  If an app has more than 1k entries, the app is going to have a lot more problems than just the cache getting fully cleared.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,277 @@
...
+    private final ClientLogger logger = new ClientLogger(CachedSchemaRegistryClient.class);
+
+    public static final Charset SCHEMA_REGISTRY_SERVICE_ENCODING = StandardCharsets.UTF_8;
+
+    static final int MAX_SCHEMA_MAP_SIZE_DEFAULT = 1000;
+    static final int MAX_SCHEMA_MAP_SIZE_MINIMUM = 10;
+
+    private final AzureSchemaRegistryRestService restService;
+    private final int maxSchemaMapSize;
+    private final HashMap<String, Function<String, Object>> typeParserDictionary;
```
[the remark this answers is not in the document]
arerlend: Changed to typeParserMap.  I will use ConcurrentHashMap here.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,277 @@
...
+        }
+        this.typeParserDictionary.put(schemaType.toLowerCase(Locale.ENGLISH), parseMethod);
+        logger.verbose(
+            "Loaded parser for '{}' serialization format.", schemaType.toLowerCase(Locale.ENGLISH));
+    }
+
+    @Override
+    public synchronized SchemaRegistryObject register(
+        String schemaGroup, String schemaName, String schemaString, String serializationType)
+        throws SchemaRegistryClientException {
```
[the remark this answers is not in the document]
arerlend: Yeah this is an artifact from when it was a checked exception.  Will remove these

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,277 @@
...
+    public synchronized void loadSchemaParser(String schemaType, Function<String, Object> parseMethod) {
+        if (schemaType == null || schemaType.isEmpty()) {
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("Serialization type cannot be null or empty."));
+        }
+        if (this.typeParserDictionary.containsKey(schemaType.toLowerCase(Locale.ENGLISH))) {
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("Multiple parse methods for single serialization type may not be added."));
+        }
+        this.typeParserDictionary.put(schemaType.toLowerCase(Locale.ENGLISH), parseMethod);
```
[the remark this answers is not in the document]
arerlend: Yeah I think I am going to make all three ConcurrentHashMaps so I can remove `synchronized`

In this case I prefer to throw if multiple parsers are added for the same serialization type, just because upserting could potentially result in a functioning app with unexpected behaviors.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,277 @@
...
+        logger.verbose(
+            "Loaded parser for '{}' serialization format.", schemaType.toLowerCase(Locale.ENGLISH));
+    }
+
+    @Override
+    public synchronized SchemaRegistryObject register(
```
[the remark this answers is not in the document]
arerlend: Removed.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/SchemaRegistryClientException.java
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.schemaregistry.client;
+
+/**
+ * Runtime exception to be returned from SchemaRegistryClient implementations.
+ */
+public class SchemaRegistryClientException extends RuntimeException {
```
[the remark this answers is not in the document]
arerlend: Changed to use AzureException

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,277 @@
...
+        logger.verbose(String.format("Cached schema string. Group: '%s', name: '%s'", schemaGroup, schemaName));
+        return schemaId.getId();
+    }
+
+    @Override
+    public String deleteSchemaVersion(String schemaGroup, String schemaName, int version)
+        throws SchemaRegistryClientException {
+        // return this.restService.deleteSchemaVersion(schemaName, version);
+        // remove from cache
+        return null;
```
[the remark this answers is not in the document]
arerlend: Throwing unsupported operation

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/SchemaRegistryObject.java
arerlend: Parsing is idempotent so multiple writes won't cause any problems

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/SchemaRegistryObject.java
arerlend: Null check was removed, parseMethod can never be passed in as null since CachedSchemaRegistryClient throws if no parser is stored for the given schema type

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/rest/AzureSchemaRegistryRestServiceClientBuilder.java
arerlend: Moved to implementation

sdk/schemaregistry/azure-schemaregistry-client/src/main/module-info.java
arerlend: Removed rest and rest.models from module-info.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/SchemaRegistryClient.java
```diff
@@ -0,0 +1,99 @@
...
+     * Registers a schema against backing schema registry store.
+     *
+     * @param schemaGroup schema group name
+     * @param schemaName schema name
+     * @param schemaString string representation of schema
+     * @param serializationType string representation of serialization format type
+     * @return SchemaRegistryObject containing information regarding registered schema.
+     * @throws SchemaRegistryClientException if registration operation fails
+     */
+    SchemaRegistryObject register(String schemaGroup, String schemaName, String schemaString, String serializationType)
```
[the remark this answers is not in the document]
arerlend: Schema registration is a bit of a special case - if schema exists, it returns the existing GUID.  Doesn't fit very cleanly into the categories in the guidelines

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClientBuilder.java
```diff
@@ -0,0 +1,284 @@
...
+     * @param parseMethod function for deserializing registry-stored schema strings to Java Objects
+     * @return The updated {@link CachedSchemaRegistryClientBuilder} object.
+     */
+    public CachedSchemaRegistryClientBuilder loadSchemaParser(
+        String schemaType, Function<String, Object> parseMethod) {
+        if (schemaType == null || schemaType.isEmpty()) {
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException("Serialization type cannot be null or empty."));
+        }
+        if (this.typeParserDictionary.containsKey(schemaType.toLowerCase(Locale.ENGLISH))) {
```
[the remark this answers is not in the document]
arerlend: We should avoid this use case - user intent when calling this method is to load (add) schema parser, not to upsert.  Overwriting previous writes may result in unpredictable behavior.

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
arerlend: Using ConcurrentSkipListMap<>(String.CASE_INSENSITIVE_ORDER)

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClientBuilder.java
```diff
@@ -0,0 +1,284 @@
...
+
+    /**
+     * Sets the service endpoint for the Azure Schema Registry instance.
+     * Supplies client defaults.
+     *
+     * @param schemaRegistryUrl The URL of the Azure Schema Registry instance
+     * @throws NullPointerException if {@code schemaRegistryUrl} is null
+     * @throws IllegalArgumentException if {@code schemaRegistryUrl} cannot be parsed into a valid URL.
+     */
+    public CachedSchemaRegistryClientBuilder(String schemaRegistryUrl) {
```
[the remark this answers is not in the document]
arerlend: Moved URL to .host()

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
```diff
@@ -0,0 +1,277 @@
...
+        AzureSchemaRegistryRestService restService,
+        HashMap<String, SchemaRegistryObject> guidCache,
+        HashMap<String, SchemaRegistryObject> schemaStringCache,
+        HashMap<String, Function<String, Object>> typeParserDictionary) {
+        this.restService = restService; // mockable
+        this.guidCache = guidCache;
+        this.schemaStringCache = schemaStringCache;
+        this.typeParserDictionary = typeParserDictionary;
+        this.maxSchemaMapSize = MAX_SCHEMA_MAP_SIZE_DEFAULT;
+    }
```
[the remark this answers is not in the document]
arerlend: As per discussion, will be fixed after private preview.  Will convert to use mockito

sdk/schemaregistry/azure-schemaregistry-client/src/main/java/com/azure/schemaregistry/client/CachedSchemaRegistryClient.java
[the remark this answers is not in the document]
arerlend: Correction - has to be public.  'serde-avro' module requires that defaults be set in serializer and deserializer builder class.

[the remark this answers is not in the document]
arerlend: *only default needs to be public.

eng/versioning/version_client.txt
```diff
@@ -26,6 +26,8 @@ com.azure:azure-identity-perf;1.0.0-beta.1;1.0.0-beta.1
 com.azure:azure-messaging-eventhubs;5.1.0;5.2.0-beta.1
 com.azure:azure-messaging-eventhubs-checkpointstore-blob;1.1.0;1.2.0-beta.1
 com.azure:azure-messaging-servicebus;7.0.0-beta.2;7.0.0-beta.3
+com.azure:azure-data-schemaregistry;1.0.0-beta.1;1.0.0-beta.1
```
JimSuplizio, line 29: nit: please put these in alphabetical order, they should be immediately after com.azure:azure-data-appconfiguration

sdk/schemaregistry/azure-data-schemaregistry-avro/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    </site>
+  </distributionManagement>
+
+  <scm>
+    <url>scm:git:https://github.com/Azure/azure-sdk-for-java</url>
+    <connection>scm:git:git@github.com:Azure/azure-sdk-for-java.git</connection>
+    <tag>HEAD</tag>
+  </scm>
+
+  <properties>
```
JimSuplizio, line 40: Please put a comment in here something like `<!-- SchemaRegistrySkip temporarily relaxing code coverage -->` to indicate that the coverage skip is temporary and file an issue to remove this entry when you've got tests/coverage numbers.

sdk/schemaregistry/azure-data-schemaregistry/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+    </site>
+  </distributionManagement>
+
+  <scm>
+    <url>scm:git:https://github.com/Azure/azure-sdk-for-java</url>
+    <connection>scm:git:git@github.com:Azure/azure-sdk-for-java.git</connection>
+    <tag>HEAD</tag>
+  </scm>
+
+  <properties>
```
JimSuplizio, line 41: Same comment as sdk/schemaregistry/azure-data-schemaregistry-avro/pom.xml, add the same comment here and ensure that issue filed includes both libraries.

sdk/schemaregistry/azure-data-schemaregistry/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-enforcer-plugin;external_dependency} -->
+        <configuration>
+          <rules>
+            <bannedDependencies>
+              <includes>
+                <include>com.azure:*</include>
```
JimSuplizio, line 100: You shouldn't need this if your only non-test dependencies are com.azure, it'll inherit that from azure-client-sdk-parent

arerlend:
https://github.com/Azure/azure-sdk-for-java/issues/11584

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/module-info.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module com.azure.data.schemaregistry.serde.avro {
+    requires transitive com.azure.core;
+
+    exports com.azure.data.schemaregistry.serde.avro;
```
alzimmermsft, line 7: This package doesn't exist, this should be `com.azure.data.schemaregistry.avro` right?

sdk/schemaregistry/azure-data-schemaregistry-avro/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <version>5.6.2</version> <!-- {x-version-update;org.junit.jupiter:junit-jupiter-params;external_dependency} -->
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
```
alzimmermsft, line 82: If you look at this PR (https://github.com/Azure/azure-sdk-for-java/pull/10195/files#diff-f42ef09d2743baecc7f1c205f31c0351R113) it includes a plugin which can generate Java source code from given schemas, should be helpful for testing.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroAsyncDeserializer.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.concurrent.Executors;
+
+/**
+ * Asynchronous registry-based deserializer implementation.
+ */
+public class SchemaRegistryAvroAsyncDeserializer {
+    private static final int DEFAULT_THREAD_POOL_SIZE = 8;
+
+    private final SchemaRegistryAvroDeserializer deserializer;
+    private final Scheduler scheduler;
```
alzimmermsft, line 20: Is there a specific reason we are creating a custom scheduler for deserialization?

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroAsyncSerializer.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Async wrapper around sync serialization operation
+     *
+     * @param object object to be serialized to bytes
+     * @return Avro byte representation of object
+     * @throws SerializationException upon serialization operation failure
+     */
+    public Mono<byte[]> serializeAsync(Object object) throws SerializationException {
+        if (object == null) {
+            return Mono.empty();
```
alzimmermsft, line 40: Should we use `Mono.empty()` or `Mono.just(new byte[0])`? These will have two very different things happen when dealing with Reactor streams. The former will complete the stream without an `onNext` operation, the latter will trigger an `onNext` (triggering downstream `map` and `flatMap` operators) and will contain an empty `byte[]` as expected when Avro serializes a `null`.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroDeserializer.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Deserializes byte array into Java object using payload-specified schema.
+     *
+     * @param data Byte array containing serialized bytes
+     * @return Java object.
+     *
+     * Object type is testable with instanceof operator.  Return value can be casted in the caller layer.
+     *
+     * @throws SerializationException Throws on deserialization failure.
+     * Exception may contain inner exceptions detailing failure condition.
```
alzimmermsft, line 44: For these Javadocs is the but about `instanceof` meant to go into the overview of the method or specific to the `return` portion? Also, Javadocs are HTML so any newlines are scrubbed, if you need line separators you'll need to add `<p>` where newlines are (except when transitioning Javadoc sections such as `param` to `param`, `return` to `throws`, etc.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroDeserializer.java
```diff
@@ -0,0 +1,49 @@
...
+import com.azure.data.schemaregistry.SerializationException;
+import com.azure.data.schemaregistry.client.CachedSchemaRegistryClientBuilder;
+
+/**
+ * A deserializer implementation capable of automatedly deserializing encoded byte array payloads into Java objects by
+ * fetching payload-specified schemas from the Azure Schema Registry store.
+ *
+ * SchemaRegistryAvroDeserializer instances should be built using the static Builder class.
+ *
+ * Pluggable with the core Azure SDK Deserializer interface.
```
alzimmermsft, line 17: Need to add `<p>` tags to have Javadocs generate with line separators.

```suggestion
 * A deserializer implementation capable of automatedly deserializing encoded byte array payloads into Java objects by
 * fetching payload-specified schemas from the Azure Schema Registry store.
 * <p>
 * SchemaRegistryAvroDeserializer instances should be built using the static Builder class.
 * <p>
 * Pluggable with the core Azure SDK Deserializer interface.
```

sdk/schemaregistry/azure-data-schemaregistry/src/main/java/com/azure/data/schemaregistry/AbstractDataDeserializer.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Common implementation for all registry-based deserializers.
+ */
+public abstract class AbstractDataDeserializer extends AbstractDataSerDe {
+    private final ClientLogger logger = new ClientLogger(AbstractDataDeserializer.class);
+
+    protected final Map<String, ByteDecoder> byteDecoderMap = new ConcurrentHashMap<>();
```
alzimmermsft, line 23: Was this made `protected` to sub-classes could interact with it? Generally we stay away from protected as it is part of the public API.

sdk/schemaregistry/azure-data-schemaregistry/src/main/java/com/azure/data/schemaregistry/AbstractDataDeserializer.java
```diff
@@ -0,0 +1,127 @@
...
+
+        try {
+            registryObject = this.schemaRegistryClient.getSchemaByGuid(schemaGuid);
+            payloadSchema = registryObject.deserialize();
+        } catch (SchemaRegistryClientException e) {
+            throw logger.logExceptionAsError(
+                new SerializationException(String.format("Failed to retrieve schema for id %s", schemaGuid), e));
+        }
+
+        // TODO: how to handle unknown formats
```
alzimmermsft, line 64: I think throwing is the correct choice here, if we can't determine the schema there isn't much we can do.

sdk/schemaregistry/azure-data-schemaregistry/src/main/java/com/azure/data/schemaregistry/ByteEncoder.java
```diff
@@ -0,0 +1,40 @@
...
+
+    // TODO: Method does not currently require schema object to be passed since schemas can be derived from
+    // Avro objects. JSON implementation would be the same.
+    /**
+     * Converts object into stream containing the encoded representation of the object.
+     * @param object Object to be encoded into byte stream
+     * @return output stream containing byte representation of object
+     * @throws SerializationException if generating byte representation of object fails
+     */
+    ByteArrayOutputStream encode(Object object) throws SerializationException;
```
alzimmermsft, line 39: Should this be switched to `OutputStream` in case we want to change the internal return type in the future.

sdk/schemaregistry/azure-data-schemaregistry/swagger/swagger.yaml
```diff
@@ -0,0 +1,378 @@
+openapi: 3.0.0
```
alzimmermsft, line 1: Should the swagger be merged into `azure-sdk-for-java`? I thought this was meant to go into `azure-rest-api-specs`.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/module-info.java
arerlend: Yeah I missed this, there have been a lot of package changes.

sdk/schemaregistry/azure-data-schemaregistry/src/main/java/com/azure/data/schemaregistry/AbstractDataDeserializer.java
arerlend: Yeah the question here is whether or not we should offer a config to skip the messages where schema was not recognized.  Right now it will throw at the registry client level since there no schema parser will have been loaded.  I think we decided to punt the final decision after preview but I am leaning heavily towards throwing.

arerlend: Going to just take the comment out.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroAsyncDeserializer.java
arerlend: Async implementation is using a threadpool to do sync work, hence the scheduler with the dedicated threadpool

arerlend: This will be changed for GA.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroSchemaUtils.java
```diff
@@ -0,0 +1,87 @@
...
+        } else if (object instanceof byte[]) {
+            return PRIMITIVE_SCHEMAS.get("Bytes");
+        } else if (object instanceof GenericContainer) {
+            return ((GenericContainer) object).getSchema();
+        } else {
+            ClientLogger logger = new ClientLogger(AvroSchemaUtils.class);
+            throw logger.logExceptionAsError(
+                new IllegalArgumentException(
+                "Unsupported Avro type. Supported types are null, Boolean, Integer, Long, "
+                    + "Float, Double, String, byte[] and IndexedRecord"));
```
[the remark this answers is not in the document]
arerlend: https://avro.apache.org/docs/1.9.2/api/java/org/apache/avro/generic/IndexedRecord.html

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroSchemaUtils.java
```diff
@@ -0,0 +1,87 @@
...
+            return PRIMITIVE_SCHEMAS.get("Integer");
+        } else if (object instanceof Long) {
+            return PRIMITIVE_SCHEMAS.get("Long");
+        } else if (object instanceof Float) {
+            return PRIMITIVE_SCHEMAS.get("Float");
+        } else if (object instanceof Double) {
+            return PRIMITIVE_SCHEMAS.get("Double");
+        } else if (object instanceof CharSequence) {
+            return PRIMITIVE_SCHEMAS.get("String");
+        } else if (object instanceof byte[]) {
```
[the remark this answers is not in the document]
arerlend: Yes it should.  added

sdk/schemaregistry/azure-data-schemaregistry/swagger/swagger.yaml
arerlend: @srnagar

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroAsyncSerializer.java
arerlend: Leaving this as is after discussion with Alan.

Thinking about this in terms of a user scenario makes it clear that returning empty byte array has a different meaning from returning null.  batch.tryAdd(null) throws.  batch.tryAdd(byte[0]) would not.  If there was an intermediate serialization step when doing batch.tryAdd(null), then the non-null contract breaks.

sdk/schemaregistry/azure-data-schemaregistry/src/main/java/com/azure/data/schemaregistry/AbstractDataDeserializer.java
arerlend: Updated to be private.  All access to this map is controlled via loadByteDecoder()

sdk/schemaregistry/azure-data-schemaregistry-avro/pom.xml
arerlend: Yeah I will add tests for this in the future.  We are planning to swap out for the pluggable implementation so testing is fairly lax here.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroSchemaUtils.java
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @param object object for which Avro schema is being returned
+     * @return Avro schema for object's data structure
+     *
+     * @throws IllegalArgumentException if object type is unsupported
+     */
+    public static Schema getSchema(Object object) throws IllegalArgumentException {
+        if (object == null) {
+            return PRIMITIVE_SCHEMAS.get("Null");
+        } else if (object instanceof Boolean) {
```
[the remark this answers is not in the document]
arerlend: This is for top level Object parameter, so they cannot be passed in as primitives.  Avro will handle if there are primitives as fields on the object.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroByteEncoder.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Returns schema name for storing schemas in schema registry store.
+     *
+     * @param object Schema object used to generate schema path
+     * @return schema name as string
+     */
+    public String getSchemaName(Object object) {
+        return AvroSchemaUtils.getSchema(object).getFullName();
+    }
```
srnagar, line 51: Both these methods can be moved to `AvroSchemaUtils` and can be made static.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroByteEncoder.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Returns ByteArrayOutputStream containing Avro encoding of object parameter
+     * @param object Object to be encoded into byte stream
+     * @return closed ByteArrayOutputStream
+     * @throws SerializationException wraps runtime exceptions
+     */
+    public ByteArrayOutputStream encode(Object object) throws SerializationException {
```
srnagar, line 59: Same here - `throws SerializationException` is not required.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroByteDecoder.java
```diff
@@ -0,0 +1,83 @@
...
+    public Object decodeBytes(byte[] b, Object object) throws SerializationException {
+        if (!(object instanceof Schema)) {
+            throw logger.logExceptionAsError(
+                new SerializationException("Attempted to decode with non-Avro schema object in AvroByteDecoder"));
+        }
+        Schema schema = (Schema) object;
+
+        if (schema.getType().equals(Schema.Type.BYTES)) {
+            return b;
+        }
```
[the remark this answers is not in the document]
arerlend: I think that is the expectation on how to handle byte-only payloads coming back from a message body.  Avro cares about the length of a byte array when working within a struct because it needs to know how long byte arrays are within the data structure so it doesn't read past the end of a byte array.  This specific case is for payloads that consist of nothing but a single byte array.  Avro can just read the entire payload into a buffer and return it to the user to manage.

I am adding a //todo to revisit this.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroByteEncoder.java
```diff
@@ -0,0 +1,85 @@
...
+                DatumWriter<Object> writer;
+                if (object instanceof SpecificRecord) {
+                    writer = new SpecificDatumWriter<>(schema);
+                } else {
+                    writer = new GenericDatumWriter<>(schema);
+                }
+                writer.write(object, encoder);
+                encoder.flush();
+            }
+            out.close();
```
srnagar, line 77: Don't need to close ByteArrayOutputStream.

From [javadoc](https://docs.oracle.com/javase/8/docs/api/java/io/ByteArrayOutputStream.html#close--):
> Closing a ByteArrayOutputStream has no effect. The methods in this class can be called after the stream has been closed without generating an IOException.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroCodec.java
```diff
@@ -0,0 +1,24 @@
...
+package com.azure.data.schemaregistry.avro;
+
+import com.azure.data.schemaregistry.Codec;
+import org.apache.avro.Schema;
+
+/**
+ * Base Codec class for Avro encoder and decoder implementations
+ */
+abstract class AvroCodec implements Codec {
+    public String schemaType() {
```
srnagar, line 13: Add @Override annotation for both these methods.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroByteDecoder.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.avro.io.DecoderFactory;
+import org.apache.avro.specific.SpecificDatumReader;
+
+import java.io.IOException;
+
+/**
+ * ByteDecoder implementation with all Avro-specific functionality required to deserialize byte arrays
+ * given an Avro schema.
+ */
+public class AvroByteDecoder extends AvroCodec
```
srnagar, line 21: Name this class `ApacheAvroByteDecoder` so that  we can support Jackson avro later if required. (If this is going away after switching to azure-core implementation, then you can leave it as-is).

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroByteEncoder.java
```diff
@@ -0,0 +1,85 @@
...
+     * @return closed ByteArrayOutputStream
+     * @throws SerializationException wraps runtime exceptions
+     */
+    public ByteArrayOutputStream encode(Object object) throws SerializationException {
+        Schema schema = AvroSchemaUtils.getSchema(object);
+
+        try {
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            if (object instanceof byte[]) {
+                out.write((byte[]) object);
```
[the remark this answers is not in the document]
arerlend: Yup, see comments there.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroAsyncDeserializer.java
```diff
@@ -0,0 +1,43 @@
...
+        this.scheduler = Schedulers.fromExecutor(Executors.newFixedThreadPool(DEFAULT_THREAD_POOL_SIZE));
+    }
+
+    /**
+     * Async wrapper around synchronous deserialization method
+     * @param data bytes containing schema ID and encoded byte representation of object
+     * @return Mono wrapper around deserialized object
+     * @throws SerializationException if deserialization operation fails
+     */
+    public Mono<Object> deserializeAsync(byte[] data) throws SerializationException {
```
srnagar, line 38: Don't need `async` suffix since the class is called `SchemaRegistryAvroAsyncDeserializer`

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroAsyncSerializer.java
srnagar: Agree. This should return `Mono.just(new byte[0])`

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroDeserializerBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+    private boolean avroSpecificReader;
+    private Integer maxSchemaMapSize;
+
+    /**
+     * Instantiates instance of Builder class.
+     * Supplies client defaults.
+     *
+     * @param registryUrl base schema registry URL for storing and fetching schemas
+     */
+    public SchemaRegistryAvroDeserializerBuilder(String registryUrl) {
```
srnagar, line 24: Consider using no-args ctor for builder and have a setter method for registryUrl

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroDeserializerBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Construct instance of {@link SchemaRegistryAvroDeserializer}
+     *
+     * @return {@link SchemaRegistryAvroDeserializer} instance
+     *
+     * @throws NullPointerException if parameters are incorrectly set.
+     * @throws IllegalArgumentException if credential is not set.
+     */
+    public SchemaRegistryAvroDeserializer buildSyncClient() {
```
srnagar, line 83: Check if `credential` and `registryUrl` are null before creating the `Deserializer`.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroSerializerBuilder.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Instantiates {@link SchemaRegistryAvroSerializer}
+     * @return {@link SchemaRegistryAvroSerializer} instance
+     *
+     * @throws NullPointerException if parameters are incorrectly set.
+     * @throws IllegalArgumentException if credential is not set.
+     */
+    public SchemaRegistryAvroSerializer buildSyncClient() {
```
srnagar, line 104: Validate inputs to check if all required fields like registryUrl and credential are set.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/module-info.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module com.azure.data.schemaregistry.serde.avro {
```
srnagar, line 4: This should be named as `com.azure.data.schemaregistry.avro`

sdk/schemaregistry/azure-data-schemaregistry/src/main/java/com/azure/data/schemaregistry/client/CachedSchemaRegistryClientBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+    /**
+     * Sets the service endpoint for the Azure Schema Registry instance.
+     *
+     * @return The updated {@link CachedSchemaRegistryClientBuilder} object.
+     * @param schemaRegistryUrl The URL of the Azure Schema Registry instance
+     * @throws NullPointerException if {@code schemaRegistryUrl} is null
+     * @throws IllegalArgumentException if {@code schemaRegistryUrl} cannot be parsed into a valid URL
+     */
+    public CachedSchemaRegistryClientBuilder endpoint(String schemaRegistryUrl) {
+        Objects.requireNonNull(schemaRegistryUrl, "'endpoint' cannot be null.");
```
srnagar, line 91:
```suggestion
        Objects.requireNonNull(schemaRegistryUrl, "'schemaRegistryUrl' cannot be null.");
```

sdk/schemaregistry/azure-data-schemaregistry/src/main/java/com/azure/data/schemaregistry/client/CachedSchemaRegistryClientBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+
+        try {
+            new URL(schemaRegistryUrl);
+        } catch (MalformedURLException ex) {
+            throw logger.logExceptionAsWarning(new IllegalArgumentException("'endpoint' must be a valid URL.", ex));
```
srnagar, line 96:
```suggestion
            throw logger.logExceptionAsWarning(new IllegalArgumentException("'schemaRegistryUrl' must be a valid URL.", ex));
```

sdk/schemaregistry/azure-data-schemaregistry/src/main/java/com/azure/data/schemaregistry/client/CachedSchemaRegistryClientBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+     * - schemaType(), which specifies schema type, and
+     * - parseSchemaString(), which parses schemas of the specified schema type from String to Object.
+     *
+     * The parseMethod argument should be a stateless, idempotent function.
+     *
+     * @param codec Codec class implementation
+     * @return The updated {@link CachedSchemaRegistryClientBuilder} object.
+     */
+    public CachedSchemaRegistryClientBuilder addSchemaParser(Codec codec) {
+        if (CoreUtils.isNullOrEmpty(codec.schemaType())) {
```
srnagar, line 220: Check if `codec` is null before checking `codec.schemaType`.

sdk/schemaregistry/azure-data-schemaregistry/src/main/module-info.java
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+module com.azure.data.schemaregistry.serde.common {
```
srnagar, line 4: This should be named as `com.azure.data.schemaregistry`

sdk/schemaregistry/azure-data-schemaregistry/src/main/module-info.java
```diff
@@ -0,0 +1,11 @@
...
+    requires com.azure.core;
```
srnagar, line 5: This should transitively require core.

sdk/schemaregistry/azure-data-schemaregistry/swagger/swagger.yaml
srnagar: Swagger will go into azure-rest-api-specs once the swagger review board approves it. Since this swagger has not yet been reviewed, I am fine with having it in this repo until then.

@arerlend Can you checkin the json version of swagger instead of yaml

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroByteDecoder.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * @param b byte array containing encoded bytes
+     * @param object schema for Avro reader read - fetched from Azure Schema Registry
+     * @return deserialized object
+     * @throws SerializationException upon deserialization failure
+     */
+    public Object decodeBytes(byte[] b, Object object) throws SerializationException {
+        if (!(object instanceof Schema)) {
```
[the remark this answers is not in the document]
arerlend: Added Objects.requireNotNull() check

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroByteEncoder.java
arerlend: These have to be available on the byte encoder interface in order to be used by the Serializer to properly store schemas in the registry, so it needs to stay.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroAsyncDeserializer.java
arerlend: Ok.  I have removed sync and async suffixes for Serializer and Deserializer

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroDeserializerBuilder.java
arerlend: I am going update this such that the CachedSchemaRegistryClient is created in the builder and is passed into the constructor of the serializer/deserializer.

Takes advantage of all the checks in CachedSchemaRegistryClientBuilder.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/AvroByteDecoder.java
arerlend: Leaving as is for now.  I think the plan will be to allow passing in the Jackson serializer, but will default to an Apache implementation.  Dependency story is okay because Jackson has a dep on Apache anyway.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroSerializerBuilder.java
arerlend: Moved CachedSchemaRegistryClient build call to this builder so it can take advantage of the validation already in the client builder.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroDeserializerBuilder.java
arerlend: Updated both serializer and deserializer builders to have separate schemaRegistryUrl() methods.

sdk/schemaregistry/azure-data-schemaregistry-avro/src/main/java/com/azure/data/schemaregistry/avro/SchemaRegistryAvroDeserializer.java
arerlend: I just removed the extra comment line.. it's was extraneous anyway.

PR #10661, opened 2020-05-01
Update qpid-proton-j-extensions and proton-j

conniey:
* Update qpid-proton-j-extensions and proton-j
* proton-j has a module name now.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/module-info.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/main/java/module-info.java
@@ -6,7 +6,6 @@ module com.azure.messaging.eventhubs {
     requires transitive com.azure.core.amqp;
 
     requires com.microsoft.azure.qpid.protonj.extensions;
-    requires org.apache.qpid.proton.j;
 
     exports com.azure.messaging.eventhubs;
     exports com.azure.messaging.eventhubs.models;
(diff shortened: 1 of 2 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/main/java/module-info.java
```diff
@@ -6,7 +6,7 @@
     requires transitive com.azure.core.amqp;
 
     requires com.microsoft.azure.qpid.protonj.extensions;
-    requires proton.j;
+    requires org.apache.qpid.proton.j;
```
JonathanGiles, line 10: This isn't required as it will come in from the azure-core-amqp module

sdk/servicebus/azure-messaging-servicebus/src/main/java/module-info.java
```diff
@@ -3,7 +3,7 @@
 
 module com.azure.messaging.servicebus {
     requires transitive com.azure.core.amqp;
-    requires proton.j;
+    requires org.apache.qpid.proton.j;
```
JonathanGiles, line 6: Same here

sdk/eventhubs/azure-messaging-eventhubs/src/main/java/module-info.java
conniey: Ahh nice! It used to complain without it. But its probably because of the non-existent module name

Issue #10434, opened 2020-04-22
Search: ListIndexes operation should return a pageable collection


heaths:
The ListIndexes operation, because it can return up to 3000 entities for P3 plans and because indexes can be large and may grow in size, could be returned in pages in the future. To accomodate this now instead of making breaking changes later or adding "overloads" that return different objects, ListIndexes should be made pageable now.

sima-zhu:
Java is using PageIterable API for all list operations.
Quote from team channel:
"In Java, paging infrastructure in a "get only one page" fashion where there is no get next page function, adding the get next page function would be an implementation code change."

Give the nature of Java paging, we would like to keep what currently have. Please feel free to reopen if you have strong feeling about this.

Issue #10272, opened 2020-04-15
Search: Proposed model and property renames


heaths:
This is a list of proposed changes for models and properties. Some of these changes may have already been done in one or more languages, but I wanted to start an issue to track all of them and, when agreed, will replicate across the different language repos.

Model | Property | Proposal | Notes
--- | --- | --- | ---
Analyzer | | LexicalAnalyzer
AnalyzerName | | LexicalAnalyzerName
AzureActiveDirectoryApplicationCredentials | | *remove* | Just uplift ApplicationId and ApplicationSecret to SearchIndexEncryptionKey
DataContainer | | SearchIndexerDataContainer
DataSource | | SearchIndexerDataSource
DataSource | Credentials | ConnectionString (string) | See notes below for DataSourceCredentials
DataSourceCredentials | | *remove* | Just uplift ConnectionString to SearchDataSource
DataSourceType | | SearchIndexerDataSourceType
DataType | | SearchFieldDataType
EncryptionKey | | SearchResourceEncryptionKey
EncryptionKey | KeyVaultUri | VaultUri | Corresponds with Key Vault naming.
EncryptionKey | KeyVaultName | KeyName | Corresponds with Key Vault naming.
EncryptionKey | KeyVaultVersion | KeyVersion | Corresponds with Key Vault naming.
Field | | SearchField
FieldBase | | SearchFieldTemplate | Probably specific to .NET - base class for convenience classes.
Index | | SearchIndex
Indexer | | SearchIndexer
IndexerExecutionInfo | | SearchIndexerStatus | Corresponds to SearchServiceClient.GetIndexerStatus
IndexerLimits | | SearchIndexerLimits
ItemError | | SearchIndexerError
ItemWarning | | SearchIndexerWarning
LengthTokenFilter | Max | LengthTokenFilter.MaxLength
LengthTokenFilter | Min | LengthTokenFilter.MinLength
Skill | | SearchIndexerSkill
Skillset | | SearchIndexerSkillset
StandardAnalyzer | | LuceneStandardAnalyzer
StandardTokenizer | | LuceneStandardTokenizer
StandardTokenizerV2 | | LuceneStandardTokenizerV2
TokenInfo | | AnalyzedTokenInfo
Tokenizer | | LexicalTokenizer
TokenizerName | | LexicalTokenizerName

On a related note, other models should be hidden and their sole collection properties uplifted as return values. Names below reflect proposed renames above.

Method | Model | Return
--- | --- | ---
[...]

PR #10670, opened 2020-05-01
Restructured search service client to sub clients with rename models

sima-zhu:
This is a giant PR.

Some of the changes are because of populating new swagger and rename model.

It is better to focus on the top level client changes for this PR.
I uploaded source jar to api tool which is easy for reviewers to understand the major workflow we target.
https://apiview.dev/Assemblies/Review/a1fd1134b32b4439b4bd886e0e10870b

We can address top level client questions first before checking details.


sima-zhu:
Accidentally put in identity code. Will make it clean.

sdk/search/azure-search-documents/README.md
```diff
@@ -69,7 +69,7 @@ Note that you will need an admin key to index documents (query keys only work fo
 
 <!-- embedme ./src/samples/java/com/azure/search/documents/ReadmeSamples.java#L54-L58 -->
 ```Java
-SearchIndexClient searchIndexClient = new SearchIndexClientBuilder()
+SearchIndexClient searchClient = new SearchIndexClientBuilder()
```
heaths, line 72: I'm not sure I see the point of renaming the local variable if the class itself wasn't renamed to `SearchClient` (the new intended class name).

sima-zhu: The IDE takes this action when refector.
Will make the readme right.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/SearchServiceAsyncClient.java
```diff
@@ -134,11 +135,12 @@ public SearchIndexerAsyncClient getSearchIndexerAsyncClient() {
...
     /**
-     * Initializes a new {@link SearchSkillsetAsyncClient} using the same configuration as the SearchServiceAsyncClient.
+     * Initializes a new {@link SearchIndexerSkillsetAsyncClient} using the same configuration as the
+     * SearchServiceAsyncClient.
      *
-     * @return a {@link SearchSkillsetAsyncClient} created from the service client configuration.
+     * @return a {@link SearchIndexerSkillsetAsyncClient} created from the service client configuration.
      */
-    public SearchSkillsetAsyncClient getSkillsetAsyncAsyncClient() {
+    public SearchIndexerSkillsetAsyncClient getSkillsetAsyncClient() {
```
heaths, line 143: Should match `get{Model}Client`, so `getSearchIndexerSkillsetAsyncClient`.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/SearchServiceClient.java
```diff
@@ -73,11 +73,11 @@ public SearchIndexerClient getSearchIndexerClient() {
...
 
     /**
-     * Initializes a new {@link SearchSkillsetClient} using the same configuration as the SearchServiceClient.
+     * Initializes a new {@link SearchIndexerSkillsetClient} using the same configuration as the SearchServiceClient.
      *
-     * @return a {@link SearchSkillsetClient} created from the service client configuration.
+     * @return a {@link SearchIndexerSkillsetClient} created from the service client configuration.
      */
-    public SearchSkillsetClient getSkillsetClient() {
+    public SearchIndexerSkillsetClient getSkillsetClient() {
```
heaths, line 80: `get{Model}Client`

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchIndexerSkillsetClient.java
```diff
@@ -76,22 +74,19 @@ public SearchIndexerSkillset getSkillset(String skillsetName) {
...
-     * comma-separated list of JSON property names, or '*' for all properties. The default is all properties
      * @param requestOptions additional parameters for the operation. Contains the tracking ID sent with the request to
      * help with debugging
      * @param context additional context that is passed through the HTTP pipeline during the service call
      * @return the list of skillsets.
      */
-    public PagedIterable<SearchIndexerSkillset> listSkillsets(String select, RequestOptions requestOptions,
-        Context context) {
-        return new PagedIterable<>(asyncClient.listSkillsets(select, requestOptions, context));
+    public PagedIterable<SearchIndexerSkillset> listSkillsets(RequestOptions requestOptions, Context context) {
```
heaths, line 88: Only `SearchIndexClient` should return a pageable for its `list` method: #10434

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchServiceResourceClientBuilder.java
```diff
@@ -102,40 +102,40 @@ public SearchServiceResourceClientBuilder() {
...
      * {@link #serviceVersion(SearchServiceVersion)} and {@link #endpoint(String) endpoint}
-     * are used to create the {@link SearchDataSourceClient client}. All other builder settings are ignored.
+     * are used to create the {@link SearchIndexerDataSourceClient client}. All other builder settings are ignored.
      * </p>
      * @return A SearchDataSourceClient with the options set from the builder.
      * @throws NullPointerException If {@code endpoint} are {@code null}.
      */
-    public SearchDataSourceClient buildDataSourceClient() {
-        return new SearchDataSourceClient(buildDataSourceAsyncClient());
+    public SearchIndexerDataSourceClient buildDataSourceClient() {
```
heaths, line 115: I thought we agreed these wouldn't be instantiatable/buildable by themselves, but can only be obtained from `SearchServiceClient`. These are not full clients but only used for "grouping" methods. I have a meeting with @tg-msft and @KrzysztofCwalina on Monday to solidify this, but was otherwise the POR.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchSynonymMapAsyncClient.java
```diff
@@ -28,24 +29,32 @@
  * Asynchronous Client to manage and query synonym map, as well as manage other resources,
  * on a Cognitive Search service.
  */
-@ServiceClient(builder = SearchServiceResourceClientBuilder.class, isAsync = true)
-public class SearchSynonymMapAsyncClient extends SearchServiceResourceAsyncClientBase {
+public class SearchSynonymMapAsyncClient {
```
heaths, line 32: We weren't renaming this class. `SynonymMap` had no complaints about being too generic. This wasn't part of the rename issue: #10272.

Everything related to the model name (client, methods, etc.) should be fixed.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchIndexerSkillsetClient.java
sima-zhu: I think we have discussed this in chat. Java use Paging even it only returns one page only. @alzimmermsft Please correct me if I am wrong.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchServiceResourceClientBuilder.java
sima-zhu: @alzimmermsft 
I tried out to have one entry point. But Java has some limitation of initializing subclients under different packages without using public constructors/builders.

eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -362,6 +362,8 @@
   <!-- Hand-written classes should be specified here to be checked -->
   <suppress checks="."
             files=".*[/\\]search[/\\]documents[/\\]implementation[/\\]((?!(GeoPointDeserializer|Iso8601DateDeserializer|SerializationUtil)).*)"/>
+  <suppress checks="."
```
conniey, line 365: Why are we suppressing all classes in implementation?

sdk/search/azure-search-documents/README.md
conniey: You can check it to only rename code instances.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchIndexerSkillsetClient.java
alzimmermsft: @sima-zhu you are correct, the general pattern in Java is to use the paging style API for non-paged service responses since it covers dual purposes of being pages but also a list style response.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/SearchServiceAsyncClient.java
```diff
@@ -64,7 +55,8 @@
      */
     private final HttpPipeline httpPipeline;
 
-    SearchServiceAsyncClient(String endpoint, SearchServiceVersion serviceVersion, HttpPipeline httpPipeline) {
+    SearchServiceAsyncClient(String endpoint, SearchServiceVersion serviceVersion,
```
conniey, line 58: Why is it wrapped here? The max length is 120 iirc.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchServiceResourceClientBuilder.java
alzimmermsft: This is the same issue we saw with Storage where Java doesn't have a module/DLL internal scoping so we either need to make the constructor public or offer a builder that would be able to create this. In Storage we went with the latter.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/SearchServiceAsyncClient.java
```diff
@@ -115,1170 +96,70 @@ public String getEndpoint() {
...
-    }
-
-    /**
-     * Runs an indexer on-demand.
+     * Initializes a new {@link SearchAsyncClient} using the given Index name and the same configuration as the
+     * SearchServiceAsyncClient.
      *
-     * @param indexerName the name of the indexer to run
-     * @return a response signalling completion.
+     * @param indexName the name of the Index for the client
```
conniey, line 102: the casing of this is wrong. "the name of the index for the client"

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/SearchServiceAsyncClient.java
```diff
@@ -115,1170 +96,70 @@ public String getEndpoint() {
...
-    Mono<Response<Void>> deleteSynonymMapWithResponse(String synonymMapName, String etag,
-        RequestOptions requestOptions, Context context) {
-        try {
-            return restClient.synonymMaps()
-                .deleteWithRestResponseAsync(synonymMapName, etag, null, requestOptions, context)
-                .map(Function.identity());
-        } catch (RuntimeException ex) {
-            return monoError(logger, ex);
-        }
+    SearchServiceResourceClientBuilder prepareBuilder() {
```
conniey, line 158: why not "getBuilder"

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/SearchServiceAsyncClient.java
```diff
@@ -1300,11 +181,11 @@ public String getEndpoint() {
...
-        return withContext(context -> getServiceStatisticsWithResponse(requestOptions, context));
+    public Mono<Response<ServiceStatistics>> getStatisticsWithResponse(RequestOptions requestOptions) {
+        return withContext(context -> getStatisticsWithResponse(requestOptions, context));
     }
 
-    Mono<Response<ServiceStatistics>> getServiceStatisticsWithResponse(RequestOptions requestOptions, Context context) {
+    Mono<Response<ServiceStatistics>> getStatisticsWithResponse(RequestOptions requestOptions, Context context) {
         try {
             return restClient.getServiceStatisticsWithRestResponseAsync(requestOptions, context)
                 .map(Function.identity());
```
conniey, line 191: Why are we mapping function.identity? We don't need a mapping operation at all.

sdk/search/azure-search-documents/src/main/java/com/azure/search/documents/indexes/SearchIndexAsyncClient.java
```diff
@@ -0,0 +1,369 @@
...
+     * help with debugging
+     * @return a response containing the created Index.
+     */
+    public Mono<Response<SearchIndex>> createWithResponse(SearchIndex index, RequestOptions requestOptions) {
+        return withContext(context -> createWithResponse(index, requestOptions, context));
+    }
+
+    Mono<Response<SearchIndex>> createWithResponse(SearchIndex index, RequestOptions requestOptions,
+        Context context) {
+        Objects.requireNonNull(index, "'Index' cannot be null");
```
conniey, line 78: We should continue this with monoError(logger, new NullPointerException("'index' cannot be null."); This will throw rhater than return that mono error we want.

sima-zhu: Good catch. Will rework on all occurrences.

samvaity: Or just include it within the `try {} catch()` so that it is raised as a monoError

conniey: Not sure if this applies to Java, but try/catch is expensive in .NET where we can just throw the exception as-is.

PR #10674, opened 2020-05-01
Add Specialized Recording Support for avro/binary Responses

### LINKED ISSUE #10665: Recording framework to handle avro files
Network requests that return avro return Content-Type: avro/binary
The recording framework currently serializes this into a String, but the data should be stored as raw bytes.

alzimmermsft:
Fixes #10665

Adds specialized recording support for `avro/binary` responses to record them as a raw byte array.

### DIFF
--- a/sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
+++ b/sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
@@ -2,13 +2,14 @@
 // Licensed under the MIT License.
 package com.azure.core.test.http;
 
+import com.azure.core.http.ContentType;
 import com.azure.core.http.HttpClient;
 import com.azure.core.http.HttpHeaders;
 import com.azure.core.http.HttpRequest;
 import com.azure.core.http.HttpResponse;
-import com.azure.core.util.UrlBuilder;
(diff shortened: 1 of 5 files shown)


sdk/core/azure-core-test/src/main/java/com/azure/core/test/http/PlaybackClient.java
```diff
@@ -111,8 +112,13 @@ public PlaybackClient(RecordedData recordedData, Map<String, String> textReplace
...
 
-            // octet-stream's are written to disk using Arrays.toString() which creates an output such as "[12, -1]".
-            if (contentType != null && contentType.equalsIgnoreCase("application/octet-stream")) {
+            /*
+             * application/octet-stream and avro/binary are written to disk using Arrays.toString() which creates an
+             * output such as "[12, -1]".
+             */
+            if (contentType != null
+                && (contentType.equalsIgnoreCase(ContentType.APPLICATION_OCTET_STREAM)
+                    || contentType.equalsIgnoreCase("avro/binary"))) {
```
srnagar, line 121: It might be better to add "avro/binary" as a constant in [ContentType.java](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/http/ContentType.java)

PR #10689, opened 2020-05-04
[Discussion] treating empty string as signal to end paging

anuchandy:
The older version of the core [treats](https://github.com/Azure/autorest-clientruntime-for-java/blob/7cef1ea3ec21b32b31fc07827155fb5aeda9b73a/azure-client-runtime/src/main/java/com/microsoft/azure/v2/PagedList.java#L61) the empty string as a signal to end the pagination. 

```java
while (nextPageLink != null && nextPageLink != "") {
```

This is the [case](https://github.com/Azure/azure-sdk-for-net/blob/a102594ee0da79ee50815b4b9317c513a05f5153/sdk/core/Azure.Core/src/Page.cs#L22) with .NET 

```xml
Gets the continuation token used to request the next
<see cref="Page{T}"/>.  The continuation token may be null or
empty when there are no more pages.
```

In python, this is handled in generated layer, when "" string continuation token is coming over the wire, it can optionally be converted to null.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/paging/ContinuablePagedFluxCore.java
@@ -209,7 +209,16 @@ public abstract class ContinuablePagedFluxCore<C, T, P extends ContinuablePage<C
          * @param token the token
          */
         void setLastContinuationToken(C token) {
-            this.isDone = (token == null);
+            if (token == null) {
+                // null continuationToken signals end of pagination.
+                this.isDone = true;
+            } else if (token instanceof String && ((String) token).length() == 0) {
+                // For String continuationToken in addition to null
+                // empty is also used to signal end of pagination.
+                this.isDone = true;
+            } else {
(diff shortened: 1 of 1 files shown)


anuchandy:
Other approaches:

1. Adding `shouldContinue(token)` in `ContinuablePagedFluxCore` but require exposing this new public/protected method.  Usually, it's PagedFlux implementor decision to decide when to terminate, not useful for consumers to see this method. 
 
2. We could take a `Predicate<C>` as ctr params which we call into to check terminal case.
 
Taking terminal token value as ctr params is not an option since == comparison is not recommended for strings.

2 seems cleaner, will update the draft.

alzimmermsft:
> We could take a `Predicate<C>` as ctr params which we call into to check terminal case.

The default predicate will only check for continuation token `!= null`, correct? Could the `String` continuation token types include a default of `!= null && != ""`?

tooptoop4:
can this be released?

skorse:
Is there an ETA when this pr will be merged and released?

anuchandy:
The feature is shipped, closing this.

PR #10698, opened 2020-05-04
Prepare Event Hubs for May release

srnagar:
Update README, Changelog and version to prepare Event Hubs modules (Event Hubs and CheckpointStore) for May release.

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/pom.xml
+++ b/sdk/eventhubs/azure-messaging-eventhubs/pom.xml
@@ -42,7 +42,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-amqp</artifactId>
-      <version>1.1.0</version> <!-- {x-version-update;unreleased_com.azure:azure-core-amqp;dependency} -->
+      <version>1.1.0</version> <!-- {x-version-update;com.azure:azure-core-amqp;dependency} -->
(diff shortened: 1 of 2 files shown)


sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Release History
 
-## 1.1.0-beta.1 (Unreleased)
+## 1.0.4 (2020-05-05)
+- Updated version of `azure-messaging-eventhubs` to `5.1.0`.
```
alzimmermsft, line 4: Should the version of Storage be updated as well?

JimSuplizio:
@srnagar is this PR jumping the gun a little? There are other dependencies, like identity and azure-storage-blobs, that should be being released first and this should be releasing with the updated dependencies.

sdk/eventhubs/azure-messaging-eventhubs-checkpointstore-blob/CHANGELOG.md
srnagar: Discussed offline with @alzimmermsft and @JimSuplizio. This change needs to wait until Identity and Storage modules are released.

PR #10726, opened 2020-05-05
Feature/storage/stg73 regenerate swagger for blobs

kasobol-msft:
- regenerate swagger from latest spec
- fix blob swagger Readme.md

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
@@ -28,6 +28,7 @@ import com.azure.storage.blob.models.BlobServiceProperties;
 import com.azure.storage.blob.models.BlobServiceStatistics;
 import com.azure.storage.blob.models.CpkInfo;
 import com.azure.storage.blob.models.KeyInfo;
+import com.azure.storage.blob.models.ListBlobContainersIncludeType;
 import com.azure.storage.blob.models.ListBlobContainersOptions;
 import com.azure.storage.blob.models.PublicAccessType;
 import com.azure.storage.blob.models.StorageAccountInfo;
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/ListContainersIncludeType.java
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+
+package com.azure.storage.blob.implementation.models;
```
alzimmermsft, line 5: Looks like this is double generated? Shouldn't there only be the `ListBlobContainersIncludeType` in the public API and this isn't needed?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/ListContainersIncludeTypeWrapper.java
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+
+package com.azure.storage.blob.implementation;
```
alzimmermsft, line 5: Should this and `ListBlobContainersIncludeTypeWrapper` both exist? Should this be removed and the XML node names be fixed in the `BlobContainers` variant?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/ListContainersIncludeType.java
kasobol-msft: Removed.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/ListContainersIncludeTypeWrapper.java
kasobol-msft: Good point. Ideally both are not needed.
However, I didn't find a way to force generator to skip `ListBlobContainersIncludeTypeWrapper`, so I end result has just one wrapper.

I can see that similar issue exist for existing `ListBlobsIncludeItemWrapper` and `ListBlobsIncludeItem` pair - the wrapper is not used but exist.

PR #10728, opened 2020-05-05
cosmosdb analyticalStorage -> analyticalStore

moderakh:
analyticalStorage -> analyticalStore


moderakh:
`Should we also change the documentation here ?
`
Good catch @kushagraThapar yes we should. will update doc.

moderakh:
@kushagraThapar done

PR #10734, opened 2020-05-05
SyncClient: Bug fix - Multiple Receive on same client

hemanttanwar:
Sync Client should be able to call receive() multiple times.

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
@@ -620,7 +620,8 @@ public final class ServiceBusReceiverClient implements AutoCloseable {
     }
 
     /**
-     * Given an {@code emitter}, creates a {@link SynchronousMessageSubscriber} to receive messages from Service Bus.
+     * Given an {@code emitter}, creates a {@link SynchronousMessageSubscriber} to receive messages from Service Bus
+     * entity.
      */
     private void queueWork(int maximumMessageCount, Duration maxWaitTime,
         FluxSink<ServiceBusReceivedMessageContext> emitter) {


sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -622,20 +628,56 @@ public void setSessionState(String sessionId, byte[] sessionState) {
...
     private void queueWork(int maximumMessageCount, Duration maxWaitTime,
         FluxSink<ServiceBusReceivedMessageContext> emitter) {
-        final long id = idGenerator.getAndIncrement();
-        final SynchronousReceiveWork work = new SynchronousReceiveWork(id, maximumMessageCount, maxWaitTime,
-            emitter);
-        final SynchronousMessageSubscriber syncSubscriber = new SynchronousMessageSubscriber(work);
 
-        logger.info("[{}]: Started synchronous message subscriber.", id);
-        asyncClient.receive(DEFAULT_RECEIVE_OPTIONS).subscribeWith(syncSubscriber);
+        if (messageProcessor.get() != null && messageProcessor.get().isDisposed()) {
```
conniey, line 649: This isn't thread safe, in the time that you check for null, someone else could have set it. Or if it is null, and then you create it, it could possibly overwrite an existing EmitterProcessor that was set in this time.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -622,20 +628,56 @@ public void setSessionState(String sessionId, byte[] sessionState) {
     @Override
     public void close() {
         asyncClient.close();
+
+        if (messageProcessor.get() != null) {
```
conniey, line 632: Will it throw if you call dispose twice?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -622,20 +628,56 @@ public void setSessionState(String sessionId, byte[] sessionState) {
...
+            }
+
+            logger.info("[{}]: Started ContinuesMessageSubscriber message subscriber for entity.",
+                asyncClient.getEntityPath());
+        }
+
+        Disposable newSubscription = messageProcessor.get()
+            .take(maximumMessageCount)
+            .timeout(maxWaitTime)
+            .doOnNext(messageContext -> {
```
conniey, line 669: These are for side effects.

conniey: doOnNext, doOnError, doOnComplete are for side effects like logging or monitoring. It shouldn't be used as part of the main chain of operations.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
hemanttanwar: This should insure against it `messageProcessor.compareAndSet(null, newProcessor)`

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -622,20 +629,63 @@ public void setSessionState(String sessionId, byte[] sessionState) {
...
+            logger.info("[{}]: Started EmitterProcessor message processor for entity.",
+                asyncClient.getEntityPath());
+        }
+
+            logger.info("[{}]: Subscribing EmitterProcessor message processor for entity.",
+                asyncClient.getEntityPath());
+
+        Disposable newSubscription = messageProcessor.get()
+            .take(maximumMessageCount)
+            .timeout(maxWaitTime)
```
srnagar, line 674: `timeout()` is not the right operator here. The `maxWaitTime` param specifies the maximum time to wait before receiving all messages upto `maximumMessageCount`. `timeout()` operator will reset the timer after each message is received. Also, if there is a timeout, it will throw an exception which is not what should happen. Instead, it should return all messages until the timeout occurred.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SyncMessageSubscriber.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    protected void hookOnError(Throwable throwable) {
+        throwable.printStackTrace();
+    }
+
+    @Override
+    protected void hookOnCancel() {
+        System.out.println("Cancelled");
+    }
```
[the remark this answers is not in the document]
hemanttanwar: I should not have checked in, I was trying something else. So  removing it.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -622,20 +627,61 @@ public void setSessionState(String sessionId, byte[] sessionState) {
...
+            })
+            .subscribe(message -> {},
+                error -> {
+                    logger.error("Error occurred while receiving messages.", error);
+                    emitter.error(error);
+                },
+                () -> emitter.complete());
+
+        Disposable oldSubscription = messageProcessorSubscription.getAndSet(newSubscription);
+        if (oldSubscription != null && !oldSubscription.isDisposed()) {
```
conniey, line 683: What if the existing subscription is still waiting for work? You're going to cancel it?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -606,20 +609,28 @@ public void setSessionState(String sessionId, byte[] sessionState) {
...
     }
 
     /**
-     * Given an {@code emitter}, queues that work in {@link SynchronousMessageSubscriber}. If the synchronous job has
-     * not been created, will initialise it.
+     * Given an {@code emitter}, creates a {@link SynchronousMessageSubscriber} to receive messages from Service Bus.
      */
-    private void queueWork(int maximumMessageCount, Duration maxWaitTime,
+    private synchronized void queueWork(int maximumMessageCount, Duration maxWaitTime,
         FluxSink<ServiceBusReceivedMessageContext> emitter) {
```
hemanttanwar, line 622: This is because if multiple client are calling receive, we need to ensure that short lived sync subscriber is thread safe.

conniey: You don't need an atomic variable then.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -606,20 +611,28 @@ public void setSessionState(String sessionId, byte[] sessionState) {
     @Override
     public void close() {
         asyncClient.close();
+
+        if (messageProcessor.get() != null) {
```
conniey, line 615: The operation should be `cancel()`, not `onComplete()`. onComplete and methods like that should be done through a sink.

```java
var processor = messageProcessor.getAndSet(null);
if (processor != null) {
   processor.cancel()
}
```

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
conniey: You can also synchronize on a `lock` object rather than on an entire class to scope down how large this critical path is.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -606,20 +611,28 @@ public void setSessionState(String sessionId, byte[] sessionState) {
...
         final SynchronousReceiveWork work = new SynchronousReceiveWork(id, maximumMessageCount, maxWaitTime,
             emitter);
         final SynchronousMessageSubscriber syncSubscriber = new SynchronousMessageSubscriber(work);
-
         logger.info("[{}]: Started synchronous message subscriber.", id);
-        asyncClient.receive(DEFAULT_RECEIVE_OPTIONS).subscribeWith(syncSubscriber);
+
+        if (messageProcessor.get() ==  null) {
+            messageProcessor.set(this.asyncClient.receive(DEFAULT_RECEIVE_OPTIONS)
+                .subscribeWith(EmitterProcessor.create(false)));
```
conniey, line 633: This overload fetches 256. I was hoping you'd use the overload that specifies the buffer size, so you don't accidentally consume more messages than the user wants to process. EmitterProcessor.create(prefetch, false);

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -606,20 +611,28 @@ public void setSessionState(String sessionId, byte[] sessionState) {
...
+        }
+
+        messageProcessor.get().subscribe(syncSubscriber);
```
conniey, line 636: The general trick is to store the value from your AtomicReference operation in a local variable so you don't keep accessing it. You had already fetched its value in line 631.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
hemanttanwar: yes: atomic is not needed  but our Checks during build time complains about it or I had to supress the warning .

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
hemanttanwar: discussed on teams and decided to not change it.

conniey:
Before this is merged, it should take into consideration that multiple receive operations can be called from different threads, and we don't want this to publish the same message to multiple downstream. It should "drain" pending receive work.

PR #10735, opened 2020-05-05
Add JUnit Based Test Parallelization

alzimmermsft:
This PR updates the version of JUnit 5 being used and adds a profile which runs JUnit 5 tests in parallel. The parallelization configuration in the client parent POM is full parallelization, in some places where this isn't safe an override configuration has been added to either make it class level parallelization or no parallelization.

The parallelization profile is only being triggered during playback test runs as they have a higher degree of concurrency safety as all resources are localized to a single test. A new flag `parallel-test-playback` has been added which will trigger tests to run concurrently, when running from source this could be activated when running against the live service. Long term the goal would be having all testing running concurrently to better leverage resources but for now it will be limited to playback only.

Notable improvements due to this change:

- KeyVault runs reduced by ~40-50%, 12-16 minutes to 6-9 minutes.
- Form Recognizer runs reduced by ~30-40%, 8-11 minutes to 5-8 minutes.

NOTE: Tests using JUnit 4, Spock, or TestNG won't be affected by this change as this uses JUnit 5 configuration variables.


sdk/search/azure-search-documents/pom.xml
```diff
@@ -149,5 +149,28 @@
...
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-surefire-plugin;external_dependency} -->
+            <configuration>
+              <systemPropertyVariables>
```
sima-zhu, line 167: Does it affect live tests? Why search is disabled?

sdk/eventhubs/azure-messaging-eventhubs/pom.xml
```diff
@@ -89,4 +89,29 @@
...
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-surefire-plugin;external_dependency} -->
+            <configuration>
+              <systemPropertyVariables>
+                <junit.jupiter.execution.parallel.enabled>false</junit.jupiter.execution.parallel.enabled>
```
JonathanGiles, line 109: If you are going to disable parallel execution, please include a comment about why this was deemed necessary.

sdk/search/azure-search-documents/pom.xml
```diff
@@ -149,5 +149,28 @@
...
+                <junit.jupiter.execution.parallel.enabled>false</junit.jupiter.execution.parallel.enabled>
```
JonathanGiles, line 168: Same here

sdk/servicebus/azure-messaging-servicebus/pom.xml
```diff
@@ -88,4 +88,28 @@
...
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-surefire-plugin;external_dependency} -->
+            <configuration>
+              <systemPropertyVariables>
+                <junit.jupiter.execution.parallel.enabled>false</junit.jupiter.execution.parallel.enabled>
```
JonathanGiles, line 107: And here...

alzimmermsft: Comment added

sdk/search/azure-search-documents/pom.xml
alzimmermsft: Comment added

sdk/eventhubs/azure-messaging-eventhubs/pom.xml
alzimmermsft: Comment added

sdk/search/azure-search-documents/pom.xml
alzimmermsft: When using the host agent parallel runs will only happen in PR validation, so this won't affect live tests. I updated the PR description to explain that this was done since playback is much safer to parallelize without an in-depth investigation due to the locality of testing resources, if you want to you could run tests in parallel locally against the service.

Parallel tests for Search has been disabled as there is some testing global state that doesn't play well with parallelization and causes intermittent test failures, I added a comment into the POM stating this.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/TestSuiteBase.java
```diff
@@ -3,36 +3,43 @@
...
-import com.azure.cosmos.models.CosmosAsyncDatabaseResponse;
-import com.azure.cosmos.models.CosmosAsyncItemResponse;
 import com.azure.cosmos.CosmosAsyncUser;
 import com.azure.cosmos.CosmosBridgeInternal;
 import com.azure.cosmos.CosmosClient;
 import com.azure.cosmos.CosmosClientBuilder;
 import com.azure.cosmos.CosmosClientException;
-import com.azure.cosmos.models.CosmosContainerProperties;
-import com.azure.cosmos.models.CosmosContainerRequestOptions;
-import com.azure.cosmos.models.JsonSerializable;
```
alzimmermsft, line 25: This was the invalid import, re-order on save happened.

mssfang:
@alzimmermsft  live tests will fail for sure. The reason is current master code should use a new endpoint that has the new feature. The fixing PR is https://github.com/Azure/azure-sdk-for-java/pull/11051#issuecomment-627045571

alzimmermsft:
> 
> 
> @alzimmermsft live tests will fail for sure. The reason is current master code should use a new endpoint that has the new feature. The fixing PR is [#11051 (comment)](https://github.com/Azure/azure-sdk-for-java/pull/11051#issuecomment-627045571)

Didn't realize that the live tests were in an unstable state, was attempting to run validation that this change didn't adversely affect Text Analytics.

Issue #7349, opened 2020-01-10
SSLEngine closed already in reactor-netty http client


kushagraThapar:
**Describe the bug**
SSLEngine closed already gets thrown by reactor-netty http client when running ReadMyWrites in Strong Https mode

***Exception or Stack Trace***
2020-01-10 20:16:08,322       [reactor-http-epoll-4] WARN  reactor.netty.http.client.HttpClientConnect - [id: 0x393d7ffd, L:/10.2.0.4:46496 - R:cdb-ms-prod-westus2-fd25.documents.azure.com/13.66.138.15:14939] The connection observed an error
javax.net.ssl.SSLException: SSLEngine closed already
        at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:834)
        at io.netty.handler.ssl.SslHandler.wrapAndFlush(SslHandler.java:797)
        at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:778)
        at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:1883)
        at io.netty.handler.ssl.SslHandler.closeOutboundAndChannel(SslHandler.java:1851)
        at io.netty.handler.ssl.SslHandler.close(SslHandler.java:729)
        at io.netty.channel.AbstractChannelHandlerContext.invokeClose(AbstractChannelHandlerContext.java:621)
        at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:605)
        at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.close(CombinedChannelDuplexHandler.java:507)
        at io.netty.channel.ChannelOutboundHandlerAdapter.close(ChannelOutboundHandlerAdapter.java:77)
        at io.netty.channel.CombinedChannelDuplexHandler.close(CombinedChannelDuplexHandler.java:318)
        at io.netty.channel.AbstractChannelHandlerContext.invokeClose(AbstractChannelHandlerContext.java:621)
        at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:605)
        at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:467)
        at io.netty.channel.DefaultChannelPipeline.close(DefaultChannelPipeline.java:969)
        at io.netty.channel.AbstractChannel.close(AbstractChannel.java:231)
[...]

PR #10741, opened 2020-05-06
Added content response on write enabled feature to Gateway and Direct TCP

kushagraThapar:
This PR: 
* Adds `contentResponseOnWriteEnabled` argument to CosmosClientBuilder and all internal client builders.  
* By default the value for the argument is set to false -> Item write operations will not return content response. 
* This feature only works in Gateway and Direct TCP mode. (Direct HTTP is implemented on the backend, but needs to be released)
* Exposes new APIs - `getETag()` on `CosmosItemResponse` and `CosmosAsyncItemResponse`
* Added configuration options for Write Throughput and Write Latency Benchmarks. 
* Updated client builders in TestSuiteBase to support `contentResponseOnWriteEnabled ` testing. 
* Updated and closed some open github issues that have already been fixed.

### DIFF
--- a/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/CollectionCRUDAsyncAPITest.java
+++ b/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/CollectionCRUDAsyncAPITest.java
@@ -81,7 +81,8 @@ public class CollectionCRUDAsyncAPITest extends DocumentClientTest {
             .withServiceEndpoint(TestConfigurations.HOST)
             .withMasterKeyOrResourceToken(TestConfigurations.MASTER_KEY)
             .withConnectionPolicy(connectionPolicy)
-            .withConsistencyLevel(ConsistencyLevel.SESSION);
+            .withConsistencyLevel(ConsistencyLevel.SESSION)
+            .withContentResponseOnWriteEnabled(true);
 
         this.client = this.clientBuilder().build();
 
--- a/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/ConflictAPITest.java
+++ b/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/ConflictAPITest.java
@@ -59,7 +59,8 @@ public class ConflictAPITest extends DocumentClientTest {
             .withServiceEndpoint(TestConfigurations.HOST)
             .withMasterKeyOrResourceToken(TestConfigurations.MASTER_KEY)
             .withConnectionPolicy(connectionPolicy)
-            .withConsistencyLevel(ConsistencyLevel.SESSION);
+            .withConsistencyLevel(ConsistencyLevel.SESSION)
+            .withContentResponseOnWriteEnabled(true);
 
         this.client = this.clientBuilder().build();
 
--- a/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/CosmosAuthorizationTokenResolverTest.java
+++ b/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/CosmosAuthorizationTokenResolverTest.java
@@ -69,7 +69,8 @@ public class CosmosAuthorizationTokenResolverTest extends DocumentClientTest {
             .withServiceEndpoint(TestConfigurations.HOST)
             .withMasterKeyOrResourceToken(TestConfigurations.MASTER_KEY)
             .withConnectionPolicy(connectionPolicy)
-            .withConsistencyLevel(ConsistencyLevel.SESSION);
+            .withConsistencyLevel(ConsistencyLevel.SESSION)
+            .withContentResponseOnWriteEnabled(true);
 
         this.client = this.clientBuilder().build();
 
--- a/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/DatabaseCRUDAsyncAPITest.java
+++ b/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/DatabaseCRUDAsyncAPITest.java
@@ -65,7 +65,8 @@ public class DatabaseCRUDAsyncAPITest extends DocumentClientTest {
             .withServiceEndpoint(TestConfigurations.HOST)
             .withMasterKeyOrResourceToken(TestConfigurations.MASTER_KEY)
             .withConnectionPolicy(connectionPolicy)
-            .withConsistencyLevel(ConsistencyLevel.SESSION);
+            .withConsistencyLevel(ConsistencyLevel.SESSION)
+            .withContentResponseOnWriteEnabled(true);
 
         this.client = this.clientBuilder().build();
     }
--- a/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/DocumentCRUDAsyncAPITest.java
+++ b/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/DocumentCRUDAsyncAPITest.java
@@ -89,7 +89,8 @@ public class DocumentCRUDAsyncAPITest extends DocumentClientTest {
             .withServiceEndpoint(TestConfigurations.HOST)
             .withMasterKeyOrResourceToken(TestConfigurations.MASTER_KEY)
             .withConnectionPolicy(connectionPolicy)
-            .withConsistencyLevel(ConsistencyLevel.SESSION);
+            .withConsistencyLevel(ConsistencyLevel.SESSION)
+            .withContentResponseOnWriteEnabled(true);
 
         this.client = this.clientBuilder().build();
 
--- a/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/DocumentQueryAsyncAPITest.java
+++ b/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/DocumentQueryAsyncAPITest.java
@@ -85,7 +85,8 @@ public class DocumentQueryAsyncAPITest extends DocumentClientTest {
             .withServiceEndpoint(TestConfigurations.HOST)
             .withMasterKeyOrResourceToken(TestConfigurations.MASTER_KEY)
             .withConnectionPolicy(connectionPolicy)
-            .withConsistencyLevel(ConsistencyLevel.SESSION);
+            .withConsistencyLevel(ConsistencyLevel.SESSION)
+            .withContentResponseOnWriteEnabled(true);
 
         this.client = this.clientBuilder().build();
 
--- a/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/InMemoryGroupbyTest.java
+++ b/sdk/cosmos/azure-cosmos-examples/src/test/java/com/azure/cosmos/rx/examples/InMemoryGroupbyTest.java
@@ -46,7 +46,8 @@ public class InMemoryGroupbyTest extends DocumentClientTest {
             .withServiceEndpoint(TestConfigurations.HOST)
             .withMasterKeyOrResourceToken(TestConfigurations.MASTER_KEY)
             .withConnectionPolicy(connectionPolicy)
-            .withConsistencyLevel(ConsistencyLevel.SESSION);
+            .withConsistencyLevel(ConsistencyLevel.SESSION)
+            .withContentResponseOnWriteEnabled(true);
 
         this.client = this.clientBuilder().build();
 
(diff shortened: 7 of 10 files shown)


sdk/cosmos/azure-cosmos-examples/src/main/java/com/azure/cosmos/examples/ChangeFeed/SampleChangeFeedProcessor.java
```diff
@@ -121,6 +121,7 @@ public static CosmosAsyncClient getCosmosClient() {
                 .key(SampleConfigurations.MASTER_KEY)
                 .connectionPolicy(ConnectionPolicy.getDefaultPolicy())
                 .consistencyLevel(ConsistencyLevel.EVENTUAL)
+                .returnMinimalResponse(false)
```
kirankumarkolli, line 124: ChangeFeedProcessor is a read-only, how minimal response impacts it?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
```diff
@@ -163,6 +166,20 @@ CosmosKeyCredential cosmosKeyCredential() {
...
+     * Gets the boolean which indicates whether to only return the headers and status code in Cosmos DB response
+     * in case of Create, Update and Delete operations on CosmosItem.
+     *
+     * If true, this removes the resource from response. It reduces networking
+     * and CPU load by not sending the resource back over the network and serializing it
+     * on the client.
+     *
+     * @return a boolean indicating whether resource will be included in the response or not
+     */
+    boolean isReturnMinimalResponse() {
```
kirankumarkolli, line 179: .NET is using name 'NoResponseContent'.

/cc: @j82w

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentServiceRequest.java
```diff
@@ -26,8 +26,6 @@
  * This is core Transport/Connection agnostic request to the Azure Cosmos DB database service.
  */
 public class RxDocumentServiceRequest implements Cloneable {
-    private static final char PREFER_HEADER_SEPERATOR = ';';
-    private static final String PREFER_HEADER_VALUE_FORMAT = "%s=%s";
```
kirankumarkolli, line 30: What are scenarios for it?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/directconnectivity/HttpTransportClient.java
```diff
@@ -84,7 +84,7 @@ public HttpTransportClient(Configs configs, Duration requestTimeout, UserAgentCo
 
         // Set requested API version header for version enforcement.
         this.defaultHeaders.put(HttpConstants.HttpHeaders.VERSION, HttpConstants.Versions.CURRENT_VERSION);
-        this.defaultHeaders.put(HttpConstants.HttpHeaders.CACHE_CONTROL, HttpConstants.HeaderValues.NoCache);
+        this.defaultHeaders.put(HttpConstants.HttpHeaders.CACHE_CONTROL, HttpConstants.HeaderValues.NO_CACHE);
```
moderakh, line 87: good catch.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
kirankumarkolli: NoContentResponseOnWrite was the current full name .NET using.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
```diff
@@ -915,6 +928,27 @@ private static void validateResource(Resource resource) {
         return headers;
     }
 
+    static Map<String, String> addPreferHeader(String preferHeaderName, String preferHeaderValue, Map<String, String> headers) {
+        if (headers == null) {
+            headers = new HashMap<>();
+        }
+        return addPreferHeader(preferHeaderName + "=" + preferHeaderValue, headers);
```
moderakh, line 935: given that currently the only preferred header is minimal response, I wonder if we can avoid string concatination here:
instead of
```java
preferHeaderName + "=" + preferHeaderValue
```
we just pass `minimanRespose=true`

sdk/cosmos/azure-cosmos-examples/src/main/java/com/azure/cosmos/examples/ChangeFeed/SampleChangeFeedProcessor.java
kushagraThapar: This is SampleChangeFeedProcessor example, which creates some documents before listening to them, so used it. This is not in the source code.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
kushagraThapar: Will go with `noContentResponseOnWrite` as the new name. Thanks.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentServiceRequest.java
kushagraThapar: Unknown scenarios, but this already exists in the SDK, and there is a unit test as well. Didn't want to break that. 
https://github.com/Azure/azure-sdk-for-java/pull/10741/files#diff-7011a46947ace6e46648f8c4027e602eL414
So to keep the functionality intact, I am re-using the logic we already have.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/ReturnMinimalResponseTest.java
```diff
@@ -0,0 +1,266 @@
...
+        safeDeleteSyncDatabase(createdDatabase);
+        for (String dbId : databases) {
+            safeDeleteSyncDatabase(client.getDatabase(dbId));
+        }
+        safeCloseSyncClient(client);
+    }
+
+    @Test(groups = {"simple"}, timeOut = TIMEOUT)
+    public void createDatabase_withMinimalResponseConfig() throws CosmosClientException {
+        CosmosDatabaseProperties databaseDefinition = new CosmosDatabaseProperties(CosmosDatabaseForTest.generateId());
```
moderakh, line 69: we usually target emulator for database and container CRUD test, because otherwise we may run out of master resource budget in prod.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentServiceRequest.java
moderakh: I am thinking if this is not exposed through public api, we should drop the generic support and only have support for minimalResponseContent scenario

kushagraThapar: I like that idea @moderakh . I will get rid of it.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/RxDocumentClientImpl.java
kushagraThapar: This doesn't get called. The other overload gets called :) 
But since the only preferred header is minimalResponse, I will remove this overload. 
Also, just to clarify, it is passed as: `prefer:return=minimal` :)

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/ReturnMinimalResponseTest.java
kushagraThapar: Yes, makes sense, will do that.

kirankumarkolli:
Would be nice to behcnmark with this change in.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -291,6 +292,39 @@ public CosmosClientBuilder keyCredential(CosmosKeyCredential cosmosKeyCredential
...
+    boolean isNoContentResponseOnWrite() {
+        return noContentResponseOnWrite;
+    }
+
+    /**
+     * Sets the boolean to only return the headers and status code in Cosmos DB response
+     * in case of Create, Update and Delete operations on CosmosItem.
+     *
+     * If set to true, this removes the resource from response. It reduces networking
+     * and CPU load by not sending the resource back over the network and serializing it on the client.
```
FabianMeiswinkel, line 316: NIT: I think pretty much all customers identifying this setting will also ask themselves whether this flag will have impact on RU charges. Would be good to document that it doesn't (at least my understanding) here - or at least in the FAQ for the feature.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -291,6 +292,39 @@ public CosmosClientBuilder keyCredential(CosmosKeyCredential cosmosKeyCredential
...
+     *
+     * If true, this removes the resource from response. It reduces networking
+     * and CPU load by not sending the resource back over the network and serializing it
+     * on the client.
+     *
+     * By-default, this is true.
+     *
+     * @return a boolean indicating whether resource will be included in the response or not
+     */
+    boolean isNoContentResponseOnWrite() {
```
moderakh, line 307: For other similar configs we are using the `Enabled` suffix

`isSessionCapturingOverrideEnabled`
`isConnectionReuseAcrossClientsEnabled`

should we change this to `isNoContentResponseOnWriteEnabled` to be consistent with the other APIs?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -291,6 +292,39 @@ public CosmosClientBuilder keyCredential(CosmosKeyCredential cosmosKeyCredential
...
+     *
+     * If set to true, this removes the resource from response. It reduces networking
+     * and CPU load by not sending the resource back over the network and serializing it on the client.
+     *
+     * By-default, this is true.
+     *
+     * @param noContentResponseOnWrite a boolean indicating whether resource will be included in the response or not
+     * @return current cosmosClientBuilder
+     */
+    public CosmosClientBuilder noContentResponseOnWrite(boolean noContentResponseOnWrite) {
```
moderakh, line 323: similarly
For other similar configs we are using the Enabled suffix

```java
SessionCapturingOverrideEnabled
ConnectionReuseAcrossClientsEnabled
```
should we change this to `NoContentResponseOnWriteEnabled` to be consistent with the other APIs?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncClient.java
moderakh: for similar configs we have "Enabled" as the prefix. To be consistent please add "Enabled" at the end. (`noContentResponseOnWriteEnabled`)

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
kushagraThapar: I like that idea, will document that "this doesn't impact RU usage for read or write operations"

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
kushagraThapar: Yes, makes sense to be consistent across these APIs, good catch @moderakh 
I will update this public surface API to reflect that.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
kushagraThapar: This will differ from what we have on .Net side. 
@kirankumarkolli  - what do you think ?

kushagraThapar:
Single region session long CI failed after running all the tests with spot bugs jar downloads. 

```
2020-05-07T15:44:56.2558748Z [ERROR] Failed to execute goal com.github.spotbugs:spotbugs-maven-plugin:3.1.12.2:spotbugs (spotbugs) on project azure-cosmos: Execution spotbugs of goal com.github.spotbugs:spotbugs-maven-plugin:3.1.12.2:spotbugs failed: Plugin com.github.spotbugs:spotbugs-maven-plugin:3.1.12.2 or one of its dependencies could not be resolved: The following artifacts could not be resolved: org.apache.bcel:bcel:jar:6.3.1, net.jcip:jcip-annotations:jar:1.0, org.dom4j:dom4j:jar:2.1.1, jaxen:jaxen:jar:1.1.6: Could not transfer artifact org.apache.bcel:bcel:jar:6.3.1 from/to ossrh (https://oss.sonatype.org/content/repositories/snapshots/): Transfer failed for https://oss.sonatype.org/content/repositories/snapshots/org/apache/bcel/bcel/6.3.1/bcel-6.3.1.jar: Connection reset -> [Help 1]
```

kushagraThapar:
Single Region Strong Https e2e failed because of lots of SSLEngine closed already reactor-netty issue. 
I think it might have caused the VM to crash. I have added more information on this issue: 
https://github.com/Azure/azure-sdk-for-java/issues/7349

PR #10838, opened 2020-05-06
Fix ci.data.yml's exclude

JimSuplizio:
ci.data.yml wasn't correctly excluding sdk/servicebus/azure-messaging-servicebus which was causing it to get kicked off when changes were made to the client library


JimSuplizio:
The linux issues with the client library are being looked at as is the issue with the analyze on the data. My change just prevents the data pipeline from running for client changes.

PR #9789, opened 2020-04-02
WIP Basic changefeed implementation 

gapra-msft:
Test situation still being worked out

This PR covers - basic CF from start to end, user settable start and end time, basic paging functionality.

PR #10839, opened 2020-05-06
Implemented Changefeed for Blobs

gapra-msft:
Original PR : https://github.com/Azure/azure-sdk-for-java/pull/9789

Easiest order to review in AvroReader, LazyDownloader, Chunk, Shard, Segment, Changefeed, ChangefeedPagedFlux, then everything else.


sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFluxTest.groovy
```diff
@@ -0,0 +1,244 @@
...
+
+import reactor.core.publisher.Flux
+import reactor.test.StepVerifier
+
+import java.time.OffsetDateTime
+
+import static org.mockito.ArgumentMatchers.any
+import static org.mockito.Mockito.*
+
+class BlobChangefeedPagedFluxTest extends HelperSpec {
```
kasobol-msft, line 15: What is `HelperSpec` ?

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/HelperSpec.groovy
```diff
@@ -0,0 +1,68 @@
...
+import spock.lang.Specification
+
+import java.nio.ByteBuffer
+import java.nio.channels.AsynchronousFileChannel
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.nio.file.StandardOpenOption
+import java.time.OffsetDateTime
+
+class HelperSpec extends Specification {
```
kasobol-msft, line 18: Should this be `MockedChangeFeedSpec` or something more concrete? 

https://softwareengineering.stackexchange.com/questions/247267/what-is-a-helper-is-it-a-design-pattern-is-it-an-algorithm

sdk/storage/azure-storage-blob-changefeed/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
kasobol-msft, line 1: what does it do?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ChunkedDownloadUtils.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * RESERVED FOR INTERNAL USE.
+ */
+public class ChunkedDownloadUtils {
+
+    /*
+    Download the first chunk. Construct a Mono which will emit the total count for calculating the number of chunks,
+    access conditions containing the etag to lock on, and the response from downloading the first chunk.
+     */
+    public static Mono<Tuple3<Long, BlobRequestConditions, BlobDownloadAsyncResponse>> getSetupMono(BlobRange range,
```
kasobol-msft, line 28: Should this be called `downloadFirstChunk`?

sdk/storage/azure-storage-blob-changefeed/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
gapra-msft: I looked up how to make Mockito work with Groovy and this is what came of it.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Implementation of {@link ContinuablePagedFlux} for Changefeed where the continuation token type is {@link String},
+ * the element type is {@link BlobChangefeedEvent}, and the page type is {@link BlobChangefeedPagedResponse}.
+ */
+public final class BlobChangefeedPagedFlux extends ContinuablePagedFlux<String, BlobChangefeedEvent,
+    BlobChangefeedPagedResponse> {
+
+    private final Changefeed changefeed;
+    private final Integer defaultPageSize = 5000;
```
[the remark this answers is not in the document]
gapra-msft: fixed

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
```diff
@@ -0,0 +1,97 @@
...
+        return byPage(null, preferredPageSize);
+    }
+
+    @Override
+    public Flux<BlobChangefeedPagedResponse> byPage(String continuationToken, int preferredPageSize) {
+
+        if (continuationToken != null) {
+            return Flux.error(new UnsupportedOperationException("continuationToken not supported. Use "
+                + "client.getEvents(String) to pass in a cursor."));
+        }
```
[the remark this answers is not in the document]
gapra-msft: I think we need to have a broader discussion about this but the reason I don't want to make that an option is because I don't want users to think they have to provide a continuationToken to get the next page in changefeed (the implementation by default has the ability to get every page lazily). In changefeed the cursor is just used to initialize the changefeed object correctly as opposed to continually be fed to get the next page (this is what users are most used to doing). Does that make sense?_

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
```diff
@@ -0,0 +1,97 @@
...
+        if (preferredPageSize <= 0) {
+            return Flux.error(new IllegalArgumentException("preferredPageSize > 0 required but provided: "
+                + preferredPageSize));
+        }
+        preferredPageSize = Integer.min(preferredPageSize, defaultPageSize);
```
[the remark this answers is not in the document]
gapra-msft: This preferred page size is kinda made up. We can have a discussion about this as well. I wanted the experience to be consistent with other service list calls where it uses the logic I used.

 If the request does not specify maxResultsPerPage or specifies a value greater than 5,000, the server will return up to 5,000 items.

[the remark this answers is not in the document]
gapra-msft: We can also add this to documentation if we want to go down this route - or we can let a user specify anything

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedIterable.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.azure.core.util.paging.ContinuablePagedIterable;
+import com.azure.storage.blob.changefeed.models.BlobChangefeedEvent;
+
+/**
+ * Implementation of {@link ContinuablePagedIterable} for Changefeed where the continuation token type is
+ * {@link String}, the element type is {@link BlobChangefeedEvent}, and the page type is
+ * {@link BlobChangefeedPagedResponse}.
+ */
+public final class BlobChangefeedPagedIterable extends ContinuablePagedIterable<String,
```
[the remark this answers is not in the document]
gapra-msft: This is one of those things where we're trying to be consistent. This seems to be the way even other services do this

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedResponse.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Gets a reference to the next page, should you want to re-initialize the BlobChangefeed.
+     *
+     * @return The {@link String cursor} that references the next page.
+     */
+    public String getContinuationToken() {
+        /* Serialize the cursor and return it to the user as a String. */
+        return cursor.serialize();
+    }
```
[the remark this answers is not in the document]
gapra-msft: See my comment above. Maybe this method (inherited from pagedresponse) should throw instead and we make a getCursor method? Depends on the route we choose to take

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/ChunkedDownloadUtils.java
gapra-msft: Yeah I can rename - I like that name.

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/HelperSpec.groovy
gapra-msft: I think I may be able to get rid of this class entirely. I'll look into that as I re-review the tests

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/ChunkFactory.java
```diff
@@ -0,0 +1,94 @@
...
+import reactor.core.publisher.Flux;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Factory class for {@link Chunk}.
+ */
+class ChunkFactory {
+
+    /* TODO (gapra): This should probably be configurable by a user. */
```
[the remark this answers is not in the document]
gapra-msft: By that comment I was referring to the chunk size (1MB). I placed it poorly. The performance of the downloads can vary so in case a customer is like - this is super slow, we can allow them to increase that threshold or decrease to improve perf depending on their situation. I think it would be a nice to have kind of thing

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
```diff
@@ -0,0 +1,88 @@
...
+    /*TODO (gapra) : It may be possible to unduplicate the code below as well to share between downloadToFile but wasnt immediately obvious to me */
+    public Flux<ByteBuffer> download() {
+        ParallelTransferOptions options = new ParallelTransferOptions()
+            .setBlockSizeLong(blockSize);
+        BlobRequestConditions requestConditions = new BlobRequestConditions();
+
+        Function<BlobRange, Mono<BlobDownloadAsyncResponse>> downloadFunc = range ->
+            client.downloadWithResponse(range, null, new BlobRequestConditions(), false);
+
+        return ChunkedDownloadUtils.getSetupMono(range, options, requestConditions, downloadFunc,
```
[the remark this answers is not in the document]
gapra-msft: Renamed to downloadFirstChunk. This method is meant to download the first chunk then get information such as the size of the blob to download the rest of it.

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFluxTest.groovy
gapra-msft: related to comment below

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedResponse.java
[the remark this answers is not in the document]
kasobol-msft: yes, both should throw or both should work.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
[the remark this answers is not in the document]
kasobol-msft: If server does that then we probably can just drop `Integer.min` and let the server do this.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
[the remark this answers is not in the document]
kasobol-msft: which is not far away from one of applications of continuation token.
normally user doesn't touch it. however, if the enumeration is pretty long they might be saving the token when error occur and just use it later to resume - i.e. use saved token to initialize iterable and let it continue without handholding the token.

If we choose to overload the meaning of continuation token then we need to make sure to add javadoc explaining how this work here.

Btw. looks like in .NET PR continuationToken == cursor. See 
https://github.com/Azure/azure-sdk-for-net/pull/11692/files#diff-c69bdde9b907d209a3c6315196fdb137
and
https://github.com/seanmcc-msft/azure-sdk-for-net/blob/feature/storage/changeFeed/sdk/storage/Azure.Storage.Blobs.ChangeFeed/README.md#resume-with-cursor

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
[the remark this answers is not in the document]
gapra-msft: By the service does it - I mean for existing listBlobs/listContainers APIs. This number here is not restricted by the service

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Chunk.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private final String chunkPath; /* Chunk path. */
+    private final ChangefeedCursor shardCursor; /* Cursor associated with parent shard. */
+    private final AvroReader avroReader;
+
+    /**
+     * Creates a new Chunk.
+     */
+    Chunk(String chunkPath, ChangefeedCursor shardCursor, AvroReader avroReader) {
+        StorageImplUtils.assertNotNull("avroReader", avroReader);
```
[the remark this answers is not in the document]
gapra-msft: Originally didnt cause I did it in the ChunkFactory but I see no harm in adding this here too

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/ChunkFactory.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final long DEFAULT_HEADER_SIZE = 4 * Constants.KB;
+    private static final long DEFAULT_BODY_SIZE = Constants.MB;
+
+    private final AvroReaderFactory avroReaderFactory;
+    private final BlobLazyDownloaderFactory blobLazyDownloaderFactory;
+
+    /**
+     * Creates a default instance of the ChunkFactory.
+     */
+    ChunkFactory() {
```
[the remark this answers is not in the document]
gapra-msft: I think this makes it easier to reason about, don't you think?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Changefeed.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Get all the events for the Changefeed.
+     * @return A reactive stream of {@link BlobChangefeedEventWrapper}
+     */
+    Flux<BlobChangefeedEventWrapper> getEvents() {
+        return validateChangefeed()
+            .flatMap(ignore -> populateLastConsumable())
+            .flatMapMany(ignore -> listYears())
```
alzimmermsft, line 69: Given these values are ignored, could we use `then` and `thenMany`?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Changefeed.java
```diff
@@ -0,0 +1,138 @@
...
+     */
+    private Mono<OffsetDateTime> populateLastConsumable() {
+        /* We can keep the entire metadata file in memory since it is expected to only be a few hundred bytes. */
+        return DownloadUtils.downloadToString(this.client, METADATA_SEGMENT_PATH)
+            /* Parse JSON for last consumable. */
+            .flatMap(json -> {
+                try {
+                    ObjectMapper objectMapper = new ObjectMapper();
+                    JsonNode jsonNode = objectMapper.readTree(json);
+                    this.lastConsumable = OffsetDateTime.parse(jsonNode.get("lastConsumable").asText());
```
alzimmermsft, line 100: Any chance this class could be used in multiple threads at once? Setting these class level variables could result in timing issues.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Changefeed.java
```diff
@@ -0,0 +1,138 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.io.IOException;
+import java.time.OffsetDateTime;
+
+/**
+ * A class that represents a Changefeed.
+ *
```
alzimmermsft, line 23: `* <p>`

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Shard.java
```diff
@@ -0,0 +1,97 @@
...
+                    /* If we hit the chunk specified in the user cursor, set pass to true and pass this chunk
+                       and any subsequent chunks through. */
+                    if (userCursor.getChunkPath().equals(chunkPath)) {
+                        pass.set(true); /* This allows us to pass subsequent chunks through.*/
+                        return true; /* This allows us to pass this chunk through. */
+                    } else {
+                        return false;
+                    }
+                }
+            });
```
alzimmermsft, line 94: Let me know if I understand this incorrectly.

So the structure of these paths would be similar to the following

```
"$blobchangefeed/log/00/2019/02/22/1810/",
"$blobchangefeed/log/01/2019/02/22/1810/"
```
This is checking if we hit a chunk with a matching cursor to what a customer passed such as `2019/02`.

Could this logic be simplified, with a small overhead in perf, to just check `chunkPath.startsWith(userCursor.getChunkPath())` and that be the only logic needed to be checked.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/implementation/util/DownloadUtils.java
```diff
@@ -0,0 +1,24 @@
...
+public class DownloadUtils {
+
+    /**
+     * Reduces a Flux of ByteBuffer into a Mono of String
+     */
+    public static Mono<String> downloadToString(BlobContainerAsyncClient client, String blobPath) {
+        return client.getBlobAsyncClient(blobPath)
+            .download()
+            .reduce(new StringBuilder(), (sb, buffer) -> {
+                sb.append(new String(buffer.array(), StandardCharsets.UTF_8));
```
alzimmermsft, line 20: Would watch out on the usage of `ByteBuffer.array` as this will throw if the `ByteBuffer` is a `DirectByteBuffer`. May want to add in using `FluxUtil.byteBufferToArray`.

```java
sb.append(new String(FluxUtil.byteBufferToArray(buffer), StandardCharsets.UTF_8));
```

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/implementation/util/TimeUtils.java
```diff
@@ -0,0 +1,122 @@
...
+            return null;
+        }
+        String[] splitPath = path.split("/");
+
+        return OffsetDateTime.of(
+            Integer.parseInt(splitPath[2]), /* year */
+            splitPath.length < 4 ? 1 : Integer.parseInt(splitPath[3]), /* month */
+            splitPath.length < 5 ? 1 : Integer.parseInt(splitPath[4]), /* day */
+            splitPath.length < 6 ? 0 : Integer.parseInt(splitPath[5]) / 100, /* hour */
+            0, /* minute */
```
alzimmermsft, line 35: Given a segment path `idx/segments/2020/05/28/1047/meta.json` why do we always set minute to `0`? Instead should we set it to the following

```java
splitPath.length < 6 ? 0 : Integer.parseInt(splitPath[5]) % 100 /* minute */
```

sdk/parents/azure-client-sdk-parent/pom.xml
```diff
@@ -534,6 +534,7 @@
               com.azure.core.test*:
               com.azure.endtoend*:
               com.azure.perf*
+              com.azure.storage.internal.avro*
```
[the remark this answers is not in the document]
gapra-msft: I think all the code is in com.azure.storage.internal.avro.implementation

Should that be safe as well?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedClientBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+     * Creates a {@link BlobChangefeedAsyncClient}.
+     *
+     * <p><strong>Code sample</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.changefeed.BlobChangefeedClientBuilder#buildAsyncClient}
+     *
+     * @return a {@link BlobChangefeedAsyncClient} created from the configurations in this builder.
+     */
+    public BlobChangefeedAsyncClient buildAsyncClient() {
+        BlobServiceVersion serviceVersion = version != null ? version : BlobServiceVersion.getLatest();
```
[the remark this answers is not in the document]
gapra-msft: Yeah you're right. This should be correctly populated by the Service Client

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFluxFactory.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Gets a new instance of {@link BlobChangefeedPagedFlux}.
+     *
+     * @param client The {@link BlobContainerAsyncClient changefeed client}.
+     * @param startTime The {@link OffsetDateTime start time}.
+     * @param endTime The {@link OffsetDateTime end time}.
+     */
+    BlobChangefeedPagedFlux getBlobChangefeedPagedFlux(BlobContainerAsyncClient client, OffsetDateTime startTime,
```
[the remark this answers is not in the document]
gapra-msft: Changed

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
```diff
@@ -0,0 +1,91 @@
...
+    public Flux<ByteBuffer> download() {
+        ParallelTransferOptions options = new ParallelTransferOptions()
+            .setBlockSizeLong(blockSize);
+        BlobRequestConditions requestConditions = new BlobRequestConditions();
+
+        Function<BlobRange, Mono<BlobDownloadAsyncResponse>> downloadFunc = range
+            -> client.downloadWithResponse(range, null, new BlobRequestConditions(), false);
+
+        return ChunkedDownloadUtils.downloadFirstChunk(range, options, requestConditions, downloadFunc,
+            Schedulers.immediate())
```
rickle-msft, line 58: Why specify the scheduler here?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloaderFactory.java
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Gets a new instance of a BlobLazyDownloader.
+     *
+     * @param client The blob client.
+     * @param blockSize The block size to download.
+     * @param offset The offset to start downloading from.
+     * @return {@link BlobLazyDownloader}
+     */
+    BlobLazyDownloader getBlobLazyDownloader(BlobAsyncClient client, long blockSize, long offset) {
```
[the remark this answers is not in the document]
gapra-msft: done

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/ChangefeedFactory.java
```diff
@@ -0,0 +1,55 @@
...
+     * Creates a SegmentFactory with the designated factories.
+     */
+    ChangefeedFactory(SegmentFactory segmentFactory) {
+        this.segmentFactory = segmentFactory;
+    }
+
+    /**
+     * Gets a new instance of a Changefeed.
+     */
+    Changefeed getChangefeed(BlobContainerAsyncClient client, OffsetDateTime startTime, OffsetDateTime endTime) {
```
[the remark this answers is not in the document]
gapra-msft: done

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
rickle-msft: It looks like we specified the scheduler in the original download to file code that you refactored. Weird. I don't recall why we would've done that.

In that case, why use immediate here instead of elastic?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
```diff
@@ -0,0 +1,94 @@
...
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.util.List;
+
+/**
+ * Implementation of {@link ContinuablePagedFlux} for Changefeed where the continuation token type is {@link String},
+ * the element type is {@link BlobChangefeedEvent}, and the page type is {@link BlobChangefeedPagedResponse}.
+ */
+public final class BlobChangefeedPagedFlux extends ContinuablePagedFlux<String, BlobChangefeedEvent,
```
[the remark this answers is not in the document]
gapra-msft: I didnt want to extend PagedFluxCore since it had some continuation state behavior that doesnt mix well with Changefeed.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
```diff
@@ -0,0 +1,91 @@
...
+                BlobRequestConditions finalConditions = setupTuple3.getT2();
+
+                int numChunks = ChunkedDownloadUtils.calculateNumBlocks(newCount, options.getBlockSizeLong());
+
+                // In case it is an empty blob, this ensures we still actually perform a download operation.
+                numChunks = numChunks == 0 ? 1 : numChunks;
+
+                BlobDownloadAsyncResponse initialResponse = setupTuple3.getT3();
+                return Flux.range(0, numChunks)
+                    .concatMap(chunkNum -> { /* TODO (gapra) : This was the biggest difference - downloadToFile does
```
rickle-msft, line 70: You might be able to factor out the lambda into your ChunkedDownloadUtils class and then here you concatMap(<refactored lambda>) and in downloadToFile we flatMap(<refactoredLambda>)?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/implementation/util/TimeUtils.java
gapra-msft: In changefeed a segment is a representation of an hour, so I just set min and sec to 0, 0

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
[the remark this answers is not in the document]
gapra-msft: discussed offline, we will throw for now and add in functionality later if required

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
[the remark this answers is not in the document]
gapra-msft: .NET is also using this behavior since it's likely to be what users are used to

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedResponse.java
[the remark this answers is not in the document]
gapra-msft: Based on offline conversation, we are keeping this

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/ChunkFactory.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param client The changefeed container client.
+     * @param chunkPath The path to the chunk blob.
+     * @param shardCursor The parent shard cursor.
+     * @param blockOffset The offset of the block to start reading from. If 0, this indicates we should read the whole
+     *                    avro file from the beginning.
+     * @param objectBlockIndex The index of the last object in the block that was returned to the user.
+     * @return {@link Chunk}
+     */
+    Chunk getChunk(BlobContainerAsyncClient client, String chunkPath, ChangefeedCursor shardCursor,
```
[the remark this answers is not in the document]
gapra-msft: done

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/ChunkFactory.java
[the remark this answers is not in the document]
gapra-msft: fixed

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFluxTest.groovy
gapra-msft: changed name and removed inheritance

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
```diff
@@ -0,0 +1,94 @@
...
+
+        if (continuationToken != null) {
+            return Flux.error(new UnsupportedOperationException("continuationToken not supported. Use "
+                + "client.getEvents(String) to pass in a cursor."));
+        }
+        if (preferredPageSize <= 0) {
+            return Flux.error(new IllegalArgumentException("preferredPageSize > 0 required but provided: "
+                + preferredPageSize));
+        }
+        preferredPageSize = Integer.min(preferredPageSize, DEFAULT_PAGE_SIZE);
```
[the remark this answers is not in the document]
gapra-msft: See this discussion https://github.com/Azure/azure-sdk-for-java/pull/10839#discussion_r425886216

I wanted to make the behavior of this the same as what customers get with service calls. If this seems wrong and we have a good reason why I can change it

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Changefeed.java
gapra-msft: Originally they werent playing well with my error logic but it seems to work now.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
gapra-msft: Changing to elastic and I'll remove the Schedulers method param

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
gapra-msft: Yeah I think that's the best bet

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
```diff
@@ -0,0 +1,91 @@
...
+import reactor.core.scheduler.Schedulers;
+
+import java.nio.ByteBuffer;
+import java.util.function.Function;
+
+/**
+ * FOR INTERNAL USE ONLY.
+ * Class to lazily download a blob.
+ */
+class BlobLazyDownloader {
```
rickle-msft, line 23: This type doesn't seem like it's actually lazy to me. While it is sequential, it will still continue to issue download requests as long as the subscription to the data is not cancelled. I haven't seen yet where this is being used, but it will probably need to apply back pressure in order to actually slow down requests and make it lazy.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Changefeed.java
gapra-msft: Changfeed wasnt designed to be threadsafe just cause it's so dependent on all these states.

gapra-msft: Now that I'm looking at it again though - I'm gonna check to see if this is the only one that is stored like that and used later - cause I think the rest of the classes arent state dependent. I should be able to just pass lastConsumable down and make this part thread safe

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Shard.java
gapra-msft: So shards will be of the form $blobchangefeed/log/00/2019/02/22/1810/, $blobchangefeed/log/01/2019/02/22/1810/

And the chunks under them will be of the form  
$blobchangefeed/log/00/2019/02/22/1810/0000.avro,
$blobchangefeed/log/00/2019/02/22/1810/0001.avro,  $blobchangefeed/log/00/2019/02/22/1810/0002.avro, etc

The user cursor will maybe have stored something that looks like chunkPath = $blobchangefeed/log/00/2019/02/22/1810/0001.avro.

and ideally the filter will return $blobchangefeed/log/00/2019/02/22/1810/0001.avro, and $blobchangefeed/log/00/2019/02/22/1810/0002.avro

So I dont think startsWith would be too applicable here.

gapra-msft: I'll add a comment with the structure of the paths

sdk/parents/azure-client-sdk-parent/pom.xml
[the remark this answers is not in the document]
alzimmermsft: Yup, that should be safe.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/implementation/models/ChangefeedCursor.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /**
+     * Serializes a {@link ChangefeedCursor} into a String.
+     *
+     * @return The resulting serialized cursor.
+     */
+    public String serialize() {
+        try {
+            return new ObjectMapper().writeValueAsString(this);
```
alzimmermsft, line 205: Should make this `ObjectMapper` instance static. For this method should use a `writer()`, for the deserialize case use a `reader()`.

Does the mapper need any configuration to handle empty/null values?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedAsyncClient.java
```diff
@@ -0,0 +1,111 @@
...
+     * Changefeed events are returned in approximate temporal order.
+     *
+     * <p>For more information, see the
+     * <a href="https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-change-feed?tabs=azure-portal">Azure Docs</a>.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.changefeed.BlobChangefeedAsyncClient.getEvents#OffsetDateTime-OffsetDateTime}
+     *
+     * @param startTime Filters the results to return events approximately after the start time.
```
rickle-msft, line 82: Should we give a little more detail about how approximate this is?

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/TimeUtilsTest.groovy
```diff
@@ -0,0 +1,126 @@
...
+        OffsetDateTime.of(2021, 1, 1, 0, 0, 0, 0, ZoneOffset.UTC) | "idx/segments/2020/01/01/0000/meta.json" | OffsetDateTime.of(2019, 1, 1, 0, 0, 0, 0, ZoneOffset.UTC) | false
+        OffsetDateTime.of(2019, 1, 1, 0, 0, 0, 0, ZoneOffset.UTC) | "idx/segments/2020/01/01/0000/meta.json" | OffsetDateTime.of(2019, 1, 1, 0, 0, 0, 0, ZoneOffset.UTC) | false
+    }
+
+    def "validYear"() {
+        expect:
+        TimeUtils.validYear(year, start, end) == valid
+
+        where:
+        start | year | end  || valid
```
rickle-msft, line 89: Formatting

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/implementation/models/BlobChangefeedEventWrapper.java
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT License.
+
+package com.azure.storage.blob.changefeed.implementation.models;
+
+import com.azure.storage.blob.changefeed.models.BlobChangefeedEvent;
+
+/**
+ * FOR INTERNAL USE ONLY.
+ * Represents a wrapper to store a BlobChangefeedEvent along with the BlobChangefeedCursor associated with it.
+ * This wrapper is required since the paging functionality does not have any information about where the
```
rickle-msft, line 11: Unfished

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/TimeUtilsTest.groovy
```diff
@@ -0,0 +1,126 @@
...
+        null                                     || null
+        "idx/segments/2019"                      || OffsetDateTime.of(2019, 1, 1, 0, 0, 0, 0, ZoneOffset.UTC)
+        "idx/segments/2019/"                     || OffsetDateTime.of(2019, 1, 1, 0, 0, 0, 0, ZoneOffset.UTC)
+        "idx/segments/2019/11"                   || OffsetDateTime.of(2019, 11, 1, 0, 0, 0, 0, ZoneOffset.UTC)
+        "idx/segments/2019/11/"                  || OffsetDateTime.of(2019, 11, 1, 0, 0, 0, 0, ZoneOffset.UTC)
+        "idx/segments/2019/11/02"                || OffsetDateTime.of(2019, 11, 2, 0, 0, 0, 0, ZoneOffset.UTC)
+        "idx/segments/2019/11/02/"               || OffsetDateTime.of(2019, 11, 2, 0, 0, 0, 0, ZoneOffset.UTC)
+        "idx/segments/2019/11/02/1700"           || OffsetDateTime.of(2019, 11, 2, 17, 0, 0, 0, ZoneOffset.UTC)
+        "idx/segments/2019/11/02/1700/"          || OffsetDateTime.of(2019, 11, 2, 17, 0, 0, 0, ZoneOffset.UTC)
+        "idx/segments/2019/11/02/1700/meta.json" || OffsetDateTime.of(2019, 11, 2, 17, 0, 0, 0, ZoneOffset.UTC)
```
rickle-msft, line 26: Is there any need to test a path like `idx/segments/2019/11/02/1730/meta.json` at all?

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/BlobLazyDownloaderTest.groovy
```diff
@@ -0,0 +1,131 @@
...
+    @Unroll
+    def "download blockSize"() {
+        setup:
+        byte[] input = uploadHelper(size)
+
+        when:
+        byte[] output = downloadHelper(new BlobLazyDownloader(bc, blockSize, 0))
+
+        then:
+        output == input
```
rickle-msft, line 49: You might be able to use a `Spy` from Spock to help check that you call download the right number of times.

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
```diff
@@ -0,0 +1,209 @@
...
+        when(mockAvroReaderFactory.getAvroReader(any(Flux.class)))
+            .thenReturn(mockAvroReader)
+
+        when:
+        ChunkFactory factory = new ChunkFactory(mockAvroReaderFactory, mockBlobLazyDownloaderFactory, mockContainer)
+        Chunk chunk = factory.getChunk(chunkPath, chunkCursor, 0, 0)
+        def sv = StepVerifier.create(chunk.getEvents().index())
+
+        then:
+        sv.expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 1234, 0) })
```
rickle-msft, line 71: Why are you using a method reference instead of just calling verifyWrapper?

rickle-msft: Can't we get the blockOffset and blockIndex based off the flux index (tuple2.getT1())? Otherwise it seems like hard coding the blockOffset and blockIndex in the test like this is only really testing that we hard coded things the same way here and when generating the mock data

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
```diff
@@ -0,0 +1,209 @@
...
+            .expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 5678, 3) })
+            .expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 9101, 0) })
+            .expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 9101, 1) })
+            .verifyComplete()
+
+        verify(mockContainer).getBlobAsyncClient(chunkPath) || true
+        verify(mockBlobLazyDownloaderFactory).getBlobLazyDownloader(mockBlob, ChunkFactory.DEFAULT_BODY_SIZE, 0) || true
+        verify(mockBlobLazyDownloader).download() || true
+        verify(mockAvroReaderFactory).getAvroReader(Flux.empty()) || true
+        verify(mockAvroReader).readAvroObjects() || true
```
rickle-msft, line 87: Why are you or'ing with true?

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
```diff
@@ -0,0 +1,209 @@
...
+            .expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 5678, 0) })
+            .expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 5678, 1) })
+            .expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 5678, 2) })
+            .expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 5678, 3) })
+            .expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 9101, 0) })
+            .expectNextMatches({ tuple2 -> this.&verifyWrapper(tuple2.getT2(), tuple2.getT1(), 9101, 1) })
+            .verifyComplete()
+
+        verify(mockContainer).getBlobAsyncClient(chunkPath) || true
+        verify(mockBlobLazyDownloaderFactory).getBlobLazyDownloader(mockBlob, ChunkFactory.DEFAULT_HEADER_SIZE) || true
```
rickle-msft, line 118: Can you add some descriptions about why you're expecting this behavior?

rickle-msft: In particular I would expect that because you're using a cursor you would get different data out, but it looks like you're maybe using the same underlying data and just expecting different calls to the internal types (an extra call to get the header, resulting in an extra call to the downloader, and using a different overload of getAvroReader?)

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Segment.java
```diff
@@ -0,0 +1,120 @@
...
+            return FluxUtil.monoError(logger, new UncheckedIOException(e));
+        }
+    }
+
+    private Flux<Shard> getShards(JsonNode node) {
+
+        /* Determine if the segment is finalized.
+           If the segment is not finalized, do not return events for this segment. */
+        JsonNode status = node.get(STATUS);
+        if (!status.asText().equals(FINALIZED)) {
```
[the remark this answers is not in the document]
gapra-msft: thanks for catching that!

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Changefeed.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Populates the last consumable property from changefeed metadata.
+     * Log files in any segment that is dated after the date of the LastConsumable property in the
+     * $blobchangefeed/meta/segments.json file, should not be consumed by your application.
+     */
+    private Mono<OffsetDateTime> populateLastConsumable() {
+        /* We can keep the entire metadata file in memory since it is expected to only be a few hundred bytes. */
+        return DownloadUtils.downloadToString(this.client, METADATA_SEGMENT_PATH)
```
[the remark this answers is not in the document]
gapra-msft: changed

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/implementation/models/ChangefeedCursor.java
gapra-msft: No we don't expect any of the values to be null ever when using ser/deser

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/TimeUtilsTest.groovy
gapra-msft: No real need but I was just including all possible paths in Changefeed Segments

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
gapra-msft: Groovy doesnt work so well with Mockito https://stackoverflow.com/questions/16678126/simple-mockito-verify-works-in-junit-but-not-spock

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
gapra-msft: Re: first comment, it's just invoking the function as a closure. https://stackoverflow.com/questions/1395166/invoke-method-as-closure

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/TimeUtilsTest.groovy
gapra-msft: fixed

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/implementation/models/BlobChangefeedEventWrapper.java
gapra-msft: fixed

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
```diff
@@ -0,0 +1,72 @@
...
+
+        return ChunkedDownloadUtils.downloadFirstChunk(range, options, requestConditions, downloadFunc)
+            .flatMapMany(setupTuple3 -> {
+                long newCount = setupTuple3.getT1();
+                BlobRequestConditions finalConditions = setupTuple3.getT2();
+
+                int numChunks = ChunkedDownloadUtils.calculateNumBlocks(newCount, options.getBlockSizeLong());
+
+                // In case it is an empty blob, this ensures we still actually perform a download operation.
+                numChunks = numChunks == 0 ? 1 : numChunks;
```
anuchandy, line 62: just curious, is the reason for making the download call still is to get a ByteBuffer of size 0 or something else?

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
rickle-msft: Shouldn't you `and` with true, though? It seems like these statements will always trivially evaluate to true

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
rickle-msft: I'm not sure I follow how invoking the method as a closure is different from just invoking the method in this case? Why not just use this.verifyWrapper?

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/TimeUtilsTest.groovy
rickle-msft: Sorry I think the intention of my question was unclear. I was highlighting 1730 as opposed to 1700. Basically something that doesn't align along an hour.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Changefeed.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Validates that changefeed has been enabled for the account.
+     */
+    private Mono<Boolean> validateChangefeed() {
+        return this.client.exists()
+            .flatMap(exists -> {
+                if (exists == null || !exists) {
+                    return FluxUtil.monoError(logger, new RuntimeException("Changefeed has not been enabled for "
+                        + "this account."));
```
anuchandy, line 88: is it the case that when change feed is enabled then storage account will have some special container to store them? so its absence means change-feed-not-enabled?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Shard.java
```diff
@@ -0,0 +1,97 @@
...
+     */
+    private Flux<String> listChunks() {
+        Flux<String> chunks = client.listBlobs(new ListBlobsOptions().setPrefix(shardPath))
+            .map(BlobItem::getName);
+        /* If no user cursor was provided, just return all chunks without filtering. */
+        if (userCursor == null) {
+            return chunks;
+        /* If a user cursor was provided, filter out chunks that come before the chunk specified in the cursor. */
+        } else {
+            AtomicBoolean pass = new AtomicBoolean(); /* Whether or not to pass the event through. */
```
anuchandy, line 80: this `pass` var get shared across all the subscribers of the returned `Flux<String>` instance, if that is not intended then we can wrap filter in a defer.

sdk/storage/azure-storage-internal-avro/src/main/java/com/azure/storage/internal/avro/implementation/AvroReader.java
```diff
@@ -0,0 +1,13 @@
...
+
+package com.azure.storage.internal.avro.implementation;
+
+import reactor.core.publisher.Flux;
+
+/**
+ * An interface that represents an AvroReader.
+ */
+public interface AvroReader {
+    Flux<AvroObject> readAvroObjects();
```
kasobol-msft, line 12: nit. I'd just call it `read` . One can deduct rest from return type.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.storage.blob.changefeed;
```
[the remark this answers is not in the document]
alzimmermsft: Agreed that `util` may not be the best package location, just wanted to call out the [guideline](https://azure.github.io/azure-sdk/java_design.html#java-pagination-subtypes).

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
gapra-msft: it is a little weird. but since the || true comes after, if the verify step errors out indicating failure, it will still error out and not just "work".

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/TimeUtilsTest.groovy
gapra-msft: I don't think we ever expect it though just cause the segments are hourly

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
gapra-msft: just doing normal method call. I'm not sure why I used the closures. I think I googled a thing and it told me to do this.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloaderFactory.java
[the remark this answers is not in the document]
gapra-msft: my bad. I'll fix this

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Changefeed.java
```diff
@@ -0,0 +1,144 @@
...
+     * Log files in any segment that is dated after the date of the LastConsumable property in the
+     * $blobchangefeed/meta/segments.json file, should not be consumed by your application.
+     */
+    private Mono<OffsetDateTime> populateLastConsumable() {
+        /* We can keep the entire metadata file in memory since it is expected to only be a few hundred bytes. */
+        return DownloadUtils.downloadToString(this.client, METADATA_SEGMENT_PATH)
+            /* Parse JSON for last consumable. */
+            .flatMap(json -> {
+                try {
+                    ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
gapra-msft: fixed

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Segment.java
```diff
@@ -0,0 +1,120 @@
...
+        return DownloadUtils.downloadToString(client, segmentPath)
+            .flatMap(this::parseJson)
+            /* Parse the JSON for shards. */
+            .flatMapMany(this::getShards)
+            /* Get all events for each shard. */
+            .concatMap(Shard::getEvents);
+    }
+
+    private Mono<JsonNode> parseJson(String json) {
+        ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
gapra-msft: fixed

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/implementation/models/ChangefeedCursor.java
gapra-msft: fixed the static comment

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedAsyncClient.java
gapra-msft: added detail

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
```diff
@@ -0,0 +1,125 @@
...
+               preferredPageSize elements. */
+            .window(preferredPageSize)
+            /* Convert the BlobChangefeedEventWrappers into BlobChangefeedEvents, and bundle them up with the last
+               element's cursor. */
+            .flatMap(eventWrappers -> {
+                /* 1. cache the Flux to turn it into a HotFlux so we can subscribe to it multiple times. */
+                Flux<BlobChangefeedEventWrapper> cachedEventWrappers = eventWrappers.cache();
+                /* 2. Get the last element in the flux and grab it's cursor. This will be the continuationToken
+                      returned to the user if they want to get the next page. */
+                Mono<ChangefeedCursor> c = cachedEventWrappers.last()
```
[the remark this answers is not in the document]
gapra-msft: Is it possible for eventWrappers to not have any data? Won't window guarantee at least one event is in the flux?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
```diff
@@ -0,0 +1,125 @@
...
+        preferredPageSize = Integer.min(preferredPageSize, DEFAULT_PAGE_SIZE);
+
+        Changefeed changefeed;
+        if (cursor != null) {
+            changefeed = changefeedFactory.getChangefeed(cursor);
+        } else {
+            changefeed = changefeedFactory.getChangefeed(startTime, endTime);
+        }
+
+        return changefeed.getEvents()
```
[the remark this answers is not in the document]
gapra-msft: Good point. Changed to concatMap

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
gapra-msft: That's perhaps a poor comment. 

We already "downloaded" the empty blob in the downloadFirstChunk method above. We want Flux.range below to return a Flux with one element - "0" in it. If you look at the internals of downloadChunk below, when chunkNum = 0, we return whatever we downloaded initially (which would be nothing in this instance) instead of making another download call.

gapra-msft: Does that make any sense?

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Changefeed.java
gapra-msft: Yes, so the changefeed feature relies on the existence of this special container that the service creates. All the blobs we read for this feature are in this container.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedAsyncClient.java
```diff
@@ -0,0 +1,114 @@
...
+            .containerName(CHANGEFEED_CONTAINER_NAME)
+            .pipeline(pipeline)
+            .serviceVersion(version)
+            .buildAsyncClient();
+        AvroReaderFactory avroReaderFactory = new AvroReaderFactory();
+        BlobLazyDownloaderFactory blobLazyDownloaderFactory = new BlobLazyDownloaderFactory();
+        ChunkFactory chunkFactory = new ChunkFactory(avroReaderFactory, blobLazyDownloaderFactory, client);
+        ShardFactory shardFactory = new ShardFactory(chunkFactory, client);
+        SegmentFactory segmentFactory = new SegmentFactory(shardFactory, client);
+        this.changefeedFactory = new ChangefeedFactory(segmentFactory, client);
```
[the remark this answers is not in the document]
gapra-msft: That sounds good.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/Shard.java
gapra-msft: fixed.

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobLazyDownloader.java
anuchandy: sounds good, thanks for clarifying

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedPagedFlux.java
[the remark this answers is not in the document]
anuchandy: The `window` operator can also forward complete event if it's upstream completes without emitting any value. If it is guaranteed that the upstream emits min 1 event then this is not a concern.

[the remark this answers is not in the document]
anuchandy: I see, `last()` is applied not directly on `window` but inside a `flatmap` following it, yes, there is no chance to hit this validation error.

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/BlobLazyDownloaderTest.groovy
gapra-msft: done

sdk/storage/azure-storage-blob-changefeed/src/main/java/com/azure/storage/blob/changefeed/BlobChangefeedAsyncClient.java
```diff
@@ -0,0 +1,114 @@
...
+     * Changefeed events are returned in approximate temporal order.
+     *
+     * <p>For more information, see the
+     * <a href="https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-change-feed?tabs=azure-portal">Azure Docs</a>.
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.changefeed.BlobChangefeedAsyncClient.getEvents#String}
+     *
+     * @param cursor Identifies the portion of the events to be returned with the next get operation.
```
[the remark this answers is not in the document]
gapra-msft: added more info

sdk/storage/azure-storage-blob-changefeed/src/test/java/com/azure/storage/blob/changefeed/ChunkTest.groovy
gapra-msft: havent addressed this quite yet. I can add in an upcoming commit. Basically since we're mocking the tests we don't care about what the AvroReader returns. We are only testing the functionality of the Chunk which asks the AvroReader for some events and the Chunks takes them and combines them in some expected way

gapra-msft: So we just need to validate that the Chunk passes in the right params and then transforms the data correctly. This is true for all the other tests as well

PR #10845, opened 2020-05-06
Fix the pom file version scanner

JimSuplizio:
There was a problem with the pom file version scanner that was exposed when the image version of the Ubuntu changed from 20200406.2 to 20200430.1. the ordering of files returned from 
`Get-ChildItem -Path $Path -Filter pom*.xml -Recurse -File | ForEach-Object` 
This ended up exposing an issue where namespace manager wasn't created in a particular case but was used later which didn't end well. The fix is to hoist the creation the namespace manager outside of the if statement.


JimSuplizio:
@weshaggard as per our conversation. I'm going to ultimately stop using the ForEach-Object funclet and put the results of Get-ChildItem array so, if something like this happens again, I get an actual line number of the error instead of the line number of the funclet.

PR #10895, opened 2020-05-07
Move JsonSerializable/Resource to implementation

xinlian12:


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/ResourceWrapper.java
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.cosmos.models;
```
moderakh, line 4: classes in `com.azure.cosmos.models` are part of our public api.
Does `ResourceWrapper` have to be in `com.azure.cosmos.models` ? can we move it to implementation?

xinlian12: Put it in models for the following reason:
Since we moved Resource class into implementation, so some access modifier has been changed from package private to public. So if I put ResourceWrapper into implementation, then the getResource method will change to public, then will expose methods to client which we did not before. 

But I can definitely remove getResource() from the interface.

xinlian12: Put in the models with the following thinking:

Since we moved Resource into implementation, so some method has changed from private package to public.  If I remove ResourceWrapper to implementation, then the getResource() will change to public which may expose methods we did not/should not? 

But it can be solved by two ways: (if move to implementation)
1. remove getResource() method
2. change to protected, have some implmentionBridge class to use this method from other places

thoughts?

xinlian12:
azp run java - cosmos - tests

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/CosmosError.java
```diff
@@ -4,6 +4,7 @@
 package com.azure.cosmos.models;
 
 import com.azure.cosmos.implementation.Constants;
+import com.azure.cosmos.implementation.Resource;
```
kushagraThapar, line 7: `CosmosError` is in models (public surface) - and is still extending internal types `Resource` and `JsonSerializable`. This is exposed using `CosmosClientException` which can lead to potential exposure of Jackson libraries.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/ResourceWrapper.java
kushagraThapar: Having ResourceWrapper in public should be fine in my opinion if it is not leaking any Jackson libraries. 
No need to do extra work through *BridgeInternals if not required. 
@moderakh  thoughts ?

sdk/cosmos/mgmt/src/main/java/com/azure/management/cosmosdb/DatabaseAccountListKeysResult.java
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+
```
kushagraThapar, line 3: This is a cosmosdb management related library change, not ours :)

kushagraThapar:
Ran `ReadLatency` benchmark on this branch. No perf regressions found. Numbers look good. 


```
2020-05-11 17:02:36,196       [main] INFO  com.azure.cosmos.benchmark.AsyncReadBenchmark - [1000000] operations performed in [130] seconds.
5/11/20 5:02:36 PM =============================================================

#Successful Operations
             count = 1000000
         mean rate = 7638.68 events/second
     1-minute rate = 7507.20 events/second
     5-minute rate = 6672.36 events/second
    15-minute rate = 6343.27 events/second
#Unsuccessful Operations
             count = 0

Latency
             count = 1000000
         mean rate = 7638.80 calls/second
     1-minute rate = 7507.83 calls/second
     5-minute rate = 6675.78 calls/second
    15-minute rate = 6347.79 calls/second
               min = 0.87 milliseconds
               max = 11.13 milliseconds
              mean = 1.28 milliseconds
            stddev = 0.62 milliseconds
            median = 1.20 milliseconds
              75% <= 1.31 milliseconds
              95% <= 1.53 milliseconds
              98% <= 2.06 milliseconds
              99% <= 2.54 milliseconds
            99.9% <= 9.33 milliseconds
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/CompositePath.java
```diff
@@ -26,7 +28,7 @@ public CompositePath() {
      * @param jsonString the json string that represents the included path.
      */
     CompositePath(String jsonString) {
-        super(jsonString);
+        this.jsonSerializable = new JsonSerializable(jsonString);
```
moderakh, line 31: I am not sure yet if we should have JsonSerializableWrapper, but if we decide to go with JsonSerializable, then you could move this code to JsonSerializable and and here just have super(jsonStirng).

xinlian12: yup, good idea to move the common code into the parent class

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosAsyncContainer.java
```diff
@@ -154,7 +154,7 @@ public String getId() {
     public Mono<CosmosAsyncContainerResponse> replace(
         CosmosContainerProperties containerProperties,
         CosmosContainerRequestOptions options) {
-        ModelBridgeInternal.validateResource(containerProperties);
+        ModelBridgeInternal.validateResource(ModelBridgeInternal.getResourceFromResourceWrapper(containerProperties));
```
mbhaskar, line 157: We are validating these resources in RxDocumentClientImpl anyway, so we could skip these altogether? same for database too. Minor change, need not worry about this in this PR.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/CompositePath.java
xinlian12: based on the discussion, will remove JsonSerializableWrapper and ResourceWrapper in  following up PR.

PR #10900, opened 2020-05-07
Container soft delete

kasobol-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -1045,4 +1045,47 @@ private boolean validateNoETag(BlobRequestConditions modifiedRequestConditions)
         }
         return modifiedRequestConditions.getIfMatch() == null && modifiedRequestConditions.getIfNoneMatch() == null;
     }
+
+    /**
+     * Restores a previously created container.  The restored container
```
alzimmermsft, line 1050: We're restoring a deleted container right?

```suggestion
     * Restores a previously delete container. The restored container
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -1045,4 +1045,47 @@ private boolean validateNoETag(BlobRequestConditions modifiedRequestConditions)
...
+     * will be renamed to the name of this {@link BlobContainerAsyncClient}.
+     * If the container associated with this {@link BlobContainerAsyncClient}
+     * already exists, this call will result in a 409 (conflict).
+     * This API is only functional is Container Soft Delete is enabled
```
alzimmermsft, line 1054:
```suggestion
     * This API is only functional if Container Soft Delete is enabled
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
kasobol-msft: done. thank you!

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
kasobol-msft: done. thank you!

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -1045,4 +1045,47 @@ private boolean validateNoETag(BlobRequestConditions modifiedRequestConditions)
         }
         return modifiedRequestConditions.getIfMatch() == null && modifiedRequestConditions.getIfNoneMatch() == null;
     }
+
+    /**
+     * Restores a previously deleted container.  The restored container
+     * will be renamed to the name of this {@link BlobContainerAsyncClient}.
```
JonathanGiles, line 1051: At a higher level, is this what the user would expect? My naive expectation would be that the container is restored into its previous container name?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerListDetails.java
```diff
@@ -42,13 +45,53 @@ public BlobContainerListDetails setRetrieveMetadata(boolean retrieveMetadata) {
...
+    public BlobContainerListDetails setRetrieveDeleted(boolean retrieveDeleted) {
+        this.retrieveDeleted = retrieveDeleted;
+        return this;
+    }
+
     /**
      * @return the listing flags
+     * @deprecated use {@link #toIncludeTypes()}
      */
+    @Deprecated
```
JonathanGiles, line 72: FYI - Deprecation needs to be called out and discussed in more detail whenever it occurs.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerListDetails.java
```diff
@@ -42,13 +45,53 @@ public BlobContainerListDetails setRetrieveMetadata(boolean retrieveMetadata) {
...
     public ListBlobContainersIncludeType toIncludeType() {
         if (this.retrieveMetadata) {
             return ListBlobContainersIncludeType.METADATA;
         }
         return null;
     }
+
+    /**
+     * @return the listing flags
```
JonathanGiles, line 81: There needs to be much more JavaDoc here.

kasobol-msft: I've added more.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerListDetails.java
kasobol-msft: I added more verbose description in `@deprecated` JavaDoc section. Is there any other place we should be calling this out?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
kasobol-msft: Yes. That's how the underlying API works. However there's nothing stopping customer from restoring into same name if nobody created container with same name.

JonathanGiles: That's not quite my question - I'm not asking how the API works, I was seeking to clarify that this aligns with the users expectations (in the absence of any particular API or implementation).

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerListDetails.java
JonathanGiles: Generally - when review requests are made, explicitly mentioning deprecations is the best practice. Eventually I would imagine tooling will be added to our repo to make it automatic, but it's still a good idea to surface it ASAP so we don't make a decision without full consideration.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
kasobol-msft: Oh I think I see what you mean. `restore` sounds like it should be acting on current `BlobContainerClient` so current implementation may sound little bit backwards. What we have here now is more of "create from deleted container". 
I think we have two options now. Either change wording here or reverse this API.

Changing wording would be:
`client.createFrom(string deletedContainerName, string deleteContainerVersion)` or something like that.

Reversing would mean two overloads:
`client.restore(string deletedVersion)` - this would preserve name
`client.restore(string deletedVersion, string destinationContainerName)` - that would land in renamed container.

@JonathanGiles  what do you think?

Also looping in @seanmcc-msft @tg-msft for this discussion. I guess we'd want to be consistent in .NET if we are changing that (PR reference https://github.com/Azure/azure-sdk-for-net/pull/11244 ). And @xiafu-msft  for Python implementation.

seanmcc-msft: The current .NET implementation is on the BlobContainerClient:

```
        /// <summary>
        /// Restores a previously created container.  The restored container
        /// will be renamed to the name of this <see cref="BlobContainerClient"/>.
        /// If the container associated with this <see cref="BlobContainerClient"/>
        /// already exists, this call will result in a 409 (conflict).
        /// This API is only functional is Container Soft Delete is enabled
        /// for the storage account associated with the container.
        /// </summary>
               public virtual Response Restore(
            string deletedContainerName,
            string deletedContainerVersion,
            CancellationToken cancellationToken = default) {...}
```

The current user experience for .NET:
```
BlobContainerClient container = 
    blobServiceClient.GetBlobContainerClient("nameToRestoreContainerTo";
await container.RestoreAsync("deletedContainerName", "deletedContainerVersion");
```
The restored container will "become" the container associated with this BlobContainerClient, and now users can interact with it via this client.

For the 3 proposed options:
`client.createFrom(string deletedContainerName, string deleteContainerVersion) `
I'm not a huge fan of this one.  I'd rather stick with "Restore" in the API name to align with the service.  Also, the name "createFrom"could be confusing to new users who are just trying to create a Container.

`client.restore(string deletedVersion)`
I could get behind this one.

`client.restore(string deletedVersion, string destinationContainerName)`
I think this one doesn't align super well with our current user experience.  Currently, a BlobContainerClient only interact with the Container it is associated with.  This would break that convention.

`BlobServiceClient.restoreContainer(string deletedContainerName, string deletedContainerVersion)`
`BlobServiceClient.restoreContainer(string deletedContainerName, string deletedContainerVersion, string destinationContainerName)`
These are ok, I'd want `restoreContainer` to return a BlobContainerClient for convenience.

I prefer the current .NET implementation the most.

JonathanGiles: I'd recommend storage discuss this internally and work with @tg-msft to land that across the languages as necessary.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -1045,4 +1045,47 @@ private boolean validateNoETag(BlobRequestConditions modifiedRequestConditions)
...
+     * If the container associated with this {@link BlobContainerAsyncClient}
+     * already exists, this call will result in a 409 (conflict).
+     * This API is only functional if Container Soft Delete is enabled
+     * for the storage account associated with the container.
+     *
```
rickle-msft, line 1056: The docs should point to the REST docs and should have a code snippet.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -1045,4 +1045,47 @@ private boolean validateNoETag(BlobRequestConditions modifiedRequestConditions)
...
+     * @param deletedContainerName The name of the previously deleted container.
+     * @param deletedContainerVersion The version of the previously deleted container.
+     * @return A reactive response signaling completion.
+     */
+    public Mono<Void> restore(String deletedContainerName, String deletedContainerVersion) {
```
rickle-msft, line 1061: Was there a discussion around restore vs undelete? Restore sounds cleaner, but we already have undelete blobs, so not sure if we want to be consistent.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
```diff
@@ -1045,4 +1045,47 @@ private boolean validateNoETag(BlobRequestConditions modifiedRequestConditions)
...
+    public Mono<Response<Void>> restoreWithResponse(String deletedContainerName, String deletedContainerVersion) {
+        try {
+            return withContext(context -> restoreWithResponse(deletedContainerName, deletedContainerVersion, context));
+        } catch (RuntimeException ex) {
+            return monoError(logger, ex);
+        }
+    }
+
+    Mono<Response<Void>> restoreWithResponse(String deletedContainerName, String deletedContainerVersion,
+                                                    Context context) {
```
rickle-msft, line 1086: I think we also want to be adding this [tracing context](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java#L333)

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
rickle-msft: This also differs from blobs in that with blobs the client must point to the deleted blob. Here the client points to the new destination. It may be worth discussing whether we depart from the rest api and have the client point to the deleted container and then do some work internally to switch things around.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/HelperTest.groovy
```diff
@@ -243,9 +242,9 @@ class HelperTest extends APISpec {
...
-        "c"           | "b"      | "id"   | OffsetDateTime.now() || "bs"             | "\n\n%s\n" + "/blob/%s/c/b\n\n\n\n" + BlobServiceVersion.getLatest().getVersion() + "\nbs\nid\n\n\n\n\n"
-        "c"           | "b"      | null   | OffsetDateTime.now() || "b"              | "\n\n%s\n" + "/blob/%s/c/b\n\n\n\n" + BlobServiceVersion.getLatest().getVersion() + "\nb\n\n\n\n\n\n"
-        "c"           | null     | null   | OffsetDateTime.now() || "c"              | "\n\n%s\n" + "/blob/%s/c\n\n\n\n" + BlobServiceVersion.getLatest().getVersion() + "\nc\n\n\n\n\n\n"
+        "c"           | "b"      | "id"   | OffsetDateTime.now() || "bs"             | "\n\n%s\n" + "/blob/%s/c/b\n\n\n\n" + BlobServiceSasVersion.getLatest().getVersion() + "\nbs\nid\n\n\n\n\n"
+        "c"           | "b"      | null   | OffsetDateTime.now() || "b"              | "\n\n%s\n" + "/blob/%s/c/b\n\n\n\n" + BlobServiceSasVersion.getLatest().getVersion() + "\nb\n\n\n\n\n\n"
+        "c"           | null     | null   | OffsetDateTime.now() || "c"              | "\n\n%s\n" + "/blob/%s/c\n\n\n\n" + BlobServiceSasVersion.getLatest().getVersion() + "\nc\n\n\n\n\n\n"
```
rickle-msft: Can you also add an entry in the changelog, please? Not sure about other platforms, but in Java we generally update the changelog with each PR

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
```diff
@@ -1423,4 +1427,170 @@ class ContainerAPITest extends APISpec {
...
     }
+
+    def "Restore Container"() {
+        given:
+        def cc1 = primaryBlobServiceClient.getBlobContainerClient(generateContainerName())
+        cc1.create()
+        def blobName = generateBlobName()
+        cc1.getBlobClient(blobName).upload(defaultInputStream.get(), 7)
+        cc1.delete()
+        def blobContainerItem = primaryBlobServiceClient.listBlobContainers(
```
rickle-msft, line 1438: Why do you need to do a listing here to get the name of the blobContainerItem? Seems like you could just access the container name from the client directly

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
```diff
@@ -1423,4 +1427,170 @@ class ContainerAPITest extends APISpec {
...
+        if (!playbackMode()) {
+            Thread.sleep(30000)
+        }
+
+        when:
+        def response = cc1.restoreWithResponse(blobContainerItem.getName(), blobContainerItem.getVersion(),
+            Duration.ofMinutes(1), Context.NONE)
+
+        then:
+        response != null
```
rickle-msft, line 1504: It would be good to validate some of the values on the response if there are any to make sure they're being set correctly.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
```diff
@@ -1423,4 +1427,170 @@ class ContainerAPITest extends APISpec {
...
+                .setPrefix(cc1.getBlobContainerName())
+                .setDetails(new BlobContainerListDetails().setRetrieveDeleted(true)),
+            null).first()
+
+        if (!playbackMode()) {
+            Thread.sleep(30000)
+        }
+
+        when:
+        cc1.restore(blobContainerItem.getName(), blobContainerItem.getVersion()).block()
```
rickle-msft, line 1527: I think we're trying to move towards using the StepVerifier for async tests instead of blocking.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
```diff
@@ -148,6 +149,32 @@ class ServiceAPITest extends APISpec {
         containers.each { container -> container.delete() }
     }
 
+    def "List deleted"() {
```
xiafu-msft, line 152: Do we have these two test cases? I didn't find them. 
1. list containers including deleted ones using sas token as credential
2. undelete containers using sas token as credential
Though it's likely that those two cases are not working, at least I haven't figured out why they are not working in Python 😀

seanmcc-msft: I also don't have those test cases .NET

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
kasobol-msft: @rickle-msft  We have a discussion started above - https://github.com/Azure/azure-sdk-for-java/pull/10900#discussion_r421834330 - please see ideas there and share your pick.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
kasobol-msft: In order to restore container  I need two things - name and version. There doesn't seem to be different way to get version of deleted container other than listing.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
rickle-msft: I think I agree with Sean and what Kamil has. Best option would be to align with undelete on blobs, but since that only restores in place, it doesn't offer much guidance here. Perhaps the next best thing is to look at copy, which has the client target the destination and then passes the source as a parameter. I think what .NET is doing most closely aligns with that, and if undelete for blobs ever allows you to move the blob as a part of that operation, we should have the source be a parameter.

I definitely think there should be an easy restore-in-place overload for cases where the source and destination are the same, so the customer just doesn't need to pass in a name at all. And I think we should rework the javadocs here to be a little clearer about what's happening. 

Also, for the sake of consistency with blobs, I think we should call this undelete. Since we link to the REST docs in our javadocs, I don't really see a likelihood that customers will get lost hunting through rest docs for the corresponding method.

tg-msft: In terms of existing APIs and patterns, this feels like a mix of [Blob's Undelete](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Blobs/api/Azure.Storage.Blobs.netstandard2.0.cs#L987) and [DataLake's Rename](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Files.DataLake/api/Azure.Storage.Files.DataLake.netstandard2.0.cs#L45) operation.

I really like Sean's suggestion because I think that removes all the confusion about how to use the API:

```C#
class BlobServiceClient
{
    Response<BlobContainerClient> RestoreContainer(
        string deletedContainerName,
        string deletedContainerVersion,
        string destinationContainerName = null);
}
```

This doesn't hide the `destinationContainerName` concept in the BlobContainerClient you use to invoke the API.  Instead it's something you can pass if you want it to change or leave null to keep the same name.

rickle-msft: @tg-msft Both undelete and rename have the client pointing to the target rather than accessing them from the parent container. If restore is similar to both of those, why would introduce a different calling pattern?

kasobol-msft: By looking at `BlobContainerClient.create` and `BlobContainerClient.delete` family (which seem to be relatives of restore - you have to create and delete container in order to restore, same story for .NET) I noticed that we have them both at `BlobContainerClient` and `BlobServiceClient`. So, I'd suggest to keep what we already have at Container level (plus extra overload that doesn't take name) and add mirroring methods to ServiceClient.

@tg-msft @seanmcc-msft what do you think?

seanmcc-msft: @kasobol-msft, sounds good.

So for STG 73, it looks like users will only be able to restore containers under their original name (based on Emily's ongoing email thread with them, there's a bug), with the restore + rename functionality coming in 75.  So we could hold off on the overload that takes new name for now.

tg-msft: I completely agree on `Undelete` vs `Restore`.  We should use the same name for everything and I've left that feedback on one of the .NET PRs.  In .NET it doesn't matter which way we move thanks to `[EditorBrowsable(Never)]`, but I'd vote for consistency with whatever the other languages are going to call this.

The reason I like moving this method up to the parent client is because the return type makes more sense.  When I call `BlobClientContainer.Rename(name)`, I get why you're returning a new client _(if I understand that clients are effectively cloud pointers)_.  If I call `BlobContainerClient.Restore()` I have no idea why it's returning a new client instance _("is it actually the same instance, or is it a new instance slightly different from my last one, or ...?")_.  If I call `BlobServiceClient.Restore(name)` it's really easy to understand that the returned client represents the restored container.  I don't feel strongly about this if everyone else prefers keeping it on `BlobContainerClient`.

The one thing I wouldn't do is put it on both `BlobServiceClient` and `BlobContainerClient`.  That pattern is an on-ramp where we have GetClient/List/Meta-operations on the parent, Create/Delete on both, and everything else on the child.  If you want to explore expanding the set of operations on both, we should push the architects to clarify the guidelines around hierarchical clients because they're not all written down today.  That discussion was a very large can of worms.  :smile:

kasobol-msft: Thanks @tg-msft @seanmcc-msft @rickle-msft for inputs.

I'm going to pull that API up to service client.

The prevailing reason for me would be that we actually need at least a tuple - name and versionid of deleted container - something that's only available at `BlobServiceClient` level through listing. So one won't be able to complete the whole experience with `BlobContainerClient` alone.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ServiceAPITest.groovy
kasobol-msft: @xiafu-msft We didn't add any new permission to SAS tokens for that. So adding test cases for that doesn't seem to bring additional coverage as we already test "write" permission elsewhere.
If that doesn't work then it's server side bug I think.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
kasobol-msft: done.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
kasobol-msft: ok.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
kasobol-msft:
```  
client.listBlobContainers(listBlobContainersOptions, null).forEach(
    deletedContainer -> {
        BlobContainerClient blobContainerClient = client.undeleteBlobContainer(
            deletedContainer.getName(), deletedContainer.getVersion());
    }
);
```

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/specialized/HelperTest.groovy
kasobol-msft: done

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
kasobol-msft: added.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
kasobol-msft: We landed on "undelete" and moving up to service client.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
kasobol-msft: I added snippets and left TODO for links as we don't have them yet.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobContainerListDetails.java
kasobol-msft: I ended up converting this to private method inside client.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobContainerAsyncClient.java
seanmcc-msft: https://github.com/Azure/azure-sdk-for-net/pull/12025

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -325,14 +325,37 @@ public String getAccountUrl() {
...
     }
 
+    /**
+     * Converts {@link BlobContainerListDetails} into list of {@link ListBlobContainersIncludeType}
+     * that contains only options selected. If no option is selected then null is returned.
+     *
+     * @return a list of selected options converted into {@link ListBlobContainersIncludeType}, null if none
+     * of options has been selected.
+     */
+    private List<ListBlobContainersIncludeType> toIncludeTypes(BlobContainerListDetails blobContainerListDetails) {
```
rickle-msft, line 342: Why is this here now instead of on the Details type? Is it so we don't have to have a public method on Details to allow access across packages?

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -717,4 +740,123 @@ private void throwOnAnonymousAccess() {
...
+     * for the storage account associated with the container.
+     * For more information, see the
+     * <a href="TBD">Azure Docs</a>. TODO (kasobol-msft) add link to REST API docs
+     *
+     * <p><strong>Code Samples</strong></p>
+     *
+     * {@codesnippet com.azure.storage.blob.BlobServiceAsyncClient.undeleteBlobContainer#String-String}
+     *
+     * @param deletedContainerName The name of the previously deleted container.
+     * @param deletedContainerVersion The version of the previously deleted container.
```
rickle-msft, line 758: Unless the rest docs already say this, maybe we should add where this version can be retrieved from

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
```diff
@@ -717,4 +740,123 @@ private void throwOnAnonymousAccess() {
...
+     *
+     * {@codesnippet com.azure.storage.blob.BlobServiceAsyncClient.undeleteBlobContainer#String-String-String}
+     *
+     * @param deletedContainerName The name of the previously deleted container.
+     * @param deletedContainerVersion The version of the previously deleted container.
+     * @param destinationContainerName The name of the destination container.
+     * @return A {@link Mono} containing a {@link BlobContainerAsyncClient} used
+     * to interact with the restored container.
+     */
+    public Mono<BlobContainerAsyncClient> undeleteBlobContainer(
```
rickle-msft, line 788: I think we can get rid of this overload and the next one. We can have just the min and then the max is WithResponse. 

Also, should we move the destinationContainerName into UndeleteContainerOptions so we don't have to add overloads later. Continuing the conversation we had around tags.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
kasobol-msft: We don't have rest docs yet so I'd hold on with that. However I've showcased how to get version id in the sample snippets. I think that should be enough.

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
kasobol-msft: This is the result of discussion on the APIView. We don't expect customer to call this and have option to hide it as there is only one place this is being used.

sdk/storage/azure-storage-blob/src/test/java/com/azure/storage/blob/ContainerAPITest.groovy
rickle-msft: That's right. I forgot about the version

sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/BlobServiceAsyncClient.java
kasobol-msft: Done

PR #10911, opened 2020-05-07
Update building and testing section to refer to wiki

srnagar:


joshfree:
I requested this change 😄 -- In general, I prefer having the contributing.md be the source of truth, but my high level goal is to make sure there isn't redundant data that gets out of sync between the contributing.md and the building and testing section in the wiki.  Having one link to the other for these instructions is my preference.

PR #10940, opened 2020-05-08
Servicebus track2 sync queue up multiple receive calls 

hemanttanwar:
Queue  up multiple receive on same instance of receiver if they come in parallel.

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
@@ -247,4 +247,8 @@ class SynchronousMessageSubscriber extends BaseSubscriber<ServiceBusReceivedMess
     long getRequested() {
         return this.requested;
     }
+
+    boolean isSubscriberInitialized() {
+        return this.subscriberInitialized;
+    }
 }
--- a/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RetriableWorkItem.java
+++ b/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/RetriableWorkItem.java
@@ -11,7 +11,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 /**
  * Represents a work item that can be scheduled multiple times.
  */
-class DeclaredRetriableWorkItem {
+class RetriableWorkItem {
     private final AtomicInteger retryAttempts = new AtomicInteger();
     private final MonoSink<Void> monoSink;
     private final TimeoutTracker timeoutTracker;
--- a/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/SynchronousMessageSubscriberTest.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/test/java/com/azure/messaging/servicebus/SynchronousMessageSubscriberTest.java
@@ -9,33 +9,23 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.mockito.ArgumentCaptor;
-import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.MockitoAnnotations;
+import org.reactivestreams.Subscription;
 import reactor.test.StepVerifier;
 
 import java.time.Duration;
-import java.time.Instant;
-import java.util.List;
 
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.ArgumentMatchers.isNull;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyZeroInteractions;
+import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.when;
 
 /**
- * Unit test for sync subscriber
+ * Unit test for sync subscriber.
  */
 public class SynchronousMessageSubscriberTest {
 
-    private static final String NAMESPACE = "my-namespace";
-    private static final String ENTITY_NAME = "my-servicebus-entity";
+    private static final int PREFETCH = 1;
 
     @Mock
     private SynchronousReceiveWork work1;
@@ -43,23 +33,10 @@ public class SynchronousMessageSubscriberTest {
     @Mock
     private SynchronousReceiveWork work2;
 
-    @Captor
-    private ArgumentCaptor<ServiceBusMessage> singleMessageCaptor;
-
-    @Captor
-    private ArgumentCaptor<List<ServiceBusMessage>> messageListCaptor;
-
-    @Captor
-    private ArgumentCaptor<Instant> scheduleMessageCaptor;
-
-    @Captor
-    private ArgumentCaptor<Long> cancelScheduleMessageCaptor;
+    @Mock
+    private Subscription subscription;
 
     private SynchronousMessageSubscriber syncSybscriber;
-
-    private static final Duration RETRY_TIMEOUT = Duration.ofSeconds(10);
-    private static final int prefetch = 1;
-    private static final String TEST_CONTENTS = "My message for service bus queue!";
 
     @BeforeAll
     static void beforeAll() {
@@ -82,22 +59,72 @@ public class SynchronousMessageSubscriberTest {
         Mockito.framework().clearInlineMocks();
     }
 
+    /**
+     * Test that if prefetch is large value, it will be the one requested.
+     */
+    @Test
+    void workAddedAndLargePrefetch() {
+        // Arrange
+        when(work1.getId()).thenReturn(1L);
+
+        // Act
+        syncSybscriber = new SynchronousMessageSubscriber(100, work1);
+
+        // Assert
+        Assertions.assertEquals(1, syncSybscriber.getWorkQueueSize());
+        Assertions.assertEquals(100, syncSybscriber.getRequested());
+
+    }
+
+    /**
+     * Test that if prefetch is small value than work, larger value be requested.
+     */
     @Test
     void workAddedInQueueOnCreation() {
+        // Arrange & Act
+        when(work1.getNumberOfEvents()).thenReturn(3);
+        syncSybscriber = new SynchronousMessageSubscriber(0, work1);
 
-       syncSybscriber = new SynchronousMessageSubscriber(prefetch, work1);
-
+        // Assert
         Assertions.assertEquals(1, syncSybscriber.getWorkQueueSize());
+        Assertions.assertEquals(3, syncSybscriber.getRequested());
+
+    }
+
+    /**
+     * A work get queued in work queue.
+     */
+    @Test
+    void queueWorkTest() {
+        // Arrange
+        syncSybscriber = new SynchronousMessageSubscriber(PREFETCH, work1);
+
+        // Act
+        syncSybscriber.queueWork(work2);
+
+        // Assert
+        Assertions.assertEquals(2, syncSybscriber.getWorkQueueSize());
         Assertions.assertEquals(1, syncSybscriber.getRequested());
 
     }
 
+    /**
+     * When we call hookOnSubscribe, the sync subscriber is initialised.
+     */
     @Test
-    void queueWork() {
+    void hookOnSubscribeTest() {
+        // Arrange
+        syncSybscriber = new SynchronousMessageSubscriber(PREFETCH, work1);
+        when(work1.getTimeout()).thenReturn(Duration.ofSeconds(10));
+        when(work1.isTerminal()).thenReturn(true);
+        doNothing().when(subscription).request(1);
 
-        syncSybscriber = new SynchronousMessageSubscriber(prefetch, work1);
-        syncSybscriber.queueWork(work2);
-        Assertions.assertEquals(2, syncSybscriber.getWorkQueueSize());
+        // Act
+        syncSybscriber.hookOnSubscribe(subscription);
+
+        // Assert
+        Assertions.assertTrue(syncSybscriber.isSubscriberInitialized());
+        Assertions.assertEquals(0, syncSybscriber.getWorkQueueSize());
         Assertions.assertEquals(1, syncSybscriber.getRequested());
 
     }


sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -41,8 +42,11 @@
         .setIsAutoCompleteEnabled(false)
         .setMaxAutoLockRenewalDuration(Duration.ZERO);
 
-    private final AtomicReference<EmitterProcessor<ServiceBusReceivedMessageContext>> messageProcessor =
+    private final AtomicReference<EmitterProcessor<ServiceBusReceivedMessageContext>> messageSource =
```
conniey, line 45: Why is this an atomic reference? If I remember correctly you are synchronizing on a lock. Only one thread can be in the method.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -41,8 +42,11 @@
...
         new AtomicReference<>();
+    private final AtomicReference<EmitterProcessor<SynchronousMessageSubscriber>> workQueueProcessor =
```
conniey, line 47: Why do you need another emitter processor? Processors are useful when it listens upstream and publishes downstream. In this case, you are using it as a sink.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -627,20 +631,40 @@ private void queueWork(int maximumMessageCount, Duration maxWaitTime,
...
 
             final SynchronousReceiveWork work = new SynchronousReceiveWork(id, maximumMessageCount, maxWaitTime,
                 emitter);
             final SynchronousMessageSubscriber syncSubscriber = new SynchronousMessageSubscriber(work);
-            logger.info("[{}]: Started synchronous message subscriber.", id);
 
-            if (emitterProcessor == null) {
-                emitterProcessor = this.asyncClient.receive(DEFAULT_RECEIVE_OPTIONS)
+            EmitterProcessor<SynchronousMessageSubscriber> workProcessor = workQueueProcessor.get();
+            workProcessor.onNext(syncSubscriber);
```
conniey, line 640: These should be done through a sink.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,166 @@
...
+import reactor.core.publisher.BaseSubscriber;
+import reactor.core.publisher.Mono;
+
+import java.util.Queue;
+import java.util.Timer;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+class LongLivedMessageSubscriber extends BaseSubscriber<ServiceBusReceivedMessageContext> {
```
conniey, line 15: This can be private static.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,166 @@
...
+
+    private final ClientLogger logger = new ClientLogger(SynchronousMessageSubscriber.class);
```
conniey, line 17: Organise variables by final and not.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,166 @@
...
+    private final AtomicBoolean isDisposed = new AtomicBoolean(false);
+    private Queue<SynchronousReceiveWork> workQueue = new ConcurrentLinkedQueue<>();
+    private SynchronousReceiveWork currentWork = null;
+    private final AtomicInteger wip = new AtomicInteger();
+    private volatile Subscription subscription;
+    private long prefetch;
+
+
+    LongLivedMessageSubscriber(long prefetch) {
+        this.prefetch = prefetch;
```
conniey, line 28: final

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * On an initial subscription, will take the first work item, and request that amount of work for it.
+     *
+     * @param subscription Subscription for upstream.
+     */
+    @Override
+    protected void hookOnSubscribe(Subscription subscription) {
+        if (this.subscription == null) {
+            this.subscription = subscription;
```
conniey, line 39: On subscribing you need to request something or it'll throw a debug message.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,166 @@
...
+        drain();
+    }
+
+    private void drain() {
+        // If someone is already in this loop, then we are already clearing the queue.
+        if (!wip.compareAndSet(0, 1)) {
+            return;
+        }
+        // Drain queue..
+        Disposable drainQueueDisposable = Mono.just(true)
```
conniey, line 99: Why just Mono.just(true) and not drainQueue?

conniey: or even Mono.empty()

hemanttanwar: I want to trigger dainQueue() asynchronously , specially for multithreaded env.  Mono.empty() would not trigger subscribe

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
hemanttanwar: I am planning to remove `prefetch` altogether. We are discussing this on teams.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
hemanttanwar: It is because checkstyle complain, I really do not need atomic reference because I synchronizing.

conniey: What does checkstyle complain about?

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,168 @@
...
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+ class LongLivedMessageSubscriber extends BaseSubscriber<ServiceBusReceivedMessageContext> {
+
+    private final ClientLogger logger = new ClientLogger(SynchronousMessageSubscriber.class);
+    private final AtomicBoolean isDisposed = new AtomicBoolean(false);
```
conniey, line 17: The default ctor sets the boolean to false.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,168 @@
...
+    private final long prefetch;
+    private final AtomicInteger wip = new AtomicInteger();
+
+    private volatile Subscription subscription;
+
+    private Queue<SynchronousReceiveWork> workQueue = new ConcurrentLinkedQueue<>();
+    private SynchronousReceiveWork currentWork = null;
+    private Disposable timeoutOperation;
+     private Disposable drainQueueDisposable;
```
conniey, line 26: nit: format your document

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
conniey: Why are you planning on removing prefetch? This would be your initial request in `hookOnSubscribe`

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
conniey: I mistook this for the rule about request has to be > 0. However, the documentation suggests `hookOnSubscribe` is overridden when you want to alter the request.
https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/BaseSubscriber.java#L80-L82

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * On an initial subscription, will take the first work item, and request that amount of work for it.
+     *
+     * @param subscription Subscription for upstream.
+     */
+    @Override
+    protected void hookOnSubscribe(Subscription subscription) {
+        if (this.subscription == null) {
```
conniey, line 40: This is only subscribed to once. The null check is unnecessary. 
https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/BaseSubscriber.java#L146

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,168 @@
...
+    protected void hookOnCancel() {
+        dispose();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void dispose() {
+        if (isDisposed.getAndSet(true)) {
```
conniey, line 153: We don't need to override dispose. This is done in the basesubscriber. We can hook into `hookOnCancel()`.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,168 @@
...
+            return;
+        }
+
+        currentWork.complete();
+        subscription.cancel();
+        if (timeoutOperation != null && !timeoutOperation.isDisposed() ) {
```
conniey, line 159: nit: format document

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,168 @@
...
+                    completeCurrentWork(currentWork);
+                }
+            });
+    }
+
+
+    @Override
+    protected void hookOnComplete() {
+        logger.info("[{}] Completed. No events to listen to.", currentWork.getId());
+        dispose();
```
conniey, line 130: The BaseSubscriber takes care of disposing for you. This isn't necessary.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -624,23 +627,23 @@ public void close() {
      * entity.
      */
     private void queueWork(int maximumMessageCount, Duration maxWaitTime,
-        FluxSink<ServiceBusReceivedMessageContext> emitter) {
+                                   FluxSink<ServiceBusReceivedMessageContext> emitter) {
```
conniey, line 630: remove indentation.

conniey: image

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/LongLivedMessageSubscriber.java
```diff
@@ -0,0 +1,168 @@
...
+import reactor.core.Disposable;
+import reactor.core.publisher.BaseSubscriber;
+import reactor.core.publisher.Mono;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+ class LongLivedMessageSubscriber extends BaseSubscriber<ServiceBusReceivedMessageContext> {
```
conniey, line 14: SynchronousMessageSubscriber is a better name.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
hemanttanwar: This is the error we get: The variable field 'messageSource ' of class 'implements' should be final. Classes annotated with @ServiceClient are supposed to be immutable. [ServiceClient]

And we can not define it as final, since we need to init it outside of constructor .

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -627,20 +627,20 @@ private void queueWork(int maximumMessageCount, Duration maxWaitTime,
...
-            if (emitterProcessor == null) {
-                emitterProcessor = this.asyncClient.receive(DEFAULT_RECEIVE_OPTIONS)
-                    .subscribeWith(EmitterProcessor.create(asyncClient.getReceiverOptions().getPrefetchCount(), false));
-                messageProcessor.set(emitterProcessor);
+            SynchronousMessageSubscriber messageSubscriber = synchronousMessageSubscriber.get();
+            if (messageSubscriber == null) {
+                messageSubscriber = asyncClient.receive(DEFAULT_RECEIVE_OPTIONS)
+                    .subscribeWith(new SynchronousMessageSubscriber(asyncClient.getReceiverOptions()
+                        .getPrefetchCount(), work));
+                synchronousMessageSubscriber.set(messageSubscriber);
```
conniey, line 638: This isn't thread-safe.
It's possible two different threads see `messageSubscriber == null`, and create the same instance of Subscriber. Then the last one wins and the first one is lost.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,162 @@
...
     private volatile Subscription subscription;
 
-    SynchronousMessageSubscriber(SynchronousReceiveWork work) {
-        this.work = Objects.requireNonNull(work, "'work' cannot be null.");
+
+    private SynchronousReceiveWork currentWork;
+    private Disposable timeoutOperation;
+    private Disposable drainQueueDisposable;
+
+    SynchronousMessageSubscriber(long prefetch, SynchronousReceiveWork initWork) {
```
conniey, line 34: rename: initWork -> initialWork. It makes it more readable.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,162 @@
...
+        logger.verbose("[{}] work completed.", currentWork.getId());
+
+        if (timeoutOperation != null && !timeoutOperation.isDisposed()) {
+            timeoutOperation.dispose();
+        }
+        if (drainQueueDisposable != null && !drainQueueDisposable.isDisposed()) {
+            drainQueueDisposable.dispose();
+        }
+
+        if (wip.decrementAndGet() != 0) {
```
conniey, line 85: There shouldn't be a need for WIP. onNext occurs serially.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,162 @@
...
+    private void completeCurrentWork(SynchronousReceiveWork currentWork) {
+
+        if (isTerminated()) {
+            return;
+        }
+
+        currentWork.complete();
+        logger.verbose("[{}] work completed.", currentWork.getId());
+
+        if (timeoutOperation != null && !timeoutOperation.isDisposed()) {
```
conniey, line 78: This can be inside the hookOnNext method. And return if the currentWork is not terminal to save the nested if.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,162 @@
...
+        if (drainQueueDisposable != null && !drainQueueDisposable.isDisposed()) {
+            drainQueueDisposable.dispose();
+        }
+
+        if (wip.decrementAndGet() != 0) {
+            logger.warning("There is another worker in drainLoop. But there should only be 1 worker.");
+        }
+
+        // After current work finished and there more receive requests
+        if (workQueue.size() > 0) {
```
conniey, line 90: Why do we need to drain? We can see if there is a next one and push it.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousReceiveWork.java
```diff
@@ -108,4 +108,12 @@ void error(Throwable error) {
         this.error = error;
         emitter.error(error);
     }
+
+    /**
+     * Indicate is an this work encountered an error.
+     * @return true if an error occured.
+     */
+    Throwable getError() {
```
conniey, line 116: Why do we need getError? the sink can complete with the error.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,162 @@
...
+    private void drainQueue() {
+        if (isTerminated()) {
+            return;
+        }
+        currentWork = workQueue.poll();
+        if (currentWork == null) {
+            return;
+        }
+
+        subscription.request(currentWork.getNumberOfEvents());
```
conniey, line 124: What if we requested 100 events, only 30 showed up in the timeout period, and the next work item wants 10? This logic will add another 10 credits (making it 80 credits) even though we don't need to add more credits to the link.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
conniey: https://github.com/reactive-streams/reactive-streams-jvm#1.3

hemanttanwar: Thanks for link, It is helpful. But It is not just onNext(),  also various timeout situation  ( we need to drain work queue after timeout ) and if there another  queueWork(work) request during timeout when where is no more work request in queue.  This is a safeguard against many such condition.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,203 @@
...
-     * @param value Event to publish.
+     * @param message Event to publish.
      */
     @Override
-    protected void hookOnNext(ServiceBusReceivedMessageContext value) {
-        work.next(value);
+    protected void hookOnNext(ServiceBusReceivedMessageContext message) {
+        if (currentWork == null) {
+            //Boundary condition(timeout cases), buffer the received message for future requests.
+            bufferMessages.add(message);
```
hemanttanwar, line 65: Example: When there is only one receive(2) request and it time out. But we got two messages after that. During this time we also got request for another   receive(2) . In this case we can use this buffer to send the messages and do not need to ask upstream.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,203 @@
...
+
+        requestCredits(currentWork.getNumberOfEvents() - sentFromBuffer);
+    }
+
+    private Disposable getTimeoutOperation() {
+        return Mono.delay(currentWork.getTimeout())
+            .subscribe(l -> {
+                if (currentWork != null && !currentWork.isTerminal()) {
+                    logger.verbose("[{}] Timeout triggered.", currentWork.getId());
+                    currentWork.complete();
```
conniey, line 168: In the case that this timeout occurs, the completeWork is completed, but it never occurs. So an onNext can happen after this completion. The `completeWork` will not be null; and it'll try to submit another message to that timed out work.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,203 @@
...
+                }
+                if (wip.decrementAndGet() != 0) {
```
conniey, line 170: This isn't how WIP is usually used. WIP is to prevent reentrant onNext/requests which may end up creating a stack overflow.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,220 @@
...
+    private final AtomicLong remaining = new AtomicLong();
 
     private volatile Subscription subscription;
 
-    SynchronousMessageSubscriber(SynchronousReceiveWork work) {
-        this.work = Objects.requireNonNull(work, "'work' cannot be null.");
+    private SynchronousReceiveWork currentWork;
+    private Disposable timeoutOperation;
+
+    SynchronousMessageSubscriber(long prefetch, SynchronousReceiveWork initialWork) {
```
conniey, line 37: Why not queue the initial request? You won't need the `initialWork` as a member variable. And you can use the prefetch.

```java
    SynchronousMessageSubscriber(long prefetch, SynchronousReceiveWork initialWork) {
        this.prefetch = prefetch;
        this.workQueue.add(initialWork)
    }

    @Override
    protected void hookOnSubscribe(Subscription subscription) {
        this.subscription = subscription;
        subscription.request(prefetch);
    }
```

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,220 @@
...
 
-    @Override
-    protected void hookOnComplete() {
-        logger.info("[{}] Completed. No events to listen to.", work.getId());
-        dispose();
+    /**
+     *
+     * @param requested credits for current {@link SynchronousReceiveWork}.
+     */
+    private void requestCredits(long requested) {
```
conniey, line 124: In general, keep the methods with the same visibility together. I organise visibility of methods in descending visibility (ie. public, package-private, private) because consumers will interact with the public/package-private ones first.

As-is, I need to jump around code to reason about what you are trying to do.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,220 @@
...
+        if (workQueue.size() == 0) {
+            return;
+        }
+
+        // If someone is already in this loop, then we are already clearing the queue.
+        if (!wip.compareAndSet(0, 1)) {
+            return;
+        }
+
+        // Drain queue..
```
conniey, line 150: The comment I was trying to make about WIP is to use it **only** for if someone is clearing the queue and to control access to the `drainQueue` method. You also modify this value in `hookOnNext`, but never decrement it when we exit the `drainQueue` method.

```java
    private void drain() {
        // If someone is already in this loop, then we are already clearing the queue.
        if (!wip.compareAndSet(0, 1)) {
            return;
        }

        try {
            drainQueue();
        } finally {
            final int decremented = wip.decrementAndGet();
            if (decremented != 0) {
                logger.warning("There should be 0, but was: {}", decremented);
            }
        }
    }
```

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,220 @@
...
      * Publishes the event to the current {@link SynchronousReceiveWork}. If that work item is complete, will dispose of
      * the subscriber.
-     *
-     * @param value Event to publish.
+     * @param message Event to publish.
      */
     @Override
-    protected void hookOnNext(ServiceBusReceivedMessageContext value) {
-        work.next(value);
+    protected void hookOnNext(ServiceBusReceivedMessageContext message) {
```
conniey, line 82: There's a lot of logic in this method which will make it hard to debug in the future... why wouldn't the following below work? And have any management of currentWork, or emitting the next work in a single place, inside the `drainQueue` method.

```java
    @Override
    protected void hookOnNext(ServiceBusReceivedMessageContext message) {
        bufferMessages.add(message);
        drain();
    }
```

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,220 @@
...
+            }
+        }
+        // timer to complete the current in case of timeout trigger
+        timeoutOperation = getTimeoutOperation(currentWork.getTimeout());
+
+        requestCredits(currentWork.getNumberOfEvents() - sentFromBuffer);
+    }
+
+    private Disposable getTimeoutOperation(Duration timeout) {
+        return Mono.delay(timeout)
```
conniey, line 186: This timeout operation is not tied to a work item, it's possible that between the currentWork and when this times out, the item has changed.

```java
    private Disposable getTimeoutOperation(SynchronousReceiveWork work, Duration timeout) {
        return Mono.delay(timeout).thenReturn(work)
            .subscribe(l -> {
                synchronized (currentWorkLock) {
                    if (currentWork == work) {
                        work.complete();
                    }
                }
            });
    }
```

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,220 @@
...
+            return;
+        }
+
+        // Drain queue..
+        drainQueue();
+
+    }
+
+
+    private void drainQueue() {
```
conniey, line 156: 1. There are 4 places in which you set `currentWork` and these can happen asynchronously. It's possible that the work you're referencing here is one that has been cancelled or timed out. Consider a `currentWorkLock` that synchronises read-write to this variable and the associated timeout operation.
2. Looking at other implementations of subscribers and your current implementation, I would put most of the heavy lifting in this method. (ie. checking to see if we have current work, updating it if need be, setting/requesting more work). As it currently is, there are several ways to emit next items and the currentWork is controlled in many places which will be a pain to debug when this fails.

conniey: You can look at any of the operators for inspiration.
https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/FluxCreate.java#L716

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusReceiverClient.java
```diff
@@ -624,23 +623,26 @@ public void close() {
...
+            if (messageSubscriber == null) {
+                long prefetch = asyncClient.getReceiverOptions().getPrefetchCount();
+                SynchronousMessageSubscriber newSubscriber = new SynchronousMessageSubscriber(prefetch, work);
+
+                if (!synchronousMessageSubscriber.compareAndSet(null, newSubscriber)) {
+                    newSubscriber.dispose();
+                    SynchronousMessageSubscriber existing = synchronousMessageSubscriber.get();
+                    existing.queueWork(work);
+                } else {
+                    asyncClient.receive(DEFAULT_RECEIVE_OPTIONS).subscribeWith(newSubscriber);
```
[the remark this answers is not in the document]
hemanttanwar: When a `newSubscriber` instance is created , we queue the work.. We can not do after `subscribeWith` is called because you may receive a message before you have anything in the queue.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,226 @@
...
+        // Acquiring the lock
+        synchronized (currentWorkLock) {
+
+            // Making sure current work not become terminal since last drain queue cycle
+            if (currentWork != null && currentWork.isTerminal()) {
+                workQueue.remove(currentWork);
+                if (currentTimeoutOperation != null & !currentTimeoutOperation.isDisposed()) {
+                    currentTimeoutOperation.dispose();
+                }
+                currentTimeoutOperation = null;
```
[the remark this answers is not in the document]
hemanttanwar: We might pick currentWork more than one time from workQueue. `currentTimeoutOperation ==  null` will indicate that are we picking up first time.  

We do not need to process  currentWork if is picked up second time and no bufferMessages to send to it.

`while ((currentWork = workQueue.peek()) != null && (currentTimeoutOperation == null || bufferMessages.size() > 0 )) {`

The time Operation is not removing currentWork from he queue, thus currentWork needs be picked up again, thus we need to threat this different and go in the loop and remove this timeout work. setting `currentTimeoutOperation = null` indicate this case also.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/SynchronousMessageSubscriber.java
```diff
@@ -5,114 +5,239 @@
```
conniey, line 5: Only thing you need are unit tests. The logic in this subscriber is complex and I can see it being hard to debug.

conniey:
I looked through the unit test and was hoping that they would test the complex logic in hookOnNext. There are various branches and failure modes that aren't exercised in the unit tests. This may come back to bite us if we don't thoroughly test it.

PR #10941, opened 2020-05-08
Spring Boot migration -- Azure AD B2C

yiliuTo:
Migrate Azure AD B2C from Azure Spring Boot project, modules include:

- azure-spring-boot-starter-active-directory-b2c

and add samples and autoconfigure code for aad b2c.


yiliuTo:
As we have discussed before, we'd like to migrate to this repo but also need to maintain the group id of track one for a period. Thus to keep the consistency as much as possible, could we move the related module statements from eng/versioning/version_data.txt to version_client.txt?

JonathanGiles:
I totally accept the position that Jim has. We all want repo purity and clear delineation between track one and track two. My understanding is that these libraries are midway through a transition from track one to track two. My guidelines are just now getting towards their final shape, and I believe there is consensus forming. For the sake of pragmatism and forward progress, my suggestion would be, barring any major technical difficulties on the engsys side, that we allow this PR to be merged as-is, and allow for it to continue its transition into the expected track two shape once it has become a member of this repository.

My expectation is that this will be a short-term intermediate state, and knowing the folks working on this, I have no doubts it will be rapidly moved into the state desired by us all.

We should all reconvene at a point further along the development process to do a top-to-bottom review of the spring integration modules for track two. I totally understand that is more difficult after a PR is merged rather than before, but I don't want to introduce a road block at this stage.

Barring any comments from others, I am, by default, ok with continuing to get this code into our repo and wired into our engineering system.

mitchdenny:
I'm inclined to agree with @JonathanGiles. We can tolerate some variation under the following conditions:

1. We all know it exists, and know that sometimes it'll bite us :)
2. We've got a plan to converge.
3. The variation doesn't leak across the rest of the ES.

Let's see if we can articulate some of the differences we are seeing and just quickly think through the impact that they are going to have in the release pipeline (which is ultimately where the rubber hits the road):

1. ```com.microsoft.azure``` group ID; I don't think that this is a problem for the release pipeline itself, we use the same release pipeline for track 1 and track 2, but there could be some version/POM enforcement scripts that need figuring out ... @JimSuplizio?

2. JavaDoc artifact missing; just checked the artifact, when you folks build this locally is it producing the JavaDoc resources? If not that needs to be fixed as Maven Central won't actually accept packages without the JavaDocs.

Those are the only two things I'm seeing. Jim do you have more?

saragluna:
@mitchdenny  We do have JavaDoc artifact generated, but only for the azure-spring-boot module since all source codes are put in that one, and the starter module only contains a pom file.

JimSuplizio:
/CC @weshaggard 

I can see the point that @JonathanGiles is making however there are consequences with using the azure sdk client as a parent, namely everything in track 2 that isn't in track 1. This hybrid approach is taking the spring libraries that are track 1 and forcing them into track 2. There are a number of things in track 2 that are track 2 specific which includes, but is not limited to, aggregate reports, coverage, dependency graph (which does not look at track 1 libraries), maven-enforcer-plugin rules and From Source runs. Further, track 2 libraries must have a Jacoco entry otherwise the exclusion required by the amd switch in the From Source runs won't work but even then, it's effectively a useless run since only track 2 versions get manipulated. The fact that the Spring libraries are track 1 with track 2 dependencies means that the track 2 dependencies needed by Spring are going to have to be declared in the external dependencies (current/dependency tags only apply to items in the same track, everything else is an external_dependency).

Part of the other problem in doing things half way, like this, is going to be support. If there needs to be a track 1 spring release prior to the track 2 release then they should be completely separate directories for support reasons. @joshfree can probably answer how long these released libraries need to be supported but I'm going to say that, historically, releasing a patch out of source that's currently building is a lot easier than branching off a SHA that's several months old. 

In all seriousness, the changes to make this track 1 would be:
1. In the various project's pom.xml files, changing the parent entry from azure-client-sdk-parent to azure-data-sdk-parent and then removal of any coverage overrides and maven-enforcer-plugin entries. 
2. The spring directory's ci.yml file would need to use the archetype-sdk-data.yml instead of archetype-sdk-client.yml. 
3. Similarly, the directories should be named accordingly (azure-spring-boot would be microsoft-azure-spring-boot) if this were going to happen.

Ultimately, all I can do is make the recommendation, the final decision is up to our architect @JonathanGiles and the java team lead @joshfree.

jialindai:
Hi @JimSuplizio , I am not quite clear about below engineer system items:

- aggregate reports
- coverage
- dependency graph (which does not look at track 1 libraries)
- maven-enforcer-plugin rules 
- From Source runs.

Do you have some reference on how these things work? And if spring modules are included, how these parts are affected?

JimSuplizio:
@jialindai I misread part of the original reply. These are keeping the track 1 groupId and using the track 2 infrastructure and if that's the case where's here is mostly correct.

Moving the library definitions from version_data.txt to version_client.txt was the correct more (and is in this PR) however every spring library that is being added should probably have an entry in eng/jacoco-test-coverage/pom.xml, not just com.microsoft.azure:azure-spring-boot. Depending on the current coverage numbers you may need to override the min line and branch coverage. An example of that is [here ](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/cosmos/azure-cosmos-examples/pom.xml#L34)if needed.

Adding the libraries with the client parent will automatically get them into the dependency graph, added to aggregate reports and everything else falls into place.

eng/jacoco-test-coverage/pom.xml
```diff
@@ -187,6 +187,11 @@
       <artifactId>azure-sdk-template</artifactId>
       <version>1.0.4-beta.13</version> <!-- {x-version-update;com.azure:azure-sdk-template;current} -->
     </dependency>
+    <dependency>
```
JimSuplizio, line 190: Every library being added should have an entry in here.

JimSuplizio:
> @mitchdenny We do have JavaDoc artifact generated, but only for the azure-spring-boot module since all source codes are put in that one, and the starter module only contains a pom file.
As @mitchdenny commented, you're going to want to have the javadocs being produced for every library. Maven wants the javadoc artifact in order to publish and we also use that artifact to populate both the githubio docs and the Microsoft docs.

mitchdenny:
Our pipeline automatically determines the way to publish based on the packages present. It looks like the starts do in fact have some files in them because JAR files are produced:

image

Our pipeline would treat this is a non-pom-only package. Maven would potentially reject it as a result as well.

saragluna:
@mitchdenny, @JimSuplizio 

We did publish the stater jars without JavaDoc artifacts, please check here https://repo.maven.apache.org/maven2/com/microsoft/azure/azure-active-directory-spring-boot-starter/2.2.3/, does this mean it's okay for these starter jars to produce no JavaDoc artifacts? 

@JimSuplizio 
If there're no source codes in the starter module, will the jacoco plugin skip such modules? If yes, do we still need to define the code coverage number in each module?

mitchdenny:
OK … that’s interesting, previously SonaType has rejected when I’ve forgotten to upload javadocs. I guess we give it a shot.

sdk/spring/azure-spring-boot-starter-active-directory-b2c/src/main/resources/aadb2c.enable.config
```diff
@@ -0,0 +1 @@
+dummy
```
JimSuplizio, line 1: Does this really need to be in here? I'm not going to block approval on the PR but I am curious.

saragluna: Yes, we'll auto-configure the AAD B2C starter when this file is on the classpath.

saragluna:
@JimSuplizio I think we'll also keep that artifact now, and change it together with the group id.

PR #10961, opened 2020-05-08
Update scan_for_unreleased_dependencies.ps1

JimSuplizio:
There was issue this week where a library was allowed to ship GA with beta dependencies. The fix here is to ensure that any dependency that not an external_dependency (aka, something we build/ship) has its version checked. If the version is beta and the library being shipped is GA, then we fail. The output would look like this:
inputGroupId=com.azure
inputArtifactId=azure-core-amqp
serviceDirectory=C:\src\azure-sdk-for-java\sdk\core
Found pom file with matching groupId(com.azure)/artifactId(azure-core-amqp), pomFile=C:\src\azure-sdk-for-java\sdk\core\azure-core-amqp\pom.xml
Library is not releasing as Beta, version=1.1.1
Error: Cannot release non-beta libraries with beta dependencies. dependency={x-version-update;com.azure:azure-core;current}, version=1.6.0-beta.1

### DIFF
--- a/eng/versioning/scan_for_unreleased_dependencies.ps1
+++ b/eng/versioning/scan_for_unreleased_dependencies.ps1
@@ -87,6 +87,21 @@ Get-ChildItem -Path $serviceDirectory -Filter pom*.xml -Recurse -File | ForEach-
                         Write-Error-With-Color "Error: Cannot release non-beta libraries with beta_ dependencies. dependency=$($versionUpdateTag)"
                     }
                     continue
+                } else {
+                    # If this is an external dependency then continue
+                    if ($versionUpdateTag -match "external_dependency}") {
+                        continue
+                    }
+                    # If this isn't an external dependency then ensure that if the dependency
+                    # version is beta, that we're releasing a beta, otherwise fail
+                    if ($versionNode.InnerText -like '*-beta.*') 
+                    {
+                        if (!$libraryIsBeta)
+                        {
+                            $script:FoundError = $true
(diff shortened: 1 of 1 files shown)


eng/versioning/scan_for_unreleased_dependencies.ps1
```diff
@@ -87,6 +87,21 @@ Get-ChildItem -Path $serviceDirectory -Filter pom*.xml -Recurse -File | ForEach-
                         Write-Error-With-Color "Error: Cannot release non-beta libraries with beta_ dependencies. dependency=$($versionUpdateTag)"
                     }
                     continue
+                } else {
+                    # If this is an external dependency then continue
+                    if ($versionUpdateTag -match "external_dependency}") {
```
weshaggard, line 92: I think it is actually a good idea to do a check for both internal and external dependencies. I'd suggest we use the SemVer.ps1 script we have under common and check to see if our version and if it is not prerelease then we should error for any dependency that is marked as a prerelease.

JimSuplizio: @weshaggard, sure. I can look and see if this is feasible for external_dependencies on Monday. There are, unfortunately, beta external dependencies. Also, I'd have to verify that the SemVer.ps1 does correctly parse all of the external dependency versions first as Maven is a bit more liberal with what it allows for versions.

weshaggard: I hope we don't depend on external prereleases in our GA packages that will definitely be a recipe for potential disaster.

JimSuplizio: We do but only in a <test> scope. I'm less concerned about external dependencies since the maven-enforcer-plugin was added and every dependency is now being scrutinized (track 2 only though).

PR #10970, opened 2020-05-08
Added Connection Config APIs for Cosmos Client Builder

kushagraThapar:
This PR : 
* Adds new Direct and Gateway Connection Config APIs for Cosmos Client Builder
`directMode(DirectConnectionConfig config)`
`gatewayMode(GatewayConnectionConfig config)`
* Moves Connection Policy to implementation package
* Adds new test for Connection Config - both Direct Connection Config and Gateway Connection Config
* Updated existing tests to use new connection configs

Few things - which still need to be done : https://github.com/Azure/azure-sdk-for-java/issues/10983

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
@@ -6,14 +6,14 @@ package com.azure.cosmos;
 import java.net.InetSocketAddress;
 import java.time.Duration;
 
-import static com.azure.cosmos.implementation.ConnectionPolicy.DEFAULT_IDLE_CONNECTION_TIMEOUT;
-import static com.azure.cosmos.implementation.ConnectionPolicy.DEFAULT_MAX_POOL_SIZE;
-import static com.azure.cosmos.implementation.ConnectionPolicy.DEFAULT_REQUEST_TIMEOUT;
-
 /**
  * Represents the connection config with {@link ConnectionMode#GATEWAY} associated with Cosmos Client in the Azure Cosmos DB database service.
  */
 public final class GatewayConnectionConfig {
+    //  Constants
+    private static final Duration DEFAULT_REQUEST_TIMEOUT = Duration.ofSeconds(60);
+    private static final Duration DEFAULT_IDLE_CONNECTION_TIMEOUT = Duration.ofSeconds(60);
+    private static final int DEFAULT_MAX_POOL_SIZE = 1000;
 
     private Duration requestTimeout;
     private int maxConnectionPoolSize;


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -327,6 +317,46 @@ public CosmosClientBuilder contentResponseOnWriteEnabled(boolean contentResponse
...
+    }
+
+    /**
+     * Gets the GATEWAY connection configuration to be used.
+     *
+     * @return gateway connection config
+     */
+    public GatewayConnectionConfig getGatewayConnectionConfig() {
+        return gatewayConnectionConfig;
+    }
```
moderakh, line 349: we are not exposing public getters for other CosmosClientBuilder apis

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -327,6 +317,46 @@ public CosmosClientBuilder contentResponseOnWriteEnabled(boolean contentResponse
...
+
+    /**
+     * Gets the DIRECT connection configuration to be used.
+     *
+     * @return direct connection config
+     */
+    public DirectConnectionConfig getDirectConnectionConfig() {
+        return directConnectionConfig;
+    }
```
moderakh, line 358: ditto

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
```diff
@@ -0,0 +1,269 @@
...
+     * default values for configuring the retry policies.  See RetryOptions class for
+     * more details.
+     *
+     * @param throttlingRetryOptions the RetryOptions instance.
+     * @return the {@link DirectConnectionConfig}.
+     * @throws IllegalArgumentException thrown if an error occurs
+     */
+    @Override
+    public DirectConnectionConfig setThrottlingRetryOptions(ThrottlingRetryOptions throttlingRetryOptions) {
+        super.setThrottlingRetryOptions(throttlingRetryOptions);
```
moderakh, line 154: why do we need to overwrite these ones? this is common for both gw and direct.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
```diff
@@ -0,0 +1,259 @@
...
+     * default values for configuring the retry policies.  See RetryOptions class for
+     * more details.
+     *
+     * @param throttlingRetryOptions the RetryOptions instance.
+     * @return the {@link GatewayConnectionConfig}.
+     * @throws IllegalArgumentException thrown if an error occurs
+     */
+    @Override
+    public GatewayConnectionConfig setThrottlingRetryOptions(ThrottlingRetryOptions throttlingRetryOptions) {
+        super.setThrottlingRetryOptions(throttlingRetryOptions);
```
moderakh, line 123: why do we need to override the common apis for both gw and direct?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
kushagraThapar: Good catch @moderakh  - will get rid of them.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
kushagraThapar: Good catch @moderakh - will get rid of them.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
kushagraThapar: Yes, they are common, and they both call their parent's class APIs.. 
The reason for overriding these ones is because of the return type. 
Since our APIs are fluent type, if you directly use parent's API (which returns `ConnectionConfig`), end users will not be able to use any APIs of `DirectConnectionConfig` or `GatewayConnectionConfig` types..

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
kushagraThapar: Yes, they are common, and they both call their parent's class APIs..
The reason for overriding these ones is because of the return type.
Since our APIs are fluent type, if you directly use parent's API (which returns ConnectionConfig), end users will not be able to use any APIs of DirectConnectionConfig or GatewayConnectionConfig types..

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ConnectionConfig.java
```diff
@@ -0,0 +1,223 @@
...
+        this.readRequestsFallbackEnabled = null;
+        this.userAgentSuffix = "";
+    }
+
+    /**
+     * Gets the connection mode used in the client.
+     *
+     * @return the connection mode.
+     */
+    public ConnectionMode getConnectionMode() {
```
kirankumarkolli, line 39: One option is to inline these into the builder.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ConnectionConfig.java
```diff
@@ -0,0 +1,223 @@
...
+     * are ordered by PreferredRegions property. Setting the property value
+     * to true has no effect until EnableMultipleWriteRegions in DatabaseAccount
+     * is also set to true.
+     * <p>
+     * DEFAULT value is true indicating that writes are directed to
+     * available writable regions of geo-replicated database account.
+     *
+     * @return flag to enable writes on any regions for geo-replicated database accounts.
+     */
+    public boolean isUsingMultipleWriteRegions() {
```
kirankumarkolli, line 85:
```suggestion
    public boolean isMultipleWriteRegionsEnabled() {
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -77,20 +80,18 @@ boolean isSessionCapturingOverrideEnabled() {
      *
      * <pre>
      * {@code
-     * ConnectionPolicy connectionPolicy = new ConnectionPolicy();
-     * getConnectionPolicy.getConnectionMode(ConnectionMode.DIRECT);
      * CosmosAsyncClient client1 = new CosmosClientBuilder()
      *         .endpoint(serviceEndpoint1)
      *         .key(key1)
-     *         .connectionPolicy(connectionPolicy)
+     *         .connectionModeDirect(DirectConnectionConfig.getDefaultConfig())
```
kirankumarkolli, line 86: directMode & gatewayMode

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
```diff
@@ -0,0 +1,269 @@
...
+    public DirectConnectionConfig setConnectionTimeout(Duration connectionTimeout) {
+        this.connectionTimeout = connectionTimeout;
+        return this;
+    }
+
+    /**
+     * Gets the idle channel timeout
+     * @return idle channel timeout
+     */
+    public Duration getIdleChannelTimeout() {
```
kirankumarkolli, line 72: connection vs channel

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
```diff
@@ -0,0 +1,269 @@
...
+
+import java.time.Duration;
+import java.util.List;
+
+/**
+ * Represents the connection config with {@link ConnectionMode#DIRECT} associated with Cosmos Client in the Azure Cosmos DB database service.
+ */
+public final class DirectConnectionConfig extends ConnectionConfig {
+    //  Constants
+    private static final Duration DEFAULT_IDLE_CHANNEL_TIMEOUT = Duration.ZERO;
```
kirankumarkolli, line 14: Will ZERO default to FX one. If so we might not have control on it. Why not be explicit?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
```diff
@@ -0,0 +1,269 @@
...
+     * <p>
+     * DEFAULT value is false indicating that writes are only directed to
+     * first region in PreferredRegions property.
+     *
+     * @param usingMultipleWriteRegions flag to enable writes on any regions for geo-replicated
+     * database accounts.
+     * @return the {@link DirectConnectionConfig}.
+     */
+    @Override
+    public DirectConnectionConfig setUsingMultipleWriteRegions(boolean usingMultipleWriteRegions) {
```
kirankumarkolli, line 194: Enable suffix

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
kushagraThapar: This section is no longer valid.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ConnectionConfig.java
kushagraThapar: Done.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/ConnectionConfig.java
kushagraThapar: Done.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
kushagraThapar: Done.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
kushagraThapar: Done.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
kushagraThapar: https://github.com/Azure/azure-sdk-for-java/issues/10983

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
kushagraThapar: https://github.com/Azure/azure-sdk-for-java/issues/10983

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/UniqueIndexTest.java
```diff
@@ -222,7 +222,7 @@ public void before_UniqueIndexTest() {
         client = new CosmosClientBuilder()
             .endpoint(TestConfigurations.HOST)
             .key(TestConfigurations.MASTER_KEY)
-            .connectionPolicy(ConnectionPolicy.getDefaultPolicy())
+            .directMode(DirectConnectionConfig.getDefaultConfig())
```
milismsft, line 225: Do these tests are always expected to run in Direct mode?

kushagraThapar: Yes, I have not changed the connection mode of the tests - they will run as it is as they were running earlier. 
Since earlier also, default was DIRECT mode.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.cosmos;
+
+import java.net.InetSocketAddress;
+import java.time.Duration;
+
+import static com.azure.cosmos.implementation.ConnectionPolicy.DEFAULT_IDLE_CONNECTION_TIMEOUT;
```
milismsft, line 9: you're consuming implementation package here and the class is public facing...

milismsft: just to be clear, this can be done after GA :-)

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -327,6 +327,223 @@ public CosmosClientBuilder contentResponseOnWriteEnabled(boolean contentResponse
...
+     * When EnableEndpointDiscovery is true, the SDK will automatically discover the
+     * current write and read regions to ensure requests are sent to the correct region
+     * based on the capability of the region and the user's preference.
+     * <p>
+     * The default value for this property is true indicating endpoint discovery is enabled.
+     *
+     * @param endpointDiscoveryEnabled true if EndpointDiscovery is enabled.
+     * @return current CosmosClientBuilder
+     */
+    public CosmosClientBuilder endpointDiscoverEnabled(boolean endpointDiscoveryEnabled) {
```
milismsft, line 412: endpointDiscoveryEnabled

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
```diff
@@ -0,0 +1,148 @@
...
+// Licensed under the MIT License.
+
+package com.azure.cosmos;
+
+import java.time.Duration;
+
+/**
+ * Represents the connection config with {@link ConnectionMode#DIRECT} associated with Cosmos Client in the Azure Cosmos DB database service.
+ */
+public final class DirectConnectionConfig {
```
moderakh, line 11: in Direct mode we still establish some connections to GW (master resource operations and query plan retrieval, etc). 
How are we controlling the GW Connection specifics in Direct mode?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
moderakh: why is this a problem?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosBridgeInternal.java
```diff
@@ -59,13 +60,20 @@ public static CosmosClientBuilder cloneCosmosClientBuilder(CosmosClientBuilder b
...
             .keyCredential(builder.getKeyCredential())
             .permissions(builder.getPermissions())
             .authorizationTokenResolver(builder.getAuthorizationTokenResolver())
             .resourceToken(builder.getResourceToken())
-            .contentResponseOnWriteEnabled(builder.isContentResponseOnWriteEnabled());
+            .contentResponseOnWriteEnabled(builder.isContentResponseOnWriteEnabled())
+            .userAgentSuffix(builder.getUserAgentSuffix())
+            .throttlingRetryOptions(builder.getThrottlingRetryOptions())
+            .preferredRegions(builder.getPreferredRegions())
+            .endpointDiscoverEnabled(builder.isEndpointDiscoveryEnabled())
```
moderakh, line 74:
```suggestion
endpointDiscoveryEnabled
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
kushagraThapar: Good catch @milismsft - You saved us V5 🤣

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
kushagraThapar: @milismsft 's point is - Since we are using default values from implementation details - changing these default values in implementation will have a direct impact on customer's app, which can be considered breaking change in some way. 
I think its a valid point, I will think what can be done for this.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -327,6 +327,223 @@ public CosmosClientBuilder contentResponseOnWriteEnabled(boolean contentResponse
...
+     * If this property is not set, the default is true for all Consistency Levels other than Bounded Staleness,
+     * The default is false for Bounded Staleness.
+     * 1. {@link #endpointDiscoveryEnabled} is true
+     * 2. the Azure Cosmos DB account has more than one region
+     *
+     * @param readRequestsFallbackEnabled flag to enable reads to go to multiple regions configured on an account of
+     * Azure Cosmos DB service.
+     * @return current CosmosClientBuilder
+     */
+    public CosmosClientBuilder readRequestsFallbackEnabled(Boolean readRequestsFallbackEnabled) {
```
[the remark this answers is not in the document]
kushagraThapar: I checked, we can convert it to primitive `boolean` with default value = true, based on this below check that we have in `GlobalAddressResolver`

```
int maxBackupReadEndpoints = (connectionPolicy.isReadRequestsFallbackEnabled() == null || connectionPolicy.isReadRequestsFallbackEnabled()) ? GlobalAddressResolver.MaxBackupReadRegions : 0;
```

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
```diff
@@ -0,0 +1,148 @@
...
+    public DirectConnectionConfig setIdleEndpointTimeout(Duration idleEndpointTimeout) {
+        this.idleEndpointTimeout = idleEndpointTimeout;
+        return this;
+    }
+
+    /**
+     * Gets the max channels per endpoint
+     * @return the max channels per endpoint
+     */
+    public int getMaxChannelsPerEndpoint() {
```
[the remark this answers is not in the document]
kushagraThapar: @David-Noble-at-work plans to fix these.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
```diff
@@ -0,0 +1,138 @@
...
+        this.requestTimeout = requestTimeout;
+        return this;
+    }
+
+    /**
+     * Gets the value of the connection pool size the client is using.
+     *
+     * @return connection pool size.
+     */
+    public int getMaxPoolSize() {
```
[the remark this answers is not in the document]
kushagraThapar: Sounds good, will do.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
```diff
@@ -0,0 +1,138 @@
...
+        this.requestTimeout = DEFAULT_REQUEST_TIMEOUT;
+    }
+
+    /**
+     * Gets the default Gateway connection configuration.
+     *
+     * @return the default gateway connection configuration.
+     */
+    public static GatewayConnectionConfig getDefaultConfig() {
+        return GatewayConnectionConfig.defaultConfig;
```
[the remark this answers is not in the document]
kushagraThapar: Yes, makes sense.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
```diff
@@ -0,0 +1,148 @@
...
+        this.maxRequestsPerChannel = DEFAULT_MAX_REQUESTS_PER_ENDPOINT;
+    }
+
+    /**
+     * Gets the default DIRECT connection configuration.
+     *
+     * @return the default direct connection configuration.
+     */
+    public static DirectConnectionConfig getDefaultConfig() {
+        return DirectConnectionConfig.defaultConfig;
```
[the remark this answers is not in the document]
kushagraThapar: Yes, makes sense.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosBridgeInternal.java
kushagraThapar: Done.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/DirectConnectionConfig.java
kushagraThapar: Good point, we are only allowing for default values configurations for `Gateway` connection properties in case of `Direct` connection mode. 
Something to think about, I will give it a thought.

kushagraThapar: @moderakh - I have exposed these gateway client properties to `CosmosClientBuilder`
So that end users can also set these even in case of direct connection mode. 
```
private Duration requestTimeoutGateway = DEFAULT_REQUEST_TIMEOUT;
private Duration idleConnectionTimeoutGateway = DEFAULT_IDLE_CONNECTION_TIMEOUT;
private int maxConnectionPoolSizeGateway = DEFAULT_MAX_POOL_SIZE;
```

kushagraThapar:
As @moderakh  pointed out, even in Direct Connection mode, end user may still need to set some of the Gateway client properties. 
For the above purpose, I have exposed these gateway client properties to `CosmosClientBuilder` so that end users can also set these even in case of direct connection mode.

```
private Duration requestTimeoutGateway;
private Duration idleConnectionTimeoutGateway;
private int maxConnectionPoolSizeGateway;
```

@moderakh  @kirankumarkolli - please check.

kushagraThapar:
> As @moderakh pointed out, even in Direct Connection mode, end user may still need to set some of the Gateway client properties.
> For the above purpose, I have exposed these gateway client properties to `CosmosClientBuilder` so that end users can also set these even in case of direct connection mode.
> 
> ```
> private Duration requestTimeoutGateway;
> private Duration idleConnectionTimeoutGateway;
> private int maxConnectionPoolSizeGateway;
> ```
> 
> @moderakh @kirankumarkolli - please check.

This has been fixed.

kushagraThapar:
> This PR exposes the notion of channel in the public api
> e.g.,`DirectConnectionConfig:setMaxChannelsPerEndpoint`
> Please check if we can use Connection instead of Channel. Less concepts better.
> 
> LGTM

Yes, that is on the list and will get in before GA.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
David-Noble-at-work: The `RntbdTransportClient` will treat the `DirectConnectionConfig` as definitive. Regardless of where we put the default values, changes to them will affect behavior, and changes to behavior must be rationalized and documented. This sounds like a good reason to move the public-facing defaults to the `ConnectionConfig` classes. It will make it easier for us to understand the impact of a change. The transport clients could then pull defaults from their respective public config classes; not the other way around.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientBuilder.java
```diff
@@ -327,6 +327,262 @@ public CosmosClientBuilder contentResponseOnWriteEnabled(boolean contentResponse
...
+        return this;
+    }
+
+    /**
+     * Sets the default DIRECT connection configuration to be used.
+     *
+     * @return current CosmosClientBuilder
+     */
+    public CosmosClientBuilder directMode() {
+        this.directConnectionConfig = DirectConnectionConfig.getDefaultConfig();
```
milismsft, line 357: does this.gatewayConectionConfig get initialized somewhere else if only "directMode() is called or is it null? also for direct mode we need a different set of default values for it, not the default ones, right?

kushagraThapar: It will be `null` - And the connection policy gets built in `buildConnectionPolicy` method - with default values for gateway configuration - with the below defaults - in `ConnectionPolicy`
```
//  Gateway connection config properties
    private int maxConnectionPoolSize = DEFAULT_MAX_POOL_SIZE;
    private Duration requestTimeout = DEFAULT_REQUEST_TIMEOUT;
    private Duration idleConnectionTimeout = DEFAULT_IDLE_CONNECTION_TIMEOUT;
```

kushagraThapar: Yes, for `DirectConnectionConfig` - we have separate default values - which I got from `RntbdTransportClient`

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/GatewayConnectionConfig.java
kushagraThapar: Done.

kushagraThapar: Updated constants to GatewayConnectionConfig.java

PR #10971, opened 2020-05-08
Update org.springframework:spring-web for CVE issue

JimSuplizio:
Updating org.springframework:spring-web to 5.2.5.RELEASE to address a CVE issue.


kurtzeborn:
Yep, this is the fix needed to address the warnings we're seeing.  LGTM.

PR #10977, opened 2020-05-08
README: Changing identity version to latest and adding some more explanation

hemanttanwar:
- Updating identity version in README
 - Adding some more key concepts

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/README.md
+++ b/sdk/servicebus/azure-messaging-servicebus/README.md
@@ -237,8 +237,8 @@ There are four ways of settling messages:
   - Abandon - releases the receiver's lock on the message allowing for the message to be received by other receivers.
   - Defer - defers the message from being received by normal means. In order to receive deferred messages, the sequence 
 number of the message needs to be retained.
-  - DeadLetter - moves the message to the Dead Letter queue. This will prevent the message from being received again. 
-In order to receive messages from the Dead Letter queue, a receiver scoped to the Dead Letter queue is needed.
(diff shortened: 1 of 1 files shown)


sdk/servicebus/azure-messaging-servicebus/README.md
```diff
@@ -142,6 +142,25 @@ on which actual message transmission takes place. The namespace often serves as
 * A **[subscription][subscription_concept]** receives messages from a topic. Each subscription is independent and
   receives a copy of the message sent to the topic.
 
+To interact with these resources, one should be familiar with the following SDK concepts:
+
+* All the Service Bus Clients are created using `ServiceBusClientBuilder`.
```
conniey: These are extraneous and implied from the "Authenticate the client" and samples. These are usually Key Concepts about using the service itself. (ie. What is a queue? What is a topic?)

sdk/servicebus/azure-messaging-servicebus/README.md
```diff
@@ -232,6 +232,14 @@ receiver.receive(10).forEach(context -> {
     receiver.complete(message);
 });
 ```
+There are four ways of settling messages:
+  - Complete - causes the message to be deleted from the queue or topic.
+  - Abandon - releases the receiver's lock on the message allowing for the message to be received by other receivers.
+  - Defer - defers the message from being received by normal means. In order to receive deferred messages, the sequence 
+number of the message needs to be retained.
+  - DeadLetter - moves the message to the Dead Letter queue. This will prevent the message from being received again.
```
conniey, line 240: Dead-letter

conniey:
```
Moves the message to the [dead-letter queue](https://docs.microsoft.com/azure/service-bus-messaging/service-bus-dead-letter-queues).
```

You can let them read the doc to understand more.

PR #10982, opened 2020-05-08
Update Storage ci.yml Additional Modules

alzimmermsft:
Due to the nature of the `-amd` switch that is used during the `From Source` builds `azure-messaging-eventhubs-checkpoint-store` fails to compile due to it, rightfully, being unable to find an unreleased version of `azure-messaging-eventhubs`. The change in this PR adds `azure-messaging-eventhubs` as an additional module that will be used during compile and test run time, this module won't show up as an option in the release pipeline. Adding this will allow the `From Source` builds to properly compile, allowing us to verify the future state when in development modules are released and dependency versions are updated. Additionally this should unblock PR #10979.

This issue was being masked, and would have been noticed earlier, but the `azure-messaging-eventhubs` and `-checkpoint-store` in development versions were incorrectly listed as `5.1.0-beta.1` and `1.1.0-beta.1` respectively. These versions were released to Maven last month for a preview build EventHubs had and should have been updated to `beta.2` but this was forgotten, additionally this didn't have a update versions PR to master as this was released from a branch. The [release process wiki](https://github.com/Azure/azure-sdk-for-java/wiki/Release-process) will be updated in the next few days to outline what should happen when a beta or hotfix is released from a branch so this scenario doesn't happen again.

@srnagar @mitchdenny @JimSuplizio @rickle-msft

### DIFF
--- a/sdk/storage/tests.yml
+++ b/sdk/storage/tests.yml
@@ -29,9 +29,6 @@ jobs:
         - name: azure-storage-queue
           groupId: com.azure
           safeName: azurestoragequeue
-      AdditionalModules:
-        - name: azure-messaging-eventhubs
-          groupId: com.azure
       TimeoutInMinutes: 120
       EnvVars:
         AZURE_TEST_MODE: LIVE


mitchdenny:
Good find!

PR #11031, opened 2020-05-11
Spring Boot migration -- JMS for Service Bus and Metrics

saragluna:
Migrate JMS for Service Bus and metrics starter from Azure Spring Boot project, modules include:

- azure-spring-boot-starter-service-bus-jms
- azure-spring-boot-starter-metrics

### DIFF
--- a/sdk/spring/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAppRoleStatelessAuthenticationFilter.java
+++ b/sdk/spring/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAppRoleStatelessAuthenticationFilter.java
@@ -54,27 +54,8 @@ public class AADAppRoleStatelessAuthenticationFilter extends OncePerRequestFilte
         final String authHeader = request.getHeader(HttpHeaders.AUTHORIZATION);
         boolean cleanupRequired = false;
 
-        if (hasText(authHeader) && authHeader.startsWith(TOKEN_TYPE)) {
-            try {
-                final String token = authHeader.replace(TOKEN_TYPE, "");
-                final UserPrincipal principal = principalManager.buildUserPrincipal(token);
-                final JSONArray roles = Optional.ofNullable((JSONArray) principal.getClaims().get("roles"))
-                    .filter(r -> !r.isEmpty())
-                    .orElse(DEFAULT_ROLE_CLAIM);
-                final Authentication authentication = new PreAuthenticatedAuthenticationToken(
-                    principal, null, rolesToGrantedAuthorities(roles));
-                authentication.setAuthenticated(true);
-                LOGGER.info("Request token verification success. {}", authentication);
-                SecurityContextHolder.getContext().setAuthentication(authentication);
-                cleanupRequired = true;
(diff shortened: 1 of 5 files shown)


eng/versioning/external_dependencies.txt
```diff
@@ -70,6 +71,7 @@ org.springframework.security:spring-security-oauth2-client;5.2.0.RELEASE
 org.springframework.security:spring-security-oauth2-core;5.2.0.RELEASE
 org.springframework.security:spring-security-oauth2-jose;5.2.0.RELEASE
 org.springframework:spring-web;5.2.5.RELEASE
+org.springframework:spring-jms;5.2.0.RELEASE
```
JimSuplizio, line 74: This should be 5.2.5.RELEASE to match the org.springframework:spring-web's version. Both libraries have spring-beans and spring-core as immediate dependencies and unless these are the same version you risk introducing a dependency ordering issue.

eng/versioning/external_dependencies.txt
```diff
@@ -227,3 +229,7 @@ media_com.microsoft.azure:adal4j;1.2.0
 
 # sdk\storage\azure-storage-blob-cryptography\pom.xml
 storage_com.microsoft.azure:azure-storage;8.4.0
+
+# sdk\spring\azure-spring-boot\pom.xml
+spring_io.micrometer:micrometer-core;1.3.0
```
JimSuplizio, line 234: /CC @kushagraThapar 
The only library that uses io.micrometer:micrometer-core and io.micrometer:micrometer-registry-azure-monitor dependencies is Cosmos. Could cosmos also be updated to use the 1.3.0 versions? I would prefer that, whenever possible, that we update the common version of an external dependency rather than declare library specific ones.

eng/versioning/version_data.txt
```diff
@@ -43,3 +43,5 @@ com.microsoft.azure:azure-media;1.0.0-beta.1;1.0.0-beta.1
 com.microsoft.azure:azure-spring-boot;2.2.4;2.2.5-beta.1
 com.microsoft.azure:azure-spring-boot-starter;2.2.4;2.2.5-beta.1
 com.microsoft.azure:azure-active-directory-spring-boot-starter;2.2.4;2.2.5-beta.1
+com.microsoft.azure:azure-servicebus-jms-spring-boot-starter;2.2.4;2.2.5-beta.1
```
JimSuplizio, line 46: Shouldn't these be being added to version_client.txt instead of version_data.txt like the other spring libraries are being added?

eng/versioning/external_dependencies.txt
saragluna: upgraded to 5.2.5.RELEASE

eng/versioning/version_data.txt
saragluna: Moved.

eng/versioning/external_dependencies.txt
kushagraThapar: Yes, I don't see a reason why cosmos can't update the micrometer dependency to 1.3.0 version. As you mentioned @JimSuplizio  this is not a blocker for this PR, and can be done later, but we should definitely have one common version.

PR #11037, opened 2020-05-11
Fix Event Hubs integration test failures

conniey:
* Make integration tests less flakey by pushing a single set of events to all partitions and fetching from them.
* Various renames.
* Adding `@Tag("integration")` so I can run a subset of tests regardless of environment variables set.
  * `-Dgroups=integration` or `-Dgroups=!integration` (JUnit 5 supports tag expressions)

### DIFF
--- a/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubProducerClientIntegrationTest.java
+++ b/sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/EventHubProducerClientIntegrationTest.java
@@ -23,7 +23,7 @@ class EventHubProducerClientIntegrationTest extends IntegrationTestBase {
     private static final String PARTITION_ID = "2";
     private EventHubProducerClient producer;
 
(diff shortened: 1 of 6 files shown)


sdk/eventhubs/azure-messaging-eventhubs/src/test/java/com/azure/messaging/eventhubs/BackCompatTest.java
```diff
@@ -29,19 +30,20 @@
...
-import static com.azure.messaging.eventhubs.TestUtils.MESSAGE_TRACKING_ID;
+import static com.azure.messaging.eventhubs.TestUtils.MESSAGE_ID;
 import static com.azure.messaging.eventhubs.TestUtils.getSymbol;
 import static com.azure.messaging.eventhubs.TestUtils.isMatchingEvent;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Integration test that verifies backwards compatibility with a previous version of the SDK.
  */
+@Tag(TestUtils.INTEGRATION)
```
srnagar, line 41: Tagging is great! Should be used in other services too!

PR #11044, opened 2020-05-12
Remove invalid com.azure.cosmos.models.JsonSerializable import

alzimmermsft:
Removes invalid import `com.azure.cosmos.models.JsonSerializable` which has been moved into `implementation`. This resolves build failures being seen in the `From Source` builds for Azure Core.


alzimmermsft:
@kushagraThapar if you could take a quick look.

PR #11046, opened 2020-05-12
Updated POM descriptions for Track 1 KeyVault.

vcolin7:
Updated all Track 1 KeyVault POMs to direct users to the Track 2 versions of the SDK.

### DIFF
--- a/sdk/keyvault/microsoft-azure-keyvault-complete/pom.xml
+++ b/sdk/keyvault/microsoft-azure-keyvault-complete/pom.xml
@@ -17,11 +17,13 @@ the MIT License. See License.txt in the project root for license information. --
   <packaging>pom</packaging>
 
   <name>Microsoft Azure Key Vault SDK Complete</name>
-  <description>This library has been replaced by new Azure SDKs, you can read about them at https://aka.ms/azsdkvalueprop. The latest libraries to interact with the Azure KeyVault service are:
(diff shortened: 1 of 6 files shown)


sdk/keyvault/microsoft-azure-keyvault-complete/pom.xml
```diff
@@ -17,7 +17,11 @@ the MIT License. See License.txt in the project root for license information. --
   <packaging>pom</packaging>
 
   <name>Microsoft Azure Key Vault SDK Complete</name>
-  <description>This encapsulate all modules of the Microsoft Azure Key Vault SDK.</description>
+  <description>This library has been replaced by new Azure SDKs, you can read about them at https://aka.ms/azsdkvalueprop. The latest libraries to interact with the Azure KeyVault service are:
+    1. https://search.maven.org/artifact/com.azure/azure-security-keyvault-keys
```
alzimmermsft, line 21: Does the description maintain newlines or do they need to be added with special characters?

vcolin7: That is actually something @g2vinay and were not sure about. Your comment made me do some research and found that we can use ${line.separator} as showed in [this StackOverflow comment](https://stackoverflow.com/a/16294529/4967034).

PR #11057, opened 2020-05-12
feat: suppress management in rev api

ChenTanyi:
Remaining problem for returning data-plane key vault class. Will ignore it in pipeline currently and discuss later.


ChenTanyi:
@JonathanGiles @mssfang Could you review whether it is ok to do such suppress in revapi?

PR #11084, opened 2020-05-12
azure-core 1.5 fixed this bug on odata.type

weidongxu-microsoft:


sdk/monitor/mgmt/src/main/java/com/azure/management/monitor/Action.java
```diff
@@ -14,7 +15,7 @@
 @JsonTypeInfo(
     use = JsonTypeInfo.Id.NAME,
     include = JsonTypeInfo.As.PROPERTY,
-    property = "odata.type",
+    property = "odata\\.type",
```
xccc-msft, line 18: Did you try to run living test for monitor locally after these changes?

weidongxu-microsoft: Already had a test for this.
https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/monitor/mgmt/src/test/java/com/azure/management/monitor/TypeSerializationTests.java#L19

yungezz:
hi @weidongxu-microsoft could you pls share more context about the change such as issue link etc? thanks.

weidongxu-microsoft:
@yungezz 

Issue on azure-core https://github.com/Azure/azure-sdk-for-java/issues/10217, which is fixed in 1.5.0

PR #11098, opened 2020-05-12
Azure External Dependencies BOM

### LINKED ISSUE #10342: Add First Level Dependencies to BOM
> Should we consider adding our direct dependencies to the BOM (JacksonXML, etc), at least temporarily?

Add the first level dependencies in the Azure SDKs for Java to `azure-sdk-bom` or into an `azure-sdk-parent-bom`.

alzimmermsft:
Fixes #10342

This PR aims to include the external dependencies that the Azure SDKs for Java use in a BOM for customer consumption.

#### Include external dependencies in `azure-sdk-bom`.

Including all external dependencies for the Azure client libraries in `azure-sdk-bom` will silo these dependencies based on the library type (client, data, management). This will create a very targeted scope which will allow for easier handling of the dependencies. This has the downside where these external dependencies are now tied to a specific track.

#### Create another BOM for external dependencies.

Creating a separate BOM separates the shipping and external dependencies. This would allow for consumers to use this separately from their Azure SDK dependencies. This has the downside when using multiple BOMs they need to be validated as compatible.

### Current State of the PR

Currently, the PR is representing the former option where external dependencies are being included in the `azure-sdk-bom`. Doing this allows consumers of the SDKs to both use compatible Azure Java SDKs and set their dependencies to match the versions the SDKs will be using. The aim for this is to reduce dependency conflicts consumers of the SDKs see.

To see the other option with having another BOM for external dependencies please view previous commits.

### DIFF
--- a/eng/.docsettings.yml
+++ b/eng/.docsettings.yml
@@ -115,7 +115,6 @@ known_content_issues:
   - ['sdk/authorization/microsoft-azure-authentication-msi-token-provider/README.md', '#3113']
   - ['sdk/batch/microsoft-azure-batch/README.md', '#3113']
   - ['sdk/boms/azure-sdk-bom/README.md', '#3113']
-  - ['sdk/boms/azure-external-dependencies-bom/README.md', '#3113']
   - ['sdk/core/README.md', '#3113']
   - ['sdk/cognitiveservices/ms-azure-cs-textanalytics/README.md', '#3113']
   - ['sdk/cosmos/changelog/README.md', '#3113']


sdk/boms/azure-external-dependencies-bom/pom.xml
```diff
@@ -0,0 +1,523 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.azure</groupId>
+  <artifactId>azure-external-dependencies-bom</artifactId>
```
joshfree, line 8: looking at prior art on maven, I think we can change this to "azure-core-dependencies" and drop the 'external' and 'bom'

eng/versioning/external_dependencies.txt
```diff
@@ -17,7 +17,6 @@ com.microsoft.azure:azure-client-runtime;1.7.3
 com.microsoft.azure:azure-core;0.9.8
 com.microsoft.azure:azure-keyvault-cryptography;1.2.2
 com.microsoft.azure:qpid-proton-j-extensions;1.2.3
-com.microsoft.azure:spotbugs-maven-plugin;1.2.1
```
mssfang, line 20: are we removing spotbugs here?

alzimmermsft: This package doesn't exist and, because it doesn't exist, this dependency tag isn't used anywhere.

sdk/boms/azure-external-dependencies-bom/pom.xml
alzimmermsft: Name changed, also adding another commit trying out adding all track 2 external dependencies into the `azure-sdk-bom`.

sdk/boms/azure-core-dependencies/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+To consume the BOM include it in the `dependencyManagement` section of your project's POM. Note that this **does not**
+result in all dependencies being included in your project.
+
+```xml
+<dependencyManagement>
+  <dependencies>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-external-dependencies-bom</artifactId>
```
srnagar, line 22: Should the name be updated to `azure-core-dependencies-bom`?

sdk/boms/azure-core-dependencies/pom.xml
```diff
@@ -0,0 +1,518 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.azure</groupId>
+  <artifactId>azure-core-dependencies</artifactId>
```
srnagar, line 8: Should this be `core-dependencies` or `external-dependencies`? I think `core` can be confusing as it seems like it's related to `azure-core` artifact.

sdk/boms/azure-core-dependencies/pom.xml
```diff
@@ -0,0 +1,518 @@
...
+        <version>1.7.28</version> <!-- {x-version-update;org.slf4j:slf4j-api;external_dependency} -->
+      </dependency>
+
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-simple</artifactId>
+        <version>1.7.25</version> <!-- {x-version-update;org.slf4j:slf4j-simple;external_dependency} -->
+      </dependency>
+
+      <dependency>
```
srnagar, line 366: Should we have all spring boot dependencies in this bom or is it better to keep this bom for only maintaining external dependencies used in client libraries?

alzimmermsft: I think the right bet going forward would be combining the SDK/Spring library being shipped and their external libraries. Having these split will just move the spot where dependency conflicts happen, or maybe we would need to create a BOM of BOMs which isn't the most ideal (nor do I know if it actually works).

sdk/boms/azure-core-dependencies/README.md
alzimmermsft: Yes, if this is being kept this should be renamed to `azure-core-dependencies`.

joshfree:
@JonathanGiles could you take a look and weigh in on the 1 vs 2 bom approach?

JonathanGiles:
I don't have a strong opinion on one vs two BOMs. I think one is sufficient for the use cases I can see. The bigger issue is ensuring there exists documentation that states that the BOM also lists our dependencies, and perhaps has a table of all such dependencies so that users are encouraged not to version these dependencies themselves.

mitchdenny:
@JonathanGiles Isn't the Spring starter kind of like a BOM anyway? If someone wants to use KeyVault with their spring application for example wouldn't they just reference the latest Spring starter which then references all the appropriate dependencies that they should be relying on?

mitchdenny:
Removed comment -- crossed threads ;)

mitchdenny:
What are the implications for release here. Do your dependencies move frequently enough that this could see us releasing a dependency BOM every time there is a CVE in our dependencies? Which would then mean we have to bump the main BOM as well?

Given that all our library POMs explicitly specify their dependency versions, doesn't this just result in two more modules that we need to release?

alzimmermsft:
@JonathanGiles, I can begin updating the README for the project to state this, additionally it could be added into the description in the POM.

@mitchdenny, that is a good question about how often we will be releasing the BOM. There are a lot of factors in play here around CVEs, libraries using it being updated, and having dependencies aligned in the SDKs so that the BOM can be released. Don't know if this will change our release process where we push more often to release to Maven just to update dependencies.

mitchdenny:
@alzimmermsft this is unblocked now. Still investigating the root cause. Could have just been a dropped message somewhere.

eng/versioning/external_dependencies.txt
JimSuplizio: Thank you @alzimmermsft!

sdk/boms/azure-sdk-bom/README.md
```diff
@@ -30,8 +32,8 @@ result in all dependencies being included in your project.
 
 ### Adding libraries to your project
 
-After adding the BOM all artifacts included in the BOM are now available to be added as a dependency without listing the
-artifact's version.
+After adding the BOM Azure client libraries included in the BOM are now available to be added as a dependency without
```
JimSuplizio, line 35: "After adding the BOM, Azure..." there should be a comma after the prepositional phrase.

sdk/boms/azure-sdk-bom/pom.xml
```diff
@@ -179,7 +374,7 @@
               <reporters>html</reporters>
               <showAll>true</showAll>
               <analyseBom>true</analyseBom>
-              <failOnDependencyConflict>true</failOnDependencyConflict>
+              <!--<failOnDependencyConflict>true</failOnDependencyConflict>-->
```
JimSuplizio, line 377: Why is this being commented out? If it needs to be removed then it should be removed entirely.

alzimmermsft: Commented it out as we want to have this enabled but a few of the external dependencies conflict in their dependencies. I've filed a few issues against @JonathanGiles's plugin to add support for dependency inclusions/exclusions when running the report to get around them. They are trivial things that are second order dependencies that aren't a priority with what we want to present here.

alzimmermsft:
> @alzimmermsft, we should chat. It may be worth me writing a script that'll climb through the pom files for client track libraries and pull out the external dependencies.

Could this be a follow-up to this change?

sdk/boms/azure-sdk-bom/README.md
alzimmermsft: Added the `,`.

eng/.docsettings.yml
```diff
@@ -115,6 +115,7 @@ known_content_issues:
   - ['sdk/authorization/microsoft-azure-authentication-msi-token-provider/README.md', '#3113']
   - ['sdk/batch/microsoft-azure-batch/README.md', '#3113']
   - ['sdk/boms/azure-sdk-bom/README.md', '#3113']
+  - ['sdk/boms/azure-external-dependencies-bom/README.md', '#3113']
```
srnagar, line 118: Now that we have combined Azure client libraries and external dependencies into a single bom, this is no longer required.

sdk/boms/azure-sdk-bom/pom.xml
```diff
@@ -47,117 +47,312 @@
...
         <artifactId>azure-storage-queue</artifactId>
-        <version>12.5.0</version>
+        <version>12.5.2</version>
       </dependency>
 
       <dependency>
         <groupId>com.azure</groupId>
         <artifactId>azure-ai-textanalytics</artifactId>
-        <version>1.0.0-beta.4</version>
+        <version>1.0.0</version>
```
srnagar, line 171: nit: it would be nice to have these dependencies listed in sorted order (separate external and azure sorted lists).

eng/.docsettings.yml
alzimmermsft: Removed.

sdk/boms/azure-sdk-bom/pom.xml
alzimmermsft: Dependencies are separated into SDK and external, sort order may not have been applied to the sections though.

sdk/boms/azure-sdk-bom/pom.xml
alzimmermsft: Removed the configuration.

PR #11101, opened 2020-05-12
File share soft delete

kasobol-msft:


sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareServiceAsyncClient.java
```diff
@@ -534,4 +538,64 @@ public String generateAccountSas(AccountSasSignatureValues accountSasSignatureVa
...
     }
+
+    /**
+     * Restores a previously deleted share.
+     * If the share associated with provided <code>deletedShareName</code>
+     * already exists, this call will result in a 409 (conflict).
+     * This API is only functional if Share Soft Delete is enabled
+     * for the storage account associated with the share.
+     * For more information, see the
+     * <a href="TBD">Azure Docs</a>. TODO (kasobol-msft) add link to REST API docs
```
alzimmermsft, line 549: Generally it is better to include TODOs outside of the comment for two reasons:

1. If we ship a beta or GA but forget to remove it we won't have silly documentation
2. IntelliJ and other IDEs offer functionality to highlight TODO comment lines.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareServiceAsyncClient.java
```diff
@@ -534,4 +538,64 @@ public String generateAccountSas(AccountSasSignatureValues accountSasSignatureVa
...
+    }
+
+    /**
+     * Restores a previously deleted share.
+     * If the share associated with provided <code>deletedShareName</code>
+     * already exists, this call will result in a 409 (conflict).
+     * This API is only functional if Share Soft Delete is enabled
+     * for the storage account associated with the share.
+     * For more information, see the
+     * <a href="TBD">Azure Docs</a>. TODO (kasobol-msft) add link to REST API docs
```
alzimmermsft, line 572: Same as the other TODO comment

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareServiceAsyncClient.java
alzimmermsft, line 549:
```suggestion
     * Restores a previously deleted share.
     * <p>
     * If the share associated with provided <code>deletedShareName</code>
     * already exists, this call will result in a 409 (conflict).
     * <p>
     * This API is only functional if Share Soft Delete is enabled
     * for the storage account associated with the share.
     * For more information, see the
     * <a href="TBD">Azure Docs</a>. TODO (kasobol-msft) add link to REST API docs
```

Javadocs are HTML paragraphs so newlines aren't maintained.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareServiceAsyncClient.java
```diff
@@ -534,4 +538,64 @@ public String generateAccountSas(AccountSasSignatureValues accountSasSignatureVa
         return new AccountSasImplUtil(accountSasSignatureValues)
             .generateSas(SasImplUtils.extractSharedKeyCredential(getHttpPipeline()));
     }
+
+    /**
+     * Restores a previously deleted share.
+     * If the share associated with provided <code>deletedShareName</code>
+     * already exists, this call will result in a 409 (conflict).
```
alzimmermsft, line 545: I'm a little confused here, will this undelete the specified share with the name of this share client? I'm guessing this will 409 if we already recreated a share with the delete name because you should rename the share instead?

sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/ShareServiceAsyncJavaDocCodeSamples.java
```diff
@@ -286,4 +288,41 @@ public void generateAccountSas() {
...
+        // BEGIN: com.azure.storage.file.share.ShareServiceAsyncClient.undeleteShare#String-String
+        ListSharesOptions listSharesOptions = new ListSharesOptions();
+        listSharesOptions.setIncludeDeleted(true);
+        fileServiceAsyncClient.listShares(listSharesOptions).flatMap(
+            deletedShare -> {
+                Mono<ShareAsyncClient> shareAsyncClient = fileServiceAsyncClient.undeleteShare(
+                    deletedShare.getName(), deletedShare.getVersion());
+                return shareAsyncClient;
+            }
+        ).then().block();
```
alzimmermsft, line 306: No need for `.then()`.

```suggestion
        ).block();
```

sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/ShareServiceAsyncJavaDocCodeSamples.java
```diff
@@ -286,4 +288,41 @@ public void generateAccountSas() {
...
+
+    /**
+     * Code snippet for {@link ShareServiceAsyncClient#undeleteShare(String, String)}.
+     */
+    public void undeleteShare() {
+        ShareServiceAsyncClient fileServiceAsyncClient = createAsyncClientWithSASToken();
+        // BEGIN: com.azure.storage.file.share.ShareServiceAsyncClient.undeleteShare#String-String
+        ListSharesOptions listSharesOptions = new ListSharesOptions();
+        listSharesOptions.setIncludeDeleted(true);
+        fileServiceAsyncClient.listShares(listSharesOptions).flatMap(
```
alzimmermsft, line 300: Are we guessing that the common pattern will be to list deleted shares to undelete them?

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileServiceAPITests.groovy
```diff
@@ -108,38 +112,46 @@ class FileServiceAPITests extends APISpec {
...
+        includeDeleted || !shares.hasNext()
 
         where:
-        options                                                                                                | limits | includeMetadata | includeSnapshot
-        new ListSharesOptions().setPrefix("fileserviceapitestslistshareswithfilter")                           | 3      | false           | true
-        new ListSharesOptions().setPrefix("fileserviceapitestslistshareswithfilter").setIncludeMetadata(true)  | 3      | true            | true
-        new ListSharesOptions().setPrefix("fileserviceapitestslistshareswithfilter").setIncludeMetadata(false) | 3      | false           | true
-        new ListSharesOptions().setPrefix("fileserviceapitestslistshareswithfilter").setMaxResultsPerPage(2)   | 3      | false           | true
+        options                                           | limits | includeMetadata | includeSnapshot | includeDeleted
+        new ListSharesOptions()                           | 3      | false           | true            | false
```
alzimmermsft, line 142: We should probably continue to maintain prefixes in cases where there is stale testing data. I know this has been an issue in the past which caused flakiness in tests.

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileServiceAPITests.groovy
```diff
@@ -235,4 +252,60 @@ class FileServiceAPITests extends APISpec {
...
+    def "Restore share min"() {
+        given:
+        def shareClient = primaryFileServiceClient.getShareClient(generateShareName())
+        shareClient.create()
+        def fileName = generatePathName()
+        shareClient.getFileClient(fileName).create(2)
+        shareClient.delete()
+        if (TestMode.PLAYBACK != getTestMode()) {
+            Thread.sleep(30000)
+        }
```
alzimmermsft, line 265: I believe there is a utility method in APISpec that does this.

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileServiceAPITests.groovy
```diff
@@ -235,4 +252,60 @@ class FileServiceAPITests extends APISpec {
...
+    def "Restore share max"() {
+        given:
+        def shareClient = primaryFileServiceClient.getShareClient(generateShareName())
+        shareClient.create()
+        def fileName = generatePathName()
+        shareClient.getFileClient(fileName).create(2)
+        shareClient.delete()
+        if (TestMode.PLAYBACK != getTestMode()) {
+            Thread.sleep(30000)
+        }
```
alzimmermsft, line 288: Same comment.

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileServiceAsyncAPITests.groovy
```diff
@@ -266,4 +272,69 @@ class FileServiceAsyncAPITests extends APISpec {
...
+        def fileName = generatePathName()
+        def delay = TestMode.PLAYBACK == getTestMode() ? 0L : 30000L
+        def shareItemMono = shareClient.create()
+            .then(shareClient.getFileClient(fileName).create(2))
+            .then(shareClient.delete())
+            .then(Mono.delay(Duration.ofMillis(delay)))
+            .then(primaryFileServiceAsyncClient.listShares(
+                new ListSharesOptions()
+                    .setPrefix(shareClient.getShareName())
+                    .setIncludeDeleted(true)).next())
```
alzimmermsft, line 288: I would break this into two parts.

1. Create the resources need using a block (or StepVerifier, either works).
2. Then use sleep if record before going into when.

Right now this feels that setup contains setup and when logic.

sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/ShareServiceAsyncJavaDocCodeSamples.java
kasobol-msft: Correct. We need two things to undelete a share - name and version. Listing API is the only way to obtain version.

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileServiceAPITests.groovy
kasobol-msft: The prefix is maintained. It just set in `given` clause instead of hardcoding.
We do this : `options.setPrefix(shareName)` where shareName is coming from `shareName = testResourceName.randomName(methodName, 60)` in the `setup()`.

This had to be changed to include some randomness in search prefix because share with softdelete enabled actually accumulates stale data for some time (i.e. 1 day).

sdk/storage/azure-storage-file-share/src/test/java/com/azure/storage/file/share/FileServiceAPITests.groovy
kasobol-msft: I've seen that in blobs but not in shares. I'll double check and add if missing.

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareServiceAsyncClient.java
kasobol-msft: This API is on `ShareServiceAsyncClient` a level up from `ShareAsyncClient` (to have parity with outcome of this discussion https://github.com/Azure/azure-sdk-for-java/pull/10900#discussion_r423420963 for blobs). Therefore `deletedShareName` is the only source of name for restored share. So, if someone created new (or already undeleted) share with same name as `deletedShareName` then restore fails as name is already taken.

The option to rename share while undeleting is not part of this PR due to server side not being ready to release it.

sdk/storage/azure-storage-file-share/src/samples/java/com/azure/storage/file/share/ShareServiceAsyncJavaDocCodeSamples.java
kasobol-msft: it's a collection (aka Flux) . I can do blockFirst()

sdk/storage/azure-storage-file-share/src/main/java/com/azure/storage/file/share/ShareServiceAsyncClient.java
```diff
@@ -534,4 +538,74 @@ public String generateAccountSas(AccountSasSignatureValues accountSasSignatureVa
...
+            return withContext(context ->
+                undeleteShareWithResponse(
+                    deletedShareName, deletedShareVersion, context));
+        } catch (RuntimeException ex) {
+            return monoError(logger, ex);
+        }
+    }
+
+    Mono<Response<ShareAsyncClient>> undeleteShareWithResponse(
+        String deletedShareName, String deletedShareVersion, Context context) {
```
rickle-msft, line 605: The rest docs seem to suggest you can rename the share like you can for containers. Is that no longer true? Or is the swagger not accurate?

kasobol-msft: Yeah they do. However, for shares, that option won't be in stg73 but later.

PR #11102, opened 2020-05-12
Removed authorization header from cosmos client exception

kushagraThapar:
* Removed authorization header from cosmos client exception
* Added tests for checking authorization header on cosmos client exception.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
@@ -226,14 +226,7 @@ public final class BridgeInternal {
 
     public static <E extends CosmosClientException> void setRequestHeaders(CosmosClientException cosmosClientException,
                                                                            Map<String, String> requestHeaders) {
-        Map<String, String> requestHeadersMap = new HashMap<>();
-        requestHeaders.forEach((key, value) -> {
-            //  Skip the authorization header
-            if (!key.equals(HttpConstants.HttpHeaders.AUTHORIZATION)) {
-                requestHeadersMap.put(key, value);
-            }
-        });
-        cosmosClientException.requestHeaders = requestHeadersMap;
+        cosmosClientException.requestHeaders = requestHeaders;
     }
 
     public static <E extends CosmosClientException> Map<String, String> getRequestHeaders(
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientException.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientException.java
@@ -14,7 +14,9 @@ import com.azure.cosmos.implementation.apachecommons.lang.StringUtils;
 
 import java.time.Duration;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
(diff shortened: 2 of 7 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
```diff
@@ -226,7 +226,14 @@ public static boolean isEnableMultipleWriteLocations(DatabaseAccount account) {
 
     public static <E extends CosmosClientException> void setRequestHeaders(CosmosClientException cosmosClientException,
                                                                            Map<String, String> requestHeaders) {
-        cosmosClientException.requestHeaders = requestHeaders;
+        Map<String, String> requestHeadersMap = new HashMap<>();
```
moderakh, line 229: This will create a new Map for the headers, and so you have to pay the additional cost of Map creation.
can we instead change the filtering when the message is logged:

You can have this in the CosmosClientException when the message is logged. this would help to pay the cost only if the message is logged.
```java
    @Override
    public String toString() {
        return getClass().getSimpleName() + "{" + "error=" + cosmosError + ", resourceAddress='"
                   + resourceAddress + '\'' + ", statusCode=" + statusCode + ", message=" + getMessage()
                   + ", causeInfo=" + causeInfo() + ", responseHeaders=" + filterSensitiveDate(responseHeaders) + ", requestHeaders="
                   + filterSensitiveDate(requestHeaders) + '}';
    }

    private List<Map.Entry<String, String>> filterSensitiveDate(Map<String, String> requestHeaders) {
        return requestHeaders.entrySet().stream().filter(entry -> !HttpConstants.HttpHeaders.AUTHORIZATION.equalsIgnoreCase(entry.getKey()))
            .collect(Collectors.toList());
    }
```

milismsft, line 229: why not removing the auth header field directly from the requestHeaders rather than making a copy of it? For the auth entry we are suppose to re-populate that every time rather than reusing it so if there are failures because of it then we have other bugs to fix :-)

kushagraThapar: Make sense !

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosClientException.java
```diff
@@ -259,7 +261,7 @@ public String toString() {
         return getClass().getSimpleName() + "{" + "error=" + cosmosError + ", resourceAddress='"
                    + resourceAddress + '\'' + ", statusCode=" + statusCode + ", message=" + getMessage()
                    + ", causeInfo=" + causeInfo() + ", responseHeaders=" + responseHeaders + ", requestHeaders="
-                   + requestHeaders + '}';
+                   + filterSensitiveData(requestHeaders) + '}';
```
milismsft, line 264: We're still leaking the auth token; just because we override the toString(), it will not prevent someone trapping and dumping the exception object via an object mapper or some other serialization method which ignores the toString() override.

The initial fix that creates a copy of the request header map is a better one if we cannot reuse the original header while removing the sensitive header parts. I checked the code and we only invoke the respective setRequestHeaders() method in exceptional cases only; the only concern here is the memory hit which could go as high as 16k per each request header.

Issue #10311, opened 2020-04-16
[BUG] ManagementClient does not support connecting via proxy


giventocode:
We have tried several approaches without any luck:
-  Setting  global proxy settings in the JVM and in code  (```https.proxyHost``` and ```https.proxyPort```)
-  Implement a custom ```ProxySelector``` and set it as a default. 

Through traces we have verified that the calls to the Service Bus endpoint are not going through the proxy.

giventocode:
Further looking into this, it seems like the issue is how the management client is instantiating the underlying http client.

```
 public ManagementClientAsync(URI namespaceEndpointURI, ClientSettings clientSettings) {
        this.namespaceEndpointURI = namespaceEndpointURI;
        this.clientSettings = clientSettings;
        DefaultAsyncHttpClientConfig.Builder clientBuilder = Dsl.config()
                .setConnectTimeout((int) CONNECTION_TIMEOUT.toMillis())
                .setRequestTimeout((int) this.clientSettings.getOperationTimeout().toMillis());
        this.asyncHttpClient = asyncHttpClient(clientBuilder);
    }
```

By default the ```asyncHttpClient``` does not use the standard Java ```ProxySelector```. 

https://github.com/AsyncHttpClient/async-http-client/blob/ac2d4bdf46abf96aeb5a61d56bc5718ac005184a/client/src/main/resources/org/asynchttpclient/config/ahc-default.properties#L18

This capability should be enabled by default in the library or it should expose a config mechanism to enable it.

joshfree:
Thanks for reporting this @giventocode.  @xccc-msft as part of moving the Managment Plane to the new azure-core (track 2) you'll get Proxy support (/cc @alzimmermsft)  - could you please follow up with @giventocode

xccc-msft:
@giventocode Would you share more details regarding which SDK and version you are using?

giventocode:
We are using version, 3.1.3. 

Also, we have found a workaround. It is possible to set the underlying HTTP client configuration in maven.

```
mvn ... -Dorg.asynchttpclient.useProxySelector=true
``` 

However, the ideal scenario is allowing us to set these setting programmatically. 

So a few options come to mind: 

1. Add proxy configuration settings to ```ClientSettings```
2. Expose the HTTP client builder.

For option 2, I have an implementation, link below, and I can submit a PR if that helps.

[...]

PR #11108, opened 2020-05-12
Enable connection via proxy for ManagementClient

DorothySun216:
Currently we don't support connection via proxy for ManagementClient, but we have proxy support for QueueClient, SubscriptionClient, MessageSender, MessageReceiver via AMQP. This change will address the feature gap. 
https://github.com/Azure/azure-sdk-for-java/issues/10311

### DIFF
--- a/sdk/servicebus/microsoft-azure-servicebus/src/test/java/com/microsoft/azure/servicebus/ManagementClientProxyTest.java
+++ b/sdk/servicebus/microsoft-azure-servicebus/src/test/java/com/microsoft/azure/servicebus/ManagementClientProxyTest.java
@@ -5,6 +5,7 @@ import com.microsoft.azure.servicebus.management.QueueDescription;
 import com.microsoft.azure.servicebus.primitives.ConnectionStringBuilder;
 import com.microsoft.azure.servicebus.primitives.TransportType;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 import java.io.IOException;
 import java.net.*;
@@ -18,6 +19,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
 public class ManagementClientProxyTest {
+    @Ignore
     @Test
     public void managementClientWithProxy() throws Exception {
         String proxyHostName = "127.0.0.1";


sdk/servicebus/microsoft-azure-servicebus/src/test/java/com/microsoft/azure/servicebus/ManagementClientProxyTest.java
```diff
@@ -0,0 +1,57 @@
...
+        String proxyHostName = "127.0.0.1";
+        int proxyPort = 8888;
+        final ProxySelector systemDefaultSelector = ProxySelector.getDefault();
+
+        ProxySelector.setDefault(new ProxySelector() {
+            @Override
+            public List<Proxy> select(URI uri) {
+                if (uri != null
+                    && uri.getHost() != null
+//                    && uri.getHost().equalsIgnoreCase(proxyHostName)
```
DorothySun216, line 30: I don't quite know why this sample code has this line. Since I am using localhost and the host name will be different from service bus host. If I enable this line, the proxy cannot be created properly. I don't think this line is necessary but just want to point out.

sdk/servicebus/microsoft-azure-servicebus/src/main/java/com/microsoft/azure/servicebus/management/ManagementClientAsync.java
```diff
@@ -92,9 +92,48 @@ public ManagementClientAsync(URI namespaceEndpointURI, ClientSettings clientSett
...
+            String proxyHostName=address.getHostName();
+            int proxyPort=address.getPort();
+
+            clientBuilder.setProxyServer(new ProxyServer.Builder(proxyHostName,proxyPort));
+        }
+
         this.asyncHttpClient = asyncHttpClient(clientBuilder);
     }
 
+    public boolean shouldUseProxy(final String hostName) {
```
DorothySun216, line 107: These three methods are taken out from ProxyConnectionHandler. Will leave as a TODO to make them as a util class and shared between these two classes.

sdk/servicebus/microsoft-azure-servicebus/src/main/java/com/microsoft/azure/servicebus/management/ManagementClientAsync.java
```diff
@@ -92,9 +92,48 @@ public ManagementClientAsync(URI namespaceEndpointURI, ClientSettings clientSett
...
                 .setConnectTimeout((int) CONNECTION_TIMEOUT.toMillis())
                 .setRequestTimeout((int) this.clientSettings.getOperationTimeout().toMillis());
+
+        if(shouldUseProxy(this.namespaceEndpointURI.getHost())){
+            InetSocketAddress address = (InetSocketAddress)this.proxies.get(0).address();
+            String proxyHostName=address.getHostName();
+            int proxyPort=address.getPort();
+
+            clientBuilder.setProxyServer(new ProxyServer.Builder(proxyHostName,proxyPort));
+        }
```
giventocode, line 102: I am wondering if it may be easier/better to set the properties in the client that enables the integration with the default mechanisms that Java networking uses for proxy configuration - documented [here](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html).  

```suggestion
                .setUseProxySelector(true)
                .setUseProxyProperties(true)
                .setRequestTimeout((int) this.clientSettings.getOperationTimeout().toMillis());
```

DorothySun216: @giventocode Thanks for pointing to this resource to offer more background. I verified that this change would have the same impact: select the default proxy set by ProxySelector, this is also what I was trying to do in my changes but don't aware that there are already system level support built into Java. I also verified this from here: https://www.javadoc.io/static/com.ning/async-http-client/1.9.31/com/ning/http/client/AsyncHttpClientConfig.Builder.html. It also discusses "If useProxySelector is set to true but setProxyServer(ProxyServer) was used to explicitly set a proxy server, the latter is preferred.", so should we just have .setUseProxySelector(true)? Have also verified that setting only this property will achieve the same goal in my test.

giventocode: Thanks for the additional insights. I further reviewed the code and in a nutshell: if there's a proxy selector, the proxy uri is retrieved from it and then is set explicitly as the proxy server to use.  So a few things to consider:

  - One of the benefits of implementing a proxy selector is that it allows you to filter/decide what proxy to use for a given URL - by overriding ``` java.util.List<Proxy> select(URI uri)```. The current implementation seems like it'd block this ability as the proxy selector won't be used at all.
 
  - Java also provides a standard way to set the proxy explicitly, as per below. It seems like this also won't be possible.  
```
System.setProperty("http.proxyHost", "webcache.example.com");
System.setProperty("http.proxyPort", "8080");
```  

As per the suggestion, if we set the flags in the underlying config of the asynchttpclient so it enables the default Java proxy setting behavior, then users can use the Java standard way of setting the proxies and extending how proxies are used programmatically.

DorothySun216: I have discussed with @giventocode offline and we have decided to add both `.setUseProxySelector(true)` (compatible with AMQP websockets to use ProxySelector for send/receive operations) and `.setUseProxyProperties(true)` (users can use the Java standard way of setting the proxies and extending how proxies are used programmatically for ManagementClient.)

sdk/servicebus/microsoft-azure-servicebus/src/test/java/com/microsoft/azure/servicebus/ManagementClientProxyTest.java
```diff
@@ -0,0 +1,57 @@
...
+import java.net.InetSocketAddress;
+import java.net.ProxySelector;
+import java.net.URI;
+import java.util.LinkedList;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+
+public class ManagementClientProxyTest {
+    @Test
+    public void managementClientWithProxy() throws InterruptedException, ServiceBusException {
```
nemakam: >managementClientWithProxy [](start = 16, length = 25)

You can ignore this test. The CI/CD machine doens't have a proxy server and hence this won't run there.

PR #11116, opened 2020-05-13
Tracing updates

samvaity:
- [ ] Update log levels in tracing code to `verbose` [Cosmos need]
- [ ] Update code to support only a single implementation of Tracer. Being consistent with other languages. Also, not a real use case for supporting multiple tracer implementations at a given time.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProxy.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProxy.java
@@ -15,13 +15,13 @@ import java.util.ServiceLoader;
  */
 public final class TracerProxy {
 
-    private static Tracer TRACER;
+    private static Tracer tracer;
 
     static {
         ServiceLoader<Tracer> serviceLoader = ServiceLoader.load(Tracer.class);
         Iterator<?> iterator = serviceLoader.iterator();
         if (iterator.hasNext()) {
-            TRACER = serviceLoader.iterator().next();
+            tracer = serviceLoader.iterator().next();
         }
     }
 
@@ -42,10 +42,10 @@ public final class TracerProxy {
      * @return An updated {@link Context} object.
      */
     public static Context start(String methodName, Context context) {
-        if (TRACER == null) {
+        if (tracer == null) {
             return context;
         }
-        return TRACER.start(methodName, context);
+        return tracer.start(methodName, context);
     }
 
     /**
@@ -57,10 +57,10 @@ public final class TracerProxy {
      * @param context Additional metadata that is passed through the call stack.
      */
     public static void setAttribute(String key, String value, Context context) {
-        if (TRACER == null) {
+        if (tracer == null) {
             return;
         }
-        TRACER.setAttribute(key, value, context);
+        tracer.setAttribute(key, value, context);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProxy.java
```diff
@@ -17,14 +17,13 @@
...
-    private static final List<Tracer> TRACERS;
+    private static Tracer TRACER;
+
     static {
         ServiceLoader<Tracer> serviceLoader = ServiceLoader.load(Tracer.class);
-        List<Tracer> tracers = new ArrayList<>();
-        for (Tracer tracer : serviceLoader) {
-            tracers.add(tracer);
+        if (serviceLoader != null) {
+            TRACER = serviceLoader.iterator().next();
```
alzimmermsft, line 25: Do we have any mechanisms available which will ensure a certain tracer is loaded if I happen to bring multiple? This can be a follow-up as it is definitely an edge case.

samvaity: Yeah, I think we could add a feature where a user preference could be preferred for a single implementation if multiple found.

srnagar: Or, if there are multiple instances of tracer implementations, should we support all of them? Create spans on each tracer implementation?

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProxy.java
```diff
@@ -84,11 +86,9 @@ public static void end(int responseCode, Throwable error, Context context) {
      * @return An updated {@link Context} object.
      */
     public static Context setSpanName(String spanName, Context context) {
```
srnagar, line 88: Why does this API return `Context`? It's inconsistent with `setAttribute()` API which returns a void. Or, should setAttribute() return Context too since `start()` is also returning `Context`?

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProxy.java
samvaity: There is no real use case to support multiple implementations, as users would always tie it with a single implementation of tracer to an exporter. 
It brings in a lot of complexity for the users of the library to actually look to configure each span with a single tracer implementation and then the corresponding exporter for that tracer. As mentioned above, not a real use case and something only Java SDK was supporting.

sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProxy.java
samvaity: `start()` should return a `Context` as we update the context to hold the span that it just created/started.
`setAttribute()`  shouldn't because we do not have anything to return. We are actually just using the values from the provided `Context`.
`setSpanName()` I think this method is redundant to set the span name as we already use the provided context for that. I see it is only used in [RestProxy here](https://github.com/Azure/azure-sdk-for-java/blob/e2a19ea5db4b63c953948e41f0e25343faf0a221/sdk/core/azure-core/src/main/java/com/azure/core/http/rest/RestProxy.java#L185).

PR #11126, opened 2020-05-13
mgmt update policy and subscriptions to latest

weidongxu-microsoft:
update policy and subscriptions to latest

reason is the ErrorResponse of old policy spec is not correct for ManagementError.
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/resources/resource-manager/Microsoft.Authorization/preview/2017-06-01-preview/policyAssignments.json#L585-L603


weidongxu-microsoft:
This is one of the preparation task for regen all services. Since regen might need to be done multiple times, I would try to have separate PRs to fix/enhance related areas.

PR #11143, opened 2020-05-13
Change preview to beta in version strings

JimSuplizio:
The contributing had older versioning strings. "preview" is no longer allowed, we changed things to "beta" due to maven's sorting which sorted preview above the GA version.


JimSuplizio:
I used the override in this PR because it's a minor update to the Java repository's CONTRIBUTING.md doc.

PR #11152, opened 2020-05-13
Generate LUIS authoring with latest swagger

jianghaolu:


jianghaolu:
Thanks @JimSuplizio  for taking a look. The 1.0.4-beta was for computer vision which I was working on at the same time and got mixed. Sorry for the confusion.

I really want to adapt to the new versioning schemes too but unfortunately we've gone down the path of using the less than ideal versioning since 2018, as you noticed, for a few versions now. And this is the versioning scheme not only for LUIS authoring, but another dozen cognitive services. Having this one library, and this one version being different doesn't feel very user-friendly.

Also this is Track 1 library. Talked to @JonathanGiles and he is also leaning towards keeping Track 1 as is for now. We are versioning our Track 2 cognitive service libraries in the correct versioning scheme: https://repo1.maven.org/maven2/com/azure/azure-ai-textanalytics/, and coincidently this differentiates well from these auto-generated Track 1 libraries.

jianghaolu:
TBH once in a while I receive a request to release one of the cognitive services Track 1 SDKs. Though they only have to click a mouse a couple times to request a release, it's a PITA for me as I have to use a generator from someone's GitHub fork who left the team couple years ago that only runs on a specific AutoRest & Node version, and apply a bunch of manual fixes, and being ashamed of why in the world I'm releasing these ugly code with this unorthodox version, I can't wait for Track 2 libraries for these as much as you do.
I've done this a couple times, including the QnA Maker last time that probably looked really offensive from your perspective of upholding engineering excellence. But I would really appreciate it if you can feel my pain here and make it a little easier for me to do this ugly work no one else wants to do. Thanks!

anuchandy:
I agree with Jianghao; If we can continue the versioning scheme already established, it will be less confusing for existing customers of track1 cognitive libs. Ref [computer-version](https://repo1.maven.org/maven2/com/microsoft/azure/cognitiveservices/azure-cognitiveservices-computervision/) track1 versions here. As we move these libs to track2, we can follow the new versioning scheme.

JimSuplizio:
At this point if @JonathanGiles is OK then go for it.

jianghaolu:
Thanks guys for keeping an extra set of eyes on this! I hope we will soon switch to Track 2 generator and deprecate all these.

PR #11195, opened 2020-05-14
share artifact staging. add release stage

scbedd:
@Azure/azure-sdk-eng

Edit: @mitchdenny, @JimSuplizio  for brevity to not handle the `-signed` versus unsigned, I just swapped anything parsing the artifacts to use the unsigned artifacts folder.

I don't believe we actually _care_ about the signed values, due to the fact we're actually going after metadata or the raw html, but I feel that it's a callout that I should make.

Edit: [Successful Release](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=396130&view=results)

### DIFF
--- a/sdk/template/azure-sdk-template/CHANGELOG.md
+++ b/sdk/template/azure-sdk-template/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 1.0.4-beta.15 (2020-05-18)
+## 1.0.4-beta.15 (Unreleased)
 - Test Java Release Pipeline
 
 ## 1.0.4-beta.14 (2020-05-18)


weshaggard:
>Edit: @mitchdenny, @JimSuplizio for brevity to not handle the -signed versus unsigned, I just swapped anything parsing the artifacts to use the unsigned artifacts folder.

Is there any reason we shouldn't do the reverse and always use the signed versions? The main reason I'm curious is if we start to use this template for other areas like the actual publishing we will want to be sure we get the signed version and if all of them can work against the signed version that will make it a little more consistent and less error prone were we don't accidently start to publish unsigned bits.

eng/pipelines/templates/stages/archetype-java-release.yml
```diff
@@ -85,21 +87,58 @@ stages:
...
+          - ${{if ne(artifact.options.skipPublishDocs, 'true')}}:
+            - deployment: PublicDocsMs
+              displayName: "Publish Updates for Docs.MS"
+              condition: and(succeeded(), ne(variables['Skip.PublishDocs'], 'true'))
+              environment: github
+              timeoutInMinutes: 5
+              dependsOn: VerifyReleaseVersion
+
+              pool:
+                vmImage: vs2017-win2016
```
weshaggard, line 113: orthogonal to your change but we probably should see about moving these to the latest windows image, windows-2019

JimSuplizio:
> > Edit: @mitchdenny, @JimSuplizio for brevity to not handle the -signed versus unsigned, I just swapped anything parsing the artifacts to use the unsigned artifacts folder.
> 
> Is there any reason we shouldn't do the reverse and always use the signed versions? The main reason I'm curious is if we start to use this template for other areas like the actual publishing we will want to be sure we get the signed version and if all of them can work against the signed version that will make it a little more consistent and less error prone were we don't accidently start to publish unsigned bits.

I would agree. The previous code in [archetype-java-release.yml](https://github.com/Azure/azure-sdk-for-java/blob/master/eng/pipelines/templates/stages/) wasn't but that was only directly being used to copy the docs to blob storage. If we were to use this template for more than just that then the signed versions would make the script more versatile.

scbedd:
Ok! That's the feedback I was looking for! Will double check that all the same artifacts are available, then leverage those.

scbedd:
Alright, merged the fix to go after the signed artifacts. Will do a test run w/ template, link the release, and merge.

scbedd:
@weshaggard @JimSuplizio @mitchdenny ready for re-check. Release confirmed.

eng/pipelines/templates/stages/archetype-java-release.yml
```diff
@@ -85,14 +87,11 @@ stages:
...
-                        Get-ChildItem $(Pipeline.Workspace)/${{parameters.ArtifactName}}-signed
-                        New-Item -Type Directory -Name ${{artifact.safeName}} -Path $(Pipeline.Workspace)
-                        Copy-Item $(Pipeline.Workspace)/${{parameters.ArtifactName}}-signed/${{artifact.name}}-[0-9]*.[0-9]*.[0-9]* $(Pipeline.Workspace)/${{artifact.safeName}}
-                        Get-ChildItem $(Pipeline.Workspace)/${{artifact.safeName}}
-                      workingDirectory: $(Pipeline.Workspace)
-                      displayName: Stage artifacts
-                      timeoutInMinutes: 5
+                    - template: /eng/pipelines/templates/steps/stage-artifacts.yml
+                      parameters:
+                        SourceFolder: ${{parameters.ArtifactName}}-signed
```
weshaggard, line 92: We should consider adding the "-signed" as part of the template.

scbedd: That's fair. I'm not making any further changes on this PR.

scbedd: That's fair. I am not going to make further changes to this PR at this time.

Issue #10416, opened 2020-04-22
[FEATURE REQ] core-mgmt - move PollerFactory out of implementation package and into polling


weidongxu-microsoft:
Move PollerFactory out of implementation package and into polling

weidongxu-microsoft:
Depends on https://github.com/Azure/azure-sdk-for-java/issues/8266

PR #11203, opened 2020-05-15
move PollerFactory to polling package; remove AzureHost

weidongxu-microsoft:
Fix https://github.com/Azure/azure-sdk-for-java/issues/10416


sdk/core/azure-core-management/src/main/java/com/azure/core/management/polling/PollResult.java
```diff
@@ -38,7 +38,7 @@ public PollResult(Error error) {
      *
      * @return the value
      */
-    public T value() {
+    public T getValue() {
```
weidongxu-microsoft, line 41: Change according to java sdk guideline.

sdk/core/azure-core-management/src/main/java/com/azure/core/management/implementation/polling/PollOperation.java
```diff
@@ -7,142 +7,28 @@
...
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
 /**
- * Factory to create PollerFlux for Azure resource manager (ARM) long-running-operation (LRO).
+ * Poll operation for Azure resource manager (ARM) long-running-operation (LRO).
  */
-public final class PollerFactory {
-    private static final ClientLogger LOGGER = new ClientLogger(PollerFactory.class);
+public final class PollOperation {
```
weidongxu-microsoft, line 27: The remaining public method (used by `polling.PollerFactory`) in the class is for pollingOperation, cancelOperation and fetchResultOperation, hence changing the class name to `PollOperation`.

Let me know if any suggestion.

weidongxu-microsoft:
One remaining question is about `AzureJacksonAdapter` and `ManagementErrorDeserializer`. They are only util classes, which can be move elsewhere (as long as they are registered to `JacksonAdapter`).

Together with another `DateTimeDeserializer` https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/resources/mgmt/src/main/java/com/azure/management/AzureServiceClient.java#L170-L190, I can move them to azure-management.

And later if we need to generate new SDK for Fluent lite, I can try generate them with other java classes (the only down side is that they will be duplicate for each mgmt service, but probably not much problem since they are only utils).

eng/versioning/version_client.txt
```diff
@@ -11,7 +11,7 @@ com.azure:azure-core-amqp;1.1.2;1.2.0-beta.1
 com.azure:azure-core-http-jdk-httpclient;1.0.0-beta.1;1.0.0-beta.1
 com.azure:azure-core-http-netty;1.5.1;1.6.0-beta.1
 com.azure:azure-core-http-okhttp;1.2.2;1.3.0-beta.1
-com.azure:azure-core-management;1.0.0-beta.8;1.0.0-beta.8
+com.azure:azure-management-core;1.0.0-beta.8;1.0.0-beta.8
```
JimSuplizio, line 14: /CC @JonathanGiles 
I'm curious. com.azure:azure-core-management has previously been released several times, is there any reason why the artifact ID is being changed?

yungezz: thanks @JimSuplizio for reviewing. Actually this is based on @JonathanGiles feedback.

JimSuplizio: /CC @mitchdenny 
Everything under sdk/core is azure-core-<something>, and their directories match the artifact name. If this is being renamed then there probably should be more changes than just renaming the artifact. @JonathanGiles is this no longer a core artifact, is this supposed to moving somewhere else or is there a different naming convention happening for management?

weidongxu-microsoft: Thanks @JimSuplizio . Personal impression on current package: no lib other than mgmt SDK would benefit for including the azure-core-management lib. So this probably no very suitable for a core lib.

Yes, there are likely follow-up actions if we do change artifactId. For now this part is only trial on a suggestion. I won't merge the artifactId change without approval from you or Jonathon.

weidongxu-microsoft:
Revert the change on artifactId.

PR #11209, opened 2020-05-15
Use Linux for Analyze step while issue is being resolved

alzimmermsft:


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -132,7 +132,7 @@ jobs:
       - template: ../variables/globals.yml
 
     pool:
-      vmImage: 'windows-2019'
```
srnagar, line 135: I like this workaround better than the other PR which changes the JDK version from 11 to 8.

PR #11211, opened 2020-05-15
Override bad java11 selection

scbedd:
`Analyze` only because that's the only one with the issue right now.

@JimSuplizio @alzimmermsft

### DIFF
--- a/eng/pipelines/templates/steps/update-version-fix.yml
+++ b/eng/pipelines/templates/steps/update-version-fix.yml
@@ -1,10 +1,7 @@
-parameters:
-  VMAgent: variables['OSVmImage']
-
 steps:
   - pwsh: |
       $java11Installs = Get-ChildItem -Path 'C:\Program Files\Java' -Filter '*azure-jdk_11*' | Sort-Object -Property Name -Descending | Select-Object -First 1
       $latestJava11Install = $java11Installs.FullName;
       Write-Host "##vso[task.setvariable variable=JAVA_HOME_11_X64]$latestJava11Install"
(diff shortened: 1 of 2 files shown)


common/smoke-tests/smoke-test.yml
```diff
@@ -80,6 +80,10 @@ jobs:
 
 
     steps:
+      - template: /eng/pipelines/templates/steps/update-version-fix.yml
+        parameters:
```
weshaggard, line 84: To reduce duplication and the number of lines we have to touch I suggest defaulting the VMAgent parameter to `$(OSVmImage)` in the template so we don't have to pass it in explicitly in all our usages.

eng/pipelines/templates/steps/update-version-fix.yml
```diff
@@ -0,0 +1,10 @@
+parameters:
+  VMAgent: variables['OSVmImage']
+
+steps:
+  - pwsh: |
+      $java11Installs = Get-ChildItem -Path 'C:\Program Files\Java' -Filter '*azure-jdk_11*' | Sort-Object -Property Name -Descending | Select-Object -First 1
+      $latestJava11Install = $java11Installs.FullName;
+      Write-Host "##vso[task.setvariable variable=JAVA_HOME_11_X64]$latestJava11Install"
+    displayName: Resolve bad selection of Java11 instead of Java13
+    condition: and(succeeded(), eq(${{parameters.VMAgent}},'windows-2019'))
```
weshaggard, line 10: I believe this needs to be `eq('${{parameters.VMAgent}}','windows-2019'))`

PR #11234, opened 2020-05-17
Regenerating track1 computer-vision for api-version 3.0

anuchandy:
Uses the existing track1 versioning scheme for cognitive, not the revised versioning. We'll be using the new versioning scheme when we move to track2. Ref: https://github.com/Azure/azure-sdk-for-java/pull/11152


sdk/cognitiveservices/ms-azure-cs-computervision/src/main/java/com/microsoft/azure/cognitiveservices/vision/computervision/models/ReadOptionalParameter.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+package com.microsoft.azure.cognitiveservices.vision.computervision.models;
+
+
+/**
+ * The ReadOptionalParameter model.
+ */
+public class ReadOptionalParameter {
+    /**
```
joshfree, line 16: Can this comment (behavior) be called out in the readme or documentation?

anuchandy: Was scanning other autogenerated cognitive SDKs, I think we don't have the Readme process defined for the "autogenerated" data SDK yet, we've it only for track2 manually-authored/convenience SDK layer. This is a good point Josh, generally, there is going to be a scalability issue on scanning through the entire "autogenerated" code and identifying these types of comments and calling them out in readme.

We already have tools to identify breaking changes but behaviors like this i.e. default server chosen value may not be detected by that tool. Maybe the one possible approach for "autogenerated" scenarios is, have the authors call out the important service behaviors in their swagger or swagger-readme and then that get picked by tools for the readme.

anuchandy:
The package is released: https://search.maven.org/artifact/com.microsoft.azure.cognitiveservices/azure-cognitiveservices-computervision/1.0.4-beta/jar

PR #11240, opened 2020-05-18
mgmt spotbugs naming fix

ChenTanyi:


ChenTanyi:
I will update it later. Since the compute has another bug.

PR #11251, opened 2020-05-18
Fix incorrect relativePath entries for data and mgmt parents

JimSuplizio:
As it turns out both data and management parents were incorrect for different reasons.
1. The data parents' relative path was **../../parents/azure-data-sdk-client** which doesn't exist. The correct relative path is ../../parents/azure-data-sdk-parent
2. The management parents' relative path of **../../../pom.management.xml** which is pointing to a file that no longer exists. The relative path needs to be point to the sdk/parents/azure-arm-parent directory however the relative path did vary based upon where the libraries were. The vast majority of entries were changed to "../../parents/azure-arm-parent" however for one mgmt parent and for a couple of libraries under /profiles in the root of the enlistment had slightly different relative paths.

### DIFF
--- a/profiles/2018-03-01-hybrid/pom.xml
+++ b/profiles/2018-03-01-hybrid/pom.xml
@@ -10,7 +10,7 @@
     <groupId>com.microsoft.azure</groupId>
     <artifactId>azure-arm-parent</artifactId>
     <version>1.2.0</version>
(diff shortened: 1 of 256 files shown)


JimSuplizio:
Killing this PR and separating out data and management into different PRs.

Issue #11194, opened 2020-05-14
Add ability to deserialize other Content-Types


conniey:
Currently, if the Content-Type is not application/xml, it will assume the Content-Type is JSON. The Service Bus APIs return "application/atom+xml", which is being incorrectly deserialized as JSON.

SerializerEncoding: https://azuresdkdocs.blob.core.windows.net/$web/java/azure-core/1.3.0-beta.1/com/azure/core/util/serializer/SerializerEncoding.html#fromHeaders-com.azure.core.http.HttpHeaders-

It would be nice to add a way to associate other content types with serializers and treat the HTTP requests and responses according to which ones can be handled.

alzimmermsft:
This problem should be resolvable by enhancing the `Content-Type` to `SerializerEncoding` mapping. Instead of checking static values we should be a little be more flexible by parsing the `Content-Type` header. `application/atom+xml` is using the suffix definition defined in [RFC 3023](https://tools.ietf.org/html/rfc3023), and this should work for other suffix types that we see in AppConfig and other services.

alzimmermsft:
Should be resolved by #11255

PR #11255, opened 2020-05-18
Adding flexibility to parsing serializer encodings.

conniey:
Currently, "application/atom+xml" is deserialized as JSON because it does not match the constant strings we check for XML.  This adds some flexibility to parsing XML and JSON content headers.

Related to #11194 .

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
@@ -6,10 +6,8 @@ package com.azure.core.util.serializer;
 import com.azure.core.http.HttpHeaders;
 import com.azure.core.util.logging.ClientLogger;
 
-import java.util.Comparator;
 import java.util.Map;
 import java.util.TreeMap;
-import java.util.function.Function;
 
 /**
  * Supported serialization encoding formats.


sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,110 @@
...
     public static SerializerEncoding fromHeaders(HttpHeaders headers) {
-        String mimeContentType = headers.getValue("Content-Type");
-        if (mimeContentType != null) {
-            String[] parts = mimeContentType.split(";");
-            if (parts[0].equalsIgnoreCase("application/xml") || parts[0].equalsIgnoreCase("text/xml")) {
-                return XML;
+        final String mimeContentType = headers.getValue(CONTENT_TYPE);
+        if (mimeContentType == null) {
+            LOGGER.warning("'{}' not found. Returning default encoding: {}", CONTENT_TYPE, JSON);
+            return JSON;
```
anuchandy, line 54: I guess you mean: 
```java
return DEFAULT_ENCODING;
```

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,110 @@
...
+                throw new IllegalArgumentException("mimeType is not properly formatted with {type}/{subtype}: "
+                    + rawMimeType);
+            }
+
+            this.type = parts[0];
+            this.subType = parts[1];
+        }
+
+        private boolean matches(String otherType, String otherSubType) {
+            return type.equals(otherType) && subType.equals(otherSubType);
```
anuchandy, line 131: I think MIME types are case insensitive but good to double-check.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,110 @@
      */
     XML;
 
+    private static final ClientLogger LOGGER = new ClientLogger(SerializerEncoding.class);
```
alzimmermsft, line 28: Going to need a Checkstyle suppression for this.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,110 @@
...
+    private static final String CONTENT_TYPE = "Content-Type";
+    private static final List<SerializerEncodingMimeTypes> SUPPORTED_MIME_TYPES;
+    private static final SerializerEncoding DEFAULT_ENCODING = JSON;
+
+    static {
+        SUPPORTED_MIME_TYPES = new ArrayList<>();
+
+        // Encodings and suffixes from: https://tools.ietf.org/html/rfc6838
+        SUPPORTED_MIME_TYPES.add(new SerializerEncodingMimeTypes(XML, "xml",
+            new MimeType("text/xml"), new MimeType("application/xml")));
```
alzimmermsft, line 38: I believe we have static properties for these values in `ContentType`.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,110 @@
...
+        SUPPORTED_MIME_TYPES.add(new SerializerEncodingMimeTypes(JSON, "json",
+            new MimeType("application/json")));
```
alzimmermsft, line 40: Should have a static property for this in `ContentType`.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,110 @@
...
+    }
+
+    private static final class MimeType {
+        private final String type;
+        private final String subType;
+
+        private MimeType(String rawMimeType) {
+            final String[] parts = rawMimeType.split("/");
+            if (parts.length != 2) {
+                throw new IllegalArgumentException("mimeType is not properly formatted with {type}/{subtype}: "
```
alzimmermsft, line 122: Will need a Checkstyle suppression for this.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
conniey: Thanks! Good catch.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
conniey: in line 59, I do a .toLowerCase(Locale.ROOT), but I can update this to be case insensitive instead.

anuchandy: ah, I missed that, thanks for the reference.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,110 @@
      */
     XML;
 
+    private static final ClientLogger LOGGER = new ClientLogger(SerializerEncoding.class);
+    private static final String CONTENT_TYPE = "Content-Type";
+    private static final List<SerializerEncodingMimeTypes> SUPPORTED_MIME_TYPES;
+    private static final SerializerEncoding DEFAULT_ENCODING = JSON;
+
+    static {
+        SUPPORTED_MIME_TYPES = new ArrayList<>();
```
alzimmermsft, line 34: Thoughts on making this more simplified using a case insensitive `TreeMap`. Begin by adding in well known Content-Types such as `application/json`, `text/xml`, and `application/xml` (plus additional in the future as we expand), then add in a simple `"json" -> SerializerEncoding.JSON` and same for XML.

During the conversion from `HttpHeaders` to `SerializerEncoding` we use the following checks:

1. Missing ContentType or it is null or empty, log a warning and return JSON.
2. Check if the ContentType exists in the `TreeMap`, if it does return that value.
3. Use string splitting logic, finding the sub-type and if it exists the `+` split.

This would simplify out a lot of the class overhead and generally have the same result. Additionally, the map could be updated with the Content-Types that weren't in the well known set to optimize them in future cases when they appear (this runs a risk of having many Content-Types if they get randomized a lot but I think that would be a bad design).

alzimmermsft: Another option instead of `TreeMap` is `ConcurrentSkipListMap` which also has the option for a key comparator.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,110 @@
...
+            LOGGER.warning("'{}' not found. Returning default encoding: {}", CONTENT_TYPE, JSON);
+            return JSON;
+        }
+
+        final String[] parts = mimeContentType.split(";");
+
+        final String mimeType = parts[0].toLowerCase(Locale.ROOT);
+        final String[] mimeTypeParts = mimeType.split("/");
+        if (mimeTypeParts.length != 2) {
+            LOGGER.warning("Content-Type '{}' does not match mime-type formatting 'type/'subtype'. "
```
alzimmermsft, line 62: nit: This uses a slightly different error message than the same check in `MimeType`.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,70 @@
...
+    private static final ClientLogger LOGGER = new ClientLogger(SerializerEncoding.class);
+    private static final String CONTENT_TYPE = "Content-Type";
+    private static final Map<String, SerializerEncoding> SUPPORTED_MIME_TYPES;
+    private static final TreeMap<String, SerializerEncoding> SUPPORTED_SUFFIXES;
+    private static final SerializerEncoding DEFAULT_ENCODING = JSON;
+
+
+    static {
+        // Encodings and suffixes from: https://tools.ietf.org/html/rfc6838
+        final Comparator<String> comparator = Comparator.comparing(Function.identity(), String::compareToIgnoreCase);
```
alzimmermsft, line 37: Could use `String.CASE_INSENSITIVE_ORDER`.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,70 @@
      */
     XML;
 
+    private static final ClientLogger LOGGER = new ClientLogger(SerializerEncoding.class);
+    private static final String CONTENT_TYPE = "Content-Type";
+    private static final Map<String, SerializerEncoding> SUPPORTED_MIME_TYPES;
+    private static final TreeMap<String, SerializerEncoding> SUPPORTED_SUFFIXES;
```
alzimmermsft, line 31: Watch out, this could hit transient concurrent modification errors. Mind trying out `ConcurrentSkipListMap`?

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,70 @@
...
+        final String subtype = mimeTypeParts[1];
+        final int lastIndex = subtype.lastIndexOf("+");
+        if (lastIndex == -1) {
+            return DEFAULT_ENCODING;
+        }
+
+        final String mimeTypeSuffix = subtype.substring(lastIndex + 1);
+        final SerializerEncoding serializerEncoding = SUPPORTED_SUFFIXES.get(mimeTypeSuffix);
+        if (serializerEncoding != null) {
+            return serializerEncoding;
```
alzimmermsft, line 86: Should `SUPPORTED_MIME_TYPES` get updated with the `Content-Type`-`SerializerEncoding` value that we deduced.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
```diff
@@ -20,21 +25,70 @@
...
+                + "Returning default: {}", parts[0], DEFAULT_ENCODING);
+            return DEFAULT_ENCODING;
+        }
+
+        // Check the suffix if it does not match the full types.
+        final String subtype = mimeTypeParts[1];
+        final int lastIndex = subtype.lastIndexOf("+");
+        if (lastIndex == -1) {
+            return DEFAULT_ENCODING;
+        }
```
alzimmermsft, line 81: This could be slightly wrong if we run across something like `text/json`, instead we should check:

```java
final String subType = mimeTypeParts[1];
final int lastIndex = subType.lastIndexOf("+");

final String mimeType = (lastIndex == -1) ? subtype : subtype.substring(lastIndex + 1);
```

conniey: text/json isn't a well-supported mimetype though, nor is it a proper suffix.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
conniey: I'm generally not a fan of making this hold some sort of mutable state, especially since this is a static helper method.

sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java
conniey: We never modify this list other than at static construction.

PR #11258, opened 2020-05-18
Update relativePath to azure-data-sdk-parent pom

JimSuplizio:
The data track libraries had the following relativePath to their parent pom: **../../parents/azure-data-sdk-client** which wasn't correct. The correct path to the parent pom should be: **../../parents/azure-data-sdk-parent**

Q. What does this mean? 
A. This means that it would have been pulling the parent pom from Maven. The difference between what is in the tree versus what is in Maven are as follows:
1. The version of com.puppycrawl.tools:checkstyle defined in the pom on Maven is 8.24 instead of 8.29. Though it's effective pom was pulling in 8.29 and only 8.29. Also verified by cleaning my .m2, building and this was the only checkstyle version.
2. There is some surfire plugin configuration that was setting a keyvault URI for testing only in source tree.
3. There were minor changes to remove profiles for the non-shipping modules and the azure-sdk-template libraries in the source. 
4. The modules list at the bottom of the file has since been removed in the source tree.

There's nothing here I found as cause for concern.

### DIFF
--- a/sdk/parents/azure-data-sdk-parent/pom.xml
+++ b/sdk/parents/azure-data-sdk-parent/pom.xml
@@ -325,6 +325,18 @@
             <failOnError>false</failOnError>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <version>3.0.0-M3</version> <!-- {x-version-update;org.apache.maven.plugins:maven-surefire-plugin;external_dependency} -->
+          <configuration>
+            <environmentVariables>
+              <test.mode>${testMode}</test.mode>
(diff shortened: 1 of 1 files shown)


sdk/parents/azure-data-sdk-parent/pom.xml
```diff
@@ -325,6 +325,18 @@
             <failOnError>false</failOnError>
           </configuration>
         </plugin>
+        <plugin>
```
JimSuplizio, line 328: @vcolin7 - So you're aware of why this particular change, in the second commit, was necessary. When this was changed back in April it was moved under the reporting instead of the pluginManagement. The reason this was passing is because when it couldn't find the data parent, it pulled the version from Maven and because that parent path was wrong it pulled the azure-sdk-parent from maven which had the correct entry.

vcolin7: Thanks for the explanation Jim! I was wondering why the my IDE would show the path all in red.

Issue #11261, opened 2020-05-18
CosmosResponseDiagnostics -> CosmosDiagnostics


kushagraThapar:
* Rename `CosmosResponseDiagnostics` -> `CosmosDiagnostics`
* Rename all public APIs to reflect this change.

kushagraThapar:
This is done.

Issue #11260, opened 2020-05-18
CosmosAsyncStoredProcedure / CosmosStoredProcedure APIs and Docs


kushagraThapar:
* Change Object[] to List of Objects in `execute()` API

kushagraThapar:
This is done.

PR #11262, opened 2020-05-18
Misc fixes for cosmos public surface APIs

kushagraThapar:
* Rename CosmosResponseDiagnostics -> CosmosDiagnostics : https://github.com/Azure/azure-sdk-for-java/issues/11261
* Fixed Stored Proc execute() API: https://github.com/Azure/azure-sdk-for-java/issues/11260
* Rename API - connectionSharingAcrossClientsEnabled
* Added @Warning annotation to *BridgeInternal classes
* Rename getRequestLatency() -> getDuration()
* Hide sessionCapturingOverrideEnabled API CosmosClientBuilder
* Hide setConsistencyLevel / getConsistencyLevel from public surface *RequestOptions
* Removed jackson dependency from pom.xml - taking from azure-core now.

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
@@ -46,175 +46,177 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 
+import static com.azure.cosmos.implementation.Warning.INTERNAL_USE_ONLY_WARNING;
+
 /**
  * DO NOT USE.
  * This is meant to be used only internally as a bridge access to classes in
  * com.azure.cosmos
  **/
-@Warning(value = "Internal use only, not meant for public usage")
+@Warning(value = INTERNAL_USE_ONLY_WARNING)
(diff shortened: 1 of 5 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/BridgeInternal.java
```diff
@@ -55,257 +51,220 @@
  * This is meant to be used only internally as a bridge access to classes in
  * com.azure.cosmos
  **/
+@Warning(value = "Internal use only, not meant for public usage")
```
moderakh, line 54: maybe we should mention that these may change in future.

moderakh, line 54: you might be able to define a static final warning String message as string and reuse, rather than repeating the warning on each method.

kushagraThapar: Good point, will do that for sure.

kushagraThapar: Done.

PR #11264, opened 2020-05-18
[TA] Deleted package private Transform class

mssfang:
- HttpResponseException include TextAnalyticsError
- Removed Transform class but add all of its methods to Utility class in the implementation pacakge.

### DIFF
--- a/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
+++ b/eng/code-quality-reports/src/main/resources/checkstyle/checkstyle-suppressions.xml
@@ -262,10 +262,6 @@
             files="com.azure.identity.implementation.WindowsCredentialApi.java"/>
 
   <!-- Use the logger in a Utility static method. -->
-  <suppress checks="com.azure.tools.checkstyle.checks.GoodLoggingCheck"
-            files="com.azure.ai.textanalytics.Transforms.java"/>
-
-  <!-- Use the logger in a Utility static method. -->
   <suppress checks="com.azure.tools.checkstyle.checks.GoodLoggingCheck" files="com.azure.ai.formrecognizer.Transforms.java" />
   <suppress checks="com.azure.tools.checkstyle.checks.GoodLoggingCheck" files="com.azure.ai.formrecognizer.training.CustomModelTransforms.java" />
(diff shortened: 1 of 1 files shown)


sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/implementation/Utility.java
```diff
@@ -94,8 +112,120 @@ public HttpHeaders getHeaders() {
     public static Throwable mapToHttpResponseExceptionIfExist(Throwable throwable) {
         if (throwable instanceof TextAnalyticsErrorException) {
             TextAnalyticsErrorException errorException = (TextAnalyticsErrorException) throwable;
-            return new HttpResponseException(errorException.getMessage(), errorException.getResponse());
+            return new HttpResponseException(errorException.getMessage(), errorException.getResponse(),
+                toTextAnalyticsError(errorException.getValue()));
```
samvaity, line 116: @srnagar Is this the same thing that management SDK's are doing too?
And you had some opinions to have a new model type for the exception rather than doing it this way?

mssfang: as discussion in the TA meeting, this change looks good.

samvaity: In `toTextAnalyticsError` .NET is using the top-level error to form the exception, but here we are using the innerError value if present.

mssfang: This is what python used to use innerError 

https://github.com/Azure/azure-sdk-for-python/pull/11155/files#diff-6814e3aaeba6362738bfde592d65bd9bR64

sdk/textanalytics/azure-ai-textanalytics/src/main/java/com/azure/ai/textanalytics/Transforms.java
```diff
@@ -1,134 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
samvaity, line 1: Just a suggestion, I would have kept the Transforms class as it explains more about the service transform layer and Utility is rather about widely generic helper functions used in the client.

Issue #11192, opened 2020-05-14
IncludedPath / ExcludedPath API and Docs


kushagraThapar:
IncludedPath -> 
* Remove constructor which takes `jsonString`
* Add constructor which takes `String path` - and call logic of `setPath(path)` in the constructor
* Change `public Collection<Index> getIndexes() {` -> `public List<Index> getIndexes() {`

ExcludedPath -> 
* Add constructor which takes `String path` - and call logic of `setPath(path)` in the constructor

kushagraThapar:
This is done.

PR #11265, opened 2020-05-18
Users/xinlian/included path and exclude path update

xinlian12:
This PR solve https://github.com/Azure/azure-sdk-for-java/issues/11192

IncludedPath ->

Remove constructor which takes jsonString
Add constructor which takes String path - and call logic of setPath(path) in the constructor
Change public Collection<Index> getIndexes() { -> public List<Index> getIndexes() 

ExcludedPath ->
Add constructor which takes String path - and call logic of setPath(path) in the constructor

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/ExcludedPath.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/ExcludedPath.java
@@ -15,18 +15,11 @@ public final class ExcludedPath {
 
     /**
      * Constructor.
-     */
-    public ExcludedPath() {
-        this.jsonSerializable = new JsonSerializable();
-    }
-
-    /**
-     * Constructor.
      *
      * @param path the excluded path.
      */
     public ExcludedPath(String path) {
-        this();
+        this.jsonSerializable = new JsonSerializable();
         this.setPath(path);
     }
 
--- a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/models/IncludedPathTest.java
+++ b/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/models/IncludedPathTest.java
@@ -2,11 +2,12 @@
 // Licensed under the MIT License.
 package com.azure.cosmos.models;
 
-import org.assertj.core.api.Assertions;
 import org.testng.annotations.Test;
 
 import java.util.ArrayList;
 import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class IncludedPathTest {
 
@@ -21,12 +22,12 @@ public class IncludedPathTest {
         includedPath.setIndexes(indexes);
 
         List<Index> includedPathIndexes = new ArrayList<>(includedPath.getIndexes());
-        Assertions.assertThat(includedPathIndexes).hasSize(2);
-        Assertions.assertThat(((RangeIndex) includedPathIndexes.get(0)).getDataType()).isEqualTo(DataType.STRING);
-        Assertions.assertThat(((RangeIndex) includedPathIndexes.get(0)).getPrecision()).isEqualTo(-1);
-        Assertions.assertThat(((RangeIndex) includedPathIndexes.get(1)).getDataType()).isEqualTo(DataType.NUMBER);
-        Assertions.assertThat(((RangeIndex) includedPathIndexes.get(1)).getPrecision()).isEqualTo(-1);
+        assertThat(includedPathIndexes).hasSize(2);
+        assertThat(((RangeIndex) includedPathIndexes.get(0)).getDataType()).isEqualTo(DataType.STRING);
+        assertThat(((RangeIndex) includedPathIndexes.get(0)).getPrecision()).isEqualTo(-1);
+        assertThat(((RangeIndex) includedPathIndexes.get(1)).getDataType()).isEqualTo(DataType.NUMBER);
+        assertThat(((RangeIndex) includedPathIndexes.get(1)).getPrecision()).isEqualTo(-1);
 
-        Assertions.assertThat(includedPath.getPath()).isEqualTo(path);
+        assertThat(includedPath.getPath()).isEqualTo(path);
     }
 }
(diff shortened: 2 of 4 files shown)


sdk/cosmos/azure-cosmos-benchmark/src/test/java/com/azure/cosmos/benchmark/ReadMyWritesConsistencyTest.java
```diff
@@ -168,7 +168,7 @@ DocumentCollection getCollectionDefinitionWithRangeRangeIndex() {
         List<IncludedPath> includedPaths = new ArrayList<>();
         IncludedPath includedPath = new IncludedPath();
         includedPath.setPath("/*");
-        Collection<Index> indexes = new ArrayList<>();
+        List<Index> indexes = new ArrayList<>();
         Index stringIndex = Index.range(DataType.STRING);
         BridgeInternal.setProperty(ModelBridgeInternal.getJsonSerializable(stringIndex), "precision", -1);
```
moderakh, line 173: can this be updated to use the public apis for setting precision.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/models/IncludedPathTest.java
```diff
@@ -1,40 +0,0 @@
...
-
-import java.util.ArrayList;
-import java.util.List;
-
-import static org.assertj.core.api.Assertions.assertThat;
-
-public class IncludedPathTest {
-
-    @Test(groups = {"unit"})
-    public void deserialize() {
```
moderakh, line 16: now that custructor for parsing json is removed, could you please add a test which sets different fields using direct api invocation?

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/ExcludedPath.java
```diff
@@ -23,10 +23,11 @@ public ExcludedPath() {
     /**
      * Constructor.
      *
-     * @param jsonString the json string that represents the excluded path.
+     * @param path the excluded path.
      */
-    ExcludedPath(String jsonString) {
-        this.jsonSerializable = new JsonSerializable(jsonString);
+    public ExcludedPath(String path) {
```
simplynaveen20, line 28: Now we have this constructor , what is the benefit of having default constructor and setPath() method explicitly .

sdk/cosmos/azure-cosmos-benchmark/src/test/java/com/azure/cosmos/benchmark/ReadMyWritesConsistencyTest.java
xinlian12: Sure can, has changed across all the places.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/models/IncludedPathTest.java
xinlian12: added

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/ExcludedPath.java
xinlian12: Good question, I am not so sure.

I think if path is required for includedPath, and we do not have any default path value, then should remove the default constructor.  (I guess same question for excludedPath).

@moderakh @kushagraThapar Any thoughts for this one?

kushagraThapar: Good point, I say we remove default constructor, and just keep the constructor which takes `String path`

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/models/IncludedPathTest.java
```diff
@@ -8,33 +8,25 @@
 import java.util.ArrayList;
 import java.util.List;
 
-import static org.assertj.core.api.Assertions.assertThat;
```
moderakh, line 11: please add back static import so we don't repeat Assertions.

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/models/IncludedPathTest.java
```diff
@@ -8,33 +8,25 @@
...
+        String path = "/*";
+        IncludedPath includedPath = new IncludedPath(path);
+
+        List<Index> indexes = new ArrayList<>();
+        indexes.add(Index.range(DataType.STRING, -1));
+        indexes.add(Index.range(DataType.NUMBER, -1));
+        includedPath.setIndexes(indexes);
+
+        List<Index> includedPathIndexes = new ArrayList<>(includedPath.getIndexes());
+        Assertions.assertThat(includedPathIndexes).hasSize(2);
```
moderakh, line 24: please have static import so we don't repeat Assertions.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/models/ExcludedPath.java
moderakh: what about setPath()? probably that shouldn't be public anymore now as path is set in the constructor.

kushagraThapar: We can leave the `setPath()` API as it is, doesn't harm to have the API - and will be in sync with `setIndexes()` API.

xinlian12: Removed default constructor from includedPath and excludedPath

sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/models/IncludedPathTest.java
xinlian12: added static import

PR #11271, opened 2020-05-19
Updated netty versions to 4.1.50.Final

kushagraThapar:
Vulnerabilities Reported:
 
* https://vuln.whitesourcesoftware.com/vulnerability/CVE-2020-11612/
 
* https://vuln.whitesourcesoftware.com/vulnerability/CVE-2019-20445/
 
* https://vuln.whitesourcesoftware.com/vulnerability/CVE-2019-20444/

### DIFF
--- a/sdk/cosmos/tests.yml
+++ b/sdk/cosmos/tests.yml
@@ -54,6 +54,71 @@ jobs:
 
       TestResultsFiles: '**/junitreports/TEST-*.xml'
 
+
+  # This template parllelizes tests that could be run on Single_Region_Session
+  # by running them against a different cosmosdb instance
+  - template: ../../eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+    parameters:
+      TestName: Single_Region_Session_2
+      ServiceDirectory: cosmos
+      TimeoutInMinutes: 240
+      EnvVars:
+        ACCOUNT_HOST: $(java-cosmos-session-singleregion-one-host)
+        ACCOUNT_KEY: $(java-cosmos-session-singleregion-one-key)
+        SECONDARY_ACCOUNT_KEY: $(java-cosmos-session-singleregion-one-secondary-key)
+        DESIRED_CONSISTENCIES: '["Session"]'
+        PROTOCOLS: '["Tcp"]'
+      MaxParallel: 1
+      Matrix:
+        # 09m 07s, expect passed
+        Fast:
+          DisplayName: Multimaster Single Region Fast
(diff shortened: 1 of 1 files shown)


sdk/cosmos/tests.yml
```diff
@@ -55,71 +54,6 @@ jobs:
 
       TestResultsFiles: '**/junitreports/TEST-*.xml'
 
-
-  # This template parllelizes tests that could be run on Single_Region_Session
-  # by running them against a different cosmosdb instance
-  - template: ../../eng/pipelines/templates/jobs/archetype-sdk-tests.yml
-    parameters:
-      TestName: Single_Region_Session_2
```
moderakh, line 63: you seem to be removing our integration tests. are the tests not working for v3?

kushagraThapar: They are: 
* just clubbed some of different CIs to single CI with both Https and Tcp protocol. 
* removed some of the extra ones which we don't need every single time for every change. 
* We might have to do somewhat similar with master tests as well.

sdk/cosmos/tests.yml
```diff
@@ -22,7 +22,6 @@ jobs:
           DisplayName: ReadMyWrites Integration Tests
           OSVmImage: 'ubuntu-18.04'
           ProfileFlag: '-Pe2e'
-          AdditionalArgs: '-DargLine="-Dcosmos.directModeProtocol=Https"'
```
moderakh, line 25: why these ones are removed? is that intentional?

sdk/cosmos/tests.yml
```diff
@@ -190,7 +189,6 @@ jobs:
           DisplayName: Single Region ReadMyWrites
           OSVmImage: 'ubuntu-18.04'
           ProfileFlag: '-Pe2e'
-          AdditionalArgs: '-DargLine="-Dcosmos.directModeProtocol=Https"'
```
moderakh, line 193: ditto

sdk/cosmos/tests.yml
kushagraThapar: Yes, currently on V3-  we are not running e2e tests on Tcp mode, which is important as compared to running e2e tests on https mode.

sdk/cosmos/tests.yml
kushagraThapar: Yes, currently on V3- we are not running e2e tests on Tcp mode, which is important as compared to running e2e tests on https mode.

PR #11272, opened 2020-05-19
Updates API based on feedback

conniey:
* Removes by default auto-lock renewal and auto-completion
* Moves auto-lock renewal into builder
* Renames peek -> browse
* Fixes snippets and docs
* Adds integration tags to run subset of tests (`-Dgroups=integration` or `-Dgroups=!integration`) without reliance on environment variables.
* Fixes bug in completion.

### DIFF
--- a/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
+++ b/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
@@ -555,7 +555,7 @@ public final class ServiceBusClientBuilder {
         }
 
         /**
-         * Enables auto-lock renewal by renewing session lock renewal until the {@code maxAutoLockRenewalDuration} has
+         * Enables auto-lock renewal by renewing each session lock until the {@code maxAutoLockRenewalDuration} has
          * elapsed.
          *
          * @param maxAutoLockRenewalDuration Maximum amount of time to renew the session lock.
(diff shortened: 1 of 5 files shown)


sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
```diff
@@ -542,16 +549,29 @@ public ServiceBusSenderClient buildClient() {
...
         private String subscriptionName;
         private String topicName;
+        private Duration maxAutoLockRenewalDuration;
 
         private ServiceBusSessionReceiverClientBuilder() {
         }
 
+        /**
+         * Enables auto-lock renewal by renewing session lock renewal until the {@code maxAutoLockRenewalDuration} has
+         * elapsed.
```
hemanttanwar, line 559: Explain little bit about `If there are multiple session` This would apply on each session separately.

sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusClientBuilder.java
```diff
@@ -734,10 +760,23 @@ private boolean isRollingSessionReceiver() {
...
 
         private ServiceBusReceiverClientBuilder() {
         }
 
+        /**
+         * Enables auto-lock renewal by renewing session lock renewal until the {@code maxAutoLockRenewalDuration} has
+         * elapsed.
+         *
+         * @param maxAutoLockRenewalDuration Maximum amount of time to renew the session lock.
+         * @return The modified {@link ServiceBusReceiverClientBuilder} object.
```
hemanttanwar, line 773: This documentation should be for non session receiver because this builder belongs to non-session entities.

PR #11300, opened 2020-05-19
Add multi tenant end-to-end tests

jianghaolu:


sdk/e2e/src/main/java/com/azure/endtoend/identity/MultiTenantTest.java
```diff
@@ -0,0 +1,164 @@
...
+            "testUserPasswordCanAccessGraph - AZURE_USER_NAME not configured in the environment.");
+        assertConfigPresence(AZURE_USER_PASSWORD,
+            "testUserPasswordCanAccessGraph - AZURE_USER_PASSWORD not configured in the environment.");
+        String clientId = CONFIGURATION.get(AZURE_CLIENT_ID);
+        String tenantId = CONFIGURATION.get(AZURE_TENANT_ID);
+        String username = CONFIGURATION.get(AZURE_USER_NAME);
+        String password = CONFIGURATION.get(AZURE_USER_PASSWORD);
+
+        UsernamePasswordCredential credential = new UsernamePasswordCredentialBuilder()
+            .clientId(clientId)
```
g2vinay, line 76: so, your instructions cover to disable 2FA for the username ?

jianghaolu: This has to be done in a separate tenant - cannot be done from the Microsoft corporate tenant. By default 2FA is turned off in any new tenant.

jianghaolu:
The instructions are at https://github.com/Azure/azure-sdk/issues/1346

sdk/e2e/pom.xml
```diff
@@ -55,6 +55,11 @@
       <artifactId>slf4j-api</artifactId>
       <version>1.7.28</version> <!-- {x-version-update;org.slf4j:slf4j-api;external_dependency} -->
     </dependency>
+    <dependency>
+      <groupId>com.microsoft.azure</groupId>
+      <artifactId>azure-mgmt-graph-rbac</artifactId>
```
weshaggard, line 60: You need to add a version tag here.

weshaggard:
I see you were trying to run "java - core - ci" unfortunately unless you have changes under core it DevOps will not let you trigger that (it is a limitation that I hate). However you can run the "java - core" from any PR to do a core build validation.

weshaggard:
@jianghaolu I don't have much context on e2e project but it might be worth adding a pipeline for it or adding it to the include path for the core ci.yml file so a pipeline is triggered and it is built when it changes.

sdk/e2e/pom.xml
jianghaolu: Thanks - is there a reason why no pipeline job is running on this PR?

weshaggard: This particular path doesn't have a pipeline setup for it. I suggest either adding one or adding it to run the core pipeline.

PR #11302, opened 2020-05-19
mgmt update default subscription logic

xseeseesee:
This PR is to update how we handle default subscription when user didn't provide subscription info.
The main logic:
- fetch list of subscriptions from the user's tenant
  - If size is 0, throw exception with link to ask user to create a subscription first.
  - If size is larger than 1, throw exception with the names & ids of subscriptions to ask user which one is desired to use. They can set in Azure profile or use withSubscription().
  - If there is only 1 subscription found, we take it as default subscription and go to next step.

### DIFF
--- a/sdk/management/azure/src/main/java/com/azure/management/Azure.java
+++ b/sdk/management/azure/src/main/java/com/azure/management/Azure.java
@@ -327,7 +327,7 @@ public final class Azure {
         @Override
         public Azure withDefaultSubscription() {
             if (profile.subscriptionId() == null) {
-                profile.withSubscriptionId(Utils.subscriptionId(this.subscriptions().list()));
+                profile.withSubscriptionId(Utils.defaultSubscription(this.subscriptions().list()));
             }
             return new Azure(httpPipeline, profile, this);
         }
--- a/sdk/resources/mgmt/src/main/java/com/azure/management/resources/implementation/ResourceManager.java
+++ b/sdk/resources/mgmt/src/main/java/com/azure/management/resources/implementation/ResourceManager.java
@@ -193,7 +193,7 @@ public final class ResourceManager extends ManagerBase implements HasInner<Resou
         @Override
         public ResourceManager withDefaultSubscription() {
             if (profile.subscriptionId() == null) {
-                profile.withSubscriptionId(Utils.subscriptionId(this.subscriptions().list()));
+                profile.withSubscriptionId(Utils.defaultSubscription(this.subscriptions().list()));
             }
             return new ResourceManager(httpPipeline, profile, sdkContext);
         }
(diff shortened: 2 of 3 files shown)


sdk/management/azure/src/main/java/com/azure/management/Azure.java
```diff
@@ -325,8 +331,25 @@ public Azure withSubscription(String subscriptionId) {
...
             if (profile.subscriptionId() == null) {
-                throw logger.logExceptionAsError(
-                    new IllegalArgumentException("Please specify the subscription ID for resource management."));
+                List<Subscription> subscriptions = new ArrayList<>();
+                this.subscriptions().list().forEach(subscription -> {
+                    subscriptions.add(subscription);
+                });
+                if (subscriptions.size() == 0) {
+                    throw logger.logExceptionAsError(
+                        new RuntimeException("Please create a subscription before you start resource management. "
```
weidongxu-microsoft, line 340: Is it possible to find a more specific exception type? RuntimeException is generally only viewed as abstract base class. Maybe IllegalStateException, or InvalidArgumentException (if we take that defaultSubscription only works for 1 subscription).

sdk/resources/mgmt/src/main/java/com/azure/management/resources/implementation/ResourceManager.java
```diff
@@ -192,8 +198,25 @@ public ResourceManager withSubscription(String subscriptionId) {
         @Override
         public ResourceManager withDefaultSubscription() {
             if (profile.subscriptionId() == null) {
-                throw logger.logExceptionAsError(
-                    new IllegalArgumentException("Please specify the subscription ID for resource management."));
+                List<Subscription> subscriptions = new ArrayList<>();
+                this.subscriptions().list().forEach(subscription -> {
+                    subscriptions.add(subscription);
+                });
+                if (subscriptions.size() == 0) {
```
weidongxu-microsoft, line 205: Is it possible to share some code?

sdk/management/azure/src/main/java/com/azure/management/Azure.java
xccc-msft: Two candidates:
- `IllegalStateException` : 
> Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.
- `UnsupportedOperationException` : 
> Thrown to indicate that the requested operation is not supported.

Maybe UnsupportedOperationException is better? The `withDefaultSubscription()` is not supported when no subscription found or more than one subscription found.

weidongxu-microsoft: `UnsupportedOperationException` is not good here. It is most commonly used in optional operation, which can be left unimplemented at all (like https://docs.oracle.com/javase/7/docs/api/java/util/Collection.html#add(E)).

So I would vote `IllegalStateException`.

xccc-msft: Got it. It's not the same as `InvalidOperationException` in csharp. So it's like the `withDefaultSubscription()` cannot be done at this time due to some restrictions.

Issue #11257, opened 2020-05-18
Wrap FeedResponseDiagnostics in CosmosDiagnostics


kushagraThapar:
* Wrap FeedResponseDiagnostics in CosmosDiagnostics
* Expose CosmosDiagnostics through FeedResponse API.

PR #11309, opened 2020-05-19
Wrapping FeedDiagnostic into CosmosDiagnostic

simplynaveen20:
To be in sync with point operation moving feedDiagnostics into cosmosDiagnostics.
Both query and point response will call getCosmosDiagnostics()

closes https://github.com/Azure/azure-sdk-for-java/issues/11257

### DIFF
--- a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosDiagnostics.java
+++ b/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosDiagnostics.java
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 package com.azure.cosmos;
 
+import com.azure.cosmos.implementation.FeedResponseDiagnostics;
 import com.azure.cosmos.implementation.apachecommons.lang.StringUtils;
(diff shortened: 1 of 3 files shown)


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/CosmosDiagnostics.java
```diff
@@ -39,6 +44,10 @@ CosmosDiagnostics clientSideRequestStatistics(ClientSideRequestStatistics client
      */
     @Override
     public String toString() {
+        if (this.feedResponseDiagnostics != null) {
+            return feedResponseDiagnostics.toString();
```
kushagraThapar, line 48: Do we not have client side request statistics for query operations ?

kushagraThapar: We should also move FeedResponseDiagnostics to implementation package.

simplynaveen20: > Do we not have client side request statistics for query operations ?

It is inside QueryMetrics type

PR #11311, opened 2020-05-19
CosmosClientException update

xinlian12:
This PR solves issue https://github.com/Azure/azure-sdk-for-java/pull/11311#issue-420341824

1. Rename CosmosClientException to CosmosException
2. Remove throws CosmosClientException from APIs.
3. Move CosmosError move to implementation
4. Change getError() of CosmosException to package private


sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/JsonSerializable.java
```diff
@@ -699,7 +698,6 @@ public ObjectNode getPropertyBag() {
             || SqlQuerySpec.class.equals(c)
             || UniqueKey.class.equals(c)
             || UniqueKeyPolicy.class.equals(c)
-            || Index.class.isAssignableFrom(c)
-            || CosmosError.class.equals(c);
+            || Index.class.isAssignableFrom(c);
```
moderakh, line 701: why CosmosError removed from here.

sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/CosmosError.java
```diff
@@ -68,7 +65,7 @@ public CosmosError(String errorCode, String message, String additionalErrorInfo)
      * @return the error code.
      */
     public String getCode() {
-        return this.jsonSerializable.getString(Constants.Properties.CODE);
+        return this.getString(Constants.Properties.CODE);
```
kushagraThapar, line 68: Though it will not make a difference for this class, but the correct implementation should be `super.set / get`
Instead of `this.getString()`, all of the APIs inside this class should be calling `super.getString()`
Here and below for other APIs - should be -> `super.get*()` and `super.set*()`

xinlian12: changed to use super.

PR #11328, opened 2020-05-20
update core mgmt

weidongxu-microsoft:


sdk/management/pom.xml
```diff
@@ -62,9 +62,9 @@
                 <version>1.5.0</version>
             </dependency>
             <dependency>
-                <groupId>com.azure</groupId>
+                <groupId>com.azure</groupId>AzureServiceClient
```
ChenTanyi, line 65: is it mistake?

weidongxu-microsoft: Yes... Fixed.

